PR #238, opened 2014-03-21
Support markdown in description

techtonik:
Compare: https://bitbucket.org/techtonik/wikify/  
And: https://pypi-preview.a.ssl.fastly.net/project/wikify/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ install_requires = [
     "guard",
     "html5lib",
     "Jinja2",
+    "markdown",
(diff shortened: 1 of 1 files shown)


shinalone:
I also think it is popular to use with markdown, not reStructuredText. 
For example, Github. 
The [reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickref.html) makes me **depressed** !

di:
@shinalone See #869, this is in progress.

PR #294, opened 2014-04-15
Adds plugin to flake8 to make sure flask.session is not abused. 

abadger:
Add onto @wickman's work to build a flake8 plugin that checks for flask.session abuse.  wickman's branch didn't pass in travis due to not checking all of the code branches.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,6 +94,9 @@ setup(
         "console_scripts": [
             "warehouse = warehouse.__main__:main",
         ],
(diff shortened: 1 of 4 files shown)


dstufft:
I'm going to close this, I've started porting Warehouse over to Pyramid to ease the concerns of the custom framework-ish thing I've got going on here. That work is on the pyramid branch inside this repository. Since we're moving to something other than Flask, this PR no longer makes sense.

PR #332, opened 2014-07-30
Update contributions.rst and add instructions for non-Python dependencies.

vladimir-v-diaz:
Include installation instruction for non-Python dependencies so newcomers to the project have an easier time setting up their environment.  Currently includes installation instructions for Ubuntu 14.04 / Debian-based environment.

* Fix installation instruction for Grunt.  The installation command was incomplete.
* Mention the Python3.4 and libffi header files required to install Warehouse.
* Add non-python-dependencies.rst, which currently contains instructions for installing / setting up non-Python dependencies in Ubuntu 14.04 or Debian-based environments.
* Link to document containing the installation instructions of non-Python dependencies.
* List the document containing non-Python dependencies installation instructions in the Table of Contents so that Warehouse documentation can be successfully generated by tox.


msabramo:
See PR #356, which makes it possible to package Warehouse as a Docker container and that takes care of Postgres and everything else automatically.

dstufft:
I'm not going to close this, but I've started porting Warehouse over to Pyramid to ease the concerns of the custom framework-ish thing I've got going on here. That work is on the pyramid branch inside this repository. That includes a development environment using Docker contains via ``fig.sh``. This takes care of a good bit of this work, however not all of it so I want to keep this open so we can port over the relevant bits still.

dstufft:
I'm actually going to close this now. Basically everything i think is encapsulated within [Docker Compose](https://docs.docker.com/compose/) now, and if anyone has any problems getting setup with the new stack we can revisit adding new installation instructions then.

PR #344, opened 2014-12-17
Give command to start postgres.

Carreau:
For us poor mortal who are not postgres fluent, to avoid searching everyday.

----- 

I suppose we should do it for redis too, but I'm not there yet.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -64,9 +64,9 @@ dependencies, and the Python development dependencies using:
 
 Before setting up the project you need to have a postgres server running.
 
-If you do not want postgres to run as a deamon you can manually launch the postgres
-server in a separate terminal with the following command (exact to path database
-path may vary depending on system):
+If you do not want postgres to run as a daemon you can manually launch the postgres
+server in a separate terminal with the following command (exact path to database may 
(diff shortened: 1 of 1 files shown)


msabramo:
See PR #356, which makes it possible to package Warehouse as a Docker container and that takes care of Postgres and everything else automatically. Even if not using Docker, you could use the `Procfile` in that PR along with [honcho](https://honcho.readthedocs.org/en/latest/) or [foreman](http://ddollar.github.io/foreman/) to manage all the processes.

dstufft:
I'm going to close this, I've started porting Warehouse over to Pyramid to ease the concerns of the custom framework-ish thing I've got going on here. That work is on the pyramid branch inside this repository. That includes using Docker via [fig](http://fig.sh/) to spin up containers for Redis, PostgreSQL, and Warehouse itself using the single command ``fig up``.

msabramo:
You might also be interested in https://github.com/msabramo/pydockerize, which is an experiment I've been doing that kind of wraps Docker and fig with some smartness to make it super easy.

Though if you already have the `Dockerfile` and `fig.yml`, it may not bring much value.

PR #354, opened 2015-01-26
Fix Travis failures (Set default ES host of 127.0.0.1:9200)

msabramo:
to prevent a bunch of test failures when doing `tox -e py34`.

Before:

    6 failed, 287 passed, 124 error in 6.70 seconds

After:

    331 passed, 86 error in 6.59 seconds

### DIFF
--- a/warehouse/search/indexes.py
+++ b/warehouse/search/indexes.py
@@ -27,7 +27,7 @@ class Index(object):
         self.db = db
         self.config = config
         self.es = Elasticsearch(
-            hosts=self.config.hosts,
+            hosts=self.config.hosts or ['127.0.0.1:9200'],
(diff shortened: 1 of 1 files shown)


msabramo:
and after fixing postgres on my machine and installing py-bcrypt:

```
417 passed in 19.60 seconds
```

msabramo:
Bump

dstufft:
I'm going to close this, I've started porting Warehouse over to Pyramid to ease the concerns of the custom framework-ish thing I've got going on here. That work is on the pyramid branch inside this repository. That doesn't currently rely on Elasticsearch (though it will in the future), but when it does we'll make sure that the tests handle it.

PR #352, opened 2015-01-24
.travis.yml: Add elasticsearch

msabramo:
per http://docs.travis-ci.com/user/database-setup/#ElasticSearch

PR #356, opened 2015-01-28
Docker

msabramo:
Changes to make it possible to package up a warehouse development environment with Docker. This makes it much easier to get started with Warehouse development, because it bundles PostgreSQL, Redis, Elasticsearch and it has pre-built cffi and such.

I have a pre-built image at https://registry.hub.docker.com/u/msabramo/warehouse/

So one could run warehouse by doing:

- `docker pull msabramo/warehouse`
- `docker run -i -t -P msabramo/warehouse`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,60 @@
+# --------------------------------------------------------------------------
+# This is a Dockerfile to build an Ubuntu 14.04 Docker image with
+# Warehouse
+#
+# Use a command like:
+#     docker build -t <user>/warehouse .
+# --------------------------------------------------------------------------
+
+FROM  orchardup/python:3.4
+MAINTAINER  Marc Abramowitz <marc@marc-abramowitz.com> (@msabramo)
+
+ENV DEBIAN_FRONTEND noninteractive
+ENV LC_ALL C
+
+RUN apt-get update && apt-get install -y software-properties-common && add-apt-repository ppa:chris-lea/node.js
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    git \
+    libffi-dev \
+    libpq-dev \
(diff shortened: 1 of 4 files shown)


msabramo:
It would be nice to have a data dump, as outlined in https://github.com/pypa/warehouse/issues/343, so the Docker container could come with some data and be a better demo.

msabramo:
I realized that I used an old version of Elasticsearch; I'm going to try to update it to the newest version.

msabramo:
Updated to use Elasticsearch 1.4.2

msabramo:
Test failures are bogus and I think that they are resolved by #352 and #354.

alex:
Running everythign in the same docker container doesn't seem like the right approach to me, it seems like you want to use something like [fig](http://www.fig.sh/) for this.

msabramo:
Yeah, it's been a while since I've looked at fig. Maybe I'll look again.

I thought that everything in one Docker container would be the simplest thing to get a dev environment going. I definitely wasn't thinking that it would be used in production like this.

Though, IIRC, fig also targets dev environments so maybe I should look at it again.

dstufft:
I'm going to close this, I've started porting Warehouse over to Pyramid to ease the concerns of the custom framework-ish thing I've got going on here. That work is on the pyramid branch inside this repository. As part of that I've set it up so that a development environment can be spun up inside of Docker containers using [fig](http://fig.sh/).

msabramo:
Check out [Docker Compose](https://blog.docker.com/2015/02/announcing-docker-compose/) -- it's the next generation of fig.

dstufft:
Open an issue to look at switching to that please!

msabramo:
https://github.com/pypa/warehouse/issues/361

PR #359, opened 2015-02-12
Use a tuple for __all__ as the good PEP tells us

sigmavirus24:


alex:
The PEP seems wrong in my opinion, tuples are not sequences, they're records, a list is the appropriate data structure for this.

dstufft:
Honestly a set is probably the right data structure for this ;)

sigmavirus24:
Hm. It seems either PEP 8 was modified to allow lists as well, or my memory is severely faulty.

PR #443, opened 2015-03-18
GNU Terry Pratchett

ncoghlan:
http://www.gnuterrypratchett.com/

### DIFF
--- a/warehouse/config.py
+++ b/warehouse/config.py
@@ -38,6 +38,16 @@ def content_security_policy_tween_factory(handler, registry):
 
     return content_security_policy_tween
 
+# http://www.gnuterrypratchett.com/
+def gnu_terry_pratchett_tween_factory(handler, registry):
+    def gnu_terry_pratchett_tween(request):
+        resp = handler(request)
+        resp.headers["X-Clacks-Overhead"] = \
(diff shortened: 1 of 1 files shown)


dstufft:
The change looks good to me, it's going to fail tests though. You'll need to add a ``pretend.call(...)`` here: https://github.com/pypa/warehouse/blob/master/tests/test_config.py#L178-L180 and you'll need to test the actual tween itself (same file, there's tests for the CSP tween there, should be pretty simple).

dstufft:
Also you'll want to run ``tox -e pep8`` to make sure it passes that.

ncoghlan:
Ah, the we learned so much about writing web services between the dawn of PyPI v1 and the dawn of Warehouse :)

dstufft:
Tests for great justice :D Wouldn't want a regression in the ``X-Clacks-Overhead: GNU Terry Pratchett`` header after all!

dstufft:
It might make sense to just close this since https://github.com/python/psf-salt/pull/41 exists now. I intend for Warehouse to be deployed behind the PSF LBs.

ncoghlan:
Heh, I just came here to close this on that basis :)

PR #453, opened 2015-03-22
Refactor the factory_boy usage to not require an explicit session

### LINKED ISSUE #429: factory_boy SubFactory doesn't pass along session
We don't have a thread local session, so we need to pass along the session when creating a SubFactory, figure out how to do this and remove the places where we create a sub factory object explicitly just to pass it in to another factory.

dstufft:
* Configure a threadlocal session for the test suite.
* Configure our base factory_boy Factory to use the threadlocal
  session.

Fixes #429

### DIFF
--- a/tests/accounts/test_models.py
+++ b/tests/accounts/test_models.py
@@ -29,16 +29,13 @@ class TestUserFactory:
         ],
     )
     def test_traversal_finds(self, db_request, username, normalized):
-        user = DBUserFactory.create(
-            session=db_request.db,
-            username=username,
-        )
+        user = DBUserFactory.create(username=username)
         root = UserFactory(db_request)
 
         assert root[normalized] == user
 
     def test_travel_cant_find(self, db_request):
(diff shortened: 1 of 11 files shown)


berkerpeksag:
Is there a reason not to use FooFactory(...) instead of FooFactory.create(...)?

dstufft:
I wasn't aware that worked tbh.

alex:
Why are thread locals any better in test than they are in code?

dstufft:
@alex A few reasons:

* The sadness is contained to stuff that doesn't happen in production, so "weird" errors in production are still less likely.
* All of the application test code still uses an explicit session, it's only used within factory_boy factories.
* factory_boy itself doesn't work the greatest when you can't stick the session as an attribute on the classes instead of passing it in.

Essentially, I'm not fond of them, but given that factory_boy seems to be written to expect it, it's only the test code (and then only for factory_boy) I'm willing to accept them in trade for factory_boy working.

PR #463, opened 2015-03-26
Fix doc link

Carreau:
warehouse.pypi.io return a 404


dstufft:
Ah, yea that's because I hit a bug with RTD with my custom domain, and I'm waiting for @ericholscher to make it work again :)

I'll go ahead and merge this so that the link is correct in the interim.

Carreau:
Thanks !

PR #467, opened 2015-03-28
Use RFC 5861 stale-while-revalidate and stale-if-error headers

dstufft:
These will trigger Fastly to serve stale responses while revalidating content or if the origin returns an error. This should make things faster and less error prone.


mattrobenolt:
:+1:

PR #476, opened 2015-04-12
Add a system for doing the legacy routing based on ?:action params

dstufft:


warehouse/legacy/action_routing.py
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def pypi_action(action):
+    def predicate(info, request):
+        if action == request.params.get(":action", None):
```
berkerpeksag: If I'm not missing something, this could be written as `return action == request.params.get(":action", None)`.

dstufft: Yea it could be, I don't have a good reason for why it wasn't other than I didn't.

PR #479, opened 2015-04-13
Fix #438.

### LINKED ISSUE #438: Implement a Security page
We need a replacement for https://pypi.python.org/security.

dachshund:
May not be perfect, but I hope it's a good start!


dstufft:
These need to be marked for translation through the ``{{ _('foo %(bar)s', bar=thing) }}`` API, it'll also need tests written.

PR #480, opened 2015-04-13
Warn when project version is not the latest

### LINKED ISSUE #404: Warn if viewing non latest project
Warehouse lets you look at the page for any version of a project, however we should provide some sort of warning if you're looking at something that isn't the latest version of that project.

ghickman:
This fixes #404 

This adds an alert div when the version being viewed isn't the latest one. It includes styles for the base `.alert` class and the `.alert-warning`. I've followed the styles used [here](https://warehouse.python.org/project/bitmath/1.0.8-1/) as they seem sensible.

It looks like this:


warehouse/templates/packaging/detail.html
```diff
@@ -20,6 +20,12 @@
   <header class="detail-header">
     <h1>{{ project.name }} <span class="version">{{ release.version }}</span></h1>
     {% if release.summary %}<p>{{ release.summary }}</p>{% endif %}
+    {% if release.version != all_releases[0].version %}
+    <div class="alert alert-warning">
+      {{ _('The latest version of this package is')
```
dstufft: This probably needs to be something like:

```
{{ _('The latest version of this package is <a href="%(version_link)s">%(version)s</a>', version_link=request.route_path('packaging.project', name=release.name), version=all_releases[0].version) }}
```

I *think* this will enable translators to put the translated content anywhere. You might also need a ``|safe()`` filter on the end of that to make it work.

warehouse/translations/en/LC_MESSAGES/warehouse.po
```diff
@@ -18,50 +18,10 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Generated-By: Babel 1.3\n"
 
-#: warehouse/templates/packaging/detail.html:87
```
dstufft: Something strange happened here... I'm not sure why all of these went missing?

ghickman: I'm not sure what happened here either, rerunning `make extract-translations && make compile-translations` seems to have regenerated the strings and gives a saner diff.

ghickman:
@dstufft – I've updated based on your comments. I couldn't see any difference to the output of `_()` with `|safe()` so left it off.

PR #483, opened 2015-04-13
Add some clarity on how to use _() in templates

ghickman:
This makes the use of `_()` more explicit.

### DIFF
--- a/docs/development/translations.rst
+++ b/docs/development/translations.rst
@@ -53,7 +53,15 @@ In Templates
 
 Inside of Jinja2 templates the standard
 `Jinja2 i18n extension <http://jinja.pocoo.org/docs/dev/extensions/#newstyle-gettext>`_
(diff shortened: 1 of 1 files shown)


docs/development/translations.rst
```diff
@@ -53,7 +53,15 @@ In Templates
 
 Inside of Jinja2 templates the standard
 `Jinja2 i18n extension <http://jinja.pocoo.org/docs/dev/extensions/#newstyle-gettext>`_
-configured with ``newstyle=True``.
+has been configured with ``newstyle=True``.
+
+You can use it like so:
+
+.. code:: python
```
dstufft: This isn't python code, it should probably either be html, nothing, or jinja2 if Sphinx has that.

dstufft:
Other than the comment about the language, this LGTM.

PR #485, opened 2015-04-13
Install dev requirements in tox by referencing requirements.txt

ryanhiebert:
This notably changes the semantics of what is included in the tox environments to use all of the requirements in in requirements.txt, even if they aren't being used. Is that acceptable, @dstufft? If not, then perhaps it would be best to create some other *-requirements.txt files so that we can control it more fine-grained, while still not having the duplication.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,14 +10,7 @@ deps =
     pretend
     pytest
     pytest-dbfixtures>=0.9.0
-    # We need to use a newer version of webob than is available on PyPI.
-    https://github.com/Pylons/webob/archive/master.zip#egg=webob
-    # We need to use a newer version of Pyramid (1.6) which isn't available on
-    # PyPI yet, so we'll install directly from GitHub.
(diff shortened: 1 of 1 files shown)


dstufft:
There's absolutely some duplication here, but long term this should just all get deleted from ``tox.ini`` and from ``dev/requirements.txt`` as things get released and Warehouse no longer depends on forked or development versions of these projects. Currently we're running 4 forked dependencies for Warehouse itself, 2/4 of those I think are planning to at least release an alpha during or shortly after the PyCon sprints, 1 of them I can probably poke the maintainer to see about getting a release, and the 4th one is Jinja2 and I'm not sure how easy it'll be to get that landed and released.

All in all, I'm not sure it's worth it to worry about it for a temporary situation.

ryanhiebert:
That reasoning makes sense. Thanks for taking the to time explain it to me.

dstufft:
No problem! Thank you for the pull request :)

PR #486, opened 2015-04-13
Setup instructions for running tox on Ubuntu without Docker.

dachshund:


docs/development/getting-started.rst
```diff
@@ -96,6 +96,8 @@ be run using tox. To run all of the tests all you have to do is:
 
 .. code-block:: console
 
+
+    $ apt-get install libffi-dev libpq-dev python3-dev postgresql postgresql-contrib
```
dstufft: This should probably have a comment before it saying something like ``# On Debian/Ubuntu systems``, so we can add in additional options for other systems.

PR #487, opened 2015-04-13
csp: add raw.githubusercontent.com for refills

domenkozar:


dstufft:
Instead of adding ``raw.githubusercontent.com`` we should just add anything we need to the ``warehouse/static/`` directory and serve it from there.

PR #488, opened 2015-04-13
 Implement a Security page (fix #438).

### LINKED ISSUE #438: Implement a Security page

trishankkarthik:


warehouse/views.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pyramid.view import view_config
+
+
+@view_config(route_name='security', renderer='security.html')
+def security(request):
+    return {
+        'administrators': [
```
domenkozar: Do we want these taken out of users model?

dachshund: @vladimir-v-diaz tells me the User model has name and email, but not keyid and fingeprint. I'm not sure whether we should add those fields to the model just to accommodate the administrators. What do you think, @domenkozar ?

domenkozar: On the long run we'll want everyone to have public pgp keys (https://keybase.io/ integration comes to mind)

dachshund: Absolutely. Hard-coding sucks, but I think it'll do just fine for the short term.

domenkozar: Agreed - gotta start somewhere :)

dstufft:
I'm going to close this since it doesn't merge cleanly and we're going to need all new content for Warehouse anyways. Thanks for the contribution, sorry it took so long to resolve this.

brainwane:
@dstufft when you say "we're going to need all new content for Warehouse anyways" I want to make sure I understand -- you mean because the security-related data in Warehouse is substantially different from the hard-coded stuff here?

dstufft:
@brainwane Yea, well it was mostly copied verbatim from https://pypi.python.org/security and that information has sufficiently bitrotted to the point that it's almost all wrong.

PR #492, opened 2015-04-14
Add basic index view

domenkozar:
No tests yet, I'll add them once we agree this is useful :)
2015-04-14-155911_2265x1366_scrot


warehouse/packaging/models.py
```diff
@@ -92,6 +93,10 @@ class Project(db.ModelBase):
         nullable=False,
         server_default=sql.func.now(),
     )
+    updated = Column(
+        DateTime(timezone=False),
+        onupdate=datetime.datetime.now
```
dstufft: I'd rather handle this inside the database using a trigger and set this to ``server_onupdate=FetchedValue()`` so that it knows to refetch it.

warehouse/packaging/models.py
```diff
@@ -107,6 +112,13 @@ def __getitem__(self, version):
             raise KeyError from None
 
     @property
+    def latest_release(self):
+        try:
+            return self.releases.order_by('created')[0]
```
dstufft: This should have a ``.limit(1)`` added to it and should use ``.first()`` instead of trying to index the returned list. There is likely a better name for this too I think, "latest" is overloaded, is it the most recently released version or the highest version number available?

warehouse/routes.py
```diff
@@ -36,6 +38,11 @@ def includeme(config):
         traverse="/{name}/{version}",
     )
     config.add_route("packaging.file", "/packages/{path:.*}")
+    config.add_route(
+        "packaging.search",
```
dstufft: I'd rather just leave the search bit out for this PR and leave it focused on the index view, the search box can be left non functional for now.

warehouse/views.py
```diff
@@ -0,0 +1,37 @@
...
+from warehouse.accounts.models import User
+
+
+@view_config(
+    route_name="index",
+    renderer="index.html",
+)
+def index(request):
+    latest_updated_projects = request.db.query(Project)\
+                                        .order_by(Project.updated)[:20]
```
dstufft: Does SQLAlchemy translate a slice notation to a ``.limit(20)``?

dstufft:
Overall looks good, added a few comments though.

warehouse/packaging/models.py
domenkozar: What is the benefit? Sounds to me that using ORM here is simpler and doesn't introduce much performance penalty.

warehouse/packaging/models.py
domenkozar: `[0]` implies `.limit(1)`, I'll add a test to cover the case when there are no releases and rename the method

warehouse/routes.py
domenkozar: k

warehouse/views.py
domenkozar: Yup!

warehouse/packaging/models.py
dstufft: So if you did this in a loop with [0] it would do a different query for each one? That's interesting. I just assumed it would do a single query and then index in Python, that's fine then.

warehouse/packaging/models.py
dstufft: We're going to deploy Warehouse and PyPI side by side for awhile, so keeping as much as possible as far as data integrity and update functionality will help ensure a consistent behavior during that time that both things are accessing the database.

Beyond that time period, I just have a preference for using the abilities of the database where possible, I'm not really concerned about performance penalty, just that I feel it's cleaner since those actions will work even when you're not going through the ORM so it's less likely that they won't happen.

domenkozar:
@dstufft alright I've simplified it and that should address all your feedback.

I haven't been able to run tests that require the db, will try to get them working inside the docker some day..

dstufft:
What problem are you having with the tests?

warehouse/views.py
```diff
@@ -0,0 +1,37 @@
...
+from warehouse.accounts.models import User
+
+
+@view_config(
+    route_name="index",
+    renderer="index.html",
+)
+def index(request):
+    latest_updated_releases = request.db.query(Release)\
+                                        .order_by(Release.created)[:20]
```
karan: Magic number!!!!

domenkozar: Isn't that 42?

domenkozar: Oh, I see: `2, 8, 20, 28, 50, 82, and 126`

berkerpeksag:
I think the logo should be updated to use the Python logo (or just a "Warehouse (pypi.python.org)" text like the current one).

warehouse/templates/index.html
```diff
@@ -0,0 +1,52 @@
...
+            <li>{{ num_users }}<span>Users</span></li>
+          </ul>
+        </div>
+    </header>
+
+    <div class="content">
+        <h1>{{ _('Recently Released') }}</h1>
+        <ul>
+            {% for release in latest_updated_releases %}
+            <div class="comment">
```
berkerpeksag: Can "comment" be replaced with a more generic name (e.g. "list", "data-list")?

warehouse/templates/index.html
```diff
@@ -0,0 +1,52 @@
...
+            <li>{{ num_files }}<span>Files</span></li>
+            <li>{{ num_users }}<span>Users</span></li>
+          </ul>
+        </div>
+    </header>
+
+    <div class="content">
+        <h1>{{ _('Recently Released') }}</h1>
+        <ul>
+            {% for release in latest_updated_releases %}
```
berkerpeksag: Needs `<li>` here.

warehouse/templates/index.html
domenkozar: data-list is pretty much as generic as comment.

warehouse/templates/index.html
```diff
@@ -0,0 +1,52 @@
...
+            <div class="comment">
+              <div class="comment-image">
+                    <img src="{{ gravatar(release.author_email, size=50) }}" alt="{{ release.author}}" height="50" width="50">
+              </div>
+              <div class="comment-content">
+                <h1><a href="{{ request.route_path('packaging.project', name=release.project.normalized_name) }}">
+                      {{ release.project.name }}</a>
+                </h1>
+                <p>{{ release.summary }}</p>
```
berkerpeksag: I think `release.summary` can be empty. Wrapping it with an `if' (just like in templates/packaging/detail.html) would be nice.

warehouse/templates/index.html
```diff
@@ -0,0 +1,52 @@
...
+    <div class="content">
+        <h1>{{ _('Recently Released') }}</h1>
+        <ul>
+            {% for release in latest_updated_releases %}
+            <div class="comment">
+              <div class="comment-image">
+                    <img src="{{ gravatar(release.author_email, size=50) }}" alt="{{ release.author}}" height="50" width="50">
+              </div>
+              <div class="comment-content">
+                <h1><a href="{{ request.route_path('packaging.project', name=release.project.normalized_name) }}">
```
berkerpeksag: Using `<h2>` or `<h3>` would be nice here.

warehouse/templates/index.html
domenkozar: fixed

warehouse/templates/index.html
domenkozar: removed `<ul>`

warehouse/templates/index.html
```diff
@@ -0,0 +1,50 @@
+{% extends "base.html" %}
+
+{% block content %}
+<section class="index-page">
+    <header class="header">
+        <div class="flash-notice">
+            <p><strong class="site-name hidden-xs hidden-sm">Warehouse (pypi.python.org)</strong>
```
dstufft: This notice needs translated, also instead of hardcoding the name it should use ``request.registry.settings['site.name']``.

PR #493, opened 2015-04-14
Add more user information to the profile page.  Fix #403

### LINKED ISSUE #403: Finish implementing the User Profile
Right now the user profiles (``/user/<username>/``) is primarily just some placeholder content that doesn't have any real function. Actually implement a reasonable profile page.

vladimir-v-diaz:


berkerpeksag:
I wouldn't add the email field (at least without using some helper to make crawlers life harder). Also, it would be nice to use more semantic markup (e.g. `<ul><li>` or `<dl>`).

dstufft:
I don't think we should put things like the user's email address, when they last logged in, their user id, or whether or not their account is active on the profile page.

A decent example of what this should probably look like is something like https://warehouse.python.org/user/dstufft/. The template that rendered that (on the old code base) is located here: https://github.com/pypa/warehouse/blob/werkzeug/warehouse/templates/accounts/profile.html#L20-L58. Hopefully that should give some guidance on what I think the profile page should contain.

dstufft:
I'm going to close this since #751 has fleshed out the user profile page and this is no longer needed.

PR #494, opened 2015-04-14
Redirect after login or logout (#fix 415).

dachshund:
Again, probably not perfect, but good enough.

I copied some code from Django! Namely, for ```warehouse.utils.http.is_safe_url```.

Someone else (probably @dstufft) needs to review legality and security, though.

### DIFF
--- a/tests/accounts/test_views.py
+++ b/tests/accounts/test_views.py
@@ -118,16 +118,24 @@ class TestLogin:
             name="unauthenticated_userid",
         )
 
+        route_path = lambda route_name, username: "/accounts/{}/".\
+                                                  format(username)
+        # TODO: Why must I mock this request method? Why is the original
+        # route_path on the request object not already aware of routes?
+        pyramid_request.route_path = pretend.call_recorder(route_path)
+
+        username = "theuser"
         form_obj = pretend.stub(
             validate=pretend.call_recorder(lambda: True),
-            username=pretend.stub(data="theuser"),
+            username=pretend.stub(data=username),
         )
         form_class = pretend.call_recorder(lambda d, login_service: form_obj)
 
         result = views.login(pyramid_request, _form_class=form_class)
 
         assert isinstance(result, HTTPSeeOther)
-        assert result.headers["Location"] == "/"
+        assert result.headers["Location"] == route_path("accounts.profile",
+                                                        username)
         assert result.headers["foo"] == "bar"
         assert pyramid_request.find_service.calls == [
             pretend.call(ILoginService, context=None),
--- a/tests/accounts/test_views.py
+++ b/tests/accounts/test_views.py
@@ -161,6 +161,86 @@ class TestLogin:
         ]
         assert pyramid_request.session.new_csrf_token.calls == [pretend.call()]
 
+    @pytest.mark.parametrize("with_user", [True, False])
+    def test_post_validate_no_redirects(self, monkeypatch, pyramid_request,
+                                        with_user):
+        remember = pretend.call_recorder(
+            lambda request, user_id: [("foo", "bar")]
+        )
+        monkeypatch.setattr(views, "remember", remember)
+
+        new_session = {}
+
+        login_service = pretend.stub(
+            find_userid=pretend.call_recorder(lambda username: 1),
+        )
+        pyramid_request.find_service = pretend.call_recorder(
+            lambda iface, context: login_service
+        )
+        pyramid_request.method = "POST"
+        pyramid_request.session = pretend.stub(
+            items=lambda: [("a", "b"), ("foo", "bar")],
+            update=new_session.update,
+            invalidate=pretend.call_recorder(lambda: None),
+            new_csrf_token=pretend.call_recorder(lambda: None),
+        )
+
+        pyramid_request.set_property(
+            lambda r: 1234 if with_user else None,
+            name="unauthenticated_userid",
+        )
+
+        route_path = lambda route_name, username: "/accounts/{}/".\
+                                                  format(username)
+        # TODO: Why must I mock this request method? Why is the original
+        # route_path on the request object not already aware of routes?
+        pyramid_request.route_path = pretend.call_recorder(route_path)
+
+        username = "theuser"
+        form_obj = pretend.stub(
+            validate=pretend.call_recorder(lambda: True),
+            username=pretend.stub(data=username),
+        )
+        form_class = pretend.call_recorder(lambda d, login_service: form_obj)
+
+        # The set of all possible next URLs. Since this set is infinite, we
+        # test only a finite set of reasonable URLs.
+        nexts = ["/security/"]
+
+        for next in nexts:
+            pyramid_request.POST["next"] = next
+
+            result = views.login(pyramid_request, _form_class=form_class)
+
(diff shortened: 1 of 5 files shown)


warehouse/utils/http.py
```diff
@@ -0,0 +1,49 @@
...
+    # so we must not allow this syntax.
+    if not url_info.netloc and url_info.scheme:
+        return False
+    # Forbid URLs that start with control characters. Some browsers (like
+    # Chrome) ignore quite a few control characters at the start of a
+    # URL and might consider the URL as scheme relative.
+    if unicodedata.category(url[0])[0] == 'C':
+        return False
+    return ((not url_info.netloc or url_info.netloc == host) and
+            (not url_info.scheme or url_info.scheme in ['http', 'https']))
```
dstufft: Small nitpick: Might as well make this {"http", "https"} and use a set literal since we're Python 3.4+ only (unlike Django).

warehouse/accounts/views.py
```diff
@@ -133,10 +143,15 @@ def logout(request):
         #       to handle this for us.
         request.session.invalidate()
 
+        redirect_to = request.POST.get(redirect_field_name,
+                                       request.GET.get(redirect_field_name))
+
+        # Security check -- don't allow redirection to a different host.
+        if not is_safe_url(url=redirect_to, host=request.host):
+            redirect_to = request.path
```
dstufft: I know this is what Django does, but it kind of seems silly to redirect back to the logout view after they've successfully been logged out just because the ``redirect_to`` isn't valid. Probably this should just use the default view. That doesn't actually exist until #492 lands, so I'd just hardcode it to ``"/"`` for now.

warehouse/accounts/views.py
```diff
@@ -96,11 +102,15 @@ def login(request, _form_class=LoginForm):
         # and we don't want to continue using the old one.
         request.session.new_csrf_token()
 
+        # If the user-originating redirection url is not safe, then redirect to
+        # the user's profile instead.
+        if not is_safe_url(url=redirect_to, host=request.host):
+            redirect_to = request.route_path("accounts.profile",
```
dstufft: I think it makes more sense to redirect users to the index view rather than their own profile in the case that the place they were trying to go was invalid. A route for this doesn't exist until #492 so for now you can just hardcode ``"/"``.

tests/utils/test_http.py
```diff
@@ -0,0 +1,57 @@
...
+
+class TestHttp:
+
+    # (MOSTLY) FROM https://github.com/django/django/blob/
+    # 011a54315e46acdf288003566b8570440f5ac985/tests/utils_tests/test_http.py
+    def test_is_safe_url(self):
+        # A None URL is not safe!
+        assert not is_safe_url(None)
+
+        for bad_url in ('http://example.com',
```
dstufft: This should use pytest parameterized tests, like:

```python

class TestIsSafeURL:

    @pytest.mark.parametrize(
        "url",
        [
            None,
            "http://example.com",
            "http:///example.com",
            ...,
        ],
    )
    def test_rejects_bad_url(self, url):
        assert not is_safe_url(url, host="testserver")

    @pytest.mark.parametrize(
        "url",
        [
            "/view/?param=http://example.com",
            "/view/?param=https://example.com",
            ...,
        ],
    )
    def test_accepts_good_url(self, url):
        assert is_safe_url(url, host="testserver")
```

This will make the tests themselves a lot more declarative and will make it so a failure on a single value early on won't cause the rest of the tests to not run.

Additionally you don't need to provide an explicit assertion message, pytest will provide a rich repr for the assertion failure that gives more details than a standard assert error.

tests/accounts/test_views.py
```diff
@@ -167,3 +263,27 @@ def test_post_forgets_user(self, monkeypatch, pyramid_request):
...
+        forget = pretend.call_recorder(lambda request: [("foo", "bar")])
+        monkeypatch.setattr(views, "forget", forget)
+
+        # The set of all possible next URLs. Since this set is infinite, we
+        # test only a finite set of reasonable URLs.
+        nexts = ["/security/"]
+
+        pyramid_request.method = "POST"
+        pyramid_request.session = pretend.stub(
+            invalidate=pretend.call_recorder(lambda: None),
```
dstufft: You don't actually need the ``call_recorder`` here since we're not testing that this view calls invalidate in this test. The only thing this test should really be concerned about testing is whether or not ``POST`` redirects or not.

tests/accounts/test_views.py
```diff
@@ -167,3 +263,27 @@ def test_post_forgets_user(self, monkeypatch, pyramid_request):
...
+        )
+
+        for next in nexts:
```
dstufft: This should use the same parametrize method as below to pass each next url into the test instead of doing the loop inside of the test itself.

tests/accounts/test_views.py
```diff
@@ -167,3 +263,27 @@ def test_post_forgets_user(self, monkeypatch, pyramid_request):
         assert result.headers["foo"] == "bar"
         assert forget.calls == [pretend.call(pyramid_request)]
         assert pyramid_request.session.invalidate.calls == [pretend.call()]
+
+    def test_post_redirects_user(self, monkeypatch, pyramid_request):
+        forget = pretend.call_recorder(lambda request: [("foo", "bar")])
```
dstufft: You also don't need this ``call_recorder``.

tests/accounts/test_views.py
dstufft: This doesn't actually test the failure case at all, I would make this something like:

```python

@pytest.mark.parametrize(
    ("next_url", "expected"),
    [
        ("/security/", "/security/"),
        ("http://example.com/", "/"),
    ],
)
def test_post_redirects_user(self, monkeypatch, pyramid_request, next_url, expected):
    ....
    pyramid_request.POST["next"] = next_url
    ...
    assert result.headers["Location"] == expected
```

tests/accounts/test_views.py
dstufft: You also don't need to worry about testing that the logout view handles the ``foo: bar`` header, since that's tested elsewhere.

tests/accounts/test_views.py
```diff
@@ -118,33 +118,129 @@ def test_post_validate_redirects(self, monkeypatch, pyramid_request,
...
+        ]
+        assert pyramid_request.route_path.calls == [
+            pretend.call(route_name, username=username),
+        ]
         assert pyramid_request.session.new_csrf_token.calls == [pretend.call()]
 
+    @pytest.mark.parametrize("with_user", [True, False])
+    def test_post_validate_no_redirects(self, monkeypatch, pyramid_request,
+                                        with_user):
+        remember = pretend.call_recorder(
```
dstufft: Much like logout, this doesn't need to duplicate all of the tests that we already have that test things like "does the login view actually login" and it should instead just focus on testing whether the login view redirects to the URL it's passed in if that url is valid, or to the default url if not. That means you can ditch most of the ``call_recorders`` and just use bare lambdas or fakes. Additionally the loop should be handled by ``parametrize`` not looping over test values inside the test case itself.

dachshund:
@dstufft : Addressed your comments, please take a look!

PR #334, opened 2014-08-08
Register Page / Registration Flow 

toumorokoshi:
Related to #322, this is a first draft at a registration page / registration flow. As of the initial pull request, it is the full workflow sans sending a confirmation e-mail, and and agreeing to a terms of use.

PR #507, opened 2015-04-26
register page - pyramids edition

toumorokoshi:
This is the pyramids edition of #334, or at least part of it.

In the interest of better feedback I'll be creating a series of smaller requests around implementing registration, rather than one giant one.

This is the UI portion of it: a simple interface that always returns a success.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ node_modules/
 
 dev/example.sql
 data/
+
+*~
--- a/tests/accounts/test_forms.py
+++ b/tests/accounts/test_forms.py
@@ -104,27 +104,33 @@ class TestRegisterForm:
 
     def _get_valid_formdata(self):
         return {
-            'email': 'myusername@username.com',
-            'username': 'my_username',
-            'password': 'foo',
-            'confirm': 'foo',
+            "email": "myusername@username.com",
+            "username": "my_username",
+            "password": "foo",
+            "confirm": "foo",
         }
 
     def test_validate_username_with_user_leading_nonalphanumeric(self):
         data = self._get_valid_formdata()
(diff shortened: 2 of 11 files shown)


toumorokoshi:
I thought this would also be a good place to ask about actually adding the user: my current plan is to create a Pyramids service to handle reg interaction with the DB, and pass that into the RegisterForm like how the LoginForm works.

My question is: can I combine Reg + Login services into a UserLoginService? Or is it preferable to have them separate.

I kind of feel like having them be the same is a good grouping, functionality-wise, but I'm not married to either way (or some third way I should be implementing reg).

dstufft:
I think it makes sense to have them in a single service, just rename ``LoginService`` to ``UserService`` or something. That way all of the user related stuff goes through a service and we can swap out services for in memory ones or whatever easily.

dstufft:
Is this dependent on #508 now?

toumorokoshi:
yes. Although I think I'll probably have to send one more pull request to do something to send a confirmation e-mail out, before this is complete. so I guess we should let this one sit and I'll pick it back uponce the other functionality exists.

karan:
Is this ready to be merged?

toumorokoshi:
Not yet, unfortunately. This is waiting on #508, more functionality is required.

ismail-s:
@toumorokoshi #508 has been merged, so is this pull request ready to be merged?

toumorokoshi:
@ismail-s sorry about that, this dropped on my radar. I haven't had a chance to tackle it all the way yet. I'll start taking a look again, but if someone else wants to take the reigns, they're welcome to.

dstufft:
This is fixed now.

PR #508, opened 2015-05-03
expanding LoginService to UserService

toumorokoshi:
The changes here add a couple methods that are necessary for the db part of #507: creating and updating user information. It does so by adding a couple method to LoginService, and also renaming it to UserService.

Aside from this, there was one more piece of functionality we need: the creation of e-mails. I was wondering about which approach is desirable here:

* creating a separate service (EmailService) to CRUD emails.
* adding the email as a string into the create_user method on IUserService (this would be the most convenient for those who want to create a user, as e-mails are necessary).

I'll hold off on adding that functionality until I hear a vote one way or the other.


.gitignore
```diff
@@ -17,4 +17,4 @@ htmlcov/
 node_modules/
 
 dev/example.sql
-data/
+data/
```
toumorokoshi: I'm not sure what this was about. I'll change it back.

warehouse/accounts/services.py
```diff
@@ -75,6 +78,19 @@ def check_password(self, userid, password):
 
         return True
 
+    def create_user(self, user):
```
dstufft: I think I would like it better if this method took all of the data and actually created the new user itself, so it would have a call signature more like:

```python
def create_user(self, username, ...):
    user = User(username=username, ...)
    ...
    return user
```

I think this matches the other methods on this service better where it accepts parameters which are not specific to the fact we're storing users in a SQLAlchemy managed database but returns a newly created user object. We can probably get away without the ``find_userid`` check here too and let the calling code deal with checking if there is already a user by the name or not.

warehouse/accounts/services.py
```diff
@@ -75,6 +78,19 @@ def check_password(self, userid, password):
...
+        if self.find_userid(user.username):
+            raise UserAlreadyExists(
+                "User with username {0} already exists!".format(user.username)
+            )
+        self.db.add(user)
+        self.db.commit()
+
+    def update_user(self, user):
```
dstufft: Similarly to above, I think this would be better off if it accepted a userid and a list of things to change, like:

```python
def update_user(self, userid, **changes):
    user = self.get_user(userid)
    for attr, value in changes.items():
        setattr(user, attr, value)
    return user  # Maybe? Not sure if we should return the user here or just let it alone
```

warehouse/accounts/services.py
```diff
@@ -75,6 +78,19 @@ def check_password(self, userid, password):
 
         return True
 
+    def create_user(self, user):
+        if self.find_userid(user.username):
+            raise UserAlreadyExists(
+                "User with username {0} already exists!".format(user.username)
+            )
+        self.db.add(user)
+        self.db.commit()
```
dstufft: We shouldn't commit the database here, we have a single transaction per request which self.db is already a part of which will either get committed if the view is successful (if it returns a value of some kind) or rollbacked otherwise (if it raises an exception of some kind, even an HTTP exception).

warehouse/accounts/services.py
```diff
@@ -75,6 +78,19 @@ def check_password(self, userid, password):
...
+
+    def update_user(self, user):
+        user = self.db.merge(user)
+        self.db.add(user)
+        self.db.commit()
```
dstufft: Similar to above, we shouldn't commit here, transactions are managed at a higher level.

dstufft:
Thanks a lot for this Pull Request!

I've made a few comments inline, assuming you don't disagree with those things, I think this will look good to merge once you address those comments.

toumorokoshi:
Thanks! I understand what changes need to be done, I'll tackle them soon.

@dstufft what are your thoughts on adding the email? It's a separate email model, but can I just add that as a parameter in the UserService create_user?

dstufft:
I would just add it as a parameter for now, if we need to expand on that later we can.

toumorokoshi:
@dstufft , I addressed the comments and made changes. Let me know what you think.

The one additional method I added was 'verify_user': this is used when a user verifies their e-mail/account.

toumorokoshi:
@dstufft have you had a chance to look at this? would love some feedback or to get this merged in.

dstufft:
Ah, oops. Missed the fact that you updated this since the last comment.

warehouse/accounts/services.py
```diff
@@ -75,6 +76,35 @@ def check_password(self, userid, password):
...
+                    is_active=False, is_staff=False, is_superuser=False):
+        user = User(username=username,
+                    name=name,
+                    password=password,
+                    is_active=is_active,
+                    is_staff=is_staff,
+                    is_superuser=is_superuser)
+        user.last_login = datetime.datetime.now()
+        self.db.add(user)
+        self.db.flush()
```
dstufft: Do we need this flush here? I'm not super good at SQLAlchemy but my understanding was that it would generally flush whenever it needed to?

dstufft: Oh, I bet you're doing this so that you can get a ``user.id`` for down below? I *think* you can do ``Email(user=user, ...)`` instead.

warehouse/accounts/services.py
```diff
@@ -75,6 +76,35 @@ def check_password(self, userid, password):
...
+        email_object = Email(email=email, user_id=user.id,
+                             primary=True, verified=False)
+        self.db.add(email_object)
+        return user
+
+    def update_user(self, user_id, **changes):
+        user = self.get_user(user_id)
+        for attr, value in changes.items():
+            setattr(user, attr, value)
+        self.db.add(user)
```
dstufft: I don't think we need ``self.db.add`` here since the user should already exist as part of the session.

warehouse/accounts/services.py
```diff
@@ -75,6 +76,35 @@ def check_password(self, userid, password):
...
+        return user
+
+    def verify_user(self, user_id):
```
dstufft: I think this would be better if it was ``verify_email(self, user_id, email_address)``, then this should just work for a future where we have multiple email addresses and we need to verify them independently.

tests/unit/accounts/test_services.py
```diff
@@ -104,11 +104,44 @@ def test_check_password_updates(self, db_session):
...
+    def test_create_user(self, db_session):
+        user = UserFactory.build()
+        email = "foo@example.com"
+        service = services.DatabaseUserService(db_session)
+        new_user = service.create_user(username=user.username,
+                                       name=user.name,
+                                       password=user.password,
+                                       email=email)
+        user_from_db = service.get_user(new_user.id)
+        db_session.flush()
```
dstufft: Do we need this flush here?

dstufft:
Ok, few more small comments but I think this looks really good and if you make those adjustments I'll be happy to merge it ASAP.

warehouse/accounts/services.py
toumorokoshi: yep! I thought I had some issues getting that working, but it works fine. I'll make the changes.

warehouse/accounts/services.py
toumorokoshi: you're right. I'll remove it. I think my understanding of how add works was mistaken, I'll be more careful in the future.

warehouse/accounts/services.py
toumorokoshi: sounds good! will do.

tests/unit/accounts/test_services.py
toumorokoshi: I think we do. Since get_user requires a userid, which doesn't exist in the session until it's flushed. I figured the correct behavior to test for was a user gets written to the database, which we don't know until the session is actually flushed in some fashion. 

conversely, I could add a test to check if the new_user returned is added to the current session. That would effectively validate the desired behavior from create_user:

    assert Session.object_session(new_user) is db_session

toumorokoshi:
@dstufft sorry for the delayed reply. I've addressed the comments.

tests/unit/accounts/test_services.py
dstufft: Ok, I'm OK with this flush then since there seems to be a good reason for it. Thanks for the explanation!

PR #511, opened 2015-05-08
Conditional header

karan:
If user is logged out, they see Sign up and Sign in button. Otherwise they see buttons to take then to their profile and to logout.

### Logged in

image

### Logged out

image


dstufft:
Hmm, something is apparently broken because this shouldn't actually be possible, accessing request.user should require decorating a view with ``@uses_session``. My plan for this was to have it done with javascript actually, the reason is that 99% of the content on a PyPI page is not user specific, so if we keep the responses user agnostic then we canc ache the response once in varnish for all users, instead of once per user.

dstufft:
Heh, I bet it's because it's accessing it inside of a template, and that happens at the rendering layer.

karan:
> accessing it inside of a template

That is true. And I don't think anything is broken yet. I forgot to add translations for the text in the buttons in the second commit, but fixed it in the third one.

I'm not sure how the js implementation would work - an ajax request perhaps, but I think that's just overcomplicating the client code.

dstufft:
The problem is, let's take the ``/`` view other than this little bit of the UI there is nothing in there that is user specific so if we can avoid adding user specific data to it we can use the same cached response for every user. This is important because PyPI gets roughly 100 million HTTP requests a day and anything we can cache at Fastly is a request that doesn't hit our backends which let us serve more traffic with less servers (and be faster at it too). So Warehouse has a goal where we don't add user specific content to pages unless that page *needs* it to function.

There are a few ways to handle this that doesn't cause us needing to cache every page for each user instead of just globally:

* As you identified, use an AJAX request to fetch the current logged in user and render this little bit of UI based on that.
* Have the login/logout view add another cookie that just states the logged in user's name. The web application will never use this cookie but a little bit of js can decide to render a logged in UI based on the existence of this cookie.
* Use edge side includes which will allow us to cache the main project page, but drop a little tag into the response which will be replaced inside of varnish with the value of some other response. This is sort of like the ``{% include %}`` tag except it happens inside of varnish and varnish can cache the main response separately (and without being user specific) while caching the user information per user.

To be clear, the broken thing is that this PR works at all, because Warehouse should error when you attempt to access ``request.user`` unless you've decorated a view with the ``uses_session`` decorator.

karan:
Ah this makes sense. Thanks for explaining this.

I'll reimplement this using the second method.

PR #518, opened 2015-05-22
Rely on pip 7's auto wheel building for caching

dstufft:


alex:
this is the best

dstufft:
Goes from ~2m8s to ~50s for a full test run in the super scientific "I ran each once". All without any need to do anything special.

PR #536, opened 2015-06-10
Set login view as forbidden_view_config

### LINKED ISSUE #417: Redirect users to the login view when they don't have permission
We need to configure Pyramid so that it will redirect users to the login view whenever they don't have permission to do something (maybe conditionally on if they are already logged in).

maxking:
Fixes #417

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -12,7 +12,7 @@
 
 from pyramid.httpexceptions import HTTPMovedPermanently, HTTPSeeOther
 from pyramid.security import remember, forget
-from pyramid.view import view_config
+from pyramid.view import forbidden_view_config, view_config
 
 from warehouse.accounts import REDIRECT_FIELD_NAME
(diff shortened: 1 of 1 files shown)


dstufft:
I don't think we should just call the login view, I think we should have a separate view which does nothing but redirects to the login view. However we also need a predicate on this so that it doesn't happen for anything registered with ``warehouse.legacy.action_routing.add_pypi_action_route``. So probably we just need a custom predicate that looks something like:

```python
def not_legacy_pypi_action(info, request):
    return not request.path.startswith("/pypi") and ":action" not in request.params
```

dstufft:
A better idea is to do something like:

```python
@forbidden_view_config()
def redirect_to_login(request):
    return Redirect To Login

@forbidden_view_config(request_param=":action")
def do403(request):
    return Do a 403
```

maxking:
Cool! I will update my pull request then.

Issue #57, opened 2013-10-08
Existing Package Upload APIs


dstufft:
- [x] submit - Submit metadata
- [x] submit_pkg_info - Submit metadata as a PKG-INFO file
- [x] file_upload - Upload a File
- [x] doc_upload - Upload documentation

dstufft:
FWIW I'm working on this now.

dstufft:
These have all now been implemented, either as a real implementation or as a stub that tells users they are no longer supported.

PR #537, opened 2015-06-13
Implement the Legacy API for Uploading Files

dstufft:
This covers the ``file_upload`` portion of #57, it still is a work in progress, but it currently does work and allows news files to be uploaded (as well as creating the resulting releases and projects if needed).


warehouse/legacy/api/pypi.py
```diff
@@ -0,0 +1,478 @@
...
+    # Buffer the entire file into memory, checking the hash of the file as we
+    # go along.
+    file_content = b""
+    file_size = 0
+    file_hash = hashlib.md5()
+    for chunk in iter(lambda: request.POST["content"].file.read(8096), b""):
+        file_size += len(chunk)
+        if file_size > MAXIMUM_FILESIZE:
+            raise _exc_with_message(HTTPBadRequest, "File too large.")
+        file_content += chunk
```
alex: Don't make a large string with `+=`, it's quadratic time. Use `io.BytesIO`

warehouse/legacy/api/pypi.py
```diff
@@ -0,0 +1,478 @@
...
+        file_hash.update(chunk)
+
+    # Get the signature if it was included.
+    if "gpg_signature" in request.POST:
+        signature = b""
+        for chunk in iter(
+                lambda: request.POST["gpg_signature"].read(8096), b""):
+            signature += chunk
```
alex: Same (you can just read the whole thing here, you're not doing anything with each chunk)

PR #539, opened 2015-06-16
Filesizeformat filter raised error on package description page, fixed!

maxking:
`filesizeformat` filter in the template raised error when the file.size if None. Check that and put up a 0 when file.size is  None.


dstufft:
This should probably just be:

```
<td>{% if file.size %}{{ size|filesizeformat() }}{% endif %}</td>
```

I don't like the implicit ``0MB`` because a ``None`` value doesn't mean it's a zero size file, it means we don't know what the size is. I think in that case it's better to just omit it.

PR #553, opened 2015-07-13
Added missing keys.

terrrydactyl:
Most keys were simply missing from the JSON statement, but requires_dist
needed to be added to the Release table. Added an alembic migration file to
handle that. Also updated tests to reflect the new JSON structure.

### DIFF
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -142,12 +142,16 @@ class TestJSONRelease:
         headers = db_request.response.headers
         assert headers["Access-Control-Allow-Origin"] == "*"
         assert headers["X-PyPI-Last-Serial"] == je.id
-
         assert result == {
             "info": {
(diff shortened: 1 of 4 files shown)


dstufft:
``requires_dist`` should already exist in the database, it's not yet been mapped to the ORM though, it's in a related table located at: https://github.com/pypa/warehouse/blob/master/warehouse/legacy/tables.py#L471-L485. This table also holds other types of dependency information, so you're going to need to map different ``kind`` to different keys in the JSON blob. 

There are other keys than this as well, you can see how the old dict was constructed here: https://bitbucket.org/pypa/pypi/src/4272c77983e871ff2b0152841758fcda90919d46/rpc.py?at=default#cl-170

berkerpeksag:
If I'm not missing something, this has been fixed in https://github.com/pypa/warehouse/commit/f40e734642c8128d0151ecc3016e34c99caf64e2.

PR #584, opened 2015-08-05
Verify URLs in forms are actually URLs. Fixes #573.

### LINKED ISSUE #573: Validate that Project URLs are actually URLs on Upload
As the title says, validate the project url metadata on upload, probably with [rfc3986](https://pypi.python.org/pypi/rfc3986).

wlonk:
Fixes #573.

Here's a cute animal picture for your troubles:

fox kits

_"Vulpes vulpes young" by Siga - Own work. Licensed under CC BY-SA 3.0 via Wikimedia Commons - https://commons.wikimedia.org/wiki/File:Vulpes_vulpes_young.JPG#/media/File:Vulpes_vulpes_young.JPG_

### DIFF
--- a/warehouse/legacy/api/pypi.py
+++ b/warehouse/legacy/api/pypi.py
@@ -20,6 +20,7 @@ import packaging.version
 import pkg_resources
 import wtforms
 import wtforms.validators
+from rfc3986 import uri_reference
 
 from pyramid.httpexceptions import HTTPBadRequest, HTTPForbidden, HTTPGone
 from pyramid.response import Response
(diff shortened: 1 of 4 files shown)


warehouse/legacy/api/pypi.py
```diff
@@ -194,7 +195,8 @@ def _validate_project_url(value):
     if not url:
         raise wtforms.validators.ValidationError("Must have an URL.")
 
-    # TODO: Actually validate that the URL is a valid URL.
+    if not rfc3986.is_valid_uri(url, require_scheme=True):
```
dstufft: This should probably include a check to make sure that the scheme we do get is either http or https. This will also accept things like ``ssh://foobar`` which I think is not something that we want to actually do here. I think it shouldn't be super hard to adjust this to check the scheme too though.

dstufft:
Thanks! One comment for some further validation, but otherwise this looks good to me.

wlonk:
Pushed changes per your feedback.

dstufft:
Woo, thanks!

PR #585, opened 2015-08-05
Long filenames break rendering.

### LINKED ISSUE #338: Long filenames break rendering
Here's how https://warehouse.python.org/project/Pillow/ looks on my 1366x768 laptop:

ekrano nuotrauka is 2014-12-03 11 48 31

The culprit appears to be

ekrano nuotrauka is 2014-12-03 11 49 22

I'm sure this can be fixed with some CSS.

wlonk:
Fixes #338.

I opted to fix this with a CSS overflow property. In most cases, it shouldn't look different, but for very long file names, such as in the Pillow project, you'll be able to see it.

I notice that the compiled CSS is under version control; that seems undesirable to me. Would there be any objections to adding something like `warehouse/static/css/*.css` to `.gitignore`? How would that interact with the `manifest.json`?

As per the ancient prophecy, here is a cute animal picture:

Aussie!

_"Hazel beim Spiel" by SamyG - Own work. Licensed under CC BY 3.0 via Wikimedia Commons - https://commons.wikimedia.org/wiki/File:Hazel_beim_Spiel.jpeg#/media/File:Hazel_beim_Spiel.jpeg)_


dstufft:
We do check in the compiled CSS, largely so that working on Warehouse doesn't require having the CSS toolchain installed unless you're actively working on the CSS. Moving away from committing the compiled CSS would likely require setting up the docker containers so that they installed the node.js tooling that we need and ran ``grunt watch`` as part of the process. Not an impossible thing but not something I had bothered dealing with yet.

wlonk:
Gotcha, thanks for the clarification. I might put some work into that, but as it's not a blocker for replacing PyPI, it comes in further down the priority list.

Issue #162, opened 2013-11-17
Properly style the error pages


alex:
This includes (but is probably not limited too):
- [ ] 500
- [ ] 404
- [ ] 410 (Maybe? Not sure if we'll ever use it.)
- [ ] 403

dstufft:
One thing that I think would be kind of cute to do, is on the 404 page, embed a link to https://www.youtube.com/watch?v=cWDdd5KKhts as an easter-egg referencing both the "unofficial" name of PyPI and the fact that the skit itself is about a cheeseshop that's missing any cheese... essentially a 404 ;)

dstufft:
Bonus points if we stick the real 404 content at the top, then lead into the embedded youtube with a "And now for something completely different".

PR #588, opened 2015-08-07
[#162] Style error pages.

wlonk:
See #162. A first stab at styling 404 and 500 pages.

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -17,7 +17,6 @@ import pretend
 import pytest
 
 from pyramid import renderers
-from pyramid.httpexceptions import HTTPMovedPermanently
 
 from warehouse import config
 from warehouse.utils.wsgi import ProxyFixer, VhmRootRemover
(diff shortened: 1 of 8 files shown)


wlonk:
There's obviously wording and style work to do here, but I thought I'd get this up for feedback and review sooner than later. Most of the work was in wiring the pages up; I'm not very familiar with Pyramid, so if there's a better way, please let me know.

wlonk:
Unsure why coverage is reporting failed when it's apparently at 100%. Will deal with it in the morning.

willingc:
@wlonk It looks like the Travis buld is failing a PEP8 check (Click on the Details link above next to `continuous-integration/travis-ci/pr):
screenshot 2015-08-10 19 46 39
It looks like you have imported something that you are not using. 

I am guessing that codecov/project fails for coverage since Travis is not happy. Perhaps double check your config.py file.

wlonk:
Thanks, @willingc! I hadn't realized coverage would complain if there were other test errors.

I had also failed to build the necessary translations files, so I fixed that, and we're green.

But I'm not at all sure that the design here is worth merging yet, so if anyone has any feedback on that, I'd appreciate it too!

dstufft:
Yea, coverage does that because if all the test jobs don't complete it might not have correct data.

wlonk:
Yeah, makes sense.

dstufft:
Going to close this since we've landed #751 which has started with a whole new design of everything. Thanks for the PR!

wlonk:
:+1:

PR #616, opened 2015-08-20
Enable NewRelic

dstufft:


ewdurbin:
may want to benchmark any performance oriented routes with this before rolling too far with it.

i haven't touched this space with python in the last year, but in previous encounters, catchall instrumentation has proved costly.

dstufft:
Do you mean specific to New Relic, or just that you've had bad experiences with the concept of catch all instruentation?

ewdurbin:
catchall instrumentation tends to affect performance similarly, but not nearly as drastically as a profiler.

consider the kinds of introspection that are occurring to provide the rich data that new relic (and similar) provide.

in the end, if the value attained from this outweighs the performance penalty it's not a problem. just note that some benchmarking may be worthwhile to know what that performance penalty is.

PR #626, opened 2015-08-24
Reword statements to not make derogatory remarks

dstufft:
It was requested that the documentation for Warehouse not make derogatory jokes about the legacy code base, so we'll go ahead and remove them.


sigmavirus24:
:+1:

PR #638, opened 2015-08-27
Switch to using eventlet instead of Nginx

dstufft:


piotr-dobrogost:
Out of curiosity - what is the reason for this switch?

dstufft:
Less stuff running in the Dyno lets me us more of the Dyno RAM serving requests.

dstufft:
Oh, and since Python will have direct control over the file, we can dynamically choose to reject a large upload instead of having a static limit in nginx. This'll let us have a default limit for all users, and then increase it on a per project account.

PR #653, opened 2015-09-03
Default new projects to use pypi-only

dstufft:


alex:
heh, test failure is relevant

PR #666, opened 2015-09-07
Redis fail gracefully

### LINKED ISSUE #665: Fail gracefully if Redis isn't Up
Right now if the redis for the statistics go down, we'll start throwing errors when rendering the page, which is obviously less than desirable. Ideally we'll want to fail gracefully for this so that if redis goes down we'll just stop rendering statistics until it comes back. A tricky thing here is that we'll want to use ESI so that we don't cache for a long time the statistics with the failure output instead of statistics. This makes sense anyways since we'll want our stats to be updated quicker than the bulk of the page anyways. Additionally, we should ensure that we log the error regardless of whether we raise an exception or not.

dstufft:
Fixes #665

### DIFF
--- a/vcl/main.vcl
+++ b/vcl/main.vcl
@@ -138,6 +138,13 @@ sub vcl_fetch {
     # If we've gotten an error after the restarts we'll deliver the response
     # with a very short cache time.
     if (beresp.status == 500 || beresp.status == 503) {
+        # If this is a ESI request, then instead of returning the error we're
+        # going to return a blank page so that our top level page acts as if it
+        # did not have ESI rather than inlining the error page.
+        if (req.url ~ "^/_esi/") {
+            error 900 "ESI Error";
+        }
+
         set req.http.Fastly-Cachetype = "ERROR";
         set beresp.ttl = 1s;
         set beresp.grace = 5s;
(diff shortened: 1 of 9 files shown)


vcl/main.vcl
```diff
@@ -135,6 +135,13 @@ sub vcl_fetch {
         return (pass);
     }
 
+    # If we've gotten an error after the restarts on an ESI page, then we'll
+    # replace the error contents with a blank page so that we don't render the
+    # error contents into the final document.
+    if (req.url ~ "^/_esi/" && (beresp.status == 500 || beresp.status == 503)) {
```
mattrobenolt: I'd personally check the status code first, since it's a more path to be a non-50x error and it's faster to compare an int vs a regular expression. :)

Alternatively, you could check against `beresp.http.Warehouse-ESI-Enable`, no? And do this for all ESI? Or is this intended for very specific endpoints?

dstufft: ``beresp.http.Warehouse-ESI-Enable`` only exists on the primary response (it's how we signal to Varnish to enable ESI for this response). Good point on the order of the if statements though.

mattrobenolt: > beresp.http.Warehouse-ESI-Enable only exists on the primary response (it's how we signal to Varnish to enable ESI for this response). 

Ah, yeah, wasn't entirely sure what that was. :)

PR #670, opened 2015-09-09
Tune uWSGI

dstufft:
We'll tune uWSGI to have better values for our environment, this is paired with a change on Heroku that will set each Dyno to run 6 uWSGI processes.

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -16,6 +16,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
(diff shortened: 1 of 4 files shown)


warehouse/templates/base.html
```diff
@@ -16,6 +16,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <meta name="blitz" content="mu-bdd143b9-97688c78-086ac034-d58d4464">
```
ewdurbin: what's this

dstufft: It lets me use blitz.io against Warehouse. I plan on removing it in a few.

ewdurbin: :+1:

PR #674, opened 2015-09-10
invoke pip.compile

alex:


requirements.txt
```diff
@@ -19,7 +19,7 @@ cffi==1.2.1               # via bcrypt
 click==5.1
 docutils==0.12            # via botocore, readme
 hiredis==0.2.0
-html5lib==0.999999        # via bleach
+html5lib==0.9999999       # via bleach
```
mattrobenolt: lol

alex: it's fine. probably no software in this stack represents versions as floats.

ewdurbin:
Does this resolve the invalid IPv6 errors? Or is it just keeping up with the releases of html5lib?

dstufft:
It should resolve them.

PR #675, opened 2015-09-10
Only restart on 502 and 503 errors

dstufft:
If our app is sending 500 errors we don't want to increase the traffic to it incase it's a load issue.

### DIFF
--- a/vcl/main.vcl
+++ b/vcl/main.vcl
@@ -108,9 +108,9 @@ sub vcl_fetch {
         error 803 "SSL is required";
     }
 
-    # If we've gotten a 500 or a 503 from the backend, we'll go ahead and retry
+    # If we've gotten a 502 or a 503 from the backend, we'll go ahead and retry
     # the request.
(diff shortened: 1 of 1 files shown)


vcl/main.vcl
```diff
@@ -110,8 +110,7 @@ sub vcl_fetch {
 
     # If we've gotten a 500 or a 503 from the backend, we'll go ahead and retry
```
mattrobenolt: Might wanna update comment too.

vcl/main.vcl
```diff
@@ -110,8 +110,7 @@ sub vcl_fetch {
...
     # the request.
-    if ((beresp.status == 500 || beresp.status == 503) &&
-            req.restarts < 1 &&
+    if (beresp.status == 503 && req.restarts < 1 &&
```
mattrobenolt: I'd personally also recommend 502 as well.

mattrobenolt:
:+1: :shipit:

PR #706, opened 2015-09-28
Efficiently handle file uploads

dstufft:
Currently we buffer the uploaded file entirely in memory. This works fine, except it's pretty heavy on the memory use, up to 60MB per upload by default. This doesn't work out so well when we're in a memory constrained environment like a Heroku dyno.

Instead we'll buffer the files to a temporary location on disk. Once they are there we'll also utilize boto3's ``upload_file`` helper which will efficiently upload the file without loading the entire thing into memory, and will also transparently switch to using multipart uploading (with retries!) when the file size gets too large.

### DIFF
--- a/tests/unit/legacy/api/test_pypi.py
+++ b/tests/unit/legacy/api/test_pypi.py
@@ -12,6 +12,7 @@
 
 import io
 import os.path
+import tempfile
 
 from unittest import mock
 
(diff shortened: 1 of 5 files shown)


tests/unit/legacy/api/test_pypi.py
```diff
@@ -989,9 +1002,12 @@ def test_upload_succeeds_with_wheel(self, pyramid_config, db_request,
             ),
         })
 
-        storage_service = pretend.stub(
-            store=pretend.call_recorder(lambda path, content: None),
-        )
+        @pretend.call_recorder
+        def storage_service_store(path, file_path):
+            with open(file_path, "rb") as fp:
+                fp.read() == b"A fake file."
```
alex: Missing an assert here

PR #738, opened 2015-11-07
create openid connect "sub" column for migrating Google Accounts

ewdurbin:


dstufft:
You'll need to add this column to the table in ``warehouse/legacy/tables.py`` as well.

dstufft:
The index too.

ewdurbin:
@dstufft is that correct?

dstufft:
Yup.

PR #741, opened 2015-11-13
Switch from uWSGI to Gunicorn

dstufft:
While uWSGI has a lot of options and is theoretically better than gunicorn, we're going to switch back to gunicorn. uWSGI breaks a lot of the rules of a Python interpreter in order to squeeze out more performance and the sum of all of these made working with it difficult.

We'll use gunicorn with the gevent worker so that we don't block threads when there are large uploads happening.


ewdurbin:
have you considered one of the asyncio worker types for gunicorn?

http://docs.gunicorn.org/en/latest/design.html#asyncio-workers

No need for gevent since this is a python 3 app.

dstufft:
The ``gaiohttp`` worker requires the entire application to be asynchronous I think since it's not running the application using a thread or anything. I couldn't figure out the parameters of the ``gthread`` worker, it doesn't look like it's actually using asyncio at all and it's just a thread pool.

ewdurbin:
sounds good, just wanted to make sure :)

piotr-dobrogost:
> uWSGI breaks a lot of the rules of a Python interpreter

This is intriguing; do you have any specific rules in mind?

dstufft:
@piotr-dobrogost A lot of the stuff can be configured back to to act like "normal", but I was still getting some weird behavior and eventually just gave up and switched.

Some of the things are like:

* It doesn't use a new dictionary for each request, instead it shares a dictionary between requests and clears it out after the request is over.
* It doesn't have a GIL enabled by default, you have to turn it on with the ``--enable-threads`` option.

There are other little weird things like that, almost always there's some configuration option that fixes it, but it just ended up being hard to track down when something weird was happening.

PR #751, opened 2015-11-19
Import the new design

dstufft:


Turbo87:
@nlhkabu @dstufft just out of curiosity: https://crates.io/ (the Rust package index) currently isn't looking all that fabulous and should probably be redesigned at some point. what are your thoughts on sharing the general design with some color theme adjustments?

nlhkabu:
@Turbo87 the design is Apache License 2.0, although the logo and footer images are not included in this.  It still has a lot of work to go, so I wouldn't recommend forking it just yet, but in principle it should be fairly easy to reskin for another use case.  All of the fonts and colours are controlled by SCSS variables, so that will help!

I am currently very busy with this project, but once the dust has settled, I am happy to help you put together an alternative colour scheme / font set for the rust community.

dstufft:
Side note, crates.io always makes me chuckle... since the original version of Warehouse (many years ago and the new code doesn't share anything with it but a name) was something I wrote to power "crate.io" - an alternative front end to PyPI (before I was a PyPI admin :D).

PR #761, opened 2015-11-20
Refactor our environment so we're no longer an installable package

dstufft:
Previously we attempted to be a Python package as well as a project directory. This made some things harder for very little benefit since we were never actually going to distribute Warehouse as an application.

Instead we'll just rip all of that out and assume we're operating from a checkout of the code. This allows us to make assumptions to make things easier to use.


dstufft:
Wrong repository.

PR #821, opened 2015-11-22
Display a relative timestamp for when a project was released

### LINKED ISSUE #791: Display how long ago the release was made
We'll want to display how long ago the release was made on the project/release detail page. Probably we'll want to implement this in javascript?

dstufft:
Show how long ago a release was made in the project detail page. This uses the [http://timeago.yarp.com/](timeago) js module to do the formatting in javascript. This means that we do not need to worry about caching these values nor the wrong value being displayed in an old tab.

This looks like:

screen-shot-2015-11-22-11-58-53

It will search for any HTML5 ``<time>`` tags with a class of ``relative`` to apply the relative display too. I'd like it if @nlhkabu could verify this is OK with her to use that class name and tag, and if it is it'd be great if she could either merge this PR or +1 it so I can.

Fixes #791

### DIFF
--- a/warehouse/static/js/main.js
+++ b/warehouse/static/js/main.js
@@ -63,4 +63,7 @@ $(document).ready(function() {
     positionWarning();
   });
 
+  // Format all of the time.relative tags to display relative time.
+  $(".-js-relative-time").timeago();
+
 });
--- a/bower.json
+++ b/bower.json
@@ -10,7 +10,8 @@
   ],
   "dependencies": {
     "jquery": "2.1.4",
-    "font-awesome": "4.4.0"
+    "font-awesome": "4.4.0",
+    "jquery-timeago": "1.4.3"
   },
   "overrides": {
     "font-awesome": {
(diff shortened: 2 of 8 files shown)


warehouse/static/js/main.js
```diff
@@ -63,4 +63,7 @@ $(document).ready(function() {
     positionWarning();
   });
 
+  // Format all of the time.relative tags to displat relative time.
```
alexwlchan: Typo: "displat" ~> "display"

dstufft: Fixed, thanks!

nlhkabu:
@dstufft, looks fine to me, except can we please prefix any JavaScript hook classes with `-js` (so this would become `-js-relative`).  Any class with `-js` will not be styled with CSS :)

If possible, could we also choose a more descriptive name, for example, `-js-relative-time`?

Thanks

dstufft:
Yup I can do that.

dstufft:
Ok. Updated to use -js-relative-time

PR #863, opened 2015-12-16
Add more validations to the legacy upload API

### LINKED ISSUE #575: Verify that uploaded files are valid files
Currently, we just blindly accept any file data that has been uploaded to Warehouse. This should change so that we attempt to verify the *contents* of the file to make sure that it is the kind of file that it claims to be and that it seems to be well formed. PyPI Legacy has some code around doing this which may be helpful.

### LINKED ISSUE #601: Uploading doesn't add a journal entry
Currently the upload API is not actually adding a journal entry, this needs to happen.

### LINKED ISSUE #620: Enable the upload view
Once the upload view is safe to use on the production database, we need to re-enable it.

dstufft:
Fixes #575
Fixes #620
Fixes #601 

Note: This won't issue purge requests to PyPI itself, just to warehouse. So anyone using this will need to do something like ``curl -XPURGE https://pypi.python.org/simple/<foo>/`` manually.

### DIFF
--- a/warehouse/legacy/api/pypi.py
+++ b/warehouse/legacy/api/pypi.py
@@ -15,6 +15,7 @@ import hmac
 import os.path
 import re
 import tempfile
+import zipfile
 
 import packaging.specifiers
 import packaging.version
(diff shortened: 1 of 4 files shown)


warehouse/legacy/api/pypi.py
```diff
@@ -640,6 +714,11 @@ def file_upload(request):
             )
 
         # TODO: Check the file to make sure it is a valid distribution file.
```
alex: Probably this TODO can go?

PR #878, opened 2015-12-30
Use the project-urls concept to allow additional links in the header

### LINKED ISSUE #783: Display project-urls at the top of the project detail page
We'll want to display the project urls at the top of the project detail page (where homepage/source code/etc are currently). Ideally this will try and select an appropriate icon based on the site.

dstufft:
We could probably add in more types of icons, but these should be good enough to get us started and adding more in the future should be pretty easy.

Fixes #783

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -10,12 +10,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import OrderedDict
+
 import pretend
 import pytest
 
 from pyramid.security import Allow
 
-from warehouse.packaging.models import ProjectFactory, File
+from warehouse.packaging.models import (
+    ProjectFactory, Dependency, DependencyKind, File,
+)
 
 from ...common.db.packaging import (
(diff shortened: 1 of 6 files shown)


nlhkabu:
Thanks @dstufft - I'm very excited to see this working!!

The only icon that is obviously missing at first glance is a download icon... it seems that many packages have this link.

The download icon should overwrite any other (i.e. GitHub, Bitbucket).  For example on this package 
https://warehouse-staging.python.org/project/django-excel-response3/ we should see a GitHub icon, then a download icon, despite the second also matching GitHub.

For this we should use this icon: http://fontawesome.io/icon/cloud-download/ (`fa-cloud-download `)

Also - it seems like we still have TODOs on some packages (see https://warehouse-staging.python.org/project/googlefinance/) - is this correct?

dstufft:
The existing TODOs will just be pages that are still cached in Fastly.

We'll want to put something like:

```
{% if name|lower == "download" %}
   {% set icon = "cloud-download" %}
```

As the first if statement in the chain before anything else to make that take precedence.

nlhkabu:
Ok, I can take a look at this.  Thanks.

PR #882, opened 2016-01-01
Replaced TODO links in footer with actual links

Stefan-Code:
Not sure how Develoment Credits are tracked as there is no AUTHORS file. I just linked to the [github contributors page](https://github.com/pypa/warehouse/graphs/contributors).

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -84,9 +84,9 @@
       <div class="footer-links">
         <ul>
           <li><h3 {{ l20n("getHelp") }}>Get Help</h3></li>
-          <li><a {{ l20n("installingPackages") }} href="TODO">Installing Packages</a></li>
-          <li><a {{ l20n("uploadingPackages") }} href="TODO">Uploading Packages</a></li>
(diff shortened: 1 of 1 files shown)


dstufft:
I'm going to let @nlhkabu take a look at this and maybe @ddbeck since they're more familiar with the documentation effort and where these should link too.

ddbeck:
Personally, I'm always happy to see placeholders get replaced with something functional (and these particular suggestions look OK to me), though with the understanding that some links may change as work proceeds on the user guide (none of that stuff has been merged to master, so there's nothing to link to yet). I'm not sure about what process is needed to make sure these things get reviewed again later, however; I'd have to defer on that question.

dstufft:
Just a new PR is enough to change them later. I'll go ahead and merge :D

nlhkabu:
This is great, thanks @Stefan-Code.  For everyone's reference, I do want to create a separate 'credits' page before launch.  I have created an issue for this here: https://github.com/pypa/warehouse/issues/890

PR #887, opened 2016-01-04
Compute and store sha256 for uploads

dstufft:


warehouse/legacy/api/pypi.py
```diff
@@ -731,21 +731,22 @@ def file_upload(request):
...
+                for hasher in file_hashes.values():
+                    hasher.update(chunk)
 
         # Actually verify that the md5 hash of the file matches the expected
         # md5 hash. We probably don't actually need to use hmac.compare_digest
         # here since both the md5_digest and the file whose file_hash we've
         # computed comes from the remote user, however better safe than sorry.
         if not hmac.compare_digest(
-                form.md5_digest.data, file_hash.hexdigest()):
+                form.md5_digest.data, file_hashes["md5"].hexdigest()):
```
sigmavirus24: So, what about allowing the uploader to specify a sha256 digest? Could this be updated to

```py
if not (hmac.compare_digest(
            form.md5_digest.data, file_hashes["md5"].hexdigest()) or
        hmac.compare_digest(
            form.sha256_digest.data, file_hashes["sha256"].hexdigest())):
```

dstufft: I plan on doing that as well yes. Was just focused on this part first.

dstufft: Although I was going to require it to match *both* of them (if both are given) and require at least one of them.

sigmavirus24: Right. If both are provided, both should match. I wasn't entirely certain what `md5_digest.data` if it wasn't provided, that's why I hedged with `or`

PR #889, opened 2016-01-04
Use SHA256 instead of MD5

### LINKED ISSUE #820: Replace md5 with sha256 on UI
Since PyPI is being rewritten, it's a good opportunity to replace md5 hashes with sha256.

NOTE: there is also #681 which talks about the upload functionality related to md5

dstufft:
Fixes #820


nlhkabu:
:tada:

PR #891, opened 2016-01-04
Allow uploading to use sha256 as well as (or instead of) md5

### LINKED ISSUE #681: Support package upload with a SHA-256 checksum
Currently package upload respects `md5_digest` as a form parameter and [validates the digest](https://github.com/pypa/warehouse/blob/master/warehouse/legacy/api/pypi.py#L646).

Discussion took place earlier this year (or last year) about migrating to using SHA-256 as the digest algorithm instead. We should begin to validate that and support it so tools (e.g., twine) can start uploading packages using this digest algorithm.

dstufft:
Fixes #681

### DIFF
--- a/tests/unit/legacy/api/test_pypi.py
+++ b/tests/unit/legacy/api/test_pypi.py
@@ -297,8 +297,25 @@ class TestMetadataForm:
     @pytest.mark.parametrize(
         "data",
         [
-            {"filetype": "sdist"},
-            {"filetpye": "bdist_wheel", "pyversion": "3.4"},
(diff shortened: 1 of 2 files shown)


sigmavirus24:
If PyPI can currently handle/ignore extra fields, I'll uncomment https://github.com/pypa/twine/blob/5c06ed2049b633fe7f0b0e27df32d2715614ab3a/twine/package.py#L124 for testing with this change

dstufft:
PyPI should just completely ignore extra fields yes.

Issue #345, opened 2014-12-20
Ability to mark a version of a package as deprecated or unsupported


alex:
Then warehouse can serve that info back when `pip` looks up a package, and `pip` can display a warning.

This allows maintainers to inform users that theyr'e using an out of date version, without needing them to modify the source of the old package.

warsaw:
This came up during a Pycon dinner conversation.  IWBNI the admins could similarly mark a package.  We were discussion the still active use of oauth despite the fact that it's been unmaintained (with no new uploads) upstream since 2009, isn't py3 compatible, there are better replacements (albeit with different APIs) on PyPI, and yet `oauth` still comes up as the top search hit.  We don't want people using the package any more, but we also don't want to break existing usages.  I'd like to see the package deprecated with a pip warning on install, and also down-rank the package in PyPI searches.

dstufft:
So given a mechanism like described in this ticket, it would be technically possible for the PyPI admins to do that. The bigger question would be whether or not we'd *want* to declare something unmaintained by fiat. Having the admins declare something unmaintained has a decent chance of blowing up if we accidentally mark something as unmaintained when it's really not unmaintained, it's just "finished" (for some value of finished).

That doesn't of course mean we can't do it, but that the biggest barrier would be determining a policy for when we consider it something that's OK to do.

/cc @r1chardj0n3s

warsaw:
Sure, you'd want to be careful, but no uploads in the last 6 years is a pretty good sign a package is abandoned. <wink>   Maybe you could even automate some of the criteria (no recent uploads for some $value of recent, etc.)

r1chardj0n3s:
I would only do such a thing if I could get the approval of the owner of
the listing.

fbender:
In addition to emitting a warning through `pip`, I'd make sure the Warehouse shows a big banner with a deprecation notice and in that banner link to a potential successor module (in the way [PEP-437 intended the `obsoleted_by` keyword](https://www.python.org/dev/peps/pep-0426/#obsoleted-by)). As a first measure, I'd add an option to the package preferences on Warehouse to name a successor package (and if there is none, e.g. because the package is no longer developed, just show a generic deprecation message on Warehous).

> Maybe you could even automate some of the criteria (no recent uploads for some $value of recent, etc.)

[...]

PR #910, opened 2016-01-09
[Draft] Implement Security Feature

jayfk:
This is a first draft about how a package security feature might look like.

The basic idea is that package maintainers can mark a release as insecure and provide a URL to a CVE or a blog post describing the vulnerability as a first step to implement something like the suggestion in https://github.com/pypa/warehouse/issues/798
# Styles
I've borrowed a lot of the excellent styles that are already in the codebase and added a few new style classes. 

If you take a look at the attached screenshots, you'll see that I've added a big red banner on top and a smaller label on the *Version History & Changelog* tab.

screen shot 2016-01-06 at 10 13 31

screen shot 2016-01-06 at 10 13 16

# Workflow

A maintainer that wants to mark a release is insecure would go to a new page where all the packages he or she maintains are listed, select one or multiple releases and provide a URL for further explanation.

screen shot 2016-01-09 at 18 39 07

# Implementation

This draft adds two new fields to the `packaging.Release` model:

- insecure *Boolean*
- insecure_url *Text*

And a new view to `accounts.views` that I've called `mark_insecure`. In its current form, the view does nothing other than passing a `dict` containing two example projects to the template.

Is this something you would consider to add to the new codebase? I'm open to all feedback for this draft and I'm willing to go down a completely different route if this doesn't fit.

If this is something you'd consider, I'd be happy to continue to work on this to fill the missing parts.

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -80,7 +80,16 @@
     <p class="help-link"><a href="#" {{ l20n("howToInstall") }}>How do I install this?</a></p>
   </div>
 </section>
-
+{% if release.insecure %}
+<section class="horizontal-section -bad -medium">
+  <div class="site-container">
+    <p {{ l20n("versionInsecure") }}>
+            This version has been marked insecure by the package maintainers. You can read more
+            about the vulnerabilites <a href="{{ release.insecure_url }}">here</a>.
+    </p>
+  </div>
+</section>
+{% endif %}
 <section class="package-status">
   <div class="site-container">
     {% if release.version != all_releases[0].version %}
@@ -206,6 +215,9 @@
                     <span class="highlight-badge" {{ l20n("thisVersion") }}>This version</span>
                   {% else %}
                     <p><a href="{{ request.route_path('packaging.release', name=release.project.name, version=hrelease.version) }}">{{ hrelease.version }}</a></p>
+                  {% endif %}
+                  {% if hrelease.insecure %}
+                    <span class="highlight-badge -bad" {{ l20n("insecure") }}>Insecure</span>
                   {% endif %}
                 </div>
 
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -174,7 +174,7 @@ def logout(request, redirect_field_name=REDIRECT_FIELD_NAME):
 @view_config(
     route_name="accounts.mark-insecure",
     renderer="accounts/mark_insecure.html",
-    #decorator=[csrf_protect("accounts.mark-insecure"), uses_session],
+    decorator=[csrf_protect("accounts.mark-insecure"), uses_session],
 )
 def mark_insecure(request):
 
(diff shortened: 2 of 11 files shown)


dstufft:
I've not looked at the code itself at all, but the feature itself seems sound to me. I wonder if it makes sense to make it more generic so that it could cover the use case of #345 as well (or maybe not, since it's possible to have an insecure version that hasn't been deprecated or had support dropped).

All design is gated on @nlhkabu so she'd have to speak to any particular design impact and how it should be displayed.

miohtama:
Instead of `insecure` boolean I suggest `marked_insecure_at` datetime, as sometimes you want to track these changes on timeline for audit purposes.

warehouse/templates/packaging/detail.html
```diff
@@ -80,7 +80,16 @@ <h3 class="title"><a href="{{ request.route_path('packaging.project', name=relea
     <p class="help-link"><a href="#" {{ l20n("howToInstall") }}>How do I install this?</a></p>
   </div>
 </section>
-
+{% if release.insecure %}
+<section class="horizontal-section -bad -medium">
```
nlhkabu: Can we try a 'thin' style here instead?  It looks a little fat to me - I think the red is distinct enough without taking up the extra screen real estate.

nlhkabu:
Thanks @jayfk. Both the design and front end code look great - the only thing I'd like to see is the banner size reduced (as per my above comment).

warehouse/templates/packaging/detail.html
```diff
@@ -80,7 +80,16 @@ <h3 class="title"><a href="{{ request.route_path('packaging.project', name=relea
...
+  <div class="site-container">
+    <p {{ l20n("versionInsecure") }}>
+            This version has been marked insecure by the package maintainers. You can read more
+            about the vulnerabilites <a href="{{ release.insecure_url }}">here</a>.
```
alexwlchan: Nitpicks:

1. "vulnerabilites" ~> "vulnerabilities"
2. It's not great to have "here" as link text, but since it's an arbitrary link it's hard to have something more descriptive that's always appropriate. Perhaps make the sentence to `Read more about the vulnerabilities`, and make the whole sentence a link?

nlhkabu: On number 2:
You could either do what @alexwlchan suggests, or simply make the whole sentence a link:
`<a href="{{ release.insecure_url }}">You can read more about the vulnerabilities here</a>`

jayfk:
> Thanks @jayfk. Both the design and front end code look great - the only thing I'd like to see is the banner size reduced (as per my above comment).

Yes! I've played around with it a little (even had the whole banner in red for a while), but I had always the impression that it was a bit too much.

> I've not looked at the code itself at all, but the feature itself seems sound to me. I wonder if it makes sense to make it more generic so that it could cover the use case of #345 as well (or maybe not, since it's possible to have an insecure version that hasn't been deprecated or had support dropped).

Maybe we could switch the fields to something like:

- deprecated_at: datetime (what @miohtama suggested so that we can keep track of it)
- deprecated_reason: ["insecure", "not maintained", None]

This way a package that is `insecure` would be a special case of a deprecated package and we have one interface to manage it all. What do you think?

jayfk:
Implementation in #1462.

PR #927, opened 2016-01-25
Lint HTML

### LINKED ISSUE #811: Lint HTML
Currently the HTML is not pure HTML5 (my mistake!!), as, for example, it still includes `/>` on image tags.

It would be great if we could install an HTML5 linter to catch this now and in future.

di:
This PR adds HTML5 linting to the `lint` target in order to enforce the Google HTML standard via https://pypi.python.org/pypi/html-linter (which was recently made Python 3.5 compatible).

This PR also fixes the resulting errors/warnings from the linter.

Fixes #811.

### DIFF
--- a/warehouse/templates/accounts/login.html
+++ b/warehouse/templates/accounts/login.html
@@ -11,16 +11,28 @@
     {% endif %}
 
     {% if form.errors.__all__ %}
-    <ul class="errors">{% for error in form.errors.__all__ %}<li>{{ error }}</li>{% endfor %}</ul>
+    <ul class="errors">
+      {% for error in form.errors.__all__ %}
+      <li>{{ error }}</li>
+      {% endfor %}
+    </ul>
     {% endif %}
 
     {% if form.username.errors %}
-    <ul class="errors">{% for error in form.username.errors %}<li>{{ error }}</li>{% endfor %}</ul>
+    <ul class="errors">
+      {% for error in form.username.errors %}
+      <li>{{ error }}</li>
+      {% endfor %}
+    </ul>
     {% endif %}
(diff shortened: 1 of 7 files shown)


warehouse/templates/accounts/login.html
```diff
@@ -11,16 +11,28 @@
     {% endif %}
 
     {% if form.errors.__all__ %}
-    <ul class="errors">{% for error in form.errors.__all__ %}<li>{{ error }}</li>{% endfor %}</ul>
+    <ul class="errors">
+      {% for error in form.errors.__all__ %}
+      <li>{{ error }}
```
sigmavirus24: Closing `</li>` disappeared (same below)

di: Hi @sigmavirus24, thanks for the feedback. The end tag can be omitted here per the HTML5 spec, see https://html.spec.whatwg.org/multipage/semantics.html#the-li-element

luzfcb: @di  I agree with @sigmavirus24 
I know that everyone should have read the HTML specification, but omitting `</li>` seems to me a special case and the content of `{{ error }}` can be a HTML.
by conclusion, I think in this case the `Special cases aren't special enough to break the rules.`

dstufft:
This PR looks pretty good to me, the one thing I'd want before we merge it is to remove the optional tags being omitted rule and don't omit them in the code base. I know it's valid HTML5 to do that, but aesthetically balanced tags are easier to read in my opinion.

di:
@dstufft I've addressed your feedback, thanks!

dstufft:
Cool, thanks!

PR #950, opened 2016-01-31
Swap version history icons

### LINKED ISSUE #949: Switch icon colours in Version History & Changelog
Currently, the icon colours in the Version history are Blue for old versions and White for the current version. This is exact opposite to that of the navigation to the left. I suggest that the icons be swapped so that the current version gets a Blue icon and older versions get a White icon.

pradyunsg:
Fixes #949


nlhkabu:
Thanks @pradyunsg.  Can you please provide a screenshot so I can quickly validate this from a design perspective?

Issue #944, opened 2016-01-29
Move package navigation to the left


techtonik:
This is the current package page on the notebook with a touchpad.

![image](https://cloud.githubusercontent.com/assets/515889/12668587/25fd6bdc-c669-11e5-909c-ad1bd45a6fe6.png)

As you see, I have to use scrollbar on the right to get to the package content. Then I have to go all the way across the screen to get to different information about package, then I have to go back to scrollbar again, because tab content is unlikely to fit page again.

nlhkabu:
Thanks for your feedback @techtonik - I have opened #1988 to address this and other similar feedback.

techtonik:
#1988 is a general issue. I am afraid that this specific "Move package navigation to the left" wish will be lost.

See https://docs.ubuntu.com/maas/2.1/en/installconfig-checklist where it is implemented.

PR #953, opened 2016-02-01
Save some vertical space on the banner

Ivoz:
Helps slightly towards #944


dstufft:
This will have to be looked at by @nlhkabu

nlhkabu:
Thanks @Ivoz.  I'm thinking that we can probably gain some extra space by also reducing the h1 size as well - down to perhaps 2rem?

When making visual changes, it would also be great if you could provide a screenshot; this would allow me to quickly validate it.

Ivoz:
fullsize
fullwidth

nexus 5
nexus5

PR #960, opened 2016-02-06
Add default search message, suggestions

### LINKED ISSUE #803: Default search message
We need to add a message for when a search doesn't match any packages.

di:
This fixes #803 by adding a default search message. It also adds a suggestion for the search term when necessary (per the discussion in #803).

@nlhkabu This is definitely missing localization stuff, I wasn't sure how that worked, please advise :)

### DIFF
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -100,7 +100,10 @@ class TestSearch:
         params = {"q": "foo bar"}
         if page is not None:
             params["page"] = page
-        query = pretend.stub()
+        suggest = pretend.stub()
+        query = pretend.stub(
+            suggest=pretend.call_recorder(lambda *a, **kw: suggest),
+        )
         request = pretend.stub(
             es=pretend.stub(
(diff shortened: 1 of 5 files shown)


sigmavirus24:
@di you have three commits with the same message. It might be worth squashing those (or editing the commit messages to not be identical)

di:
@sigmavirus24 I think you just caught me while rebasing, there is only one commit.

sigmavirus24:
@di that's why I removed my comment ;)

nlhkabu:
thanks for this @di - as I said on the issue - it's awesome. @dstufft set up the translations, so he is in a better position to advise on this than me.

FYI @dstufft I am happy for this to be merged on the design side ;D

di:
@dstufft I found the documentation on translations and took a shot at it.

@nlhkabu FYI, I noticed that long package names cause some wrapping of the "Order Packages By" elements:

screen shot 2016-02-08 at 10 02 19 am

PR #975, opened 2016-02-19
Consistent output of keywords (tags) in package detail pages

### LINKED ISSUE #855: Make tag display consistent
tags-alt

tags

We need to make the output of tags consistent.  Each tag should be followed by a comma and space `, ` i.e `tag1, tag2, tag3`

rjwebb:
This fixes #855 by splitting the string containing the keywords into a list (using a jinja2 tag). The keywords are then formatted as comma separated <span> elements.

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -175,6 +175,10 @@ def test_urlparse():
             ["python", "finance", "data", "code", "test automation"]
         ),
         (
+            "'python'; finance; \"data\";        code    ; test automation",
+            ["python", "finance", "data", "code", "test automation"]
+        ),
+        (
             "a \"b\" c   d  'e'",
             ["a", "b", "c", "d", "e"]
         ),
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -165,3 +165,24 @@ def test_urlparse():
     inp = "https://google.com/foo/bar?a=b"
     expected = urllib.parse.urlparse(inp)
     assert filters.urlparse(inp) == expected
+
+
+@pytest.mark.parametrize(
+    ("inp", "expected"),
+    [
+        (
+            "'python', finance, \"data\",        code    , test automation",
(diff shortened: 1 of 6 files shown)


dstufft:
This looks pretty good to me. The one thing I'm not sure about is whether the ``,`` should be inserted as part of the HTML or if it should be inserted using CSS (as you've done). I don't have a strong opinion but maybe @nlhkabu does?

dstufft:
Also, you have linting errors.

nlhkabu:
Hey Bob!  I'm glad you found the project!
Yes, I'd much prefer to keep the commas out of the CSS and instead insert them as part of the HTML.

nlhkabu:
Also, would you mind providing a screenshot of the new rendering?  Thanks ;D

rjwebb:
screenshot from 2016-02-19 10 54 40

This is what the keyword tags now look like.

di:
@rjwebb Looks good! Like I mentioned on #855, some of the tags are delimited by semicolons and not commas, it would be nice to support splitting those as well, e.g.: `natural language processing; thesaurus` becomes `['natural language processing', 'thesaurus']`.

rjwebb:
@di Whoops, I missed your comment about semicolons! I agree, I should modify ```format_tags``` to accommodate that.

What if we have a string that contains commas _and_ semicolons? Should one take precedence, or should we just split on both of them? (e.g. ```"finance; business, data"``` -> ```['finance', 'business', 'data']```).

rjwebb:
Hm, I think that checking for commas, then checking for semicolons, then splitting on whitespace is a fine way to go. I couldn't find any cases in the database where people had mixed commas and semicolons.

di:
@rjwebb I agree that commas should take precedence. LGTM! :+1:

nlhkabu:
Ready to merge on my side @dstufft

dstufft:
Thanks!

PR #978, opened 2016-02-19
Only show first tab if hash is empty

### LINKED ISSUE #976: Opening a project detail page shows all of the tabs
Currently, if you go to a ``/project/foo/`` page, you get the contents of all of the tabs one after another instead of just the first tab. For Example:

screen-shot-2016-02-19-07-11-21

I assume that this is a regression caused by #972 so maybe @di has an idea :)

di:
Fixes #976. (Sorry about that).


dstufft:
No Problem! Thanks for fixing it quickly :)

PR #981, opened 2016-02-22
Add padding to large-logo

### LINKED ISSUE #977: No spacing between large logo and home page content
This may be intended, but it looks a bit odd to me. There isn't any spacing between the large logo and the home page text. This isn't immediately obvious because the real logo is mostly transparent background on it's right hand side, but you can see it with the TM in the top right corner. 

screen-shot-2016-02-19-07-15-37

di:
This fixes #977 by adding padding to the `.large-logo` class. The 50px margin was being ignored due to the `display: table-cell` on the element.

From https://developer.mozilla.org/en-US/docs/Web/CSS/margin: 
> Applies to: all elements, except elements with table display types other than table-caption, table and inline-table.

Before (in Chrome version 50.0.2655.0):
screen shot 2016-02-22 at 11 11 08 am

After:
screen shot 2016-02-22 at 11 12 30 am

@nlhkabu, I'm curious what this looks like in whatever browser you were using which was showing a margin.

### DIFF
--- a/warehouse/static/sass/components/_about-pypi.scss
+++ b/warehouse/static/sass/components/_about-pypi.scss
@@ -9,8 +9,8 @@
     text-align: left;
 
     @media only screen and (max-width: 550px){
-       display: block;
-       width: 100%;
+      display: block;
+      width: 100%;
     }
 
     @media only screen and (max-width: 550px){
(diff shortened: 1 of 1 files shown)


nlhkabu:
Thanks @di.

For reference, it looks like this on FF 43.0.3

screenshot from 2016-02-27 18-15-18

Issue #854, opened 2015-12-06
When you visit the homepage, the search box should have focus by default


alex:

nlhkabu:
Yes, I agree.  However, we have had some early user feedback that suggests this is not great on mobile, where the focus prompts the keyboard, which can take up to 50% of the screen real estate.

Perhaps the solution is to only focus the input for viewports > 400px.

alex:
that sounds like a good compromise!

nlhkabu:
Update on this (in case anyone else is thinking on jumping in).  I have a fix for this and will make a PR soon.

carlcarl:
Seems this should be closed?

miohtama:
This still bugs on tablet, the soft keyboard appears, as one cannot discriminate tables and hardware keyboard devices from each other based on viewports. Furthermore HDPI mobile screens also may cause the keyboard appear.

The "search focus by default" paradigm is unfortunately not very good approach considering many devices have soft keyboards nowadays. The case has different 10 years ago with different accessibility standards.

I suggest rolling back this as sudden appear of soft keyboard without user interaction is a major buster until we can detect hardware keyboard only devices via JavaScript.

nlhkabu:
The objective here wasn't to avoid prompting a soft keyboard *ever*, but rather to avoid prompting it on small (mobile) screen sizes, where the soft keyboard would take a disproportionate amount of space.

miohtama:
Please take a look on the attached screenshot of Warehouse front page. This is a horrible decision. Please do not have autofocus ever as it is against current usability principles as explained above and this makes me cry with my 20 years of web development experience and I will jump from the bridge if this happens for my beloved PyPi.

![screenshot_20160224-221357](https://cloud.githubusercontent.com/assets/49922/13309625/2edb8bc2-db44-11e5-9637-a5236888d8d7.png)

nlhkabu:
> I will jump from the bridge if this happens for my beloved PyPi.

If you are genuinely considering self harm over this issue, please seek medical help.  Otherwise it is neither necessary nor appropriate to escalate this issue with such language.

I agree that this is a problem.  If you would like to see if changed, please submit a PR.

Refs: #983

dstufft:
I mean, it's not like this is a new thing for PyPI. The current version of pypi.python.org has the autofocus attribute set on the home page. as far as I know, nobody has complained about that so maybe just using the HTML5 is less error prone.

miohtama:
[...]

PR #985, opened 2016-02-25
Use HTML5 autofocus attribute instead of javascript

### LINKED ISSUE #983: Search field autofocused on package pages
On package pages the search field is automatically focused (for example https://warehouse.python.org/project/simplejson/).  This means that once the page loads you must unfocus the search field before you can scroll via the keyboard.

Automatically focusing the search field is arguably reasonable for the front page, where the search field is front and center, but it's a misfeature for pages where search is just part of the site header.  Worse, the field is focused via JavaScript, so it doesn't obey the user's configured autofocus preferences.

dstufft:
The original issue (#854) wanted this to not happen on mobile devices (or more specifically, devices with a soft keyboard) to prevent the soft keyboard from opening and covering up a significant part of the page. This was using an imperfect detection method of using the width of the screen to determine if something had a soft keyboard or not.

Using javascript for this appears to be a bit prone to issues, however there is an ``autofocus`` attribute in HTML5 that can be used for this exact purpose, which we are already using on legacy PyPI without any complaints (to my knowledge). In my limited testing, the ``autofocus`` attribute seems to not bring up a soft keyboard on a mobile device (Mobile Safari on iOS 9.2.1) so I think we can just unconditionally set this attribute and let the UserAgent decide how to handle it.

This may mean that people with an old enough browser don't get autofocus support, but that seems reasonable to me.

/cc @nlhkabu 

Fixes #983

### DIFF
--- a/warehouse/static/js/main.js
+++ b/warehouse/static/js/main.js
@@ -78,18 +78,4 @@ $(document).ready(function() {
   // });
   $(".-js-relative-time").timeago();  // Add back to document.l10n.ready
 
-  // Add focus to search bar.  Only do this above 750px, so we avoid
-  // prompting the keyboard on mobile (which can get in the way).
-
-  function focusSearch(){
-    var screenWidth = $(window).width();
-    if (screenWidth > 750) {
-      $('#search').focus();
-    }
-  }
-
-  focusSearch();
-
-
-
 });
--- a/warehouse/templates/index.html
+++ b/warehouse/templates/index.html
@@ -36,7 +36,7 @@
 
     <form class="search-form -large -fullwidth" action="{{ request.route_path('search') }}">
       <label for="search" class="sr-only">Search PyPI</label>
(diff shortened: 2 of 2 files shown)


dstufft:
A coworker checked this out on android for me, and said that both Chrome and Ghostery did not bring up a soft keyboard for him (but it did have the cursor focused and blinking in the text box, which is different than what iOS did for me, which just didn't have the input focused at all as far as I can tell).

I don't know of anyone with a screen reader to see how this behaves there, but I assume it'll be at least equivalent to the javascript solution, if not superior since the screenreader can make decisions based on the attribute rather than having it happen in JS where it has less control.

nlhkabu:
@miohtama can you please check this on your nexus tablet?

dstufft:
@nlhkabu Any opposition to merging this and seeing if anyone complains?

nlhkabu:
Happy to merge it for now and see what the result looks like.  I'll need to test it across devices before launch to see if there are any issues.

dstufft:
\o/

PR #986, opened 2016-02-25
Remove search field autofocus

miohtama:
Autofocusing fields gives bad user experience on mobile and tablet devices with soft keyboard.

Please ref #854 for discussion.


di:
Hi @miohtama, it looks like this is already being addressed in #985.

dstufft:
Closed via #985

PR #990, opened 2016-02-29
Add RSS feeds

### LINKED ISSUE #390: Implement Legacy RSS Feeds
Determine if we need/still want the legacy RSS feeds and if so implement them (both "last hour" and "packages").

### LINKED ISSUE #794: Work out what to do with RSS
We do not have any links to RSS on the new design.

screenshot-pypi python org 2015-11-20 17-11-13

We need to work out:
a) Are we still supporting them
b) If yes, from where should we link to them (footer maybe??)

di:
This fixes #794 and fixes #390 by adding duplicates of PyPI's two RSS feeds:

* https://pypi.python.org/pypi?:action=rss
* https://pypi.python.org/pypi?:action=packages_rss

It also adds redirects from the old feed URLs to the new ones.

_Note_: If you look at these feeds in a browser such as Chrome, the XML will probably look a little more like garbage than you're used to. This is because our Content Security Policy is not allowing the browser to inline JS or CSS, but it still rewrites the HTML expecting the styles and scripts to be there. You'll see an error in the console such as:

```
Refused to apply inline style because it violates the following Content Security
Policy directive: "style-src 'self' fonts.googleapis.com". Either the
'unsafe-inline' keyword, a hash
('sha256-nx7T2vM3FG2GqDCK5NErQR6tT6iOS/rBjaBaUvmACC0='), or a nonce
('nonce-...') is required to enable inline execution.
```

### DIFF
--- a/tests/unit/i18n/test_filters.py
+++ b/tests/unit/i18n/test_filters.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import babel.dates
+import email.utils
 import pretend
 
 from warehouse.i18n import filters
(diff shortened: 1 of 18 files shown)


dstufft:
Ok, that's kind of hilarious the the browser's own UI is being blocked by CSP.

dstufft:
Thanks!

Issue #1005, opened 2016-03-04
Dynamically alter CSP


demianbrecht:
Currently CSP is set statically in warehouse/config.py. We'll need to add a mechanism for views or templates to alter this, preferably in some generalized fashion.

Use case: I'm adding recaptcha to the user registration page. Amongst source additions to `script-src` et al, it also requires the addition of `"unsafe-inline"` to `style-src` (see [here](https://developers.google.com/recaptcha/docs/faq#recently-my-recaptcha-widget-started-displaying-invalid-site-key-whats-happening for details)). Obviously, this isn't something that we want to add statically for every page and should only be added when absolutely necessary.

A couple different methods I can think of to potentially solve this:

- Add nonces to the *-src headers and make them accessible to templates. This would mean moving the CSP policy formatting to the tween function rather than being in the factory method which would cause a per-render performance impact (albeit super minor). This also wouldn't solve the issue of "unsafe-inline" or other header values not directly rendered in templates.
- Add a decorator to add header value changes as needed per view. This will require moving from setting `request.response.headers` in warehouse.config explicitly to using `request.response.headerlist` so that multiple `Content-Security-Policy` headers can be included in the response. Based on what I've read, browsers *should* support intersecting multiple headers, but that'll have to actually be tested.

Thoughts?

(Maybe there's something already allowing for all this in warehouse and I just haven't found it, which would be wonderful ;))

di:
(I think you actually meant to link to https://developers.google.com/recaptcha/docs/faq#im-using-content-security-policy-csp-on-my-website-how-can-i-configure-it-to-work-with-recaptcha)

I don't think nonces would work here, because the styles/scripts are coming from Google, not from somewhere where we would be able to set a nonce on them.

I don't really see a better way to solve this than to allow `style-src 'unsafe-inline'` on any page using the recaptcha, which probably means any recaptcha should be on it's own page.

dstufft:
Maybe what we want is to move to ``warehouse/csp.py`` and do something like:

```python
def content_security_policy_tween_factory(handler, registry):
    def content_security_policy_tween(request):
        resp = handler(request)

        # We don't want to apply our Content Security Policy to the debug
        # toolbar, that's not part of our application and it doesn't work with
        # our restrictive CSP.
        if request.path.startswith("/_debug_toolbar/"):
            return resp

        policy = registry.settings.get("csp", {})
        policy = some_hypothetical_dict_merge(
            policy,
            request.find_service(name="csp"),
[...]

PR #1007, opened 2016-03-05
Added csp service

demianbrecht:
This change allows for modification of CSP settings at runtime,
defaulting to configuration settings.

Addresses #1005.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ dev/example.sql
 warehouse/.commit
 warehouse/static/components
 warehouse/static/dist
+
+tags
--- a/warehouse/csp.py
+++ b/warehouse/csp.py
@@ -1,0 +1,60 @@
+import collections
+import copy
+
+
+def _serialize(policy):
+    return "; ".join([
+        " ".join([k] + [v2 for v2 in v if v2 is not None])
+        for k, v in sorted(policy.items())
+        if [v2 for v2 in v if v2 is not None]
+    ])
+
+
+def content_security_policy_tween_factory(handler, registry):
+    policy = registry.settings.get("csp", {})
+    
+    def content_security_policy_tween(request):
+        resp = handler(request)
+
+        try:
+            policy = request.find_service(name="csp")
+        except ValueError:
+            policy = collections.defaultdict(list)
+
+        # We don't want to apply our Content Security Policy to the debug
+        # toolbar, that's not part of our application and it doesn't work with
+        # our restrictive CSP.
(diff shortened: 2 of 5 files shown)


warehouse/csp.py
```diff
@@ -0,0 +1,58 @@
...
+        try:
+            policy = request.find_service(name="csp")
+        except ValueError:
+            policy = collections.defaultdict(list)
+
+        # We don't want to apply our Content Security Policy to the debug
+        # toolbar, that's not part of our application and it doesn't work with
+        # our restrictive CSP.
+        if not request.path.startswith("/_debug_toolbar/") and policy:
+            resp.headers["Content-Security-Policy"] = _serialize(policy)
```
dstufft: This should be ``_serialize(policy).format(request=request)``. We use a slightly dynamic CSP in development based on the request attributes. So right now this CSP is:

```
$ curl -I http://192.168.99.100/
HTTP/1.1 200 OK
Server: gunicorn/19.4.5
Date: Sat, 05 Mar 2016 16:10:54 GMT
Connection: keep-alive
Content-Type: text/html; charset=UTF-8
ETag: "zzqeSeiFKitpuq9Sj56MPw"
Content-Length: 11602
Content-Security-Policy: connect-src 'self'; default-src 'none'; font-src 'self' fonts.gstatic.com; frame-ancestors 'none'; img-src 'self' {request.scheme}://{request.domain}:9000/ https://secure.gravatar.com; referrer origin-when-cross-origin; reflected-xss block; script-src 'self'; style-src 'self' fonts.googleapis.com
Vary: Accept-Encoding
```

Which we obviously don't want values like ``{request.scheme}://{request.domain}:9000/``.

.gitignore
```diff
@@ -23,3 +23,5 @@ dev/example.sql
 warehouse/.commit
 warehouse/static/components
 warehouse/static/dist
+
+tags
```
dstufft: What's this about?

dstufft:
Two comments, otherwise looks good to me.

warehouse/csp.py
demianbrecht: Ah cool, will change.

.gitignore
demianbrecht: I use per-project ctags. `tags` is the default output file.

demianbrecht:
Comments addressed and added a test to guard against regression of the csp formatting for development purposes.

dstufft:
Looks great, thanks!

Issue #1009, opened 2016-03-06
Allow OS X Wheels


dstufft:
We need to allow OS X wheels to be uploaded.

njsmith:
This is somewhat complicated by the fact that there are a *lot* of different valid OS X platform tags.

dstufft:
Here's a snippet that I *think* will generate all possible OS X platform tags:

```python
["_".join(["macosx", v, a]) for v, a in itertools.product(["10_{}".format(i) for i in range(12)], ["ppc", "ppc64", "i386", "x86_64", "fat", "fat32", "fat64", "intel", "universal"])]
```

A downside of this, is that it also generates some *invalid* OS X platform tags. For example, there was no such thing as a ``intel`` or ``i386`` or ``x86_64`` binary on 10.0 (and others). We could just ignore that as a mildly amusing edge case where someone could upload a wheel that doesn't actually make sense and couldn't actually exist in real life (and which pip won't actually use), or we could try to complicate this a lot more and only generate ``ppc`` on 10.0 through to 10.6, ``ppc64`` on 10.5 (and maybe 10.4, though it only partially worked there), ``i386`` on 10.4+, and ``x86_64`` on 10.5+ (maybe 10.4+, but it only partially worked on 10.4). Then we'd also have to allow the multi architecture tags based on what archiectures are contained within them. I'm not sure it's worth it to actually do that though versus just allowing all of the architectures for all of the versions through.

This would also of course bake a maximum OS X version into Warehouse which would need to be bumped whenever OS X releases a new version. This could be a bit annoying, though it might also be a good thing, if they drastically alter something in 10.12 so that we need to hold of on allowing them, that will be baked in right away, but I'm not sure that there is much of a chance of that actually happening. We could work around it by using ``range(99)`` or something, or by adding more runtime logic so that ``macosx`` wheels use something like:

```python
re.match("macosx_10_[0-9]+_(ppc|ppc64|...")
```

Dunno!

njsmith:
The approach currently in #1011 makes the decisions to:
* don't bother trying to restrict to valid (version, architecture) tuples
* do restrict to valid architectures (might at least catch typos like "itnel")
* do use a regexp to allow 10.12, 10.13, etc.

Ideally I guess this code could be shared with pip but meh...

xavfernandez:
> Ideally I guess this code could be shared with pip but meh...

Well, I think pep425 tags implementation will certainly end up in `packaging` package so this could be done :)

Issue #1008, opened 2016-03-06
Allow manylinux1 wheels


dstufft:
We need to allow the new ``manylinux1`` wheels to be uploaded.

njsmith:
These are pretty simple -- the only two platform tags allowed are exactly `manylinux1_x86_64` and `manylinux1_i686`.

PR #1011, opened 2016-03-06
Add support for OS X and manylinux platform tags

njsmith:
Also removed "win-amd64" and "win-ia64" from the whitelist, since they
violate PEP 425.

Fixes: gh-1008
Fixes: gh-1009

### DIFF
--- a/warehouse/legacy/api/pypi.py
+++ b/warehouse/legacy/api/pypi.py
@@ -46,9 +46,30 @@ MAX_FILESIZE = 60 * 1024 * 1024  # 60M
 MAX_SIGSIZE = 8 * 1024           # 8K
 
 
-ALLOWED_PLATFORMS = {
-    "any", "win32", "win-amd64", "win_amd64", "win-ia64", "win_ia64",
+# Wheel platform checking
+# These platforms can be handled by a simple static list:
+_allowed_platforms = {
+    "any",
+    "win32", "win_amd64", "win_ia64",
+    "manylinux1_x86_64", "manylinux1_i686",
 }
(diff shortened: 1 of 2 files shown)


dstufft:
You have linting errors here.

njsmith:
Fixed, tests are passing

warehouse/legacy/api/pypi.py
```diff
@@ -786,11 +807,12 @@ def file_upload(request):
...
-            if set(plats) - ALLOWED_PLATFORMS:
-                raise _exc_with_message(
-                    HTTPBadRequest,
-                    "Binary wheel for an unsupported platform.",
-                )
+            for plat in plats:
+                if not _valid_platform_tag(plat):
+                    raise _exc_with_message(
+                        HTTPBadRequest,
+                        "Binary wheel for an unsupported platform.",
```
ogrisel: I think it would be helpful for the users to report the offending paltform tag(s) in the exception message.

dstufft: I agree, though that would probably require building a dictionary that maps platform tag to Valid/Invalid and then doing an all() on the values of that dictionary or something.

@njsmith If you don't mind making this change in this PR, I think that would be a great addition.

ogrisel: My suggestion was just to do something like:

```python
raise _exc_with_message(
    HTTPBadRequest,
    "Binary wheel %s for an unsupported platform: %s" % (filename, plat),
)
```

njsmith:
Improved error message and tests are passing -- ready to go I think.

Issue #707, opened 2015-09-29
Rate limit login attacks


dstufft:
We need to enable rate limiting on the log in attempts so people cannot easily brute force them.

miohtama:
I'm not a Warehouse contributor (yet), but below related things.

* Here is Pyramid code I use in another project: https://gist.github.com/miohtama/86c86b4caa61a5615e44

* Background story:  http://opensourcehacker.com/2014/07/09/rolling-time-window-counters-with-redis-and-mitigating-botnet-driven-login-attacks/

*  More about the topic: http://operationssecurity.org/en/latest/user/index.html#brute-force-login-prevention

demianbrecht:
I'm going to be taking a stab at authentication/registration and obviously falls in that ballpark. I'm thinking that rate limiting isn't something that we want to do (unless there's a technique that I'm missing).

If the account in question is used as the key by which we're rate limiting, it's easily possible to effectively DoS a legitimate user. Even if an account lockout can be reset by a password reset via email, it's still a painful process for the user.

If we use the source address as the key, then we're potentially DoSing all users behind a proxy.

Rather than rate limit, I'm wondering if the form should use either a captcha or a secret question to which an answer is given during registration (although that would pose a problem for account migration) when attempts >= N (3 seems pretty common).

In addition to the captcha/secret question, a delay could be added to the login flow, which would drastically increase the time required for automated attacks.

Thoughts?

miohtama:
There are three ways of rate limiting for login, sign up and password reset

* Limit by IP address. This is pretty much useless as in future IPv6 gives attackers basically endless addresses and you need fancy way to upscale address matching to network block http://serverfault.com/questions/631160/banning-ipv6-addresses 

* Limit by username: this prevents spearhead attacks against individual users. Show CAPTCHA after N attempts, force cooldown after M attempts. This is the scenario where delay can help. However in the contemporary world of cyber threads this is rarely the most common attack method.

* Global attempt threshold: require CAPTCHA for all attempts (even for first login/password) if the system attempt threshold excess normal values (which one obtains from active system monitoring). In this case the system is most likely under a botnet attack and attackers are going through email (username)/password combo lists bought from blackmarkets and having thousands of IPs in their possession.  The attackers just cycle to the next email (username) / password / IP combo for every attempt so cooling down cannot help here. This is what bad guys tend to do against high value targets which I would see PyPi to be. It's likely millions of different username combos from thousands of IPs.

[...]

Issue #322, opened 2014-05-20
Warehouse needs a register page


toumorokoshi:
This is a proposal to add a register page to warehouse. The base functionality would mirror what's provided on the current pypi register page:

https://pypi.python.org/pypi?%3Aaction=register_form

karan:
+1 to "Sign Up with Github".

demianbrecht:
FWIW, having the ability to register an arbitrary authentication policy would be super helpful for those planning to adopt warehouse behind a corporate firewall and similar environments. In these cases, it might be helpful to be able to strip /all/ auth mechanisms other than some custom method.

dstufft:
This should actually be completed now.

PR #1012, opened 2016-03-06
Add user registration functionality

demianbrecht:
New items:
- Recaptcha service (conditionally functional, dependent on env vars)
- forms.RegistrationForm
- Reusable recaptcha template

Also added the responses lib as a dependency for mocking recaptcha tests

Features:
- [x] reCaptcha service
- [x] Reusable reCapatcha template
- [x] ``forms.RegistrationForm``
- [x] ``views.register``

Review items:
- [x] Address SSL verification issue
- [x] Expose reCaptcha keys through make targets and docker-compose
- [x] Move from base ``requests.post`` to a thread local session implementation
- [x] Handle timeouts & other lower level socket errors in recaptcha.py
- [x] Security review on session handling in registration handler
- [x] Bottom load recaptcha scripts
- [x] Create a threadlocal requests session 
- [x] Configure sane timeouts for ``requests.Session`` instance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,10 +38,10 @@ build:
 	touch .state/docker-build
 
 serve: .state/docker-build
-	docker-compose up
+	RECAPTCHA_SITE_KEY=$(RECAPTCHA_SITE_KEY) RECAPTCHA_SECRET_KEY=$(RECAPTCHA_SECRET_KEY) docker-compose up
 
 debug: .state/docker-build
-	docker-compose run --service-ports web
+	RECAPTCHA_SITE_KEY=$(RECAPTCHA_SITE_KEY) RECAPTCHA_SECRET_KEY=$(RECAPTCHA_SECRET_KEY) docker-compose run --service-ports web
 
 tests:
 	docker-compose run web env -i ENCODING="C.UTF-8" bin/tests --dbfixtures-config tests/dbfixtures.conf $(T) $(TESTARGS)
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,8 @@ commands you can use:
 
     $ # Start up a local environment
     $ make serve
+    $ # Start up a local environment in debug mode (pdb enabled)
+    $ make debug
     $ # Initialize the database and fill it with test data
     $ make initdb
     $ # Run the tests
@@ -39,6 +41,10 @@ commands you can use:
     $ make docs
     $ # Run the various linters
     $ make lint
+
+.. note:: reCaptcha is featured in authentication and regisration pages. To
+          enable it, pass ``RECAPTCHA_SITE_KEY`` and ``RECAPTCHA_SECRET_KEY``
+          through to ``serve`` and ``debug`` targets.
 
 
 Discussion
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ commands you can use:
     $ # Run the various linters
     $ make lint
 
-.. note:: reCaptcha is featured in authentication and regisration pages. To
+.. note:: reCaptcha is featured in authentication and registration pages. To
           enable it, pass ``RECAPTCHA_SITE_KEY`` and ``RECAPTCHA_SECRET_KEY``
           through to ``serve`` and ``debug`` targets.
 
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -9,6 +9,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 
 import wtforms
 import wtforms.fields.html5
@@ -33,6 +34,21 @@ class CredentialsMixin:
     def __init__(self, *args, user_service, **kwargs):
         super().__init__(*args, **kwargs)
         self.user_service = user_service
+
+
+# XXX: This is a naive password strength validator, but something that can
+# easily be replicated in JS for client-side feedback.
+# see: https://github.com/pypa/warehouse/issues/6
+PWD_MIN_LEN = 8
+PWD_RE = re.compile(r"""
+^                                                       # start
+(?=.*[A-Z]+.*)                                          # >= 1 upper case
+(?=.*[a-z]+.*)                                          # >= 1 lower case
+(?=.*[0-9]+.*)                                          # >= 1 number
+(?=.*[.*~`\!@#$%^&\*\(\)_+-={}|\[\]\\:";'<>?,\./]+.*)   # >= 1 special char
+.{""" + str(PWD_MIN_LEN) + """,}                        # >= 8 chars
+$                                                       # end
+""", re.X)
 
 
 class RegistrationForm(CredentialsMixin, forms.Form):
@@ -79,6 +95,14 @@ class RegistrationForm(CredentialsMixin, forms.Form):
             # TODO: log error
             # don't want to provide the user with any detail
             raise wtforms.validators.ValidationError("Recaptcha error.")
+
+    def validate_password(self, field):
+        if not PWD_RE.match(field.data):
+            raise wtforms.validators.ValidationError(
+                "Password must contain 1+ upper case letter, 1+ lower case "
+                "letter, 1+ number, 1+ special character and be at least "
+                "%d characters in length" % PWD_MIN_LEN
+            )
 
 
 class LoginForm(CredentialsMixin, forms.Form):
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -11,12 +11,12 @@
 # limitations under the License.
 
 import wtforms
+import wtforms.fields.html5
 
-from warehouse import forms
+from warehouse import forms, recaptcha
 
 
-class LoginForm(forms.Form):
-
+class CredentialsMixin:
     username = wtforms.StringField(
         validators=[
             wtforms.validators.DataRequired(),
@@ -30,19 +30,67 @@ class LoginForm(forms.Form):
         ],
     )
 
-    def __init__(self, *args, login_service, **kwargs):
+    def __init__(self, *args, user_service, **kwargs):
         super().__init__(*args, **kwargs)
+        self.user_service = user_service
 
-        self.login_service = login_service
+
+class RegistrationForm(CredentialsMixin, forms.Form):
+    password_confirm = wtforms.PasswordField(
+        validators=[
+            wtforms.validators.DataRequired(),
+            wtforms.validators.EqualTo(
+                "password", "Passwords must match."
+            ),
+        ],
+    )
+
+    full_name = wtforms.StringField()
+
+    email = wtforms.fields.html5.EmailField(
+        validators=[
+            wtforms.validators.DataRequired(),
+            wtforms.validators.Email(),
+        ],
+    )
+
+    g_recaptcha_response = wtforms.StringField()
+
+    def __init__(self, *args, recaptcha_service, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.recaptcha_service = recaptcha_service
 
     def validate_username(self, field):
-        userid = self.login_service.find_userid(field.data)
+        if self.user_service.find_userid(field.data) is not None:
+            raise wtforms.validators.ValidationError(
+                "Username exists.")
+
+    def validate_email(self, field):
+        if self.user_service.find_by_email(field.data) is not None:
+            raise wtforms.validators.ValidationError("Email exists.")
+
+    def validate_g_recaptcha_response(self, field):
+        # do required data validation here due to enabled flag being required
+        if self.recaptcha_service.enabled and len(field.data) <= 0:
+            raise wtforms.validators.ValidationError("This field is required.")
+
+        try:
+            self.recaptcha_service.verify_response(field.data)
+        except recaptcha.RecaptchaError:
+            # TODO: log error
+            # don't want to provide the user with any detail
+            raise wtforms.validators.ValidationError("Recaptcha error.")
+
+
+class LoginForm(CredentialsMixin, forms.Form):
+    def validate_username(self, field):
+        userid = self.user_service.find_userid(field.data)
 
         if userid is None:
             raise wtforms.validators.ValidationError("Invalid user.")
 
     def validate_password(self, field):
-        userid = self.login_service.find_userid(self.username.data)
+        userid = self.user_service.find_userid(self.username.data)
         if userid is not None:
-            if not self.login_service.check_password(userid, field.data):
+            if not self.user_service.check_password(userid, field.data):
                 raise wtforms.validators.ValidationError("Invalid password.")
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -71,8 +71,8 @@ class RegistrationForm(CredentialsMixin, forms.Form):
 
     def validate_g_recaptcha_response(self, field):
         # do required data validation here due to enabled flag being required
-        if self.recaptcha_service.enabled and len(field.data) <= 0:
-            raise wtforms.validators.ValidationError("This field is required.")
+        if self.recaptcha_service.enabled and not field.data:
+            raise wtforms.validators.ValidationError("Recaptcha error.")
 
         try:
             self.recaptcha_service.verify_response(field.data)
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -66,7 +66,7 @@ class RegistrationForm(CredentialsMixin, forms.Form):
                 "Username exists.")
 
     def validate_email(self, field):
-        if self.user_service.find_by_email(field.data) is not None:
+        if self.user_service.find_userid_by_email(field.data) is not None:
             raise wtforms.validators.ValidationError("Email exists.")
 
     def validate_g_recaptcha_response(self, field):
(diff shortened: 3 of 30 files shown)


demianbrecht:
Some tests are broken, a few still need to be written, and some finishing touches have to be made yet, but I figured this would be as good a time as any to get some feedback on current progress.

demianbrecht:
This work is aimed to be partially implement #322 (email verification and Github-based registration still required) and #707 (captcha will still be required for login).

warehouse/accounts/forms.py
```diff
@@ -30,19 +30,67 @@ class LoginForm(forms.Form):
...
+
+    def validate_g_recaptcha_response(self, field):
+        # do required data validation here due to enabled flag being required
+        if self.recaptcha_service.enabled and not field.data:
+            raise wtforms.validators.ValidationError("Recaptcha error.")
+
+        try:
+            self.recaptcha_service.verify_response(field.data)
+        except recaptcha.RecaptchaError:
+            # TODO: log error
```
dstufft: If this is a "user did not type the recaptcha correctly" style error, then we probably don't want to log anything here (or at the most, we'd want debug level logging, not error level). Our errors get logged to sentry and in general errors should be actionable for us/operators.

warehouse/accounts/services.py
```diff
@@ -78,9 +92,10 @@ def check_password(self, userid, password):
 
     def create_user(self, username, name, password, email,
                     is_active=False, is_staff=False, is_superuser=False):
+
         user = User(username=username,
                     name=name,
-                    password=password,
+                    password=self.hasher.encrypt(password),
```
dstufft: oof, did I really have this saving the unencrypted password to the database?

warehouse/recaptcha.py
```diff
@@ -0,0 +1,137 @@
...
+        payload = {
+            "secret": settings["secret_key"],
+            "response": response,
+        }
+        if remote_ip is not None:
+            payload["remoteip"] = remote_ip
+
+        # TODO: maybe add a session service? there doesn't seem to be anything
+        # calling out to external services atm though
+        # TODO: turn verify back on once the certs are fixed
```
demianbrecht: Will remove this line (verification is back on)

warehouse/accounts/views.py
```diff
@@ -169,3 +169,42 @@ def logout(request, redirect_field_name=REDIRECT_FIELD_NAME):
...
+
+
+@view_config(
+    route_name="accounts.register",
+    renderer="accounts/register.html",
+    decorator=[csrf_protect("accounts.register"), uses_session],
+)
+def register(request, _form_class=forms.RegistrationForm):
+    if request.authenticated_userid is not None:
+        return HTTPSeeOther("/")
```
dstufft: I guess this raises the question of what exactly should a /register/ page *do* if there is a logged in user... I suspect redirecting to something is a reasonable thing though. This should use ``request.route_path("index")`` though to avoid hardcoding the path.

warehouse/accounts/views.py
```diff
@@ -169,3 +169,42 @@ def logout(request, redirect_field_name=REDIRECT_FIELD_NAME):
...
+
+    user_service = request.find_service(IUserService, context=None)
+    recaptcha_service = request.find_service(name="recaptcha")
+    recaptcha_service.add_to_csp_policy()
```
dstufft: I feel like this would be cleaner if it was ``recaptcha_service.add_to_csp_policy(request.find_service(name="csp"))`` and then the ``add_to_csp_policy`` didn't need to do anything except mutate the dictionary.

warehouse/accounts/views.py
```diff
@@ -169,3 +169,42 @@ def logout(request, redirect_field_name=REDIRECT_FIELD_NAME):
...
+
+    if request.method == "POST" and form.validate():
+        user = user_service.create_user(
+            form.username.data, form.full_name.data, form.password.data,
+            form.email.data
+        )
+
+        headers = remember(request, user.id)
+
+        request.session.new_csrf_token()
```
dstufft: We should probably extract the session invalidation/csrf token/remember() call out of the login view and just call that function both from the login view and this view. Just a module scoped function in ``warehouse/accountsviews.py`` with:

```python
def _login_user(request, user_id):
    # We need to protect against session fixation attacks, so make sure
    # that we create a new session (which will cause it to get a new
    # session identifier).
    if (request.unauthenticated_userid is not None and
            request.unauthenticated_userid != user_id):
        # There is already a userid associated with this request and it is
        # a different userid than the one we're trying to remember now. In
        # this case we want to drop the existing session completely because
        # we don't want to leak any data between authenticated userids.
        request.session.invalidate()
    else:
        # We either do not have an associated userid with this request
        # already, or the userid is the same one we're trying to remember
        # now. In either case we want to keep all of the data but we want
        # to make sure that we create a new session since we're crossing
        # a privilege boundary.
        data = dict(request.session.items())
        request.session.invalidate()
        request.session.update(data)

    # Remember the userid using the authentication policy.
    headers = remember(request, user_id)

    # Cycle the CSRF token since we've crossed an authentication boundary
    # and we don't want to continue using the old one.
    request.session.new_csrf_token()

    return headers
```

should be fine I think.

warehouse/csp.py
```diff
@@ -32,13 +31,15 @@ def content_security_policy_tween(request):
 
 
 def csp_factory(_, request):
-    return copy.deepcopy(request.registry.settings.get("csp", {}))
+    return collections.defaultdict(
+        list, request.registry.settings.get("csp", {})
```
dstufft: I think you still want this to be a deepcopy else modifying the lists involved will persist I think.

warehouse/csp.py
```diff
@@ -32,13 +31,15 @@ def content_security_policy_tween(request):
...
+    )
 
 
 def includeme(config):
     config.register_service_factory(csp_factory, name="csp")
     # Enable a Content Security Policy
-    config.add_settings({
+    config.add_settings(collections.defaultdict(list, {
```
dstufft: Given the defaultdict in the csp factory, I don't think this is needed is it?

warehouse/templates/includes/input-recaptcha.html
```diff
@@ -0,0 +1,14 @@
+{% if request.find_service(name="recaptcha").enabled %}
+  {% block top_script %}
+  <script src="www.google.com/recaptcha/api.js"></script>
```
dstufft: I guess recaptcha doesn't work if it's not included at the top? Otherwise it'd be preferable to include it at the end of the body to prevent blocking the rendering.

dstufft:
Overall looks pretty good, I made some comments on the code that I saw. As far as the TODO items, we shouldn't use a module scoped session. IIRC requests sessions are not thread safe (and in general, module scoped connection pools are not really great anyways). This should either be scoped to the service (so it'd be created in the service's ``__init__()`` method) or it should be created in the ``includeme()`` and stashed in the registry and passed into the service when it's created (see ``warehouse/db.py`` for an example). Another option is to register a HTTP Client service so that multiple different things can share the requests session (might not be worth it though).

In any case, since the sessions aren't compeltely threadsafe, any use of session that isn't scoped to a specific request will probably need a wrapper around it so that a single session is actually a session per thread. 

/cc @sigmavirus24 @lukasa since they know requests behavior here better than me :)

demianbrecht:
Yeah, the point about moving to a module-scope instance really should have been "more generalized handling of sessions", but good to know about threading issues.

I'll follow the same pattern as warehouse/db.py for sessions. I'd considered an HTTP client service initially, but thought it would be a tad overkill.

Thanks for the review! Will get the comments as soon as I have time.

Lukasa:
The thread-safety of `Session` objects is a tricky one. The things that put their thread-safety at risk are primarily the cookie dictionary and the (hateful) 301 cache (there are some problems with users manipulating `Session` properties without being careful about synchronising access, but I care less about that). If you aren't using either of those, you can actually force both of them to be no-ops and obtain the thread-safet you need.

However, a `Session` in thread-local storage is a really good idea. Having a `Session` live as long as possible is a good idea (the connection pooling works best that way). Thread-local storage can do a very good job of that.

dstufft:
Hmm, maybe requests-toolbelt should get a ``ThreadLocalSession``.

Lukasa:
@dstufft Not sure that would be all that useful: how would it work in a way that is more useful than just using the standard thread local storage logic in Python?

sigmavirus24:
I have the same question as @Lukasa but wouldn't be opposed to it if we could figure out how it's an improvement.

dstufft:
Benefits:

* It'd have the same interface as a requests ``Session`` allowing it to be used transparently anyplace (or most anyplace) something depends on a session. Which means the code doesn't have to be aware it's using a thread local.
* Thread local storage doesn't have a creation function, so that means anytime you attempt to use the the thread's session, you have to first check to see if the thread even has a session created yet, and if not create it.

Essentially, I'd think it'd basically be:

```python
import threading
import requests

class ThreadLocalSession:
    def __init__(self):
        self._threadlocal = threading.local()
   
    def _get_session(self):
        try:
            return self._threadlocal.session
        except AttributeError:
            self._threadlocal.session = requests.Session()

    def __getattr__(self, name):
        return getattr(self._get_session(), name)
```

You'd need something that would let you have settings on the Session though, so maybe the ``__init__()`` would accept keyword arguments that would get ``setattr()``'d onto a new session whenever it was created. Another option would be, in the spirit of attempting to mimic the ``Session`` API, add a dict that stores things like ``ThreadLocalSession().verify = False`` and propagates that to all of the sessions somehow.

Dunno!

warehouse/accounts/forms.py
demianbrecht: Ah woops, yeah, I meant debug level log of the exception here :)

warehouse/accounts/services.py
demianbrecht: heh yep, there was even a test for it ;)

warehouse/accounts/views.py
demianbrecht: Ah yeah, thanks for the catch on the hardcoded path. I was tossing up either index or profile pages, either make sense to me. Easy enough to change later if needed though.

warehouse/accounts/views.py
demianbrecht: Actually, maybe a little cleaner would be to implement an `merge` method to the csp service and then just have the recpatcha policy as a constant. With that, from the viwe it would look like:

`request.find_service(name="csp").merge(recaptcha_service.CSP_POLICY)`

Thoughts?

warehouse/accounts/views.py
demianbrecht: Yeah, was considering that. Makes sense to me, will do.

warehouse/csp.py
demianbrecht: Whoops, yeah you're right. Updated.

warehouse/csp.py
demianbrecht: It is now that the csp factory has been reverted back to `copy.deepcopy`.

demianbrecht: I take that back. the `defaultdict` doesn't hold when applied to `request.registry.settings` because of the `update`. So no, this doesn't have to be a `defaultdict` (will change it), but I'll need to update the factory function.

sigmavirus24:
>  Another option would be, in the spirit of attempting to mimic the Session API, add a dict that stores things like ThreadLocalSession().verify = False and propagates that to all of the sessions somehow.

Perhaps we should discuss this in an issue over on the toolbelt, but I'm not sure there's a great way to do what you're asking for here.

warehouse/http.py
```diff
@@ -0,0 +1,27 @@
+import threading
+import requests
+
+
+class ThreadLocalSessionFactory(threading.local):
```
demianbrecht: This doesn't actually give us what I was after: A persistent, thread local session, it will be a thread local, per-request session (so additional complexity that has no benefit). Will fix.

demianbrecht:
@dstufft: This should be functionally complete now.

Makefile
```diff
@@ -38,10 +38,10 @@ build:
 	touch .state/docker-build
 
 serve: .state/docker-build
-	docker-compose up
+	RECAPTCHA_SITE_KEY=$(RECAPTCHA_SITE_KEY) RECAPTCHA_SECRET_KEY=$(RECAPTCHA_SECRET_KEY) docker-compose up
```
dstufft: Are these actually needed? If ``RECAPTCHA_SECRET_KEY`` is defined in the environment, won't it automatically be inherited by the ``docker-compose up`` call?

README.rst
```diff
@@ -40,6 +42,10 @@ commands you can use:
     $ # Run the various linters
     $ make lint
 
+.. note:: reCaptcha is featured in authentication and regisration pages. To
```
dstufft: Typo here in "registration".

Makefile
dstufft: Urf, nevermind me. These are makefile variables not environment variables.

warehouse/accounts/views.py
```diff
@@ -118,7 +86,9 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME,
 
         # Now that we're logged in we'll want to redirect the user to either
         # where they were trying to go originally, or to the default view.
-        return HTTPSeeOther(redirect_to, headers=dict(headers))
+        return HTTPSeeOther(
+            redirect_to,
+            headers=dict(_login_user(request, userid)))
```
dstufft: Stylistically, I think it would be nicer to split this out so that you have a ``headers = _login_user(request, userid)`` line before this redirect. If only to make the actual step where a user is logged in more obvious and look less like a side effect of generating the redirect.

warehouse/http.py
```diff
@@ -0,0 +1,34 @@
+import threading
+import requests
+
+
+_local = threading.local()
```
dstufft: I think it would be better to have this be encapsulated inside of the ``ThreadLocalSessionFactory`` instead of having it be a global thread local. I'd just move this line into ``ThreadLocalSessionFactory.__init__``. That should still give thread local sessions, but have separate pools per instance of ``ThreadLocalSessionFactory``, which by default we only create one of, but there's no reason we couldn't do more.

warehouse/recaptcha.py
```diff
@@ -0,0 +1,142 @@
...
+)
+
+
+class Service:
+    def __init__(self, request):
+        self.request = request
+
+    @property
+    def csp_policy(self):
+        if not self.enabled:
```
dstufft: I'm wondering if it wouldn't make sense to just remove this conditional and just always include this CSP on these pages. The only time this should skip the CSP is in development, and it doesn't seem harmful to have the CSP policy include this in development.

warehouse/templates/includes/input-recaptcha.html
```diff
@@ -0,0 +1,29 @@
...
+    </ul>
+    {% endif %}
+  {% endif %}
+{%- endmacro %}
+
+{% macro recaptcha_src() -%}
+  {% if request.find_service(name="recaptcha").enabled %}
+    <script type="text/javascript">
+      var config = new Object();
+      config.sitekey = "{{ request.registry.settings["recaptcha"]["site_key"] }}";
```
dstufft: I think that having this inline script here is generally not the way we want to go about this. We should be able to do something like:

```
<script
    id="recaptcha-js"
    src="https://www.google.com/recaptcha/api.js?onload=renderCaptcha&render=explicit"
    data-site-key="{{ request.registry.settings["recaptcha"]["site_key"] }}"
    async
    defer>
</script>
```

Then in our static JS files we can do something like:

```javascript
renderCaptcha = function() {
    var config = new Object();
    config.sitekey = $("script#recaptcha-js").data("site-key");
    grecaptcha.render($("#recaptcha-container"), config);
};
```

This should make it so that we remove all of the inline scripts (which protects against XSS, since we won't need unsafe-inline in the CSP policy) and doesn't rely on using HTML escaping inside of a javascript block.

dstufft:
Added some comments about changes, this is also failing HTML linting... but one of the changes I mentioned should resolve that issue.

README.rst
demianbrecht: Fixed.

warehouse/accounts/views.py
demianbrecht: Yeah, I was on the fence about that. Generally I prefer to use local variables only if they're used in multiple places (and it doesn't make the code look bad), but agreed on the side effect, so will change.

warehouse/templates/includes/input-recaptcha.html
demianbrecht: Whoops, something I totally intended to change before pushing for review, thanks for the catch.

Why use anonymous functions? FWIW out of habit, I never use anonymous functions anymore because of browser-based debugging. Chrome developer tools would trace [anonymous] or something similar rather than something meaningful. So even when defining a method bound to an object, I'd do something like:

    myobject.prototype = {
        mymethod: function MYOBJECT_mymethod() {
            // do something magical
        }
    };

Maybe things have gotten a little better since I last did anything full stack, but that's what I remember doing in the past.

warehouse/recaptcha.py
demianbrecht: I'm easy either way, no harm either way really.

warehouse/http.py
demianbrecht: Yeah, I was on the fence about this one. The thing that I like about the global is that it's a little more resilient to external changes then. i.e. If someone was to set `reify` to `False` for some reason, then we'd start getting a new session per call. Minor as it really /shouldn't/ happen, but that was the reasoning. In any event, I've updated it.

demianbrecht:
@dstufft: Review issues addressed and tests are passing.

PR #1031, opened 2016-03-18
Add recaptcha to login

demianbrecht:
This should address #707

### DIFF
--- a/tests/unit/accounts/test_forms.py
+++ b/tests/unit/accounts/test_forms.py
@@ -18,19 +18,69 @@ from warehouse.accounts import forms
 from warehouse import recaptcha
 
 
-class TestLoginForm:
+class RecaptchaTestMixin:
+    def test_recaptcha_disabled(self):
+        form = self.form(
+            data={"g_recpatcha_response": ""},
+            user_service=pretend.stub(),
+            recaptcha_service=pretend.stub(
+                enabled=False,
+                verify_response=pretend.call_recorder(lambda _: None),
+            ),
+        )
+        assert not form.validate()
+        # there shouldn't be any errors for the recaptcha field if it's
+        # disabled
+        assert not form.g_recaptcha_response.errors
+
+    def test_recaptcha_required_error(self):
+        form = self.form(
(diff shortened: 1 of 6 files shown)


dstufft:
I don't think we want a captcha on every single login, that would be somewhat annoying I think. It would be better if it only required a captcha after N failed login attempts for that IP address or for that user in the last Y minutes (rolling).

It's also not a complete solution because we also have the basic auth which is used in automated systems, so we probably want the two systems to share attempt counters, and have it so that for basic auth if you get too many failed attempts it just rejects your login, and for regular auth it shows a captcha.

demianbrecht:
Yep, missing basic auth throws a bit of a wrench into this for sure. Captcha doesn't help much in that scenario and I imagine that's the path that would most likely be targeted by attackers.

Per-IP is going to be an issue with IPv6 (noted [here](https://github.com/pypa/warehouse/issues/707#issuecomment-190015037)). Also, I'm not sure how fond I am of per-IP blocks on auth given it has the potential to DoS legitimate users. Is this approach one used in PyPI?

So, I'll keep what's done so far, but I'll also add global system, per user and potentially per-IP sliding windows that will either trigger recaptcha in the FE or lock out accounts in the BE.

Sound reasonable?

dstufft:
Right now in PyPI we have separate counters for user and for IP address. I think it's probably OK to drop the per IP address if we replace it with a global threshold that starts to trigger a captcha for all logins if the global failed logins reach some sort of threshold. We would (probably?) ignore that in the basic auth, but that's OK because longer term I plan to deprecate the basic auth and move to something generated by the system with high entropy (API key, Client TLS Certificate? Something Else?!) that shouldn't be vulnerable to that and at that point the via-basic-auth bypass for the global failed logins should be closed.

dstufft:
To be clear, I'd suggest that we *only* ignore the global threshold when authenticating with basic auth, the per user block should not be ignored.

di:
#707 was fixed in #1530 so this can be closed.

Thanks for taking a shot at this @demianbrecht!

PR #1034, opened 2016-03-18
Relax CSP for XML pages

di:
This PR relaxes the CSP on XML pages to allow browsers such as Chrome to inline styles and images.

Before:
screen shot 2016-03-18 at 6 25 17 pm

After:
screen shot 2016-03-18 at 6 24 27 pm

### DIFF
--- a/tests/unit/rss/test_views.py
+++ b/tests/unit/rss/test_views.py
@@ -12,11 +12,23 @@
 
 import datetime
 
+import pretend
+
 from warehouse.rss import views as rss
 from ...common.db.packaging import ProjectFactory, ReleaseFactory
 
 
 def test_rss_updates(db_request):
+    db_request.find_service = pretend.call_recorder(
+        lambda *args, **kwargs: pretend.stub(
+            enabled=False,
+            csp_policy=pretend.stub(),
+            merge=lambda _: None,
+        )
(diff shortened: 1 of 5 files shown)


dstufft:
I'm a bit concerned about the allowance of ``data:`` here. If someone manages to get unescaped output onto this page that renders an image of some kind that could possibly allow a giffar style attack. Maybe this isn't very likely, I'd have to think about it, but I ask because maybe it doesn't look very bad without adding ``data:`` to the img-src and just the ``unsafe-inline`` for styles would be enough?

di:
It wouldn't look too bad, it's only used it render the arrows.

di:
@dstufft I've updated the PR. Here's what it looks like without the `img-src` policy `data:`:

screen shot 2016-03-20 at 11 03 22 am

dstufft:
So coming back around to look at this again, and I think that the ``XMLService`` may be overkill here. All we really need is a static dictionary somewhere that can be accessed. Invoking the service machinery for this seems to be a waste. Any reason not to just stick:

```python
XML_CSP = {"style-src": ["'unsafe-inline'"]}
```

somewhere and then reference that?

di:
@dstufft Addressed.

PR #1038, opened 2016-03-21
Upgrade ES version to latest

di:
An alternative to #1037

(Closes #1037)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ rabbitmq:
   image: rabbitmq:latest
 
 elasticsearch:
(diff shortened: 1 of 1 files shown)


dstufft:
This should probably pin to at least ``2`` explicitly, if not ``2.2``.

dstufft:
(For the record, 2.2.1 is what we're using in production right now for Warehouse)

di:
@dstufft I amended it to pin to `2.2`, would you prefer `2.2.1`?

dstufft:
Nah, ``2.2`` should be fine.

PR #1044, opened 2016-03-30
Use pip's hashing mode in our requirements files

dstufft:
In any of the requirement files that we use for deployment, we'll use the hashing mode in pip to ensure that we're getting the exact files that we expect and not compromised files.


erikrose:
If that reqs file is used anywhere but in our production deployment, we might add hashes for alternative architectures (for binary things like gevent), so contribs on other platforms don't get left in the lurch. https://pypi.python.org/pypi/hashin/ can help with this.

dstufft:
I actually used hashin to generate all of this! Although the only *supported* way to run Warehouse is on Heroku or in our docker containers, both of which run Linux.

PR #1070, opened 2016-04-16
Move all package.json devDependencies to dependencies

dstufft:
Heroku only installs regular dependency in a Node.js application, however we need all of our dependencies to build our static files, so we'll just make them all regular dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,21 +3,16 @@
     "node": "5.5.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


ryanhiebert:
Yeah, I wish there was a separate spot for build dependencies...

dstufft:
Yea that would be great!

PR #1164, opened 2016-05-04
Fix #1163

### LINKED ISSUE #1163: error 404 when searching for a package name with underscores
when I try to search for a package name which contains underscores I get a 404 error

example: https://pypi.io/search/?q=test_test

di:
This PR fixes #1163 by only returning a 404 when there is a `page_count` (e.g., there are results) and the `page_num` exceeds it. When there are no results, it should not 404.

Additionally, when testing this, I discovered that it is possible to raise a 500-level error if the search term is such that it would not generate any "suggestions", for example, just `_`. This PR fixes that bug as well.

### DIFF
--- a/tests/unit/utils/test_paginate.py
+++ b/tests/unit/utils/test_paginate.py
@@ -47,14 +47,17 @@ class FakeResult:
 
 class FakeSuggestResult(FakeResult):
 
-    def __init__(self, data, total, options):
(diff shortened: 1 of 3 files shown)


dstufft:
✨ We have that 500 error in sentry actually, kept forgetting to file a bug about it.!

PR #1165, opened 2016-05-04
Use a 302 redirect instead of a 307 redirect

dstufft:
Not all of our downstreams currently understand a 307 redirect, so we'll instead use a 302 which is more common but slightly semantically vague.

### DIFF
--- a/tests/unit/legacy/test_views.py
+++ b/tests/unit/legacy/test_views.py
@@ -13,7 +13,7 @@
 import pretend
 import pytest
 
-from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPNotFound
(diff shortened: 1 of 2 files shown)


Sp1l:
This would be very helpful for FreeBSD ports!

Primarily for documentation purposes: It is OK to use https://pypi.io/packages links for ports users to download packages?

dstufft:
It's OK to use them, eventually they'll be the "real" links that get linked too from PyPI and the old ``pypi.python.org`` links will get redirected. The only downside is right now pypi.io is considered pre-production code so nobody is getting alerted if it goes down and I won't lose sleep to fix it if it's down etc. That should hopefully be going away in the near future though.

PR #1167, opened 2016-05-05
Fixes #1112

GrappigPanda:
CHANGE:
  - warehouse.raven.py:includeme() now hooks into the Celery errorhandler and 
    will log any errors at logging.ERROR level.

PR #1181, opened 2016-05-09
Log celery errors for Sentry

### LINKED ISSUE #1112: Errors in celery don't get sent to Sentry

di:
This is the patch from #1167 with tests added. Fixes #1112.

I can't actually verify that this works, mind taking a look @dstufft?

### DIFF
--- a/tests/unit/test_celery.py
+++ b/tests/unit/test_celery.py
@@ -21,12 +21,27 @@ from warehouse.config import Environment
 
 
 def test_configure_celery(monkeypatch):
-    configure = pretend.call_recorder(lambda: None)
+    client = pretend.stub()
+    getitem = pretend.call_recorder(lambda *a: client)
+    registry = pretend.stub(__getitem__=getitem)
+    configure = pretend.call_recorder(lambda: pretend.stub(registry=registry))
+    register_logger_signal = pretend.call_recorder(lambda x: None)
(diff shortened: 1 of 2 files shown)


dstufft:
I'll have to poke this a bit tomorrow, thanks!

dstufft:
Ok, finally got a chance to take a look at this (sorry!).

I think this looks mostly right, except that I'm not sure what the effect would be on calling this unconditionally in the ``includeme``, which will get used during non-celery code too. The Raven docs suggest using the ``on_configure`` hook of Celery to "encapsulate behavior". So I think we should move this to that?

I think all it should need to do is move the two lines of code over to [warehouse/celery.py#L28-L30](https://github.com/pypa/warehouse/blob/d57186c3b5f783e2235b7d5be610302adce23c6f/warehouse/celery.py#L28-L30) and make it look something like:

```python
@celeryd_init.connect
def _configure_celery(*args, **kwargs):
    config = configure()
    celery.register_logger_signal(config.registry["raven.client"])
    celery.register_signal(config.registry["raven.client"])
```

I'll try to take a poke at doing that and seeing if it works unless you feel like refactoring and beat me to it.

di:
@dstufft Thanks! I updated this PR.

dstufft:
Awesome, thanks!

Issue #702, opened 2015-09-26
Integrate search design


dstufft:
We need a page that will allow users to search via the web UI for projects to aid in the discovery of projects.

dstufft:
#739 implements a very basic version of this, but I know the design will require more features.

nlhkabu:
@dstufft I am going to remove the design tag for this and change the title to 'integrate' search page to indicate that my part is done

dstufft:
This will need things like:

* Faceting
* Sorting
* Pagination
* Number of total results

di:
@dstufft Looks like we already have pagination and total # of results, and #1004 adds sorting by release date.

Are there additional types of sorting we might want to support? Number of downloads might be useful.

For faceting, what else besides "Development Status"?

dstufft:
Number of downloads is a bit interesting, though it'll take some thought maybe. Right now we're regenerating the search index every 15 minutes or so, but in the future I plan to make a full regeneration a rare thing and just have it add/remove stuff from the current index based on what is actually changing. That means there won't be a great "hook" to trigger updating the download counts as downloads happen. That's not an insurmountable thing though, we could either just periodically regenerate the entire index, or have a task that looks for the N oldest projects which were last indexed at least X time ago and reindex those.

I would guess probably ordering by name wouldn't be the worst thing to do.

For faceting, I'm not sure! I guess maybe we want to make it possible to facet on any of the classifier values? That would effectively replace the old browse page then as well as add some real functionality to classifiers.

dstufft:
FWIW, you can see a list of all available classifiers here: https://pypi.python.org/pypi?%3Aaction=list_classifiers

nlhkabu:
The original design specified that each classifier should work as a facet, with only one group being open at a time (https://github.com/pypa/warehouse/issues/755)

![screenshot from 2016-03-06 07-43-32](https://cloud.githubusercontent.com/assets/3323703/13552968/371f45a2-e36f-11e5-819b-9c484b30fb56.png)

On the HTML side, the checkboxes should be nested (see https://github.com/pypa/warehouse/issues/754).

For the ordering, no. of downloads was definitely a consideration when I put together the design...

Another idea is to calculate package 'popularity' by working out how many other packages depend on that package.  Unless this is already how 'relevance' works??

dstufft:
Relevance is a full text search, it doesn't take into account anything special about the fact these are packages which depend on each other.

[...]

Issue #701, opened 2015-09-26
Incremental search indexing


dstufft:
Currently, the only support way to update the search index is via bulk indexing periodically. Ideally we'll have something that can subscribe to the changes made to the database and trigger a new update (or removal in the case of deletion) of a project in the Elasticsearch index.

robhudson:
Does the system in use (or sqlalchemy) have anything like the Django signals? That is how you would typically set something like this up on a Django project.

If I had some guidance on the signals part of this bug I could help implement the rest.

dstufft:
Yes! It's a tad bit more complicated than in Django, but you can see something similar already done at [warehouse/cache/origin/\_\_init\_\_.py#L20-L49](https://github.com/pypa/warehouse/blob/ce07436eb0384d860b9af52a767d7683fe92c1ee/warehouse/cache/origin/__init__.py#L20-L49). SQLAlchemy only tells you what has changed during a flush (which is distinct from committing the transaction) so you need to have one hook to record what things need reindexed, and another thing to actually trigger that.

ALTHOUGH, now that I think about it, since this would be done in Celery (or should be anyways) you only need the flush hook, because our Celery tasks automatically get buffered in memory and get sent _after_ the transaction is committed.

brainwane:
(In case @robhudson doesn't know: during the Warehouse team's current [grant-funded effort to improve Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), we're prioritizing getting it to the point where we can redirect pypi.python.org to pypi.org so the site is more sustainable and reliable; more details at [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap).)

Because I figure we'll want to get this sorted out before we flip the switch, I'm marking this for [the launch milestone](https://github.com/pypa/warehouse/milestone/1), but @ewdurbin can of course override me if I'm wrong.

brainwane:
In our [meeting this week](https://wiki.python.org/psf/PackagingWG/2018-02-26-Warehouse) we decided this belongs in a future milestone on [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) and does not have to block the launch.

[...]

PR #1182, opened 2016-05-10
Add search result ordering by download count

di:
This PR adds search result ordering by download count as discussed in #702.

### DIFF
--- a/tests/unit/cli/search/test_reindex.py
+++ b/tests/unit/cli/search/test_reindex.py
@@ -42,6 +42,7 @@ def test_project_docs(db_session):
                 "name": p.name,
                 "normalized_name": p.normalized_name,
                 "version": [r.version for r in prs],
+                "downloads": p.downloads,
             },
         }
         for p, prs in sorted(releases.items(), key=lambda x: x[0].name.lower())
--- a/tests/unit/packaging/test_search.py
+++ b/tests/unit/packaging/test_search.py
@@ -27,6 +27,7 @@ def test_build_search():
                 pretend.stub(version="3.0"),
                 pretend.stub(version="4.0"),
             ],
+            downloads=9001,
         ),
         summary="This is my summary",
         description="This is my description",
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -153,6 +153,10 @@ class Project(SitemapMixin, db.ModelBase):
 
         return request.route_url("legacy.docs", project=self.name)
 
+    @property
+    def downloads(self):
+        return sum([release.downloads for release in self.releases])
+
 
 class DependencyKind(enum.IntEnum):
 
@@ -338,6 +342,10 @@ class Release(db.ModelBase):
     @property
     def has_meta(self):
         return any([self.keywords])
+
+    @property
+    def downloads(self):
+        return sum([file_.downloads for file_ in self.files])
 
 
 class File(db.Model):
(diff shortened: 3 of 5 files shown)


dstufft:
So I like this, but there's one problem with it-- I think it's going to trigger half a million queries when reindexing against the production database.

di:
@dstufft I added a commit here that should significantly cut down the number of queries required for reindexing, however I noticed that in the month since I opened this PR, download statistics were removed from PyPI (https://bitbucket.org/pypa/pypi/commits/f54).

I'm not sure how much PyPI and Warehouse have in common at this point, but I did notice that in the prod DB snapshot I have, some projects report 0 downloads for their latest releases, e.g.:

```
>>> [r.downloads for r in db.query(Project).filter(Project.name=='Django').first().releases][:10]
[0, 0, 0, 20177, 270956, 264277, 3568, 13574, 10831, 6499]
```

So I'm wondering if this is still worth adding!

dstufft:
Download counts are temporarily disabled on PyPI because they broke and I haven't fixed them yet.

The current state of our statistics pipeline is that we're archiving all of the download events to an S3 like storage that DreamHost provides, and we're writing download events into a BigQuery database that Google provides. It doesn't exist yet, but there's probably going to be some sort of cronjob that will take data from BigQuery and inject them into the existing downloads field.

What does that mean for this PR? It means if total number of downloads is something we think we want to sort by, then this PR is still useful as that's still going to work. However BigQuery offers us some additional possible queries we can do in the future (like Daily/Monthly/Weekly counts) that maybe we think will give us better numbers (or maybe we don't).

An additional consideration is what effect this will have on #701-- Maybe the answer is we need to periodically do a full reindex. Or maybe we need a Celery cron task that will take N items from Elasticsearch that were indexed the longest time ago, and reindex them.

di:
> However BigQuery offers us some additional possible queries we can do in the future (like Daily/Monthly/Weekly counts) that maybe we think will give us better numbers (or maybe we don't).

I think that the "average" daily counts (for the last release) in this PR are about as useful if not more useful than the daily/monthly/weekly stats we could get from BigQuery. However, it would remove the need to reindex to get the latest stats (for downloads).

dstufft:
We can get average daily counts for any release from BigQuery.

di:
@dstufft Gotcha. It doesn't seem like we're using BigQuery in Warehouse at the moment, any thoughts/direction on adding it? Do you know if it's possible to emulate locally?

dstufft:
@di We're not using it in Warehouse at the moment, I have a branch locally that half way starts to add it.

I don't think it's possible to emulate locally, but for the record, the BigQuery data is completely public. See https://mail.python.org/pipermail/distutils-sig/2016-May/028986.html.

jaddison:
> The current state of our statistics pipeline is that we're archiving all of the download events to an S3 like storage that DreamHost provides

I'm late to the party here, but if you have the download data in S3 (or S3-like storage), I assume it is stored in datetime-based filenames. Can you not just run a cron job every 15 mins or hour to grab the most recent, unprocessed download logs - and then run elasticsearch bulk document in-place updates via https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-update (scripted field updating)?

I assume that BigQuery is used for other purposes (statistics related?), but if not it could be removed from the infrastructure to simplify things.

Lots of assumptions in my statements, I'm willing to admit. Apologies if this is just noise.

di:
This is sufficiently old and dusty, so I'm going to close this as it's not a priority.

PR #1185, opened 2016-05-16
Update getting-started.rst

Samureus:
Place "docker start" last?

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -15,9 +15,9 @@ Quickstart for Developers with Docker experience
 
     $ git clone git@github.com:pypa/warehouse.git
     $ cd warehouse
-    $ docker start
     $ make serve
     $ make initdb
+    $ docker start
 
 View Warehouse in the browser at ``http://localhost:80/`` (Linux) or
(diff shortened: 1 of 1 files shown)


dstufft:
Hey!

Thanks for the pull request, unfortunately I don't believe we can accept it. ``docker start`` is needed to start the docker daemon, which needs to be running prior to running ``make serve`` or ``make initdb``.

Samureus:
Sorry for not replying earlier, my bad, I guess I meant to say that the command was not needed, as far as I can tell the command to start the docker daemon is either

```
$ sudo systemctl start docker
# or on older distributions, you may need to use
$ sudo service docker start
```

(assuming you did not enable it to automatically start on boot), on the other hand, just running docker start shows an error.

```
~/Dev/warehouse$ docker start
docker: "start" requires a minimum of 1 argument.
See 'docker start --help'.

Usage:	docker start [OPTIONS] CONTAINER [CONTAINER...]

Start one or more stopped containers
```

From what I understood of the Makefile, I believe you have to first `make build`, or better yet, `make serve`,  (`docker-compose up` will already have the containers running) then `make initdb`.

dstufft:
Hmm. I think actually the ``docker start`` is just bad documentation and should be removed.

PR #1187, opened 2016-05-16
Select latest releases for display on profile pages

joealcorn:
I noticed when viewing my own profile that the date shown as the latest release was in fact the date of the first release.

The release that's pulled out of the database is unlikely to actually be the latest release because we were only sorting based on the project name, so presumably it'll pull the first result it finds in the database, most likely the first.


dstufft:
Good catch!

PR #1196, opened 2016-05-19
Add :z to the volume mounts to avoid permission issues with SELinux

d0ugal:


docker-compose.yml
```diff
@@ -26,10 +26,10 @@ app:
     # to make this work is to share multiple, smaller directories. These cover
     # the important things that we want to share, but changes to requirements
     # or any file not in these directories will require a rebuild.
-    - ./dev:/app/dev
-    - ./docs:/app/docs
-    - ./warehouse:/app/warehouse
-    - ./tests:/app/tests
+    - ./dev:/app/dev:z
```
dstufft: Can you add a comment about about why the ``:z`` exists so that a month from now I don't forget?

d0ugal: Done, I wasn't quite sure how best to word it, but hopefully it provides enough context.

PR #1203, opened 2016-05-23
Test and fix missing titles. Fixes #784

### LINKED ISSUE #784: Audit pages for missing titles
We'll want to make sure all of our pages actually have titles.

atodorov:
Here is a quick test for #784 and fixes to a number of templates. Notes:

* `pages/legal.html`, `pages/security.html` and `pages/sponsors.html` are missing titles at the moment but are essentially empty. 
* assert will stop at the first failure.

On legacy PyPI the legal link points to http://www.python.org/about/legal. If we have the text I can add it to this template or modify the code to redirect to the same URL. 

Do we want to copy the text from https://pypi.python.org/security for `security.html` ? 

`sponsors.html` contains the following comment
```
{# Note: The actual sponsors page for PyPI is defined in pypi-theme #}
```

Please advise how to proceed with this PR. 

Regards,
@MrSenko

### DIFF
--- a/tests/functional/test_templates.py
+++ b/tests/functional/test_templates.py
@@ -1,0 +1,53 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import warehouse
+from jinja2 import Environment, FileSystemLoader
+
+
+def test_templates_for_empty_titles():
+    """
+        Test if all HTML templates have defined the title block. See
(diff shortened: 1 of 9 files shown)


dstufft:
Hey @atodorov 

What I would do for those three pages is just add a dummy title to them, something like ``{% block title %}Legal{% endblock %}`` for each one. That should get the tests passing, and then people can just fill in with the real title when the content for those pages get filled out (and this test will make sure they don't forget to do that if they delete the title block).

tests/functional/test_templates.py
```diff
@@ -0,0 +1,54 @@
...
+from jinja2 import Environment, FileSystemLoader
+
+
+def test_templates_for_empty_titles():
+    """
+        Test if all HTML templates have defined the title block. See
+        https://github.com/pypa/warehouse/issues/784
+    """
+    dir_name = os.path.dirname(__file__)
+    dir_name = os.path.join(dir_name, '..', '..', 'warehouse', 'templates')
```
dstufft: This should probably ``import warehouse`` and then do ``os.path.join(os.path.dirname(warehouse.__file__), "templates")`` to make this more resilient against the test code getting shuffled around.

dstufft:
One small adjustment, and then looks good to me to merge. Thanks!

dstufft:
✨ Awesome 🍰

PR #1215, opened 2016-06-03
Add prefix query on name to search

robhudson:
This also refactors what was the multi_match query to explicit queries
to allow for prefix queries.

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -181,7 +181,7 @@ def search(request):
         if len(q) > 1:
             should.append(Q('prefix', normalized_name=q))
 
-        query = request.es.query("bool", should=should, minimum_should_match=1)
+        query = request.es.query("dis_max", queries=should)
         query = query.suggest("name_suggestion", q, term={"field": "name"})
     else:
         query = request.es.query()
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -18,6 +18,7 @@ from pyramid.httpexceptions import (
 from pyramid.view import (
     notfound_view_config, forbidden_view_config, view_config,
 )
+from elasticsearch_dsl import Q
 from sqlalchemy import func
 from sqlalchemy.orm import aliased, joinedload
 
@@ -29,6 +30,19 @@ from warehouse.classifiers.models import Classifier
 from warehouse.packaging.models import Project, Release, File
 from warehouse.utils.row_counter import RowCount
 from warehouse.utils.paginate import ElasticsearchPage, paginate_url_factory
+
+
+SEARCH_FIELDS = [
+    "author", "author_email", "description", "download_url", "home_page",
+    "keywords", "license", "maintainer", "maintainer_email", "normalized_name",
+    "platform", "summary",
+]
+SEARCH_BOOSTS = {
+    "normalized_name": 10,
+    "description": 5,
+    "keywords": 5,
+    "summary": 5,
+}
 
 
 @view_config(context=HTTPException)
@@ -152,21 +166,23 @@ def index(request):
     ],
 )
 def search(request):
-    if request.params.get("q"):
-        query = request.es.query(
-            "multi_match",
-            query=request.params["q"],
-            fields=[
-                "author", "author_email", "description^5", "download_url",
-                "home_page", "keywords^5", "license", "maintainer",
-                "maintainer_email", "normalized_name^10", "platform",
-                "summary^5",
-            ],
-        ).suggest(
-            name="name_suggestion",
-            text=request.params["q"],
-            term={"field": "name"}
-        )
+
+    q = request.params.get("q", '')
+
+    if q:
+        should = []
+        for field in SEARCH_FIELDS:
+            kw = {"query": q}
+            if field in SEARCH_BOOSTS:
+                kw["boost"] = SEARCH_BOOSTS[field]
+            should.append(Q("match", **{field: kw}))
+
+        # Add a prefix query if ``q`` is longer than one character.
+        if len(q) > 1:
+            should.append(Q('prefix', normalized_name=q))
+
+        query = request.es.query("bool", should=should, minimum_should_match=1)
+        query = query.suggest("name_suggestion", q, term={"field": "name"})
     else:
         query = request.es.query()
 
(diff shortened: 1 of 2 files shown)


robhudson:
Note: Because of the pretend.stub test failures I've not fixed the real failures. I'm not familiar with pretend.

I'll continue working on this tomorrow but if there's time I would appreciate a preliminary review on the direction and help with pretend. Thanks.

dstufft:
Re: pretend, it looks like your failures fall into two categories

The first is where you have a ``assert db_request.es.query.calls == [ ... ]`` that is failing. For this you just need to update the RHS to match the new calls you're making. If you're making a call that looks something like ``request.es.query(foo="bar")``, then you just need to update that so the RHS looks like ``[pretend.call(foo="bar")]``.

The second is a lack of a suggest stub, this should be pretty easy to add, you basically slightly changed it so that instead of ``suggest()`` only being called when there was a query, you're calling it all of the time. Either that should be adjusted so it's only called when there is actually a query (in which case, these failures should go away) or the tests that test when there is no ``?q=...`` will need to be adjusted to match the other tests and their use of a ``suggest`` stub.

warehouse/views.py
```diff
@@ -152,29 +153,47 @@ def index(request):
...
+    if q:
+        for field in fields:
+            kw = {"query": q}
+            if field in boosts:
+                kw["boost"] = boosts[field]
+
+            should.append(Q('match', **{field: kw}))
+
+    # Add a prefix query if ``q`` is longer than one character.
+    if len(q) > 1:
```
dstufft: Although the difference should be minuscule, it probably makes sense to move this to under ``if q:``, no reason to evaluate ``if len(q) > 1`` if we already know that ``q`` is ``False``.

warehouse/views.py
```diff
@@ -152,29 +153,47 @@ def index(request):
...
+        should.append(Q('prefix', normalized_name=q))
```
dstufft: Probably want to do something like ``normalized_name=packaging.utils.cannoncalize_name(q)`` here so that we're comparing like to like (normalized to normalized). For that matter, we probably want to do the same on the ``normalized_name`` field up above (Although that's not a regression in this PR, so you don't need to add that unless you want to).

warehouse/views.py
```diff
@@ -152,29 +153,47 @@ def index(request):
...
+
+    if should:
+        query = request.es.query("bool", should=should, minimum_should_match=1)
     else:
         query = request.es.query()
 
+    # Add suggestions.
+    query = query.suggest(name="name_suggestion", text=q,
```
dstufft: As mentioned earlier, this has slightly changed the behavior of this so that instead of only getting suggestions when a query was provided, it now adds suggestions *always*. Is there a reason to change this behavior, or should this be gated on an ``if q:``?

warehouse/views.py
```diff
@@ -189,12 +208,12 @@ def search(request):
     available_filters = collections.defaultdict(list)
 
     for cls in request.db.query(Classifier).order_by(Classifier.classifier):
-        first, *_ = cls.classifier.split(' :: ')
+        first = cls.classifier.split(' :: ')[0]
```
dstufft: Any reason to make this change? Not that I disagree with it.

di: Also curious... I would slightly prefer the original syntax.

warehouse/views.py
```diff
@@ -152,29 +153,47 @@ def index(request):
...
+            if field in boosts:
+                kw["boost"] = boosts[field]
+
+            should.append(Q('match', **{field: kw}))
+
+    # Add a prefix query if ``q`` is longer than one character.
+    if len(q) > 1:
+        should.append(Q('prefix', normalized_name=q))
+
+    if should:
```
dstufft: This whole section could probably be simplified as:

```python
if q:
    should = []
    for field in fields:
        ...

    if len(q) > 1:
        should.append(Q('prefix', normalized_name=q))

    query = request.es.query("bool", should=should, minimum_should_match=1)
    query = query.suggest(name="name_suggestion", text=q)
else:
    query = request.es.query()
```

I think that's a bit simpler in terms of branches and such too and I *think* it matches what you're doing here (unless I'm misunderstanding something).

dstufft:
Cool, so I did a review, the overall approach looks pretty reasonable to me (at least as far as I understand ES). I made a few inline comments, but overall this looks like a good change.

warehouse/views.py
robhudson: Mostly because the other variables are never used. I'm happy to change it back tho.

warehouse/views.py
robhudson: Definitely. I didn't catch that the `.suggest` was only if a `q` is provided. Thanks.

warehouse/views.py
robhudson: Checking into this a bit more. I don't think `q` as the users type it is going to look like a `normalized_name`. So I should probably use `name` here, which, I believe, is also in the index.

dstufft: Yea-- but we can normalize any name that users type already using ``packaging.utils.cannoncalize_name()``. Is there any reason not to do that? I know that ES does some of it's own normalizing already so maybe it's redundant I don't know.

robhudson: Also, are there other fields besides project names that people may want a partial match on?

dstufft: I don't think so? I'm not sure to be honest :]

robhudson: I don't see other fields that stand out for this so I'll leave it to name.

The `normalized_name` has a custom analyzer that tokenizes (lowercase, delimits words) so RHS processing on the input isn't necessary, IMO. It probably isn't necessary at index time either but since it's a postgresql thing it doesn't hurt anything. It is the right choice of fields since the `name` field in ES has no custom analyzers.

dstufft: Okay

robhudson:
Updated after code review.

robhudson:
Found a bug here. The changes to the elasticsearch query from `multi_match` to `bool` queries results in exact matches no longer being at the top. I'm moving it to a `dis_max` query which is what `multi_match` does under the hood and testing that.

robhudson:
Updated again. Ready for full review pass.

dstufft:
✨

PR #1220, opened 2016-06-03
Disable autocorrect in the username field, issue #1198

### LINKED ISSUE #1198: Disable autocorrect/autocomplete/spellcheck on username field
When going to https://pypi.io/ and entering my username "xistence" my computer will happily recommend that I change my username to "distance" and while that may be a neat username, it certainly isn't my username...

Adding these to the `<input>` should do the trick: `autocorrect="off" autocapitalize="off" spellcheck="false"`.

karan:
Prevents a device from autocorrecting the username.

Fixes #1198

### DIFF
--- a/warehouse/templates/includes/input-credentials.html
+++ b/warehouse/templates/includes/input-credentials.html
@@ -5,7 +5,7 @@
   {% endfor %}
 </ul>
 {% endif %}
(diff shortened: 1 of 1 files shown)


dstufft:
Awesome, thanks!

Issue #819, opened 2015-11-22
Integrate tooltip library


nlhkabu:
Requirements:
- We should be able to invoke the tooltip on hover or click
- The tooltip should work with all directions (top, left, bottom, right)

It would be nice if the plugin could automatically calculate the best direction, though this is not essential.

Organisation:
Please put all relevant CSS into a new file at:
`https://github.com/pypa/warehouse/tree/master/warehouse/static/sass/components/_tooltip.scss`

karan:
OK to close this @nlhkabu?

Issue #818, opened 2015-11-22
Copy pip command and sha256 hash on click (relies on #819)


nlhkabu:
Clicking on the copy icon should copy the content to the clipboard:

On the package page:
![pip](https://cloud.githubusercontent.com/assets/3323703/11322702/5b3bd4ea-90f0-11e5-88f6-967a7878c8bc.png)

On the downloads tab on the package page:
![md5](https://cloud.githubusercontent.com/assets/3323703/11322701/5541d8e6-90f0-11e5-85fd-b845eb8699b5.png)

I would like this to work with a tooltip as on Github:
![copy-to-clipboard](https://cloud.githubusercontent.com/assets/3323703/11322714/ce826af4-90f0-11e5-83bf-99589936aa5a.png)

![copied](https://cloud.githubusercontent.com/assets/3323703/11322713/ce7f1eda-90f0-11e5-85f3-ec0737d3f2b8.png)

Relies on #819

domenkozar:
Could we instead of insecure `md5` use `sha256`? New PyPI could be a chance to fix this.

nlhkabu:
@domenkozar something to talk to @dstufft about :)  - Can you please open another issue for this, as I don't want that discussion to take over this thread.

nlhkabu:
@domenkozar in fact - this is already being discussed at #681

domenkozar:
I've opened https://github.com/pypa/warehouse/issues/820, sorry for the noise here.

alex:
https://clipboardjs.com/ is pleasant to use from my experience

dstufft:
This is implemented now in #1245!

PR #1223, opened 2016-06-03
Copy pip command to clipboard

karan:
#818 so far...

Does not implement #819.

### DIFF
--- a/Gulpfile.babel.js
+++ b/Gulpfile.babel.js
@@ -56,6 +56,9 @@ let webpackConfig = {
   },
   resolve: {
     modules: [ path.resolve(staticPrefix, "js"), "node_modules" ],
+    alias: {
+      "clipboard": "clipboard/dist/clipboard",
+    },
   },
 };
 
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "babel-preset-es2015": "6.6.0",
     "babel-preset-es2015-native-modules": "6.6.0",
     "babel-register": "6.7.2",
+    "clipboard": "1.5.10",
     "del": "2.2.0",
     "exports-loader": "0.6.3",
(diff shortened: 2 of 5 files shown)


dstufft:
Ok, after playing with this, for some reason this seems to work if you import it like:

```javascript

import Clipboard from "clipboard/dist/clipboard";
```

I'm not entirely sure *why*, but it seems to work.

dstufft:
Ok, figured out why, and a better way of handling this... Basically ``clipboard.js``'s ``package.json`` is a bit hinky and it is pointing to an ES6 module even though it should be a CommonJS module. That doesn't fully work *BUT* they also include the babel compiled CommonJS module inside of ``dist/``. So that uses that instead of their ES6 module.

This can be better fixed with this patch to your branch:

```diff
diff --git a/Gulpfile.babel.js b/Gulpfile.babel.js
index b7e283e..8acc441 100644
--- a/Gulpfile.babel.js
+++ b/Gulpfile.babel.js
@@ -56,6 +56,9 @@ let webpackConfig = {
   },
   resolve: {
     modules: [ path.resolve(staticPrefix, "js"), "node_modules" ],
+    alias: {
+      "clipboard": "clipboard/dist/clipboard",
+    },
   },
 };
 
```

This will tell webpack to replace ``"clipboard"`` with ``"clipboard/dist/clipboard"`` whenever used in an import statement.

karan:
Oh wow - that's beyond my knowledge of babel. But thanks for digging that up.

I've updated the PR with the change. (Added an empty file for tooltip)

dstufft:
There is a lint failure here, otherwise this looks pretty good to me.

karan:
ah lints. fixed it!

dstufft:
✨

PR #1226, opened 2016-06-04
[WIP] Add some optimizations to reindex

robhudson:
WORK IN PROGRESS

I consider the code to be relatively complete (besides the tests), but local testing is revealing a lot of timeout errors connecting to the Elasticsearch cluster. I don't know if a change here causes this or something else.

### DIFF
--- a/warehouse/cli/search/reindex.py
+++ b/warehouse/cli/search/reindex.py
@@ -57,6 +57,7 @@ def reindex(config, **kwargs):
     """
     client = config.registry["elasticsearch.client"]
     db = Session(bind=config.registry["sqlalchemy.engine"])
+    num_replicas = config.registry.get("elasticsearch.replicas", 0)
 
     # We use a randomly named index so that we can do a zero downtime reindex.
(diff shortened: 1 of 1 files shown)


dstufft:
Just gonna say that the general idea and the changes look good to me, once you work out the issues :]

warehouse/cli/search/reindex.py
```diff
@@ -68,14 +69,21 @@ def reindex(config, **kwargs):
...
         using=client,
         shards=config.registry.get("elasticsearch.shards", 1),
-        replicas=config.registry.get("elasticsearch.replicas", 0),
+        replicas=0,
     )
     new_index.create()
+    client.indices.put_settings(
+        index=new_index_name,
+        body={"index": {"refresh_interval": "-1"}}
+    )
```
di: @robhudson Is there any reason why this setting would nee to be set _after_ the index is created?

robhudson: If there's a way to set the refresh interval when the index is created, I think that would be better. I don't recall if that was possible when I started this or not.

PR #1230, opened 2016-06-04
Add a tooltip library (and add it to pip install button)

karan:
#818 and #819

This is based on github's implementation and is distributed with MIT license.

phone

The one little snafu is that after copying, if the user hover off of the button and then back on the button, it says "Copied" - it should say "Copy to clipboard". Haven't found a good solution to that yet.

Next up will be integrating copy to clipboard to sha sums.

### DIFF
--- a/warehouse/static/sass/warehouse.scss
+++ b/warehouse/static/sass/warehouse.scss
@@ -96,6 +96,7 @@
 @import "components/statistics-bar";
 @import "components/vertical-tabs";
 @import "components/wh-buttons";
+@import "components/tooltip";
 
 
 // TRUMPS LAYER
--- a/warehouse/static/sass/warehouse.scss
+++ b/warehouse/static/sass/warehouse.scss
@@ -94,9 +94,9 @@
 @import "components/site-header"; // refactor
 @import "components/sponsor-footer";
 @import "components/statistics-bar";
+@import "components/tooltip";
 @import "components/vertical-tabs";
 @import "components/wh-buttons";
-@import "components/tooltip";
 
 
 // TRUMPS LAYER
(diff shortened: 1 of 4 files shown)


dstufft:
Linting again :)

Since this has some design work in it, it'll need @nlhkabu's OK, but it looks good to me!

warehouse/static/sass/warehouse.scss
```diff
@@ -96,6 +96,7 @@
 @import "components/statistics-bar";
 @import "components/vertical-tabs";
 @import "components/wh-buttons";
+@import "components/tooltip";
```
nlhkabu: Can you please add this to the list in alphabetical order. Thanks.

nlhkabu:
Looks great!  Thanks @karan! Can you please also add it to the SHA link on the downloads tab?

screenshot from 2016-06-04 16-46-43

https://github.com/pypa/warehouse/blob/master/warehouse/templates/packaging/detail.html#L297

warehouse/templates/packaging/detail.html
```diff
@@ -71,7 +71,7 @@ <h3 class="title"><a href="{{ request.route_path('packaging.project', name=relea
 
     <p class="pip-instructions">
       <span id="pip-command">pip install {{ release.project.normalized_name }}</span>
-      <a href="#" class="copy-pip-command" data-clipboard-target="#pip-command">
+      <a href="#" class="copy-pip-command tooltipped tooltipped-s" data-clipboard-target="#pip-command" aria-label="Copy to clipboard">
```
nlhkabu: also, can we please namespace the js hook with `-js`. e.g `-js-copy-pip-command`.

karan:
Thanks for reviewing @nlhkabu @dstufft. I fixed the lint and namespace issues.

I was going to add the sha copy button in a separate PR to make the history cleaner.

dstufft:
Awesome!

Let me just say too, how awesome it was to have a little gif showing off the effects of this. It made seeing what was going to change way easier for me at least.

nlhkabu:
:tada: :dancers:

PR #1237, opened 2016-06-05
Handle autohiding like legacy PyPI does

### LINKED ISSUE #1222: Uploads to pypi.io don't trigger auto-hide old releases functionality in pypi.python.org
It appears that [uploads to pypi.io are only syncing json metadata](https://github.com/pypa/setuptools/issues/589#issuecomment-223615124) and not updating the latest package [at the canonical location](https://pypi.python.org/pypi/setuptools).

I've tried [managing the project](https://pypi.python.org/pypi?%3Aaction=pkg_edit&name=setuptools) to see what releases might be present, but that URL returns 500 after 30 seconds (probably a timeout), so I'm unable to assess exactly why pypi.python.org isn't showing the latest release.

dstufft:
Fixes #1222

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -1710,6 +1710,160 @@ class TestFileUpload:
             "403 Invalid or non-existent authentication information."
         )
 
+    def test_autohides_old_releases(self, pyramid_config, db_request):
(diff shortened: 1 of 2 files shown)


jaraco:
Nice. Thanks!

Issue #1234, opened 2016-06-04
After copying "pip install package" the tooltip says "Copied!" for forever


alex:
After a few seconds it should switch back to "Copy to clipboard"

dstufft:
Even better would be if we can detect when the ``tooltipped:hover``, ``.tooltipped:active``, and ``.tooltipped:focus`` events stop happening and change it then, so that it only changes once someone has moused off the thing.

I think this would be on the ``blur`` or ``focousout`` event (for ``:active``) and ``mouseout`` or ``mouseleave`` (for ``:hover``), but someone would need to figure out what the right events are. This would mean that click and keeping your mouse there would keep it at ``Copied!``... until you mouse away, then it would turn back to ``Copy to Clipboard``.

An alternative would be to move away from the pure CSS tooltips and instead switch to JS based tooltips, at least for this, which would give us tooltip show/hide events we can subscribe to.

karan:
Yeah that's a snafu I mentioned in the PR - and haven't been able to find a decent solution to (I'm not a JS person - just know enough to be dangerous).

I like the idea of listening to focusout and mouseout events. I'll keep looking and trying other things though (got my last exam on Monday).

karan:
@alex @dstufft This can be closed now.

PR #1240, opened 2016-06-06
Set original tooltip text when hover out

karan:
Fix for #1234 

Tested on desktop Chrome stable and Firefox. `setAttribute` does not work in Safari.

phone

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -39,4 +39,22 @@ docReady(() => {
     e.trigger.setAttribute("aria-label", "Copied!");
     e.clearSelection();
   });
+
+  // Get all elements with class "tooltipped" and bind to focousout and
+  // mouseout events. Change the "aria-label" to "original-label" attribute
+  // value.
+  let setOriginalLabel = (element) => {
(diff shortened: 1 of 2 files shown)


warehouse/static/js/warehouse/index.js
```diff
@@ -39,4 +39,22 @@ docReady(() => {
     e.trigger.setAttribute("aria-label", "Copied!");
     e.clearSelection();
   });
+
+  // Get all elements with class "tooltipped" and bind to focousout and
+  // mouseout events. Change the "aria-label" to "original-label" attribute
+  // value.
+  let setOriginalLabel = (element) => {
+    element.setAttribute("aria-label", element.getAttribute("original-label"));
```
dstufft: Instead of using an attribute called ``original-label``, I think it would be better to use one called ``data-original-label``, and then instead of ``element.getAttribute`` you should be able to use (I think) ``element.dataset.originalLabel``.

dstufft:
One small adjustment to better match what I think is a better practice in the modern, HTML5 web, but overall it looks good to me. Great work!

karan:
Done!

dstufft:
✨

PR #1242, opened 2016-06-07
return 400 if search page num is non-int

### LINKED ISSUE #1236: non-integer page params should result in a 4xx not a 500
https://github.com/pypa/warehouse/blob/master/warehouse/views.py#L195

Right now if you pass anything that's not an int, you get a 500.

karan:
Fixes #1236

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -192,7 +192,11 @@ def search(request):
     if request.params.getall("c"):
         query = query.filter("terms", classifiers=request.params.getall("c"))
 
-    page_num = int(request.params.get("page", 1))
+    try:
+        page_num = int(request.params.get("page", 1))
+    except ValueError:
+        raise HTTPNotFound
+
     page = ElasticsearchPage(
         query,
         page=page_num,
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -196,7 +196,7 @@ def search(request):
     try:
         page_num = int(request.params.get("page", 1))
     except ValueError:
-        raise HTTPBadRequest
+        raise HTTPBadRequest("'page' must be an integer.")
 
     page = ElasticsearchPage(
         query,
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -14,6 +14,7 @@ import collections
 
 from pyramid.httpexceptions import (
     HTTPException, HTTPSeeOther, HTTPMovedPermanently, HTTPNotFound,
+    HTTPBadRequest,
 )
 from pyramid.view import (
     notfound_view_config, forbidden_view_config, view_config,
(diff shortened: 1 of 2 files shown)


alex:
I think this should be a 400, not a 404, do you agree @dstufft? Also, probably needs a test.

karan:
@alex @dstufft how can i run the test suite locally?

dstufft:
Yea, a 400 is what I think it should be.

warehouse/views.py
```diff
@@ -192,7 +193,11 @@ def search(request):
     if request.params.getall("c"):
         query = query.filter("terms", classifiers=request.params.getall("c"))
 
-    page_num = int(request.params.get("page", 1))
+    try:
+        page_num = int(request.params.get("page", 1))
+    except ValueError:
+        raise HTTPBadRequest
```
dstufft: It'd probably be nice to add a message to this, like ``raise HTTPBadRequest("'page' must be an integer, was {!r}".format(request.params["page"])`` or something like that. I'm pretty sure Pyramid will expose that information in the response so people can see what they did wrong.

alex: You don't want to just use a format like that though, possibly opens you up
to XSS.

On Mon, Jun 6, 2016 at 9:03 PM, Donald Stufft <notifications@github.com>
wrote:

> In warehouse/views.py
> <https://github.com/pypa/warehouse/pull/1242#discussion_r65995809>:
>
> > @@ -192,7 +193,11 @@ def search(request):
> >      if request.params.getall("c"):
> >          query = query.filter("terms", classifiers=request.params.getall("c"))
> >
> > -    page_num = int(request.params.get("page", 1))
> > +    try:
> > +        page_num = int(request.params.get("page", 1))
> > +    except ValueError:
> > +        raise HTTPBadRequest
>
> It'd probably be nice to add a message to this, like raise
> HTTPBadRequest("'page' must be an integer, was
> {!r}".format(request.params["page"]) or something like that. I'm pretty
> sure Pyramid will expose that information in the response so people can see
> what they did wrong.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/pypa/warehouse/pull/1242/files/163c42f9716be0173deb7b22968495f5ce79ad57#r65995809>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAADBIOwuSVf6TX1nC4S40FzUFIYFBY9ks5qJMN4gaJpZM4Ivb7P>
> .
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: D1B3 ADC0 E023 8CA6

karan:
@dstufft Thanks for the suggestions. I added the message.

Sorry this took many revisions - but hey at least I know how to run the tests locally. :)

dstufft:
No problems! It looks great :shipit:

PR #1246, opened 2016-06-08
Add classifiers and links in project detail page.

### LINKED ISSUE #71: Link classifiers in the project description
The classifiers need to be linked to some sort of "browse" page.

wlonk:
Maybe this fixes #71?

Is this helpful? Is this in the right direction? I hope so! Let me know.

_According to the traditions of my people, every pull request should come with a cute animal picture:_

armadillo

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -12,7 +12,10 @@
 
 import enum
 
-from collections import OrderedDict
+from collections import (
+    OrderedDict,
+    defaultdict,
+)
 
 from citext import CIText
 from pyramid.security import Allow
@@ -264,6 +267,15 @@ class Release(db.ModelBase):
         order_by=Classifier.classifier,
     )
     classifiers = association_proxy("_classifiers", "classifier")
+
+    @property
+    def structured_classifiers(self):
+        structured = defaultdict(list)
+        for classifier in self.classifiers:
+            key, *value = classifier.split(' :: ', 1)
+            if value:
+                structured[key].append(value[0])
+        return structured
 
     files = orm.relationship(
         "File",
--- a/warehouse/static/sass/components/_sidebar-section.scss
+++ b/warehouse/static/sass/components/_sidebar-section.scss
@@ -6,15 +6,6 @@
   h3 {
     font-size: 18px;
   }
-
-  dt {
-    font-weight: bold;
-    margin-top: 0.3em;
-  }
-
-  dd {
-    padding-left: 1em;
-  }
 }
 
 .sidebar-section.status {
--- a/warehouse/static/sass/components/_sidebar-section.scss
+++ b/warehouse/static/sass/components/_sidebar-section.scss
@@ -6,6 +6,14 @@
   h3 {
     font-size: 18px;
   }
+
+  dt {
+    font-weight: bold;
+  }
+
+  dd {
+    padding-left: 1em;
+  }
 }
 
 .sidebar-section.status {
--- a/warehouse/static/sass/components/_sidebar-section.scss
+++ b/warehouse/static/sass/components/_sidebar-section.scss
@@ -9,6 +9,7 @@
 
   dt {
     font-weight: bold;
+    margin-top: 0.3em;
   }
 
   dd {
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -25,6 +25,9 @@ from ...common.db.packaging import (
     ProjectFactory as DBProjectFactory, ReleaseFactory as DBReleaseFactory,
     FileFactory as DBFileFactory, RoleFactory as DBRoleFactory,
 )
+from ...common.db.classifiers import (
+    ClassifierFactory as DBClassifierFactory,
+)
 
 
 class TestProjectFactory:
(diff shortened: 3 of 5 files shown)


nlhkabu:
Hi @wlonk - thanks for the PR and the pic :heart: 
Are you able to add a screenshot of what the new sidebar looks like?

wlonk:
Sure thing. It's, um, _functional_ to put it kindly.

screenshot 2016-06-08 01 09 30

See the added thing on the left-hand sidebar area.

nlhkabu:
Given the format, I'm not sure that we can make these look anything other than "functional" :) 

In the short term, maybe we can move them to be the last sidebar section (under the tags)?

dstufft:
This might be a bad idea, I'm not good at design, but one thing that struck me is that maybe we could have less wrapped text if we changed

* Development Status :: 4 - Beta
* Environment :: Console
* Intended Audience :: Developers
* Intended Audience :: End Users/Desktop
* Natural Language :: English
* Operating System :: OS Independent
* Topic :: System :: Archiving

into 

* Development Status
  * 4 - Beta
* Environment
  * Console
* Intended Audience
  * Developers
  * End Users/Desktop
* Natural Language
  * English
* Operating System
  * OS Independent
* Topic
  * System
    * Archiving

I don't know if that's better or worse though.

dstufft:
Another option could be something like:

<dl>
  <dt>Development Status</dt>
  <dd>
    <ul>
      <li>4 - Beta</li>
    </ul>
  </dd>

  <dt>Environment</dt>
  <dd>
    <ul>
      <li>Console</li>
    </ul>
  </dd>

  <dt>Intended Audience</dt>
  <dd>
    <ul>
      <li>Developers</li>
      <li>End Users/Desktop</li>
    </ul>
  </dd>

  <dt>Natural Language</dt>
  <dd>
    <ul>
      <li>English</li>
    </ul>
  </dd>

  <dt>Operating System</dt>
  <dd>
    <ul>
      <li>OS Independent</li>
    </ul>
  </dd>

  <dt>Topic</dt>
  <dd>
    <ul>
      <li>System :: Archiving</li>
    </ul>
  </dd>

  <dt>Topic (Alternate 1)</dt>
  <dd>
    <ul>
      <li>
        System
        <ul>
            <li>Archiving</li>
        </ul>
      </li>
    </ul>
  </dd>

  <dt>Topic (Alternate 2)</dt>
  <dd>
    <dl>
      <dt>System</dt>
      <dd>
        <ul>
            <li>Archiving</li>
        </ul>
      </dd>
    </dl>
  </dd>
</dl>

wlonk:
@nlhkabu: I'll move it to the last sidebar position. That's better, thanks for the feedback.

@dstufft: I'm gonna be traveling today, but I'll try to grab some time in the airport or on the airplane to try this out and see how it looks. I have an old-school fondness for seeing `Foo :: Bar :: Baz` on classifiers, but given that it doesn't have enough space to not-line-wrap on the sidebar, I think that trying alternatives is especially worthwhile.

nlhkabu:
@dstufft good idea! In fact, it's the way we display the classifiers on the search result filter sidebar - so there is a precedent :)

dstufft:
@nlhkabu I'd be lying if I didn't say that's what made me think of the idea :)

dstufft:
I'm not sure which is better, the nested ul's is more consistent maybe, but there's never going to be a single, top level without a nested ul, so a DL might fit better, and do we nest classifiers with > 2 sections further or do we only do 1 level of nesting? I dunno!

wlonk:
For no good reason, my gut preference is DL with no deep nesting. I'll ponder why I feel that way.

wlonk:
There's some more styling to do, but as a skeleton, what do you think of this, @nlhkabu and @dstufft?

screenshot 2016-06-08 16 19 19

warehouse/packaging/models.py
```diff
@@ -265,6 +268,15 @@ def __table_args__(cls):  # noqa
...
     classifiers = association_proxy("_classifiers", "classifier")
 
+    @property
+    def structured_classifiers(self):
+        structured = defaultdict(list)
+        for classifier in self.classifiers:
+            key, *value = classifier.split(' :: ', 1)
+            if value:
+                structured[key].append(value[0])
+        return structured
```
di: This is pretty similar to what's happening here (though not exactly the same):
https://github.com/pypa/warehouse/blob/32b27e77c5caaf2a0aafe450f200a076b3135c24/warehouse/views.py#L210-L214

I wonder if there might be a way to share this functionality between both places.

wlonk: It's almost the same, except I need the tail, not the whole classifier, as the value. I don't see a _great_ way of unifying them right now.

nlhkabu:
Looking good @wlonk ! On the styling front, maybe we can an indent to the `<dt>`s and a little margin under each item:

classifiers

Otherwise, I am open to other ideas :)

wlonk:
I've added those changes, but do we have any baseline grid concerns I should be aware of? I know enough to know that that's a thing, but not how to make good choices around it.

nlhkabu:
> do we have any baseline grid concerns I should be aware of?

Not currently - this is something I need to clean up across the CSS, so it'sok to do something custom for now and I will make a fix later :)

nlhkabu:
My mistake - I found my vertical baseline :p

I am using a `$spacing-unit` across the CSS codebase - for the vertical basline, I am using half of this unit, e.g:

```
padding-bottom: $spacing-unit / 2;
```

warehouse/static/sass/components/_sidebar-section.scss
```diff
@@ -6,6 +6,15 @@
...
+
+  dt {
+    font-weight: bold;
+    margin-top: 0.3em;
+  }
+
+  dd {
+    padding-left: 1em;
+  }
 }
```
nlhkabu: It would be better to make this CSS generic by moving it into base/_lists.scss :)

wlonk:
Changes made!

nlhkabu:
Ok to merge for me.  Ping @dstufft for final review.

dstufft:
Thanks! I'm going to merge this, but I'm also going to submit a second PR that shuffles things around just a bit to move the classifier munging off of the data model, since it doesn't really need to be there.

dstufft:
Ok, I just submitted https://github.com/pypa/warehouse/pull/1264 which shows the tweaks I made. They're basically just moving the classifier stuff off the data model which means we can test it without the database and it's a bit cleaner of code since it's now a "pure" function.

PR #1249, opened 2016-06-09
fixed relative upload date

### LINKED ISSUE #987: Relative upload time wrong by 8 hours (UTC vs local time ?)
- Publish package
 - go to https://warehouse.python.org/project/yourpackage/
 - be in non UTC timezone ?

It says:
```
Uploaded about 8 hours ago by <me> <myavatar>
```

html:

```
<time class="-js-relative-time" datetime="2016-02-26T22:00:26" title="Feb 26, 2016">about 8 hours ago</time>
```

It's locally 14h00 (Pacific Time), so my guess it's a JS side only issue.
Not really important I guess, as you don't upload that often.

But in case something else is wrong with dates....

nkanaev:
resolves #987

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -110,7 +110,7 @@
         {% endfor %}
       </nav>
       <p class="author">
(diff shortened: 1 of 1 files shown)


dstufft:
Oops! Sorry I totally missed this PR. Thanks a lot for this ✨

Issue #1228, opened 2016-06-04
Implement "forgot password" feature


karan:
- A new page that will be linked to from the sign-in and register page.
- Asks for username
- Email the user with a temporary link to change the password

nlhkabu:
I've added this to the launch milestone because it would be really nice to get this working soon :)

gsb-eng:
@dstufft @nlhkabu @karan Anyone started working on this? If not, can I implement this?

dstufft:
As far as I know nobody has started working on it, so feel free :)

karan:
@gsb-eng  I have not. I'll be out of town without a computer so please go for it. :)

gsb-eng:
I'm working on this.

sfdye:
+1
Need this feature, when I try to register an account with my primary email, it says an account with the email already exists, but I don't remeber the password

todbot:
+1 me too. I cannot publish my packages any more since it seems like https://pypi.python.org won't accept new uploads.

brainwane:
In our [meeting today](https://wiki.python.org/psf/PackagingWG/2017-12-19-Warehouse) we decided this belongs in [the first milestone, the MVP for package maintainers](https://github.com/pypa/warehouse/milestone/8).

PR #1262, opened 2016-06-11
Reset password functionality.

gsb-eng:
#1228 Implementation.


gsb-eng:
@dstufft Do we have any "send_email" configured for warehouse project?

nlhkabu:
thanks for your work on this so far @gsb-eng .
FYI - when it comes to the templates, feel free to output the content without any styling - I will add style to it when I do the login and register pages.

gsb-eng:
@ncoghlan Sure, BTW I'm not good in styling :)

warehouse/accounts/forms.py
```diff
@@ -117,3 +117,40 @@ def validate_password(self, field):
...
+        # Instead of again defining username field, we are using LoginForm
+        # to get the username field and poping password field
+        self._fields.pop('username')
+
+    def validate_password(self, field):
+        userid = self.user_service.find_userid(self.username)
+        if userid is not None:
+            if self.user_service.check_password(userid, field.data):
+                raise wtforms.validators.ValidationError(
+                    "Password shouldn't match with previous one.")
```
hroncok: When I forget a password and then click a link to reset it and set the old password, why it is not ok to have the same password as before?

gsb-eng: @hroncok In my view It's all about how we can limit the damage of a password leak, If the system administrator is aware of the leak then users can be notified and other steps can be taken. Since passwords are open, system administrator would expect users to change their passwords, then forbidding password reuse is quite logical: if users reuse old passwords, technically they are not really changing their passwords.

Above scenario is logical in case of password leak and an unauthorized person tried to access the system. If we think in a real user perspective,  reusing the old password means: logically he/she still remember the password. The question arises here is, Why would somebody change their password that is still there in their memory?

hroncok: It happened to me more than once.

gsb-eng:
@dstufft : I've completed the core functionality, I just wanted to make sure that my implementation for password reset is valid or not. Could you please check and let me know if any changes required.

gsb-eng:
@dstufft  Could you please verify whether the implementation is good or not? so that I can start unit-tests.

dstufft:
Looking at this, I see one major thing that I really think we should attempt to avoid. In particular, it is attempting to use Redis to store tokens, presumably to ensure that a single token is only good for one use rather than being able to be used over and over again. This is a good thing to try and do, however by relying on storing valid tokens we make this a bit more fragile than it otherwise could be. I would prefer to instead use a mechanism that does not rely on any server side storage.

I took a look at what Django does here, and I think we could do something similar, but to step back a second, fundamentally what we need for password reset is something that can generate a token that has the following properties:

* Uniquely identifies a particular user.
* Cannot be spoofed or modified by an attacker.
* Can only be used once.

With this in mind, I think we could generate tokens using something like:

```python

import hmac
import itsdangerous

from pyblake2 import blake2b

def generate_user_hash(user_id, last_login, user_password):
    return blake2b("|".join(map(str, [user_id, last_login, user_password])).encode("utf8")).hexdigest()

def generate_reset_token(user_id, last_login, user_password):
    s = itsdangerous.URLSafeTimedSerializer('secret-key')
    return s.serialize({
        "user.id": user_id,
        "user.hash": return generate_user_hash(user_id, last_login, user_password),
    })

def verify_reset_token(token):
    s = itsdangerous.URLSafeTimedSerializer('secret-key')
    data = s.loads(token)
    user = get_user_from_id_or_none(data["user.id"])  # Fake function!

    if user is not None:
        return False

    if hmac.compare_digest(data["user.hash"], generate_user_hash(user.id, user.last_login, user.password)):
        return True

    return False
```

This code can't be lifted or used directly, and it's largely taken from the idea that exists in Django, but it has a few important properties that makes this work:

* It bakes the user.id into a signed token that is passed to the end user via email to recover the account. This ensures that we don't need to map the OTK to a particular user, since itsdangerous will ensure that nobody can tamper with this, or generate their own token.
* It includes a hash of some state of the current user, which if it changes will ensure that any existing tokens no longer work. This state includes:
    * The id of the user, this just ensures that the hash has some value that ties it to the specific user itself rather than just be side along side.
    * The last login of the user, anytime we log in the user (including via password reset) we should be using the ``warehouse.accounts.views:_login_user`` function (this should maybe be moved to warehouse.accounts and made not private) which will ensure that the token can only be used once (since reseting a password should log in the user once the password has reset, and that will change the last_login).
    * The password hash, this will ensure that if the user changes their password any existing password reset tokens for that user will be invalidated. This is safe to do because blake2 is a one way hash, so nobody can reverse the existing hash from it. The best they can do is find some combination of (user.id, last_login, and password_hash) that matches but they still can't reverse engineer the password from that, and since they don't know the real last_login or the salt for the password_hash they can't discover any information from it. We *could* use ``User.password_date`` instead of ``User.password`` for this instead which would achieve basically the same thing I believe without even having the possibility of leaking a password hash.

All together, I think this kind of thing is a better solution since it doesn't require any sort of storage on our end of things. We might still want a password reset service, just to deal with the fact it's parametrized by a config value, but that could be something like:

```python
import hmac

from warehouse.utils.crypto import BadData, URLSafeTimedSerializer


class InvalidPasswordResetToken(Exception):
    pass

class PasswordResetService:
    
    def __init__(self, key, max_age, user_service):
        self.signer = URLSafeTimedSerializer(key)
        self.max_age = max_age

    def _generate_hash(self, user):
        ...   # Actually generate a hash

    def generate(self, user_id):
        user = self.user_service.get_user(user_id)
        return self.signer.dumps({
            "user.id": user.id,
            "user.hash": self._generate_hash(user),
        })

    def verify(self, token):
        try:
            data = self.signer.loads(token, max_age=self.max_age)
        except BadData:
            raise InvalidPasswordResetToken

        user = self.user_service.get_user(data.get("user.id"))
        if user is None:
            raise InvalidPasswordResetToken

        user_hash = self._generate_hash(user)
        if not hmac.compare_digest(data.get("user.hash"), user_hash):
            raise InvalidPasswordResetToken

        return True
```


Does all of that make sense?

gsb-eng:
@dstufft I totally agree for this.

> since reseting a password should log in the user once the password has reset, and that will change the last_logi

Do we really have to perform login just after reset?

dstufft:
> Do we really have to perform login just after reset?

Isn't it a bit unfriendly to ask them to re-enter (for a third time!) the password they just reset?

gsb-eng:
Yeah, will implement that too.

dstufft:
Great!

As an additional thing, after thinking about it more, I think we should definitely use ``User.password_date`` rather than ``User.password`` in the user hash here. It will achieve the same thing (invalidate any existing password reset tokens after the password has been reset, or changed at all) with two beneficial properties:

* It will work even the person doesn't change their password at all and simply resets it to the same thing they already had (or resets it, then changes it back).
* It completely eliminates any ability to even theoretically figure out what the password hash value might be, at most you can theoretically figure out what the ``User.password_date`` might be.

I *think* if we use that, then we don't *need* to use ``User.last_login`` either, unless we feel like a user logging in should invalidate any existing password reset tokens. I'm sort of on the fence about that, but I think it makes sense to invalidate reset tokens on login anyways.

gsb-eng:
> It will work even the person doesn't change their password at all and simply resets it to the same thing they already had (or resets it, then changes it back).

One thing here is: we don't change the password if user tries to resets to the old password. Instead we'll throw an error "Password shouldn't match with previous one", In this case user might just login with old password or might try reset again with new password.

**Case-1**: Login with old password.

   > I think if we use that, then we don't need to use User.last_login

 I think in this case it's better to have `User.last_login`, so that we can make that existing OTK invalid just after login.

**Case-2**: Try again with new password.
  Since we are doing login just after reset, `User.last_login` also gets updated (along with `User.password_date`). We might not require `last_login` here(`password_date` is enough to invalidate), but it helps to invalidate the token in `Case-1`.

gsb-eng:
@dstufft I think I've implemented everything that we've discussed here, Could you please have a final look. So that I can make sure all the tests are passed.

gsb-eng:
@dstufft My apologies for the delay, I've added tests for this feature. I think this is good to go now.

gsb-eng:
@dstufft Could you please verify and merge this?

gsb-eng:
@dstufft : Could you please check this and merge?

gsb-eng:
@dstufft : Any outstanding issues with this?

nlhkabu:
Thanks for your work on this @gsb-eng and sorry for our delay in getting back to you.

@dstufft if you are happy with this, I will review / update the templates to match the new styles in #1980

gsb-eng:
@dstufft could you please merge this?

gsb-eng:
@dstufft  Please let me know if you can merge. I'll make sure there are no conflicts.

brainwane:
Thanks for your pull request and sorry for the slow response!

The folks working on Warehouse have just gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and will in December kick off work towards [the first milestone: redirecting pypi.python.org to pypi.org](https://github.com/pypa/warehouse/milestone/1). We will have more responses to your work after that kickoff.

Thanks again.

brainwane:
@gsb-eng: in the Warehouse developer [meeting today](https://wiki.python.org/psf/PackagingWG/2017-12-19-Warehouse) we decided that issue #1228 definitely belongs in [the first milestone we want to deliver, the MVP for package maintainers](https://github.com/pypa/warehouse/milestone/8).

So this feature is a priority for us, and we'd love it if you had time to rebase and update this branch in the next few days! And then go ahead and ping Donald here and/or in IRC (`#pypa-dev` on Freenode) to let him know it's ready to review.

Thanks!

gsb-eng:
I'll do this over the weekend.

gsb-eng:
@dstufft  This is green now, could you please check?

nlhkabu:
ping @di or @ewdurbin - are either of you able to validate for @gsb-eng ?

di:
This is next up on my to-do list. I think this PR is mostly there but there are a few small things I think will need addressed before this is merge-able. I will likely just push directly to @gsb-eng's branch here if that's ok with them!

There's also some things that need done before this can be deployed, like configuring `pypi.org` with our SMTP service, which I believe @ewdurbin was taking care of (or might have already taken care of).

nlhkabu:
great, thanks for the update @di :grin:

ewdurbin:
@di mailgun domain setup and credentials added to Heroku for warehouse-test and warehouse-prod.

gsb-eng:
@di yes you can do that.

di:
I've moved these commits into #2764. Thanks for your contribution @gsb-eng!

PR #1263, opened 2016-06-11
Css refactor

nlhkabu:
Major CSS refactor moving codebase over to BEM naming convention.  Cleaned up components and added documentation.


nlhkabu:
html-linter is throwing linter errors because of the new BEM naming convention.
It does not look immediately obvious to me how to get around this :(

dstufft:
We can probably just make html linting to shut up about that particular kind of error.

nlhkabu:
@dstufft - FYI we will need to merge this PR on pypi-theme https://github.com/python/pypi-theme/pull/2 at the same time, so that the HTML/CSS remains in sync.

dstufft:
@nlhkabu pypi-theme doesn't auto deploy, so the easiest thing to do will be to merge that PR, then I'll make a release of pypi-theme and this PR can be updated to change the pypi-theme version.

dstufft:
By the way, I think you can tell html_linter.py to shut about about this by adding ``names`` to the disabled list in the [Makefile#L118](https://github.com/pypa/warehouse/blob/5584c79226559b7635ac0c9e8423988761f03b02/Makefile#L118).

nlhkabu:
> By the way, I think you can tell html_linter.py to shut about about this by adding names to the disabled list in the Makefile#L118.

aha!  Thank-you :)

I just added a few more things to my todo list on this PR (see https://github.com/pypa/warehouse/issues/813#issuecomment-225411735), but hopefully we can get it merged in the next few days :)

nlhkabu:
ok @dstufft this is ready to merge with a new release of pypi-theme :)

Closes #1263

Issue #1060, opened 2016-04-07
Author of package not shown


kennethreitz:
My package, [Requests](https://warehouse.python.org/project/requests/#description) is registered and owned by myself. I also have two maintainers who contribute significantly to the project.

Currently, in the Warehouse UI, there appears to be no indication that I am the owner of my package, or that it is owned by me. This is very important to me, as a package author.

kennethreitz:
By contrast, the current PyPi explicitly shows "Author: Kenneth Reitz".

kennethreitz:
So, perhaps this isn't as much about surfacing ownership of the package namespace as it is surfacing the "Author" metadata.

kennethreitz:
I personally think that should be one of the most visible pieces of information presented for a package — it's one of the first things I typically look at. 

In my mind, it belongs up in the blue area, next to the package name.

e.g.

```
requests 2.9.1
by Kenneth Reitz
Python HTTP for Humans.

dstufft:
To be clear, ``Author: Kenneth Reitz`` is a different thing than the Owners/Maintainers. What now looks like:

<img width="299" alt="screen-shot-2016-04-07-13-44-01" src="https://cloud.githubusercontent.com/assets/145979/14360990/d63a7074-fcc6-11e5-904b-50441cd8c38c.png">

used to look like:

<img width="437" alt="screen-shot-2016-04-07-13-45-39" src="https://cloud.githubusercontent.com/assets/145979/14361033/1406bd22-fcc7-11e5-83f3-d1ccb3ed555f.png">

As far as the "What users can manage this project", I personally think making the list "flat" and making Owners and Maintainers (in the PyPI permissions sense) show up the same is a reasonable thing to do.

That being said, there is some metadata that we're not currently exposing which are: Platform, Supported-Platform, Author, Author-Email, Maintainer, Maintainer-Email, License, Classifier, Requires-Dist, Provides-Dist, Obsoletes-Dist, Requires-Python, Requires-External, Requires, Provides, Obsoletes (but some of these we're not exposing by design).

Related to this specific issue, we're not exposing the Author and Maintainer information and we should be. In the current metadata Author, Author-Email, Maintainer, and Maintainer-Email just take a single arbitrary string so we mostly just assume it's a single thing, however in the future we will likely make this able to take a list of values, so whatever design we pick should be able to handle multiple entries.

[...]

PR #1267, opened 2016-06-13
Omit the uploader information from the UI

dstufft:
This is inspired partially by #1060, and it removes the "uploaded by" information from everywhere in the UI. I think we should maybe not remove it from everywhere, but we should instead remove it from everywhere it is surfaced *now* and instead add it in as part of the "Version History & Changelog" view so people can get a list of the history of the releases of the project, when they were releases, and by whom they were released.

This might need a bit of tweaking still, the top bar looks like this now:

screen-shot-2016-06-12-21-58-06

However, I'm not sure that the ``Uploaded <on>`` thing actually makes sense still, is that information that important to have always available? Should we just rely on the "Version History & Changelog" page for this and give the project links more room to flow over to the right in? In any case, if we keep it there it's currently in a ``<p class="author">`` and that should probably be changed to something else since it doesn't expose the author anymore.

I don't know if this is the right thing to do or not, so I'll leave this for @nlhkabu to either merge, reject, or ask for further refinement.

### DIFF
--- a/warehouse/packaging/views.py
+++ b/warehouse/packaging/views.py
@@ -12,7 +12,6 @@
 
 from pyramid.httpexceptions import HTTPMovedPermanently, HTTPNotFound
 from pyramid.view import view_config
-from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 
 from warehouse.accounts.models import User
@@ -40,7 +39,6 @@ def project_detail(project, request):
     try:
         release = (
             request.db.query(Release)
-                      .options(joinedload(Release.uploader))
                       .filter(Release.project == project)
                       .order_by(Release._pypi_ordering.desc())
                       .limit(1)
--- a/warehouse/static/sass/blocks/_package-links.scss
+++ b/warehouse/static/sass/blocks/_package-links.scss
@@ -1,15 +1,11 @@
 /*
   A horizontal bar on the package detail page that includes the main links
-  for each page, as well as author information.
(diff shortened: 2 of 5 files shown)


nlhkabu:
I agree that we should remove the uploaded on altogether, and only expose it in the changelog.  As you mention, this will provide more space for links to overflow.  This means we can remove the author class and any css that's creating the columns to split.

dstufft:
@nlhkabu It occurs to me that we really have two pieces of information we can surface here. We have "who created the *release*" and "who uploaded one or more *files*". For your average case where a single person uploads all of the files and nobody goes out of there way to create a release without uploading a file, then this will all be the same person. However there are cases where you can image having multiple people involved in an upload (maybe one person uploads the sdist and individual people upload wheels for different archiectures, or maybe one person uploads the sdist and then a bot downloads it, compiles it, and uploads wheels).

So I guess one question to ask ourselves when it comes time to expose this information is what do we mean by "uploader", do we mean the person who created the release (typically automatically created when the first file is uploaded) or do we mean who uploaded the first file, or who uploaded *any* file?

It might be reasonable to support multiple uploaders in the design and have it just do ``unique(<list of users who both created the release and uploaded any files>)`` where this will typically always be a single user but would reflect cases where it's not without overcomplicating the UI.

It might also be reasonable to make the changelog only surface who created the release, and add information to the file downloads about who uploaded each individual file.

Anyways, I updated this PR so it should be good to merge if it all looks good to you.

PR #1268, opened 2016-06-13
Try to update getting started instructions on OS X.

Carreau:
I have no clue what I am doing, but with boot2docker I get an
incompatible client/server error message.

With this changes I'm currently at the step where docker at least accept
to pull images.

---

Docker is currently trying to install stuff in the VM. I changed a lot of stuff so unsure if this is the only thing that needs to be done for instructions to work, but at least a step forward.


dstufft:
✨

Carreau:
Thanks ! Now that I have a setup that almost works, I might be able to give (useful) feedback and maybe make some patches.

Issue #1252, opened 2016-06-09
Decide on terminology, document what we decide, be consistent 


dstufft:
In #1250 (implemented in #1251) @nlhkabu asked for all of the references to "Project" be changed to "Package", at the risk of incorrectly summarizing her reasoning, she did this because we had introduced the word "package" in a few locations in Warehouse (namely the help page), and because packaging.python.org was using "package" as well. Thus, switching project to package was an attempt to unify the terminology.

I originally purposely used the word project here because it was the supporting definition used in PEP 426. It was the supporting definition in PEP 426 in an attempt to disambiguate the term "package". Currently the term "package" can refer to:

* Something you import.
* A name on PyPI (such as "Django").
* A file that you download and install from PyPI (such as "Django-1.9.tar.gz").
* A generalized name of "something you work on".

So the term "Project" was chosen to represent the second of those things.

There are a couple issues with using the term Project however:

* A lot of people are still going to use the ambiguous term, regardless of what we call it.
* PyPI is named, "Python Package Index" (although that name is bad for other reasons as well).
* The PyPug uses the term "packages" in some locations (though I think it also uses the term "projects" in some locations.

There are several relevant issues on the PyPug issue tracker like, pypa/python-packaging-user-guide#89, pypa/python-packaging-user-guide#87, pypa/python-packaging-user-guide#199, pypa/python-packaging-user-guide#235, pypa/python-packaging-user-guide#106, and pypa/python-packaging-user-guide#25 and more on distutils-sig that I don't feel like digging up right now (yea, there's some history here).

For my part, I think the terminology of "Project" on Warehouse is clearer than the term package and we should go back to using it (and we should be consistent about using it). This leaves the name "PyPI" as a kind of unfortunate vestige, but @ncoghlan suggested tweaking the name to Python Packaging Index, which steps around that particular ambiguity. I'm not sure how I feel about that in particular, but I do know that I feel that the change as implemented in #1251 makes PyPI more difficult to understand. Even on the very front page of Warehouse now, we have:

<img width="658" alt="screen-shot-2016-06-09-09-38-06" src="https://cloud.githubusercontent.com/assets/145979/15931488/f7709558-2e25-11e6-80e6-4f8e122e67d3.png">

Which makes people have to sit there and figure out what kind of "package" this is (is it "Django" or is it "Django-1.0.tar.gz"?).

[...]

PR #1278, opened 2016-06-16
Copyedit the help page

ddbeck:
Following up on #1252, @nlhkabu asked me about revising the help page. I gave the page a once over and here's what I was going for:

- make the transition from clicking the "Bugs & Feedback" footer link to the question text on the help page a little easier
- improve the skimmablility of the page
- reemphasize the CoC because I cannot get enough of _that_

Plus I made a couple other consistency and style changes while I was at it! I also tried to keep the commits small, so it should be easy to keep only the changes that read well to another person.

### DIFF
--- a/warehouse/templates/pages/help.html
+++ b/warehouse/templates/pages/help.html
@@ -28,7 +28,7 @@
 
       <h3 id="installing" class="common-question">What is a package and how do I install one?</h3>
 
-      <p>A 'package' in the context of PyPI is a bundle of software to be installed.</p>
+      <p>A <em>package</em> in the context of PyPI is a bundle of software to be installed.</p>
       <p>Packages on PyPI are developed and shared by other members of the Python community so that you can use them in your own projects.</p>
       <p>To learn how to install packages, visit the <a href="//packaging.python.org/en/latest/installing/">installation tutorial</a> on the <a href="//packaging.python.org/">Python Packaging User Guide</a>.</p>
 
@@ -61,7 +61,7 @@
 
       <h3 id="contributing" class="common-question">How can I help?</h3>
 
-      <p>We have a huge amount of work to do to continue to maintain and improve PyPI (aka the Warehouse project), and would love to see some new faces working on the project.</p>
+      <p>We have a huge amount of work to do to continue to maintain and improve PyPI (also known as the Warehouse project), and would love to see some new faces working on the project.</p>
       <p>You <strong>do not</strong> need to be an experienced open-source developer to make a contribution - in fact, we'd love to help you make your first open source pull request!</p>
       <p>If you have skills in Python, HTML, SCSS, or JavaScript, then please take a look at the <a href="//github.com/pypa/warehouse/issues">issue tracker</a>. If you're interested in working on a particular issue, leave a comment and we can guide you through the contribution process.</p>
 
--- a/warehouse/templates/pages/help.html
+++ b/warehouse/templates/pages/help.html
@@ -43,11 +43,13 @@
 
       <h3 id="feedback" class="common-question">Where can I report a bug or provide feedback?</h3>
 
-      <p>We welcome <em>constructive</em> feedback. If you would like to suggest an improvement, please use our <a href="//goo.gl/forms/IaPbMVAKnYDP2xV92">feedback form</a>. Common feedback will be transferred to our issue tracker to be addressed by the development team.</p>
+      <p>We welcome <em>constructive</em> feedback and bug reports.
 
-      <p>If you are experiencing a <strong>bug</strong>, please log an issue directly on our <a href="//github.com/pypa/warehouse/issues">issue tracker</a>.</p>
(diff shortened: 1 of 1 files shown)


nlhkabu:
Thanks for this PR @ddbeck - all of the changes are great.

I wonder whether or not we should also change the first question (or add a new section) to deal with the projects/release/package distinction discussed in https://github.com/pypa/warehouse/issues/1252 ?

ddbeck:
I tried rewriting it a few times, but every time I changed it, it seemed like I was changing it for the sake of changing it. Is there something about it you think is inadequate at this point?

nlhkabu:
Mainly the distinction between a project and a package. I think the average punter is going to confuse the two (I know I did).

Maybe we could break it down into two questions? For example:

- What is the difference between a project and a package?
- How do I install packages?

ddbeck:
OK, I'll give it another try soon!

ddbeck:
Alright, I separated the "What's a package?" question from the "How do I install a package?" question and then I expanded the former significantly. I tried—perhaps unsuccessfully—to bridge the gap on *file* and *package* here. We could conceivably go a step more and rewrite the second question as "How do I install a file that's on PyPI?" Happy to hear feedback on this.

Also, it's only fair to say that I cribbed from @dstufft's commentary on #1252. Thanks, Donald!

Oh yeah, I used requests as an example project, but that was just a guess at something popular on my part—maybe it's in our interest to reserve an example project for documentation purposes (à la [RFC 6761](https://tools.ietf.org/html/rfc6761)).

dstufft:
https://pypi.io/project/spam/ ?

dstufft:
Oh, and this look good to me, I'll let @nlhkabu read over it and merge it if she agrees :D

warehouse/templates/pages/help.html
```diff
@@ -13,53 +13,76 @@
...
-      <p>A 'package' in the context of PyPI is a bundle of software to be installed.</p>
-      <p>Packages on PyPI are developed and shared by other members of the Python community so that you can use them in your own projects.</p>
+      <h3 id="packages" class="common-question">What's a package, project, or release?</h3>
+
+      <p>We use a number of terms to describe software available on PyPI, like <em>project</em>, <em>release</em>, <em>file</em>, and <em>package</em>. Sometimes those terms are confusing because they're used to describe different things in other contexts. Here's how we use them on PyPI:</p>
+
+      <p>A <em>project</em> on PyPI is the name of a collection of releases and files, and information about them. Projects on PyPI are made and shared by other members of the Python community so that you can use them.</p>
+
+      <p>A <em>release</em> on PyPI is a specific version of a project. For example, the <a href="//pypi.io/project/requests/">requests</a> project has many releases, like <em>requests 2.10</em> and <em>requests 1.2.1</em>. A release consists of one or more <em>files</em>.</p>
+
```
nlhkabu: @dstufft is there a better way to make an internal link to a package here?

warehouse/templates/pages/help.html
```diff
@@ -13,53 +13,76 @@
...
+      <p>A <em>file</em>, also known as a <em>package</em>, on PyPI is something that you can download and install. Because of different hardware, operating systems, and file formats, a release may have several files (packages), like an archive containing source code or a binary <a href="//pypi.io/project/wheel/#files">wheel</a>.</p>
+
```
nlhkabu: and here?

warehouse/templates/pages/help.html
```diff
@@ -13,53 +13,76 @@
...
+
+      <h3 id="installing" class="common-question">How do I install a package from PyPI?</h3>
```
nlhkabu: Maybe given the definition above we can change this to:

"Howto I install a file (package) from PyPI?

warehouse/templates/pages/help.html
```diff
@@ -13,53 +13,76 @@
...
+
       <p>To learn how to install packages, visit the <a href="//packaging.python.org/en/latest/installing/">installation tutorial</a> on the <a href="//packaging.python.org/">Python Packaging User Guide</a>.</p>
```
nlhkabu: And again here:

"to learn how to install files (packages), visit the ...."

warehouse/templates/pages/help.html
dstufft: You can do ``{{ request.route_path('packaging.project', name='requests') }}`` to autogenerate a thing like ``/project/requests/``. This will automatically update if we move URLs around, *but* it requires you to know the name of a package that exists on both ``test.pypi.io`` and ``pypi.io`` else it'll link to a page that 404s.

dstufft:
I'd like to get this merged, do you think you'd be able to get the couple of comments sorted out @ddbeck or should I take this over?

ddbeck:
Oh, I'm sorry! I'll update this PR tomorrow. Sorry for letting this get
away from me.

ddbeck:
OK, I've checked in the terminology fixes. I didn't change the URLs, because I don't think relying on the existence of a project name on both `test.pypi.io` and `pypi.io` to be such a hot idea.

Hopefully this PR is ready to go now, but let me know if there are more changes to make. Thanks!

nlhkabu:
The copy changes look good, thanks @ddbeck :)
I'll let @dstufft comment on the URLS, but from my side this is ready to merge.

PR #1279, opened 2016-06-17
Upgrade pypi-theme to 1.29

dstufft:


dstufft:
Forcing this to merge since it's just bumping the pypi-theme dependency.

PR #1282, opened 2016-06-17
Add 'make serve' to the detailed installation instructions

mbacchi:
This was added to the "Quickstart" section of the Getting Started guide, but not the detailed instructions.  Hoping this will make it easier for beginners.


dstufft:
✨

Issue #1272, opened 2016-06-15
Configure send_mail.


gsb-eng:
Warehouse doesn't have any send_mail functionality configured, Sending mail is necessary for functionalities like password recovery, user registration confirmation, etc.

**Initial thoughts:**

- Sending email should happen asynchronously.
- Should use celery to do the actual sending.
- Should handle failure cases.

gsb-eng:
@dstufft  : Could you please add if anything else needs to be considered?

gsb-eng:
I'll be working on this.

gsb-eng:
Broadly we can categorize warehouse email requirement into two types.

1) **Instant Emails**:

 **Use Cases** :  Password recovery, registration confirmation, alert when a new maintainer or author added, etc.  I think 100% of emails goes through PyPi comes under this category.

   **Implementation:** We can use a celery task to send email asynchronously, Worker should be always running to handle queued tasks.

   **Handling Failures:**  If any exception occurs in the first attempt, we've to handle two cases here.
   1. retry = True:
        We'll try re-sending for 3 to 4 times (consensus required) and give up after that.
   2. retry = False:
       Give up.

Sample celery task:
```
@app.task(bind=True,
def send_mail(task, retry=True, max_retries=3):
     try:
         mailer.send_immediately(message)
    except Exception, exc:
        if retry:
            task.retry(countdown=1, max_retries=max_retries, exc=exc)
        raise
```
   

2) **Scheduled Emails**:

   **Use Cases:** News letters, some mass updates, etc.

   **Implementation:** We can use a celery task to send email asynchronously, Worker should run at defined time to handle the queued tasks.

 **Handling Failures:** We can maintain a different queue for the failed tasks and a worker should be running always to handle the failure queue.

Sample celery task:
```
@app.task(bind=True)
def send_email(task):
     try:
         mailer.send(message)
    except Exception, exc:
        send_email(args, queue='failure_queue')
```

Consensus required on below queries:
1. Do we really need **Scheduled Emails**?
2. If a instant email fails, how many times we've to try re-send the email?

gsb-eng:
@dstufft : Can I close this issue now?

PR #1286, opened 2016-06-20
Adding send_email feature.

gsb-eng:
Implementation for #1272.


gsb-eng:
@dstufft  This is the basic implementation of email feature, Could you please verify and let me know your thoughts?

gsb-eng:
@dstufft : Could you please review and let me know your thoughts?

warehouse/email.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pyramid_mailer import get_mailer
+from pyramid_mailer.message import Message
+
+from warehouse import celery
+
+
+@celery.task(bind=True, ignore_result=True, acks_late=True)
```
dstufft: I wonder if we really want ``acks_late`` here? The trade off is basically what behavior do we want on certain failures (particularly, a worker node dying). With ``acks_late=True`` we're almost certainly going to deliver the message at least once, but we might deliver it 2 or more times. With ``acks_late=False`` we're only going to deliver the message at most once, but we might deliver it zero times.

So the underlying question becomes then whether it's better to send extra emails in failure conditions or send no emails in failure conditions.

warehouse/email.py
```diff
@@ -0,0 +1,37 @@
...
+def send_email_task(task, request, body, recipients, subject, countdown,
```
dstufft: Passing ``countdown`` and ``max_retries`` into this function is a bit weird. Celery already handles that for the general case for us. Is there any reason why sending email needs special handling for these two parameters?

warehouse/email.py
```diff
@@ -0,0 +1,37 @@
...
+    try:
+        mailer = get_mailer(request)
+
+        message = Message(
+            body=body,
+            recipients=recipients,
+            sender=request.registry.settings['mail.sender'],
+            subject=subject
+        )
+        mailer.send_immediately(message)
```
dstufft: This is probably the only line that should be wrapped inside of the ``try ... except`` block here, because it's the only thing that should have transient failures. ``get_mailer()`` and ``Message()`` should either always work or should always fail so retrying here probably doesn't matter.

warehouse/email.py
```diff
@@ -0,0 +1,37 @@
...
+    except Exception as exc:
+        task.retry(countdown=countdown, max_retries=max_retries, exc=exc)
```
dstufft: As mentioned above, I'm not sure that we need special treatment for ``countdown`` and ``max_retries`` for this task, so just letting celery handle this itself seems to be a reasonable way of implementing that. This would make this line just need to be:

```python
task.retry(exc=exc)
```

warehouse/email.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def send_mail(body, recipients, subject, countdown=1, max_retries=3):
```
dstufft: I don't think we need this shim function here to "hide" the fact that sending email is done asynchronously in calling code. I think it's perfectly reasonable for views and such to simply do ``send_mail.delay(...)``.

dstufft:
Hey!

Sorry for the delay in reviewing this. I've gone through and reviewed it and made some inline comments. Overall this looks like it's going in a good direction.

warehouse/email.py
gsb-eng: Lets consider this situation in different contexts, so that it's easy to decide whether to send multiple emails or no email in the event of a worker failure.

If we consider `acks_late=False`:

**1. Password reset email:**  After successful completion of reset request, we'll show a message to the user to verify their email for the process to be followed. Then they don't receive any email in the event of failure, it's fair that user can come back and request for reset again. May be this time we'll deliver the message and user will'll be happy :)

Here we've to consider, what if the email fails to deliver multiple times?
  
**2. User registration confirmation:**  May be it's fine if the email fails to deliver, but if we decide like it has to be delivered: Unlike password reset it's not fair that user can try register again, The only possibility is that re-queue the message(I think acks_late=True does the same).

**3. New author/maintainer added notification:** I think this is similar to registration confirmation.

If we consider `acks_late=True`:

In the event of worker failure, message will be re-queued for all the scenarios mentioned above and there is a possibility that messages will be delivered multiple times.

**My opinion:** 
1. If we fail to deliver the message, user might loose the trust on the system. 
2. If we deliver multiple emails, there are two possibilities.
  a) User may feel like it's spam.
  b) Since it's a known email, he/she may be fine with that.

**Note: ** I think we've to consider this "worker is not known to crash" :)

warehouse/email.py
gsb-eng: :+1: Sure

warehouse/email.py
gsb-eng: Considering `send_mail` should be a generic function, I thought it's better to make this flexible, so that this can be used anywhere in the application with required configuration (Some times we might require to retry more than 3 times or less than 3 times).

warehouse/email.py
gsb-eng: Yeah that was my intention. But I think you are right, it's better to make that visible so that it's easy to understand. I'll change this.

warehouse/email.py
dstufft: I don't think we need that control here. I'd prefer to leave it out until/if we end up needing it.

warehouse/email.py
dstufft: Yea, I'm not opposed to sending extra emails (particularly since it's only in the crashed worker case), just thinking it out if we want that or not. It seems like a reasonable choice.

gsb-eng:
@dstufft : I think this is complete now, could you please review and let me know if anything else needs to changed?

dstufft:
✨

PR #1289, opened 2016-06-24
Provide canonical URL for package detail pages

### LINKED ISSUE #968: Define canonical URLs on package detail pages
One issue with the current PyPI is search engines delivering links to the details page for a specific old version of a package, rather than the latest version.

I was looking into the use of canonical URLs for the CPython documentation, and realised they may be an appropriate solution for the "search engines pointing to old versions" problem on PyPI as well: http://docs.readthedocs.org/en/latest/canonical.html

hroncok:
...as well as a mechanism to provide it anywhere when required

Fixes #968 

I gave it a thought and according to [this](https://audisto.com/insights/guides/28/#choose-full-urls-over-shortened-urls-for-rel-canonical-29f0d99af867efd897fa2074fcb81bec), absolute URLs should be used in rel=canonical links. There might be an issue, that the canonical URL currently changes whether the webpage is served over http:// or https:// but https://pypi.io/ does the redirect, so I don't think it should be a problem.

I did not accompanied this change with a test, as I'm not sure how to properly test it. My idea is to test that rendered HTML pages with specific package version always have the appropriate rel=canonical link, however I'm not sure where would be the good starting point for this, help is appreciated.

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -31,6 +31,9 @@
 
     <link rel="alternate" type="application/rss+xml" title="RSS: 40 latest updates" href="{{ request.route_path('rss.updates') }}">
     <link rel="alternate" type="application/rss+xml" title="RSS: 40 newest packages" href="{{ request.route_path('rss.packages') }}">
+    {% if self.canonical_url() %}
+    <link rel="canonical" href="{% block canonical_url %}{% endblock %}">
+    {% endif %}
 
     {# TODO: We need a solution to https://github.com/pypa/warehouse/issues/833
      #       before we can progress on this. Currently this breaks html_include
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -60,6 +60,8 @@
 
 {% block title %}{{ release.project.name }}{% endblock %}
 
+{% block canonical_url %}{{ request.route_url('packaging.project', name=release.project.name) }}{% endblock %}
+
 {% block content %}
 <section class="banner">
   <div class="package-header">


hroncok:
My idea for the test is something this:

```python
@pytest.mark.parametrize("package", [("Sphinx", "1.3.5")])
def test_canonical_url_is_present_in_package_detail(webtest, package):
    resp = webtest.get("/project/{}/{}/".format(*package))
    assert '<link rel="canonical" href="http://localhost/project/{}/">'.format(package[0]) in resp.body.decode('utf-8')
```

I just need to put the package in there, as during the test no packages are present :(

dstufft:
Hey, sorry for the delay in getting back to you here.

You should be able to add packages by doing something like:

```python
p = ProjectFactory.create()
rs = [ReleaseFactory.create(project=p) for _ in range(3)]
```

(Those classes are in ``tests/common/db/packaging.py`` file).

That should create a project, and then create three releases to go with it.

One other note, I think it'd be better to actually use xpath or csselect via webtest rather than doing a simple string test. That way we protect against say, accidentally commenting it out.

hroncok:
Thanks for help. However, as soon as I do this:

```python
def test_canonical_url_is_present_in_package_detail():
    proj = ProjectFactory.create()
    assert True
```

I get:

```python
sqlalchemy.exc.UnboundExecutionError: Could not locate a bind configured on mapper Mapper|Project|packages or this Session
```

dstufft:
Try adding a ``db_session`` keyword arg to the test.

hroncok:
Actually, db_request, but it helped, thanks a lot.

dstufft:
Ok, something very broken is going on here and after a few hours I've not yet figured out wtf is going on with the test suite here. I'm going to just commit this as is and later I'll see if I can't figure out why webtest and db_session don't play nicely together.

hroncok:
Thanks.

For the reference:

```python
def test_canonical_url_is_present_in_package_detail(db_session, webtest):
    release = ReleaseFactory.create(project=project)
    db_session.flush()  # ?
    db_session.commit()  # ?
    resp = webtest.get("/project/{}/{}/".format(release.name, release.version))
    assert '<link rel="canonical" href="http://localhost/project/{}/">'.format(release.name) in resp.body.decode('utf-8')  # needs a better check (DOM level)
```

PR #1290, opened 2016-06-25
Rename 'Version History & Changelog' to 'Release History'. Closes #1254

### LINKED ISSUE #1254: Rename 'Version History & Changelog' to 'Release History & Changelog'
This will also require renaming some of the CSS hooks.
What do you think @dstufft ?

timofurrer:
Renames 'Version History & Changelog to 'Release History' in Project
Detail Page according to #1254.

@nlhkabu you mentioned that this wording change also affects CSS hooks. What did you mean by that? Which hooks?


nlhkabu:
Thanks for the PR @timofurrer - It looks good :)

I'm thinking that it might make sense to change:

- `.history-timeline` to `.release-timeline`
- `.history-section` to `.release`

It would also make sense to change `.release-description` to `.project-description` to avoid confusion.

Each change would need to 
- Update the HTML, including the [BEM](http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/) elements (e.g. `.history-section__meta` would become `.release__meta`
- Rename the SCSS file
- Update the SCSS file, including comments
- Change the import in `warehouse.scss` to match the new file name

timofurrer:
@nlhkabu I've just renamed the CSS part of it. I've also changed the `.release-description` to avoid the confusion you mentioned.

warehouse/static/sass/components/_project-description.scss
```diff
@@ -2,12 +2,12 @@
   Styles the main description panel for an individual package.
```
nlhkabu: While we are here, can we also change this to say "Styles the main description panel for an individual project"

nlhkabu:
Thanks @timofurrer :)

PR #1293, opened 2016-06-28
pagination design

### LINKED ISSUE #1200: Use buttons for pagination
Replace current pagination on the search page with a button set.

The current pagination does not have a left and right arrow, but it would be nice to add them as per below.

```
<div class="button-set">
  <a class="wh-button disabled">Previous</a>
  <a href="" class="wh-button -highlight">1</a>
  <a href="" class="wh-button">2</a>
  <a href="" class="wh-button">3</a>
  <a href="" class="wh-button disabled">...</a>
  <a href="" class="wh-button">500</a>
  <a href="" class="wh-button">Next</a>
</div>
```
We'll need to translate 'previous' and 'next'.

Also, 'previous/next buttons should only be disabled if you are on the first/last page :)

nkanaev:
resolves #1200 

result:
screen shot 2016-06-28 at 14 29 42

### DIFF
--- a/warehouse/static/sass/blocks/_button.scss
+++ b/warehouse/static/sass/blocks/_button.scss
@@ -74,7 +74,7 @@
     border-color: $brand-color;
     background-color: $brand-color;
     color: $white;
-    z-index: index($z-index-scale, 'button');
+    z-index: index($z-index-scale, 'highlight-button');
 
     &:hover {
       background-color: darken($brand-color, 2);
--- a/warehouse/static/sass/settings/_z-index.scss
+++ b/warehouse/static/sass/settings/_z-index.scss
@@ -9,4 +9,4 @@
   See: https://www.smashingmagazine.com/2014/06/sassy-z-index-management-for-complex-layouts/
 */
 
-$z-index-scale: 'history-line', 'history-node', 'callout-block', 'callout-block-border', 'button', 'active-button', 'sticky-top', 'dark-overlay', 'filter-panel';
+$z-index-scale: 'history-line', 'history-node', 'callout-block', 'callout-block-border', 'button', 'active-button', 'highlight-button', 'sticky-top', 'dark-overlay', 'filter-panel';
--- a/warehouse/templates/search/results.html
+++ b/warehouse/templates/search/results.html
@@ -48,6 +48,49 @@
 <option {{ l20n(l20n_id) }} value="{{ value }}" {{ 'selected' if value == order else ''}}>{{ text }}</option>
 {%- endmacro %}
 
+
+{% macro pagination(page) -%}
+{% if page.items and page.page_count > 1 %}
+  {% set map = page.link_map() %}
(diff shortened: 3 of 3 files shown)


dstufft:
It looks like the test had failed due to some network hickups or something. I want ahead and restarted them and they're all green now. This looks good from my end, however @nlhkabu is going to have to review this since it's front end/design work.

nlhkabu:
Hi @nkanaev, thanks for this PR.

My preference here is to use a button group, so that we don't have to add specific CSS for the pagination (less code, less to maintain). It's also more visually consistent with the rest of the design :)

Based on the work you have done, is there any reason why the button group markup wouldn't work?

nkanaev:
i do agree that less css is better. i'd actually implemented it as a button set at first. but then noticed making "..." elements as unclickable links doesn't make sense, so I needed a logical way to add `span` element. in the end i've ended up implementing something bem'ish (though it's not Gallagher's one).

do have any suggestions on spans? maybe we should just use something like:

    <span class="wh-button">...</span>

nlhkabu:
Yes, let's try using a span with a button class.  If necessary, we can modify the button BEM, but it *should* work out of the box :)

You can also apply a disabled modifier to the span, e.g.

```
<span class="button button--disabled">...</span>
```

This will make it grey and give it a `not-allowed` cursor.

warehouse/static/sass/blocks/_button-set.scss
```diff
@@ -0,0 +1,26 @@
...
+    margin-left: -2px;
+    border-radius: 0;
+
+    &:first-child {
+      border-radius: 3px 0 0 3px;
+    }
+
+    &:last-child {
+      border-radius: 0 3px 3px 0;
+      z-index: 5;
```
nkanaev: hacky, had to create to resolve the overlapping border (grey color) of the last element while on the last page.

nkanaev:
final result:
screen shot 2016-07-04 at 14 43 29

nlhkabu:
Thanks for upading this @nkanaev :)

I'm still not convinced that a new button **set** is needed.  I'd much rather modify the existing button **group** to see if we can make it work.

Does this markup do the trick?

```
<div class="button-group"
  <a class="button button-group__button">Previous</a>
  <a class="button button-group__button">1</a>
  <a class="button button-group__button button--highlight">2</a>
  <a class="button button-group__button">3</a>
  <a class="button button-group__button">4</a>
  <span class="button button-group__button button--disabled">...</a>
  <a class="button button-group__button">382</a>
  <a class="button button-group__button">Next</a>
</div>
```

nkanaev:
thanks, totally have missed `button-group`.
had to adjust a little bit - other buttons were overlapping highlight buttons on mouse hover, which is visible only when overlapping button is set to disabled.

nlhkabu:
Merged :sparkles: 

Thanks very much for this @nkanaev :)

PR #1341, opened 2016-07-02
Don't disallow linking directly to https pages

### LINKED ISSUE #1340: Reasses 'no prefix' rule on HTML linter
As per https://github.com/pypa/warehouse/pull/1339#issuecomment-230109083 we should consider turning off the `no-prefix` rule on the HTML linter so that we can specify a link should use https.

We will also need to update any links that do not currently have a prefix, as well as the HTML styleguide documentation.

dstufft:
Fixes #1340

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ saucelabs:
 lint: .state/env/pyvenv.cfg
 	$(BINDIR)/flake8 .
 	$(BINDIR)/doc8 --allow-long-titles README.rst CONTRIBUTING.rst docs/ --ignore-path docs/_build/
-	$(BINDIR)/html_lint.py --disable=optional_tag,names `find ./warehouse/templates -path ./warehouse/templates/legacy -prune -o -name '*.html' -print`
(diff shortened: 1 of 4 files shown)


nlhkabu:
thanks @dstufft - can we also update `frontend.rst` with:

```
HTML Code Style
---------------

Warehouse follows the
`Google HTML style guide <https://google.github.io/styleguide/htmlcssguide.xml>`_,
which is enforced via linting with
`HTML Linter <https://github.com/deezer/html-linter>`_.

Exceptions:

- Protocols can be included in links - with ``https`` preferred
- All HTML tags should be closed

We also allow both dashes and underscores in our class names, as we follow the
`Nicholas Gallagher variation
<http://nicolasgallagher.com/about-html-semantics-front-end-architecture/>`_
of the `BEM naming methodology <https://en.bem.info/>`_.

More information on how BEM works can be found in
`this article from CSS Wizardry
<http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/>`_.

```

nlhkabu:
assuming, of course, you are ok with the text :)

PR #1344, opened 2016-07-02
Integrate statuspage.io into footer

### LINKED ISSUE #604: Integrate statuspage.io
We'll want to integrate statuspage.io so that people can get a link to the status.python.org incidents when one is occuring. They have a web API that we are currently using via javascript on pypi.python.org.

See: https://status-page-blog.herokuapp.com/embed-system-status-within-your-webapp

berkerpeksag:
Fixes #604

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,6 +50,7 @@ app:
     FILES_BACKEND: "warehouse.packaging.services.LocalFileStorage path=/app/data/packages/ url=http://files.example.com/packages/{path}"
     RECAPTCHA_SITE_KEY: "${RECAPTCHA_SITE_KEY}"
     RECAPTCHA_SECRET_KEY: "${RECAPTCHA_SECRET_KEY}"
+    STATUSPAGE_URL: https://2p66nmmycsj3.statuspage.io
 
 web:
   extends:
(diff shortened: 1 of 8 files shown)


berkerpeksag:
Fixed test failure and linting issue.

nlhkabu:
Looks great from my side.

dstufft:
Reviewing this, I'd prefer it if we didn't bake the statuspage.io ID into the code here. I'd prefer it if instead it was made a setting. So you'd want to add a new setting to the list in ``warehouse/config.py`` and then read that setting in ``warehouse/csp.py``.

For passing that data into javascript, you'd then do something like:

```
{% if request.registry.settings.get("statuspage.domain") $}
<li><a href="https://status.python.org/">Status: <span data-statuspage-domain="{{ request.registry.settings['statuspage.domain'] }}">All Systems Operational</span></a></li>
{% endif %}
```

Then you could adjust the JS to pull this data out of this using something like what was doing for HTML including in ``warehouse/static/js/warehouse/utils/html-include.js``.

berkerpeksag:
Thanks for the idea. I have just implemented it.

berkerpeksag:
Fixed the merge conflict.

PR #1345, opened 2016-07-02
Update hashes of cffi, boto3 and msgpack-python

berkerpeksag:
I'm getting a huge warning when I try to install
dependencies manually.

I'm not sure if this is the correct fix, but it
did the trick for me.


dstufft:
@berkerpeksag Are you using a different version of CPython than 3.5 or something?

berkerpeksag:
Yes, I'm using CPython 3.4.3+ (I think I built it myself a while ago)

dstufft:
Ah, that explains why you got the error messages then. I only bothered to add the hashes for CPython 3.5 and the sdists. I don't have any problem including the other hashes though.

PR #1351, opened 2016-07-02
Move version number (Closes #1337)

### LINKED ISSUE #1337: Reduce package card height
Now that we have removed uploader information from the card, I think the version looks a bit daft on its own line:

screenshot from 2016-07-02 10-39-36

Suggested fix: move version to next to the package name.

Given that this will make the cards shorter, we can also afford to reduce the margin on each card, which will compress the interface somewhat.

controversial:
I moved the version number from below the title to be inline with it as described in #1337.

I took a few *creative liberties* beyond what's strictly described in the issue:
- The version number is floated to the right. I couldn't decide if it would be better to have the version number maintain the same distance from the right as the indentation level on the left, but the text overflows past this level, so I positioned the version number as in the screenshot below.
- The version number is prefixed with a `v`

if either is a problem I'll happily take it out.

The result:

I haven't yet played with the margins yet, but I'll do that now.

### DIFF
--- a/warehouse/static/sass/blocks/_package-snippet.scss
+++ b/warehouse/static/sass/blocks/_package-snippet.scss
@@ -40,6 +40,7 @@
 
   &__version {
     font-weight: $bold-font-weight;
+    float: right;
   }
 
   &__description {
--- a/warehouse/static/sass/blocks/_package-snippet.scss
+++ b/warehouse/static/sass/blocks/_package-snippet.scss
@@ -15,7 +15,7 @@
 .package-snippet {
   @include card;
   padding: ($spacing-unit / 2) 20px ($spacing-unit / 2) 75px;
-  margin: 10px 0 20px;
+  margin: 5px 0 15px;
   // Use png fallback
   background: $white url('../images/white-cube.png') no-repeat 0 50%;
   // Or svg if the browser supports it
--- a/warehouse/static/sass/blocks/_package-snippet.scss
+++ b/warehouse/static/sass/blocks/_package-snippet.scss
@@ -41,6 +41,9 @@
   &__version {
     font-weight: $bold-font-weight;
     float: right;
+    &::before {
+      content: "v";
+    }
   }
 
   &__description {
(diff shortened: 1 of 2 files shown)


TheJJ100100:
`creative liberties`

TheJJ100100:
@nlhkabu

controversial:
I'm going back and forth on the prefix of `v`. I'll do whatever someone official says. I do like the `float: right;` though

sigmavirus24:
There shouldn't be a prefix of `v`. [Version strings in Python](https://www.python.org/dev/peps/pep-0440/#version-scheme) don't have it and it may end up being misleading to others (regardless of its aesthetic appeal).

controversial:
@sigmavirus24 Thanks, fixed.

controversial:
@sigmavirus24 @dstufft Does everything look ok here?

dstufft:
@nlhkabu will have to review/merge this.

controversial:
Ok, sounds good.

nlhkabu:
Thanks for the PR @controversial.

Whilst I agree that the version on the right probably *looks* better, from a usability perspective, it makes more sense to have it directly after the package name. This will allow our users to read the content more easily, without having the scan the entire width of the card.

I would also like to keep the version as a `p` rather than a `h3` to indicate that it is less important than the package title.

It looks as though we can get rid of `package-snippet__meta` and turn `package-snippet__version` into a `p`.  We could then apply the font size currently applied to `package-snippet__meta` to `package-snippet__version`.

From a CSS perspective, you can probably avoid floats by using `display: inline-block` on both the title and version.  You might also want to put a small left margin on the version to add a little spacing between the two.

controversial:
Ok. I'll do this when I get a chance. Maybe tomorrow. Thanks!

JosephBywater:
@controversial It's been a while ;)

sahildua2305:
@controversial Are you still working on this? If not, I can take over and fix the issue 😄

controversial:
Sorry, I'm not, you can do it.

sahildua2305:
Thanks for the quick reply, @controversial 👍

PR #1352, opened 2016-07-03
Only show links bar if links are present for the project (Closes #1347)

### LINKED ISSUE #1347: Do not show links bar if no links are present on the project
We want to avoid this situation (see second grey - empty - div):

screenshot from 2016-07-02 21-32-31

This should be as simple as wrapping the div in an if statement checking that links exist.

controversial:
Fixes #1347 

This can be tested on the `pyrfr` repo on the front page.

### DIFF
--- a/warehouse/filters.py
+++ b/warehouse/filters.py
@@ -143,4 +143,4 @@ def format_classifiers(classifiers):
 def contains_valid_uris(items):
     """Returns boolean representing whether the input list contains any valid
     URIs"""
-    return any([is_valid_uri(i) for i in items])
+    return any(is_valid_uri(i) for i in items)
--- a/warehouse/filters.py
+++ b/warehouse/filters.py
@@ -29,6 +29,7 @@ from pyramid.threadlocal import get_current_request
 
 from warehouse.utils.http import is_valid_uri
 
+
 def _camo_url(camo_url, camo_key, url):
     camo_key = camo_key.encode("utf8")
     url = url.encode("utf8")
--- a/warehouse/filters.py
+++ b/warehouse/filters.py
@@ -142,5 +142,6 @@ def format_classifiers(classifiers):
 
 def contains_valid_uris(items):
     """Returns boolean representing whether the input list contains any valid
-    URIs"""
+    URIs
+    """
     return any(is_valid_uri(i) for i in items)
--- a/warehouse/filters.py
+++ b/warehouse/filters.py
@@ -140,5 +140,5 @@ def format_classifiers(classifiers):
 
 
 def filter_valid_uris(items):
-    """ Filter out a list so that it only contains valid urls """
+    """Filter out a list so that it only contains valid URLs."""
     return [i for i in items if is_valid_uri(i)]
--- a/warehouse/filters.py
+++ b/warehouse/filters.py
@@ -27,6 +27,7 @@ import readme_renderer.txt
 
 from pyramid.threadlocal import get_current_request
 
+from warehouse.utils.http import is_valid_uri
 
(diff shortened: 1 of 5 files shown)


warehouse/filters.py
```diff
@@ -136,3 +137,8 @@ def format_classifiers(classifiers):
     structured = collections.OrderedDict(sorted(structured.items()))
 
     return structured
+
+
+def filter_valid_uris(items):
+    """ Filter out a list so that it only contains valid urls """
```
sigmavirus24: I don't think docstrings tend to start or end with whitespace. They should end with a period.

controversial: Ok, sure. I always start / end mine with a single space because I think it makes the code appear less cluttered (and both are PEP8 acceptable), but if this team omits it that's fine by me.

warehouse/templates/packaging/detail.html
```diff
@@ -90,6 +90,7 @@ <h1 class="package-header__name">{{ release.project.name }} {{ release.version }
   </div>
 </section>
 
+{% if release.urls.values()|filter_valid_uris|length > 0 %}
```
sigmavirus24: It seems like you want to print this if there are *any* valid URLs, right? So why not instead write the filter to return a boolean, e.g.,

```py
def any_valid_uris(uris):
     return any(is_valid_uri(uri) for uri in uris)
```

And use it like:

```jinja2
{% if release.urls.values() |any_valid_urls %}
...
{% endif %}
```

warehouse/filters.py
sigmavirus24: You should read PEP 257 if you're trying to follow PEP-based style-guides.

warehouse/templates/packaging/detail.html
controversial: Ok, I guess. I did it this way for robustness, but your way is probably better for the short term.

warehouse/filters.py
controversial: Hadn't seen that, thanks! (Now I have to update https://github.com/controversial/livejson/ so that I can still brag about PEP compliance...)

warehouse/templates/packaging/detail.html
sigmavirus24: Could you explain why your way is more robust?

controversial:
My whole docker container seems to have decided to rebuild everything, so I may not be able to make the other change you suggested for about 45 minutes (which is what it took last time) :(

controversial:
@sigmavirus24 I just pushed a commit to fix the style thing, but how do I fix the other failure? I don't see from where this arises, it looks like the tests fail to recognize my new filter. It works correctly on my machine, though. Any advice?

warehouse/filters.py
```diff
@@ -136,3 +138,9 @@ def format_classifiers(classifiers):
     structured = collections.OrderedDict(sorted(structured.items()))
 
     return structured
+
+
+def contains_valid_uris(items):
+    """Returns boolean representing whether the input list contains any valid
+    URIs"""
+    return any([is_valid_uri(i) for i in items])
```
sigmavirus24: You don't need a list comprehension here.  `any` accepts a generator.

controversial: Ok. Will do. What's causing the test failure, though?

controversial:
Fixed that too @sigmavirus24. Any insight on the failure now?

controversial:
@sigmavirus24 All checks all pass. Can this be merged?

apollo13:
LGTM

controversial:
Thanks!

Issue #1353, opened 2016-07-03
Update margin on `project-header__pip-instructions`


nlhkabu:
Currently: `10px 0 15px` 

Update to: `10px 0 20px`

PR #1357, opened 2016-07-03
Change margin for issue #1353

TheJJ100100:
Change margin from 15 to 20. @nlhkabu


nlhkabu:
Thanks for the PR @TheJJ100100.  I'd prefer not to use the spacing unit in this case - in all other instances we use either the full value or half value - we never divide it by other units.

Are you ok to modify this PR to simply use the pixel value?

TheJJ100100:
@nlhkabu Like that?

nlhkabu:
Thanks @TheJJ100100 

:sparkles:

PR #1359, opened 2016-07-03
Reorder trove classifiers on search sidebar

berkerpeksag:
Perhaps moving order to database would be a better solution in long term, this would do the trick for now.

I have also fixed https://github.com/pypa/warehouse/issues/1298#issuecomment-229862426 in a separate commit.

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -44,6 +44,17 @@ SEARCH_BOOSTS = {
     "keywords": 5,
     "summary": 5,
 }
+SEARCH_FILTER_ORDER = (
+    "Programming Language",
+    "License",
+    "Framework",
+    "Topic",
+    "Intended Audience",
+    "Environment",
+    "Operating System",
+    "Natural Language",
+    "Development Status",
+)
 
 
 @view_config(context=HTTPException)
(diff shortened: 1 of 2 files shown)


warehouse/views.py
```diff
@@ -211,11 +222,17 @@ def search(request):
         first, *_ = cls.classifier.split(' :: ')
         available_filters[first].append(cls.classifier)
 
+    def filter_key(item):
+        try:
+            return SEARCH_FILTER_ORDER.index(item[0])
+        except ValueError:
+            return item[0]
```
dstufft: This is wrong I believe? If passed a classifier that isn't in ``SEARCH_FILTER_ORDER`` it's going to raise an exception because this function is going to return an int if it's in ``SEARCH_FILTER_ORDER`` and a str if it's not.

Perhaps this function should look more like:

```python
def filter_key(item):
    try:
        return (0, SEARCH_FILTER_ORDER.index(item[0], item[0])
    except ValueError:
        return (1, 0, item[0])
```

That should ensure that even if it gets a classifier it doesn't know that (A) it comes last in the list and (B) sorts everything by the defined order, then by alphabetical.

berkerpeksag: You're right, this would raise a TypeError. I've now fixed it. Thanks!

PR #1362, opened 2016-07-03
Add format_number filter to warehouse.i18n

berkerpeksag:


dstufft:
This should use Babel instead of naively formatting with a comma. You'd want to write a new filter like the ones you see in [warehouse/i18n/filters.py](https://github.com/pypa/warehouse/blob/6db7a5cbdc8b00d783887225e578dda3e3d6b8bf/warehouse/i18n/filters.py) which would just be a thin wrapper around [babel.numbers.format_number](http://babel.pocoo.org/en/stable/api/numbers.html#babel.numbers.format_number).

berkerpeksag:
Thanks for the review. I have now addressed your comment.

PR #1363, opened 2016-07-03
Hide "pip install <package>" if there is nothing to download

berkerpeksag:


dstufft:
This seems like a good idea to me, @nlhkabu ?

nlhkabu:
:tada:

PR #1381, opened 2016-07-06
Fix link to checksum explanation. Closes #1318

### LINKED ISSUE #1318: Link to checksum docs on downloads table
screenshot from 2016-06-30 19-20-43

The original idea of this design is to have a popover that contains help information appear when the user clicks on the help icon.  This is not going to be possible for launch, so we should instead find an appropriate place to link to.  Ideally, a page that explains what a checksum is and how to use it.

timofurrer:
I've set the link to the pip documentation as @berkerpeksag suggested.


nlhkabu:
Thanks @timofurrer

PR #1383, opened 2016-07-08
convert project detail view js to es6 without jquery

wismer:
I'd say this is done but I made some slight changes that I would need to test to be sure it didn't break anything. The new Docker update has borked my setup, forcing me to reinstall / rebuild.

Please let me know if there is anything that doesn't seem right or was not addressed. I'm also not quite sure what the `hashchange` event is supposed to do.

### DIFF
--- a/warehouse/static/js/main.js
+++ b/warehouse/static/js/main.js
@@ -2,4 +2,3 @@
 //       ported to using the ES6 syntax and is part of the warehouse package
 //       this will be removed. New code should go in
 //       warehouse/static/js/warehouse and not here.
-


dstufft:
This has merge conflicts.

warehouse/static/js/warehouse/index.js
```diff
@@ -28,6 +28,15 @@ import * as formUtils from "warehouse/utils/forms";
 import Clipboard from "clipboard";
 import PositionWarning from "warehouse/utils/position-warning";
 import Statuspage from "warehouse/utils/statuspage";
+import projectTabs from "warehouse/utils/project-tabs";
+
+// project detail tabs
+docReady(()=> {
+  var headingBtn = document.querySelector(".-js-vertical-tab-content");
+  var mobileBtn = document.querySelector(".-js-vertical-tab-mobile-heading");
+  var styleProps = getComputedStyle(mobileBtn, null);
```
dstufft: Should these ``var`` be ``let``?

berkerpeksag: or even `const`

wismer: mm, I think `let` is more for wrangling with scope but `const` would make more sense since they are not being reassigned anywhere in this scope.

dstufft:
Heh, merge conflicts again :(

wismer:
Something doesn't seem right - I rebased this off of master but there's nothing different?

wismer:
I think this was already merged in?

wismer:
actually, I think you can close this one. The code was included in another PR.

PR #1384, opened 2016-07-08
es6 conversion with filter panel in search without jquery

wismer:
as with #1383, some feedback would be great. I'm hoping it keeps the same intended functionality. From what I've been able to test, it appears that way but I am uncertain.

### DIFF
--- a/warehouse/static/js/main.js
+++ b/warehouse/static/js/main.js
@@ -13,41 +13,6 @@ $(document).ready(function() {
     $(this).closest(".accordion").toggleClass("accordion--closed");
   });
 
-  function setTab(tab) {
-    if (tab) {
-      $(".-js-vertical-tab-content").hide();
-      tab.show();
-      $(".vertical-tabs__tab--is-active").removeClass("vertical-tabs__tab--is-active");
-      $("a[href^='#"+tab[0].id+"']").addClass("vertical-tabs__tab--is-active");
-    }
-  }
-
-  function getTab(selector) {
-    var tab = $(".-js-vertical-tab-content" + selector);
-    return (selector && tab.length) ? tab : null;
-  }
-
-  window.onhashchange = function() {
-    setTab(getTab(location.hash) || getTab(":first"));
-  };
-
-  // Set the tab if the hash is valid, otherwise show the first tab
-  setTab(getTab(location.hash) || getTab(":first"));
-
-  // If in tab mode
-  $(".-js-vertical-tab").click(function(event) {
-    event.preventDefault();
-    history.pushState(null, "", $(this).attr("href"));
-    setTab(getTab(location.hash));
-  });
-
-  // If in accordion mode
-  $(".-js-vertical-tab-mobile-heading").click(function(event) {
-    event.preventDefault();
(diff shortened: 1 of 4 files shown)


warehouse/static/js/warehouse/utils/search-filter-toggle.js
```diff
@@ -0,0 +1,29 @@
...
+      el.style.display = display;
+    }
+  };
+
+  const toggleEvent = (element, display) => {
+    element.addEventListener("click", togglePanelDisplay.bind(null, display), false);
+  };
+
+  const toggleAccordion = (el) => {
+    el.className = el.className === "accordion" ? "accordion accordion--closed" : "accordion";
```
berkerpeksag: Can this be replaced with `el.classList.toggle("accordion--closed");`?

warehouse/static/js/warehouse/utils/search-filter-toggle.js
```diff
@@ -0,0 +1,29 @@
+export default () => {
+  const togglePanelDisplay = (display, event) => {
+    event.preventDefault();
+    var elements = document.querySelectorAll(".-js-dark-overlay, .-js-filter-panel");
+    for (var el of elements) {
+      el.style.display = display;
```
berkerpeksag: I understand that this is basically what the jQuery version does, but perhaps we could defer `el.style.display` part to a new CSS class (something like `.filter-panel--closed { display: none; }`)?

warehouse/static/js/warehouse/utils/search-filter-toggle.js
wismer: @berkerpeksag oh good call. much simpler.

warehouse/static/js/warehouse/utils/search-filter-toggle.js
wismer: @berkerpeksag That would be easier. Should I run the design/css lead or just put it in anyway?

berkerpeksag: I'd say wait for Nicole or Donald's approval before doing more work on this :)

wismer: oops, I had omitted a few words that really changed the meaning. I meant "run it by the lead"

warehouse/static/js/main.js
```diff
@@ -49,25 +43,13 @@ $(document).ready(function() {
...
-    $(".-js-filter-panel").hide();
-  });
 
   $.timeago.settings.cutoff = 7 * 24 * 60 * 60 * 1000;  // One week
 
-  // document.l10n.ready.then(function() {
-  //   // Format all of the time.relative tags to display relative time.
-  //   $(".-js-relative-time").timeago();
-  // });
+  document.l10n.ready.then(function() {
```
dstufft: This should probably just be deleted, but it certainly shouldn't be uncommented.

warehouse/static/js/warehouse/utils/search-filter-toggle.js
dstufft: @nlhkabu would have to answer this question.

PR #1387, opened 2016-07-08
es6 native replacement for timeago jquery plugin

wismer:
Definitely Not Quite Ready For Prime Time - getting an error or two 😢

### DIFF
--- a/warehouse/static/js/main.js
+++ b/warehouse/static/js/main.js
@@ -61,13 +61,4 @@ $(document).ready(function() {
     $(".-js-dark-overlay").hide();
     $(".-js-filter-panel").hide();
   });
-
-  $.timeago.settings.cutoff = 7 * 24 * 60 * 60 * 1000;  // One week
-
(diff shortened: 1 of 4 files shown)


warehouse/static/js/main.js
```diff
@@ -61,13 +61,4 @@ $(document).ready(function() {
     $(".-js-dark-overlay").hide();
     $(".-js-filter-panel").hide();
   });
-
-  $.timeago.settings.cutoff = 7 * 24 * 60 * 60 * 1000;  // One week
```
berkerpeksag: Could you also remove `import "timeago";` (line 7) and update `package.json`?

warehouse/static/js/warehouse/utils/timeago.js
```diff
@@ -0,0 +1,39 @@
+const enumerateTime = (timestampString) => {
+  var now = new Date(),
```
berkerpeksag: These variables could use `let`.

Issue #1305, opened 2016-06-30
Remove legal page, link to https://www.python.org/about/legal/


nlhkabu:
Instead of linking to a separate page, the legal link in the footer can link to https://www.python.org/about/legal/

dstufft:
Let me poke Van about this real quick to make sure that still makes sense to do so here.

nlhkabu:
Note that as per the HTML styleguide, any link should not contain a protocol, e.g. `//mylink.com` not `http://mylink.com`

PR #1390, opened 2016-07-13
Update legal link in footer to python.org legal page

brianfoley81:
As mentioned in #1305, remove the legal page within the app and link to the legal page on python.org.  I'm not sure, based on the comments in the issue, if this change needs to be done, but I figured I would give it a shot just in case it does!


dstufft:
Hey!

Thanks for this PR. I'm currently waiting on @VanL to tell me what exactly we're supposed to do for the legal content, whether it should be a link to the existing page or new content or not.

dstufft:
On second thought, I'm going to go ahead and merge this. If @VanL wants us to do something different bringing it back is trivial.

PR #1392, opened 2016-07-14
bump around gravatar sizes to make things look a little nicer

ewdurbin:
- Bump up size on main profile
- Double the requested size from gravatar so that things don't look so grainy

### DIFF
--- a/warehouse/templates/accounts/profile.html
+++ b/warehouse/templates/accounts/profile.html
@@ -7,7 +7,7 @@
   <div class="left-layout">
     <div class="left-layout__sidebar">
       <div class="author-profile">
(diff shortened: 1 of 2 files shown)


ewdurbin:
maintainer view:

  - before: 
screen shot 2016-07-14 at 10 07 21 am
  - after: 
screen shot 2016-07-14 at 10 07 38 am

profile view:

  - before:
screen shot 2016-07-14 at 10 07 44 am
  - after: 
screen shot 2016-07-14 at 10 08 02 am

dstufft:
This will be something for @nlhkabu

nlhkabu:
thanks @ewdurbin :tada:

Issue #827, opened 2015-11-23
Project Summary cards have variable/inconsistent height


dstufft:
Someone on IRC mentioned that the project summary cards (like on the index page) show up in variable heights that don't really match each other based on how long the content inside of them is. It'd be nice to make them all the same height or, failing that, make them some sort of multiple of each other like servers in a rack (1U, 2U, 4U, etc) so that they line up better.

nlhkabu:
Servers in a rack?  Sorry, I don't understand what you mean... can you draw me an example? :smile:

miohtama:
Please see below. Rows are unaligned, making it look unprofessional.

The solution is to use fixed CSS height and then cut out description, so that it doesn't overflow. CSS does not support ellipsis, so often equal height is achieved using a fade effect at the bottom of the box or just trimming the text to certain width and making an assumption it doesn't exceed the given space.

<img width="1198" alt="screen shot 2015-11-23 at 12 38 07" src="https://cloud.githubusercontent.com/assets/49922/11349071/19d63fc8-91df-11e5-9a78-5abad3e6beb6.png">

nlhkabu:
> often equal height is achieved using a fade effect at the bottom of the box or just trimming the text to certain width and making an assumption it doesn't exceed the given space

I'm not keen on this idea - I think it dramatically reduces the usability of the page.

Instead, let's try addressing this issue using a truncate plugin that adds a 'more' link so that users *can* expand the content if they want to.  The link should be styled to be consistent with the 'more' links on the package history panel:

![show-more](https://cloud.githubusercontent.com/assets/3323703/11359301/c5ce931e-9272-11e5-9d9a-279c619c5b3e.png)

This plugin might be a candidate: https://github.com/jedfoster/Readmore.js

Note that this should only be applied on pages where we have multiple columns of packages (just the index page for now) as this really isn't a problem when we have just one column on display (as on the author page).  In addition, the plugin should be disabled and the full text rendered when the viewport is < 800px, as this is when this layout becomes single column.

dstufft:
That seems like a reasonable solution.

Ivoz:
This actually looks like the perfect use case for a CSS table. http://caniuse.com/#feat=css-table

controversial:
@Ivoz: [Never use tables for layout](http://stackoverflow.com/questions/83073/why-not-use-tables-for-layout-in-html), a lot of people feel really strongly about that 😛 

[...]

PR #1394, opened 2016-07-15
truncate long package description text

nkanaev:
discussed here: #827 

i think using a single line description everywhere makes card heights more consistent and visually pleasing.
also other repositories (like [rubygems](https://rubygems.org/search?utf8=%E2%9C%93&query=test) or [metacpan](https://metacpan.org/search?q=test)) are okay with ellipses. and this solution plays nicely with responsive design.

result:
image


dstufft:
@nlhkabu will need to take a look at this, but I think using ellipses here is a great idea.

nlhkabu:
thanks @nkanaev :)

PR #1407, opened 2016-07-24
Remove unused classifiers from filter list

### LINKED ISSUE #1386: Remove unused classifiers from filter list
We currently show all trove classifiers in the search filter panel, despite the fact that some are not applied to any projects in the DB.

It would be better to only show those classifiers that are actually applied to projects, so we avoid filtering by a classifier and returning an empty result.

berkerpeksag:
Fixes #1386

### DIFF
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -293,6 +293,12 @@ class TestSearch:
         classifier2 = ClassifierFactory.create(classifier="foo :: baz")
         classifier3 = ClassifierFactory.create(classifier="fiz :: buz")
 
+        project = ProjectFactory.create()
+        release1 = ReleaseFactory.create(project=project)
+        release1.created = datetime.date(2011, 1, 1)
+        release1._classifiers.append(classifier1)
+        release1._classifiers.append(classifier2)
+
         page_obj = pretend.stub(page_count=(page or 1) + 10)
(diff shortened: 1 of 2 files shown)


dstufft:
Other than ``make lint`` failing, this looks good to me.

tests/unit/test_views.py
```diff
@@ -293,6 +293,13 @@ def test_with_classifiers(self, monkeypatch, db_request, page):
         classifier2 = ClassifierFactory.create(classifier="foo :: baz")
         classifier3 = ClassifierFactory.create(classifier="fiz :: buz")
 
+        project = ProjectFactory.create()
+        release1 = ReleaseFactory.create(project=project)
+        release1.created = datetime.date(2011, 1, 1)
+        release1._classifiers.append(classifier1)
+        release1._classifiers.append(classifier2)
+        release1._classifiers.append(classifier3)
```
berkerpeksag: Should I left `classifier3` alone and add an assert to test that it's not in `search(db_request)['available_filters']`? Right now, we are not actually testing the new behavior.

dstufft: Yea that sounds good.

berkerpeksag:
Fixed the style issues and tweaked the test case. Thanks.

PR #1418, opened 2016-08-05
WIP: Add support for legacy Register API

sigmavirus24:
Copy relevant bits out of form_upload.

See https://github.com/pypa/twine/issues/200 for a bit more detail

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -986,10 +986,131 @@ def _legacy_purge(status, *args, **kwargs):
     require_methods=["POST"],
 )
 def submit(request):
-    return _exc_with_message(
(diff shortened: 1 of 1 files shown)


dstufft:
It'd be great if we could deduplicate the ``submit()`` view with the ``file_upload()`` view, since the two are sharing a lot of code at the moment.

sigmavirus24:
> It'd be great if we could deduplicate the submit() view with the file_upload() view, since the two are sharing a lot of code at the moment.

Yep. I just want to find out what we need to get it working first. Then I'll deduplicate. =)

dstufft:
Closing this, we're not going to be implementing this API in Warehouse.

PR #1420, opened 2016-08-05
Replace pypi.io with pypi.org in a few additional locations

dstufft:


di:
Also needs updating on this line: https://github.com/pypa/warehouse/blob/master/vcl/main.vcl#L98

dstufft:
Ah, good catch!

PR #1424, opened 2016-08-13
Expose postgresql port.

Carreau:
+ add info on docker.

With this I was able to get a local version of Pypi-legacy to display a page !

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,7 @@
 db:
   image: postgres:9.5
+  ports:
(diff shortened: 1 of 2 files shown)


berkerpeksag:
Note that `docker-compose up` won't start if you already have PostgreSQL installed and looks for the same port number. Since the port number was mainly exposed for pypi-legacy, is there a chance to change it to something else?

dstufft:
Ah, yea we should just do that then :)

PR #1429, opened 2016-08-16
OpenSearch support

### LINKED ISSUE #1399: Add support for opensearch
http://www.opensearch.org/

Lets you (at least from Chrome) type `pypi.io`, hit tab, and then let you type what you're searching for from the address bar instead of having to go to the site first to use the search box.

nkanaev:
resolves #1399

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -101,6 +101,23 @@ def robotstxt(request):
 
 
 @view_config(
+    route_name="opensearch.xml",
+    renderer="opensearch.xml",
+    decorator=[
+        cache_control(1 * 24 * 60 * 60),         # 1 day
+        origin_cache(
+            1 * 24 * 60 * 60,                    # 1 day
+            stale_while_revalidate=6 * 60 * 60,  # 6 hours
+            stale_if_error=1 * 24 * 60 * 60,     # 1 day
+        )
+    ]
+)
+def opensearchxml(request):
+    request.response.content_type = "text/xml"
+    return {}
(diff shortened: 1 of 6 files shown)


warehouse/views.py
```diff
@@ -101,6 +101,23 @@ def robotstxt(request):
 
 
 @view_config(
+    route_name="opensearch.xml",
+    renderer="opensearch.xml",
+    decorator=[
+        cache_control(1 * 24 * 60 * 60),         # 1 day
```
sigmavirus24: Instead of calculating these twice, why not have:

```
ONE_DAY = 1 * 24 * 60 * 60
SIX_HOURS = 6 * 60 * 60

@view_config(
    route_name="opensearch.xml",
    renderer="opensearch.xml",
    decorator=[
        cache_control(ONE_DAY),
        origin_cache(ONE_DAY, stale_while_revalidate=SIX_HOURS, stale_if_error=ONE_DAY),
    ]
)
```

di:
Thanks @nkanaev!

I've verified that this works in the following browsers:
- Chrome Version 52.0.2743.116 (64-bit)
- Chrome Version 54.0.2830.0 canary (64-bit)
- Firefox 48.0

warehouse/views.py
dstufft: This is pretty standard for how the rest of Warehouse does it. I don't have a strong opinion one way or another in general, but for this specific PR I think that consistency is better.

PR #1436, opened 2016-08-23
Update [pytest] section to [tool:pytest]

di:
This fixes the following warning from the test suite:

```
=============================== pytest-warning summary ===============================
WC1 None [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.
```


di:
Looks like we can't do this yet, see https://github.com/pytest-dev/pytest/issues/1831

PR #1438, opened 2016-08-23
Add intermediate containers

### LINKED ISSUE #1221: Prevent rebuilding of all containers when changing package.json
I added a dep in package.json and wanted to install the new package. Sadly, `make serve` causes the container(s) to install basically from scratch (from `apt-get update` on). Is there a better way to install additional packages or add new frontend js dependencies?

di:
Changes to the working directory can force a rebuild of the containers after the `COPY . /app/` step. At the moment, this includes downloading a fair amount of dependencies, none of which need to happen after copying the Warehouse source into the container.

This PR moves the installation of these dependencies into a series of intermediate containers before finally performing the `COPY . /app/` step.

Fixes #1221.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,38 +23,42 @@ RUN set -x \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-# Copy the directory into the Dockerfile
-COPY . /app/
-
-# Set our work directory to our app directory
-WORKDIR /app/
-
-# Uncomment the below line and add the appropriate private index for the
-# pypi-theme package.
-# ENV PIP_EXTRA_INDEX_URL ...
-
-# Install Warehouse
 RUN set -x \
     && apt-get update \
     && apt-get install inotify-tools wget bzip2 gcc g++ make libpq-dev libjpeg-dev libffi-dev --no-install-recommends -y \
-    && rm -rf node_modules \
     && wget https://saucelabs.com/downloads/sc-4.3.14-linux.tar.gz -O /tmp/sc.tar.gz \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -23,38 +23,49 @@ RUN set -x \
...
+
+COPY package.json /tmp/package.json
+
+# Install NPM dependencies
+RUN set -x \
     && npm install -g phantomjs-prebuilt gulp-cli \
+    && cd /tmp \
     && npm install \
+    && mkdir /app \
+    && cp -a /tmp/node_modules /app/
```
dstufft: Huh, does this actually work?

Dockerfile
```diff
@@ -23,38 +23,49 @@ RUN set -x \
...
     # Uncomment the below line if you're working on the PyPI theme, this is a
     # private repository due to the fact that other people's IP is contained
     # in it.
     # && pip install -c requirements/main.txt -r requirements/theme.txt \
     && find /usr/local -type f -name '*.pyc' -name '*.pyo' -delete \
-    && rm -rf ~/.cache/ \
+    && rm -rf ~/.cache/
+
+# Clean up build dependencies
+RUN set -x \
```
dstufft: If we're not going to do this inline with when we install them, then there's not a lot of point to ever uninstalling them. It's just a trick to keep the image size smaller, but that won't happen if you don't install, use, uninstall all in the same ``RUN`` command.

di: Good point. I'm forgetting now why I thought it should be a separate command... 🙂

Dockerfile
di: Indeed!

PR #1441, opened 2016-08-24
add support for opengraph tags to warehouse

### LINKED ISSUE #677: Handle pinning on Pinterest
See https://bitbucket.org/pypa/pypi/issues/336/python-package-pages-pin-poorly-on

> @audreyr wrote:
> For example, when I try to pin Cookiecutter 1.0.0 on Pinterest:
> * it pins as "pypi.python.org" with no mention of the package name or version
> * the image doesn't get picked up.
> * the default text image should be the package name, not PYPI PYTHON ORG
> 3811756009-screen shot 2015-09-09 at 9 37 24 pm
> Feel free to close this out if it's too low-priority. I just figured you might want to know and have it tracked, since a lot of people (esp women) use Pinterest for all their bookmarking.

reaperhulk:
These tags are used for Facebook, Slack, Pinterest, and other sites and provide richer metadata when sharing links.

This may resolve #677.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,10 @@ saucelabs:
 lint: .state/env/pyvenv.cfg
 	$(BINDIR)/flake8 .
 	$(BINDIR)/doc8 --allow-long-titles README.rst CONTRIBUTING.rst docs/ --ignore-path docs/_build/
-	$(BINDIR)/html_lint.py --disable=optional_tag,names,protocol `find ./warehouse/templates -path ./warehouse/templates/legacy -prune -o -name '*.html' -print`
+	# TODO: Figure out a solution to https://github.com/deezer/template-remover/issues/1
+	#       so we can remove extra_whitespace from below.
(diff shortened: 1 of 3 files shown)


dstufft:
This looks good to me, but you'll need to fix the linting issues.

reaperhulk:
PR updated. For some reason it's still failing with `17:1: Error: Encoding: The meta charset should be set to utf-8: Add the tag <meta charset="utf-8">.` even though that tag is present already (line 17 of base.html)

di:
@reaperhulk Unfortunately the linter is not very sophisticated and just looks at each `.html` file individually, it doesn't know that the tag has already been added in `base.html`, thus it fails when it sees `meta` tags in `detail.html`

You could fix it by adding another meta tag with the charset in `detail.html`, but I think it would be better to add a macro in `base.html` for generating the `opengraph_tags` block from a title and description, and then call that from `detail.html` instead.

dstufft:
Huh, maybe we should just disable the ``<meta charset>`` rule? Something that only has to exist in one file probably isn't something we need to really lint for?

di:
@dstufft That's a good idea.

@reaperhulk You can disable that specific check by adding `charset` to the `--disable` flag here: https://github.com/pypa/warehouse/blob/master/Makefile#L118

PR #1443, opened 2016-08-25
Add data-requires-python attribute

### LINKED ISSUE #1403: Implement data-requires-python on Simple
As per python/peps#56 we need to add the ``data-requires-python`` field to our simple API.

di:
Fixes #1403.


takluyver:
Thanks! I can see this working on https://pypi.org/simple/flit/ :-)

Issue #1389, opened 2016-07-13
Avatars via Gravatar: inform users


brainwane:
A friend of mine who maintains some Python packages took a look at pypi.io and mentioned:

> It was surprising to me to see my Gravatar image used on my profile—the old site doesn't even have landing pages for user profiles, as far as I can tell?
> It might be good to let users opt into having their Gravatar image show up on their profile page, or at least tell the user that Gravatar is the source of the image when they log in and look at their profile page (I had to do a little digging to figure it out).

So I think this is two related tasks:
- [ ] Find some way to inform maintainers that Gravatar is where this image is being sourced from (mouseover/hover text on the image? image caption? a message on the user profile/login page? I'm not sure how to do it)
- [ ] Provide a way for users to opt in or out of this (my preference is for this to be opt-**in** since showing photos of people without asking them first can feel intrusive and can be a factor that increases harassment by jerks)

I'd predict that, if we don't have these set up at launch, at least a dozen people are going to unhappily say "hey I don't want my picture there" or "where's that coming from," so it's worth it to just implement this ahead of launch to save all of us time and frowns. :)

Potentially related to #815.

boxed:
Seems very strange to opt out of a service you've signed up for whose entire purpose is exactly the behavior they got. 

I absolutely agree that it would be nice if it was obvious that the image comes from gravatar. Maybe just a title attribute?

ssbarnea:
The entire purpose of Gravatar was to put an avatar (not necessarily your real face) near a username. I do not see any reason for adding this feature because if someone doesn't want to use a photo they have at least two options: to not use gravatar or to user an alternative email on pypi. Adding a parameter to your real email is enough to hide gravatar (john.doe+pypi@gmail.com will not render the gravatar unless is registered explicitly on gravatar).

For the sake of reaching M1 and avoiding feature-creeping, I would vote to close this as wontfix or to move it to another milestone.

brainwane:
How about just adding some hover text or something on the image so it is a little more discoverable that it comes from Gravatar, to make it more obvious where it's coming from, so people who find it an unpleasant surprise don't waste time hunting around their PyPI settings fruitlessly?

dstufft:
I'm going to kick this out of launch, I don't think it is really needed as a high priority item and can be added later.

nlhkabu:
Closing this as it was implemented in https://github.com/pypa/warehouse/pull/1444

PR #1444, opened 2016-08-25
Inform Gravatar users that their images are from Gravatar

di:
This PR should resolve the concerns mentioned in #1389 by:
1. Updating the `alt` attribute to mention Gravatar;
2. Adding a `title` attribute which mentions Gravatar;
3. Add a link below the user's image (when logged in) linking to their Gravatar profile, which will allow them to change (or remove) the image.

Screenshots:
screen shot 2016-08-25 at 5 42 23 pm
screen shot 2016-08-25 at 5 44 11 pm


warehouse/accounts/views.py
```diff
@@ -32,6 +32,7 @@
 @view_config(
     route_name="accounts.profile",
     renderer="accounts/profile.html",
+    uses_session=True,
```
dstufft: Hmm, I'd prefer not to set ``uses_session`` on the profile, that will make it hard to cache the profile view (effectively impossible for logged in users). Let me think on this a few.

di: I suppose we could do something similar to the `current-user-indicator`...

di:
@dstufft Since we're not giving users the ability to log in at launch, how about I drop https://github.com/pypa/warehouse/pull/1444/commits/6643af60b89f889fc36b1dd73bd156ac3a7af405 from this PR and add an issue to deal w/ the user's profile page being able to tell when the user is logged in?

PR #1447, opened 2016-08-29
Make it possible to copy&paste from readme

szepeviktor:
On my Windows workstation is it possible to triple-click to select the whole line.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,18 +16,18 @@ commands you can use:
 
 .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


di:
This README section is somewhat lacking, I think it would be better to just point towards https://warehouse.readthedocs.io/development/getting-started/ instead of duplicating content here.

@dstufft thoughts?

dstufft:
Not duplicating seems like a reasonable goal to me.

PR #1457, opened 2016-09-22
Remove note related to Docker Machine

jayfk:
Removed a note about using `docker-machine env` to determine the ip address. That's no longer possible using Docker for Mac/Windows.

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -136,12 +136,6 @@ Viewing Warehouse in a browser
 
 Web container is listening on port 80. It's accessible at ``http://localhost:80/``.
 
-.. note::
-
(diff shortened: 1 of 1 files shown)


di:
Hi @jayfk, thanks for your PR.

Docker machine still exists, because the latest version of Docker does not support some older OSX and Windows architectures. Thus it's still possible that some Warehouse developers might still be using docker-machine (or just haven't migrated yet).

I think it'd be better to modify the note to say something like "If you use `docker-machine`, the warehouse application..."

Also, the build is failing due to #1458 -- sorry about that.

jayfk:
> I think it'd be better to modify the note to say something like "If you use docker-machine, the warehouse application..."

👍

PR #1458, opened 2016-09-22
Add version to drop_postgresql_database call

di:
Due to https://github.com/ClearcodeHQ/pytest-dbfixtures/pull/192

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ sudo: false
 addons:
   hosts:
     - warehouse.local
-  postgresql: "9.4"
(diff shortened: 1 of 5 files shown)


dstufft:
Hmm, why are we switching to 9.4? Production is using 9.5.

di:
@dstufft Ah. Since the Travis builds run against 9.4, https://github.com/pypa/warehouse/blob/master/.travis.yml#L38

And there's this note in `getting-started.rst`:
> PostgreSQL 9.4 is required because of pgcrypto extension

I thought we were using 9.4, but now I see it's due to Travis not supporting 9.5 yet. (https://github.com/travis-ci/travis-ci/issues/4264)

I'll update this.

PR #1459, opened 2016-09-22
Upgrade Javascript Pipeline

### LINKED ISSUE #1456: Warehouse can be used as a CPU burn test on mac OS
It'd be funny if it wasn't so sad.

Running `make serve` on mac OS with the latest Docker for Mac causes extremely high CPU usage.

screen shot 2016-09-22 at 10 46 31
*Yep, that's celsius. On a freaking hexa core Mac Pro*

From what it looks like, `gulp watch` isn't using any filesystem APIs by default causing the CPU to be under constant load to look for file changes: https://forums.docker.com/t/docker-compose-volumes-gulp-watch-100-cpu-usage/10192

I've played around with the various things that can be passed to the watch function, but couldn't fix it. I had no way to debug it properly as I'm not very familiar with gulp.

Could someone please take a look at it?

dstufft:
Fixes #1456


jayfk:
I can confirm this fixes #1456. Thanks @dstufft.

PR #1464, opened 2016-09-26
Automatically generate a coverage html report

jayfk:
Creates a coverage html report with each test run and mounts it at `./htmlcov`.


di:
Thanks @jayfk!

PR #1471, opened 2016-10-06
Optimize reindex

### LINKED ISSUE #1224: Optimize reindexing
It's possible to tweak the index settings before and after a full reindex speed up reindexing.

Things you can do before reindexing:
* Set `num_replicas` to `0`. During reindexing this will stop the network traffic of Elasticsearch replicating your data across the nodes. This can happen in bulk when you set `num_replicas` back to a value greater than `0` after reindexing is finished.
* Set `refresh_interval` to `-1`. I.e. disable index refreshes completely during bulk indexing.  (See: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html#bulk)

Things you can do after reindexing:
* Perform a `force_merge` on the index. The defaults should be fine. See: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html
* Set the `num_replicas` back to its default value.
* Set the `refresh_interval` back to its default value.

di:
This PR finishes work by @robhudson in #1226 to optimize ES reindexing, as well as prevent the occurrence of the occasional `ReadTimeoutError`.

Closes #1224, closes #1226.

### DIFF
--- a/tests/unit/cli/search/test_reindex.py
+++ b/tests/unit/cli/search/test_reindex.py
@@ -84,6 +84,12 @@ class FakeESIndices:
                 else:
                     raise ValueError("Unknown action: {!r}.".format(action))
 
+    def put_settings(self, index, body):
+        pass
+
+    def forcemerge(self, index):
+        pass
+
 
 class FakeESClient:
 
--- a/tests/unit/cli/search/test_reindex.py
+++ b/tests/unit/cli/search/test_reindex.py
@@ -54,6 +54,9 @@ class FakeESIndices:
         self.indices = {}
         self.aliases = {}
 
+        self.put_settings = pretend.call_recorder(lambda *a, **kw: None)
+        self.forcemerge = pretend.call_recorder(lambda *a, **kw: None)
+
     def create(self, index, body):
         self.indices[index] = body
 
@@ -83,12 +86,6 @@ class FakeESIndices:
                     self.remove_alias(values["alias"], values["index"])
                 else:
                     raise ValueError("Unknown action: {!r}.".format(action))
-
-    def put_settings(self, index, body):
-        pass
-
-    def forcemerge(self, index):
-        pass
 
 
 class FakeESClient:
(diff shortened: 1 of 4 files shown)


tests/unit/cli/search/test_reindex.py
```diff
@@ -84,6 +84,12 @@ def update_aliases(self, body):
                 else:
                     raise ValueError("Unknown action: {!r}.".format(action))
 
+    def put_settings(self, index, body):
+        pass
```
dstufft: This, and the ``forcemerge`` should be testable somehow to ensure they've been used. Might just be adding a settings attribute to ``FakeESIndices`` that records what the current settings are or something like that.

warehouse/cli/search/reindex.py
```diff
@@ -57,6 +57,8 @@ def reindex(config, **kwargs):
     """
     client = config.registry["elasticsearch.client"]
     db = Session(bind=config.registry["sqlalchemy.engine"])
+    number_of_replicas = config.registry.get("elasticsearch.replicas", 0)
+    refresh_interval = config.registry.get("elasticsearch.interval", "5s")
```
dstufft: Is there any reason you've chosen ``5s`` here? I'm not against that number just curious if there's some significance to it or not.

di: Simply because this is what it had been set to in the original PR: https://github.com/pypa/warehouse/pull/1226/files#diff-80840db14f707f8d7c8db4ee49bfd4efR110

It seems like the default is actually "1s": https://www.elastic.co/guide/en/elasticsearch/reference/2.4/index-modules.html#dynamic-index-settings

I don't see any reason why it shouldn't go back to the default version, unless @robhudson has a reason.

robhudson:
Thanks for picking this up and finishing it, @di.

warehouse/cli/search/reindex.py
dstufft: Currently it only gets updated during bulk indexing, however once we kill legacy PyPI then I plan on making it so that bulk indexing is only a rare thing and most updates happen as updates happen to the source data.

That being said, even when we are doing incremental updates as the source data changes, we don't really need super realtime data. If something shows up within a few minutes that should be more than quick enough.

dstufft:
I've approved these changes, it all looks good to me, the only open question I think is the one of what the refresh interval should be. I don't really have a good sense of what it should be so I'll let @di pick if he thinks 1s is good enough for now or if he wants to adjust it.

PR #1473, opened 2016-10-11
Remove creation of new index when reindexing

di:
This PR removes the `new_index.create()` call from the reindexing command. I think the explicit call to create a new index is unnecessary, because [elasticsearch will automatically create an index if it has not been created before](https://www.elastic.co/guide/en/elasticsearch/reference/2.2/docs-index_.html#index-creation). 

Furthermore, it is causing problems when reindexing in production:
- Prior to #1471, this line was producing connection timeout errors nearly every time it was called; 
- Since #1471, the call can return before timing out, which produces a `index_already_exists_exception` instead (but still occasionally times out).

The latter issue leads me to believe that this due to a race condition with the automatic index creation -- since our "temporary" index is given a randomly-generated name every time, I don't see any other way that the index could already exist.

I can't reproduce either exception in local development. However, I have verified that reindexing still creates a new index and runs as expected without this line.

### DIFF
--- a/tests/unit/cli/search/test_reindex.py
+++ b/tests/unit/cli/search/test_reindex.py
@@ -56,12 +56,7 @@ class FakeESIndices:
 
         self.put_settings = pretend.call_recorder(lambda *a, **kw: None)
         self.forcemerge = pretend.call_recorder(lambda *a, **kw: None)
-
-    def create(self, index, body):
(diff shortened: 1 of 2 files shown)


di:
@dstufft, are you able to take a look at this? As noticed in #1489, the search index is getting somewhat stale due to this issue.

PR #1479, opened 2016-10-17
Reduce package card height

### LINKED ISSUE #1337: Reduce package card height

k-nut:
This fixes #1337 by moving the version number of the package to the same line as the name. It also reduces the margins as proposed.
I think this change should also be applied to the markup in `warehouse/templates/packaging/detail.html` but I could not find out where that is displayed. Can someone provide me with a link so that I can test locally and if applicable push another change?

screen shot 2016-10-17 at 12 12 52


di:
Hey @k-nut, thanks for taking care of this!

The `project_snippet` macro in `warehouse/templates/packaging/detail.html` was originally used in the "Dependencies" tab, [which is currently commented out](https://github.com/pypa/warehouse/blob/master/warehouse/templates/packaging/detail.html#L203-L229). I think it's fine if you update that markup as well, but you won't see anything change.

This LGTM with or without that change, but @nlhkabu should review as well.

k-nut:
Cool, thank you for the quick feedback! I pushed another change that changes the markup in `detail.html` as well.

k-nut:
May I ask what the status of this is? Is there anything that I can do to help move this forwards?

dstufft:
Thanks! Sorry for the delay. I'm going to go ahead and merge this for now. I'll tag it so that @nlhkabu can review it later to ensure she didn't want any changes done to it.

PR #1480, opened 2016-10-17
Restore margin between maintainer avatars

### LINKED ISSUE #1313: Add spacing between maintainer avatars
This is a regression from the CSS refactor

screenshot from 2016-06-30 07-24-59

k-nut:
This fixes #1313
It also sets the max-height to 40px in the responsive view.
This is to make sure that the images always
stay square and do not become distorted (50px height vs 40px width)

screen shot 2016-10-17 at 21 58 47


di:
This seems to add more horizontal spacing than vertical:

screen shot 2016-10-18 at 9 48 50 am

k-nut:
True! Good catch. I pushed a new version that uses flexbox which seems to work better. Would that be something that you can imagine using?
screen shot 2016-10-18 at 20 39 22

di:
Thanks @k-nut! This looks good to me but @nlhkabu will need to approve.

dstufft:
Thanks! Sorry for the delay here, I've gone ahead and merged this, and tagged it so @nlhkabu can review it later.

nlhkabu:
:+1:  thanks for this PR @k-nut 

Flexbox is indeed a good solution.  I'll let you know if we have any troubles when we cross browser test :)

PR #1513, opened 2016-12-02
Prevent inconsistency between release and release_files tables.

Carreau:
The work in PR #1448 was meant to replicate the information of the
`requires_python` of the `release` table to the `release_files` table
for efficiency when generating the list of available packages for pip.

While the work on #1448 seem sufficient for warehouse itself, it need to
consider that legacy-pypi also access the same database, and legacy-pypi
violate some constraint.

In particular when using `setup.py register` followed by `twine upload`,
the file upload insert files into `release_files` after inserting into
`releases`. Thus the value in releases is not propagated, leading to an
inconsistency and a listing in pip missing information about
python-version compatibility.

While I doubt there are any packages released between the merge of #1448
and a fix, this update the tables, and bind an already existing trigger
to update the information during insertion in `release_files`

---  

See discussion pass [this comment](https://github.com/pypa/pypi-legacy/pull/506#issuecomment-264256260) on the merged pr in PyPI legacy. 

cc @dstufft.

### DIFF
--- a/warehouse/migrations/versions/99291f0fe9c2_update_release_file_on_insert.py
+++ b/warehouse/migrations/versions/99291f0fe9c2_update_release_file_on_insert.py
@@ -1,0 +1,60 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+
+The release.requires_python and release_files.requires_python can be out of
+sync if a file is uploaded after the release has been created. It's
+requires_python field will stay empty.
+
+We need to :
+ - Fix potentially existing mistakes
+ - Create a trigger that ensure consistency
+
+Revision ID: 99291f0fe9c2
+Revises: e7b09b5c089d
+Create Date: 2016-12-02 00:58:53.109880
+"""
+
+from alembic import op
+
+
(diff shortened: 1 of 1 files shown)


dstufft:
Other than the lint failure this looks fine to me.

Carreau:
I can try to write  test that make sure releases files is updated even if
insertion is made after the release conterpart.

I'm in a shaky train right now but I can try to do that tomorrow.

Carreau:
I'm unsure how to write test for that. it seem to require changes in `ReleaseFactory` and other things I'm uncomfortable with and clearly don't understand all the magics of the metaclasses.

I should have fixed the linting error though.

dstufft:
No Problem!

dstufft:
@Carreau The migrations have forked since you created this PR. You'll need to either do the logic equivalent of a "rebase", or do a merge with alembic. The rebase is probably simplest, just update the ``down_revision`` to ``e7b09b5c089d`` I think.

Carreau:
> The migrations have forked since you created this PR. You'll need to either do the logic equivalent of a "rebase", or do a merge with alembic. The rebase is probably simplest, just update the down_revision to e7b09b5c089d I think.

Done with a rebase. Travis seem to be happy.

Carreau:
Thanks !

Carreau:
Is that also automatically deployed to testpypi ?

dstufft:
The code is automatically deployed, but database migrations are not automatically ran. I've gone ahead and ran the migrations now.

Carreau:
> The code is automatically deployed, but database migrations are not automatically ran. I've gone ahead and ran the migrations now.

Thanks I appreciate !

Old version of py3onlytest now have the data-requires-python field, and the new one I just uploaded has is as well ! Pip also installs the right version[*]. 

[*] well I had one version where the requires_python in the db was not matching the one in the package metadata, but I'll put that on me fiddling with the debugger in twine. And that allowed me to check that even then the error message is good. 

Thanks for guiding me through the process ! And for working on Python packaging in general !

PR #1575, opened 2016-12-15
fix invalid syntax

silenius:


dstufft:
I believe this syntax is fine. It's passing tests anyways. This is a Python 3.5 codebase so this should be a mandatory keyword only argument.

silenius:
right :)

PR #1591, opened 2016-12-17
Allow only one sdist to be uploaded per release

dstufft:


jaraco:
This change breaks setuptools (see pypa/setuptools#893).

dstufft:
I'm at the airport, but this change was proposed / accepted as part of PEP 527. If setuptools hardcodes the .zip it could just upload zip files instead of .tar.gz?

jaraco:
Yes, that's what I've done for now. I expect Unix folks will complain that there's no .tar.gz but that's for another day.

PR #1594, opened 2016-12-17
Only limit uploads on sdists not any file

dstufft:


tests/unit/forklift/test_legacy.py
```diff
@@ -1540,6 +1540,113 @@ def storage_service_store(path, file_path, *, meta):
             ),
         ]
 
+    def test_upload_succeeds_with_wheel_after_sdift(self, tmpdir, monkeypatch,
```
reaperhulk: typo here

PR #1623, opened 2016-12-28
Select the latest non-prerelease version for json

### LINKED ISSUE #1599: json urls object may contain prereleases
I think the urls object in the JSON output should reference downloads for the latest stable release; even after #1519, the urls object may contain references to prerelease versions. For example, https://pypi.org/pypi/wheel/json has

```json
{
  "urls": [
    {
      "filename": "wheel-0.30.0a0-py2.py3-none-any.whl",
      "url": "https://files.pythonhosted.org/packages/83/53/e120833aa2350db333df89a40dea3b310dd9dabf6f29eaa18934a597dc79/wheel-0.30.0a0-py2.py3-none-any.whl"
    }, ...
```

I think it should point to 0.29.0 instead, like the project page does.

dstufft:
Fixes #1599

### DIFF
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -66,6 +66,41 @@ class TestJSONProject:
         assert resp is response
         assert json_release.calls == [pretend.call(release, db_request)]
 
+    def test_with_prereleases(self, monkeypatch, db_request):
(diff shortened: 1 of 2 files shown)


tdsmith:
Thanks Donald!

PR #1625, opened 2016-12-29
Python 3.6.0

di:


dstufft:
I've been holding off on this a bit, trying to decide if it made sense to migrate to 3.6.0 or not. While 3.6.0 is the newest version and there's little reason not to upgrade CPython 3.5.2 to CPython 3.6.0, in the upcoming months it's likely that PyPy3 will be released with production level completion for Python 3.5. That means if we stick with Python 3.5.2 we should be able to pretty easily move over to PyPy3 when that is released.

However, I'm not sure if it makes much sense to care a whole lot about using PyPy here since we purposely offload as much as we can to Varnish. Another consideration is that I can't think of a whole lot in Python 3.6 that is really compelling to move, but I'm happy to consider something if I'm forgetting it.

di:
Ok! I just did it now because the Docker image was just released and I didn't really have a good reason _not_ to move to 3.6, but I wasn't aware that there was a plan to move to PyPy.

I'll close this PR for now so it's not hanging around!

dstufft:
Absolutely! My PyPy plan is not fully formed, more like a "It would be nice if...." kind of deal. If there was something really compelling about 3.6 (or 3.7, or 3.8, etc) and PyPy didn't have plans to support that version yet I'd be fine with moving to it too. It's really just a combination of nothing super compelling in 3.6 and PyPy being zoom zoom.

PR #1748, opened 2017-02-15
Add hash for chardet==2.3.0 wheel

di:


di:
Moved this into #1749 instead.

PR #1774, opened 2017-02-26
fixes #1773 fix filter-panel css overrides during toggling

### LINKED ISSUE #824: Fix toggle of filter panel on search results page
How to reproduce this bug:
1. Visit https://warehouse.python.org/search/?q=search
2. Reduce the screen size down to mobile (so the sidebar dissapears)
3. Click 'add filter' button
4. Close panel by clicking on the 'x'
5. Increase the screen size back to desktop

There is a conflict between the JS hiding the panel:
https://github.com/pypa/warehouse/blob/master/warehouse/static/js/main.js#L48-L52

and the CSS:
https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/components/_panel-overlay.scss

We need to refactor this so that instead of using `hide()` we apply a class that can be controlled by CSS media queries.

### LINKED ISSUE #1773: filter-panel disappearing after  clicking on "Add Filter" and resizing window
if you are within the "(max-width: 1000px)" breakpoint, you are shown a "Add Filter" button.  This  button displays the filter panel and overlay.  After clicking on the close button, if you resize the window larger than this breakpoint region, you do not recover filter panel due to overriding elemental css properties.  The fix for this is, instead of giving the filter-panel a display of "none" give it a display of "".  This will erase the elemental style and let the css properties regain control.

sohail288:
Fixes #1773 
Fixes #824


dstufft:
Can you provide more information about what the problem you're fixing is here?

sohail288:
The problem involves window resizing after clicking on the "Add Filter" button which is displayed when the window size is less than 1000px.  After clicking on "Add Filter" and subsequently clicking on the "close" button for the filter panel, if you resize the window to a size larger than 1000px, the filtering options that are normally displayed for large screen sizes do not reappear.  
 
The problem can be reproduced by following these steps:

1) open the site on a browser with a window width of at least 1000px
2) search for something
3) you should see the faceted search options on the left side
4) narrow window to a width less than 1000px
5) click on the now visible "Add Filter" button
6) click on the close button for the filter panel
7) resize the window to a size larger than 1000px
8) notice that the original filter options on the left side are gone.

di:
I believe this is the same issue described in #824.

dstufft:
Ok great. I verified locally and this works to solve the problem. Thanks! Will merge once tests pass.

dstufft:
Doing an override merge because codecov isn't reporting the status for some reason. On the website it is 100% (and this change couldn't possibly change it anyways).

PR #1837, opened 2017-03-12
Display a list of "trending" projects

### LINKED ISSUE #775: Have top projects use the last 30 days or so instead of all time
Right now the "top projects" thing on the index page uses the all time download counts. This is OK but it means that they are unlikely to _ever_ really change. Instead it'd be nice if we used the last N days or something like that.

dstufft:
Instead of using an absolute download count over all time (which has been broken for awhile anyways), we will instead compute which projects were "trending" yesterday (the last full day) compared to their typical download counts over the last 30 days.

To do this, we basically compute a zscore for each project, or basically we figure out what the average downloads per day is for a project, what the stddev for downloads is over the last 30 days, and then we compute a number that tells us how many standard deviations away from average yesterday was in either the positive or negative direction. We then store this information in PostgreSQL as ``Project.zscore`` (it is a floating point number).

To display trending we then order all the projects by that zscore (putting ``NULL`` last, since a new project just released today could not be trending yet!) and finally by a random value to tie break (and also to handle cases like TestPyPI where there will be no trending data).

This isn't perfect, but it should hopefully display some new projects that are actually interesting to people. One possible problem with it is that it will be far easier for a relatively unknown project with few downloads to get a high zscore than it will be for a well known project (e.g. it's easier to go from 5 downloads to 100 downloads than it is to go from 50,000 downloads to 1,000,000). We could maybe weight this somehow by the number of downloads in general a project has to give well known, well downloaded projects a bit of a boost, but I'm not sure how best to do that. There is currently a limit that a project must have at least 5000 total downloads in the past month to be considered as having a zscore at all which should hopefully clear out some of the extreme ends of this.

This still needs cleaned up (in particular I need to figure out how best to handle authentication to BigQuery) as well as getting the computation task scheduled for once a day. It also needs tests written/fixed!

Fixes #775


warehouse/packaging/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+
+        zscores.update(dict(rows))
+
+        if not page_token:
+            break
+
+    # We're going to "reset" all of our zscores to a steady state where they
+    # are all equal to ``None``. The next query will then set any that have a
+    # value back to the expected value.
+    request.db.query(Project).update({Project.zscore: None})
```
alex: This should probably have a `where(Project.zscore.is_not(None))` qualifier on it.

warehouse/packaging/tasks.py
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from warehouse import tasks
+from warehouse.packaging.models import Project
+
+
+@tasks.task(bind=True, ignore_result=True, acks_late=True)
+def compute_trending(task, request):
```
alex: What actually invokes this?

dstufft: Nothing right now, it will be setup to auto run via celery beat.

warehouse/packaging/tasks.py
dstufft: Yea good point, I'll fix that.

PR #1864, opened 2017-03-26
Update user service should hash the password

gsb-eng:
Update user service should always hash the password. This change helps in several other features like reset password, update user profiles etc


gsb-eng:
@dstufft : Could you please check and merge this?

dstufft:
Looks good! Will merge once tests pass again, thanks.

PR #1899, opened 2017-04-04
Autoincrement trove classifier id.

Carreau:
This should allow to not have to bother about getting the id right when
inserting new.

Legacy PyPI does a :

```python
cursor = st.get_cursor()
    cursor.execute("select max(id) from trove_classifiers")
    id = cursor.fetchone()[0]
    if id:
        id = int(id) + 1
    else:
        id = 1
```

Then insert, which IIUC is subject to race conditions.

---- 

That's still imperfect, as for new root-classifiers `l2` is supposed to be equal to id, so you would need to : 
```
INSERT into trove_classifiers (classifier, l2, l3, l4, l5) VALUES ('NEW TROVE', something temporary, 0, 0, 0);
```

Then : 
```
UPDATE trove_classifiers set l2=id where classifier='NEW TROVE';
```

(which you can skip for non root new trove classifiers, so this is already an improvement)

I'm unsure if you can use something like:
```
INSERT into trove_classifiers (classifier, l2, l3, l4, l5) VALUES ('NEW TROVE', __autodertermine this from self.id__ , 0, 0, 0);
```

Which would be great. But anyway that only affect further Pull requests that would add trove classifiers. I don't see any further test to add as inserting without increasing the primary key is already a violation of the uniqueness constraint.


Carreau:
Failure is unrelated (can't found NPM ??) can anyone restart the tests ?

PR #1927, opened 2017-04-18
Auth pages

nkanaev:
* Styled up login/register pages

image

image

* "Sign out" link on current user indicator now works without redirecting user
* Minor ui tweaks (removed outline, since input fields have border color for focus state)

before: 

image

after:

image


dstufft:
Thanks a lot, this looks great to me!

PR #1931, opened 2017-04-19
Automatically run migrations during Heroku's "Release" Phase

### LINKED ISSUE #661: Automatically Run Database Migrations
Right now there's no way to automatically run our database migrations, however it would be great if there were (I think?). At the moment whenever there is a migration to apply I have to manually run it. Given that we're on Heroku and we're using pipelines I'm not sure we have very many options to run our migrations. The only idea I can think of is to run them as part of the application starting up, which has problems with the fact we have multiple Dynos running in production.

So I guess the question is, should we run them automatically and if we should, is there anyplace better to run them than on application startup?

dstufft:
This change ensures that only one alembic command can be run at any one time (and they will queue up behind the currently running command waiting to acquire the lock). This means that two independently run migration commands won't have one of them failing to commit the transaction.

In addition this uses the Heroku [Release Phase](https://devcenter.heroku.com/articles/release-phase) to run the database migrations and deploy the Fastly configuration automatically when a new release is created. That new release won't deploy until after the release command has been run (and if it unsuccessful, the deploy will fail).

This will prevent the Fastly config from slowing down start up as well.

One important caveat here is that obviously with an already running website both the Fastly and the DB migrations will have to not break compatibility with the currently running code unless we're fine with getting errors during that window.

Fixes #661

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
-web bin/start-web python -m twisted web -n -p tcp:port=$PORT --wsgi warehouse.wsgi.application
+release: bin/release
+web: bin/start-web python -m twisted web -n -p tcp:port=$PORT --wsgi warehouse.wsgi.application
 worker: bin/start-worker celery -A warehouse worker -B -S redbeat.RedBeatScheduler -l info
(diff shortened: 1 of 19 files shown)


ewdurbin:
the only suggestion i'd have is documenting the requirement of backwards compatible database migrations (at least on a step by step basis) in the developer docs. similar for fastly configuration.

PR #1980, opened 2017-05-13
Design updates

nlhkabu:
Clean up footer:
screenshot from 2017-05-13 10-21-13

Improve styling of login and registration forms:
screenshot from 2017-05-13 10-21-07
screenshot from 2017-05-13 10-20-56

Also includes minor CSS cleanups


warehouse/templates/base.html
```diff
@@ -140,19 +140,13 @@
...
-          <li><a href="{{ request.route_path('policy.terms-of-use') }}">Terms of Use</a></li>
-        </ul>
       </div>
 
       <hr class="footer__divider">
 
       <div class="footer__text">
         <p>Developed and maintained by the Python community, for the Python community.</p>
-        <p>© {{ now()|format_date('yyyy') }} Python Software Foundation.</p>
+        <p>© {{ now()|format_date('yyyy') }} Python Software Foundation. <a href="{{ request.route_path('policy.terms-of-use') }}">Terms of Use</a></p>
```
dstufft: It might make sense to keep this in it's own footer item, if only because rather than "Legal", it'd be a convenient to have a place to list other kinds of policies like name ownership and what not (like [PEP 541](https://www.python.org/dev/peps/pep-0541/)).

nlhkabu: I'm happy to have a fourth column - but it seems silly for it to only have one item (for now)
Let's re-add it when we add these pages?

dstufft: Seems fair

Issue #2003, opened 2017-05-17
[Feature Request] Support Uploading ARM binary Wheels


kpayson64:
Currently, attempting to upload a ARM binary will return a 400 error.

bennuttall:
Great news! I've attempted to build ARM wheels of every package of PyPI with reasonable success: http://piwheels.bennuttall.com/

PR #2010, opened 2017-05-18
Add linux_armv7l as a supported wheel type

kpayson64:
Addresses #2003

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -54,6 +54,7 @@ _allowed_platforms = {
     "any",
     "win32", "win_amd64", "win_ia64",
     "manylinux1_x86_64", "manylinux1_i686",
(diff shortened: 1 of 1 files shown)


dstufft:
I don't have any way to verify this is the right tag, but it looks like it could be so I'll go ahead and merge it.

ncoghlan:
@dstufft We really shouldn't be allowing any new ABI compatibility tags without broad discussion of the ecosystem implications - while wanting to help out the Raspberry Pi project is admirable, we now have ARMv6 and ARMv71 wheels on PyPI that claim to work with any Linux distro, when they're actually specific to Raspbian :)

https://github.com/pypa/warehouse/issues/3668 is the follow-up issue to work out how we may be able to resolve this without breaking the Raspberry Pi UX.

PR #2012, opened 2017-05-18
Add linux_armv6l as a supported wheel type

kpayson64:
https://github.com/pypa/warehouse/issues/2003

armv6 is used by Raspberry Pi 2 and Zero


bennuttall:
FYI armv6 is used by Raspberry Pi 1 and Raspberry Pi Zero. armv7 is used by Raspberry Pi 2 and 3 (though technically Pi 3 is armv8, it generally runs an armv7 OS).

PR #2018, opened 2017-05-19
improve/add copywriting, cleanup related css

nlhkabu:
Closes https://github.com/pypa/warehouse/issues/1983 and https://github.com/pypa/warehouse/issues/1312

screenshot from 2017-05-19 18-59-33


dstufft:
Just an FYI, github doesn't understand the "Closes FOO and BAR" syntax, and it'll only automatically close Bar. I think you can do like Closes Foo, Closes Bar but I normally put them each on their own line personally.

PR #2021, opened 2017-05-21
Add browser warning. Allow messages to stack.

nlhkabu:
screenshot from 2017-05-21 15-26-17

Shows for ie9 and below

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -62,7 +62,18 @@
 
   <body>
 
-    {% include "warehouse:templates/includes/warning-banner.html" %}
+    <section class="stick-to-top js-stick-to-top">
+      {% include "warehouse:templates/includes/warning-banner.html" %}
+
+      <!-- Add browser warning. Will show for ie9 and below -->
+      <!--[if IE]>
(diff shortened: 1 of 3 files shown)


warehouse/templates/base.html
```diff
@@ -28,6 +28,15 @@
     <link rel="stylesheet" href="{{ request.static_path('warehouse:static/dist/css/font-awesome.css') }}">
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,400italic,600,600italic,700,700italic|Source+Code+Pro:500">
 
+    <!-- Add browser warning. Will show for ie9 and below -->
+    <!--[if IE]>
+	    <style>
```
dstufft: In general inline styles like this isn't going to work because of our CSP policy. I think this will only work because IE9 and older don't support CSP at all. Might it make more sense to use the conditional around the HTML instead of the CSS?

nlhkabu: yep, that will work.  I will update :)

nlhkabu:
updated

Issue #1923, opened 2017-04-17
Suggested search results give the wrong count


alexwlchan:
If I run a search on Warehouse that's slightly misspelt – e.g. "betamazx" (https://pypi.org/search/?q=betamazx) – because there aren't any results, it helpfully suggests that maybe I meant something else:

<img width="826" alt="There were no results for '*betamazx*'. Did you mean 'betamax'? (4 results)" src="https://cloud.githubusercontent.com/assets/301220/25083034/89af2008-234c-11e7-9110-b285a846d770.png">

And indeed, the suggestion is helpful – I click the link, and I see search results for "betamax", including the package I was looking for. But the count is wrong – the final page has 14 results, not 4.

Other mistyped searches:

* reqwests suggests "requests (83 results)", but a search for "requests" yields 5,951 results
* caverage suggests "coverage (13 results)", but a search for "coverage" yields 4,916 results
* hiper (3) sends me to hyper (142)

di:
Thanks for the report! It seems like the "suggestions" that elasticsearch gives us is only based on the project name, but our actual search results include hits on other fields as well.

bstutes:
The "actual" search is returning a list of modules that mention they use the "betamax" module somewhere in the project description.  Are you sure you want to label this as a "bug" or should the additional modules be listed just in case as user wants to also see where the module being searched for is used?  I just wanted to be sure before I took a shot at working on this "bug".

di:
This isn't a bug per se, but it definitely appears to the user that the count is off, by saying "4 results" and not something more (or less) specific.

It's going to be near-impossible to get the suggested "results" and the actual number of results to agree. Currently we search across quite a lot of fields:
```python
SEARCH_FIELDS = [
    "author", "author_email", "description", "download_url", "home_page",
    "keywords", "license", "maintainer", "maintainer_email", "normalized_name",
    "platform", "summary",
]
```
We would need to add an individual suggestion to the query for each of these fields, somehow determine which is the "top" suggestion from all of them aggregated together. Then, to get the actual number of results, we would have sum the results for each field, but ensure that we're not double-counting packages that have hits for the "top" suggestion in multiple fields.

It all seems like overkill for a feature that's not particularly useful in the first place -- we only give the user one suggestion, so the number of results doesn't actually matter.

I'd be in favor of just not showing the # of suggested results instead.

alexwlchan:
> I'd be in favor of just not showing the # of suggested results instead.

[...]

PR #2028, opened 2017-05-23
Bug #1923 search results give the wrong count

bstutes:
Made simple change agreed upon in the discussion thread for this bug.  Also found the <label...> was not properly associated with the <select....> tag to specify the sort order for returning projects found in a search.  Added "id=" attribute to properly associate the label.

Also cleaned up comments that document all changes made.  That's the reason for commits 85c2b0c, 5682ad0 and 87b1cc7.


warehouse/templates/search/results.html
```diff
@@ -35,9 +35,6 @@ <h3 class="package-snippet__title">
```
di: Looks like the line above this one (`% set freq = best_guess.freq %}`) can be removed as well.

bstutes: I'll make the additional change and create a new pull request.

di: @bstutes Making a new PR shouldn't be necessary, you should be able to just push a new commit to your existing branch.

warehouse/templates/search/results.html
```diff
@@ -143,7 +139,7 @@ <h2 class="no-top-padding">Filter Projects</h2>
           <div>
             <input id="search" type="hidden" name="q" value="{{ term }}">
             <label for="order">Order by &nbsp;</label>
-            <select class="-js-form-submit-trigger" name="o">
+            <select class="-js-form-submit-trigger" id="order" name="o">
```
dstufft: What's the purpose of this change?

di: It correctly makes the label correspond to the form element via the `for` attribute.

dstufft: Ahh, I missed that. Makes sense.

PR #2034, opened 2017-05-25
Remove double border with sticky banner

nlhkabu:
## Before

screenshot from 2017-05-25 08-15-57

## After

screenshot from 2017-05-25 08-15-39

Also fixed up the browser warning code, as there was an orphan `</div>` :)


nlhkabu:
Further testing found that this is not working as I want.  Closing :/

Issue #2052, opened 2017-05-28
Project tabs stop working after screen resize


nlhkabu:
To replicate:

- Open the screen at any desktop resolution
- Visit any project page
- Reduce the screen size until the layout changes and the side tabs stack on top of each other
- Click on the tabs

Notice that while the *content* changes (as it should), the correct tab is not highlighted.

This bug happens in the reverse order too: e.g. if you start on a small resolution and resize to desktop.

I believe the code we need to edit is here: `warehouse/static/js/warehouse/utils/project-tabs.js`

joedougherty:
I believe that this issue is being caused by two things:

1.) The `projectTabs` function accepts an argument `inMobileState`, but this variable is only computed on the initial page load. Any change to the window size will not alter the initial value of this variable.

2.) The `projectTabs` function defined in `warehouse/static/js/warehouse/utils/project-tabs.js` is only fired once (on `docReady` in `warehouse/static/js/warehouse/index.js`). This explains why the tab selected code only works in the context of the original window size. 

I believe this can be solved by moving the code that sets `inMobileState` from `index.js` to the function declaration itself. This way, `inMobileState` will be re-evaluated every time `projectTabs()` is called. Secondly, `projectTabs()` needs to be called on window resize.

I've created PR #2075 to help. Please let me know if there is anything more I can do.

dstufft:
Thanks for your excellent overview of what was causing this issue and the PR that fixed it!

Closing this since https://github.com/pypa/warehouse/pull/2075 is merged and solves the problem.

nlhkabu:
thanks @joedougherty :man_dancing:

PR #2075, opened 2017-06-02
Fixes project tabs screen resize problem

joedougherty:
This PR is designed to address #2052 

These changes move evaluation of `inMobileState` to the `projectTabs` function itself. They also call `projectTabs()` on window resize.

These changes do not includes tests as it does appear as though there are js-specific tests. I took a look at https://warehouse.readthedocs.io/development/submitting-patches but did not see anything about front-end testing. I've tested these changes manually in Firefox 52.0.3 and Chrome 57.0.2987.133 on Ubuntu GNOME 16.04. Please let me know if there is some additional testing I can do.

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -42,10 +42,9 @@ docReady(() => {
 docReady(()=> {
   const headingBtn = document.querySelector(".-js-vertical-tab-content");
   if (headingBtn) {
-    const mobileBtn = document.querySelector(".-js-vertical-tab-mobile-heading");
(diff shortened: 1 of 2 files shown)


dstufft:
Awesome! Thanks a lot for your contribution.

We do have a couple of Selenium tests, but out front end testing is largely aspirational at this point. It's not something I have a lot of experience with, so I haven't yet really figured out a good mechanism for handling it.

I've verified this solves the problem manually, so I'm going to merge this and it should start showing up on live within 10 minutes or so (though pages get cached so the roll out will be on a page by page basis as cache expires).

Thanks again!

PR #2124, opened 2017-06-20
Update user profile page

nlhkabu:
screenshot from 2017-06-20 09-06-11
screenshot from 2017-06-20 09-05-59

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -53,6 +53,7 @@ def failed_logins(exc, request):
 @view_config(
     route_name="accounts.profile",
     renderer="accounts/profile.html",
+    uses_session=True,
     decorator=[
         origin_cache(
             1 * 24 * 60 * 60,                 # 1 day
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -53,7 +53,6 @@ def failed_logins(exc, request):
 @view_config(
     route_name="accounts.profile",
     renderer="accounts/profile.html",
-    uses_session=True,
     decorator=[
         origin_cache(
             1 * 24 * 60 * 60,                 # 1 day
@@ -282,3 +281,12 @@ def _login_user(request, userid):
         user_service.update_user(userid, last_login=datetime.datetime.utcnow())
 
         return headers
+
+
+@view_config(
+    route_name="includes.current-user-profile-callout",
+    renderer="includes/accounts/profile-callout.html",
+    uses_session=True,
+)
+def profile_callout(user, request):
+    return {"user": user}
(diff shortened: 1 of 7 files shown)


warehouse/accounts/views.py
```diff
@@ -53,6 +53,7 @@ def failed_logins(exc, request):
 @view_config(
     route_name="accounts.profile",
     renderer="accounts/profile.html",
+    uses_session=True,
```
di: @dstufft mentioned [here](https://github.com/pypa/warehouse/pull/1444#r76335326) that he'd prefer not to turn on sessions for this page, however I'm not sure what the alternative would be.

dstufft: You'd mix it in with javascript instead, it's something that's easy enough to fail gracefully back if that fails for whatever reason. I think to do it really nicely would require a bit of work on the ``html_include()`` functionality so that it can have a default value that gets replaced with the fetched data.

So you'd end up with template code like:

```
{% htmlinclude request.route_path("includes.user-uploaded-message") %}
<p>{{ user.name|default(user.username, true) }} has not uploaded any projects to PyPI, yet.</p>
{% endhtmlinclude %}
```

Which would initially render the bit that's inside the blocks, and would fetch on the client side using JS the URL given and replace the block with the results of that.

Please note this will NOT currently work, but if that sounds reasonable I could work up the functionality of the current ``html_include()`` into something like that, which would work then.

nlhkabu: thanks for catching this @di.
sounds good @dstufft.  Would you make a separate PR for this, or include it in this one? (FWIW, I'm totally happy for you to edit this PR directly)

dstufft: Ok, if you merge master into this branch you can do this now in a template:

```
{% csi "/path/to/some/url/" %}
Content That will Render before the include has happened, or if JS is disabled for some reason.
{% endcsi %}
```

The work that's left to do in this PR then is to remove the stuff that touches the current logged in user on the user profile page, and replace it with the client side include (csi) block that has the "default" message on it and add a new URL that just renders the bit that you would like replaced and would be user dependent. You'll need to handle both the logged in and logged out case in that included URL, because it will be included unconditionally (unless the end user doesn't have JS or something) and it will completely replace whatever you put inside the ``{% csi %}{% endcsi %}`` blocks.

You can see an example of such a view at https://github.com/pypa/warehouse/blob/master/warehouse/views.py#L287-L293 which really just exists to render the template (thus why it's body is completely empty besides returning an empty dict). If you need help getting that setup I can help with that though.

nlhkabu: thanks @dstufft - can you please ping me on slack when you are around and we can go though this together.  I'm not sure I understand what I need to do here :)

nlhkabu:
rebased with master :)

dstufft:
FYI this has merge conflicts now.

nlhkabu:
ready to merge on my side @dstufft.  If you are happy, can you please go ahead?

Issue #780, opened 2015-11-20
Style the logged in / logged out indicator


dstufft:
We'll want to fix things so that there is a logged in indicator when the user is logged in that can list all of their projects and take them to the management pages for their projects.

dstufft:
With #837 this now only needs for the design to incorporate a logged in indicator. That bit of the design is controlled by the ``warehouse/templates/includes/current-user-indicator.html`` template.

di:
Now that #1437 is merged, should we take this out of the "Launch" milestone?

dstufft:
This should be good enough for launch yes.

nlhkabu:
Adding this back to launched milestone as we need to address this (along with the extra links) on mobile :)

![screenshot from 2017-05-25 11-41-34](https://cloud.githubusercontent.com/assets/3323703/26447170/2476a1b4-413f-11e7-983e-5018c3ed0197.png)

PR #2125, opened 2017-06-20
Add user dropdown menu

nlhkabu:
## Logged out
screenshot from 2017-06-20 19-37-14

## Logged in

screenshot from 2017-06-20 19-37-39

### DIFF
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -47,7 +47,7 @@ class TestLoginIndicator:
             a.get("href")
             for a in document.findall(".//nav[@id='user-indicator']/a")
         ]
-        assert urls == ["/account/login/",
-                        "/account/register/",
-                        "/help/",
-                        "https://donate.pypi.io"]
+        assert urls == ["/help/",
+                        "https://donate.pypi.io",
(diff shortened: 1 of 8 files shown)


nlhkabu:
Closes #780

dstufft:
Feel free to tell me I'm wrong, but it feels to me like the blue on blue makes it easy to "lose" the dropdown menu into the header image. Maybe a white border around the left/right/bottom would help distinguish it? I dunno, maybe it's fine too :)

nlhkabu:
I did try a white background, but it didn't feel consistent / part of the menu.  I'll try a white border, or maybe it's just a case of increasing the shadow.

FYI, I've also removed the not-essential links (donate, help) on mobile:
screenshot from 2017-06-20 20-41-29

i think we need to leave a fair bit of room here, because eventually we'll be translating these and they might become quite long.

nlhkabu:
screenshot from 2017-06-20 20-48-10

Subtle, but definitely better - thanks for the suggestion @dstufft :)

dstufft:
Yea, that stands out more as a separate entity and I don't feel like my eyes are getting lost. 💯

PR #2138, opened 2017-06-23
Add missing license information to JS files

nlhkabu:
- Cleaned up spacing in JS files
- Add licenses to missing files

@dstufft - we don't have licenses on each SCSS file.  Do we need to add these also?


dstufft:
@nlhkabu Yea, I guess we should add licenses to the SCSS files, they're covered under the license anyways.

Issue #2139, opened 2017-06-23
menu overflow


nlhkabu:
![screenshot from 2017-06-23 07-42-11](https://user-images.githubusercontent.com/3323703/27469454-8abce8d4-57e7-11e7-9d21-8ca66f2191d3.png)

@813px wide.

Will need to give more space here, because these links may become longer when translated.

pradyunsg:
Could someone close this?

Off Topic: @nlhkabu "Closes <issue number>" needs to be mentioned in the first post of the PR, for GitHub to auto close things. :)

PR #2153, opened 2017-06-28
reduce width of search to facilitate longer menu

nlhkabu:
Display at 801px (breakpoint is at 800)
screenshot from 2017-06-28 18-51-26


nlhkabu:
Closes #2139

PR #2156, opened 2017-06-29
remove overlay on desktop

### LINKED ISSUE #1775: dark-overlay remains after window is resized larger than the breakpoint
another issue that I found while messing around is that when you have the window sized small enough to show the "Add Filter" button, and then click it, and then resize the window to a larger size than the breakpoint (at 1000px currently), the screen remains unusable with no way of getting rid of the overlay.  You have to resize the window below 1000px and then click on the "close" button.

I believe it would be best to restructure the css code a bit, but currently, I can only come up with the following:

```
window.onresize = function () {
  if (window.screen.width > 1000) {
     document.querySelector('.-js-dark-overlay').style.display = '';
 	 document.querySelector('.-js-filter-panel').style.display= '';
  }
}
```

nlhkabu:
Closes #1775

This *should* work, but I can't confirm as I can't test this locally.  I am getting a `elasticsearch.exceptions.NotFoundError` when using the search.  @dstufft can you please ping me to help me get this sorted?

### DIFF
--- a/warehouse/static/sass/blocks/_applied-filters.scss
+++ b/warehouse/static/sass/blocks/_applied-filters.scss
@@ -28,10 +28,6 @@
   margin-top: 10px;
   @include clearfix;
 
-  @media only screen and (max-width: $tablet){
-    margin-top: 20px;
-  }
-
   &__add-button {
     display: none;
 
--- a/warehouse/static/sass/blocks/_dark-overlay.scss
+++ b/warehouse/static/sass/blocks/_dark-overlay.scss
@@ -19,13 +19,15 @@
   <div class="dark-overlay"></div>
 */
 
-.dark-overlay {
-  display: none;
-  background-color: transparentize($black, 0.4);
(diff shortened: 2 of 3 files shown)


dstufft:
Huh, did you try ``make initdb``? That should reset the DB and search index back to the default.

nlhkabu:
that's worked. thanks @dstufft

nlhkabu:
not sure why CI is failing here, error is `apt-get install failed`

nlhkabu:
Looks like master is failing on CI @dstufft https://travis-ci.org/pypa/warehouse/builds/248529070

As I have rebased this PR, it is also failing.  Otherwise this is ready to go on my side.

di:
@nlhkabu It seems like the cause of that failure is an issue with Travis: https://github.com/travis-ci/travis-ci/issues/7987

nlhkabu:
oh hu - thanks @di

PR #2163, opened 2017-07-01
add some information about github repo

AyumuKasuga:
If homepage in package is a github repo, we show some information about
that repo, like stars, forks, open issues, last push.

image


pradyunsg:
I really like the idea.

I'm not a maintainer of this project so I won't comment on the design or implementation parts.

dstufft:
I'll have to think on the implementation here a bit. Pushing it to the client side has benefits, but also downsides and I'll have to think through the trade off. It's possible we'll want to move it to server side to also allow us to provide different backends for different VCSs (in future PRs, this PR wouldn't be blocked on that).

While I do that though, @nlhkabu should probably take a look at the general concept and see what she thinks about it and it's overall picture in the design of Warehouse.

AyumuKasuga:
Hello @dstufft 
What downsides are you talking about?
GitHub has ratelimits for their api, and for this endpoint it's 60 per hour (maybe we can increase it, but it will still be limited i think). In frontend implementation we do not care about limits, if user reaches maximum requests per hour -- this information pane just will be invisible for him/her (and only for him/her), and that okay, guys from npmjs also use this behavior on their package page (https://www.npmjs.com/package/webpack for example). And of course we also don't load backend with excess work, and don't care about cache and cache invalidation and so on.

nlhkabu:
hi @AyumuKasuga - thanks for this PR.

Are you able to show me a larger screenshot, so I can see the change in the context of the whole sidebar?

Thanks

AyumuKasuga:
Hi @nlhkabu 
There is a full size screenshot 
github-stats-example

warehouse/templates/packaging/detail.html
```diff
@@ -134,6 +134,32 @@ <h3 class="sidebar-section__title">Project Links</h3>
         </div>
         {% endif %}
 
+        {% if release.github_repo_info_url %}
+        <div class="sidebar-section github-repo-info hidden" data-url="{{release.github_repo_info_url}}">
+          <h3 class="sidebar-section__title">GitHub stats</h3>
```
nlhkabu: Please capitalize and use full word "Statistics" rather than "Stats"

i.e. "GitHub Statistics"

warehouse/templates/packaging/detail.html
```diff
@@ -134,6 +134,32 @@ <h3 class="sidebar-section__title">Project Links</h3>
...
+            <span class="github-repo-info__item" data-key="stargazers_count"></span>
+          </p>
+          <p>
+            <i class="fa fa-code-fork" aria-hidden="true"></i>
+            <strong>Forks: </strong>
+            <span class="github-repo-info__item" data-key="forks"></span>
+          </p>
+          <p>
+            <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
+            <strong>Open issues: </strong>
```
nlhkabu: capitalize: "Open Issues"

warehouse/templates/packaging/detail.html
```diff
@@ -134,6 +134,32 @@ <h3 class="sidebar-section__title">Project Links</h3>
...
+            <span class="github-repo-info__item" data-key="open_issues_count"></span>
+          </p>
+          <p>
+            <i class="fa fa-clock-o" aria-hidden="true"></i>
+            <strong>Last push: </strong>
```
nlhkabu: capitalize: "Last Push"

AyumuKasuga:
@nlhkabu thank you for your review. I just capitalized all this things :)

nlhkabu:
ok on my side

warehouse/templates/packaging/detail.html
```diff
@@ -150,6 +150,32 @@ <h3 class="sidebar-section__title">Project Links</h3>
         </div>
         {% endif %}
 
+        {% if release.github_repo_info_url %}
+        <div class="sidebar-section github-repo-info hidden" data-url="{{release.github_repo_info_url}}">
+          <h3 class="sidebar-section__title">GitHub Statistics</h3>
+          <p>
```
di: Could you add these classes to this and following `<p>` tags?

```
vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed
```

warehouse/packaging/models.py
```diff
@@ -358,6 +359,19 @@ def urls(self):
...
+        if self.home_page:
+            parsed = urlparse(self.home_page)
+            if parsed.netloc == 'github.com' and parsed.path.count('/') >= 2:
+                splited_path = parsed.path.split('/')
+                user_name, repo_name = splited_path[1], splited_path[2]
+                tpl = "https://api.github.com/repos/{user_name}/{repo_name}"
+                return tpl.format(
+                    user_name=user_name,
+                    repo_name=repo_name
+                )
```
di: It is possible for users to specify multiple project URLs, any of which could be links to a github repo. Could you update this to look at all the URLs in `self.urls`, not just `self.home_page`?

di: Also I'm not sure what `tpl` means here, perhaps this could just be:

```python
return f"https://api.github.com/repos/{user_name}/{repo_name}"
```

warehouse/packaging/models.py
```diff
@@ -407,6 +408,14 @@ def urls(self):
 
         return _urls
 
+    @property
+    def github_repo_info_url(self):
+        for parsed in [urlparse(url) for url in self.urls.values()]:
+            segments = parsed.path.strip('/').rstrip('/').split('/')
+            if parsed.netloc == 'github.com' and len(segments) >= 2:
```
bskinn: Would it make sense to change this to something like:

    if parsed.netloc.endswith('github.com') and ...

That way the feature would twig on URLs using `www.github.com/user/repo`, not only `github.com/user/repo`.

di: @bskinn Great catch. Could you file an issue for this? Welcome to take a shot at fixing it as well, let us know if you need any help.

bskinn: Sure thing, filed as #3643. I'll take a shot at putting together a PR for it, too.

Issue #284, opened 2014-04-15
Determine new API URL structure for warehouse (starting with new JSON API)


ctheune:
At the PyCon2014 sprint I have started to make bandersnatch easier to cache. This means moving away from XML-RPC in general.

I'm leveraging the existing /pypi/<package>/json API which already helps, but I'll need two more endpoints:
- get a list of all packages and their most recent serial
- get the changelog

I implemented the necessary code on a branch for PyPI: 
https://bitbucket.org/ctheune/pypi/branch/ctheune-bandersnatch-json

However, I don't wanna force this through but have a decision how the URLs should look like.

Ideally we can implement this in both warehouse and PyPI in a way that bandersnatch can support both of them without breaking when you guys switch the public server (and I might be on vacation. ;) )

brainwane:
As I understand it, this issue (designing and implementing a new Warehouse API) is a prerequisite for integrating `twine` into `pip` and thus dealing with pypa/packaging-problems#76 and pypa/packaging-problems#60 , per @dstufft's comment in pypa/twine#127. Is that correct? If so, I'd suggest we add this to one of our upcoming milestones.

brainwane:
We talked about this issue [in today's bug triage meeting](https://wiki.python.org/psf/PackagingWG/2018-01-29-Warehouse) and folks explained to me: Even though this may be necessary for some Twine improvements, this is not a ticket we will address before launch. This is a new feature and is best suited for post-launch; Warehouse needs to be done before we can improve `twine`.

phildini:
Hello! Quick call-out that some of us would _really_ enjoy the JSON API containing info like owners/maintainers before the XMLRPC API is shut down. See ticket linked right above. Cheers, thanks for all your work!

brainwane:
I've marked #2914 as something we should address before shutting down legacy PyPI, but developing the structure for the new API can wait till after we shut down the legacy site.

brainwane:
As we develop the new API we should consider #347 as well. And I've added this issue to [the list of things we might work on at the PyCon sprints](https://wiki.python.org/psf/PackagingSprints).

jonparrott:
I would also love an API for managing both my account and my projects. For some examples of where this is useful:

1. We have an account that owns all of the projects our organization publishes. I want to rotate its password every week. 
2. Likewise, I want to audit all of my organization's project and verify that no more than n people have admin access to it.
[...]

PR #2165, opened 2017-07-02
Package update feed

### LINKED ISSUE #1683: Package update feeds
Was just having a thought about a neat feature that would be really helpful for me and probably others. Thinking of downstream/secondary package maintainers for instance though others may find it valuable too. My primary personal use case is helping keep [conda-forge]( http://conda-forge.github.io/ ) up to date. Though maybe Linux packagers and others would be interested

It would be really neat to have some kind of feed that would be published to when a particular package is updated. Mainly interested in version bumps, but potentially other events could be valuable if people are interested. It would be good if this feed was designed so that a person or bot could subscribe ideally without an account. Also it might be nice to have some way to filter out dev/prereleases from stable releases.

### LINKED ISSUE #2551: RSS feed for new file uploads to existing packages
Hi!

Currently there are two RSS feeds available (added by #990):
* https://pypi.org/rss/updates.xml -> "40 latest updates"
* https://pypi.org/rss/packages.xml -> "40 newest packages"

The former allows for monitoring of when a new version of a package is released, however doesn't show when new uploads to an existing package occur - which is a problem, since it's not unusual for files to be uploaded to a package in a staggered manner.

For example:
* twine uploading sdist and multiple wheels sequentially over a slow connection.
* projects using a combined Travis+AppVeyor CI solution to build the sdist and manylinux/Windows/OS X wheels (since no one CI solution covers all platforms), where the two CI systems get out of sync due to backlogs.
* projects uploading a wheel for the first time to an existing release that was previously sdist only.

This causes issues for projects using pip's `--require-hashes` mode along with tooling that automatically updates dependencies in requirements files (such as https://pyup.io/ ), since:

1. pyup.io sees the updates.xml RSS feed saying a new package version has been published
2. a PR is opened updating the requirements file with the new package version and hashes of whatever file uploads exist at that point in time
3. the tests may or may not pass at that point (depending on if the sdist or an appropriate wheel exists; in the aiohttp case the sdist wasn't uploaded for hours later)
4. even if the tests pass then, they may break later after the PR was merged, if an appropriate wheel becomes available, since pip prefers wheels over the sdist, and the requirements file will only contain the hash of the sdist

If an RSS feed were available of "40 newest file uploads" or even "40 newest changes of any type, whether that be metadata or uploads", then tools like pyup.io could use it to automatically update the requirements files hashes as uploads occurred (presumably with a slight delay to batch changes). 

See:
pyupio/pyup#250

Note: This is different from #1683 / #2165, since they are about a feed for just one package (and I think still only at the releases level, not the uploads level), rather than a feed for all packages site-wide.

CC @jayfk

AyumuKasuga:
Fixes #2551, fixes #1683.
Url like `/rss/{package_name}/updates.xml` with last 10 releases.


nlhkabu:
Hi @AyumuKasuga - thanks for this PR.
Can you please provide a screenshot of each of the template updates so I can validate?  Thanks

AyumuKasuga:
Hi @nlhkabu 
There is no visual changes in templates in this PR, it's only adds `<link>` tag in template [here](https://github.com/pypa/warehouse/pull/2165/files#diff-2300880d801924899c92e292b10886ccR60)

And RSS feed is just xml like this:
localhost-rss-eniarbiter-updates xml

warehouse/templates/rss/project_updates.xml
```diff
@@ -0,0 +1,14 @@
+{% extends "base.xml" %}
+{% block title %}Recent Updates of {{ project_name }} package{% endblock %}
+{% block description %}Recent Updates of {{ project_name }} to the Python Package Index{% endblock %}
```
di: It might be better if this title and description said something like "Latest releases of the `{{ project_name }}` package".

At the very least, this should say "updates **to the** `{{ project_name }}` package" instead of "updates **of** `{{ project_name }}` package"

warehouse/templates/packaging/detail.html
```diff
@@ -56,6 +56,10 @@ <h3 class="package-snippet__title">
 
 {% block description %}{{ release.summary }}{% endblock %}
 
+{% block extra_rss %}
+  <link rel="alternate" type="application/rss+xml" title="RSS: updates of {{ release.project.normalized_name }}" href="{{ request.route_path('rss.project_updates', name=release.project.normalized_name) }}">
```
di: Same comment here about "updates of".

warehouse/rss/views.py
```diff
@@ -46,6 +48,43 @@ def rss_updates(request):
...
+        .filter(
+            Project.normalized_name == func.normalize_pep426_name(project_name)
+        )
+        .order_by(Release.created.desc())
+        .limit(10)
+        .all()
+    )
+
+    if not latest_releases:
+        return HTTPNotFound()
```
di: I think there's a difference here between the case when project hasn't been registered yet, and the case when the project exists, but doesn't have any releases.

For the former, I would expect a 404, but for the latter, I would expect an RSS feed with no `<item>`s.

This would be handled by using `traverse` as mentioned in my other comment.

tests/unit/test_routes.py
```diff
@@ -141,6 +141,8 @@ def add_policy(name, filename):
             "https://files.example.com/packages/{path}",
         ),
         pretend.call("rss.updates", "/rss/updates.xml", domain=warehouse),
+        pretend.call("rss.project_updates",
+                     "/rss/{name}/updates.xml", domain=warehouse),
```
di: If you set the following:

```python
traverse="/{name}",
factory="warehouse.packaging.models:ProjectFactory",
```

you can expect a `project` argument to the `rss_project_updates` function, and a query will be unnecessary. This will also handle 404-ing if the project doesn't exist.

warehouse/rss/views.py
```diff
@@ -46,6 +48,43 @@ def rss_updates(request):
...
+
+    return {
+        "project_name": project_name,
+        "latest_releases": latest_releases
+    }
```
di: If we made the necessary changes to get a `project` via the `ProjectFactory`, we could just pass the `project` variable to the template here, and use `project.name` and `project.releases` (which is ordered by creation date) directly in the template.

AyumuKasuga:
Hello @di 
Thanks for the review. I changed templates and improved `project_updates` view using `ProjectFactory`

warehouse/templates/rss/project_updates.xml
```diff
@@ -0,0 +1,14 @@
+{% extends "base.xml" %}
+{% block title %}Latest releases of the {{ project.name }} package{% endblock %}
```
di: Nitpick: this should probably say "project", not package. It would be good to put the project name in quotes too.

warehouse/templates/rss/project_updates.xml
```diff
@@ -0,0 +1,14 @@
...
+{% block description %}Latest releases of the {{ project.name }} to the Python Package Index{% endblock %}
```
di: Same here, let's add quotes and "project".

AyumuKasuga:
@di thanks, fixed!

jakirkham:
Anything else needed for this PR? Seems very useful (though I'm biased 😄).

rce:
Hello @di! Is there something preventing merge and can I do something to make it happen?

di:
Yes, sorry for not following up. I had some reservations about this after approving, but haven't had a chance to address them (my time has mostly been spent in favor of development which helps fully us transition from pypi-legacy to Warehouse), so this PR has been on the backburner.

One thing: I wanted to check/ensure that this view is either not cached, or purged when a new release is made.

Another thing is that right now it's possible to delete releases (via pypi-legacy, and soon via Warehouse as well) and it would be ideal to have an entry in this feed for when a release is deleted as well. As it stands right now, the feed introduced by this PR just shows all the existing releases for a project. It might make more sense to look at the Journal entries for a project instead, which would include new releases and deletions.

I don't really want to start @AyumuKasuga off on revising this PR until I've really had a chance to think about it again (and other PyPI maintainers have weighed in) but perhaps they could do a little investigation of those ideas and report back.

brainwane:
I'm also sorry for the slow response!

To second Dustin: The folks working on Warehouse have gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and have kicked off work towards [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) -- the most urgent task is, as Dustin said, to improve Warehouse to the point where we can redirect pypi.python.org to pypi.org so the site is more sustainable and reliable. Since this feature isn't something that the legacy site has, I've moved it to a future milestone. But Dustin, Ernest, and other developers are interested in giving @AyumuKasuga feedback as they proceed, here or in IRC.

@di mentioned that the right approach might be to provide a feed of Journal updates -- @AyumuKasuga, you should check out #2871 to see that work progress, and maybe build on that!

Thanks and sorry again for the wait.

brainwane:
@AyumuKasuga Now that #2871 is finished, would you like to return to this PR and talk with us about how to proceed? Perhaps using the Journal entries?

As always, if you have questions along the way as you work on this, please feel free to ask them here, in [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev). Thank you!

brainwane:
@AyumuKasuga Ping :)

AyumuKasuga:
@brainwane  sorry, currently don't have a time to work on this.

di:
No worries @AyumuKasuga. Given that, I'm going to close this PR in favor of a different, yet-to-be-implemented API (see #284) which will ideally have the features I mentioned in https://github.com/pypa/warehouse/pull/2165#issuecomment-362689152.

Anyone that was hoping for this to be merged: Feel free to file a new issue with your specific use-case and we can try to help you make do with [the existing APIs](https://warehouse.readthedocs.io/api-reference/) (this will also help us consider your use-case when creating the new API).

PR #2172, opened 2017-07-06
support gpg keys

graingert:


graingert:
@dstufft not sure how to add a migration for this? I sort of assumed it would work if it used to be in legacy.

dstufft:
Hey there!

Sorry for the delay in getting back to you on this. I'm actually going to close this PR, because I don't think that we want to support PGP keys in the long term so improving them is not something that I'm super interested in at this point.

However, thanks for the contribution even if it wasn't ultimately accepted!

brainwane:
As far as I can tell this is reply by Donald is the closest thing we have to a record of the Warehouse policy decision that we will not be displaying GPG/PGP signatures for packages and will not give users a way to manage their public GPG keys. Also see https://github.com/pypa/warehouse/issues/2935#issuecomment-372515634 and the followups, and:

* 0df1ed21a327800ec192a40a25c02aca6f218ec0 in 2015 which included signatures in the package display
* c9c8ad9a19a72848c74107c29318e2b1bc1e5143 which took out those signatures
* 2ac719c5f32b26a76473b43ebd4f0ac8a5e81df8 which took out several gpg/pgp checks

Noting this for others' reference in case anyone is doing research.

brainwane:
Yes, people discussed the question in a few places. The [comment above](https://github.com/pypa/warehouse/pull/2172#issuecomment-330006945) seems to be the place where you publicly communicated that the decision was now final.

dstufft:
Yea that is probably about right.

PR #2176, opened 2017-07-09
Revise legacy upload error message

ddbeck:
As evidenced by pypa/python-packaging-user-guide#346, uploading a file is not so "simple." I've tried to change the message to be a little more direct about what to do instead.


ncoghlan:
The new section I just added to the migration guide about this change may be a better reference: https://packaging.python.org/guides/migrating-to-pypi-org/#registering-package-names-metadata

ncoghlan:
As far as a revised error message goes, I think the main problem here is the reference to "This API", which may be thoroughly ambiguous when displayed by a client. Here's a posssibly clearer alternative:

    Project pre-registration is no longer required or supported, so please continue directly to uploading files.

I'm less sure about including a link given that more precise wording, though - I've changed the migration page to actually quote the error message (so search engines should figure it out as long as we keep the messages in sync), and we really don't have much more to say about the matter than what's in that message (short of referring people off to the gory details in https://github.com/pypa/warehouse/issues/1627, and I'm not sure we want to do that)

PR #2193, opened 2017-07-14
Add newlines between links on simple pages

asottile:
Compare:
- [old pypi](https://pypi.python.org/simple/pre-commit)
- [new pypi](https://pypi.org/simple/pre-commit/)

This makes new pypi more like old pypi in this regard (and more human readable!)


warehouse/templates/legacy/api/simple/detail.html
```diff
@@ -19,7 +19,7 @@
   <body>
     <h1>Links for {{ project.name }}</h1>
     {% for file in files -%}
-    <a href="{{ request.route_url('packaging.file', path=file.path) }}#sha256={{ file.sha256_digest }}"{% if file.release.requires_python %} data-requires-python="{{ file.release.requires_python }}"{% endif %}>{{ file.filename }}</a>
+    <a href="{{ request.route_url('packaging.file', path=file.path) }}#sha256={{ file.sha256_digest }}"{% if file.release.requires_python %} data-requires-python="{{ file.release.requires_python }}"{% endif %}>{{ file.filename }}</a><br/>
```
xavfernandez: `<br>` should be enough

asottile: mostly went with `<br/>` since that's what old pypi had -- I agree for html `<br>` is sufficient -- I'll change this for ya

dstufft:
Thanks ✨

brainwane:
@asottile I wanted to thank you for this fix and also ask whether you'd like a suggestion of another bug to fix?

Issue #2206, opened 2017-07-19
Sort user page project list by last release date


mahmoud:
Right now the project listing on user page is sorted alphabetically:

![screenshot from 2017-07-19 01-42-04](https://user-images.githubusercontent.com/130193/28358448-c0df6b58-6c23-11e7-8194-8715a6686302.png)

I just released chert a couple days ago, but it's below ashes, released months ago.

Sorting by popularity, pinning, and other configurability might make sense later, but in the meantime I think a more sensible sort order would be by the recency of release. Makes the content seem fresher and more social, which is kind of the point of a user page with a big honkin gravatar, I figure.

Happy to contribute this if folks seem to agree.

---

**Good First Issue**: This issue is good for first time contributors. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

dstufft:
That seems reasonable to me.

mahmoud:
#2213 addresses this, give it a gander when you get a chance :)

di:
Just want to point out to any potential contributors that this issue is still up for grabs.

di:
Since this issue has been around for a while without being resolved, here's a pointer in the right direction for any potential first-time contributors:

We can modify the query in [`warehouse/accounts/views.py`](https://github.com/pypa/warehouse/blob/master/warehouse/accounts/views.py) to sort the projects by release date as follows:

```diff
diff --git a/warehouse/accounts/views.py b/warehouse/accounts/views.py
index c6ea043..26a636a 100644
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -75,13 +75,12 @@ def profile(user, request):
         )

     projects = (
-        request.db.query(Release)
-                  .options(joinedload(Release.project))
-                  .join(Project)
-                  .distinct(Project.name)
-                  .filter(Project.users.contains(user))
-                  .order_by(Project.name, Release._pypi_ordering.desc())
-                  .all()
+        request.db.query(Project)
+                  .filter(Project.users.contains(user))
+                  .options(joinedload(Project.releases))
+                  .join(Release)
[...]

PR #2213, opened 2017-07-20
Sorting projects on user pages by creation date

mahmoud:
Per #2206, this one-line change sorts projects on a user's page by time of last release. Works in the compose locally, all tests passing, as expected.

Couple of notes on implementation: I took the approach of doing a simple `.sort()` on the distributions because we're already fetching all the records anyways, and postgres's `ORDER BY` requires the first field to equal that of the `DISTINCT` clause when it's present ([docs](https://www.postgresql.org/docs/9.1/static/sql-select.html#SQL-DISTINCT)). I don't think it'll be a big deal, and I'd rather not complicate the query, just to do the sort on the db.

Hope this works for you!


dstufft:
You're missing coverage-- I'm guessing because the lambda isn't getting called.

dstufft:
But to be explicit, this looks fine to me and the rationale for doing it in memory instead of in the DB seems solid.

mahmoud:
Ah, so the test that covered the query returned 0 results, I guess. I'll see if I can track that down.

brainwane:
Thanks for your pull request, @mahmoud!

The folks working on Warehouse have just gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and will in December kick off work towards [the first milestone: redirecting pypi.python.org to pypi.org](https://github.com/pypa/warehouse/milestone/1). So if you have time to revise this PR, then after that kickoff we'll be pretty responsive with feedback.

mahmoud:
Hey @brainwane! How fun, running into you here :)

So this would-be-1-line PR died for me because I couldn't run tests locally, even when the site ran. Unfortunately, the coverage requirement is very high, but the parity between tests and site behavior is low. Specifically, the fake users created for the tests don't have any projects, so my one line was only half-hit. In any case, I followed the instructions in the guide, I got all manner of errors and I just ended up going down the rabbit hole of the Makefile.

I'd like to see some of those resources go toward making that workflow better, because if it's this hard to make a small change to Warehouse, it's pretty hard to tell myself and others that it's a good way to spend their time. 

Anyways, I'll keep an eye on this and am happy to give it another go once things improve.

brainwane:
Hi @mahmoud and thanks for the feedback! I'm happy to run into you here as well, and I'm sorry you had this trouble. :(

So it sounds like one problem you ran into is that -- even on `master` -- you couldn't get the tests to run (is that right?) and another problem you ran into is that the fake (test) users didn't have any projects. I see (via `git log -- tests/unit/`) that in the last two months we've added a lot more to our test coverage and the available test classes, and I think we've addressed that issue; thanks for bringing it up.

So if you wouldn't mind trying to get the tests to run on `master` and letting us know -- in IRC in #pypa-dev on Freenode, or in a new issue -- if you still have trouble running them, that would be great! We have, just in the last couple months, smoothed out some of the new contributor snags, and we'd be glad to find and smooth out any more.

Thanks.

brainwane:
@mahmoud I won't bother you about this today, but new contributors generally have an easy time of it now, getting the tests to run, so I hope you'll come back and try again if you have other future improvements to make. :)

Issue #2185, opened 2017-07-11
Upload fails if classifier in setup.py is tuple rather than list


gilessbrown:
When I run `python setup.py sdist upload` I get an error message:

```
(wextracto) gsbrown$ python setup.py sdist upload
running sdist
running egg_info
writing requirements to Wextracto.egg-info/requires.txt
writing Wextracto.egg-info/PKG-INFO
writing top-level names to Wextracto.egg-info/top_level.txt
writing dependency_links to Wextracto.egg-info/dependency_links.txt
writing entry points to Wextracto.egg-info/entry_points.txt
reading manifest file 'Wextracto.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'Wextracto.egg-info/SOURCES.txt'
running check
creating Wextracto-0.14.0
creating Wextracto-0.14.0/Wextracto.egg-info
creating Wextracto-0.14.0/wex
creating Wextracto-0.14.0/wex/js
copying files to Wextracto-0.14.0...
<lots of "copying ..." lines snipped>
Writing Wextracto-0.14.0/setup.cfg
Creating tar archive
removing 'Wextracto-0.14.0' (and everything under it)
running upload
Submitting dist/Wextracto-0.14.0.tar.gz to https://upload.pypi.org/legacy/
Upload failed (400): classifiers: 'FieldStorage('classifiers', 'Development Status :: 4 - Beta')' is not a valid choice for this field
error: Upload failed (400): classifiers: 'FieldStorage('classifiers', 'Development Status :: 4 - Beta')' is not a valid choice for this field
```

That is with a tuple for classifiers as per:
https://github.com/eBay/wextracto/blob/cfa19d36c8054670987705cef172f7e84c1f934c/setup.py#L31

If I change from `(` to `[` (and the matching pair) to make it a list rather than a tuple it uploads successfully.

(Originally raised as pypa/pypi-legacy#670)

bucknerns:
Same issue here thought the Intended Audience classifier was removed based on the error.  


Copying sshaolin.egg-info to build/bdist.linux-x86_64/wheel/sshaolin-0.0.5-py2.7.egg-info
running install_scripts
creating build/bdist.linux-x86_64/wheel/sshaolin-0.0.5.dist-info/WHEEL
running upload
[...]

PR #2255, opened 2017-07-30
Support uploading tuple of classifiers

gtback:
Added a failing test for #2185, and added a *potential* fix. I'm not a SQLAlchemy expert, so I expect there are likely better ways to fix it, but the test is (hopefully) useful regardless.

@gifforda helped me out with this.

### DIFF
--- a/tests/common/db/base.py
+++ b/tests/common/db/base.py
@@ -31,7 +31,17 @@ class WarehouseFactory(SQLAlchemyModelFactory):
 
     @classmethod
     def _create(cls, *args, **kwargs):
-        r = super()._create(*args, **kwargs)
+        new_kwargs = {}
+        for k, v in kwargs.items():
+            if isinstance(v, tuple):
+                v = list(v)
+            new_kwargs[k] = v
+        new_args = []
+        for a in args:
+            if isinstance(a, tuple):
+                a = list(a)
+            new_args.append(a)
+        r = super()._create(*new_args, **new_kwargs)
         session = cls._meta.sqlalchemy_session
         session.flush()
         return r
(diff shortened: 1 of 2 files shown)


dstufft:
Hmm, I don't think this is going to fix #2185, because this is just modifying the test functions and not any of the actual code.

gtback:
Yikes, I didn't realize that was just the test code until just now.

dstufft:
No worries! FWIW I *think* the issue with #2185 might actually be client side? I haven't dig into it to figure it out, but I think that distutils/setuptools sends different content for tuple vs list there and I'm not sure why. It might be something we can work around in Warehouse though, I'm not sure.

gtback:
I haven't used SQL Alchemy for anything non-trivial, but I would imagine there's some easy way for a Many-to-Many relationship to accept any iterable, rather than just a `list` (I just don't know what that is). It's possible that better client-side support would mitigate the bug, but Warehouse could also probably silently cast a tuple into a list. 

I tried adding `collection_class=list,` into [here](https://github.com/pypa/warehouse/blob/0d01a752790fbc04967d8355a49f754c6cf10acc/warehouse/packaging/models.py#L272), but that didn't seem to make a difference (at least in terms of the failing test), giving me `TypeError: Incompatible collection type: tuple is not list-like`.

di:
Hey @gtback, thanks for taking a stab at this. Like Donald said, this looks like a distutils issue rather than an issue with Warehouse, so I'm going to close this PR.

If you're still interested in contributing to Warehouse, take a look at [these issues which are good for first-time contributors](https://github.com/pypa/warehouse/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).

gtback:
Thanks, @di, and sorry I let this languish. The original approach didn't work, and I never spent the time to find a better one. I'll look for other issues to help out with. 😄

PR #2311, opened 2017-08-13
Add aria roles to search, tabs

### LINKED ISSUE #810: Add ARIA roles to HTML
From our accessibility audit by @xavierdutreilh.

We should add [ARIA](http://www.w3.org/TR/wai-aria/roles#landmark_roles) roles to our HTML to make it more accessible for users using screen readers.

Of particular concern are:
- The search form (`<form ... role="search">`)
- The vertical tabs on the packages page (http://www.w3.org/TR/wai-aria/roles#tab)

We also need to add `aria-hidden="true"` to the link icons on the package page as they are not readable by screen readers.  See https://github.com/FortAwesome/Font-Awesome/issues/6133 for more information.

nlhkabu:
Closes #810 

From what I can tell, these are the only ARIA roles we should need.  Most others are covered implicitly by using semantic HTML5. E.g. `nav` automatically takes the role of `navigation`.


nlhkabu:
Also addresses accessibility on our font awesome icons, as per the advice [here](http://fontawesome.io/accessibility/).  This should close https://github.com/pypa/warehouse/issues/1202

PR #2357, opened 2017-09-04
use request.remote_addr rather than request.client_addr

ewdurbin:
warehouse.utils.wsgi.ProxyFixer goes to great lenghts in combination with our CDN VCL to resolve the correct User IP Address into REMOTE_ADDR, see https://github.com/pypa/warehouse/issues/256

the behavior of client_addr is not quite correct, leading to JournalEntry with values for submitted_from which are Heroku IP Addresses see https://github.com/pypa/pypi-legacy/issues/603


alex:
Looks like there's some test fixtures needing updates.

PR #2373, opened 2017-09-08
Add description_content_type field to Release

di:
This PR adds the `description_content_type` field to the `Release` model and captures it on upload.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -327,7 +327,7 @@ class MetadataForm(forms.Form):
                 # Note: This isn't really Metadata 2.0, however bdist_wheel
                 #       claims it is producing a Metadata 2.0 metadata when in
                 #       reality it's more like 1.2 with some extensions.
-                ["1.0", "1.1", "1.2", "2.0"],
+                ["1.0", "1.1", "1.2", "2.0", "2.1"],
                 message="Unknown Metadata Version",
             ),
         ],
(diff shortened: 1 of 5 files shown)


hallazzang:
I'm waiting for this PR to be merged

di:
@hallazzang Thanks for your interest. You might want to subscribe to https://github.com/pypa/warehouse/issues/869 where I've been tracking most of the work necessary (including this PR), specifically in [this comment](https://github.com/pypa/warehouse/issues/869#issuecomment-340928703).

warehouse/forklift/legacy.py
```diff
@@ -16,6 +16,7 @@
 import re
 import tempfile
 import zipfile
+from cgi import parse_header
 
 from cgi import FieldStorage
```
merwok: Seems weird to use old cgi module functions where Pyramid provides the needed parsers.

di: @merwok Thanks for the review. Assuming you're talking about `parse_header`, would you mind pointing me to the equivalent function which Pyramid provides?

di:
For those watching, this PR does not get you descriptions in Markdown, [there are still a few outstanding blockers here](https://github.com/pypa/warehouse/issues/869#issuecomment-340928703) (but we're close).

warehouse/forklift/legacy.py
merwok: Sorry for missing this!  I’m talking about both parse_header and FieldStorage; Pyramid parses these and surely exposes a way to get Python objects for them, but I can’t look for it right now.

PR #2374, opened 2017-09-09
first pass at an admin UI for projects

ewdurbin:
A really dumb first pass at an admin screen for Projects.

This will at least give admins quick access to Maintainers for a project, releases, and journal entries.

### DIFF
--- a/warehouse/admin/views/projects.py
+++ b/warehouse/admin/views/projects.py
@@ -13,10 +13,12 @@
 import shlex
 
 from paginate_sqlalchemy import SqlalchemyOrmPage as SQLAlchemyORMPage
-from pyramid.httpexceptions import HTTPBadRequest, HTTPNotFound
+from pyramid.httpexceptions import (
+    HTTPBadRequest,
+    HTTPMovedPermanently,
+)
 from pyramid.view import view_config
 from sqlalchemy import or_
-from sqlalchemy.orm.exc import NoResultFound
 
 from warehouse.accounts.models import User
 from warehouse.packaging.models import Project, Release, Role, JournalEntry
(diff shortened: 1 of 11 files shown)


ewdurbin:
screen shot 2017-09-09 at 4 06 47 pm

ewdurbin:
screen shot 2017-09-09 at 4 13 25 pm

ewdurbin:
screen shot 2017-09-09 at 5 36 01 pm
screen shot 2017-09-09 at 5 36 04 pm
screen shot 2017-09-09 at 5 36 10 pm
screen shot 2017-09-09 at 5 36 15 pm
screen shot 2017-09-09 at 5 36 19 pm

ewdurbin:
need to get test coverage in place, but ready for a review.

warehouse/admin/views/projects.py
```diff
@@ -0,0 +1,201 @@
...
+    q = request.params.get("q")
+    project_name = request.matchdict["project_name"]
+
+    try:
+        page_num = int(request.params.get("page", 1))
+    except ValueError:
+        raise HTTPBadRequest("'page' must be an integer.") from None
+
+    project = (request.db.query(Project)
+               .filter(or_(Project.name == project_name,
```
dstufft: Project detail only works with the normalized name, this view works with either. Probably we should be consistent one way or the other. You can actually use the ``ProjectFactory`` to do that though. Take a look at th enormal project detail pages to see how that works. It also includes redirecting to the canonical URL which we probably want to do here too.

ewdurbin:
screen shot 2017-09-10 at 9 50 05 am
screen shot 2017-09-10 at 9 50 09 am
screen shot 2017-09-10 at 9 50 12 am
screen shot 2017-09-10 at 9 50 16 am
screen shot 2017-09-10 at 9 50 21 am

warehouse/admin/views/projects.py
dstufft: Alternatively we might want to consider adding a UUID pk to these URLs and using that in the admin like was done for Users. That way our admin URLs are stable even if something gets renamed etc. Downside to that of course is that our Admin URLs are not as "pretty" and don't encode anything human readable in them.

ewdurbin: @dstufft good tip, but the ProjectFactory approach appears to be interfering with the permissions. any pointers? i'm not really all that familiar with pyramid and it's terminologies.

dstufft: Oh, I think you need to add the admin permission to the Admin group for the Project. In the Model for Projects there is a ``__acl__`` method which returns a list of ACLs, you need to add one like: ``(Allow, "group:admins", "admin")``, which will say to Allow anyone with "group:admins" principal (e.g. if you're a site admin you have that principal assigned to you) actions on this object that require the ``"admin"`` permission.

dstufft: In Pyramid, views have contexts, so what you're doing with that ProjectFactory is setting the context for that view, and the permission system uses the ACLs of the context to decide if the current user has permissions to do this view or not.

It works by default, because Pyramid has a default No-op context, which exists at https://github.com/pypa/warehouse/blob/c6055f056faa6170706bc54c691be231b26c17eb/warehouse/config.py#L54-L64 that I setup to already have that ACL added.

ewdurbin:
@dstufft only remaining question i have is how to keep the new routes from giving `404`s instead of `403`s for non existent packages.

For instance, a non logged in user calling:

`GET http://localhost/admin/projects/requests/` gets a 403
`GET http://localhost/admin/projects/doesnotexist/` gets a 404

Seems to be something with the way the `ProjectFactory` works, but again my lack of Pyramid chops is showing.

dstufft:
Yea, it's because the layer that finds the object (and raises a 404) occurs at a different layer then the one that checks permissions. Personally I don't think it's a big deal for our use case here. The primary purpose for wanting routes like these to give 403 instead of 404 is to avoid leaking information about what objects exist (but that you don't have permission for).

If we think we need/want 403s in all cases instead of 404's, then for the admin UI, this is going to be better solved generically by registering a new handler for 404 exceptions that only works for the admin URLs. I did that at https://github.com/pypa/warehouse/blob/c6055f056faa6170706bc54c691be231b26c17eb/warehouse/admin/views/core.py#L18-L20 for turning 403s into a redirect to the admin login page. So it would be something similar to that.

I don't think it really matters here though, and if it did I think it'd make sense to make that it's own PR.

ewdurbin:
works for me @dstufft.

Issue #2377, opened 2017-09-11
Project Admin shows no journals for some projects


dstufft:
Projects named like "Django" do not show any journals whereas projects like "setuptools" do. The apparent common thread amongst those that do not show journals is that the real name and the normalized name differ.

PR #2380, opened 2017-09-12
resolves journal display issue from #2377

ewdurbin:
@clgarciga and I investigated #2377 and determined that there was a
mismatch in the journals_query for the admin.project.journals view.

JournalEntry is stored with the non-normalized name for the project, and
the query previously was comparing Project.normalized_name to JournalEntry.name

one thing to note, I'm not sure if journals are updated when projects have been
renamed in the past. It might be worth adding a normalized_name to JournalEntry
for safety, or add uuids as @dstufft alluded in the past

Co-authored-by: <cglanza@gmail.com>


dstufft:
Looks good to me! ✨

Issue #2281, opened 2017-08-05
Mention PEP 527 on `Invalid file extension`


boppreh:
[PEP 527](https://www.python.org/dev/peps/pep-0527/) forbids the upload of any sdist with extension other than `.tar.gz` or `.zip`. Warehouse rejects such uploads with the message `400 Invalid file extension.` ([implementation](https://github.com/pypa/warehouse/blob/2a99b6f8b0ffa66a13d52b493c756cecc8307be1/warehouse/forklift/legacy.py#L763), [test](https://github.com/pypa/warehouse/blob/57f70e977c979f0ca2a2fe640a9362eaf72898af/tests/unit/forklift/test_legacy.py#L904)).

I suggest adding more information to this message, preferably mentioning [PEP 527](https://www.python.org/dev/peps/pep-0527/) and that some older projects have been grandfathered for a temporary period. This prevents users from getting confused by the seemingly inconsistent behavior, and to distinguish between files rejected by PEP 527 and other cases.

di:
Fixed in #2387.

PR #2387, opened 2017-09-14
Mention PEP 527 in “Invalid file extension” response

jonemo:
Addresses https://github.com/pypa/warehouse/issues/2281

This updates the HTTP response when uploading files named in violation of PEP 527 from

```
400 Invalid file extension.
```

to

```
400 Invalid file extension. PEP 527 requires .tar.gz or .zip (https://www.python.org/dev/peps/pep-0527/).
```

Original request:

> PEP 527 forbids the upload of any sdist with extension other than .tar.gz or .zip. Warehouse rejects such uploads with the message 400 Invalid file extension. (implementation, test).

> I suggest adding more information to this message, preferably mentioning PEP 527 and that some older projects have been grandfathered for a temporary period. This prevents users from getting confused by the seemingly inconsistent behavior, and to distinguish between files rejected by PEP 527 and other cases.

After reviewing the related code, it appears to me that the issue author's concern about confusion "between files rejected by PEP 527 and other cases" is in fact not a problem because `400 Invalid file extension` is only ever raised for PEP 527 incompatible file names. This makes this change seem less necessary, but in my opinion still worthwhile.


dstufft:
Hey there!

Thanks for this PR, it looks good except it should probably include ``.whl`` and ``.egg`` too.

jonemo:
Heh, looks like both the author of #2281 and I misread PEP 527 :) https://github.com/pypa/warehouse/pull/2387/commits/082737e1a926abe026f677d364f22af14d24ae05 adds a mention of `.egg` and `.whl` to the message.

Out of curiosity: How far along is the transition described in PEP 527? i.e. how much longer will there be package that need support for the other file extensions?

dstufft:
I probably need to finish it, I forgot about it to be honest.

dstufft:
Thanks! ✨

PR #2396, opened 2017-09-16
Databased Backed Blacklists

dstufft:
This removes the blacklist that was hardcoded, and instead replaces it with one that is stored in the database. It also provides an admin UI to view and add new blacklists.

This is a somewhat of a dangerous workflow, because blacklisting a name also involves removing all files, releases, etc that exist for that (essentially the same thing as hitting remove package on legacy PyPI). Due to the danger, adding a blacklist requires confirmation.

This still requires a few things:

- [x] Add the ability to remove existing blacklists to allow users to claim the name again.
- [x] Make the confirmation page harder to blindly click though.
  - This might be requiring typing the name of the project that would be deleted into an input box?
  - This might require typing in a random string?
- [x] Write tests

Throwing this up here so people can take a look at it before I finish it.


warehouse/admin/templates/admin/blacklist/confirm.html
```diff
@@ -0,0 +1,72 @@
...
+  <div class="row">
+      <div class="col-md-2"></div>
+      <div class="col-md-8">
+        <div class="box box-primary">
+          <div class="box-body">
+              {% if existing.project %}
+              <p>
+                Blacklisting {{ blacklist.project }} will irreversibly delete
+                the {{ existing.project.name }} project along with
+                {{ existing.releases|length() }} releases and
```
ewdurbin: worth linking to the releases view?

warehouse/admin/templates/admin/blacklist/confirm.html
```diff
@@ -0,0 +1,72 @@
...
+{% block content %}
+  <div class="row">
+      <div class="col-md-2"></div>
+      <div class="col-md-8">
+        <div class="box box-primary">
+          <div class="box-body">
+              {% if existing.project %}
+              <p>
+                Blacklisting {{ blacklist.project }} will irreversibly delete
+                the {{ existing.project.name }} project along with
```
ewdurbin: worth linking to the project detail view?

warehouse/admin/templates/admin/blacklist/confirm.html
```diff
@@ -0,0 +1,72 @@
...
+                {{ existing.releases|length() }} releases and
+                {{ existing.files|length() }} files. This project has
+                {{ existing.users|length() }} users able to access it, listed
+                below:
+
+                <ul>
+                  {% for user in existing.users %}
```
ewdurbin: i don't think it'd be trivial but exposing the role_name here would be nice.

warehouse/admin/views/blacklist.py
```diff
@@ -0,0 +1,251 @@
...
+    # Verify that the user has confirmed the request to blacklist.
+    confirm = request.POST.get("confirm")
+    if not confirm:
+        request.session.flash(
+            "Must confirm the blacklist request.",
+            queue="error",
+        )
+        return HTTPMovedPermanently(request.current_route_path())
+    elif canonicalize_name(confirm) != canonicalize_name(project_name):
+        request.session.flash(
```
ewdurbin: the flash produced is pretty much identical in color to the header in admin, not a huge deal but threw me off when reviewing. 
screen shot 2017-09-16 at 7 34 29 pm

warehouse/admin/views/blacklist.py
```diff
@@ -0,0 +1,251 @@
...
+        request.db.query(Project)
+                  .filter(Project.normalized_name ==
+                          func.normalize_pep426_name(project_name))
+                  .first()
+    )
+    if project is not None:
+        request.db.add(
+            JournalEntry(
+                name=project.name,
+                action="remove",
```
ewdurbin: should the action here be `blacklisted`?

warehouse/admin/views/blacklist.py
```diff
@@ -0,0 +1,251 @@
...
+                                   project.name)))
+        request.db.query(Release).filter(Release.name == project.name).delete()
+        request.db.query(Project).filter(Project.name == project.name).delete()
+
+    request.session.flash(
+        f"Successfully blacklisted {project_name!r}",
+        queue="success",
+    )
+
+    return HTTPMovedPermanently(request.route_path("admin.blacklist.list"))
```
ewdurbin: maybe just a redirect?

warehouse/admin/views/blacklist.py
dstufft: I'm going to leave this alone, because other things might be reading our journal entries and inferring things from them, and we want them to still go ahead and remove these projects if they are. It's not really super important to keep in the log that a project was blacklisted either, since it's a binary state and admins can see when it was and who did it (as well as any relevant comments).

dstufft:
Feedback handled. I'm a little nervous about the idea of a bulk blacklisting, since blacklisting means deleting packages and is generally irreversible, perhaps once we get some more experience with the single entrypoint we can explore that?

ncoghlan:
As far as a bulk feature goes, that may be more appropriate to tackle as a "bulk quarantine" (i.e. shuffle all the affected artifacts off to a different S3 bucket, without deleting them outright) rather than as bulk automation of the existing removal behaviour (quarantine also has the benefit of preserving the artifacts for security analysis, similar to the way anti-virus quarantining works).

dstufft:
We don't actually delete the files from S3 even with bulk delete. The cost of storing the files is minuscule it's actually serving the files that actually costs the bulk of our S3 "bill". This also makes it easier for us to deal with transactionally deleting these projects since it's all contained inside of the database.

So deleting here is really just deleting the data from the database (although maybe at some point we'll have some process to garbage collect data from S3 that is no longer referenced in the DB).

ncoghlan:
OK, so even today the database entries for "deleted" releases could potentially be backfilled from the artifacts retained in S3. I think that would already mitigate a lot of the risk associated with removal errors, even if the scripts for state reconstruction don't exist yet.

warehouse/forklift/legacy.py
```diff
@@ -639,6 +631,17 @@ def file_upload(request):
...
     except NoResultFound:
+        # Before we create the project, we're going to check our blacklist to
+        # see if this project is even allowed to be registered. If it is not,
+        # then we're going to deny the request to create this project.
+        if request.db.query(exists().where(
+                BlacklistedProject.name ==
+                func.normalize_pep426_name(form.name.data))).scalar():
+            raise _exc_with_message(
+                HTTPBadRequest,
+                "The name {!r} is not allowed.".format(form.name.data),
```
benjaoming: Hi @dstufft - glad you started something!

Would suggest the message to be a little more clear?

    For concerns of security (namesquatting and typosquatting), '{!r}' is not an allowed name for a project. Please use a different name.

ncoghlan: I added that question to https://github.com/pypa/warehouse/issues/2401#issuecomment-330107332 (as I think it's related to the general UX of how we communicate invalid name declarations to end users)

PR #2400, opened 2017-09-17
Disable autocomplete

dstufft:


dstufft:
It works in Chrome and it's on MDN so that should be Chrome and Firefox at the very least.

PR #2404, opened 2017-09-17
add blacklist form to admin.project.detail view

ewdurbin:


ewdurbin:
this is suboptimal as it's a full copy-pasta of the form, only thing to note is that the project name field is not intended to be modified.

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -108,4 +108,29 @@ <h3 class="project-name text-center">{{ project.name }}</h3>
...
+  <div class="box box-secondary">
+    <div class="box-header with-border">
+      <h3 class="box-title">Blacklist Project</h3>
+    </div>
+
+    <form method="GET" action="{{ request.route_path('admin.blacklist.add') }}">
+      <div class="box-body">
+        <div class="form-group col-sm-4">
+          <label for="blacklistedProject">Project Name</label>
+          <input name="project" class="form-control" id="blacklistedProject" readonly="readonly" value="{{ project.normalized_name }}" placeholder="{{ project.name }}">
```
dstufft: Maybe just ditch this field all together and make it a hidden field? We should already know what project we're blacklisting since it's all over the page and we're going to have to type the name to confirm it anyways. That would narrow this form down to just comment + the submit button.

ewdurbin: 👍

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -108,4 +108,28 @@ <h3 class="project-name text-center">{{ project.name }}</h3>
...
       </div>
   </div>
+  <div class="box box-secondary">
+    <div class="box-header with-border">
+      <h3 class="box-title">Blacklist Project</h3>
+    </div>
+
+    <form method="GET" action="{{ request.route_path('admin.blacklist.add') }}">
+      <div class="box-body">
+        <div class="form-group">
```
dstufft: We no longer need this div, since this is a hidden field now it can just leave directly under the ``<form>``.

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -108,4 +108,28 @@ <h3 class="project-name text-center">{{ project.name }}</h3>
...
+          <input name="project" class="form-control" id="blacklistedProject" type="hidden" readonly="readonly" value="{{ project.normalized_name }}" placeholder="{{ project.name }}">
```
dstufft: The ``class``, ``id``, ``readonly``, and ``placeholder`` attributes are no longer needed now that this is a hidden field.

PR #2409, opened 2017-09-17
Block stdlib names

ewdurbin:
Prohibits new projects with names which collide on canonicalized name with standard library modules from being created.

Uses stdlib_list from @jackmaney for programatic access to a hopefully up to date list of standard library names.

RE: #2151


alex:
There are several packages in the stdlib that have backports on PyPI, or were originally on PyPI. For example:

- https://pypi.org/project/asyncio/
- https://pypi.org/project/ipaddress/
- https://pypi.org/project/wsgiref/

What does this do with them?

ewdurbin:
@alex, no impact on existing projects, this just stops any new projects with such a name from being created without admin intervention.

existing projects can upload new releases without any intervention.

jonemo:
In the discussion in #2151 the requirement to have an escape hatch came up, for example for packages that backport stdlib modules. How would those be handled?

ewdurbin:
my opinion is that any new backports should be explicitly named as such. in the rare case that a package name included in the standard library module requires registration with the precise name, an administrative tool in this project should allow that.

warehouse/forklift/legacy.py
```diff
@@ -46,6 +50,17 @@
...
 PATH_HASHER = "blake2_256"
 
+ALL_STDLIB = set(
+    chain.from_iterable(
+        [stdlib_list.stdlib_list(version)
+         for version in stdlib_list.short_versions]
+    )
+)
+
+STDLIB_PROHIBITTED = [
```
dstufft: Small nitpick, but this would probably be better as a set, it will save some duplication for like ``ConfigParser`` vs ``configparser`` and will also make the ``in`` check later on a tiny bit faster.

warehouse/forklift/legacy.py
```diff
@@ -46,6 +50,17 @@
 
 PATH_HASHER = "blake2_256"
 
+ALL_STDLIB = set(
+    chain.from_iterable(
+        [stdlib_list.stdlib_list(version)
```
dstufft: This would be slightly better as ``(...)`` instead of ``[...]`` to make this is a generator expression. That will prevent us from creating an intermediate list.

warehouse/forklift/legacy.py
```diff
@@ -46,6 +50,17 @@
 
 PATH_HASHER = "blake2_256"
 
+ALL_STDLIB = set(
```
dstufft: Any reason for having ``ALL_STDLIB`` exist instead of just nesting this logic inside the comprehension for ``STDLIB_PROHIBITED``? We could ditch the "inner" ``set()`` then and just have ``STDLIB_PROHIBITED`` be a set comprehension.

warehouse/forklift/legacy.py
```diff
@@ -642,6 +657,14 @@ def file_upload(request):
                 "The name {!r} is not allowed.".format(form.name.data),
             ) from None
 
+        # Also check for collisions with Python Standard Library modules.
+        if (packaging.utils.canonicalize_name(form.name.data) in
+                STDLIB_PROHIBITTED):
+            raise _exc_with_message(
+                HTTPBadRequest,
+                "The name {!r} is not allowed.".format(form.name.data),
```
dstufft: Should we have a distinct message here? Even if it's just including "because it is the name of a Python standard library module" in the existing message?

ncoghlan:
For future backporting-without-admin-approval, note that the backports namespace package already exists as a self-documenting target for that purpose: https://pypi.python.org/pypi/backports/

(While that backport namespace itself could also be targeted, it isn't vulnerable to the naive "pip install urllib" problem that afflicts stdlib module names, so it's inherently less interesting as an attack vector)

VanL:
We should probably add some common "typosquatting" names to the blocklist as well - urrlib, etc.

dstufft:
Well urllib is part of the stdlib so it will get included automatically with this PR. There is another PR that let us add a custom blacklist in the DB that we've pre-filled with previous typosquatting packages that we've taken ownership of.

PR #2410, opened 2017-09-18
updates to /security and /help, lead people to FAQ when project-name conflicts arise.

ewdurbin:


warehouse/templates/pages/security.html
```diff
@@ -32,11 +32,12 @@
 
       <br>
 
-      <p>Instead, please email Donald Stufft or Richard Jones directly, providing as much relevant information as possible.</p>
+      <p>Instead, please email Donald Stufft, Ernest W. Durbin III, or Richard Jones directly, providing as much relevant information as possible.</p>
```
dstufft: Doesn't need to block merging this, but we should probably reach out to Richard to ask him if he wants to still receive these emails or not since he's stepped back from PyPI.

Issue #2333, opened 2017-08-25
Include user name in 403 error message from file upload


di:
Currently when a user attempts to upload a file to a project for which they don't have proper permissions, they get the following error message:

> "You are not allowed to upload to `{project.name}`."

It might be helpful to the user here to include the user name they have authenticated with in the error message, e.g. for cases like in #2243. For example:

> "The user '`{request.user.username}`' is not allowed to upload to '`{project.name}`'."

alexmuller:
I think this issue can be closed :)

ewdurbin:
thanks @alexmuller

PR #2416, opened 2017-09-19
Include user name in 403 error message from file upload #2333

mbacchi:


ewdurbin:
resolves #2333

ewdurbin:
thanks @mbacchi

Issue #2284, opened 2017-08-07
Ignore duplicate files when uploading, instead of throwing error


rsyring:
Original complaint: https://gist.github.com/boppreh/ac7522b3a4ac46b4f6010eecddc57f21

> Uploading distributions to https://upload.pypi.org/legacy/
> Uploading package-1.2.3.zip
> HTTPError: 400 Client Error: File already exists. for url: https://upload.pypi.org/legacy/
> 
> It's the exact same file. You could add a check for it and ignore, but no, it had to be an error.

And Donald's email reply on Distutils:

> > It's the exact same file. You could add a check for it and ignore, but no, it had to be an error.
> 
> This seems like a reasonable feature request, please file an issue with Warehouse asking for it so I don’t forget it.

dstufft:
Just to be clear, this should only no-op on files with the exact same hash, if someone tries to upload a different file with the same name that should still get the error as it is today.

dstufft:
Fixed in https://github.com/pypa/warehouse/pull/2419.

PR #2419, opened 2017-09-19
Proposed fix for GH-2284

benjaminws:
Here is a very naive attempt to address #2284. I'm not sure this is the right direction for this fix, but really want to help out.

- https://github.com/pypa/warehouse/issues/2284
- Move duplication check to happen after the temp file is persisted, so a hash can be generated from that
- Simple function to query for existing file and check hash
- Update unit tests to generate hash from file content, not name
- Move file size check to happen after duplication check


warehouse/forklift/legacy.py
```diff
@@ -572,6 +572,14 @@ def _is_valid_dist_file(filename, filetype):
     return True
 
 
+def _is_duplicate_file(request, filename, sha256_hash):
+    """ Check to see if file already exists, and if it's content matches
+    """
+    requested_file = request.db.query(File).filter(File.filename == filename)
+    return (request.db.query(requested_file.exists()).scalar() and
```
dstufft: We can probably avoid issuing two queries by just loading the existing file from the database using something like:

```python
file_ = (
    request.db.query(File)
              .filter(File.filename == filename)
              .first()
)

return file_ is None or file_.sha256_digest == sha256_hash
```

If we want to minimize what data we pull from the DB, we could change the query to be:

```python
file_ = (
    request.db.query(File)
              .options(load_only("sha256_digest"))
              .filter(File.filename == filename)
              .first()
)
```

warehouse/forklift/legacy.py
```diff
@@ -870,8 +871,6 @@ def file_upload(request):
             for chunk in iter(
                     lambda: request.POST["content"].file.read(8096), b""):
                 file_size += len(chunk)
-                if file_size > file_size_limit:
```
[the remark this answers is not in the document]
benjaminws: I originally moved it because the test was failing with an error about file size, which I naively assumed was due to an order of operations thing. But now that I'm looking at it, the test fails because the size of the bytes buffer used is `legacy.MAX_FILESIZE + 1`. I'm assuming that was copied and pasted from the previous test (that checks file size). I'll shore up the test and move things back.

warehouse/forklift/legacy.py
benjaminws: Ooo, love removing redundant queries. Also, I was wondering if we'd want to check all the hashes like we do later in this code path., that makes lots of sense. Thanks for the feedback! Give me a bit and I'll fix this up.

dstufft: By the way, if you're not aware you can click the little red box on the right hand side of the screen to get taken to the debug toolbar, and you can see all the queries that got executed as part of a HTTP request. It records the last N HTTP requests so you can still view the ones for file upload, you just need to load the page in the browser and then navigate to the upload request in the debug toolbar.

I normally like to do that to try and look for queries that seem close enough to duplicates I might be able to share a single query for, and also when using strategies like ``load_only()`` to make sure I'm not accidently triggering extra queries because I forgot something in my ``load_only()`` (since SQLAlchemy will implicitly execute extra queries if you say, did the ``load_only("sha256_digest")`` from above, and then accessed ``file_.size``).

warehouse/forklift/legacy.py
```diff
@@ -572,6 +572,14 @@ def _is_valid_dist_file(filename, filetype):
     return True
 
 
+def _is_duplicate_file(request, filename, sha256_hash):
```
dstufft: One other thing, it'd probably be a tad bit nicer to have this take a ``Session()`` object instead of a ``Request()`` object. That would mean that you just pass in ``request.db`` instead of ``request``, but it then means that (A) this function can be tested independently of a request and (B) if we ever want to re-use this function in a context we don't have a request, we can. It's also just a bit more hygienic since the only reason we need the ``Request()`` object is to access the ``Session()`` object hanging off it at ``request.db``.

warehouse/forklift/legacy.py
benjaminws: That's an A+ tip, I completely forgot about the toolbar.  This will help tremendously.

warehouse/forklift/legacy.py
benjaminws: I felt bad passing around a `Request()` as it were, so this is the feedback I was looking for. Thanks again!

benjaminws:
Finally got some time to work on this. Think I covered the cases you pointed out.

warehouse/forklift/legacy.py
```diff
@@ -571,6 +571,22 @@ def _is_valid_dist_file(filename, filetype):
...
+    """ Check to see if file already exists, and if it's content matches
+    """
+
+    file_ = (
+        db_session.query(File)
+                  .filter(File.filename == filename)
+                  .first()
+    )
+
+    if file_:
```
dstufft: I'd prefer to explicitly check for ``None`` here, it shouldn't matter but just in case I think it makes sense to make this ``if file_ is not None:``. Doesn't hurt at least.

warehouse/forklift/legacy.py
```diff
@@ -571,6 +571,22 @@ def _is_valid_dist_file(filename, filetype):
...
+        return (file_.sha256_digest == hashes["sha256"] and
+                file_.md5_digest == hashes["md5"] and
+                file_.blake2_256_digest == hashes["blake2_256"])
+
+    return False
```
dstufft: Thinking this through, I wonder if this function shouldn't have 3 distinct return files, ``True``, ``False``, and ``None``. The return values would then mean:

* ``True``: This file is a duplicate and all further processing should halt.
* ``False``: This file exists, but it is not a duplicate.
* ``None``: This file does not exist.

Then further down we can do:

```python
is_duplicate =  _is_duplicate_file(request.db, filename, file_hashes):
if is_duplicate:
    raise _exc_with_message(HTTPBadRequest, "File already exists.")
elif is_duplicate is not None:
    return Response()
```

This means we get to skip a whole bunch of processing and uploading files to S3 etc for something that we know is a duplicate and we want to no-op anyways. It also prevents anything weird from happening like people lying to us about the contents of the file in order to change the data in the DB.

What do you think?

warehouse/forklift/legacy.py
benjaminws: A+, makes sense to be safe.

warehouse/forklift/legacy.py
benjaminws: I really like the idea of short-circuiting to prevent wasted calls/resources/etc. I'll write up a unit test for and see how it feels.

benjaminws: I wrote up some unit tests for this function and modified it to support the above use case. Also, following the logic you suggested, this will now bail early if the file exists but isn't exactly the same. I'm not sure if we want to return a different response in that scenario or not.

Thoughts?

dstufft: You mean it will bail early if the file exists and *is* exactly the same, right?

benjaminws: Right, so as it is written now the code will bail with an exception if a file exists and is exactly the same. 

This latest commit adds what you proposed above, which will bail early _without an error_ if the file exists, but not the same content. I could have totally misinterpreted your intent with that, so feel free to straighten me out.

dstufft: Oh right I see. Basically the goal is:

* If the file isn't the same file as we previously had, then bail early with an error. Since we already have a file we don't want to continue, but since this file doesn't match what was already uploaded we want to treat this as an error code.
* If the file is the same file as we previously had, just bail early with a success response. Since we've already had this file uploaded and it matches what was already uploaded we can just pretend we did the rest of the processing and it was successful.
* If we don't already have a file by this name, continue processing.

I might have switched around my conditional when I wrote the example code, but it sounds like we have the two first cases switched, so maybe we need to add a ``not`` to the return line of the function.

benjaminws: Ah right, I'm following now. Making a sandwich then I'll fix this up.

benjaminws: Feel like I have it covered. Take a look when you can.

benjaminws:
Happy with this now that I got the linter working locally, except that apparently, the config differs slightly than what is used on travis haha.

benjaminws:
Linting issues were on my end, ignore that previous comment.

dstufft:
Thanks!

ofek:
@dstufft All my tests are now failing https://travis-ci.org/ofek/hatch/jobs/278753948

Does Twine's --skip-existing do nothing now?

dstufft:
Looks like there's a bug in this that didn't get exposed by the tests. It appears that:

https://github.com/pypa/warehouse/blob/2cf4cfb35a1e7bebf9b0edcf2dfa80445b7e48c5/warehouse/forklift/legacy.py#L846-L855

Needs to get moved to immediately after:

https://github.com/pypa/warehouse/blob/2cf4cfb35a1e7bebf9b0edcf2dfa80445b7e48c5/warehouse/forklift/legacy.py#L920-L925

Otherwise all duplicate uploads are still going to fail because the filename still exists in the filename log. If nobody (@benjaminws? @ofek? Someone else?) gets to this I can try to get it fixed later today or tomorrow.

benjaminws:
@dstufft @ofek I'm working on this right now

PR #2420, opened 2017-09-19
Make classifiers 'sticky' on /search/ results page

### LINKED ISSUE #1415: Search with "sticky" trove classifier in UI
Currently warehouse works as follows:
1. Select trove classifier (e.g. Topic::Scientific/Engineering)
2. Type something in search field and press enter
3. Warehouse forgets the selected trove classifier

I would suggest to make the classifiers "stickier":
1. Select trove classifier (e.g. Topic::Scientific/Engineering)
2. Type something in search field and press enter
3. Warehouse shows results for the currently "active" classifiers
4. Add a button in the UI for removing all filters (currently, you can only remove them one by one). Also, maybe reset the classifier set if returning to the main page.

E.g. if you are interested only in scientific Python packages, it would be nice to have a more concentrated view of packages of interest more like CRAN etc., where search results are not mixed together with irrelevant stuff. Such portals can in principle be maintained by 3rd parties, but I think with small usability tweaks Pypi is good enough.

pv:
Change the template for /search/ so that the search form uses the
currently active classifiers also for the next search, instead of
clearing the classifier set.

The search forms on other pages are not changed, so the user can reset
the search by e.g. returning to the main page --- or, of course,
removing the classifiers one by one.

Closes #1415

I think the UX after this change is preferable to the one before. It's a bit of a hassle to activate the classifiers in the first place, and in the "before" UX your choice gets forgotten immediately when searching, whereas in the "after" you don't have to worry losing your selection if e.g. a word was mistyped (moreover, you can select the classifiers also before hitting the "Search" button). The active classifiers are shown with nice boxes with big `X` next to them so it's clear which ones are active, and it's also easy to remove them.

This is a template-only change. I'm not sure what to do with tests --- it seems the current set of functional tests is quite light, so I did not try to go adding a functional test for searching. So I tested this manually locally and it seemed to work fine.


ewdurbin:
@pv thanks for your contribution. adding screenshots in the future for previewing UI changes will generally speed up reviews.

Here are some!

Mobile width:

Initial search:
screen shot 2017-09-20 at 6 50 37 am
Add classifier:
screen shot 2017-09-20 at 6 50 25 am
Change search term:
screen shot 2017-09-20 at 6 50 31 am
✨ 

Desktop width:

Initial search:
screen shot 2017-09-20 at 6 50 43 am
Add classifier:
screen shot 2017-09-20 at 6 50 50 am
Change search term:
screen shot 2017-09-20 at 6 50 55 am
✨

di:
Thanks @pv!

PR #2429, opened 2017-09-21
Test if tabs exist before running JavaScript

alexmuller:
This JavaScript executes on every page so if there's no mobile button element on the page the call to `getComputedStyle` will cause an error. This change catches that case.


alex:
Code looks reasonable to me, does it make more sense to simply not run this function on pages it's not relevant on?

alexmuller:
Yep that's possible, I think the best approach would be:

- Add a new `block` just before the `</body>`
- Change this file to return a function
- On the packaging detail page, add a bit of JavaScript to the end of the body block to run the function

As far as I can tell it's just the packaging detail page which has these tabs. I'll have a go at what that solution could look like.

di:
Seems like this is still happening somehow.

di:
Urgh this is a PR not an issue...

PR #2435, opened 2017-09-22
Move file log check to after duplication check

benjaminws:
- Move `Filename` log check to after duplication check
- Fix up test


dstufft:
Thanks!

ofek:
Still failing, but now new error https://travis-ci.org/ofek/hatch/jobs/278753948

PR #2441, opened 2017-09-24
Classifier-based search should AND, not OR, in the search criteria

pv:
Make warehouse search show items that match all specified categories, rather than any of them.

This is how it works for the old pypi (see eg https://pypi.python.org/pypi?:action=browse&c=40&c=256), and presumably this is the intended behavior also for pypi.org

Screenshot before (shows entries matching either of the two classifiers):
before

Screenshot after (shows only entries matching both of the two classifiers):
after

### DIFF
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -446,7 +446,8 @@ class TestSearch:
             ),
         ]
         assert es_query.filter.calls == [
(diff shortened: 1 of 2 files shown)


ewdurbin:
Excellent! Thanks @pv!

Issue #2017, opened 2017-05-19
Dedicated mechanism for requesting a large file limit


dstufft:
Right now we limit uploads to 60MB in size, but some projects legitimately have a hard time fitting themselves into that size. With Warehouse we have the ability to increase that limit on a per project basis, however doing that currently involves manually running some SQL.

What would be great is a UI (logged in on the project admin page) that allows maintainers to request a limit increase that includes what limit they would like as well as a place for them to justify the increase. We could then have a similar page in the Warehouse admin that allows us to go through and hit approve or deny on each request.

brainwane:
Related: #3175.

pradyunsg:
Until we have this, would it make sense to do the following?

- Add an issue template for limit requests, requesting the relevant information.
- Add an FAQ entry when folks hit the limit, that points to that issue template.
- Update the error message from PyPI for upload limits to point to the FAQ entry.

brainwane:
@pradyunsg I agree.

PR #2470, opened 2017-10-06
Add ability for admins to set the upload limit on projects

jonparrott:
@dstufft this is a quick implementation to make it easier for you and @ewdurbin to field upload limit size requests until #2017 is done.

I couldn't get the one test I wrote to work, I'm pretty unfamiliar with testing pyramid apps, so a little help there would be awesome. I did test this feature manually locally and it appears to work.

Stacktrace from test output:

```python
_________________________________________________________________________________________________ TestProjectSetLimit.test_sets_integer_limit _________________________________________________________________________________________________

self = <tests.unit.admin.views.test_projects.TestProjectSetLimit object at 0x7f85b27ee828>, db_request = <pyramid.testing.DummyRequest object at 0x7f85b27f89e8>

    def test_sets_integer_limit(self, db_request):

        db_request.user = UserFactory.create()
        db_request.current_route_path = lambda: "/admin/projects/",
        db_request.session = pretend.stub(
            flash=pretend.call_recorder(lambda *a, **kw: None),
        )
        db_request.matchdict["project_name"] = project.normalized_name
        db_request.POST["upload_limit"] = "12345"

>       views.set_upload_limit(project, db_request)

tests/unit/admin/views/test_projects.py:360:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
warehouse/admin/views/projects.py:233: in set_upload_limit
    'admin.project.detail', project_name=project.normalized_name))
/usr/local/lib/python3.6/site-packages/pyramid/url.py:312: in route_path
    return self.route_url(route_name, *elements, **kw)
/usr/local/lib/python3.6/site-packages/pyramid/url.py:261: in route_url
    mapper = reg.getUtility(IRoutesMapper)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Registry global>, provided = <InterfaceClass pyramid.interfaces.IRoutesMapper>, name = ''

    def getUtility(self, provided, name=u''):
        utility = self.utilities.lookup((), provided, name)
        if utility is None:
>           raise ComponentLookupError(provided, name)
E           zope.interface.interfaces.ComponentLookupError: (<InterfaceClass pyramid.interfaces.IRoutesMapper>, '')

/usr/local/lib/python3.6/site-packages/zope/interface/registry.py:286: ComponentLookupError
```

### DIFF
--- a/tests/unit/admin/views/test_projects.py
+++ b/tests/unit/admin/views/test_projects.py
@@ -17,7 +17,6 @@ from pyramid.httpexceptions import HTTPBadRequest, HTTPMovedPermanently
 
 from warehouse.admin.views import projects as views
 
-from ....common.db.accounts import UserFactory
 from ....common.db.packaging import (
     JournalEntryFactory,
     ProjectFactory,
@@ -346,11 +345,11 @@ class TestProjectJournalsList:
 
 
 class TestProjectSetLimit:
-    def test_sets_integer_limit(self, db_request):
-        project = ProjectFactory.create()
+    def test_sets_limitwith_integer(self, db_request):
+        project = ProjectFactory.create(name="foo")
 
-        db_request.user = UserFactory.create()
-        db_request.current_route_path = lambda: "/admin/projects/",
+        db_request.route_path = pretend.call_recorder(
+            lambda *a, **kw: "/admin/projects/")
         db_request.session = pretend.stub(
             flash=pretend.call_recorder(lambda *a, **kw: None),
         )
@@ -361,8 +360,38 @@ class TestProjectSetLimit:
 
         assert db_request.session.flash.calls == [
             pretend.call(
-                "Successfully set the upload limit on 'foo' to '12345'",
+                "Successfully set the upload limit on 'foo' to 12345",
                 queue="success"),
         ]
 
         assert project.upload_limit == 12345
+
+    def test_sets_limit_with_none(self, db_request):
+        project = ProjectFactory.create(name="foo")
+        project.upload_limit = 12345
+
+        db_request.route_path = pretend.call_recorder(
+            lambda *a, **kw: "/admin/projects/")
+        db_request.session = pretend.stub(
+            flash=pretend.call_recorder(lambda *a, **kw: None),
+        )
+        db_request.matchdict["project_name"] = project.normalized_name
+
+        views.set_upload_limit(project, db_request)
+
+        assert db_request.session.flash.calls == [
+            pretend.call(
+                "Successfully set the upload limit on 'foo' to None",
+                queue="success"),
+        ]
+
+        assert project.upload_limit is None
+
+    def test_sets_limit_with_bad_value(self, db_request):
+        project = ProjectFactory.create(name="foo")
+
+        db_request.matchdict["project_name"] = project.normalized_name
+        db_request.POST["upload_limit"] = "meep"
+
+        with pytest.raises(HTTPBadRequest):
+            views.set_upload_limit(project, db_request)
--- a/tests/unit/admin/views/test_projects.py
+++ b/tests/unit/admin/views/test_projects.py
@@ -17,6 +17,7 @@ from pyramid.httpexceptions import HTTPBadRequest, HTTPMovedPermanently
 
 from warehouse.admin.views import projects as views
 
+from ....common.db.accounts import UserFactory
 from ....common.db.packaging import (
     JournalEntryFactory,
     ProjectFactory,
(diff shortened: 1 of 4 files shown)


warehouse/admin/views/projects.py
```diff
@@ -201,3 +202,32 @@ def journals_list(project, request):
...
+def set_upload_limit(project, request):
+    upload_limit = request.POST.get("upload_limit", "")
+
+    # If the upload limit is an empty string or othrwise falsy, just set the
+    # limit to None, indicating the default limit.
+    if not upload_limit:
+        upload_limit = None
+
+    # Update the project's upload limit.
+    project.upload_limit = int(upload_limit)
```
dstufft: This isn't going to work if ``upload_limit`` is ``None``. Maybe do

```python

if not upload_limit:
    project.upload_limit = None
else:
    try:
        project.upload_limit = int(upload_limit)
    except ValueError:
        raise HTTPBadRequest(f"Invalid value for upload_limit: {upload_limit}) from None
```

That should cover all of the cases I think?

tests/unit/admin/views/test_projects.py
```diff
@@ -342,3 +343,26 @@ def test_non_normalized_name(self, db_request):
...
         with pytest.raises(HTTPMovedPermanently):
             views.journals_list(project, db_request)
+
+
+class TestProjectSetLimit:
+    def test_sets_integer_limit(self, db_request):
+        project = ProjectFactory.create()
+
+        db_request.user = UserFactory.create()
+        db_request.current_route_path = lambda: "/admin/projects/",
```
dstufft: You'll want to add a similar thing for ``db_request.route_path()`` to fix the testing error. I normally use a ``pretend.call_recorder()`` so I can make sure that it is called with the correct arguments.

jonparrott: Ah I get now. Thanks. Done and added two additional tests for coverage.

warehouse/admin/views/projects.py
jonparrott: Yep, done - without this SQLAlchemy will raise an error, but this is more explicit and raises a proper HTTP error.

dstufft: Yea, SQLAlchemy errors get bubbled up into Sentry and alerted whereas HTTP errors don't. So when there is an expected error we try to catch it and turn it into HTTP errors to keep sentry cleaner. Not really a super big deal here because I doubt the 2 or 3 admins are likely to start throwing a bunch of random garbage into fields, but whatever!

jonparrott: Better safe than alerted.

jonparrott:
Thanks for reviewing, @dstufft, let me know if there's anything else I need to do to get this merged.

dstufft:
Part of me is wondering if it would make sense to deviate from the database schema here and have the input be in MB instead of bytes. It's in bytes in the DB schema because that was easier for the code, but it's kind of crummy for humans to have to manually multiple the value by ``1024 * 1024`` to put into this form. I think every request we've gotten has been in terms of MB so that's typically what we think of... although it might be nice to allow input like ``100MB`` and convert that to bytes, but that's a bit more effort then just mandating it be in MB.

It also might be a good idea to flash an error and redirect back to the detail page if you try to set a limit less then the current default minimum.

jonparrott:
@dstufft all reasonable concerns, but considering this will be only used by admins and largely superseded by #2017 do you feel like it's okay as is for now?

I'm also happy to send follow-up PRs for those items to keep this one small.

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -132,4 +145,30 @@ <h3 class="box-title">Blacklist Project</h3>
...
+        <button class="btn btn-box-tool" data-widget="collapse" data-toggle="tooltip" title="Expand"><i class="fa fa-plus"></i></button>
+      </div>
+    </div>
+
+    <form method="POST" action="{{ request.route_path('admin.project.set_upload_limit', project_name=project.normalized_name) }}">
+      <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+      <div class="box-body">
+        <div class="form-group col-sm-12">
+          <label for="uploadLimit">Upload limit (in bytes)</label>
+          <input type="text" name="upload_limit" class="form-control" id="uploadLimit" rows="3" value="{{ project.upload_limit | default('', True)}}">
```
dstufft: This would be a bit nicer as ``<input type="number" min={{ THE MINMUM SIZE }} ...>`` instead of as a text field. It might also make sense to include ``step="10"`` or something, which i *think* means that the arrows that browsers render this with will increment/decrement this by 10, but you can still type in whatever step you want directly into the box.

dstufft:
I don't think it's superseded by #2017, because it's useful for Admins to be able to set this manually in either case. But I'm OK merging this as is if you can send a follow up PR with the above stuff.

dstufft:
Oh and talked with @ewdurbin and Mark (the other two folks who do admin stuff) and they both agreed that just accepting input in MB is fine, and just hiding the fact it's being converted to bytes.

dstufft:
Yea, basically just getting Warehouse to do the ``1024 * 1024`` math instead of expecting people to do it :D. We'd also convert the other way to turn bytes into MB for rendering in the form.

PR #2473, opened 2017-10-06
UI improvements to the admin interface for setting project upload limits

### LINKED ISSUE #2471: Admin package upload size text control should use MB instead of bytes
Context: https://github.com/pypa/warehouse/pull/2470#issuecomment-334852617

Decision:

> The input in should be in MB and should just be converted to/from bytes in the backend.

Also:

> It also might be a good idea to flash an error and redirect back to the detail page if you try to set a limit less then the current default minimum.

Additionally:

> [The form field] would be a bit nicer as <input type="number" min={{ THE MINMUM SIZE }} ...> instead of as a text field. It might also make sense to include step="10" or something

jonparrott:
* Uses a number field instead of a text field.
* The text field now uses megabytes instead of bytes.
* The detail page now shows the value in megabytes as well as the full value in bytes.
* The detail page now shows "Default" as well as indicating what the default limit is in MB instead of just displaying "None".
* The form field should prevent numbers lower than the default, but the backend also now gives an invalid request error if the value is lower than the default.
* Simplified the flash message as the full value will be displayed on the detail page.

Unfortunately, I couldn't use "step" in the form field as it would have limited the input to multiples of the step value.

Fixes #2471

### DIFF
--- a/warehouse/admin/templates/admin/projects/detail.html
+++ b/warehouse/admin/templates/admin/projects/detail.html
@@ -54,7 +54,13 @@
                 </tr>
                 <tr>
                   <td>Upload limit</td>
-                  <td>{{ project.upload_limit }}</td>
+                  <td>
+                    {% if project.upload_limit %}
+                      {{ project.upload_limit / ONE_MB }} MB ({{ "{:,.0f}".format(project.upload_limit) }} bytes)
+                    {% else %}
+                      Default ({{ MAX_FILESIZE / ONE_MB }} MB)
+                    {% endif %}
+                  </td>
                 </tr>
               </table>
             </div>
@@ -159,8 +165,13 @@
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
       <div class="box-body">
         <div class="form-group col-sm-12">
-          <label for="uploadLimit">Upload limit (in bytes)</label>
-          <input type="text" name="upload_limit" class="form-control" id="uploadLimit" rows="3" value="{{ project.upload_limit | default('', True)}}">
+          <label for="uploadLimit">Upload limit (in Megabytes)</label>
+          {% if project.upload_limit %}
+            {% set upload_limit_value = project.upload_limit / ONE_MB %}
+          {% else %}
+            {% set upload_limit_value = '' %}
+          {% endif %}
+          <input type="number" name="upload_limit" class="form-control" id="uploadLimit" min="{{ MAX_FILESIZE / ONE_MB }}" value="{{ upload_limit_value }}">
         </div>
       </div>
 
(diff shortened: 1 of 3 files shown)


dstufft:
> Unfortunately, I couldn't use "step" in the form field as it would have limited the input to multiples of the step value.

A bummer, I was hoping it would just control the arrows and not be part of the client side validation. Ah well.

jonparrott:
> A bummer, I was hoping it would just control the arrows and not be part of the client side validation. Ah well.

I tried my best. :)

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -54,7 +54,13 @@ <h3 class="project-name text-center">{{ project.name }}</h3>
                 </tr>
                 <tr>
                   <td>Upload limit</td>
-                  <td>{{ project.upload_limit }}</td>
+                  <td>
+                    {% if project.upload_limit %}
+                      {{ project.upload_limit / ONE_MB }} MB ({{ "{:,.0f}".format(project.upload_limit) }} bytes)
```
dstufft: To format the bytes, I think it'd be better to use ``project.upload_limit|format_number()``, it's more comprehensible then the terrible Python mini language :D

I'm torn on the display of the MB itself, we have ``filesizeformat(binary=True)`` which will do the same thing, except work nicer if we go > 1GB... but I don't think that we're likely to do that so I don't know how much it matters. Since I'm already asking for you to fix the number formatting for the bytes, I think unless you feel strongly about it using ``{{ project.upload_limit|filesizeformat(binary=True) }}`` would be my preference.

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -159,8 +165,13 @@ <h3 class="box-title">Set upload limit</h3>
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
       <div class="box-body">
         <div class="form-group col-sm-12">
-          <label for="uploadLimit">Upload limit (in bytes)</label>
-          <input type="text" name="upload_limit" class="form-control" id="uploadLimit" rows="3" value="{{ project.upload_limit | default('', True)}}">
+          <label for="uploadLimit">Upload limit (in Megabytes)</label>
+          {% if project.upload_limit %}
+            {% set upload_limit_value = project.upload_limit / ONE_MB %}
```
dstufft: Unlike the display issue, this needs to use the same logic as the python code to go from bytes to MB, so this makes sense to keep this regardless of what happens above.

warehouse/admin/templates/admin/projects/detail.html
jonparrott: Done.

warehouse/admin/templates/admin/projects/detail.html
jonparrott: Noted.

dstufft:
Great, thanks a lot!

jonparrott:
Happy to help

PR #2535, opened 2017-10-26
Switch the build process to use Debian w/ multi stage builds

### LINKED ISSUE #2523: Alpine Linux segfaults in dev environment
Since #2474, I am having an issue in local dev in which `hupper` reloads or crashes on every request, with either something like:
```
Killing server with PID 7.
Starting monitor for PID 25.
```
or:
```
Server with PID 43 exited with code -11.
```
I bisected to determine that this is in fact due to a change in #2474, and it specifically seems like the cause is the switch to `python:3.6.2-alpine3.6`, because changing the base image back to `python:3.6.1-slim` resolves the problem.

~I poked around with `hupper` a bit, and it seems like the cause is that something is writing to the worker's pipe on every request with no value, and thus `self.worker.pipe.recv` is `None` and this triggers a reload for `hupper`. It also doesn't seem to be coming from `hupper` itself.~

I also tried upgrading to `python:3.6.3-alpine3.6` (we should probably do this anyways since it was missed in #2508), but the problem remains.

@dstufft, are you able to reproduce this? My next steps are going to try to minimally reproduce this to share the issue with the `hupper` devs, unless you have any insight.

dstufft:
This switches us back to using a Debian based Docker image and also refactors the build support to further take advantage of multi stage builds. The resulting image is 394MB large (compared to 261MB with Alpine) when built with DEVEL=yes.

This also restructures things a bit so that instead of ``/app/`` and ``/usr/local`` we move everything into ``/opt/warehouse``.

Fixes #2523 (I hope?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 # steps appended onto the end.
 FROM node:6.11.1 as static
 
-WORKDIR /app/
+WORKDIR /opt/warehouse/src/
 
 # The list of C packages we need are almost never going to change, so installing
 # them first, right off the bat lets us cache that and having node.js level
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -4,102 +4,152 @@
...
 
-# Install System level Warehouse requirements, this is done before everything
-# else because these are rarely ever going to change.
+# This is a work around because otherwise libpq-dev bombs out trying
+# to create symlinks to these directories.
 RUN set -x \
-    && apk --no-cache add libpq libxml2 libxslt \
-        $(if [ "$DEVEL" = "yes" ]; then echo 'bash libjpeg postgresql-client'; fi)
+    && mkdir -p /usr/share/man/man1 \
+    && mkdir -p /usr/share/man/man7
```
di: I think this can be removed? Pretty sure it's only `postgresql-client` that needs these to exist.

PR #2539, opened 2017-10-27
Delete Project Admin UI

di:
This PR adds the ability for admins to to delete a project.

### DIFF
--- a/warehouse/admin/utils.py
+++ b/warehouse/admin/utils.py
@@ -1,0 +1,73 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from packaging.utils import canonicalize_name
+from pyramid.httpexceptions import HTTPSeeOther
+
+from warehouse.packaging.models import (
+    Project, Release, Dependency, File, Role, JournalEntry, release_classifiers
+)
+
+
+def confirm_project(project, request):
+    confirm = request.POST.get("confirm")
+    project_name = project.normalized_name
+    if not confirm:
+        request.session.flash(
+            "Must confirm the request.",
+            queue="error",
+        )
+        raise HTTPSeeOther(
+            request.route_path(
+                'admin.project.detail',
+                project_name=project_name
+            )
+        )
+    if canonicalize_name(confirm) != project.normalized_name:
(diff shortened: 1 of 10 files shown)


warehouse/admin/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    request.db.query(Role).filter(Role.project == project).delete()
+    request.db.query(File).filter(File.name == project.name).delete()
+    (request.db.query(Dependency).filter(Dependency.name == project.name)
+               .delete())
+    (request.db.execute(release_classifiers.delete()
+                        .where(release_classifiers.c.name ==
+                               project.name)))
+    request.db.query(Release).filter(Release.name == project.name).delete()
+    request.db.query(Project).filter(Project.name == project.name).delete()
+    request.db.add(
```
ewdurbin: are we repeating lines 52-68 verbatim in lines _nice_-85?

warehouse/admin/views/blacklist.py
```diff
@@ -166,37 +166,17 @@ def add_blacklist(request):
         )
     )
 
-    # Go through and delete anything that we need to delete so that our
-    # blacklist actually blocks things and isn't ignored (since the blacklist
-    # only takes effect on new project registration).
-    # TODO: This should be in a generic function somewhere instead of putting
```
ewdurbin: One nice to have would be thorough tests around the newly factored `remove_project` function. Right now in both places we call it (`add_blacklist`, and new `delete_project`) our tests are fairly naive.

warehouse/admin/utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+from packaging.utils import canonicalize_name
+from pyramid.httpexceptions import HTTPSeeOther
+
+from warehouse.packaging.models import (
+    Project, Release, Dependency, File, Role, JournalEntry, release_classifiers
+)
+
+
+def confirm_project(project, request):
```
ewdurbin: I think this could be generalized a bit more to cover the other use case at `https://github.com/pypa/warehouse/pull/2539/files#diff-2246831731f72e538a306ad3a32ac674L146` if we set the redirect to `request.current_route_path()` and allowed supplying a custom `message` for the flash at line 26.

di: I was originally hoping to do that, but the blacklist confirmation is slightly different:

* It doesn't require a `project` to exist, it just needs a project name;
* The blacklist form `POST`s to the same route that the form lives at, so it can redirect back to `current_route_path` if the confirmation fails. The delete form, however, just lives on the project detail page, `POST`s to the `admin.project.delete` route, and needs to redirect back to the project detail page instead, since `current_route_path` wouldn't correspond to a page that we could `GET`. (I hope that makes sense)

I left this as a separate utility function so that it could be more easily tested, and potentially reused if we need to confirm other actions for an existing project, but I think it's different enough from the blacklist confirmation that we should let them be separate.

di:
Hey @ewdurbin, do you mind taking a second look at this?

ewdurbin:
@di sorry! I didn't mean to block this so long, thanks for the bump.

PR #2542, opened 2017-10-29
Added password strength gauge in registration form.

### LINKED ISSUE #6: Validate Complexity of Passwords
Ensure that user passwords always match some level of password complexity/strength.

yeraydiazdiaz:
Fixes #6

Added vendor JS gulp step to include zxcvbn.js and include it in the register template.
Added password strength gauge in registration form.

[](https://postimg.org/image/71yn2g8o2j/)


yeraydiazdiaz:
In order to follow the [zxcvbn recommendations](https://github.com/dropbox/zxcvbn#browserify--webpack) I made some changes on the frontend pipeline, specifically creating a `vendor` subdirectory in `warehouse/static/js`, adding a gulp step to simply copy them to a `js/vendor` subdirectory in `dist` and finally ignoring it in the linting. Seemed like the best approach to achieve the recommendations although we're now adding a specific version of zxcvbn.js on the repo that will need to be manually updated, I couldn't think of a better approach so suggestions are welcome.

The UI is of course up for discussion and there's no validation step, it's simply a graphical representation of the result of zxcvbn score, any comments and suggestions are appreciated.

di:
@yeraydiazdiaz This is great, thanks!

I think vendoring the library is definitely the right approach. [We do something similar for the Admin UI](https://github.com/pypa/warehouse/tree/master/warehouse/admin/static/js).

> The UI is of course up for discussion and there's no validation step, it's simply a graphical representation of the result of zxcvbn score, any comments and suggestions are appreciated.

The UI looks great to me but @nlhkabu will have to approve. We do validate this in the backend (the score must be >= 2) so it would be reasonable to disable form submission if the password doesn't meet this score.

yeraydiazdiaz:
Cool, I didn't realise we were doing it already. I'll add the disabling of the submit button for low scores to this PR and wait for feedback on the UI.

yeraydiazdiaz:
I've updated this to include a slightly more advanced validation in the frontend side, enabling the submit button only if all required fields have values, the password score is >= 2 and the passwords match. Here's a screengrab of the flow:

[](https://postimg.org/image/1puye9di0r/)

dstufft:
The gif of this change looks really good to me, I'll try to get it reviewed this week and if @nlhkabu hasn't had a chance to review it before then, I feel comfortable merging it and tagging it for her to do a post-commit review.

di:
@yeraydiazdiaz After taking a second look at this, I think that I wasn't clear enough when I said " it would be reasonable to disable form submission".

What I meant was that attempting to submit the form when the password doesn't meet these requirements should not submit the form, and instead show a tooltip indicating that the password is insufficient, similar to how it works for other fields:

screen shot 2017-11-07 at 4 47 15 pm

By disabling the submit button, the form no longer gives this feedback when a non-password field is invalid, which is not a great UX.

yeraydiazdiaz:
@di I agree it's not great UX. I did consider adding information via the tooltip, however it is actually implemented using the [`:required` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:required) which is browser specific and could not find a way to hook into it.

I do think disabling the submit button if the required fields are missing helps the UX though, if you agree maybe the easiest way forward is to show a message and/or color code the guage to show that the password is not secure enough?

yeraydiazdiaz:
I've repurposed this PR implementing the behaviour [described by @di](https://github.com/pypa/warehouse/pull/2542#issuecomment-342650529). I had to build a custom tooltip system in a similar fashion to that of Chrome but I think the flow is now quite nice. Here's the result:

[](https://postimg.org/image/yt6ndhi6p/)

yeraydiazdiaz:
Huh, didn't realise the above gif shows a backend validation error on the password strength while the frontend allows submission. I must have had local changes in the backend. Here's the actual flow:

[](https://postimg.org/image/w0m9crmox/)

Currently tooltips are only cleared when attempting another form submission. I am wondering if they should disappear after a timeout, an event (maybe when focusing on the infringing or any field in the form) or both?

It'd be good to introduce some testing around frontend, not sure if there's plans for that.

brainwane:
Thanks for your PR @yeraydiazdiaz and sorry for the slow response!

The folks working on Warehouse have gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and have kicked off work towards [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap).

Issue #6 is [in our "shut down Legacy PyPI" milestone](https://github.com/pypa/warehouse/milestone/7) which we might not get to during the next few months of grant-funded effort, but we are working to review open PRs while we have multiple maintainers concentrating on improving Warehouse right now. So if you fix the merge conflict on this PR and let us know that you've done so by leaving a comment, the next few months are a good time to get it re-reviewed.

Thanks and sorry again for the wait.

yeraydiazdiaz:
Hi @brainwane, no worries, I've rebased and updated the PR, let me know if it needs any changes.

brainwane:
@lgh2 the next time you and I chat in IRC, let's talk about you reviewing this PR. Thanks!

lgh2:
Sounds good!

lgh2:
Hi @yeraydiazdiaz -- thank you very much for the PR! 

I am using Chrome Version 64.0.3282.119 (Official Build) beta (64-bit) on Linux Mint 18.1 (Ubuntu Xenial 16.04)

I tested it and when I entered "12345678" as a password, the tooltip appeared up at the top of the screen, and I had to scroll up in order to see it: 

wandering-tooltip

Since I block ads pretty aggressively in Chrome, Sumana suggested that some of my plugins might be causing the issue, so I created a new profile in Chrome for testing, and had the same problem: 

wandering-tooltip-2

The form did not submit when the  "password strength" bar was red or orange, but would submit when it was yellow or green.  I think moving the tooltip to a more visible place might help users see why the form is not submitting.

yeraydiazdiaz:
Hi @lgh2, thanks for looking into the PR. Turns out there was a bug where the tooltip was misplaced if the user had scrolled in the page, I've fixed it now.

nlhkabu:
hi @yeraydiazdiaz - thanks for this PR :grin: 

A couple of comments on making the HTML and CSS consistent with the rest of the codebase:

I think the strength indicator justifies it's own block in the SCSS.  I suggest we move `.pw-strength` into a new file e.g. `warehouse/static/sass/blocks/_password-strength.scss`.  I would also prefer that we rename it to `password-strength` to be more explicit.  The guage should also be an `element` child of the block - so `password-strength__guage`

For the percentages, I'd consider using a modifier, e.g. `password-strength__guage--0`, `password-strength__guage--1`, etc.

For more information on BEM (the naming convention we use), you can refer to our docs here: https://warehouse.readthedocs.io/development/frontend

For the tooltip, we should instead use the tooltips already in the codebase, located here: https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/blocks/_tooltip.scss

You can see an example of their use here: https://github.com/pypa/warehouse/blob/master/warehouse/templates/packaging/detail.html#L78

I hope this helps!  Let me know if any of this feedback is unclear.  Again, thanks for your work on this, I'm excited to see it moving forward again!

ThomasWaldmann:
I still remember when I wanted to contribute some pw strength measurement function to `passlib` and digged into theory and misc. implementations for this.

I finally discarded the whole idea as futile:

- you would need a lot of world knowledge to do a realistic pw strength measurement (like dictionaries of all languages, slang, common sayings, ...) for an international audience. for an "english minded" pw strength meter, some word(s) from another language might look really strong although it is just a dictionary word, a popular saying, whatever.

- besides that, a password generator might generate a random password that does not have uppercase, digits, or special chars inside (so does not conform to your rules). Do you want to reject it and force the user to generate another random one?

- one needs to be aware that by enforcing specific criteria (e.g. at least 1 number, at least 1 uppercase) one is in fact reducing the size of the valid password space, so an attacker has less values to try when brute forcing.

So, my conclusion back then was that it is impossible to tell whether a password is good.
You can only tell when it is very bad. Stuff like 1234, aaaa, q1q1q1q1q1q1.

So, add a hint below the pw field reminding people to choose a good one.
Check len(p) and len(set(p)) to reject the very bad ones.

di:
@ThomasWaldmann Not sure if you took a look at the code in this PR, but we're using [zxcvbn](https://github.com/dropbox/zxcvbn), a well-known password strength estimator which is "good enough" for our purposes.

We also have the following message below the password field (which is in the screenshots above):

>  Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.

ThomasWaldmann:
@di "well known" doesn't mean much. Does it address the concerns I raised?

Just a few examples from a quick glance at the code:
- https://github.com/dropbox/zxcvbn/blob/master/data-scripts/build_keyboard_adjacency_graphs.py has US and dvorak keyboard. quite limited world knowledge.
- https://github.com/dropbox/zxcvbn/tree/master/data guess you can see there what I meant with an "english minded pw strength checker".

brainwane:
@ThomasWaldmann Yeah, I poked around the issue tracker for zxcvbn and saw that you're not the only one who wishes it were more multilingual. Still, I agree with @di that adding it to Warehouse and stopping users from registering with super-awful-in-English passwords is a good step. We can see whether we can improve from there.

yeraydiazdiaz:
@nlhkabu I've implemented your suggestions, in particular using the existing tooltip system which I somehow managed completely miss before 😅

di:
@yeraydiazdiaz This looks great! One small change before merging:

The "Confirm Password" input is starting to look a little out of place with everything between it and the "Password" input. Could you change the class for it's label from `sr-only` to `form-group__label` so it is displayed?

warehouse/static/sass/blocks/_password-strength.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  <div class="form-group">
+    <p class="form-group__help-text">
+      Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
+      Password strength: <span class="password-strength"><span class="password-strength__guage"></span></span>
+    </p>
+  </div>
+*/
+
+.password-strength {
```
nlhkabu: perfect :)

warehouse/static/sass/blocks/_password-strength.scss
```diff
@@ -0,0 +1,62 @@
...
+/*
+  A password strength indicator:
+
+  <div class="form-group">
+    <p class="form-group__help-text">
+      Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
+      Password strength: <span class="password-strength"><span class="password-strength__guage"></span></span>
+    </p>
+  </div>
+*/
```
nlhkabu: We should not document the containing block here.  The following is enough:

```
<span class="password-strength">
  <span class="password-strength__guage"></span>
</span>
```

warehouse/templates/accounts/register.html
```diff
@@ -77,9 +77,12 @@ <h1 class="page-title">Create an account on PyPI</h1>
...
           </div>
-          {{ form.new_password(placeholder="Select a password", required="required", class_="form-group__input", autocomplete="new-password") }}
+          <div>
+            {{ form.new_password(placeholder="Select a password", required="required", class_="form-group__input", autocomplete="new-password") }}
+          </div>
           <p class="form-group__help-text">
             Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
+            Password strength: <span class="password-strength"><span class="password-strength__guage"></span></span>
           </p>
           {% if form.new_password.errors %}
```
nlhkabu: Can we split this out into 2 `<p>` s?

```
<p class="form-group__help-text">
  Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
</p>
<p class="form-group__help-text">
  <strong>Password strength</strong>
  <span class="password-strength">
    <span class="password-strength__guage"></span>
  </span>
</p>

```

nlhkabu:
Hi @yeraydiazdiaz and @di, I've just added a couple of nitpick comments.

One thing we're not considering here is the experience of users who are using screen readers.
Would it be possible to do something like this:

```
<span class="password-strength">
  <span class="password-strength__guage">
    <span class="sr-only">weak</span><!-- text is updated as password becomes stronger -->
  </span>
</span>
```

If this is too much to include in this PR, we can merge without and open a separate ticket and PR later?

warehouse/templates/accounts/register.html
```diff
@@ -77,9 +77,12 @@ <h1 class="page-title">Create an account on PyPI</h1>
...
             </label>
           </div>
-          {{ form.new_password(placeholder="Select a password", required="required", class_="form-group__input", autocomplete="new-password") }}
+          <div>
+            {{ form.new_password(placeholder="Select a password", required="required", class_="form-group__input", autocomplete="new-password") }}
+          </div>
           <p class="form-group__help-text">
             Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
+            Password strength: <span class="password-strength"><span class="password-strength__guage"></span></span>
           </p>
```
nlhkabu: Can we split this out into 2 `<p>` s?

```
<p class="form-group__help-text">
  Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters. Avoid common words or repetition.
</p>
<p class="form-group__help-text">
  <strong>Password strength</strong>
  <span class="password-strength">
    <span class="password-strength__guage"></span>
  </span>
</p>

```

di:
@nlhkabu We are setting the `aria-label` on the parent of the input at the same time as when we show the tooltip, does that suffice or do we need to have the `<span class="sr-only">` bit as well (or instead)?

nlhkabu:
@di, I think we should do both.  The experience for someone using a screen reader will be to hear "password strength" and then ... nothing.  Adding something here to indicate the strength before they submit the form would be useful :)

di:
Thanks @yeraydiazdiaz!

brainwane:
Thank you so much, @yeraydiazdiaz! Let us know if you're interested in an idea for the next issue you could work on. :)

yeraydiazdiaz:
No worries at all, I tried out pypi.org last night and it was an excellent experience, really happy to contribute, just point me to an issue 👍

nlhkabu:
hi @yeraydiazdiaz - I was actually thinking about this this morning - it would be great if we could add the password strength indicator onto 2 other pages:

1. the page where you choose a new password, during the 'reset password' workflow
2. the account settings page (where a logged in user can change their password)

What do you think?

PR #2574, opened 2017-11-12
simple/<package_name> sort versions correctly

### LINKED ISSUE #2217: Improve sorting on simple page
I'd like to submit a patch for this but I have a few questions :)

First I'll describe what I'd like to do...

## sort by version number

See https://pypi.org/simple/pre-commit/

You'll notice that `0.10.0` erroneously sorts *before* `0.2.0` (I'd like to fix this)

## investigation

I've found the code which does this sorting [here](https://github.com/pypa/warehouse/blob/3bdfe5a89cc9a922ee97304c98384c24822a09ee/warehouse/legacy/api/simple.py#L76-L89)

This seems to just sort by filename, but by inspecting and viewing [this page](https://pypi.org/simple/pre-commit-mirror-maker/) I notice it seems to ignore `_` vs. `-` (which is good, that's what I want to continue to happen but I'm just not seeing it from the code!)

## other questions

The `File` objects which come back from the database contain a `.version` attribute that I'd like to use to participate in sorting, my main question is: **Can I depend on this version to be a valid [PEP440](https://www.python.org/dev/peps/pep-0440/) version and use something like `pkg_resources.parse_version`?**

I'd basically like to replicate something close to the sorting which @chriskuehl's [dumb-pypi](https://github.com/chriskuehl/dumb-pypi) does [here](https://github.com/chriskuehl/dumb-pypi/blob/fd0f93fc2e82cbd9bae41b3c60c5f006b2319c60/dumb_pypi/main.py#L77-L91).

Thanks in advance :)

---

**Good First Issue**: This issue is good for first time contributors. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

cryvate:
It now sorts based on name, pkg_resource parsed version and reversed filename (keep extensions together). Fixes #2217.

Is there a way to extract the extension of the file cleanly? And the remaining metadata for e.g. wheels? Then we can sort on that instead of the 'hacky' reversed filename.


cryvate:
I tested this with ``aws_kpl_deagg`` which has versions 1.0.1 to 1.0.10 inclusive in the server I got from following the development guide.

cryvate:
I will add unit tests if this looks good, I am just pondering how to deal with the remainder of the filename properly (the extension etc.).

warehouse/legacy/api/simple.py
```diff
@@ -47,6 +49,16 @@ def simple_index(request):
...
 
 
+def get_sort_key_file(file):
+    return (
+        file.name,
+        # need to parse so 1, 10, 2 compare correctly
+        pkg_resources.parse_version(file.version),
+        # get same extensions together
+        file.filename[::-1]
+    )
```
di: I don't think this needs to sort on `file.name`, as this is going to be the same for all  files within a project.

Instead of sorting on the reversed filename as a secondary sort, you can use `file.packagetype` instead.

At this point, the key function should be short enough that you can just do it inline, e.g.:

```python
    files = sorted(
        request.db.query(File)
        ...
        .all(),
        key=lambda f: (pkg_resources.parse_version(f.version), f.packagetype)
    )
```

di:
@cryvate Apologies that this took so long to review! If you are still able to work on this, and get it merged, we'd be able to send you some PyPI stickers as part of [Ernest's PyPI Sticker Challenge](https://twitter.com/EWDurbin/status/951436735040376832).

cryvate:
@di I made the suggested change, and tried to update one of the tests. I cannot run the tests locally at the moment, so not sure it is actually testing what I've fixed. I'd love some stickers!

di:
@cryvate Thanks! I think instead of modifying an existing test, it'd be better to add a new test here that would have failed before you made this change.

I'm thinking something like:

```python
def test_file_sorting(self, db_request):
    releases = [
        ReleaseFactory.create(project=project)
        for version in ["1.0", "2.0", "10.0"]
    ]
    for release in releases:
        FileFactory.create(
            release=release,
            packagetype="sdist",
            filename="{}-{}.tar.gz".format(project.name, release.version),
        )
        FileFactory.create(
            release=release,
            packagetype="bdist_wheel",
            filename="{}-{}.whl".format(project.name, release.version),
        )
    files = simple.simple_detail(project, db_request)["files"]
    
    # TODO: make an assertion based on the order of `files`
```
I think the easiest way to make the assertion for that test would be to just look at the order of the filenames for each file.

Also, if you want to debug what's wrong with testing environment, feel free to ask questions here. It's a lot faster to be able to run tests locally!

cryvate:
@di work on this is stalled until the 5th as I do not have internet at the moment, and the testing infrastructure does not seem very happy with that!

di:
Hi @cryvate, hope you're back online! Are you able to run the tests locally yet?

cryvate:
I did have a stab at it, but I couldn't get it to work. What's the best way to get feedback on what's going wrong, pypa IRC?

di:
@cryvate Sure, or `pypa-dev` would maybe be more appropriate.

brainwane:
@cryvate I wanted to check in -- if you are working on this issue and have questions, please feel free to ask them here, in [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or on the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev). And then we'd love for you to add any troubleshooting tips we find together in a PR to improve https://warehouse.readthedocs.io/development/getting-started/#running-tests-and-linters .

brainwane:
I've just pinged @cryvate via email.

brainwane:
@cryvate letting you know about my comment https://github.com/pypa/warehouse/issues/2217#issuecomment-371545217 . We're very happy to help you with this issue or with another issue if you're up for it, now or when you have time to help with Warehouse. Thank you!

warehouse/legacy/api/simple.py
```diff
@@ -84,8 +86,8 @@ def simple_detail(project, request):
                           .with_entities(Release.version)
             )
         )
-        .order_by(File.filename)
-        .all()
+        .all(),
+        key=lambda f: (pkg_resources.parse_version(f.version), f.packagetype)
```
cryvate: So there is a potential for a problem here, if f.packagetype is None problems occur (is that valid? I can make it happen in the tests but not sure it can happen in a real deployment)

cryvate: And similarly, if f.version is None? Looking at recent changes this happens relatively regularly.

di: `File.version` and `File.packagetype` will not be `None` here, and if they are, we've got bigger problems.

Also, I think I would prefer if we used `packaging.version.parse` here instead, from the `packaging` package.

cryvate: We should be good to go then

brainwane:
@cryvate is this ready for re-review?

cryvate:
@brainwane: yes

lgh2:
It looks like some searches are still returning older versions of packages in certain circumstances. It looks like it might be something related to [this comment](https://github.com/pypa/warehouse/issues/2217#issuecomment-316762937).  

This is a search run on `pypi.io`: 

pypi_io

And this is the same search run on the branch: 

warehouse_search

di:
@lgh2 This issue is about the order of links on the simple index for a project, e.g. https://pypi.org/simple/pip/

The search on `pypi.org`/`pypi.io` and the development environment is always going to differ, because the dev environment just has a snapshot (and, it's of TestPyPI).

brainwane:
@cryvate Thank you for this fix and thank you for persisting through all the developer environment troubles you had!

Unfortunately I think @ewdurbin is now out of stickers :( (Ernest tell me if I'm wrong.)

@cryvate I really appreciate your work on this and would be happy to recommend other things around Warehouse that would benefit from your attention.

PR #2616, opened 2017-11-25
pypi.python.org/pypi?:action=purge requires a GET

ewdurbin:
Should address https://github.com/pypa/pypi-legacy/issues/739

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1059,7 +1059,7 @@ def file_upload(request):
 
 def _legacy_purge(status, *args, **kwargs):
     if status:
-        requests.post(*args, **kwargs)
+        requests.get(*args, **kwargs)
 
 
 @view_config(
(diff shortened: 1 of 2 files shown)


ewdurbin:
Alternative option is to update pypa/pypi-legacy to accept a POST here.

ewdurbin:
Ah, we need this to be a `POST` to bypass CDN... need to fix in pypi-legacy :)

warehouse/forklift/legacy.py
```diff
@@ -1059,7 +1059,7 @@ def file_upload(request):
 
 def _legacy_purge(status, *args, **kwargs):
     if status:
-        requests.post(*args, **kwargs)
+        requests.get(*args, **kwargs)
```
di: I think this needs to remain a `POST`? I'm not familiar with the legacy API but I'm pretty sure this won't correctly send the `:action` and `project` via the `data` kwarg.

di: Oops, beat me to it. 👍

ewdurbin: confirmed, i did a dumb :)

PR #2624, opened 2017-11-29
adds troubleshooting docs for linter

lgh2:
This commit adds instructions to the docs for installing Node.js, npm, on Debian, Ubuntu and Mint, and using npm to install eslint and sass-lint to fix makefile errors in the linter.

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -257,46 +257,49 @@ Troubleshooting
 Errors when executing ``make lint``
 ------------------------------------
 
-The linter depends on Node.js and the npm packages eslint and sass-lint. If the above command fails with an error like the following:
+The linter depends on Node.js and the npm packages eslint and sass-lint.
+If the above command fails with an error like the following:
 
 .. code-block:: console
 
     ./node_modules/.bin/eslint 'warehouse/static/js/**
-    ' make: ./node_modules/.bin/eslint: Command not found 
-    Makefile:88: recipe for target 'lint' failed 
+    ' make: ./node_modules/.bin/eslint: Command not found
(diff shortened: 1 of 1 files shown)


di:
Glad to see you here already @lgh2! 👋 

Since all of the `Makefile` commands actually depend on having `npm` and `node` properly installed (not just `make lint`) we should probably just tell the user to check that they have these installed further up (i.e. before "Building the Warehouse Container"), rather than just in the linting section.

Also, if the user _does_ have `npm` and `node` properly installed, running `make lint` (or any of the other commands) _should_ install all the necessary dependencies (including `eslint` and `sass-lint`) so those instructions might be unnecessary.

I tested this by deleting my `node_modules` and then running `make lint`, and it works for me. If that doesn't work for you, we should probably figure out why.

lgh2:
Ok, I can move the instructions so that people install Node and npm before the container.  I think that makes more sense, anyway.
What os are you on? I am on Mint and had to install the dependencies manually.  I ran `make lint` after installing Node and eslint, and the sass linter didn't get installed.

di:
@lgh2 I'm using macOS, but it shouldn't make a difference. If you run:

```
$ rm -rf .state
$ rm -rf node_modules
$ make lint
```
Does it give the same 'Command not found' error? Is there anything in `node_modules` afterwards?

di:
Hey @lgh2, I just merged #2642 which should mean that a Warehouse dev should not need to install `node`, `npm`, or any of the JS dependencies locally anymore to do linting and such, as they're now done inside the docker container. 

As such, I don't think these doc changes are necessary anymore, but thanks for raising the issue!

lgh2:
Ok, thank you!

PR #2661, opened 2017-12-11
Add 'Show password' checkbox

### LINKED ISSUE #1984: Show password input (checkbox)
from https://www.nngroup.com/articles/password-creation/

screenshot from 2017-05-13 11-07-02

It would be great if we can add a checkbox that shows/hides the passwords on our register and login forms.

di:
Fixes #1984. This works in [all the browsers we support](https://github.com/pypa/warehouse/blob/master/docs/development/frontend.rst#browser-support).

dec-11-2017 11-30-44

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -106,7 +106,6 @@ docReady(() => {
   window.addEventListener("resize", onResize, false);
 });
 
-
 docReady(() => {
   if (document.querySelector(".-js-autoplay-when-visible")) {
     YouTubeIframeLoader.load((YT) => {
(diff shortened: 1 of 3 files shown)


nlhkabu:
Ok on my side, however we will need someone else to review the JS.  Ping @dstufft, @ewdurbin ?

nlhkabu:
@di is there any reason not to include this also on the registration page?  Maybe one checkbox "show passwords" that reveals the password and password confirmation inputs?

di:
@nlhkabu That makes sense!

nlhkabu:
In which case, it would be better to add the checkbox to the right hand side of the field label, something like this:

screenshot from 2017-12-13 18-21-20

If we do this, we should move the login 'show password' checkbox to the same location (for consistency)?

di:
@nlhkabu That's a bit outside of what I know how to do 🙂 

How about I add the checkbox to the registration page in this PR, and you style it in a separate PR?

nlhkabu:
sounds good @di :)

di:
@nlhkabu This should be ready for you to take over! You should be able to push to my branch but let me know if you have any issues.

nlhkabu:
hi @di - I think it will be easier for me to merge this and I will open a new / separate PR.

PR #2664, opened 2017-12-12
add more detail for autogenerating migrations

mbacchi:


docs/development/database-migrations.rst
```diff
@@ -1,10 +1,16 @@
...
+Modifying database schemata will need database migrations (even for adding and
+removing tables). To autogenerate migrations::
 
-    $ docker-compose run web python -m warehouse db revision
+    $ docker-compose run web python -m warehouse db revision --autogenerate --message "text"
+
+Verify your migration was generated by looking at the output from the command
+above:
+
+    ``Generating /opt/warehouse/src/warehouse/migrations/versions/390811c1c\
```
ewdurbin: can we remove the line break or fix the way it displays after rendering?

docs/development/database-migrations.rst
```diff
@@ -9,7 +9,8 @@ removing tables). To autogenerate migrations::
 Verify your migration was generated by looking at the output from the command
 above:
 
-    ``Generating /opt/warehouse/src/warehouse/migrations/versions/390811c1cdbe_.py ... done``
+    ``Generating /opt/warehouse/src/warehouse/migrations/versions/390811c1c\
```
ewdurbin: It doesn't look like this line break is being properly rendered. I'm not an RST expert though, so I'm not sure how to resolve the rendered `\ `

ewdurbin: Apparently this renders correctly when viewed _not on github_ ✨

di:
Thanks @mbacchi!

Issue #2666, opened 2017-12-12
release_dependencies with empty string for specifier


ewdurbin:
`devpi-postgresql==0.1.1` was released this morning and was throwing 500s when trying to render on both https://pypi.python.org/pypi/devpi-postgresql/0.1.1 as well as https://pypi.org/project/devpi-postgresql/0.1.1

Investigation showed that the issue is being caused by an empty string in the `specifier` field for a `project_url` type dependency:

```
=> select * from release_dependencies where specifier='' and name="devpi-postgresql";
name      | devpi-postgresql
version   | 0.1.1
kind      | 8
specifier | 
id        | 670291c1-74fe-4500-906e-376439f33b09
```

Nothing stands out in the `setup.py` at https://github.com/devpi/devpi/commit/b54c65ccbc089241f7691c339e4866ad7397669b#diff-8a6f387167cc787557161cb96aad255e

@fschulze: can you specifiy what tools and versions you used to package and release `devpi-postgresql==0.1.1`?

I note that there is another example of an empty specifier from August... see: https://gist.github.com/anonymous/f88f6b767ade67baf8b89a2d0a2e3890

ewdurbin:
Note: I removed the errant specifier to stop the steady flow of 500s PyPI and Warehouse were generating as a result:

```
=> delete from release_dependencies where id = '670291c1-74fe-4500-906e-376439f33b09';
DELETE 1
```

And things are all better :-D

ewdurbin:
Warehouse Sentry Error: https://sentry.io/python-software-foundation/warehouse-production/issues/417262933/

fschulze:
I use devpi-server which can push releases to pypi/warehouse. I recently added support for additional metadata in https://github.com/devpi/devpi/pull/481, because of https://github.com/pytest-dev/pytest/issues/2966.

The data currently submitted in the post on push looks like this:
```python
{'author': 'Florian Schulze, Holger Krekel',
 'author_email': 'florian.schulze@gmx.net',
[...]

PR #2667, opened 2017-12-12
Validate Project-URL metadata

ewdurbin:
RE: #2666

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -264,7 +264,7 @@ def _construct_dependencies(form, types):
 class ListField(wtforms.Field):
 
     def process_formdata(self, valuelist):
-        self.data = [v.strip() for v in valuelist if v]
+        self.data = [v.strip() for v in valuelist if v.strip()]
 
 
 # TODO: Eventually this whole validation thing should move to the packaging
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -251,11 +251,8 @@ def _validate_project_url(value):
 
 
 def _validate_project_url_list(form, field):
-    if not isinstance(field.data, list):
-        raise wtforms.validators.ValidationError("Must be a list.")
-    else:
-        for datum in field.data:
-            _validate_project_url(datum)
+    for datum in field.data:
+        _validate_project_url(datum)
 
 
 def _construct_dependencies(form, types):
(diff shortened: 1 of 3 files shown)


warehouse/forklift/legacy.py
```diff
@@ -251,8 +251,11 @@ def _validate_project_url(value):
 
 
 def _validate_project_url_list(form, field):
-    for datum in field.data:
-        _validate_project_url(datum)
+    if isinstance(field.data, str):
+        _validate_project_url(field.data)  # Or raise ValidationError?
```
ewdurbin: @dstufft i'm not sure if we want to just accept a string or not here.

di: `Project-URL` is a multiple-use field, so this should always be a list. I think it should just fail validation if it isn't.

ewdurbin: oh, you're darn tootin.

https://github.com/pypa/warehouse/blob/6fe519b7506702bdf961631bdb2373670842bf21/warehouse/forklift/legacy.py#L464-L469

di:
Wouldn't this be an issue for any `ListField`-type field, not just `project_urls`?

ewdurbin:
yeah, let's see if I can put a validator on the `ListField` itself, just now realized we own that class :-D

di:
I think we might want a failing test like this on `TestListField`:

```python
@pytest.mark.parametrize(
    ("value", "expected"),
    [
        ("", []),
        ("foobar", ["foobar"]),
    ]
)
def test_coerce_string_into_list(self, value, expected):
    class MyForm(Form):
        test = legacy.ListField()

    form = MyForm(MultiDict({'test': value}))

    assert form.test.data == expected
```

The issue seems to be that `process_formdata` isn't throwing away values that end up as empty strings.

ewdurbin:
yep! I'm retooling to give `ListField` a `pre_validate` method that should do the job nicely.

warehouse/forklift/legacy.py
```diff
@@ -264,7 +264,7 @@ def _construct_dependencies(form, types):
 class ListField(wtforms.Field):
 
     def process_formdata(self, valuelist):
-        self.data = [v.strip() for v in valuelist]
+        self.data = [v.strip() for v in valuelist if v]
```
di: This still could give us empty strings if one of the values is just whitespace:
```
>>> [v.strip() for v in ['', ' ', 'a'] if v]
['', 'a']
```

PR #2672, opened 2017-12-13
Add clues and shutdown directions to Getting Started

brainwane:
cc @lgh2


lgh2:
Would it be better to recommend people to use the `docker-compose stop` and `docker-compose start` commands instead of CTRL+C? Or does it not make a difference? I have done a little bit with Docker, but not enough to confidently recommend one command over the other. 

Here are the[ Docker docs](https://docs.docker.com/compose/reference/stop/)

docs/development/getting-started.rst
```diff
@@ -142,6 +149,15 @@ Web container is listening on port 80. It's accessible at
     docker container with ``docker-machine env``
 
 
+Stopping Warehouse and other services
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the terminal where `make serve` is running, use `Control-C` to
+gracefully stop the Warehouse application.
+
+Or, from another terminal, use ``docker-compose stop``.
```
di: There is also the Makefile command `make stop` available, not sure which should be preferred here.

docs/development/getting-started.rst
```diff
@@ -142,6 +149,15 @@ Web container is listening on port 80. It's accessible at
     docker container with ``docker-machine env``
 
 
+Stopping Warehouse and other services
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the terminal where `make serve` is running, use `Control-C` to
+gracefully stop the Warehouse application.
```
di: "gracefully stop all Docker containers" perhaps?

brainwane: Fixing, thanks.

docs/development/getting-started.rst
brainwane: `make stop`, because that filters for `warehouse` things to stop. Thanks. Fixing.

docs/development/getting-started.rst
```diff
@@ -155,6 +176,21 @@ anything private. The password for every account has been set to the string
...
+Running your developer environment after initial setup
+------------------------------------------------------
+
+You won't have to initialize the database after the first time you do
+so, and you will rarely have to re-run `make build`. Ordinarily, to
+access your developer environment, you'll:
+
+.. code-block:: console
+
+    $ cd warehouse
```
di: This might be confusing because there's another `warehouse` directory within the top-level `warehouse` directory.

Let's assume that all of the commands in this guide are run from the top level of the repo, so the dev will never have to `cd warehouse`.

brainwane: Good point! Will revise. Thanks.

Issue #2663, opened 2017-12-12
ipython shell not available


mbacchi:
Trying to use ipython from cli/shell.py I get the following errors:

```
[mbacchi@hostname warehouse-mbacchi-3]$ make shell --type=ipython
make: unrecognized option '--type=ipython'
...
[mbacchi@hostname warehouse-mbacchi-3]$ docker-compose run --rm web python -m warehouse shell --type=ipython
Starting warehousembacchi3_db_1 ... done
Starting warehousembacchi3_redis_1 ... done
Starting warehousembacchi3_rabbitmq_1 ... 
Starting warehousembacchi3_rabbitmq_1 ... done
{"logger": "raven.base.Client", "level": "INFO", "event": "Raven is not configured (logging is disabled). Please see the documentation for more information.", "thread": 140115770255104}
Error: The 'ipython' shell is not available.
```

I think this solution means adding ipython to the requirements, but not sure how that is generated at the moment.

mbacchi:
I provided a fix for this (if it indeed is something that maintainers want 'fixed', as it pulls in quite a number of dependencies) by adding requirements.in and requirements.txt in PR #2674.

What I am not sure about is whether I generated the hashes correctly in requirements.txt. My method was to download the package and its dependencies via "pip download ipython"(some were wheels, one was a .gz), perform a "pip hash" on them, and add it manually to the requirements.txt file. This was a somewhat error prone process, and I'm guessing someone may have a tool to speed it up and prevent errors.

Anyway, it functions as expected afterwards by running:
```
mbacchi@host warehouse-mbacchi-4]$ docker-compose run --rm web python -m warehouse shell --type=ipython
Starting warehousembacchi4_rabbitmq_1 ... 
Starting warehousembacchi4_redis_1 ... e
Starting warehousembacchi4_rabbitmq_1 ... done
{"logger": "raven.base.Client", "level": "INFO", "event": "Raven is not configured (logging is disabled). Please see the documentation for more information.", "thread": 139861917742848}
Python 3.6.3 (default, Nov  4 2017, 14:24:48) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]:                                                                                                                             
Do you really want to exit ([y]/n)? y
```

di:
Hey @mbacchi, thanks for looking into this!

[...]

PR #2674, opened 2017-12-14
Add ipython to requirements to fix issue #2663

mbacchi:


mbacchi:
I'm going to attempt another solution that allows a user to enable ipython mode.

PR #2693, opened 2017-12-21
Move show password checkbox

nlhkabu:
screenshot from 2017-12-21 08-14-28
screenshot from 2017-12-21 08-14-16

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -122,19 +122,23 @@ docReady(() => {
 });
 
 docReady(() => {
-  let passwordField = document.querySelector("#password");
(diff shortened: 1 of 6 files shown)


nlhkabu:
@di I also noticed that on the register page, only the first password is being revealed.  Can we refactor the JS so that the checkbox shows both the password *and* the password confirmation?

I started doing this, but I got myself into a JS mess :/

di:
Hey @nlhkabu, can you check the "Allow edits from maintainers" box on this PR so I can push a commit to your branch?

nlhkabu:
Hi @di - looks like it's already checked :)

di:
Oops, that was my mistake. I pushed a commit which makes the checkbox toggle both input fields.

PR #2702, opened 2017-12-24
Add ipython requirements to fix issue #2663 - second attempt.

mbacchi:
This will allow you to enable ipython but is disabled by default so that
if you don't want it you won't install extra packages, but for folks that
do you can get an ipython shell.
    
To enable it do the following 3 things:
    
1. In docker-compose.yml set the value services->web->build->args->IPYTHON
to "yes"
2. In the Dockerfile, set the ARG IPYTHON variable to yes (with no quotes).
3. In the Makefile set the IPYTHON variable to yes (with no quotes).


Dockerfile
```diff
@@ -91,6 +96,9 @@ COPY requirements /tmp/requirements
 RUN set -x \
     && if [ "$DEVEL" = "yes" ]; then pip --no-cache-dir --disable-pip-version-check install -r /tmp/requirements/dev.txt; fi
 
+RUN set -x \
+    && if [ "$IPYTHON" = "yes" ]; then pip --no-cache-dir --disable-pip-version-check install -r /tmp/requirements/ipython.txt; fi
```
ewdurbin: Can we add a condition on `$DEVEL` here also to ensure this doesn't accidentally get set in a deployable asset?

`if [ "$IPYTHON" = "yes" ] &&  "$DEVEL" = "yes" ]; then ...`

mbacchi: Totally, I'll do that in the next couple days. I just wanted to get this replacement solution out there so I didn't let it languish too long.

mbacchi:
@ewdurbin when you get a minute, take a look and let me know what you think about these updates.

ewdurbin:
@mbacchi: Overall this looks good, I just tested it it out locally and all went swimmingly.

My only concern is having to set a variable in 3 places to make it happen. Is there any way to pull this from an environment variable? (In my earlier pass I had assumed that it was an Environment Variable, but your helpful and accurate doc update convinced me otherwise)

ewdurbin:
If an Environment Variable is possible, I would suggest `WAREHOUSE_IPYTHON_SHELL=1` or similar. I don't have time to play with it at the moment, but it seems like it'd doable. I'll check back here tomorrow morning and if you haven't responded I'll take a crack at it :)

mbacchi:
Good idea, I'll look at this in the morning.

docs/development/getting-started.rst
```diff
@@ -219,6 +219,17 @@ db     The SQLAlchemy ORM ``Session`` object which has already been configured
...
+the following 3 changes:
+
+* In docker-compose.yml, set the value IPYTHON to "yes".
+
+* In the Dockerfile, set the ARG IPYTHON variable to yes (with no quotes).
+
+* In the Makefile set the IPYTHON variable to yes (with no quotes).
+
+Then run the ``make shell`` command as you would above and you'll get the
+Ipython shell.
```
pradyunsg: nit: IPython

mbacchi:
@ewdurbin This isn't as tidy as I had wanted it to be, but illustrates the issues with passing variables to the docker build command. Apparently you must provide a service name if you use --build-arg. I think what I have done building each docker-compose service is the right way to go but if you have other ideas I'm all ears.

Also I wasn't sure why the DEVEL arg was set to "yes" in docker-compose.yml but if you want to revert those changes that's doable. I figured setting it in the Makefile to override the default docker-compose.yml setting was the right way to go from the standpoint of automating it (vs. manual modification).

And I suppose this same thing could be done for the DEVEL flag itself, allow the Makefile to look for an environment variable and then set the flags for DEVEL="yes" instead of doing it manually.

docker-compose.yml
```diff
@@ -27,7 +27,8 @@ services:
     build:
       context: .
       args:
-        DEVEL: "yes"
+        DEVEL: "no"
```
ewdurbin: Docker Compose is only used locally, so setting DEVEL=yes is always the right call here :)

Let's revert the changes to DEVEL args in docker-compose.yml and this is good to go!

mbacchi:
Great, thanks for the review @ewdurbin. All set now. Happy new year!

Makefile
```diff
@@ -3,6 +3,15 @@ TRAVIS := $(shell echo "$${TRAVIS:-false}")
 PR := $(shell echo "$${TRAVIS_PULL_REQUEST:-false}")
 BRANCH := $(shell echo "$${TRAVIS_BRANCH:-master}")
 DB := example
+DEVEL := no
```
ewdurbin: RE: #2728

Issue #423, opened 2015-03-14
Enable users to modify their own account


dstufft:
We need pages that allow users to modify their own account (email, password, etc).

nlhkabu:
This is not going to be ready for launch, but should be one of the first items addressed after launch.

nlhkabu:
- [ ] Update email
- [ ] Update password
- [ ] Update PGP ?
- [ ] Link OpenID to account 
- [ ] Delete Account

nlhkabu:
Current design/layout:
![pypi-update-details](https://cloud.githubusercontent.com/assets/3323703/16544325/9c922e94-40f9-11e6-8d2d-7bbfd71777f2.png)

gsb-eng:
@dstufft : Could you please confirm below points?
1. A user can have a single email registered with the current register feature, correct? (But, Schema has been designed to accommodate multiple emails)
2. The update email here is supposed to be updating primary email, correct?
3. How about adding additional security here, Verifying current password for any update?

nkanaev:
what's the status of the ticket? @gsb-eng, are you working on this?

gsb-eng:
@nkanaev Yes, I'm working on this.

WhyNotHugo:
Not sure if this is the right place to ask: Given that account deletion is currently not implemented, how would one delete one's account? Is there some alternative channel where this can be requested? I've an old account (which even still has some co-ownerships!!) that I'd like to get rid of.

brainwane:
In our [meeting today](https://wiki.python.org/psf/PackagingWG/2017-12-19-Warehouse) we decided this belongs in the first milestone, the MVP for package maintainers.

di:
I broke this up as follows:
* Change public info: #2829 
* Change emails: #2830 
* Change password: #2831 
* Delete account: #2832 

PGP ids [have gone away in legacy](https://github.com/pypa/pypi-legacy/pull/762) and [we are not currently supporting federated auth](https://github.com/pypa/warehouse/issues/61).

di:
@WhyNotHugo Sorry for the late reply: Account deletion still works for [legacy-pypi](https://pypi.python.org/pypi?%3Aaction=user_form), you should be able to delete your account there once you transfer ownership of any projects (or remove the projects entirely)

PR #2705, opened 2017-12-26
Role Management

### LINKED ISSUE #956: [Feature Request] UI for adding maintainers
The idea here is simple, but I have no clue how hard it would be to implement. That said, it would be great if warehouse had a user interface for current maintainers of a project to add other maintainers.

### LINKED ISSUE #2654: Role maintenance for package owners
Owners need to be able to add "Maintainer"/"Owner" roles to the packages they own.

Legacy:
screen shot 2017-12-08 at 11 45 57 am

di:
Fixes #2654, fixes #956.

This adds the ability for a project Owner to add/remove roles for their projects.

This adds two new logged-in views:
* "Manage Profile" page (just stubbed out with some profile info for now)
* "Manage Projects" page (lists the logged in user's projects)

For individual projects, there are two logged-in views per project:
* "Options" page (just stubbed out w/ some project info)
* "Collaboration" page (this is where role management lives)

@nlhkabu Please feel free to reorganize/restyle as much as you want -- I just did the bare minimum to get this feature functional (especially, please revert 1d9704e).


nlhkabu:
Hi @di.  I've just pushed some (fairly large) changes to the styles and the way that the manager login works.

If it's ok with you, I'll keep working on this branch to continue this work, as it's not there yet.

The main thing I currently need your help with is changing the way that the roles form works.  My new design looks like this:

screenshot from 2018-01-08 18-19-02

The added feature is that a user can change the role for an existing collaborator, rather than having to remove and then re-add them.

Is it possible to add some new form code to enable this functionality?

It would also be *really awesome* if we could make the 'username' text input searchable - the way GitHub does it:

screenshot from 2018-01-08 18-15-53

I realise this second request is fairly large, so I'm happy to open a separate issue for this if you prefer :)

di:
@nlhkabu The reason the form was like that originally is because it's technically possible (right now, on legacy-pypi) for a user to have both the "Maintainer" and "Owner" role -- they are not mutually exclusive.

I don't see any reason why a user would need to have both roles (since all the privileges a Maintainer has, an Owner also has) but since it's possible that we have users right now with both, I wanted this to be able to render correctly when that's the case.

I think we could possibly add some logic in here that hides the "Maintainer" role if the "Owner" role also exists for that user. Presumably when we shut down legacy-pypi, we could run a migration that removes any "duplicate" roles so we won't need that logic forever.

> It would also be really awesome if we could make the 'username' text input searchable - the way GitHub does it:

I agree, that would be awesome, and I think it's doable. But we should probably save that for post-launch work since it's not critical.

warehouse/templates/packaging/detail.html
```diff
@@ -98,10 +98,17 @@ <h1 class="package-header__name">
...
+{% if release.summary %}<!-- todo: or logged in user can edit package -->
 <section class="horizontal-section horizontal-section--grey horizontal-section--thin">
   <div class="site-container">
-    <p class="package-description">{{ release.summary }}</p>
+    <div class="split-layout split-layout--middle">
+      {% if release.summary %}
+        <p class="package-description">{{ release.summary }}</p>
+      {% endif %}
+      <!-- todo: if logged in user can edit package -->
+        <a href="{{ request.route_path('manage.project.settings', name=project.normalized_name) }}" class="button button--highlight">Edit Project</a>
```
di: If we really want an edit link on this page, we'll need to do a client-side include so the page doesn't require the use of a session (mostly adding this as a note to myself)

dstufft:
> I think we could possibly add some logic in here that hides the "Maintainer" role if the "Owner" role also exists for that user. Presumably when we shut down legacy-pypi, we could run a migration that removes any "duplicate" roles so we won't need that logic forever.

I think doing this makes a lot of sense, I don't think it makes much sense at all for someone to be both an owner and a maintainer.

di:
OK, I created #2745 to capture what will need done once legacy is shut down.

I'll rework the form in this PR to handle these duplicate roles once @nlhkabu is ready to hand it back to me.

nlhkabu:
> I think we could possibly add some logic in here that hides the "Maintainer" role if the "Owner" role also exists for that user. Presumably when we shut down legacy-pypi, we could run a migration that removes any "duplicate" roles so we won't need that logic forever.

This would be great.  I absolutely agree - it makes no sense that someone should be able to have both roles... so if we can find a way of hiding this in the UI, that would be ideal.

> I agree, that would be awesome, and I think it's doable. But we should probably save that for post-launch work since it's not critical.

Great! :D I'll open a ticket.

di:
@nlhkabu OK, I've made the select boxes functional now, and hidden any "duplicate" roles. I just realized that I still need to add `JournalEntry` additions for all these views, so I still have some work to do, but that won't affect anything on the UI end of things.

nlhkabu:
hi @di - the role form is working well - thank you :)

One small request - are we able to order the collaborators by when they were added to the project?

In this example, I used Ernest's login - and then added Donald to the project.  What was weird was that Donald wasn't added to the bottom of the list - but at the top above Ernest.

screenshot from 2018-01-17 06-52-24

What do you think?

Also, FYI, I have added quite a lot of UI stuff to this branch that needs more work:

### Overall UI

screenshot from 2018-01-17 06-56-35

### Project Options

screenshot from 2018-01-17 06-56-39

### Deleting a Project
screenshot from 2018-01-17 06-56-47

### Release options
screenshot from 2018-01-17 06-57-12

### Deleting a release
screenshot from 2018-01-17 06-57-18

Overall the TODO list is:

1. Add a documentation page for the documentation tab - cc @ewdurbin 
2. Make the delete project modal do something - fix 'show password' checkbox
3. Make the delete release modal do something  - fix 'show password' checkbox (also, remove repetition in template if possible)
4. Add text instructions for making a new release
5. Add a 'release detail' page (which will show the release files and journal). Link to it from the releases page
6. Add help text on how to update a project (on the help page)

Are we happy to merge this PR with dummy UI and make separate PRs for each of these issues?

warehouse/routes.py
```diff
@@ -87,11 +87,36 @@ def includeme(config):
...
+    # Management (views for logged-in users)
+    config.add_route("manage.profile", "/manage/profile/", domain=warehouse)
+    config.add_route("manage.projects", "/manage/projects/", domain=warehouse)
     config.add_route(
-        "accounts.edit_gravatar",
-        "/user/{username}/edit_gravatar/",
-        factory="warehouse.accounts.models:UserFactory",
-        traverse="/{username}",
+        "manage.project.settings",
+        "/project/{name}/settings/",
```
ewdurbin: should this and the following three routes be namespaces under `/manage`?

it feels valid to me to keep all "edit" activity under a single top level path.

perceived advantages:

  - easier to spot/search in logs
  - easier to manage routing/caching/etc in the layers above warehouse

di:
@nlhkabu @ewdurbin responses to both of you below:

> One small request - are we able to order the collaborators by when they were added to the project?

Currently we don't have the "date created" for any existing roles. We can certainly add it, but I'm not really sure what the ordering would look like when some roles have a timestamp and others don't, and anything we choose would probably be pretty confusing to the user without indicating _why_ the sorting is the way it is.

Currently it's alphabetical based on username, which I think is ideal for finding a given user in a long list of collaborators.

> Also, FYI, I have added quite a lot of UI stuff to this branch that needs more work:

I'd like to keep this PR just focused on role management as much as we can. Perhaps we can feature-flag or comment out anything in the UI that is not actually working and re-enable it when we go to ship those issues?

> Are we happy to merge this PR with dummy UI and make separate PRs for each of these issues?

I don't want to merge this with UI components that don't work, I think that there's a strong chance that they'll be around for a significant amount of time and confuse users.

> Couple things I'd like to see:

Added `TODOs` to the `GH-2745` comments and namespaced all the "manage" routes.

> Looking really good and I think we can get this out to keep making progress. Some functionality I note that it is missing (but obviously does not strictly fall into role management):

These are captured in #423 and should probably be a part of a separate PR which fully addresses that issue.

nlhkabu:
thanks @di and @ewdurbin :D

I've opened tickets for each of the missing UI components and commented them out (adding TODOs linking to the tickets) - so on my side, we're good to go.

I agree with Ernest - maybe listing the owners first is better?  I don't see it as a blocker though - let's merge this PR and open a separate ticket for this?

di:
Thanks for creating all those sub-issues @nlhkabu!

I've resolved the merge conflicts and fixed two small things.

As far as role management goes I think this is good to go, however I'd like to resolve the outstanding UI decisions in #2734 before this is merged.

PR #2728, opened 2018-01-04
Don't set DEVEL in Makefile

di:
This was added in #2702 but the effect is that by default, `DEVEL=no` and `--build-args` was overriding what was in `docker-compose.yml`. Thus running `make serve` with no other environmental vars set would not build the web container in development mode as it originally did.

(cc @mbacchi @ewdurbin)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,12 @@ TRAVIS := $(shell echo "$${TRAVIS:-false}")
 PR := $(shell echo "$${TRAVIS_PULL_REQUEST:-false}")
 BRANCH := $(shell echo "$${TRAVIS_BRANCH:-master}")
 DB := example
-DEVEL := no
 IPYTHON := no
 
 # set environment variable WAREHOUSE_IPYTHON_SHELL=1 if IPython
 # needed in development environment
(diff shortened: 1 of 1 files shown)


ewdurbin:
@di I _think_ this is right, particularly given that we have no intention of using `docker-compose` outside of development.

mbacchi:
Thanks for catching this @di, @ewdurbin. I thought my testing showed this worked correctly but that obviously wasn't the case.

Can you explain to me how it works in production though? I would have assumed that docker-compose was used, and I know about the push for k8s, but is the live warehouse site currently using manual docker configuration possibly, or something else?

Thanks in advance!

ewdurbin:
currently warehouse is deployed to Heroku.

di:
> currently warehouse is deployed to Heroku.

To expand on that, live Warehouse doesn't use docker or docker-compose at all. They are only for local development.

You can see how it's actually being run [here](https://github.com/pypa/warehouse/blob/master/Procfile#L2).

The case where `DEVEL=no` in the `Dockerfile` is a part of our transition to k8s and not part of the current development environment or current deployment.

PR #2764, opened 2018-01-10
Password Reset

### LINKED ISSUE #1228: Implement "forgot password" feature

di:
Fixes #1228.

I've squashed @gsb-eng's original PR into a few commits here that represent the entirety of #1262, and then I've added commits for all the changes I felt were necessary.

@nlhkabu This is ready for you to add additional styling to.


di:
@nlhkabu I should mention that since #2709, any email that would be sent to the user will appear in the console like so:

screen shot 2018-01-10 at 2 51 39 pm

(including a link which should be valid for your environment and will take you to a password reset page for that user)

nlhkabu:
Thanks @di:

I've made the following design updates to this PR:

screenshot from 2018-01-11 07-54-23

screenshot from 2018-01-11 07-51-20

screenshot from 2018-01-11 07-52-00


One small thing - when the user successfully resets their password, they are automatically logged in.  Can we please display a message in the flash banners at the top of the page that says: "You have successfully reset your password".  It should be green (have the `--success` class).

gsb-eng:
@di : This is just a heads-up to make sure SMTP config data is provided. 

I remember that, when I've added "send_email" I didn't request for setting-up SMTP config data in config.py.

di:
Thanks @nlhkabu, your changes LGTM and I've added a commit which adds a flash on successful reset.

warehouse/accounts/services.py
```diff
@@ -55,6 +61,11 @@ def __init__(self, session, ratelimiters=None):
             argon2__parallelism=6,
             argon2__time_cost=6,
         )
+        self.serializer = URLSafeTimedSerializer(
+            settings["password_reset.secret"],
+            salt="password-reset",
+        )
+        self.token_max_age = 21600  # 6 hours
```
ewdurbin: This is a _little_ tight which I'm OK with but perhaps should be configurable.

Also may be worth letting users know in flashed message as well as the contents itself that they only got 21600 seconds to get the job done.

warehouse/accounts/services.py
```diff
@@ -63,6 +74,42 @@ def get_user(self, userid):
...
+    def get_user_by_otk(self, otk):
+        if not otk:
+            raise InvalidPasswordResetToken
+
+        try:
+            data = self.serializer.loads(
+                otk,
+                self.token_max_age
+            )
+        except BadData:
```
ewdurbin: Regarding expiry, I would wonder if we can determine failure reason here. The current expiration TTL of 6 hours will probably lead to a few users wondering WTF happened when they submit an expired token.

warehouse/templates/email/password-reset.body.txt
```diff
@@ -0,0 +1,9 @@
+Someone, perhaps you, has made a password reset request for your PyPI account
+'{{ username }}'.
+
+If you wish to proceed with this request, please follow the link below to reset
+your password:
+
+  {{ request.route_url('accounts.reset-password', _query={'otk': otk}) }}
+
+If you did not make this request, you can simply ignore this email.
```
ewdurbin: Or respond to let the administrators know?

nlhkabu: @ewdurbin what email would we use to contact admins?  Or maybe we should link to the security policy instead?

warehouse/accounts/services.py
nlhkabu: +1 on this- @di if you are happy to output a message on the page, I can style it.

warehouse/accounts/services.py
nlhkabu: so, as I understand it, a user has 6 hours between requesting a reset and completing it?  If so, i agree, we should communicate it in the UI.  Also, in the reset email.  "This link will expire in 6 hours".

warehouse/templates/email/password-reset.body.txt
ewdurbin: that's a really good point a link to the security policy and a note saying a response is OK as well would be solid.

the email will be delivered from an address that _is_ repayable and will go to PyPI Admins though.

warehouse/accounts/services.py
ewdurbin: that understanding is correct @nlhkabu

warehouse/accounts/services.py
```diff
@@ -63,6 +74,42 @@ def get_user(self, userid):
...
+
+        # Compare the otk hash values for the user
+        user_hash = self._generate_otk_hash(user)
+        if not hmac.compare_digest(data.get("user.hash"), user_hash):
+            raise InvalidPasswordResetToken
+
+        return user
+
+    @functools.lru_cache()
+    def get_user_by_username(self, username):
```
dstufft: If we're going to add a public method like this, it needs to be added to the ``IUserService`` interface as well. 

One thing to check, I believe the way SQLAlchemy works, you could implement this method as just doing:

```python
def get_user_by_username(self, username):
    user_id = self.find_userid(username)
    return None if user_id is None else self.get_user(user_id)
```

SQLAlchemy should still have the ``User`` class in it's cache and can just return that existing instance rather than re-querying the database (I think anyways!). That should make the implementation of this method a little more DRY. It also raises the question of if this method should be part of ``IUserService`` or if it makes sense to just use embed those two lines of code into the view. Neither answer is wrong here (and adding it to the ``IUserService`` does make it possible to optimize the queries if I'm wrong about SQLAlchemy or we switch to another system that doesn't do that).

di:
I think I've addressed everyone's feedback:
* `token_max_age` is now configurable (but still 6 hours by default);
* The token age is now included in both the email and the message displayed after the email is sent;
* The token creation and checking is no longer part of the `UserService`;
* Improved error messages for all the cases when the token is invalid;
* Added a line telling users they can respond to the reset email.

nlhkabu:
Small copywriting update:

screenshot from 2018-01-17 07-29-12

warehouse/accounts/services.py
```diff
@@ -167,6 +180,50 @@ def verify_email(self, user_id, email_address):
...
+        if not token:
+            raise InvalidPasswordResetToken("Invalid token")
+
+        try:
+            data = self.serializer.loads(token, max_age=self.token_max_age)
+        except SignatureExpired:
+            raise InvalidPasswordResetToken("Expired token")
+        except BadData: #  Catch all other exceptions
+            raise InvalidPasswordResetToken("Invalid token")
+
```
nlhkabu: Could we make these error messages a bit more helpful to the user? e.g.

"This reset token has expired -  please request a new link"
"This reset token is invalid - please request a new link"

warehouse/accounts/services.py
```diff
@@ -167,6 +180,63 @@ def verify_email(self, user_id, email_address):
...
+        if str(user.last_login) > last_login:
+            raise InvalidPasswordResetToken(
+                "Invalid token - User has logged in since this token was "
+                "requested"
+            )  # TODO: track and audit this, seems alertable
+
+        password_date = data.get("user.password_date")
+        if str(user.password_date) > password_date:
+            raise InvalidPasswordResetToken(
+                "Invalid token - Another password reset has been completed "
```
dstufft: Slight tweak here, we might want to just say that the password has changed since the token was requested (or use an OR statement). If they're already logged in and manually change their password they'll hit this case too.

ewdurbin:
@gsb-eng @di thanks for putting this together!

@gsb-eng that was a substantial effort over a lengthy period of time for an important feature, and we appreciate your persistence.

we recently had some [warehouse logo stickers](https://ernest.ly/stickers/warehouse-sticker-pack) printed up, and I'd like to mail some to you. please email ernest@python.org 💖

PR #2772, opened 2018-01-12
NPM dependency locking

di:
This PR pins back the version of `zopflipng-bin` we install in our static environment because [it was released recently](https://github.com/imagemin/zopflipng-bin/releases) and seems to be causing issues.

This PR also adds a `package-lock.json` which pins all our dependencies and sub-dependencies to specific versions, so that broadly-specified dependencies don't change on us and break things.

@alanbato, could you pull down this branch and confirm if it resolves your issues?


alanbato:
Thanks! I got it to work only after running the following commands:
```
make build
mkdir -p .static
touch .state/docker-build
docker-compose run static
```
Tests are passing now :)

PR #2780, opened 2018-01-13
Fix broken recaptcha testing variables in dev

jonafato:
Docker's env-files don't support variable substitution in this manner.
With the current configuration, the site and secret keys are passed to
the application as the literal strings `"${RECAPTCHA_SITE_KEY}"` and
`"${RECAPTCHA_SECRET_KEY}"` respectively. By only including the
environment variable's name in the env-file rather than giving it a
value as well, the variable is propagated from the local environment as
desired.


di:
Thanks @jonafato!

PR #2782, opened 2018-01-13
Add registration and login autocomplete hints

### LINKED ISSUE #411: Use new autofills
HTML5 comes with new autofilles for new passwords and existing passwords to help aid browsers and extensions to know what needs to go into those boxes.

jonafato:
- Add appropriate `new-password` and `current-password` autocomplete values to password inputs on registration and login pages.
- Add autocomplete values for usernames and emails
- Add missing `required` attributes on the affected fields

Fixes #411.

### DIFF
--- a/warehouse/admin/templates/admin/login.html
+++ b/warehouse/admin/templates/admin/login.html
@@ -51,7 +51,7 @@
       </div>
 
       <div class="form-group{% if form.password.errors %} has-error{% endif %}">
-        {{ form.password(class="form-control", placeholder="Password") }}
+        {{ form.password(class="form-control", placeholder="Password", autocomplete="current-password") }}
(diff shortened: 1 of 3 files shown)


di:
@jonafato Thanks! Looks like this fully resolves #411 but it would probably make sense to add `email` and `username` autocomplete attributes as well while we're at it, would you mind doing that too?

jonafato:
@di Makes sense, updated. I also added `required="required"` on a couple of the affected inputs that were missing this attribute.

di:
Thanks @jonafato!

Send an email to ernest@python.org with a link to this PR and your address and he'll send  you some of these sweet PyPI stickers as a thank-you for your contribution:

image

PR #2813, opened 2018-01-18
Strip whitespace from Project-URL values

### LINKED ISSUE #1419: [Request] Package documentation link(s)
I'm following up [this issue on ReadTheDocs about browsing docs offline with 3rd party apps](https://github.com/rtfd/readthedocs.org/issues/662) - it looks like no one asked anyone here for help, but the issue is still outstanding.

There's currently no way to map from here to documentation in a format other applications can use. Details of why this would be useful are in the above ticket, but essentially there are examples of documentation names being different to the package names.

`pypi-legacy` accepts three URLs from a user, `Home page`, `Download URL` and `Bugtrack url`. `warehouse` only takes `url`, meant for the homepage.

Is there room for optional documentation URLs (plural as docs may be in multiple places and a user can choose which they prefer)?

[Also on pypi-legacy here](https://github.com/pypa/pypi-legacy/issues/501), as I'm not sure if this would be applied to `pypi-legacy` or kept for `warehouse`.

di:
Fixes #1419.

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -160,6 +160,14 @@ class TestRelease:
                 ]),
             ),
             (
+                None,
+                None,
+                ["Source Code, https://example.com/source-code/"],
+                OrderedDict([
+                    ("Source Code", "https://example.com/source-code/"),
+                ]),
+            ),
+            (
                 "https://example.com/home/",
                 "https://example.com/download/",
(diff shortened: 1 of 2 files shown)


fungi:
Thanks! I'd gone diving in the WH source and couldn't spot what was missing to get them to show up. This will be exceedingly useful.

PR #2820, opened 2018-01-22
Fix manage projects view when there are 0 releases

di:


pradyunsg:
I was about to file an issue for this. Thanks for the quick fix @di! ^>^

di:
@pradyunsg It takes around 15 minutes for CI to run and the deployment to be released. I'm not seeing any new exceptions, could you try again?

pradyunsg:
Yep. We're back up. :)

PR #2822, opened 2018-01-22
Style updates

nlhkabu:
Quick PR to fix some visual regressions and simplify some of the design:

## All search now uses a search icon, which also allows us to deprecate the dark button and large buttons :)

screenshot from 2018-01-22 22-05-15

## Padding of the login button fixed

screenshot from 2018-01-22 19-23-49

screenshot from 2018-01-22 19-24-00

## Logo size fixed

screenshot from 2018-01-22 22-18-02

screenshot from 2018-01-22 22-18-15

## About PyPI alignment fixed

screenshot from 2018-01-22 22-20-12

screenshot from 2018-01-22 22-19-48

## Skip to content

I've also made the 'skip to content' link visible by tabbing through the page.  Try it :)


warehouse/templates/error-base.html
```diff
@@ -18,7 +18,9 @@
 {% block body %}
 <div class="viewport-section viewport-section--dark">
   <div class="viewport-section__content">
-    <img alt="{{ request.registry.settings['site.name'] }}" src="{{ request.static_path('warehouse:static/dist/images/logo-small.svg') }}">
+    <a href="/">
```
di: Should this be `href="{{ request.route_path('index') }}` instead?

nlhkabu:
thanks @di - I will do a quick cross browser check to see what's going on :)

warehouse/templates/error-base.html
nlhkabu: yes, will update :)

nlhkabu:
The issue with the 'skip to content' link threw me for a while - until I realised that we had an `autofocus` attribute on the homepage search bar, which was completely throwing out the tab index of the page (e.g. the curstor started at the search input).  Some browsers were ignoring `autofocus` where others were implementing it... thus the cross browser issue.

It should be working now on FF @di - can you please confirm?  I tested it locally on my version, but I'm not sure which version you are on.  I've also tested with IE(11) via browser stack.

di:
@nlhkabu Still no luck. A single tab press takes me to the search bar. I'm on Firefox 58.0.

If you want to merge this anyways and file that away as an issue for later, that's fine with me!

nlhkabu:
Ok @di, let's merge - I added a comment to https://github.com/pypa/warehouse/issues/1317#issuecomment-360371922, so we can sort it out then :)

PR #2823, opened 2018-01-22
Add edit project button on project detail page

### LINKED ISSUE #2810: Link to 'edit project' when logged in user is owner
I have added a button in the UI to link *back* the the manage project page, when the user is viewing their project.

screenshot from 2018-01-18 07-25-07

This is currently commented out.

di:
Fixes #2810.

The button is only shown for logged-in project owners via a client-side include.

### DIFF
--- a/tests/unit/packaging/test_views.py
+++ b/tests/unit/packaging/test_views.py
@@ -231,3 +231,12 @@ class TestReleaseDetail:
         result = views.release_detail(release, db_request)
 
         assert result["license"] == "BSD License, MIT License"
+
+
+class TestEditProjectButton:
+
+    def test_edit_project_button_returns_project(self):
+        project = pretend.stub()
+        assert views.edit_project_button(project, pretend.stub()) == {
+            'project': project
+        }
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -100,6 +100,13 @@ def test_routes(warehouse):
             traverse="/{username}",
             domain=warehouse,
         ),
+        pretend.call(
+            "includes.edit-project-button",
+            "/_includes/edit-project-button/{project_name}",
+            factory="warehouse.packaging.models:ProjectFactory",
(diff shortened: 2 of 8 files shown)


nlhkabu:
thanks for this @di.

One flaw in the template - if a project does not have a `release.summary` then the whole grey horizontal bar will not show, _including_ the edit button.

In this case, we want the bar to show, but with the edit button alone.

Therefore the logic for the whole `horizontal-section` should be (excuse my pseudo code)  `if project.release OR edit project button`.

di:
@nlhkabu Unfortunately it's not that simple, because the edit project button come from a completely separate request. We only know whether it will be displayed or not within the `edit-project-button.html` template, not within the parent.

A few things we could do:
* Always show the grey bar, and show some fallback text if there is no summary instead of hiding it, like "No description provided."
  * Downside: Can't save that little bit of screen real-estate in the edge case where the project has no summary and the button isn't there) 
* Move the "Edit Project" button somewhere else
  * Downside: I don't know where else it could go
* Move the entire summary & grey bar into the `edit-project-button.html` templage
  * Downside: I'd rather keep the client-side-include as simple as possible

I think the first one makes the most sense to me, thoughts?

nlhkabu:
Thanks @di - I've updated the template to use fallback text - the first option you suggested.  I've also done a quick style update to ensure this looks good / stacks well on mobile, with and without a button :)

## Desktop with button
screenshot from 2018-01-25 07-08-01

## Desktop without button
screenshot from 2018-01-25 07-07-40

## Mobile with button
screenshot from 2018-01-25 07-08-12

## Mobile without button
screenshot from 2018-01-25 07-08-20

nlhkabu:
@ewdurbin I believe this is ready.  Can you please review and merge if you are satisfied?  Thx

ewdurbin:
Looks great, thanks @nlhkabu!

PR #2833, opened 2018-01-23
Change public info

### LINKED ISSUE #2829: Manage Account: Change public information
From #423: From the 'manage account' page, a logged-in user should be able to change the following information displayed on their "public" profile:

* Gravatar image (display the image and provide a link to gravatar to change it)
* Name

di:
Fixes #2829.

@nlhkabu This will need some styling, I was hoping that the gravatar image could be placed to the right of the form.

Also, I think we need to adjust the size of the user's real name on the profile page. A `h2` looks nice to me:

| screen shot 2018-01-23 at 3 24 48 pm | screen shot 2018-01-23 at 3 25 18 pm |
|-|-|

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -566,3 +566,11 @@ class TestProfileCallout:
         request = pretend.stub()
 
         assert views.profile_callout(user, request) == {"user": user}
+
+
+class TestEditProfileButton:
+
+    def test_edit_profile_button(self):
+        request = pretend.stub()
+
+        assert views.edit_profile_button(request) == {}
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -27,81 +27,85 @@ from ...common.db.packaging import ProjectFactory, RoleFactory, UserFactory
 
 class TestManageProfile:
 
-    def test_manage_profile(self):
-        request = pretend.stub()
+    def test_manage_profile(self, monkeypatch):
+        user_service = pretend.stub()
+        name = pretend.stub()
+        request = pretend.stub(
+            find_service=lambda *a, **kw: user_service,
+            user=pretend.stub(name=name),
+        )
+        save_profile_obj = pretend.stub()
+        save_profile_cls = pretend.call_recorder(lambda **kw: save_profile_obj)
+        monkeypatch.setattr(views, 'SaveProfileForm', save_profile_cls)
+        view_class = views.ManageProfileViews(request)
 
-        assert views.manage_profile(request) == {}
+        assert view_class.manage_profile() == {
+            'save_profile_form': save_profile_obj,
+        }
+        assert view_class.request == request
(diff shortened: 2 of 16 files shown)


nlhkabu:
thanks for this PR @di :)

Updated UI (tested on desktop and mobile):

screenshot from 2018-01-27 11-09-26

It would be great if we could tailor the gravatar message based on whether the user has an account or not.  I've added some commented out HTML:

```html
<!-- if gravatar account -->
<a href="{{ gravatar_profile(user.email) }}" target="_blank" class="button">
  Change image on gravatar.com
  <i class="fa fa-external-link" aria-hidden="true"></i>
  <span class="sr-only">Opens in new window</span>
</a>
<!-- else
<a href="https://gravatar.com/" target="_blank" class="button">
  Create an avatar on gravatar.com
  <i class="fa fa-external-link" aria-hidden="true"></i>
  <span class="sr-only">Opens in new window</span>
</a>
endif -->
```

Is this technically possible?

Another thing I noticed - we don't have an issue for editing a user's username?  Is there a reason we don't offer this?  If so, I think we should still display the username here somewhere. wdyt?

nlhkabu:
hi @di, I have restyled the public profile page:

## Desktop
screenshot from 2018-01-29 07-49-46

## Tablet
screenshot from 2018-01-29 07-49-56

## Mobile
screenshot from 2018-01-29 07-50-06

I added an edit button for when the profile is the same as the logged in user:

```
<!-- if this user is logged in -->
  <a href="{{ request.route_path('manage.profile') }}" class="button button--primary author-profile__edit-button">Edit Profile</a>
<!-- endif -->
```

Could we please do the same with this as we did with the 'edit project' button?

nlhkabu:
> I don't think this is possible -- all we can do is generate the URL to the image.

Ok, so let me remove my pseudo code in the template :)

> This isn't currently a feature of legacy. I don't really see any reason why it wouldn't be possible, but since it's a new feature it should probably wait until post-launch.

Yep, I 100% agree. However, I think we can*display* the username on this page in the short term - it seems odd that the user can't at least see it here.

> Yup, we can do this
\o/

nlhkabu:
@di, I've pushed up a couple of changes - over to you :)

di:
@nlhkabu I think I'm good here. I made a small tweak to display the user's primary email address used for the Gravatar image:

screen shot 2018-01-31 at 11 42 16 am

Also, I opted to hide the "full name" if it's empty on the public profile, rather than have it fall back on the username. I think it's confusing to the user that this would sometimes be their full name, and sometimes be their username.

Without the header, there's not enough padding between the image and the metadata, so you might need to add a small tweak here:

screen shot 2018-01-31 at 11 48 58 am

Other than that, I think this is good to go.

nlhkabu:
## Profile with and without name

screenshot from 2018-01-31 20-18-23
screenshot from 2018-01-31 20-18-04

PR #2846, opened 2018-01-26
Return empty response when include fails

### LINKED ISSUE #2843: duplicate footer appears in package page only when not logged in
When viewing a package page (appears to apply to any package), the (blue) footer ("Get Help", "About PyPI", "Contributing to PyPI") occurs twice: once below a login prompt and a search text box that appear below the package's title (with the package details below these and the footer), and once at the bottom of the page (the latter as expected).

This behavior occurs only when logged out. In contrast, when logged in, the layout seems as intended: a search text box at the top, and the package title and details contiguously below that search box, with a single footer at the bottom.

di:
Fixes #2843.

If the client side include fails for some reason, return an empty response. For example, if there is a permissions error, don't redirect to the login page, just show an empty view.

### DIFF
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -25,7 +25,7 @@ from warehouse import views
 from warehouse.views import (
     SEARCH_BOOSTS, SEARCH_FIELDS, current_user_indicator, forbidden, health,
     httpexception_view, index, robotstxt, opensearchxml, search, force_status,
-    flash_messages
+    flash_messages, forbidden_include
 )
 
 from ..common.db.accounts import UserFactory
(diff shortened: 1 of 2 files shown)


tests/unit/test_views.py
```diff
@@ -147,6 +147,19 @@ def test_logged_out_redirects_login(self):
...
 
+class TestForbiddenIncludeView:
+
+    def test_forbidden_include(self):
+        exc = pretend.stub()
+        request = pretend.stub()
+
+        resp = forbidden_include(exc, request)
+
+        assert resp.status_code == 200
```
dstufft: Do we still want this to be a ``403`` response? I think the HTML include will work fine in either case, but it feels a little cleaner to keep the response code as ``403`` here? What do you think?

PR #2864, opened 2018-01-31
Generic Token Service

di:
This PR turns the `UserTokenService` added in #2764 into a generic token service that can be
reused, and register the existing service in a way that allows for multiple token services to coexist.

The goal of this is to be able to add a token service for email confirmation that shares the same serialization and configuration scheme, but can be configured and used independently.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -154,7 +154,11 @@ def user_service(db_session, app_config):
 
 @pytest.yield_fixture
 def token_service(app_config):
-    return services.TokenService("secret", "salt", 21600)
+    return services.TokenService(
+        secret="secret",
+        salt="salt",
+        max_age=21600,
+    )
 
 
 class QueryRecorder:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -107,7 +107,6 @@ def app_config(database):
             "elasticsearch.url": "https://localhost/warehouse",
             "files.backend": "warehouse.packaging.services.LocalFileStorage",
             "files.url": "http://localhost:7000/",
-            "password_reset.secret": "insecure secret",
             "sessions.secret": "123456",
             "sessions.url": "redis://localhost:0/",
             "statuspage.url": "https://2p66nmmycsj3.statuspage.io",
@@ -154,10 +153,8 @@ def user_service(db_session, app_config):
 
 
 @pytest.yield_fixture
-def token_service(app_config, user_service):
-    return services.UserTokenService(
-        user_service, app_config.registry.settings
-    )
+def token_service(app_config):
+    return services.TokenService("secret", "salt", 21600)
 
 
 class QueryRecorder:
--- a/warehouse/accounts/services.py
+++ b/warehouse/accounts/services.py
@@ -21,7 +21,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from zope.interface import implementer
 
 from warehouse.accounts.interfaces import (
-    InvalidPasswordResetToken, IUserService, IUserTokenService,
+    IUserService, ITokenService, TokenExpired, TokenInvalid, TokenMissing,
     TooManyFailedLogins,
 )
 from warehouse.accounts.models import Email, User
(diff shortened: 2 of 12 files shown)


warehouse/accounts/services.py
```diff
@@ -180,61 +180,30 @@ def verify_email(self, user_id, email_address):
...
         return self.serializer.dumps({
-            "user.id": str(user.id),
-            "user.last_login": str(user.last_login),
-            "user.password_date": str(user.password_date),
+            key: str(value)
+            for key, value in data.items()
         })
 
-    def get_user_by_token(self, token):
+    def extract_data(self, token):
```
ewdurbin: The names `generate_token` and `extract_data` feel kinda unharmonious to me.

The names `dumps` and `loads` from its dangerous itself are generic enough for our purposes, particularly on the generalized service. This is mostly a nit and not a blocker.

tests/conftest.py
```diff
@@ -154,10 +153,8 @@ def user_service(db_session, app_config):
 
 
 @pytest.yield_fixture
-def token_service(app_config, user_service):
-    return services.UserTokenService(
-        user_service, app_config.registry.settings
-    )
+def token_service(app_config):
+    return services.TokenService("secret", "salt", 21600)
```
ewdurbin: This is more a question than it is a suggestion, but can the service be initialized with `kwargs`? Particularly when run across someplace randomly in the codebase that doesn't have such conveniently obvious values, those can certainly help someone understand what's happening.

warehouse/accounts/views.py
```diff
@@ -285,18 +291,49 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
...
-    except InvalidPasswordResetToken as e:
-        # Fail if the token is invalid for any reason
-        request.session.flash(e.message, queue="error")
+    def _error(message):
+        request.session.flash(message, queue="error")
         return HTTPSeeOther(
             request.route_path("accounts.request-password-reset"),
         )
 
+    try:
```
ewdurbin: This block kinda has me wondering if a `UserTokenService` subclassing the `TokenService` makes sense to bundle all this exception handling up into one place. Would this be a better pattern to create in the codebase rather than having somewhat sensitive code that shouldn't change often in the views file, which will?

warehouse/accounts/services.py
```diff
@@ -255,8 +224,18 @@ def database_login_factory(context, request):
...
-    )
+class TokenServiceFactory:
+
+    def __init__(self, name, service_class=TokenService):
+        self.name = name
+        self.service_class = service_class
+
+    def __call__(self, context, request):
+        secret = request.registry.settings[f"token.{self.name}.secret"]
+        salt = self.name
```
ewdurbin: A little implicit for my tastes but not a problem. The first thing I looked for was appropriate differentiation of salts for each service, and it was kinda tough to find where that was happening.

tests/conftest.py
di: Absolutely can do this.

warehouse/accounts/services.py
di: Yeah, I don't love this either. Do you think it makes sense to make the salt configurable and come from settings instead? Felt like that would be overkill. Open to other ideas here.

ewdurbin: Yeah, [itsdangerous recognizes the confusion this parameter name invokes](http://pythonhosted.org/itsdangerous/#the-salt) so it's probably best for us to have _some_ better name for it.

I'm by no means opposed to leaving this as is.

warehouse/accounts/views.py
di: Eh, I'm not sure. This only happens (and will only happen) here in this view, so I'm partial to just keeping it here.

Also, we wouldn't want to be calling `session.flash` or returning redirects from within the service, so we would still need to have some kind of try/except thing happening here anyways, even if it's just catching a generic exception, I think.

Issue #2824, opened 2018-01-22
Login page autofocus & tabbing


di:
It would be nice if [the login page](http://warehouse.local/account/login/) had the following features:

* the `username` field is auto-focused on page load;
* a `<tab>` from the `username` field went to the `password` field and not the "show password" checkbox (perhaps via [the `tabindex` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex))

---

**Good First Issue**: This issue is good for first time contributors. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

PR #2917, opened 2018-02-11
Add autofocus & tabbing to the login page.

Volcyy:
This PR changes the login page `/account/login/` to autofocus the username field and proceed to the password field once the user presses tab, as suggested in #2824.


nlhkabu:
Hi @Volcyy thanks for your PR :)

Your solution is working when I reproduce this locally, however, *after* the password field, the tab index jumps back to the top of the page. Ideally, the user could tab through all of the fields in the form before heading back to the main menu.

Perhaps a better solution is to set the tab index of each part in the form in this order:
Username, password, submit, view password, forgot password.

I would recommend you test this by trying to use only your keyboard to navigate the page - and checking that it makes sense.

Thanks agian :)

Volcyy:
@nlhkabu Thank you for the extensive reply. I've added the `tabindex` attributes to the elements you mentioned, and tabbing now cycles through the form parts as you suggested.
However, when tabbing after `Forgot Password?` is selected, the user will tab through the other elements of the page. According to [this Stack Overflow answer](https://stackoverflow.com/questions/5192859/how-to-ignore-html-element-from-tabindex#5192919), this could be prevented by assigning `tabindex="-1"` to all the other "tab-able" elements, but judging from your reply mentioning "heading back to the main menu", I guess that's intended?

di:
@nlhkabu This looks good to me, could you merge if your concerns have been addressed?

nlhkabu:
Thank you @Volcyy for the PR! :tada:

brainwane:
@Volcyy Also: is there any chance we'll see you [at the PyCon North America sprints this year](https://mail.python.org/pipermail/distutils-sig/2018-February/031986.html)?

di:
Thanks @Volcyy!

Send an email to ernest@python.org with a link to this PR and your address and he'll send you some of these sweet PyPI stickers as a thank-you for your contribution:

image

Volcyy:
@brainwane I'll look into it but I doubt it since I'm in Europe and have uni at that time. Thanks for the invitation though!
@di Thanks for telling me about these, I'll do that! 😄

Issue #2882, opened 2018-02-03
Redirect to project page when user logs in


nlhkabu:
When a user first logs in (or successfully resets their password) , they are currently redirected to the homepage.

It would be nice if they were instead redirected to their projects page at `http://localhost/manage/projects/`

ping @di for feedback and labelling.  I assume we can tag this with 'good first issue'?

di:
Agreed. 👍

pradyunsg:
https://github.com/pypa/warehouse/blob/64510f16148cefe77e7ae69809575c36e484e475/warehouse/accounts/views.py#L119

Just a matter of updating this line, correct?

di:
Yes, and the corresponding tests.

waseem18:
@di @nlhkabu I'm interested in contributing to `warehouse` and would like to start with this good first issue.

di:
Fixed in #2919.

PR #2919, opened 2018-02-11
Redirects to projects page when user logs in

waseem18:
Fix for #2882 

Changed `/` to `/manage/projects` - changed required tests as well


warehouse/accounts/views.py
```diff
@@ -116,7 +116,7 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME,
         # the index instead.
         if (not redirect_to or
                 not is_safe_url(url=redirect_to, host=request.host)):
-            redirect_to = "/"
+            redirect_to = "/manage/projects"
```
di: I think it would be better if we used `request.route_path()` to generate a path to the `manage.projects` route instead of just setting this to be the string `"/manage/projects"`.

The behavior wouldn't change, but if this route ever changed in the future, we wouldn't have to come back here and update this line.

waseem18: Thanks for the review @di 

I'll address the reviews.

waseem18:
@di addressed

tests/unit/accounts/test_views.py
```diff
@@ -170,7 +170,7 @@ def test_post_validate_redirects(self, monkeypatch, pyramid_request,
 
         assert isinstance(result, HTTPSeeOther)
 
-        assert result.headers["Location"] == "/"
+        assert result.headers["Location"] == "/manage/projects"
```
di: This kind of introduces the same problem as my previous comment: if this path ever changes, we'll have to come and update this string to patch it.

Since we're not actually interested in testing that `route_path` works as expected, what we usually do in these tests is mock it out like so:

```py
pyramid_request.route_path = pretend.call_recorder(lambda a: '/the-redirect')
```

And then later, we can assert that it was called with the right route identifier:

```py
assert pyramid_request.route_path.calls == [pretend.call('manage.projects')]
```

And that the response has the path we stubbed out:

```py
```

tests/unit/accounts/test_views.py
```diff
@@ -202,7 +202,7 @@ def test_post_validate_redirects(self, monkeypatch, pyramid_request,
         ("expected_next_url, observed_next_url"),
         [
             ("/security/", "/security/"),
-            ("http://example.com", "/"),
+            ("http://example.com", "/manage/projects"),
```
waseem18: @di Do I keep this "/manage/projects" as it is or shall I change it? I see this test failing for some reason.

di: The tests are failing because it's trying to call `route_path`, you'll need to mock it out for this test as well, and this value will become something like `'/the-redirect'` (or whatever you set the mocked path to be).

waseem18: @di Just a quick question :

If I have 

`assert pyramid_request.route_path.calls == [pretend.call('manage.projects')]` wouldn't it fail the test case with the parameter `/securiry` ?

di: We don't necessarily need to assert that `route_path` is called here, I think it'd be fine to omit that assertion.

waseem18:
@di addressed the changes in `tests`.

tests/unit/accounts/test_views.py
```diff
@@ -222,11 +224,10 @@ def test_post_validate_no_redirects(self, pyramid_request,
             username=pretend.stub(data="theuser"),
         )
         form_class = pretend.call_recorder(lambda d, user_service: form_obj)
-
+        pyramid_request.route_path = pretend.call_recorder(lambda a: observed_next_url)
```
di: Let's make this:
```
pyramid_request.route_path = pretend.call_recorder(lambda a: '/the-redirect')
```
Just to ensure that this is only being called in the second test case.

waseem18: Okay!

waseem18: addressed :+1:

di:
@waseem18 You've got linting errors on a few lines that are too long:
```
./tests/unit/accounts/test_views.py:166:80: E501 line too long (85 > 79 characters)
./tests/unit/accounts/test_views.py:174:80: E501 line too long (84 > 79 characters)
./tests/unit/accounts/test_views.py:227:80: E501 line too long (85 > 79 characters)
```

waseem18:
@di Will correct them

di:
Thanks @waseem18!

Send an email to ernest@python.org with a link to this PR and your address and he'll send you some of these sweet PyPI stickers as a thank-you for your contribution:

image

waseem18:
Thanks @di for reviewing and merging my PR.

Starting with this `good-first-issue` I would love to contribute more to `warehouse`. Will start will resolving `unassigned bugs` first and then will continue to feature requests.

Will plan to spend at least an hour daily working on `warehouse` after my office. :)

Thanks for the support

PR #2922, opened 2018-02-12
Implement Account Recovery with Email (#2887)

### LINKED ISSUE #2887: Account recovery by Email
If you try to do "Request Password Reset" you'll face a problem. The form requires your login that most of the users don't know or forget. It'll be very useful to do account recovery by Email, isn't it?

alanbato:
Fixes #2887.

This PR is still a work in progress, since I've only implemented the field for the user to enter their email.

However, I have some questions:
1. How can we make one (any) of the two fields obligatory? Right now it still asks for a username
2. How can we check that the email is linked to the user? (Maybe a custom validation on the email field?)
3. Should we have an EmailMixin just like we have for Username? I see the email field defined and used in several forms
4. Where is the bathroom in this place?
5. Why do we have an username->user method and an email->userid method but not an email->user method? Should we create one? (using email->userid->user)
6. Is there an end to this list of questions?
7. Yes

Any help is appreciated, I'll keep working on trying to figure some of this things out meanwhile. Just wanted to let you know work is being done to implement this feature :)
🐍


waseem18:
@alanbato Instead of having another field for email wouldn't it be better to have just one single field with the place holdeder `Username or Email` and identify through regex if the entered content is whether username or email and then proceed further?

Please correct me if I'm wrong somewhere.

Thanks

alanbato:
@waseem18 Yes, that is also an option. However, that would mean dropping the use of the UsernameMixin in the form, and I'm not sure if we're making any additional validations there.
But yes, it would make the interface simpler to the user at the cost of more validation and preprocessing steps in the back. 
I'm not sure which option is best to be honest, haha.

Thanks for the feedback!

di:
Great questions @alanbato, responses below:

> How can we make one (any) of the two fields obligatory? Right now it still asks for a username

I think the right thing to do here is keep this a single field that accepts anything (thus relaxing the validation that the username exits) and modifying the view to look for either:

```python
username_or_email = wtforms.StringField(
         validators=[wtforms.validators.DataRequired()]
)
```

> How can we check that the email is linked to the user? (Maybe a custom validation on the email field?)

We probably don't want to do this validation in the form, as it would reveal whether a given email address is tied to a user account or not. It'd be better for this view to always flash something like "If that username or email address exists, we have sent a reset email" and then in the view, we do:

* check if a user exists with a username corresponding to the input, if so send the email;
* if not, check if a user exists with an email corresponding to the input, if so send the email;
* if not, do nothing

> Should we have an EmailMixin just like we have for Username? I see the email field defined and used in several forms

Possibly, but I don't think it's necessary for this PR.
    
> Why do we have an username->user method and an email->userid method but not an email->user method? Should we create one? (using email->userid->user)

It just hasn't been created yet -- this might be a good time to do so.

di:
Hey @alanbato, how's it going? Need any more direction on this PR?

alanbato:
Hi @di, thank you for your help. I've had that much time to implement the rest of it, but I think I now have a pretty good idea of how to do it. :)

I should be pushing an update to this and my other PR by the end of the day.
Thanks again for the feedback! 🐍

warehouse/accounts/views.py
```diff
@@ -252,7 +252,9 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
     form = _form_class(request.POST, user_service=user_service)
 
     if request.method == "POST" and form.validate():
-        user = user_service.get_user_by_username(form.username.data)
+        user = user_service.get_user_by_username(form.username_or_email.data)
+        if user is None:
+            user = user_service.get_user_by_email(form.username_or_email.data)
```
di: Looks like we'll need to add some tests around this conditional. We also don't want to call `send_password_reset_email` if `user` is `None`, so we'll need another conditional there as well (and tests for that condition).

alanbato: Were we not calling `send_password_reset_email` with `None` when a user with that username was not found? That's why I kept it that way, I assumed that the function did a noop when None was passed in as a user.

How would I go about testing that last condition? As in, testing that nothing is done if a user is not found.

di: The `UsernameMixin` was validating that the username existed, so previously the form validation would fail. This won't happen here anymore. (BTW, `RequestPasswordResetForm` is still inheriting from `UsernameMixin`, which it shouldn't be anymore)

To test the last condition, I would monkeypatch `send_password_reset_email` the same way we do in other tests, and just assert that it is not called.

alanbato: Yep, I wanted to remove `UsernameMixin` from the form, but some tests were relying on its validation of the username and I didn't know if I should remove them.
I guess instead of verifying that the username exists I'll try to change the tests to check if the result is the `{'n_hours': n_hours}` dict or `None`

As for the last condition, I'm adding another test to assert `send_password_reset_email` is not being called.

Funny how all this `pretend` functions work :) I still need to learn more about testing frameworks and modules

di: Yeah, we should remove it and update the tests -- we don't need the `username` field on this form anymore.

Hmm, we might have to tweak this a little bit so that `n_hours` can still be passed to the template even if `send_password_reset_email` is not called. We can get it from the `TokenService` like so:

```python
token_service = request.find_service(ITokenService, name='password')
n_hours = token_service.max_age // 60 // 60
```
Let's just use this instead of what `send_password_reset_email` sends back.

warehouse/accounts/services.py
```diff
@@ -76,6 +76,11 @@ def get_user_by_username(self, username):
         user_id = self.find_userid(username)
         return None if user_id is None else self.get_user(user_id)
 
+    @functools.lru_cache()
+    def get_user_by_email(self, email):
```
di: This method needs added to the `IUserService` interface as well.

warehouse/accounts/views.py
```diff
@@ -252,9 +252,12 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
     form = _form_class(request.POST, user_service=user_service)
 
     if request.method == "POST" and form.validate():
-        user = user_service.get_user_by_username(form.username.data)
-        fields = send_password_reset_email(request, user)
-        return {'n_hours': fields['n_hours']}
+        user = user_service.get_user_by_username(form.username_or_email.data)
+        if user is None:
+            user = user_service.get_user_by_email(form.username_or_email.data)
+        if user is not None:
```
di: This can just be `if user:`

pradyunsg: or it can be just `else:` :)

di: @pradyunsg Actually, we do want to test `user` again, since `get_user_by_email` could return either valid user or `None`, and we don't want to send the email if it's `None`.

pradyunsg: Whoops! I missed the assignment in the body.

_hides behind a paper bag_

di:
Hey @alanbato, how's it going with this PR?

Since #2887 is in our End-User MVP milestone, I may want to help you out a bit here and add a commit or two to get this merged if you don't have bandwidth to work on it sooner than later (which is totally understandable).

alanbato:
Hi @di, thank you for bearing with 😅 My schedule is a little bit constrained on the weekdays and I'm sorry if sometimes I don't reply right back, I feel more comfortable doing the requested changes before replying.

I've implemented the changes requested, and I'm currently figuring out how to use stubs to make the fake token service work correctly in the tests with the new return value 😅

tests/unit/accounts/test_views.py
```diff
@@ -418,6 +418,109 @@ def test_request_password_reset(
...
+        token_service.dumps = pretend.call_recorder(lambda a: "TOK")
+        user_service.get_user_by_username = pretend.call_recorder(
+            lambda a: None
+        )
+        user_service.get_user_by_email = pretend.call_recorder(
+            lambda a: stub_user
+        )
+        pyramid_request.find_service = pretend.call_recorder(
+            lambda *a, **kw: user_service,
+        )
```
di: You probably want something like this here: https://github.com/pypa/warehouse/blob/5783a2678f6a36ed5fbd5427ed112887e0baff1a/tests/unit/accounts/test_views.py#L438-L443

alanbato: Yes, I figured I needed to change that, but came upa less elegant solution:
``` python
pyramid_request.find_service = pretend.call_recorder(
            lambda *a, **kw:
            token_service if 'name' in kw else user_service,
        )
```
I'm debating as to overriding the `max_age` attribute of the `token_service` with a simple `3600` and not a stub. Any opinions?

di: Making it an integer makes sense to me, since the view is now doing actual math with it.

alanbato:
Hey @di, I'm not sure how to modify the `test_get()` test, since apparently it was counting on getting the form back as a response and I've changed that so it always returns the `n_hours` parameter to the view. Maybe I need to put the form back in in some way?

tests/unit/accounts/test_views.py
alanbato: It wasn't required after all /shrugs

warehouse/accounts/views.py
```diff
@@ -259,11 +259,15 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
...
-    return {"form": form}
+        user = user_service.get_user_by_username(form.username_or_email.data)
+        if user is None:
+            user = user_service.get_user_by_email(form.username_or_email.data)
+        if user:
+            fields = send_password_reset_email(request, user)
+
+    token_service = request.find_service(ITokenService, name='password')
+    n_hours = token_service.max_age // 60 // 60
+    return {"n_hours": n_hours}
```
di: These three lines should only happen inside the `if request.method == ...` statement above, we should still `return {"form": form}` at the end of this view.

alanbato: Oh I see! Sorry, I misunderstood your earlier remark on the return value of the view. Making the changes as of now :)

alanbato:
Finally, the promised land! The green ✔️ !

Thoughts on the PR @di? Are we missing anything? :)

di:
Thanks @alanbato, nice work!

alanbato:
Thanks for all your help and patiance @di! Couldn't have done it without you :) :snake:

Issue #2794, opened 2018-01-16
Add architecture explanation to application structure overview


brainwane:
Let's update our [application structure overview](https://warehouse.readthedocs.io/application/) with a writeup like [Zulip's architecture summary](https://zulip.readthedocs.io/en/latest/overview/architecture-overview.html) or a curated list of links to conference talks, blog posts, etc. that would get us 30% of the way towards a history and application overview like [this MediaWiki overview](http://aosabook.org/en/mediawiki.html). We'd mention frameworks and components we use, like:

* Pyramid
* Alembic
* Postgres

and engineering approaches we recommend people know about as they learn Warehouse.

Reasoning: Developers who are new to a codebase need is to know *the design rationale* of confusing bits -- why it was made this way, what decisions are embedded in particular choices, whether particular components are the result of a feature request, a quick fix after an outage, an experiment, etc. (This is based on research summarized in [*Making Software*](http://aosabook.org/en/makingsoftware.html).)

[Discussed a bit on the pypa-dev mailing list.](https://groups.google.com/forum/#!topic/pypa-dev/U6NO55bExbE)

brainwane:
@lgh2 just chatted with Ernest and got some notes that I'll be turning into a PR. Here are those notes for reference -- they are very rough because I requested very quick notes, so that's my fault, not hers:

# The Warehouse codebase

Warehouse uses the Pyramid web framework, the SQLAlchemy ORM, and Postgres for its database. Warehouse's front end uses Jinja2 templates.

The application exists within two Docker containers, one of which contains static files for the website, and the other which contains the Python web application code running in a virtual environment and the database. In the development environment, Docker Compose manages running the containers and the connections between them.

The top-level directory of the Warehouse repo contains a number of files. Among them are the license file, `contributing.rst` and `readme`. The `requirements.txt` file is for the Warehouse virtual environment. The Dockerfile creates the Docker containers that Warehouse runs in, and the docker-compose yml file configures docker compose.  Test configuration is in `setup.cfg`. Heroku uses `runtime.txt`. The makefile contains commands to spin up Docker compose and the Docker containers. There are also some files associated with Warehouse's front end. 

# add files


Since Warehouse was built on top of a pre-existing database, some of the code in the ORM may not look like code from SQLAlchemy’s documentation in order to make it fit the existing tables. There are some places where joins are done using logic instead of a foreign key. 

[...]

PR #2937, opened 2018-02-13
Add stub of new architecture overview

brainwane:
@lgh2 interviewed @ewdurbin and got down some rough notes that I've polished a bit. I know this is just a sketch and not very detailed, but I suggest we review this for accuracy and get this merged, and then iteratively improve this page as Laura and other newer contributors learn more and use their experiences to add to it.

It's probably best to review this commit-by-commit so you see the `git mv` of the legacy URL structure doc separately from the new architecture overview.

Partially gets at #2794 but not entirely.

### DIFF
--- a/docs/application.rst
+++ b/docs/application.rst
@@ -1,0 +1,80 @@
+The Warehouse codebase
+======================
+
+Warehouse uses the
+`Pyramid <https://docs.pylonsproject.org/projects/pyramid/en/latest/index.html>`__
+web framework, the
+`SQLAlchemy <https://docs.sqlalchemy.org/en/latest/>`__ ORM, and
+`Postgres <https://www.postgresql.org/docs/>`__ for its database.
+Warehouse's front end uses Jinja2 templates.
+
+The production deployment for Warehouse is in progress and currently
+doesn't use any containers, although we may change that in the
+future. In the development environment, we use several `Docker
+<https://docs.docker.com/>`__ containers, and use `Docker Compose
+<https://docs.docker.com/compose/overview/>`__ to `manage
+<https://github.com/pypa/warehouse/blob/master/docker-compose.yml#L3>`__
+running the containers and the connections between them. In the future
+we will probably reduce that number to two containers, one of which
+contains static files for the website, and the other which contains
+the Python web application code running in a virtual environment and
+the database.
+
+Since Warehouse was built on top of a pre-existing database and
+developers had to fit our ORM to the existing tables, some of the code
+in the ORM may not look like code from SQLAlchemy’s documentation. There
+are some places where joins are done using name-based logic instead of a
+foreign key (but this may change in the future).
+
+Warehouse also uses `Pyramid's hybrid URL traversal and dispatch
+<https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hybrid.html>`__.
+Using factory classes, resources are provided directly to the views
+based on the URL pattern.
+
+Since reads are *much* more common than writes (much more goes out than
+goes in), we try to cache as much as possible. This is a big reason
+that, although we have supported localization in the past, `we currently
+don't <https://github.com/pypa/warehouse/issues/1453>`__.
+
+File and directory structure
+----------------------------
+
+The top-level directory of the Warehouse repo contains files including:
+
+-  ``LICENSE``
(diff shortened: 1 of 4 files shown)


docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
+Warehouse's front end uses Jinja2 templates.
 
-At the moment it *just* lists the legacy structure and none of the intended
-new structure.
+The application exists within two `Docker <https://docs.docker.com/>`__
+containers, one of which contains static files for the website, and the
+other which contains the Python web application code running in a
+virtual environment and the database. In the development environment,
+`Docker Compose <https://docs.docker.com/compose/overview/>`__ manages
+running the containers and the connections between them.
```
di: Technically this isn't true _yet_ -- we currently don't use any containers in production, and we use [far more than two containers in development](https://github.com/pypa/warehouse/blob/master/docker-compose.yml#L3), but this will be true some day in the near future.

It might be worth disambiguating the difference between how we use containers in a development environment vs. how we (will eventually) use them in production.

docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
-  - `rss` and `packages_rss` will be in a bunch of peoples` RSS readers
-  - `doap` is most likely referred to
-  - `show_md5` ?
+Directories within the repository:
 
-**can be deprecated carefully**
-  - `submit`, `upload`, `doc_upload`, `file_upload`,
+::
+
+    bin/ - high-level scripts for Docker
```
di: Also used for Travis and Makefile commands

docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
+        legacy/ - most of the read-only APIs implemented here
+        forklift/ - APIs for upload
+        accounts/ - user accounts
+        admin/ - application-administrator-specific
+        cache/ - caching
+        classifiers/ - frame trove classifiers
+        cli/ - entry scripts and [the interactive shell](https://warehouse.readthedocs.io/development/getting-started/#running-the-interactive-shell)
+        i18n/ - internationalization
+        locales/ - internationalization
+        manage/ - DB
```
di: This is actually for everything a logged-in user can do, (i.e. manage account and owned projects)

docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
+        migrations/ - DB
+        packaging/ - models
+                    - rate limiting to prevent abuse
+                    - RSS feeds
+                    - site maps
```
di: Seems like the directory names are missing here.

docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
-/raw-packages nginx implementation specific hackery (entirely internal; not to
-              be retained)
-/stats        Web stats. Whatever. Probably dead.
-/local-stats  Package download stats. All the legacy mirrors have this.
-/static       Static files (CSS, images) in support of the web interface.
-============= =================================================================
+Warehouse also uses `Pyramid's hybrid URL traversal and
+dispatch <https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hybrid.html>`__.
+Using factory classes, URLs are pre-populated before the view is
+requested.
```
di: I would rephrase this line a bit, I'm not sure 'URLs are pre-populated' makes sense, maybe something like "Using factory classes, resources are provided directly to the views based on the URL pattern"

docs/application.rst
```diff
@@ -1,80 +1,75 @@
...
+-  ``LICENSE``
+-  ``CONTRIBUTING.rst`` (the contribution guide)
+-  ``README.rst``
+-  ``requirements.txt`` for the Warehouse virtual environment
+-  ``Dockerfile``: creates the Docker containers that Warehouse runs in
+-  ``docker-compose.yml`` file configures Docker Compose
+-  ``setup.cfg`` for test configuration
+-  ``runtime.txt`` for Heroku
+-  ``Makefile``: commands to spin up Docker Compose and the Docker
+   containers
```
di: And run tests, and linting, etc.

di:
Also @brainwane you've got a bad doc reference that's failing CI:
```
/home/travis/build/pypa/warehouse/docs/development/legacy-application-structure.rst:20:unknown document: api-reference/legacy
```

brainwane:
Thanks @di. I've fixed all those issues.

PR #2942, opened 2018-02-14
block top-level module namespaces not covered by stdlib-list

### LINKED ISSUE #2940: Blocking of stdlib names has holes?
Hi!

Pull request #2409 introduced a blacklist of stdlib names to counter to protect against [pytosquatting](https://www.pytosquatting.org/), aiming to plug #2151.

I found a hole with it. While names like `xmlrpc.client` and `encodings.idna` are blocked because they appear in some `stdlib-list` list, some parent packages like `encodings` are not in their list and do not get blocked by warehouse as a consequence. I was able to register these four packages:
* [encodings](https://pypi.org/project/encodings/)
* [xml.etree](https://pypi.org/project/xml.etree/)
* [xml.parsers](https://pypi.org/project/xml.parsers/)
* [xmlrpc](https://pypi.org/project/xmlrpc/)

According to [pytosquatting.org](https://www.pytosquatting.org/), package `encodings` had about 20 hits per day in the past, not sure how much hits the others would get. Personally, I would have considered at least `encodings` and `xmlrpc` worth blocking.

Should `stdlib-list` be completed or rather parent package names auto-included at warehouse level for a generic fix? What do you think?

PS: I'm happy to delete these four packages from PyPI once this hole is plugged. I would ask to keep them up until then, please feel free to verify their whitehat nature. Thank you!

Best, Sebastian

CC: @benjaoming @hannob

ewdurbin:
fixes #2940, initial work failed to take this into consideration.

one other thing to consider is situations noted in the initial report...

should we block packages like:

`encodings.magicencoding`
or
`xml.painville`

that "squat" by appearing to belong to top level namespaces?

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -883,6 +883,9 @@ class TestFileUpload:
                                       "encodings.utf_8_sig",
                                       "distutils.command.build_clib",
                                       "xmlrpc", "xmlrpc.server",
+                                      "xml.etree", "xml.etree.ElementTree",
+                                      "xml.parsers", "xml.parsers.expat",
+                                      "xml.parsers.expat.errors",
                                       "encodings.idna", "encodings",
                                       "CGIHTTPServer", "cgihttpserver"])
     def test_fails_with_stdlib_names(self, pyramid_config, db_request, name):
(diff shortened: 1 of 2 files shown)


ewdurbin:
was going to ship changes in https://github.com/pypa/warehouse/pull/2942/commits/c07fbb30d43d2bb1f3e101aeae82fe0d59f8a913 separately, but just pushed them to try to rattle codecov.io loose.

PR #2948, opened 2018-02-15
Changed match query to term query - changed in tests as well

### LINKED ISSUE #1886: Search method in XML-RPC API treats {"summary": "fix code"} as {"summary": ["fix", "code"]}
As stated in the title, the `search` method of [PyPI/Warehouse's XML-RPC API](https://warehouse.readthedocs.io/api-reference/xml-rpc/) splits the values of its `spec` argument on whitespace, resulting in the method returning (for the example in the title) packages with either "fix" or "code" in their summary, not just those with both "fix" and "code" (let alone just those that contain the exact string "fix code").  This happens even if the `spec` value is passed as a one-element list (e.g., `{"summary": ["fix code"]}`), and while I haven't tested every field, it is not limited to just "summary".

On the off chance that this is the intended behavior, the documentation should at least be updated.

waseem18:
Fixes #1886 

Also added test with spaces in value of field of spec.


waseem18:
@di Can you please review this?

Thanks

di:
Thanks @waseem18!

Issue #2943, opened 2018-02-14
`make clean` fails 


lgh2:
It appears that `make clean` fails when run on the Warehouse repository on my machine.  

Here is the error message I got: 
```
$ make clean
rm -rf warehouse/static/components
rm -rf warehouse/static/dist
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.a7bbeb58.woff': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.1fb8a705.eot.gz': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.e1dd9adb.ttf.br': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.1fb8a705.eot.br': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/FontAwesome.b5744b64.otf.gz': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/FontAwesome.b5744b64.otf': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.912ec66d.svg.gz': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.e1dd9adb.ttf.gz': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.912ec66d.svg': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.68560c51.woff2': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.a7bbeb58.woff.gz': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/FontAwesome.b5744b64.otf.br': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.e1dd9adb.ttf': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.1fb8a705.eot': Permission denied
rm: cannot remove 'warehouse/static/dist/fonts/fontawesome-webfont.912ec66d.svg.br': Permission denied
rm: cannot remove 'warehouse/static/dist/manifest.json.br': Permission denied
rm: cannot remove 'warehouse/static/dist/images/favicon.6a76275d.ico': Permission denied
[...]

PR #2950, opened 2018-02-15
Add troubleshooting guidance for make clean failure

brainwane:
Followup to #2943 documenting how to recover from a `make clean` error. Thanks to @lgh2 and @di for working it out.

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -241,6 +241,32 @@ into the message "no space left on device", try running the following command
 (Solution found and further details available at
 https://github.com/chadoe/docker-cleanup-volumes)
 
+"``rm: cannot remove``" errors when running ``make clean``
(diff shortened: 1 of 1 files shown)


di:
Hmm, I'm not really sure that this will happen often enough to merit an entry in the troubleshooting guide... though I'm still not really sure why it happened in the first place. Thoughts @brainwane?

brainwane:
@di Thanks for the assessment. This happened to both me and Laura so I'm inclined to say that it's a thing that might also happen to other people. But I'm also ok with leaving this PR un-merged and closing it, since, as you say, we aren't *actually* sure what causes it, and I don't want to be inaccurate in our docs.

And then: if you are from the future and you find that you're running into this error, and you found this PR by searching for the text of the error, please talk with us here, in [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or on the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev), and help us find out what causes it!

PR #2967, opened 2018-02-17
Implement an Admin Flag model and some Emergency Circuit breakers

ewdurbin:
Allows us to query any flag and it will default to `False` until created and enabled.


warehouse/accounts/views.py
```diff
@@ -215,6 +216,13 @@ def register(request, _form_class=RegistrationForm):
     if request.authenticated_userid is not None:
         return HTTPSeeOther("/")
 
+    if AdminFlag().is_enabled(request.db, 'disallow-new-user-registration'):
+        request.session.flash(
+            "User Registration Temporarily Disabled",
+            queue="error",
+        )
+        raise HTTPForbidden()
```
di: Could be wrong here but I'm not sure the flashed message will appear on the 403 page.

warehouse/forklift/legacy.py
```diff
@@ -740,6 +741,18 @@ def file_upload(request):
...
+        # Check for AdminFlag set by a PyPI Administrator disabling new project
+        # registration, reasons for this include Spammers, security
+        # vulnerabilities, or just wanting to be lazy and not worry ;)
+        if AdminFlag().is_enabled(
+                request.db,
+                'disallow-new-project-registration'):
+            raise _exc_with_message(
+                HTTPForbidden,
+                ("New Project Registration Temporarily Disabled "
+                 "See https://pypi.org/help#admin-intervention for details"),
```
di: Maybe use a `request.route_path` for the URL here instead?

ewdurbin: probably a good refactor for _all_ of our `/help` down the line.

warehouse/accounts/views.py
ewdurbin: hrm... yeah. good catch.

ewdurbin: We _certainly_ need to message correctly to users what's up whenever we use these. I'm going to see what options we have.

di: Could just redirect to a page that shows the flashed messages (like the index).

ewdurbin: Addressed!

warehouse/accounts/views.py
```diff
@@ -215,6 +216,14 @@ def register(request, _form_class=RegistrationForm):
     if request.authenticated_userid is not None:
         return HTTPSeeOther("/")
 
+    if AdminFlag().is_enabled(request.db, 'disallow-new-user-registration'):
```
alex: You don't need to instantiate `AdminFlag` here, should just be `AdminFlag.is_enabled(...)`. Same with the other callsite.

warehouse/migrations/versions/7165e957cddc_create_table_for_warehouse_.py
```diff
@@ -0,0 +1,52 @@
...
+#       not break compatibility with the *previous* version of the code base.
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
dstufft: Not a big deal, but I normally remove these lines, at least as a marker that I looked at the migration and it's doing the right thing.

warehouse/utils/admin_flags.py
```diff
@@ -0,0 +1,33 @@
...
+class AdminFlag(db.Model):
+
+    __tablename__ = "warehouse_admin_flag"
+
+    id = Column(Text, primary_key=True, nullable=False)
+    description = Column(Text, nullable=False)
+    enabled = Column(Boolean)
+
+    @classmethod
+    def is_enabled(cls, session, flag_name):
```
dstufft: This should probably either be a static method, or it should use ``cls`` in the body instead of ``AdminFlag``.

warehouse/utils/admin_flags.py
```diff
@@ -0,0 +1,33 @@
...
+from warehouse import db
+
+
+class AdminFlag(db.Model):
+
+    __tablename__ = "warehouse_admin_flag"
+
+    id = Column(Text, primary_key=True, nullable=False)
+    description = Column(Text, nullable=False)
+    enabled = Column(Boolean)
```
dstufft: Is there a reason to allow this to nullable? What state does that represent? AFAIK there is only two states for a flag, either on or off?

warehouse/utils/admin_flags.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @classmethod
+    def is_enabled(cls, session, flag_name):
+        flag = (session.query(AdminFlag)
+                       .filter(AdminFlag.id == flag_name)
+                       .first())
+        if flag is None:
```
dstufft: I'm torn on this, on one hand it's nice that by surviving a flag not existing in the database, that we're able to handle things like the code running prior to migrations having being ran. However I think that generally code can require a migration being ran-- but migrations have to be compatible with the previous iteration of Warehouse right?

In that case I think I would suggest making it a hard error to not have the flag exist in the database, and prepopulate the database in the migration with any new admin flags. That also leads the way to an eventual admin interface for these, where the admin interface itself doesn't allow creating new admin flags (since they require code support, and getting the ID correct is important) but does allow toggling the flag on/off.It also means there is a single place we can see all of the admin flags (in the DB).

warehouse/accounts/views.py
ewdurbin: d'oh missed that in factoring to a `classmethod`. thanks.

warehouse/utils/admin_flags.py
ewdurbin: yeah, also missed while factoring to a classmethod. do you have a preference between classmethod or static method for this?

warehouse/utils/admin_flags.py
ewdurbin: nope, clear oversight :)

warehouse/utils/admin_flags.py
dstufft: I don't really. I'd maybe use a classmethod just to make inheritance work... but I don't think that really matters since we're never going to use inheritance here anyways.

warehouse/utils/admin_flags.py
ewdurbin: I'd certainly rather failsafe than make it a hard error if a flag isn't available. Consider the case where a release is being canaried and has _optional_ migrations or works around those in some fashion.

But, I did add the flags we're using into the migration, as indeed it significantly reduces the complexity the Admin interface implementation.

ewdurbin:
screen shot 2018-02-18 at 12 37 10 pm

PR #2972, opened 2018-02-18
Require at least one verified email address to register new projects.

ewdurbin:
Great policy anyway, but now that we have a working implementation, let's use it.

An ongoing spam wave is using a wide range of accounts, ip addresses, and variants of their content. This should dramatically increase the effort required on their part to continue their efforts.


ewdurbin:
screen shot 2018-02-18 at 9 24 28 am

Issue #2976, opened 2018-02-18
Admin: Spam cleanup and prevention tooling


ewdurbin:
Historically spammers have been rather naive by registering tons of packages from a single user account or IP address.

Our latest wave of spam has been coming from a multitude of users and IP addresses :(

I propose a couple parts in the Warehouse Admin UI for:

# ~~Complete delete of a user~~ (done)

  - Must display a list of all projects/releases to be deleted before execution
  - Must require explicit confirmation in some manner before execution
  - Should remove all projects and releases they have "contributed"
  - Should disable and trash the account

# Bulk delete from supplied package name list

  - Must display a list of all projects/releases to be deleted before execution
  - Must require explicit confirmation in some manner before execution
  - Should remove all projects and releases associated

brainwane:
I discussed this with Ernest & Dustin and we decided that it's not in Milestone 1; if there are bits of this that we need to do by any particular milestone, such as Launch, then we'll break out issues and prioritize and do them separately.

di:
This has been sufficiently resolved.

PR #2977, opened 2018-02-19
Admin feature: Nuke user

di:
Towards fixing #2976.

Remove the user account, any projects they own, associated releases, roles, emails, etc.

### DIFF
--- a/warehouse/admin/views/users.py
+++ b/warehouse/admin/views/users.py
@@ -23,7 +23,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from warehouse import forms
 from warehouse.accounts.models import User, Email
-from warehouse.packaging.models import Role
+from warehouse.packaging.models import JournalEntry, Role
 from warehouse.utils.paginate import paginate_url_factory
 from warehouse.utils.project import remove_project
 
@@ -151,5 +151,13 @@ def user_delete(request):
         remove_project(project, request, flash=False)
 
     request.db.delete(user)
+    request.db.add(
+        JournalEntry(
+            name=f'user:{user.username}',
+            action=f'nuke user',
+            submitted_by=request.user,
+            submitted_from=request.remote_addr,
+        )
+    )
     request.session.flash(f'Nuked user {user.username!r}.', queue='success')
     return HTTPSeeOther(request.route_path('admin.user.list'))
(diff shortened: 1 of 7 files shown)


warehouse/admin/views/users.py
```diff
@@ -125,3 +126,23 @@ def user_detail(request):
...
+    require_csrf=True,
+)
+def user_delete(request):
+    user = request.db.query(User).get(request.matchdict['user_id'])
+
+    # Delete projects one by one so they are purged from the cache
+    for project in user.projects:
+        remove_project(project, request, flash=False)
+
+    request.db.delete(user)
```
ewdurbin: this will implicitly cascade to their associated emails?

warehouse/admin/views/users.py
```diff
@@ -125,3 +126,23 @@ def user_detail(request):
...
+    permission='admin',
+    uses_session=True,
+    require_csrf=True,
+)
+def user_delete(request):
+    user = request.db.query(User).get(request.matchdict['user_id'])
+
+    # Delete projects one by one so they are purged from the cache
+    for project in user.projects:
+        remove_project(project, request, flash=False)
```
ewdurbin: Need a `JournalEntry`

warehouse/admin/views/users.py
ewdurbin: also probably want a `JournalEntry`

di: Yes, this will delete emails and roles for that user.

I thought about adding a `JournalEntry` here, but it seems like they're project-specific? E.g., we don't make them when a user account is created, just when stuff changes with a specific project.

warehouse/admin/views/users.py
di: There's already a `JournalEntry` happening in `remove_project`.

warehouse/admin/views/users.py
ewdurbin: fair, I would like some record of this feature. the journal is freeform. we should keep a record that an admin used it.

ewdurbin:
```
JournalEntry(
    submitted_by=request.user,
    action='purge_user',
    name=f'user:{user.username}),
)
```

would work and can't collide with project names.

warehouse/admin/views/users.py
ewdurbin: confirmed, sorry I tested the button locally on a user with no projects :)

PR #2978, opened 2018-02-19
Delete account

### LINKED ISSUE #2832: Manage Account: Delete Account
From #423: From the 'manage account' page, a logged-in user should be able to delete their account.

If the user is the sole owner of any projects, this should be disabled and they should be informed that they need to transfer ownership or delete the projects before they can delete their account.

A notification email should be sent to the primary email address when the account is successfully deleted.

di:
Fixes #2832.

@nlhkabu This just needs a quick glance.

With projects solely owned by the user:

screen shot 2018-02-18 at 6 16 50 pm

Without any projects solely owned by the user:
screen shot 2018-02-18 at 6 17 19 pm

Corresponding modal:
screen shot 2018-02-18 at 6 19 38 pm

### DIFF
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -22,7 +22,7 @@ from webob.multidict import MultiDict
 
 from warehouse.manage import views
 from warehouse.accounts.interfaces import IUserService
-from warehouse.packaging.models import JournalEntry, Project, Role
+from warehouse.packaging.models import JournalEntry, Project, Role, User
 
 from ...common.db.accounts import EmailFactory
 from ...common.db.packaging import (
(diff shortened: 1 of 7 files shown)


nlhkabu:
Updated and ready on my side @di :)

## Many projects

screenshot from 2018-02-19 07-30-49

## One project
screenshot from 2018-02-19 07-29-51

## No projects
screenshot from 2018-02-19 07-28-40

nlhkabu:
Also, not sure if this is just my local install... I'm getting a `sqlalchemy` error on `/account/register/` on this branch:

screenshot from 2018-02-19 07-37-51

```
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) relation "warehouse_admin_flag" does not exist
LINE 2: FROM warehouse_admin_flag 
             ^
 [SQL: 'SELECT warehouse_admin_flag.id AS warehouse_admin_flag_id, warehouse_admin_flag.description AS warehouse_admin_flag_description, warehouse_admin_flag.enabled AS warehouse_admin_flag_enabled \nFROM warehouse_admin_flag \nWHERE warehouse_admin_flag.id = %(id_1)s \n LIMIT %(param_1)s'] [parameters: {'id_1': 'disallow-new-user-registration', 'param_1': 1}] (Background on this error at: http://sqlalche.me/e/f405)
```

nlhkabu:
Another thing @di - I noticed that there is no confirmation message when a user deletes their account.  Is this normal?

di:
@nlhkabu Looks great! The error is because there was a migration in #2967, you'll need to do a `docker-compose run web python -m warehouse db upgrade head`.

I'm not entirely sure what you mean by confirmation message, but I think you mean something like a green "Account successfully deleted" flash or something similar. We can't flash a message here because once the user is deleted, they are logged out and their session is destroyed.

We do have the modal, and there is an email sent to the primary email address with a notification that the account was deleted, so I think that should suffice.

nlhkabu:
ok, thanks @di.  I was thinking of a "Account successfully deleted" flash, but if we have no session, that makes sense :D

PR #2979, opened 2018-02-19
Cascade release deletion to description_urls

di:


ewdurbin:
pretty sure we can 🔥 the `description_urls` table at some point.

PR #2980, opened 2018-02-19
Transfer journals to deleted-user when nuking

di:


ewdurbin:
it's gross, but it's "feature parity" hooray "feature parity"

PR #2987, opened 2018-02-19
POC: Rename profile to account settings

nlhkabu:
Not sure if you agree @di?  I feel as though most of the content on this page is now related to the user's account, rather than their profile... thus the rename.  wdyt?


di:
Yeah, I think I agree. Should probably refactor the route and view names as well, if we're gonna do it.

nlhkabu:
I've updated the route, views, templates too @di.  Can you please double check I've not missed anything?  Thx

Issue #2936, opened 2018-02-13
Redirect from login page if user is logged in


di:
Currently the ` accounts.login` view will still render even if there's already user logged in.

Since it doesn't make much sense for a user to log in twice, this view should check if there is a logged in user, and redirect to the post-login view if there is. (See #2882 for which view this should be).

---

**Good First Issue**: This issue is good for first time contributors. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

brainwane:
@waseem18 Would you like to take this on next?

(Until he says yes, this is still up for grabs.)

waseem18:
Absolutely Yes!

I'll start looking into it right away

brainwane:
Cool, thanks, @waseem18 -- if you jump into IRC I have a few more suggestions for you.

waseem18:
Sure, Hopping in

waseem18:
This issue also exists for the view `accounts.register`

Issue #2921, opened 2018-02-12
Pagination for Project History


di:
Per #2911: Currently the 'Project History' page just lists all the journal entries for a project, but this could be quite long. We should add pagination for this page.

cand126:
Hi @di, I would like to ask some questions.

1. Is there a mockup for pagination?
2. Does the current API support pagination?

di:
@cand126 Thanks for the questions. This is specifically about the "Project History" in the logged-in UI for project maintainers, it looks like this:

<img width="1154" alt="Screen Shot 2019-03-25 at 11 13 30 AM" src="https://user-images.githubusercontent.com/294415/54935835-1fd95d80-4eef-11e9-8558-1401f145826a.png">

---

<img width="1164" alt="Screen Shot 2019-03-25 at 11 13 35 AM" src="https://user-images.githubusercontent.com/294415/54935840-21a32100-4eef-11e9-9bf2-07c602e3fca2.png">

The pagination buttons are there, but they don't work, and it looks like right now the results are getting truncated as well. This might just require hooking the pagination query parameter up to the query for the Journal entries.

cand126:
@di I created a package with lots of realeases, and the pagination feature seems fine. Each page has 25 items, and the I can navigate to different pages by click the button. The url also indicates the page number.
![Screenshot from 2019-03-29 23-54-19](https://user-images.githubusercontent.com/6241921/55266669-89a29000-527e-11e9-97a5-ac3620346d30.png)
![Screenshot from 2019-03-29 23-54-24](https://user-images.githubusercontent.com/6241921/55266674-8c9d8080-527e-11e9-88fa-312753036fd6.png)
![Screenshot from 2019-03-29 23-54-30](https://user-images.githubusercontent.com/6241921/55266676-90c99e00-527e-11e9-9dd4-7dd069f2003f.png)

di:
Sorry for the goose chase @cand126. I realized that this was due to a misconfiguration of our CDN, which was stripping out URL parameters for these pages. I just updated and it's now fixed.

Thanks for helping debug!

Issue #2673, opened 2017-12-13
Add readiness notification/link to localhost:80 to `make serve` logs


brainwane:
After a developer runs `make serve`, eventually the process seems to hang; that's when the application's ready for you to view it at localhost:80, but that's not obvious unless you know (see also #2672 which adds a note to the docs to tell you what to look for). The logs say:

~~~
web_1            | 2017-12-13T21:52:01+0000 [-] Site starting on 8000
web_1            | 2017-12-13T21:52:01+0000 [twisted.web.server.Site#info] Starting factory <twisted.web.server.Site object at 0x7fbb6dcc4240>
web_1            | 2017-12-13T21:52:01+0000 [twisted.application.runner._runner.Runner#info] Starting reactor...
~~~

which doesn't mention `localhost`, implies port 8000 which doesn't let you see the site, and doesn't provide a clickable link.

@lgh2 suggested we add some message to stdout when it starts the server and it's ready for the developer to view it, like "starting server on http://localhost:80". Is this possible?

brainwane:
From `#pypa-dev` just now:

`<mbacchi>` sumanah: not sure it will be easy or desirable to make the web worker say 'ready...' but worth opening a ticket anyway
`<mbacchi>` the messages you're seeing are all the logs aggregated from the docker compose nodes
`<lghampton>` Why would it not be desirable?
`<sumanah>` mbacchi: are you thinking it might be misleading to tell the dev it's ready?
`<di_codes>` i agree with mbacchi that it’s likely not easy. there isn’t really a great way to determine when all the containers are “ready”
`<mbacchi>` you'd have to make the server process output a ready message but is that just on startup or other operations? I'd suggest that the docs should indicate what to watch for rather than the logs/application spitting out some specific message because over time the ready message will be scrolled up in the history
`<sumanah>` mbacchi: di_codes: I'm fine with us not specifically saying "ok now it's go time!" if we put more explicit "here's what to look for" in the docs, and especially if we provide a clickable link to localhost:80 somewhere nearby in the terminal output
`<sumanah>` https://zulip.readthedocs.io/en/latest/development/setup-vagrant.html#step-3-start-the-development-environment is what I'll probably want to pattern after -- "You will see several lines of output starting with something like:" and "And ending with something similar to:" and then "Now the Zulip server should be running and accessible. Verify this by navigating to http://localhost:9991/ in the browser on your main machine."
`<di_codes>` sumanah: agreed, that is probably the best option
[...]

Issue #1388, opened 2016-07-11
404 for registered package with no release


inglesp:
This was [originally raised](https://github.com/pypa/pypi-legacy/issues/456) against PyPI.

When a package has been registered, but no release has been uploaded, Warehouse returns a 404.  It would be more helpful if it displayed information about package.

For an example of this, see https://pypi.io/project/spider/.

reversefold:
How does one find out information about a registered package which has no releases? I tried to upload a package for which I thought the name was unused but got an error (see the linked ticket). I'd like to be able to contact the person who registered it as well as seeing how old the registration is but am not sure how to do so.

brainwane:
Thanks for the bug report! And I'm sorry for the slow response.

In [a Warehouse core developers' meeting](https://wiki.python.org/psf/PackagingWG/2018-03-06-Warehouse) earlier this month, we decided to pare down our near-future milestones on [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) so they really only contain the essential bugfixes and features we need to launch, replace legacy PyPI, and shut down the old site. I see this behavior is currently the same on legacy PyPI and on the new site. So I'm moving this issue into a milestone further in the future; sorry for the wait.

I also think this may be related to [the policy questions addressed in PEP 541](https://www.python.org/dev/peps/pep-0541/) so I'm marking it with that tag as well, so that when that gets accepted, there'll be a little more clarity on how to properly address this.

Thank you again.

di:
I'm not sure I would call this a "bug", but it's definitely un-ergonomic and can cause confusion when a new package author thinks a name has not been taken.

I think we either need to create a new page for this, or make the current project detail view handle a project with zero releases, either of which should indicate that the name is registered but has no releases (and possibly link to our PEP 541 policy/procedure).

pradyunsg:
Does this still need more discussion?

di:
Yes, I think so. There are challenges to both:

* **Create a new page for this:** what does this look like? How does it communicate to the user the state of this project? Do we really need a new page?
* **Make the current project detail view handle a project with zero releases:** The "project" page is actually rendering a release, so without a release to render, it doesn't really work as intended.

pradyunsg:
I was looking at the code and, separately from the discussion above, the relevant page we'd add for this would need to be rendered at this point in the code:

[...]

PR #2988, opened 2018-02-19
Redirect authenticated user from reset pw pages to index

### LINKED ISSUE #2802: Redirect from password reset pages if user is logged in
Currently the two password reset views will still render even if there's a user logged in:

* `accounts.request-password-reset`
* `accounts.reset-password`

Since it doesn't make much sense for a logged in user to request a password reset, or reset their password, these views should check if there is a logged in user, and redirect to the `index` route if there is.

HndrkMkt:
Resolves #2802.


di:
Thanks @HndrkMkt! We'll probably want to remove [this link](https://github.com/pypa/warehouse/blob/bb57e6133bea8a71de61aa24a096ead1a7ca31bf/warehouse/templates/accounts/login.html#L73) as well, since it links to one of these pages for a logged in user.

HndrkMkt:
Hi @di, don't we redirect from that page if the user is logged in as per #2936?

di:
Right, what I mean is that the link is only shown to a logged-in user (on the "manage account" page) and since it will always redirect, we should just remove it, there's no point in having it if a logged-in user can't do a password reset.

di:
Hi @HndrkMkt, does that clarify what I'm asking for in this PR?

HndrkMkt:
Hi @di, sorry for the late reply. I checked the link you provided and as far as I can tell that page is shown if you want to login to the page. Did you mean [this one](https://github.com/pypa/warehouse/blob/master/warehouse/templates/manage/account.html#L236) by any chance? That should be the one displayed on the "manage account" page.

di:
Ah, you're absolutely right @HndrkMkt, good catch. Yes that's the one that should be removed. Sorry for the confusion!

HndrkMkt:
No problem :) I removed the link in my last commit. 
In case there's some other issues you'd recommend working on next, please let me know!

di:
Thanks @HndrkMkt! I'll let @brainwane direct you towards other potential areas for contribution.

brainwane:
@HndrkMkt thanks for your contribution! I'm going to list here a few issues that I think may be good ones for you to work on, based on this PR and other work you've done.

* #2673
* #1307
* #2921 
* #1388

Go ahead and comment on any one of them to say you're working on it!

PR #2995, opened 2018-02-20
Only index pypi.org... test.pypi.org and any other instances have no need

### LINKED ISSUE #916: Provide for robots.txt disabling indexing altogether (eg. for test repos)
testpypi.python.org is currently indexed by google.  This isn't too useful, and also has the chance of "leaking" test releases of packages which might not be ready to ship.  A simple way to address this would be using robots.txt to disable indexing on the entire warehouse instance.

ewdurbin:
closes #916

### DIFF
--- a/tests/functional/test_basic.py
+++ b/tests/functional/test_basic.py
@@ -10,21 +10,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


ewdurbin:
thanks for fixing those tests @di!

PR #2997, opened 2018-02-20
Added help entry for File already exists error

### LINKED ISSUE #2701: "File already exists." Error could use help links
#2670 brings to light that our error when uploading a filename which has already been used, but no longer exists is unclear.

```
HTTPError: 400 Client Error: File already exists. for url: https://upload.pypi.org/legacy/
```

Here are the two scenarios and messaging I believe would be more helpful. We'd need to create the help entry for `https://pypi.org/help/#file-name-reuse` as well.

- Filename has been used, file exists:

```HTTPError: 400 Client Error: File already exists, see https://pypi.org/help/#file-name-reuse. for url: https://upload.pypi.org/legacy/```

- Filename has been used, file no longer exists:

```HTTPError: 400 Client Error: Filename cannot be reused, see https://pypi.org/help/#file-name-reuse. for url: https://upload.pypi.org/legacy/```

### PR DESCRIPTION
**Pull request by waseem18**
Fixes #2701.

Need to add link to help entry in the the error message. I guess this change goes into `twine`.

### CODE REVIEW

**Comment by di** [comment]
The error message is [here](https://github.com/pypa/warehouse/blob/ec17830ba20a5e963558a4f4337b7638304bca2b/warehouse/forklift/legacy.py#L1026), we can update it to include a link to this help section.

**Comment by waseem18** [comment]
Sure @di Will do

File: warehouse/templates/pages/help.html [hunk 1]
```diff
@@ -268,6 +269,19 @@ <h2>{{ admin_intervention() }}</h2>
           When the PyPI Administrators are overwhelmed by spam <b>or</b> determine that there is some other threat to PyPI, new user registration and/or new project registration may be disabled. Check <a href="https://status.python.org">our status page</a> for more details, as we'll likely have updated it with reasoning for the intervention.
         </p>
       </section>
+      <section id="file-name-reuse" class="common-question">
+        <h2>{{ file_name_reuse() }}</h2>
+        <p>
+          The error <i>HTTPError: 400 Client Error: File already exists</i> is caused because of one of the below two reasons:
```
brainwane: I suggest you rephrase as:

> The error <i>HTTPError: 400 Client Error: File already exists</i> happens for one of two reasons:

waseem18: Sure - Will rephrase

warehouse/forklift/legacy.py
```diff
@@ -1033,7 +1036,8 @@ def file_upload(request):
             raise _exc_with_message(
                 HTTPBadRequest,
                 "This filename has previously been used, you should use a "
-                "different version.",
+                "different version. "
+                "See https://pypi.org/help/#file-name-reuse",
```
di: Could you generate these URLs with `request.route_url('help', _anchor='file-name-reuse')` instead? Otherwise these will be wrong for https://test.pypi.org/

waseem18: Sure, Will address.

waseem18: @di What about `_scheme and  _host` ? Wouldn't `request.route_url('help', _anchor='file-name-reuse')` just generate the path?

di: [`route_path`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_path) would generate a path, like `/help/#file-name-reuse`, [`route_url`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_url) generates a fully qualified URL, like  `https://pypi.org/help/#file-name-reuse`, so you don't need to specify a host or scheme.

waseem18:
@di After I put `route_url` I'm getting below error when I run `make tests`. I tried a lot to know the cause of it and resolve but in vain. Could you please help me with this?

error

di:
@waseem18 The issue is that your tests don't mock out `db_request.route_url`, you'll need to add something like:

```python
db_request.route_url = pretend.call_recorder(lambda route, **kw: "/the/help/url/")
```
And then your assertion will become:
```python
assert resp.status == (
    "400 This filename has previously been used, you should use a "
    "different version. See /the/help/url/"
) 
```
since we don't actually need to test if `route_url` works as it should.

di:
And, additionally, you can test that `route_url` was called as expected:

```python
assert db_request.route_url.calls == [pretend.call('help', _anchor='file-name-reuse')]
```

waseem18:
Thanks for the links @brainwane 
I'll surely follow :+1:

di:
Thanks @waseem18!

PR #3013, opened 2018-02-21
add separate Procfile entry for uploads

ewdurbin:
we need to cordon the expensive upload processes off from mainline web to be able to scale them independently.


dstufft:
Does this actually work? I thought you could only have 1 web accessing process in an app.

ewdurbin:
yeah, this is a no go, was just getting stuff out there that might work :-p

PR #3014, opened 2018-02-21
s/twisted/gunicorn/g

ewdurbin:
we'll need to `heroku create --buildpack https://github.com/heroku/heroku-buildpack-nginx.git` before deploying.

we're experiencing memory leaks in `twisted.web` for uploads that require us to swap uwsgi servers.


ewdurbin:
Ready when approved @dstufft !
screen shot 2018-02-21 at 1 26 03 pm

ewdurbin:
k! I think we're set. we can tune after we see how this behaves on real PyPI

PR #3027, opened 2018-02-22
Catch DetachedInstanceError from DebugToolbar

di:
Occasionally, the DebugToobar attempts to `__repr__` a SQLAlchemy instance that has become detached from it's session. Instead, catch the exception and return a different representation when it's detached and we can't get the attributes anymore.


di:
Dunno what's up with codecov again. This is at 100% per https://codecov.io/gh/pypa/warehouse/pull/3027 so I'm going to merge it.

PR #3034, opened 2018-02-22
Add a `Project.latest_release` property

di:
This PR provides a `latest_release` property, which gives a pared-down `Release`-like named tuple for when we don't need the full release (and all the memory consumption that comes with it).

This PR also adds two more lightweight queries where we were originally using `project.releases` and didn't need to be.

### DIFF
--- a/warehouse/manage/views.py
+++ b/warehouse/manage/views.py
@@ -29,7 +29,9 @@ from warehouse.manage.forms import (
     AddEmailForm, ChangePasswordForm, CreateRoleForm, ChangeRoleForm,
     SaveAccountForm,
 )
-from warehouse.packaging.models import File, JournalEntry, Project, Role
+from warehouse.packaging.models import (
+    File, JournalEntry, Project, Release, Role,
+)
 from warehouse.utils.project import confirm_project, remove_project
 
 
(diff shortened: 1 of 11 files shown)


warehouse/manage/views.py
```diff
@@ -337,7 +339,16 @@ def _key(project):
     effective_principals=Authenticated,
 )
 def manage_project_settings(project, request):
-    return {"project": project}
+    n_releases = (
```
ewdurbin: nit: `release_count` might be a more human friendly and obvious name than `n_releases`

warehouse/packaging/models.py
```diff
@@ -175,6 +176,17 @@ def documentation_url(self):
 
         return request.route_url("legacy.docs", project=self.name)
 
+    @property
+    @functools.lru_cache()
```
ewdurbin: _smart_! my only concern here is that _technically_ this can grow unbounded without a `maxsize` parameter attached, and I'm uncertain exactly how long these will linger in the garbage collector... any ideas?

ewdurbin: I guess it gets GC'd with the object... so per request?

di: > I guess it gets GC'd with the object

Nope, it doesn't! Good catch.

PR #3035, opened 2018-02-23
load minimum amount of data for json releases/files

ewdurbin:


ewdurbin:
For projects with _long_ release histories we were pulling altogether too much data to construct the `releases` key which just a dict of files for all previous releases.

ewdurbin:
Report on codecov.io shows 100% I promise.

PR #3037, opened 2018-02-23
Don't install g++ from a PPA in travis

alex:
I have no idea if this will work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ python: 3.6.1
 env:
   global:
     - TRAVIS_NODE_VERSION=6.6.0
-    - CXX=g++-4.8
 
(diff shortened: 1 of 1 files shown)


alex:
The default C++ compiler appears just fine.

This speeds up the tests by a bit.

ewdurbin:
🙏 bless you!

ewdurbin:
for record keeping: these were added in https://github.com/pypa/warehouse/pull/1137 and appear to have been in support of some node dependencies for static builds. it looks like brotli may not have worked with older compilers... but we're good now :)

ewdurbin:
particularly as we don't actually use any assets from travis for deployment, I'm gonna say 🚀

PR #3038, opened 2018-02-23
Make dropdowns tabbable

nlhkabu:
This change makes it so that the browser no longer ignores the hidden links in a dropdown when tabbing through the page with a keyboard.

However, this PR is not complete, because the dropdown is still hidden while the active link is focused by keyboard. (to see what I mean, log in and tab through the page to the user dropdown).

Unfortunately, because CSS has no parent selector, I cannot display the dropdown with CSS alone.  We need some JS:

- When `.dropdown__link` is `:active` by keyboard
- add `.dropdown__content--visible` to the containing (`.dropdown__content`)

- when the user tabs off the dropdown
- remove `.dropdown__content--visible`

@di would you be able to help with this?  Have I provided enough information?  Does it make sense? Thx

### DIFF
--- a/warehouse/static/sass/blocks/_dropdown.scss
+++ b/warehouse/static/sass/blocks/_dropdown.scss
@@ -44,21 +44,21 @@
 
   &__content {
     position: absolute;
-    right: 0;
+    right: 100000px;
     margin-bottom: -4px;
     box-shadow: 1px 1px 2px 1px rgba(0, 0, 0, 0.05);
     z-index: index($z-index-scale, "dropdown");
     border: 1px solid $border-color;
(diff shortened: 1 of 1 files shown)


di:
@nlhkabu Yup, I think we can do this.

di:
@nlhkabu I looked into this a bit more. I also found https://www.w3.org/WAI/tutorials/menus/flyout/ which recommends something slightly different, I wonder if you have seen this and have any thoughts?

nlhkabu:
Cool @di - thanks for the link to the tutorial.  I'd be happy to restructure this component to follow this guideline.  If I update the HTML, could you provide the JS?

nlhkabu:
Closing this PR - I will open a new one to follow recommended structure :)

PR #3042, opened 2018-02-23
Optimize Release History view for projects with long release histories.

ewdurbin:
for large project release histories, repetitively rendering url for cubes is costly

Locally, debug toolbar showed a 10x render improvement for `cs18-sidecar` which has >1000 releases.

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -248,6 +248,8 @@
           <h2 class="page-title">Release History</h2>
 
           <section class="release-timeline">
+            {% set blue_cube = request.static_url('warehouse:static/dist/images/blue-cube.svg') %}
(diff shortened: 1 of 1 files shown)


ewdurbin:
btw I have no idea why the call to `request.static_url('warehouse:static/dist/images/{blue,white}-cube.svg')` is so slow.

di:
@ewdurbin I don't think there's any reason for these to be `static_url`, they could be `static_path` which probably would be cheaper.

ewdurbin:
@di quick test locally shows that `static_path` is no faster than `static_url`, so for now I think this PR is good enough to ship?

dstufft:
It would be interesting to see if this is related to stuff we've done in Warehouse, or if the problem can be exposed in Vanilla Pyramid. If the latter it might make sense to open an issue with Pyramid.

PR #3043, opened 2018-02-23
reduce columns pulled for main project view in admin

ewdurbin:
followup to #3036 for "admin.project.detail" route.


ewdurbin:
screen shot 2018-02-23 at 11 44 57 am

memory spike from viewing a big project in admin 😬

PR #3045, opened 2018-02-23
Defer the Release.description column

dstufft:
The vast bulk of the time, when querying against a Release we do not want the description column, because the column is very large and can easily spike memory and make queries take longer, particularly when loading multiple Releases.

Instead we will defer the column so that fetching it requires making another query or explicitly un-defering it. This will cause an extra query to happen on views that do not have direct control over the query (e.g. ones using the Pyramid Context objects) but the impact of a single extra query on these pages is pretty low while the impact of loading this column by default is playing a whack-a-mole game with any query against Release to add a load_only() or defer() option.

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -297,7 +297,7 @@ class Release(db.ModelBase):
         server_default=sql.func.now(),
     )
 
-    # We deferr this column because it is a very large column (it can be MB in
+    # We defer this column because it is a very large column (it can be MB in
     # size) and we very rarely actually want to access it. Typically we only
     # need it when rendering the page for a single project, but many of our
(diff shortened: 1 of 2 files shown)


warehouse/packaging/models.py
```diff
@@ -298,6 +297,15 @@ def __table_args__(cls):  # noqa
         server_default=sql.func.now(),
     )
 
+    # We deferr this column because it is a very large column (it can be MB in
```
ewdurbin: nit: typo `deferr`

dstufft:
I'll create a second PR for the ``load_only()`` calls, wanted to keep them separate because some of the ``load_only()`` calls may still make sense.

ewdurbin:
🚢

PR #3051, opened 2018-02-23
Revert "Add a `Project.latest_release` property (#3034)"

dstufft:
This reverts commit 889a2b216daf5015fdd1964d61ea0220c677d7f1.


warehouse/manage/views.py
```diff
@@ -339,16 +337,7 @@ def _key(project):
     effective_principals=Authenticated,
 )
 def manage_project_settings(project, request):
-    release_count = (
-        request.db.query(Release)
-        .filter(Release.project == project)
-        .count()
-    )
```
di: Still might want to do this eventually, it's kinda wasteful to load all the releases into memory just to count them.

PR #3055, opened 2018-02-23
Document legacy upload API endpoint

brainwane:
Followup to https://github.com/pypa/warehouse/issues/2930#issuecomment-367155334 .

Also improve some headers and tables of contents organization and display.


docs/api-reference/legacy.rst
```diff
@@ -122,3 +122,14 @@ The Simple API implements the HTML-based package index API as specified in `PEP
...
 .. _`PEP 503`: https://www.python.org/dev/peps/pep-0503/
+
+Upload API
+----------
+
+The API endpoint served at `upload.pypi.org/legacy/
+<https://upload.pypi.org/legacy/>`_ is Warehouse's emulation of the
+legacy PyPI upload API. This is the endpoint that tools such as `twine
+<http://twine.readthedocs.io/>`_ and `setuptools
+<https://setuptools.readthedocs.io/en/latest/setuptools.html#upload-upload-source-and-or-egg-distributions-to-pypi>`_
```
di: Technically it's `distutils`, not `setuptools` that's hitting this endpoint.

brainwane: OK. Will fix. Everything else here is accurate?

di: 👍

brainwane:
Fixed.

PR #3056, opened 2018-02-24
Disable delete button if input does not match

### LINKED ISSUE #2884: Disable 'delete confirm' button until confirmation word is correct
We currently have a modal on `warehouse/templates/manage/settings.html`, that allows the user to confirm that they want to delete their project:

screenshot from 2018-02-03 14-43-29

The user is required to enter the project name as an extra security measure.  If they get it wrong, we show them this error:

screenshot from 2018-02-03 14-44-19

## Proposal

It would be really nice if we could `disable` the delete button until the correct project name is given, e.g.

screenshot from 2018-02-03 14-46-02

screenshot from 2018-02-03 14-46-25

## Notes

We will have several other delete confirmation modals on other pages, sometimes with multiple modals on a single page (e.g. delete release, delete file) - so the code will need to be written to take this into account.

di:
This PR fixes #2884 by disabling all 'delete' buttons if the input does not match the expected confirmation string:

feb-24-2018 00-58-27

This PR also introduces [Stimulus](https://stimulusjs.org/) as a lightweight JS framework for our front-end. I've been experimenting with a few front-end frameworks and I think this is a really good fit for our existing codebase.

Most of this PR is refactoring to create macros for our confirmation buttons/respective modals, and some boilerplate to add Stimulus as a dependency and set up the app.

Once that's out of the way, the actual "fix" for #2884 is entirely within ace5aee. We sprinkle our existing HTML with a few carefully placed `data-` attributes, and write a pretty straightforward controller, and we're done, all without a single `document.querySelector`.

The other nice thing is that this gracefully degrades -- if the user has JS disabled and the controller is never started, this behaves just like it did prior to this PR.

---

@nlhkabu There's one thing to tweaked here: it looks like some primary buttons have a higher z-index than the modal overlay:

feb-24-2018 00-55-45

But looking at `_z-index.scss`, everything seems correct, so I'm not sure what's causing this.


nlhkabu:
thanks @di 

After much digging I worked out what the z-index problem is.

The delete account modal is inside the `.callout-block` div.  This div has it's own x-index applied (4), which is creating a new z-index "stacking context" (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

Basically, instead of the modal z-index being 10, it becomes 4.1 (a child z-index of the callout block).  I will push a fix now :)

di:
Ah, thanks @nlhkabu! TIL about child z-indices 🙂

PR #3071, opened 2018-02-27
Add "Your Profile" link to user dropdown

berkerpeksag:
Currently, the only way I found to go to your public
profile is to click the "public profile" links at
https://pypi.org/manage/account/

### DIFF
--- a/warehouse/templates/includes/current-user-indicator.html
+++ b/warehouse/templates/includes/current-user-indicator.html
@@ -29,6 +29,10 @@
           Admin
         </a>
         {% endif %}
+        <a class="dropdown__link" href="{{ request.route_path('accounts.profile', username=request.user.username) }}">
+          <i class="fa fa-cube" aria-hidden="true"></i>
+          Your Profile
+        </a>
         <a class="dropdown__link" href="{{ request.route_path('manage.projects') }}">
           <i class="fa fa-cube" aria-hidden="true"></i>
           Your Projects


nlhkabu:
Hi @berkerpeksag thanks for this PR :)

What is the benefit of linking prominently to this page from here?  All of the information on this page is available via the management UI (with more detail).  Is there something missing from the management UI that you get on this page?

I also think we need to test this with some users before merging it - my concern is that the user might confuse the 'profile' link with the 'account settings' link.

berkerpeksag:
The management UI is only for me so I can't share it on my Patreon page or add it to my CV. Also, adding a link to a user's public profile is a common feature on many websites:

python.org (see "View your public profile"):



GitHub (see "Your Profile"):



Twitter (see "Profile"):



Last.fm (see "View profile"):

warehouse/templates/includes/current-user-indicator.html
```diff
@@ -29,6 +29,10 @@
           Admin
         </a>
         {% endif %}
+        <a class="dropdown__link" href="{{ request.route_path('accounts.profile', username=request.user.username) }}">
+          <i class="fa fa-cube" aria-hidden="true"></i>
```
di: This should probably have a different icon than the "Your Projects" link.

berkerpeksag: Good point, thanks. Do you have any suggestions?

di: We use [`user-circle`](https://fontawesome.com/icons/user-circle?style=solid) for this elsewhere, that might be the best choice but I'll leave it to @nlhkabu.

warehouse/templates/includes/current-user-indicator.html
```diff
@@ -29,6 +29,10 @@
...
+          Your Profile
```
berkerpeksag: Should I rename this to "Your Public Profile"?

warehouse/templates/includes/current-user-indicator.html
nlhkabu: +1 for `user-circle`

warehouse/templates/includes/current-user-indicator.html
nlhkabu: If it fits, that would be better @berkerpeksag

nlhkabu: @berkerpeksag if "Your public profile" does not fit, I think "Public profile" would suffice.  IMO, it's clearer than "Your profile"

berkerpeksag:
Superseded by #3255. I accidentally deleted the fork in my profile instead of an organization so I couldn't edit this PR via GitHub.

PR #3072, opened 2018-02-27
Make project names clickable in projects.html

berkerpeksag:
It would be nice if we could click project's name if we
want to go to its page. IMO, the "View" link is located
at too right to be useful.

### DIFF
--- a/warehouse/templates/manage/projects.html
+++ b/warehouse/templates/manage/projects.html
@@ -27,8 +27,10 @@
     <div class="package-snippet">
       <div class="split-layout split-layout--table split-layout--wrap-on-tablet">
         <div>
-          <h3 class="package-snippet__title">{{ project.name }}</h3>
           {% if release %}
+          <h3 class="package-snippet__title">
+            <a href="{{ request.route_path('packaging.project', name=project.normalized_name) }}">{{ project.name }}</a>
+          </h3>
           <p class="package-snippet__meta">
             <em>Last released on {{ release.created|format_date() }}</em>
           </p>
@@ -36,6 +38,7 @@
             {{ release.summary }}
           </p>
           {% else %}
+          <h3 class="package-snippet__title">{{ project.name }}</h3>
           <p class="package-snippet__meta">
             <em>Created on {{ project.created|format_date() }}</em>
           </p>


nlhkabu:
thanks for this PR @berkerpeksag.  The reason that we're not currently linking from the project name is that it can be confusing for the user - should this link be a 'view' link, or should it be an 'edit' link (given we are in the management UI)?  This is why we have instead chosen to have buttons with explicit names on the right hand side.

User testing has so far shown no problem with the edit/view links located to the right.

berkerpeksag:
> The reason that we're not currently linking from the project name is that it can be confusing for the user - should this link be a 'view' link, or should it be an 'edit' link (given we are in the management UI)?

I personally expect to see similar visual elements behave same in a website. Currently, a project's name links to `/project/<name>/` at https://pypi.org/ and https://pypi.org/user/berkerpeksag/ so if I see a similar visual elements (same box, same icon, project name, project description) in a different page, I'd expect it behave same as others.

On https://pypi.org/:

screenshot from 2018-02-28 11 54 33

On https://pypi.org/user/berkerpeksag/:

screenshot from 2018-02-28 11 55 08

On https://pypi.org/manage/projects/:

screenshot from 2018-02-28 11 55 44

IMO, the "View" button can be be replaced with a link from the project's name in the last image. I've done this via Chrome devtools:

screenshot from 2018-02-28 12 05 14

berkerpeksag:
Here's another example from GitHub:

screenshot from 2018-02-28 12 15 52

Clicking an organization's name always sends me to organization's public page and if I want to do something different, for example leaving an organization (or editing a project in PyPI's context), then I can just click to the "Leave" button.

nlhkabu:
hi @berkerpeksag just to let you know - I'm currently testing this concept in user tests.  So far, we've had mixed results - some people have expected the link to redirect to the public pages, while others have expected it to go to the 'edit' page when they are in the 'edit' part of the app.

I'll keep testing and come back to you when I have more feedback :)

nlhkabu:
Hi @berkerpeksag - Results from user testing: It was a 50/50 split on where this link should go - some users thought it should go to the public page, while others thought that it should go to the edit page, given they were in the edit part of the app.

Given this, I think it's safer if we **do not** add a link.  The buttons on the right were universally understood and utilised.

berkerpeksag:
@nlhkabu can this be closed now?

nlhkabu:
Yes, let's close.  Thanks again for this PR @berkerpeksag

PR #3076, opened 2018-02-27
Datadog instrumentation

ewdurbin:
We probably want to abstract this behind an interface/service.

But for now this demonstrates the sort of stuff we want to be doing.


ewdurbin:
immediately this gets us:

screen shot 2018-02-27 at 10 13 11 am

PR #3099, opened 2018-02-28
Add canonical version column

di:
And set it on new release creation.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2249,8 +2249,7 @@ class TestFileUpload:
                         "platform tag .*",
                         resp.status)
 
-    def test_upload_succeeds_creates_release(
-            self, pyramid_config, db_request, monkeypatch):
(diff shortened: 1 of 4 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -2294,6 +2295,11 @@ def test_upload_succeeds_creates_release(self, pyramid_config, db_request):
         storage_service = pretend.stub(store=lambda path, filepath, meta: None)
         db_request.find_service = lambda svc: storage_service
 
+        canonicalize = pretend.call_recorder(lambda a: 'canonical-version')
+        monkeypatch.setattr(
```
dstufft: We *probably* don't need to monkeypatch here since this is a "pure" function and we can know a priori what the outcome is going to be (similarly to why we wouldn't monkeypatch a call to ``.lower()``or similar). This monkeypatch mostly just serves to make the test more fragile and tied to the implementation details of the method itself. We don't really care in the test *how* the canonical version is being populated, just that it is.

PR #3102, opened 2018-02-28
Data migration for `canonicalize_version`

di:


warehouse/migrations/versions/1fdf5dc6bbf3_data_migration_for_canonical_version_.py
```diff
@@ -0,0 +1,60 @@
...
+                canonical_version=canonicalize_version(release.version),
+            )
+        )
+
+    op.alter_column('releases', 'canonical_version', nullable=False)
+
+
+def downgrade():
+    op.alter_column('releases', 'canonical_version', nullable=True)
+    op.get_bind().execute(releases.update().values(canonical_version=None))
```
dstufft: I would just block downgrading here, I've done it in the past , see:

https://github.com/pypa/warehouse/blob/master/warehouse/migrations/versions/1e2ccd34f539_move_existing_blacklisted_projects_into_.py#L42-L43

PR #3111, opened 2018-03-01
Support exact phrase searching

### LINKED ISSUE #2850: Support searching for an exact phrase
The search box does not appear to have a way to search for a specific exact phrase.

Based on how other search engines work, putting a phrase in quotes should require an exact phrase match.

However, currently searches with and without quotes produce the same result:

https://pypi.org/search/?q=%22Image+processing+routines+for+SciPy%22
and 
https://pypi.org/search/?q=Image+processing+routines+for+SciPy

I would have expected the first one with quotes to only produce the result containing the exact phrase.

---

**Good First Issue**: This issue is good for first time contributors. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

waseem18:
Fixes #2850 

- Removed `index_options="docs"` from [normalized_name attribute](https://github.com/pypa/warehouse/blob/2925b28c3c4c35be0b9c169e2f1d7bdc0bace0ae/warehouse/packaging/search.py#L37) as `index_options="positions"` is the default index option for analyzed strings.
- Added `filter_query` method in `views.py` which returns quoted and unquoted strings.
- Added `form_query` method in `views.py` which returns a list of query objects while taking the arguments as `query_type` and `query`.
- Added test case `test_with_exact_phrase_query`

More information on approach:
- Search query `django authorization` will use `match` query.
- Search query `"django authorization"` will use `match_phrase` query type.
- Search query `"django authorization" automated` will use `match_phrase` query type for the string `django_authorization` and `match` for `automated` and then combines the results.


waseem18:
@di @HonzaKral Will be happy to receive your feedback on this.

HonzaKral:
It looks like this produces a query which is `OR`'d together where maybe the default should be `AND`?

What I think we should do is create a `bool` query for each term (whether quoted or not) and them require all of them to match. That would mean rewriting `_form_query` to:

```python
    def _form_query(self, query_type, query):
        """
        Create a query matching a single term/phrase across all SEARCH_FIELDS.
        """
        return Q('multi_match', fields=SEARCH_FIELDS, query=query, type=query_type)
```

and then `_gather_es_queries` can create the query itself:

```python
    def _gather_es_queries(self, q):
        quoted_strings, unquoted_strings = self._filter_query(q)
        queries = [
            self._form_query("match_phrase", q) for q in quotes_strings
        ] + [
            self._form_query("match", q) for q in unquoted_strings
        ]

        # we require all queries to match - all terms must be present in a document
        query = Q('bool', must=queries)

        # if query was more than a letter, allow use to optionally match on prefix
        if len(q) > 1:
            query = query | Q("prefix", normalized_name=q)
        return query
```

waseem18:
Thanks for taking time and reviewing @HonzaKral I'll follow the suggestions.

However as `multi_match` doesn't accept type as `match` and `match_phrase` we can go with the default type `best_fields` (or `most_fileds`?) and `phrase` (which executes `match_phrase` query).

HonzaKral:
@waseem18 you are right, my apologies, we should be using the default and `phrase`, thanks!

waseem18:
@HonzaKral I've updated the code which addresses the above comments.

di:
@waseem18 Thanks for this. I haven't had time yet to give it a thorough review, but it is on my list.

tests/unit/test_views.py
```diff
@@ -207,16 +208,30 @@ def test_esi_flash_messages():
...
 class TestSearch:
 
+    def _filter_query(self, s):
+        matches = re.findall(r'(?:"([^"]*)")|([^"]*)', s)
+        result_quoted = [t[0].strip() for t in matches if t[0]]
+        result_unquoted = [t[1].strip() for t in matches if t[1]]
+        return result_quoted, result_unquoted
+
+    def _form_query(self, query_type, query):
+        return Q('multi_match', fields=SEARCH_FIELDS,
```
HonzaKral: we should combine the fields with boosts here. `multi_match` accepts a list of fields as strings with an optional `^BOOST` suffix, so something like `["normalized_name^10", "description^5"]`

warehouse/views.py
```diff
@@ -250,20 +251,36 @@ def index(request):
...
+
+        bool_query = Q('bool', must=must)
+
+        # Allow to optionally match on prefix
+        # if ``q`` is longer than one character.
         if len(q) > 1:
-            should.append(Q('prefix', normalized_name=q))
+            bool_query = bool_query | Q('prefix', normalized_name=q)
+
+        query = request.es.query('bool', must=bool_query)
```
HonzaKral: there is no need to wrap the query in a `bool`. It doesn't do anything.

HonzaKral:
I left some comments on the code, my apologies as some of the confusion there was caused by my incomplete answer earlier where I just produced hints without mentioning how the finished thing should fit together.

I am also confused about the code duplication between `tests/unit/test_views.py` and `warehouse/views.py` - it seems like the same code is used in both where it should probably be abstracted away into it's own method/function somewhere in `views` and imported into the tests.

please feel free to reach out directly and we can cooperate more on this this and next week as I finally have more time.

Thank you!

warehouse/views.py
waseem18: @HonzaKral  The reason why I did it is because in the next line we're using the `suggest` method on query and on the `Q` object i.e `bool_query` doesn't have a suggest method and would throw an error.

waseem18:
Thanks for the review and suggestions @HonzaKral  I'll address the reviews

tests/unit/test_views.py
waseem18: I missed this - I'll put it in.

warehouse/views.py
HonzaKral: yes, you need to use ` Search` object, but that accepts any query, so just `request.es.query(query)` should do the trick.

waseem18:
@HonzaKral I've addressed the reviews - I see a test failing which is no way related to this issue. I'll look into it.

brainwane:
@waseem18 is this ready for re-review?

waseem18:
Yes @brainwane

di:
At first glance this looks good, however I am going to back-burner fully reviewing/merging this PR until post-launch. I've put it in the appropriate milestone.

Sorry @waseem18, and thanks in advance for your patience.

brainwane:
@robhudson if you have any time to spend looking at this, that would be great.

HonzaKral:
@waseem18 Looks good to me! There is still some duplication of code between tests and views but the search specific parts look good! Thanks for working on this!

tests/unit/test_views.py
```diff
@@ -208,15 +208,18 @@ def test_esi_flash_messages():
...
-                kw["boost"] = SEARCH_BOOSTS[field]
-            queries.append(Q("match", **{field: kw}))
+        quoted_strings, unquoted_strings = views.filter_query(q)
+        queries = [
+            views.form_query("phrase", i) for i in quoted_strings
+        ] + [
+            views.form_query("best_fields", i) for i in unquoted_strings
+        ]
+
+        query = Q('bool', must=queries)
```
HonzaKral: this block is the same as in `views.py`, could we perhaps make it into a function in views that we would import here? Otherwise we are not really testing the implementation.

waseem18:
Addressed reviews @honzakral @di

HonzaKral:
Thank you, @waseem18 for the work!

waseem18:
Your welcome and thank you for you helping me and reviewing this PR.

PR #3120, opened 2018-03-02
Document tests depending on `manifest.json`

brainwane:
Document workaround for #1536.

Along the way, improve `docs/development/getting-started.rst` with more specific Sphinx and `.rst` markup.

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -33,6 +33,8 @@ Quickstart for Developers with Docker experience
 
 View Warehouse in the browser at ``http://localhost:80/``.
 
+.. _dev-env-install:
+
 Detailed Installation Instructions
 ----------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/development/getting-started.rst
```diff
@@ -356,6 +358,16 @@ To run all tests, all you have to do is:
...
 
+.. tip::
+   Currently, running ``make tests`` from a clean checkout of
+   Warehouse (namely, before trying to compile any static assets) will
+   fail multiple tests because the tests depend on a file
+   (:file:`/app/warehouse/static/dist/manifest.json`) that gets
+   created during deployment. So until we fix `bug 1536
+   <https://github.com/pypa/warehouse/issues/1536>`_, you'll need to
+   install Warehouse in a developer environment before running tests;
+   see :ref:`dev-env-install` for instructions.
```
di: I think it might be worthwhile to be explicit about what needs to happen to create this file (once the dev environment is set up), which is that `make serve` needs to be run -- just setting up the environment is not enough.

PR #3133, opened 2018-03-03
Fix infinite redirect...

di:
“Two things are infinite: the universe and human stupidity; and I'm not sure about the universe.”
― Albert Einstein

### DIFF
--- a/warehouse/packaging/views.py
+++ b/warehouse/packaging/views.py
@@ -67,6 +67,14 @@ def project_detail(project, request):
 def release_detail(release, request):
     project = release.project
 
+    if release.version != request.matchdict.get("version", release.version):
+        return HTTPMovedPermanently(
+            request.current_route_path(
+                name=project.name,
+                version=release.version,
+            ),
+        )
+
     if project.name != request.matchdict.get("name", project.name):
         return HTTPMovedPermanently(
             request.current_route_path(name=project.name),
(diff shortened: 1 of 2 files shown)


warehouse/packaging/views.py
```diff
@@ -67,13 +67,19 @@ def project_detail(project, request):
 def release_detail(release, request):
     project = release.project
 
-    if not {project.name, release.version} <= set(request.matchdict.values()):
+    if release.version != request.matchdict.get("version", release.version):
```
dstufft: It might be reasonable to add a comment to explain why this is using ``.get()``, I was looking at it and couldn't figure it out at first until I remembered the ``project_detail()`` is implemented in terms of ``release_detail()``.

di: Yep. I'll add that separately since I'd like to get this merged ASAP.

dstufft: Yea that's fine, you might also mention that we're using this case to ensure both project name *and* version are correct (and thus reducing redirects). I almost made a comment about removing the ``name=project.name`` kwarg since it will inherit it from the request itself, till I realized it's saving an extra redirect in cases where both the version and the project need to be redirected to their "real" value.

di: Yeah, that was my original goal... unfortunately I was trying to be a little too clever. 🙂

PR #3136, opened 2018-03-05
Update "show password" checkbox to use Stimulus

### LINKED ISSUE #3087: Update "show password" checkbox to use Stimulus
In #3056 I added [Stimulus](https://stimulusjs.org/) as a front-end framework, however there are still a number of places where we're not using it.

One of them is the "Show Password" checkbox, which reveals the password fields on the current page. Right now, it's managed by this piece of Javascript.
https://github.com/pypa/warehouse/blob/f9acc7c8cea658618d6c8b4ce306eb25dc87a2a9/warehouse/static/js/warehouse/index.js#L130-L151

It'd be great to take this and convert this into a Stimulus controller. 

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

aalmazan:
Fixes #3087. 

This is my first issue -- please let me know of any feedback, especially with controller/target naming.


warehouse/static/js/warehouse/controllers/password_controller.js
```diff
@@ -0,0 +1,30 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+import { Controller } from "stimulus";
+
+export default class extends Controller {
+  static targets = ["showPassword", "newPassword", "passwordConfirm"];
```
di: The `data-target` attribute doesn't need to be unique. If we set it to be something like `data-target="password"` for all these password fields, we can access them all by using `this.passwordTargets` (note the pluralization) which would greatly simplify the controller to something like:

```js
togglePasswords() {
  for (let field of this.passwordTargets) {
    field.type = this.showPasswordTarget.checked ? "text" : "password";
  }
}
```

This would also let us reuse this on other forms/pages that contain passwords.

warehouse/static/js/warehouse/index.js
```diff
@@ -139,15 +139,6 @@ docReady(() => {
       showPasswordCheck.checked = false;
```
di: I think we can add a `connect()` function to the controller (which will be called when it first loads) and move these 'initialization' bits there so that we can completely eliminate this entire block of JS.

warehouse/templates/accounts/register.html
```diff
@@ -70,46 +70,50 @@ <h1 class="page-title">Create an account on PyPI</h1>
...
-          </p>
-          <p class="form-group__help-text">
-            <strong>Password strength:</strong>
-            <span class="password-strength">
+        <div data-controller="password">
+          <div class="form-group">
+            <div class="split-layout">
+              <label for="password" class="form-group__label">Password</label>
+              <label for="show-password">
+                <input data-action="click->password#showPasswords" data-target="password.showPassword"
```
di: Also, I think technically the better event to bind to here is the `change` event, and since this is the default event for an input element, this can just become `data-action="password#togglePasswords"`

warehouse/static/js/warehouse/controllers/password_controller.js
```diff
@@ -0,0 +1,30 @@
...
+  passwordFields = ["newPasswordTarget", "passwordConfirmTarget"];
+
+  showPasswords() {
```
di: Since this function is actually showing _and_ hiding passwords, I think a name like `togglePasswords` might make more sense.

warehouse/templates/accounts/register.html
```diff
@@ -70,46 +70,50 @@ <h1 class="page-title">Create an account on PyPI</h1>
...
-          {% if form.password_confirm.errors %}
-          <ul class="form-errors">
-            {% for error in form.password_confirm.errors %}
-            <li>{{ error }}</li>
-            {% endfor %}
-          </ul>
-          {% endif %}
+          <div class="form-group">
+            <label for="password_confirm" class="form-group__label">Confirm Password</label>
+            {{ form.password_confirm(placeholder="Confirm password", required="required", class_="form-group__input", **{ 'data-target': 'password.passwordConfirm' }) }}
```
di: WTForms will automatically change underscores to dashes in parameters it doesn't recognize, so this could just be `data_target="password.passwordConfirm"`

brainwane:
@aalmazan thanks for your contribution! Is there any chance we'll see you [at the PyCon sprints in May](https://us.pycon.org/2018/community/sprints/)?

aalmazan:
@di Thank you for the feedback! I must have missed a few things while going through Stimulus's docs. Thank you for pointing them out. Changes are commited on: 7a0de64.

warehouse/templates/accounts/register.html
```diff
@@ -70,46 +70,50 @@ <h1 class="page-title">Create an account on PyPI</h1>
...
+            <div>
+              {{ form.new_password(placeholder="Select a password", required="required", class_="form-group__input", autocomplete="new-password", data_target="password.password") }}
+            </div>
+            <p class="form-group__help-text">
+              Choose a strong password that contains letters (uppercase and lowercase), numbers and special characters.
+              Avoid common words or repetition.
+            </p>
+            <p class="form-group__help-text">
+              <strong>Password strength:</strong>
+              <span class="password-strength">
```
di: Looks like the following four lines need more indentation.

warehouse/static/js/warehouse/controllers/password_controller.js
```diff
@@ -0,0 +1,34 @@
...
+  static targets = ["showPassword", "password"];
+
+  connect() {
+    if (this.passwordTargets && this.showPasswordTarget) {
+      // Reset these so they don't persist between page reloads
+      for (let field of this.passwordTargets) {
+        field.type = "password";
+        this.showPasswordTarget.checked = false;
+      }
+    }
```
di: This function is only going to get called if we've set the `data-controller="password"` attribute on some field, so we probably don't need the checks anymore as this will only run on pages that we already know have the targets. It's also unnecessarily setting `this.showPasswordTarget.checked`  for every field instead of just once (but this was an issue from the original implementation). I think this can just become something like:

```js
connect() {
    // Reset the targets so they don't persist between page reloads
    this.showPasswordTarget.checked = false;
    this.togglePasswords();
}
```

aalmazan:
@di  Ah, apologies. Didn't pay attention that the checkbox was on the other pages. These should now be updated, as well as the other 2 requested changes.

di:
Thanks @aalmazan!

brainwane:
Thanks, @aalmazan! Would you like a suggestion for a next issue to work on?

Issue #2935, opened 2018-02-13
Reach out to 3rd party services about new pypi.org domain


di:
We should reach out to any 3rd party service that depends on the `pypi.python.org` domain and encourage them to update to `pypi.org`.

A list of services that currently depend on `pypi.python.org`:

- [x] https://badge.fury.io/for/py (clicking the badge directs the user to `pypi.python.org`)
- [x] https://pyup.io/ (PRs currently link to `pypi.python.org`)
- [x] [TravisCI](https://docs.travis-ci.com/user/deployment/pypi/) ("can automatically release your Python package to [PyPI](https://pypi.python.org/) after a successful build." May fix travis-ci/dpl#377?)
- [x] [CircleCI](https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/) (similar)
- [x] [sphinx-pypi-upload](https://bitbucket.org/jezdez/sphinx-pypi-upload/)
- [x] [the cheeseshop extension to Sphinx](https://bitbucket.org/birkenfeld/sphinx-contrib/src/8e295053a27d5914127fbb01e2a8fb57421897a7/cheeseshop/?at=default) (["easily link to PyPI packages"](http://www.sphinx-doc.org/en/master/develop.html#extensions))
- [x] [Pydoc.io](https://blog.readthedocs.com/announcing-pydoc-io/) ([site](https://www.pydoc.io/), [code](https://github.com/rtfd/pydoc.io); grabs PyPI packages and builds API docs)
- [x] possibly [PythonAnywhere](https://www.pythonanywhere.com/) which gives people hosted Python environments & links to [projects on legacy PyPI](https://www.pythonanywhere.com/batteries_included/) & may run a mirror? to handle [`pip install`](https://help.pythonanywhere.com/pages/InstallingNewModules)
- [x] the Debian [`watch`](https://wiki.debian.org/debian/watch) folks and [`devscripts` (including `uscan`)](https://manpages.debian.org/stretch/devscripts/uscan.1.en.html) folks, such as Piotr Roszatycki, as they hit legacy PyPI ([`uscan`](https://manpages.debian.org/stretch/devscripts/uscan.1.en.html#PyPI), [`watch`](https://wiki.debian.org/debian/watch#PyPI)) and produce pypi.debian.net ([example](http://pypi.debian.net/arrow)) and [have asked about Warehouse supporting `uscan` more cleanly](https://github.com/pypa/warehouse/issues/358#issuecomment-337233792)
[...]

PR #3153, opened 2018-03-07
migration guide for third-party services

brainwane:
a step towards #2935

### DIFF
--- a/warehouse/__about__.py
+++ b/warehouse/__about__.py
@@ -37,7 +37,7 @@ else:
     __commit__ = None
 
 __author__ = "The Python Packaging Authority"
-__email__ = "donald@stufft.io"
+__email__ = "admin@mail.pypi.org"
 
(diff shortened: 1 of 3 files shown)


di:
LGTM, what is this blocked on? BTW you have linting errors.

brainwane:
As I was testing this I realized the copyright date on the documentation was 2015 because that's what was in `__about__.py`, so I updated that as well.

warehouse/__about__.py
```diff
@@ -37,7 +37,7 @@
     __commit__ = None
 
 __author__ = "The Python Packaging Authority"
-__email__ = "donald@stufft.io"
+__email__ = "ernest@python.org"
```
di: Hmm, maybe we should make this `admin@mail.pypi.org` instead?

brainwane: Fixed. Ready for re-review.

PR #3155, opened 2018-03-07
Send email when a new collaborator has been added to the project.

### LINKED ISSUE #1000: Email all owners when a new owner/maintainer is added
To make account compromise more obvious, it would be great if all owners of a package were emailed when a new owner or maintainer was added to a package.

In addition, if a user is removed from a role, then that user should also be emailed (eg to prevent an attacker who has compromised one owner, from silently removing other owners of a package, prior to uploading a new malicious package, thereby circumventing #997).

Mariatta:
- send the email to the newly added collaborator
- send the email to other owners
- email not sent to the person who added the collaborator

Fixes #1000


warehouse/manage/views.py
```diff
@@ -547,6 +549,22 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
                     submitted_from=request.remote_addr,
                 ),
             )
+
+            owners = request.db.query(Role).filter(Role.role_name == 'Owner',
+                                                   Role.project == project)
+            email_recipients = [owner.user.email for owner in owners]
```
di: We can reduce some of the memory required to do this by moving the `.user` part into the DB with a join:

```python
owners = (
    request.db.query(Role)
    .join(User)
    .filter(Role.role_name == 'Owner', Role.project == project)
)
owner_emails = [owner.email for owner in owners]
```

That way we just `User`s back from our query, not `Roles`.

warehouse/templates/email/collaborator-added.body.txt
```diff
@@ -0,0 +1,5 @@
+'{{ username }}' has been added as a collaborator with '{{ role }}' role to the PyPI
+project: '{{ project }}' by '{{ submitter }}'.
```
di: I've been taking care to keep these emails at a fixed 80-character limit, which requires a bit of crafting of the template.

For this, we probably want to start out with the same "greeting" every time rather than the username as well.

So maybe something like:

```
A new collaborator has been added to a project you own on PyPI:

  Username: {{ username }}
  Role: {{ role }}
  Collaborator for: {{ project }}
  Added by: {{ submitter }}

If this was a mistake, you can reply to this email directly to communicate with
the PyPI administrators.
```

warehouse/email.py
```diff
@@ -19,15 +19,18 @@
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, body, recipients, subject):
+def send_email(task, request, body, recipients, subject, send_to_bcc=False):
```
di: I think I would prefer if this more closely mimicked the arguments that `Message` takes, so something like:

```python
def send_email(task, request, body, subject, recipients=None, bcc=None)
```
(FYI, changing the order here might require updates elsewhere where we're passing this positional arguments)

And then since the default arguments for `Message` are already `None`, we can call it like:

```python
message = Message(
    body=body,
    recipients=recipients,
    bcc=bcc,
    sender=...,
    subject=subject,
)

File: warehouse/templates/email/added-as-collaborator.body.txt [hunk 4]
```diff
@@ -0,0 +1,3 @@
+You have been added as {{ role }} to the PyPI project {{ project }} by {{ submitter }}.
+
+Congratulations!
```
di: This needs the same "you can reply to the admins" message.

warehouse/manage/views.py
```diff
@@ -551,6 +552,31 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
+                form.role_name.data,
+                owner_emails
+            )
+
+            send_added_as_collaborator_email(request,
+                                             request.user,
+                                             project.name,
+                                             form.role_name.data,
+                                             user.email
+                                             )
```
di: This is a nitpick, but could you make the indentation here consistent? (e.g. like the call to `send_collaborator_added_email` above)

warehouse/manage/views.py
```diff
@@ -551,6 +552,31 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
                     submitted_from=request.remote_addr,
                 ),
             )
+
+            owners = (
+                request.db.query(Role)
+                .join(User)
+                .filter(Role.role_name == 'Owner', Role.project == project)
+            )
```
di: Sorry, I gave you a bad example query before. This should be:

```python
owners = (
    request.db.query(User)
    .join(Role.user)
    .filter(Role.role_name == 'Owner', Role.project == project)
)

warehouse/email.py
Mariatta: Ok it makes sense to do it that way. Thanks.
That'll be quite a big change, since this is being called everywhere as positional argument. 
I will prepare a separate PR just to update `recipients` into keyword only argument in `send_email` method.

brainwane:
@Mariatta looks like [tests are failing](https://travis-ci.org/pypa/warehouse/jobs/351780712) -- take a look?

Mariatta:
Thanks! The tests are passing now. Please re-review when you have the chance.

lgh2:
Hi @Mariatta!  I tested this PR and it looks like the emails are being sent to the new collaborator, but not to the other owners.

Mariatta:
Thanks @lgh2! With the latest commit (https://github.com/pypa/warehouse/pull/3155/commits/ca04e7c7f6368b175ed0805c1aff138f0fd61e46) the other owners should receive the email notification now.

di:
Thanks @Mariatta!

Mariatta:
Thanks for the patience in reviewing this :)

PR #3158, opened 2018-03-07
Added functionality to send email whenever primary email is changed

### LINKED ISSUE #998: Send confirmation when email address changed
Currently on PyPI, if a user changes their contact email address, there is no notification to the old email address, meaning account compromise can go unnoticed.

This means that an attacker can silently prevent other important security notifications (eg #997) from being sent to the real owner.

waseem18:
Fixes #998 

- Added code to send an email to primary email whenever primary email is changed.
  - If `A` is the current primary email and email `B` has been made primary then an email will be sent to the email `A`.
- Added email templates
- Added tests


waseem18:
@di Can you please review this PR?!

Thanks

warehouse/templates/email/primary-email-change.body.txt
```diff
@@ -0,0 +1,5 @@
+Someone, perhaps you, has changed the primary email for your PyPI account
+'{{ username }}'.
```
di: I think it'd be nice here if this email included the emails it changed from/to. So something like 

```
The primary email for your PyPI account:

  '{{ username }}'

been changed from this address:

  {{ old_email }}

to this address:

  {{ new_email }}

If you did not make this change, you can reply to this email directly to
communicate with the PyPI administrators.
```

tests/unit/manage/test_views.py
```diff
@@ -379,8 +379,8 @@ def test_change_primary_email(self, monkeypatch, db_request):
         user = UserFactory()
         old_primary = EmailFactory(primary=True, user=user)
         new_primary = EmailFactory(primary=False, verified=True, user=user)
-
```
di: Also, this is a very small issue, but it seems like your PR removes some empty lines like this one here -- could you make sure they remain?

waseem18:
Addressed reviews @di

tests/unit/manage/test_views.py
```diff
@@ -389,7 +391,14 @@ def test_change_primary_email(self, monkeypatch, db_request):
         )
         view = views.ManageAccountViews(db_request)
 
+        send_email = pretend.call_recorder(lambda *a: None)
+        monkeypatch.setattr(
+            views, 'send_primary_email_change_email', send_email
+        )
         assert view.change_primary_email() == view.default_response
+        assert send_email.calls == [
+            pretend.call(db_request, db_request.user, email.email)
```
waseem18: @di After we merged this PR, I created [another PR](https://github.com/pypa/warehouse/pull/3111) with the latest master - but I see my [Travis test failing](https://travis-ci.org/pypa/warehouse/jobs/350764085) at this line which is no way related to #3111 . I'm not able to replicate the test failure locally.  

Could you please help me with this? Thanks

waseem18: Strange! Now when I updated my branch there wasn't any test failure.

tests/unit/manage/test_views.py
```diff
@@ -379,8 +379,10 @@ def test_change_primary_email(self, monkeypatch, db_request):
         user = UserFactory()
         old_primary = EmailFactory(primary=True, user=user)
         new_primary = EmailFactory(primary=False, verified=True, user=user)
+        email = EmailFactory(user=user)
```
di: @waseem18 The issue is that we're adding an extra email here. Since we've already created the `old_primary` email above, and since `EmailFactory` defaults to making the created email primary as well, we had created two primary emails for the user in this test.

The way `User.email` works is that it assumes there might be multiple primary emails, and returns the first primary email it finds for the user (not totally why this is the case, but I'm guessing backwards compatibility with legacy).

In this test, most of the time, `user.email` would refer to this `email`, but sometimes (not sure why) they might be created out of order, so `user.email` would be `old_primary` instead, causing the test to fail.

The fix is in #3178: just create one primary email for the user, and test that email is the one that gets the message.

tests/unit/manage/test_views.py
di: See my comment here: https://github.com/pypa/warehouse/pull/3158#discussion_r173226379

tests/unit/manage/test_views.py
waseem18: Thank you so much for the detailed response and the PR @di

di: No worries! Sorry that I missed it in review and you spent so much time trying to debug it.

PR #3169, opened 2018-03-08
Advise project maintainers on Project Links

### LINKED ISSUE #3119: Add help text for maintainers about project URL list
From IRC conversation with @alanbato and @ewdurbin and @di. 

Add help text in what's currently "Project Settings" explaining that the project URL list comes from the package's `project-url` field as exaplained in https://packaging.python.org/specifications/core-metadata/#project-url-multiple-use and pointing to the bit of packaging.python.org that says where to put this info (`setup.py`?).

~~~
<alanbato> I couldn't find any way to change to change the repo linked in the project description
<sumanah> alanbato: so that's the url field in the setup.py file
<sumanah> https://packaging.python.org/tutorials/distributing-packages/#url
<sumanah> interestingly it looks like you can give an OrderedDict if you want to provide multiple URIs! although there's a bug related to that right now https://github.com/pypa/warehouse/issues/3097
<alanbato> Mmm, then I should change the link in the setup.py file and it should be changed, right?
<sumanah> alanbato: that is my understanding, yes!
<alanbato> Alright, I'll try that, thanks! I'll go take a look at the twine documentation :)
<sumanah> alanbato: maybe we should put a bit of help text in https://pypi.org/manage/project/<project>/settings/ explaining this.
<di_codes> sumanah: actually `url` (aka `Home-page`) is different than `project_urls` (aka `Project-URL`)
<sumanah> OHH
<di_codes> `url` can’t be a `OrderedDict`
<sumanah> di_codes: ok so I'm reading https://packaging.python.org/specifications/core-metadata/ more closely now.
<alanbato> sumanah: Yes, that's where I started looking and couldn't find the option to edit it. But I'm not sure we want to be able to override the setup.py fields from the website
<sumanah> alanbato: oh I am not suggesting we override!
<sumanah> alanbato: I'm suggesting that, if Project Settings is where maintainers expect to be able to figure it out, then that's where the help text should live :)
<EWDurbin> alanbato: release data is immutable. In order to update that information you’d need a new release.
<alanbato> Then maybe we can work to improve docs and help text in the settings page, so new project maintainers like me do not get confused :)
<EWDurbin> alanbato: agreed :)
<sumanah> I think that's a fine idea alanbato. Shall I file the issue or shall you?
~~~

brainwane:
Add help text in Project Settings (where maintainers would naturally look for it) telling them how to update the Project Links sidebar using fields in `setup.py`.

Fixes #3119.

project-url-help-screenshot

### DIFF
--- a/warehouse/templates/manage/settings.html
+++ b/warehouse/templates/manage/settings.html
@@ -21,6 +21,21 @@
   <h2>Project Settings</h2>
 
   <div class="callout-block">
+    <h3>Project Links</h3>
+    <p>
+      To set the Project Links in the '{{ project.name }}' sidebar for
+      your next release, use
+      the <a href="https://packaging.python.org/tutorials/distributing-packages/#url"><tt>url</tt></a>
+      and <a href="https://packaging.python.org/tutorials/distributing-packages/#project-urls"><tt>project_urls</tt></a>
+      fields in your <tt>setup.py</tt> file. Updating these fields
+      will not change the Project Links for past
+      releases. See <a href="https://packaging.python.org/tutorials/distributing-packages/">the
+      Python Packaging User Guide</a> for more help.
+    </p>
+  </div>
+
+
+  <div class="callout-block">
     <h3>Delete Project</h3>
     <p>
     {% if project.releases %}


di:
I think this PR runs the risk of being a little too specific. There's a lot of other things on the "project" page besides the project URLs (like author name/email, classifiers, etc) that come from the metadata and cannot be changed w/in Warehouse.

It might make sense to generalize this message a bit for users that haven't quote grokked the fact that "this metadata comes from my distribution and only my distribution".

PR #3175, opened 2018-03-08
Advise packagers on requesting upload size limit increase

brainwane:


warehouse/forklift/legacy.py
```diff
@@ -993,10 +993,11 @@ def file_upload(request):
                     raise _exc_with_message(
                         HTTPBadRequest,
                         "File too large. " +
-                        "Limit for project {name!r} is {limit}MB".format(
+                        "Limit for project {name!r} is {limit}MB. ".format(
                             name=project.name,
-                            limit=file_size_limit // (1024 * 1024),
-                        ))
+                            limit=file_size_limit // (1024 * 1024)) +
+                        "See https://pypi.org/help/#file_size_limit",
```
di: Instead of hardcoding this link, could you generate it like `request.route_url('help', _anchor='file-file-size-limit')`, otherwise it will be wrong for TestPyPI.

warehouse/templates/pages/help.html
```diff
@@ -152,6 +154,17 @@ <h2>{{ project_name_claim() }}</h2>
         </p>
       </section>
 
+      <section id="file_size_limit" class="common-question">
```
di: I think our frontend generally prefers `kebab-case` for ids rather than `snake_case`.

warehouse/templates/pages/help.html
```diff
@@ -152,6 +154,17 @@ <h2>{{ project_name_claim() }}</h2>
...
+        <h2>{{ file_size_limit() }}</h2>
+        <p>
+          If you can't upload your release to PyPI because you're hitting the upload file size limit, please <a href="https://github.com/pypa/warehouse/issues/new">file an issue</a> and tell us:</p>
+        <ul>
+          <li>The name of the project</li>
+          <li>The size of your release (example: 300 MB)</li>
```
di: 300MB is kind of on the larger end of our normal requests. Instead of providing an example we could just say:

> The size of your release, in megabytes

warehouse/templates/pages/help.html
```diff
@@ -152,6 +154,17 @@ <h2>{{ project_name_claim() }}</h2>
...
+          <li>Which index/indexes you need the increase for (PyPI, Test PyPI, or both)</li>
```
di: It's probably also worth adding somewhere (perhaps here) that if they are trying to upload the _first_ release for their project, and that release is larger than the default limit, they need to make an initial release that is smaller than the default liit before we are able to increase the limit for the given name.

warehouse/forklift/legacy.py
di: (Also I realize there are a few other places where this isn't being done... perhaps you'd like to fix those too? 🙂)

warehouse/templates/pages/help.html
brainwane: Got it. Fixing.

warehouse/templates/pages/help.html
brainwane: Got it. Fixing.

warehouse/forklift/legacy.py
brainwane: AHA. I tried to figure that out - evidently not hard enough. :) Thanks for the tip. Fixing.

warehouse/templates/pages/help.html
brainwane: Thanks. Fixed.

brainwane:
Failing a test because I am not properly plumbing the route call through Pyramid's URL mapping:

~~~
warehouse/forklift/legacy.py:1001: in file_upload
    'help', _anchor='file-size-limit'
../lib/python3.6/site-packages/pyramid/url.py:261: in route_url
    mapper = reg.getUtility(IRoutesMapper)
~~~

Looks like we should also fix https://github.com/pypa/warehouse/blob/master/warehouse/forklift/legacy.py#L833 to use URL mapping.

Going to fix after lunch, would welcome tips.

di:
Yes, you'll need to mock out `route_url` in the test, something like this:

```
db_request.route_url = pretend.call_recorder(
    lambda route, **kw: "/the/help/url/"
)
```

brainwane:
@di Thanks for the tip. I have fixed the test for the filesize message, and will look at the other upload error messages that need URL substitution in a separate PR. I believe this is ready to merge.

PR #3181, opened 2018-03-08
Only calculate streaming in conditional logic

SethMichaelLarson:
I'm also thinking we might be able to streamline the buffering logic in `conditional_http_tween` with the `compression_tween` as they both do buffering it seems. If that change is desirable I'll make it in this PR as well.

### DIFF
--- a/warehouse/cache/http.py
+++ b/warehouse/cache/http.py
@@ -73,8 +73,6 @@ def conditional_http_tween_factory(handler, registry):
         if response.last_modified is not None:
             response.conditional_response = True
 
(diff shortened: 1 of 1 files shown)


di:
Hi @SethMichaelLarson, I'm not sure I understand the goal of this change. What is this fixing?

SethMichaelLarson:
Since this is run per-request we could combine the buffering logic in the compression tween and the identical buffering logic in the compression tween. I switched the `if` conditions around so the faster one breaks before the "slower" one.

PR #3182, opened 2018-03-08
Cascade Role deletion

di:
Similar to #2979, cascade deletion to roles if the referenced project or user gets deleted.


ewdurbin:
a nice simplification rather than explicitly deleting. 👍

My understanding is that if a project or user is deleted, their rows in the `roles` table will be as well.

PR #3183, opened 2018-03-08
Make two upload tests URL-independent

brainwane:
Followup to #3175. Change the upload permissions check and project name/Python standard library collision check error messages refer to the URL of the PyPI in question, thus ensuring that we send error messages that, as appropriate, refer to pypi.org, test.pypi.org, or any other future PyPI.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -1885,8 +1885,24 @@ class TestFileUpload:
             ),
         })
 
-        with pytest.raises(HTTPForbidden):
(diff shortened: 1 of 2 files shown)


brainwane:
It was reasonably mechanical to make these changes, so if we approve this, I'll set up new Good First Issue bug reports to fix `test_fails_with_admin_flag_set`, `test_upload_requires_verified_email`, and `test_fails_with_invalid_names`, and point to this PR for guidance on how to do so.

di:
👍 to the good first issues.

Issue #1190, opened 2016-05-17
Add support for changing username


jleclanche:
This is a sorely needed feature on pypi. Let me know if I can help implement it in warehouse.

brainwane:
This is a new feature that legacy PyPI does not support, and I also think it would be a good idea -- you're welcome to help implement it (although it might depend on #414 ?).

nlhkabu:
#414 is now closed, so this is fine to go ahead.

brainwane:
@jleclanche thanks for your note and sorry for the slow response!

The folks working on Warehouse have gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and have kicked off work towards [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) -- the most urgent task is to improve Warehouse to the point where we can redirect pypi.python.org to pypi.org so the site is more sustainable and reliable. Since, as @di noted in https://github.com/pypa/warehouse/pull/2833#issuecomment-361297268 , this feature isn't something that the legacy site has, I've moved it to a future milestone. But we'd be happy if a volunteer helped make it happen sooner than that!

Thanks and sorry again for the wait.

brainwane:
@jleclanche You mentioned you'd be interested in implementing it -- right now is a great time to [get started working on Warehouse](https://warehouse.readthedocs.io/development/getting-started/), since there are multiple people working actively on Warehouse for the next few months who can help you get started and review your code. [Ernest W. Durbin III has multiple livechat mentorship slots available each week](https://twitter.com/EWDurbin/status/955415184339849217) to help if you're interested in joining us!

brainwane:
We'd appreciate a volunteer's help with this feature! if you have questions along the way as you work on this, please feel free to ask them here, in [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

In [today's core Warehouse developers' meeting](https://wiki.python.org/psf/PackagingWG/2018-03-06-Warehouse) we decided to pare down our near-future milestones on [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) so they really only contain the essential bugfixes and features we need to launch, replace legacy PyPI, and shut down the old site. So I'm moving this issue into a milestone further in the future.

di:
[...]

PR #3189, opened 2018-03-08
User cache purge

### LINKED ISSUE #2861: Deleting a project should purge user's profile cache
Currently deleting a project does not purge the profile page of the project's owners/maintainers, which contains a link to the project.

di:
Fixes #2861.

This PR enables purging pages related to the `User` model whenever either the `User` model changes, or any projects/releases related to that user.

To do so, we need to be able to generate cache keys from some iterable attribute on the changed object (like `Project.users` or `Release.project.users`).


warehouse/packaging/__init__.py
```diff
@@ -38,13 +40,36 @@ def includeme(config):
     # Register our origin cache keys
     config.register_origin_cache_keys(
         Project,
-        cache_keys=["project/{obj.normalized_name}"],
-        purge_keys=["project/{obj.normalized_name}", "all-projects"],
+        cache_keys=[
+            key_factory("project/{obj.normalized_name}"),
+            key_factory("user/{itr.username}", iterate_on='users'),
```
dstufft: What changes to a user would get reflected on pages that use the ``Project`` model?

warehouse/packaging/__init__.py
```diff
@@ -38,13 +40,36 @@ def includeme(config):
...
+            key_factory("all-projects"),
+        ],
     )
     config.register_origin_cache_keys(
         Release,
-        cache_keys=["project/{obj.project.normalized_name}"],
-        purge_keys=["project/{obj.project.normalized_name}", "all-projects"],
+        cache_keys=[
+            key_factory("project/{obj.project.normalized_name}"),
+            key_factory("user/{itr.username}", iterate_on='project.users'),
```
dstufft: Same question as above, what change to a user requires invalidating the cache for anything that uses the ``Release`` model?

warehouse/packaging/__init__.py
di: Technically if the user changes their primary email, their avatar would change, which is shown on the "Maintainers" section for project/release pages.

Also if we ever ship #1190, the URLs to maintainers would be wrong.

dstufft: My primary concern is that there is a limit to how many surrogate keys (what these are) we can attach to a single HTTP response. Since we have no control over how many users can get added to a single project, that would mean that a page may have more surrogate keys than we can support in a single response (and I believe what happens then is that we just silently lose some of the surrogate keys).

I *think* probably it doesn't make sense to support ``iterate_on`` for cache keys because we can't control how many keys are going to get added. However ``iterate_on`` for purge keys makes total sense, and this example here could easily be implemented by, instead of adding a cache key for user to each page, adding a purge key to ``User`` that is something like ``key_factory("project/{itr.normalized_name}", iterate_on="projects")``. That should have the same effect, without the unbounded surrogate key problem.

dstufft: Oh, and if you make that change, add a comment please about why one supports it and one doesn't, otherwise it'll probably be an attractive nuisance for people wanting to add support for it to cache keys and relying on tribal knowledge for why.

di:
@dstufft This is ready for re-review.

PR #3202, opened 2018-03-09
Make notification bar + callout block components dismissable (optionally)

aalmazan:
This is a quick first pass for this issue. I do have a question before applying it to the rest of the notification templates: 

If we need to remember that the user dismissed more than one notification (that is shown to all users), we'll need to keep a track of those notifications via some sort of unique key? Or will there only be one, sort of, "global" notification at one time? 

Once that gets figured out, as well as any other feedback, I'll start adding in the Stimulus handling on the elements in `warehouse/templates/includes/flash-messages.html`.


dstufft:
There are probably two different types of dismissible messages, ephemeral and persistent.

Something like the flash messages would be ephemeral, there's no need to persist them because they're gone forever after a page reload anyways, dismissing them only lets you get rid of them on the current load of the page without issuing a reload.

Then there is things like the donation banner which would be persistent, and would need a unique key assigned to them that would remember to keep them dismissed for all time. I *think* the ideal here is that wether something has been dismissed or not is stored in localstorage. Local Storage will be more persistent than cookies (which people are more likely to clear from time to time) and LocalStorage isn't sent back to the server on every single request, removing some needless bytes on the wire when we don't really care about what things the user has dismissed.

This does mean that persistent dismissals would only be persistent per browser per machine, which I think is perfectly fine for the initial feature work. If people want their dismissals to persist across browsers/machines (at least when they are signed into Warehouse) we can look at possibly expanding the feature in the future to additionally persist that data server side.

dstufft:
One thing that might be interesting to include is the idea of a version into the unique key. This could then allow the site developers to say "We think the information in this callout has changed significantly enough to re-trigger everyone seeing it, even if they've persisted it previously". Although arguably you could simply say that if it's changed that much, it should just get a new unique id.

di:
Agree w/ Donald, this should use LocalStorage and not cookies. Sorry for the misguidance in the original issue.

I was imagining that the "dismiss key" for each target would be it's ID. As this PR stands right now, it would dismiss all of the "dismissables" or none. I like Donald's idea of versioning, maybe the key could be a combination of the `id` attribute and a `data-version` attribute.

aalmazan:
Updated with feedback from @di and @dstufft:

- Dismissed notifications that should be persisted must have an `id` and `data-notification-version`
- If either `id` or `data-notification-version` doesn't exist, the dismissal will not persist
- Also, added notification dismissing in the flash-messages template

On my end, I see a quick little flash from a dismissed notification (since the element flips to `display: none` on load). For notifications such as the donate one, maybe we should default it to being hidden and switch to visible after the controller initializes to prevent this?

di:
> For notifications such as the donate one, maybe we should default it to being hidden and switch to visible after the controller initializes to prevent this?

The issue with this is that if a user has JS disabled, the notifications will always be hidden.

nlhkabu:
Perhaps we can add a CSS transition?  It could be a nice way of handling the dismiss gently **and** it might give the JS enough time to load to avoid that flicker?

dstufft:
> The issue with this is that if a user has JS disabled, the notifications will always be hidden.

If we made the dismissal a CSS class instead of setting ``display: none``  inline, we could use the ``<noscript>`` tag to load a set of CSS rules that only come into play when there isn't JS active. We could use this to both make it so that the dismissal class is a ``no-op`` AND hide the elements that make things look dimissable.

aalmazan:
I think adding `<noscript>` would probably do the trick. Testing this out, however, I'm getting the following when trying to do a quick test:

```html
<!-- In <head> -->
<noscript>
    <style>
        body {
            background-color: red !important;
        }
    </style>
</noscript>

// Error from console (Firefox)
Content Security Policy: The page’s settings blocked the loading of a resource at self (“style-src http://localhost http://fonts.googleapis.com”). Source: background-color: red;.

// Contents of CSP header
Content-Security-Policy: base-uri 'self'; block-all-mixed-content; connect-src 'self' https://2p66nmmycsj3.statuspage.io; default-src 'none'; font-src 'self' fonts.gstatic.com; form-action 'self'; frame-ancestors 'none'; frame-src 'none'; img-src 'self' "http://localhost:9000/" https://secure.gravatar.com; script-src 'self' www.google-analytics.com; style-src 'self' fonts.googleapis.com
```

`'self'` is already part of the `style-src` whitelist. Anyone have any ideas why this is still getting blocked?

di:
@aalmazan It's because the `'self'` CSP does not allow for inline styles, for that we would need `'unsafe-inline'` which, as it's name implies, could be unsafe.

So we'd need to make noscript load a special "noscript.css" style instead.

aalmazan:
Updated PR to include additional feedback:

* Added `noscript.{scss,css}` to gulpfile build process.
* Included `noscript.css` in base template within a `<noscript>` tag directly after all other css.
* Added `--dismissable` CSS modifier to indicate a notification's dismissability.
    * I'm reluctant to default all notifications to `display: none`, so this is the direction I ended up going. Even though it's sort of redundant to have this, I feel it's easier to reason about for future notifications. 
* Notifications are now always visible for users without javascript enabled.
* Solves the issue of the harsh flicker, at least from my testing. It now looks more "normal" when looking at the page load as a whole.

di:
Code-wise, this looks good to me (aside from a small linting error).

@nlhkabu will likely have some tweaks to make to the styling, for example I'm noticing that the "x" isn't vertically centered:

screen shot 2018-03-15 at 10 59 17 am

warehouse/templates/includes/flash-messages.html
```diff
@@ -13,19 +13,28 @@
...
 </div>
 {% endfor %}
 
 {% for message in request.session.pop_flash(queue="success") %}
-<div class="notification-bar notification-bar--success">
-  <span class="notification-bar__message">{{ message }}</span>
+<div class="notification-bar notification-bar--success" data-controller="notification" data-target="notification.notification">
+  <div class="site-container">
+    <span class="notification-bar__message">{{ message }}</span>
+    <span data-target="notification.dismissButton" data-action="click->notification#dismiss" class="notification-bar__dismiss">×</span>
```
di: Could you add a `title` attribute to these, something like "Dismiss this notification"?

warehouse/templates/includes/flash-messages.html
```diff
@@ -16,7 +16,7 @@
 <div class="notification-bar notification-bar--warning" data-controller="notification" data-target="notification.notification">
   <div class="site-container">
```
nlhkabu: 1. can we please remove the 'site container'?   I understand that you did this to position the close button near the middle of the page, but I fear that it will make some of our longer messages wrap over two lines.

warehouse/templates/includes/flash-messages.html
```diff
@@ -16,7 +16,7 @@
...
     <span class="notification-bar__message">{{ message }}</span>
-    <span data-target="notification.dismissButton" data-action="click->notification#dismiss" class="notification-bar__dismiss">×</span>
```
nlhkabu: Could we please make this `<span>` a `<button>` and use a font awesome icon? Also, we could probably add some aria in there:

```
<button data-target="notification.dismissButton" data-action="click->notification#dismiss" class="notification-bar__dismiss" aria-label="close">
  <i class="fa fa-times"></a>
</button>
```

warehouse/static/sass/blocks/_notification-bar.scss
```diff
@@ -39,6 +39,10 @@
   color: $white;
   padding: ($spacing-unit / 2);
 
+  .site-container {
+    position: relative;
```
nlhkabu: Given my earlier comment, this could be moved up to the notification-bar instead :)

warehouse/static/sass/blocks/_notification-bar.scss
```diff
@@ -54,6 +58,25 @@
...
+    cursor: pointer;
+    font-size: 1.5rem;
+  }
+
+  // Indicates that a notification is dismissable.
+  // Since this defaults to not displaying the element, one must make sure the
+  // notification controller is being used.
+  &--dismissable {
+    display: none;
+  }
```
nlhkabu: it might be nice to add a CSS transition here, so it fades in :)

warehouse/templates/includes/flash-messages.html
aalmazan: Sure thing. I'm going to assume having the close button at the top right is what we're expecting now.

warehouse/static/sass/blocks/_notification-bar.scss
aalmazan: I'm not too familiar with animations -- which property should we be animating? Based on [this Stack Overflow question](https://stackoverflow.com/a/3332179/6553469), we can't transition the `display` property. Should we have it transition `opacity` and/or `height` instead?

aalmazan:
@nlhkabu Made your requested changes excluding the transition -- please see my comment above. Screenshot below as well of the home page. If that all looks good, I'll remove the ability to dismiss the donate notification since that isn't meant to be closed. 

selection_016

warehouse/static/sass/blocks/_notification-bar.scss
nlhkabu: yes, you're right.  Let's leave it for now :)

nlhkabu:
Thanks for the updates @aalmazan !

One thing I noticed, if the text is long, it will overlap on the button:

screenshot from 2018-03-19 07-26-39

We can avoid this by adding more right padding to `.notification-bar--dismissable`, e.g.

screenshot from 2018-03-19 07-32-52

screenshot from 2018-03-19 07-32-28-2

brainwane:
@aalmazan I'm just dropping a note to thank you for your cooperation and help over these many revisions! Much appreciated!

aalmazan:
@nlhkabu Oops, was meaning to send this earlier but: updated css to fix overlapping text.

nlhkabu:
Awesome @aalmazan !  Could you please remove this from the donate notification?  Once this is merged, we'll work out which notifications we should apply it to :)

aalmazan:
Sorry for the delay @nlhkabu. Reverted the donate notification back.

nlhkabu:
Merged! 

Thanks for all your work on this @aalmazan !!!!

I'm going to leave https://github.com/pypa/warehouse/issues/3126 open for now, as we still want to be able to dismiss our `callout-block` component too - but this should be much easier to do with this PR as a guide :)

PR #3214, opened 2018-03-10
Make "recipients" a keyword only argument in send_email()

Mariatta:


warehouse/email.py
```diff
@@ -19,7 +19,7 @@
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, body, recipients, subject):
+def send_email(task, request, body, subject, recipients=None):
```
dstufft: This doesn't actually make it keyword only right? To do that it'd need to be ``*, recipients=None)`` right?

Mariatta: Thanks! You're right. I've added the `*, `

PR #3217, opened 2018-03-10
Add test pypi styles

### LINKED ISSUE #3104: Make Test PyPI visually more distinct from canonical PyPI
https://test.pypi.org/ looks so much like https://pypi.org/ visually -- same colors, same `favicon.ico`, same logo, same `<title>`, etc. -- that when the two are in tabs next to each other, it's harder to distinguish them than I'd like, which might lead to accidents.

Can we make them look more different visually while still preserving the functional parity between Test PyPI and canonical PyPI?

nlhkabu:
Closes #3104

screenshot from 2018-03-10 10-34-49

- Adds sticky warning
- Adds 'testing' bg image on header


di:
I really like the background image.

At first I didn't notice the red warning because it looks just like the pre-production warning that I've become used to. Maybe it's unnecessary? We're starting to have a lot of these piling up.

warehouse/static/sass/blocks/_site-header.scss
```diff
@@ -33,4 +33,9 @@
       max-width: 50px;
     }
   }
+
+  // Test PyPI style
+  &--testing {
+    background: $header-background-color url("../images/testing.png") 0 50%;
```
di: I think we want to be able to cache-bust this image instead. We'll have to specify this in-line in the template so we can use `request.static_path('warehouse:static/dist/images/testing.png')`

nlhkabu: ok, no problem :)

nlhkabu:
Thanks for your feedback @di - I originally only had the background image, but then I realised that anyone with a visual impairment would miss this information.  Also, the notification bar is sticky, where the header is not.

I think the banner blindness that you describe should disappear once we remove the pre-production warning - e.g it won't be normal to have a big red banner at the top all the time.

nlhkabu:
Ah - one thing I just realised might be a problem - if we decide to translate the site in the future, using a word as a background image might not be ideal.  I suppose we can cross that bridge when/if we come to it.

di:
I think we can localize the image just as easily if we do it in the template -- we'll just need to make a new image for every language.

dstufft:
I think images referenced in CSS is still cache busted?

di:
> I think images referenced in CSS is still cache busted?

Oh yup, they are. Sorry for the bad advice here @nlhkabu.

brainwane:
Could `<title>PyPI - the Python Package Index · Warehouse (TestPyPI)</title>` change to `<title>TestPyPI - the Python Package Index · Warehouse</title>` or similar, so that the truncated title in a browser tab starts with "test"?

And could the `favicon` be slightly different?

nlhkabu:
Yes, I can make these updates @brainwane.

@di - please advise on cache busting.  Should I remove the image altogether?  I think with the banner, combined with @brainwane's suggestions, we probably don't need it.

brainwane:
Here's a suggestion. Could Test PyPI, instead of saying

> Find, install and publish Python packages with the Python Package Index

say something like this?

> Test Python package publishing with the Test Python Package Index

di:
@nlhkabu What you originally had (background image in CSS) was fine. I like the background image and think we should keep it, but I'll leave it up to you.

nlhkabu:
Updated:

screenshot from 2018-03-28 18-11-15

screenshot from 2018-03-28 18-11-30

screenshot from 2018-03-28 18-11-23

nlhkabu:
@di or @ewdurbin - ready for re-review

warehouse/templates/base.html
```diff
@@ -102,6 +107,15 @@
         <span class="notification-bar__message">You are using an unsupported browser, please upgrade to a newer version.</span>
       </div>
       <![endif]-->
+      {% if testPyPI %}
+      <div class="notification-bar notification-bar--warning">
+        <span class="notification-bar__icon">
+          <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+          <span class="sr-only">Warning:</span>
+        </span>
+        <span class="notification-bar__message">You are using testPyPI - a separate instance of the Python Package Index that allows you to try distribution tools and processes without affecting the real index.</span>
```
pradyunsg: *TestPyPI

nlhkabu: good catch @pradyunsg

ewdurbin:
looks great! I really like the repeat in the header!

warehouse/templates/base.html
nlhkabu: fixed :)

PR #3220, opened 2018-03-10
Displays message when username/email not found in request password reset form

### LINKED ISSUE #3089: Tell user when email is not found when resetting password
Right now if a user tries to recover their account by providing an email address, we don't give any indication whether the email address is actually attached to an account.

Since attempting to register with a already "used" email address will provide the same information (that the email is in use by _some_ PyPI user) there's no real sense in protecting this bit of information here.

Instead we should tell the user that the email was not found, so that they can try a different email they might own, rather than mistakenly thinking that it was the correct email and that the email was not delivered (or that resetting a password is somehow broken).

waseem18:
Fixes #3089 

- Shows error message when username/email not found in request password reset form


waseem18:
Thank You @ewdurbin for helping me find the cause of code coverage failure. I'll fix that.

waseem18:
@di Can you please review this?

Thanks

lgh2:
Looks to me like it does what it says on the tin:

email-not-found2
email-not-found

PR #3239, opened 2018-03-13
RSS Feed Fixes

### LINKED ISSUE #3238: Legacy and Warehouse RSS feeds differ
@andrew asked in https://github.com/librariesio/libraries.io/issues/2024#issuecomment-372638824 about Warehouse's RSS feeds:

> * https://pypi.org/rss/updates.xml
> * https://pypi.org/rss/packages.xml

> Which I expected the contents to match the old ones but currently don't:

> * https://pypi.python.org/pypi?%3Aaction=rss
> * https://pypi.python.org/pypi?%3Aaction=packages_rss

I've verified through visual inspection that the data in the legacy RSS feed and the data in the Warehouse RSS feeds differ.

This is a bug in the feeds or a bug in the docs. Currently our [feeds documentation](https://warehouse.readthedocs.io/api-reference/feeds/) and [Warehouse migration guide](https://warehouse.readthedocs.io/api-reference/integration-guide/) don't say anything about deliberate differences between the legacy and Warehouse RSS feeds. We can update the docs if there's a deliberate reason for the difference.

di:
Fixes #3238.

* Updates deprecated DTD
* Use the release creation date for `pubDate` in `updates.xml`
* Set the `all-projects` cache key for RSS feeds
* Indentation/whitespace fixes.

### DIFF
--- a/warehouse/templates/rss/packages.xml
+++ b/warehouse/templates/rss/packages.xml
@@ -4,7 +4,7 @@
 {% block items -%}
   {% for project in newest_projects %}
     <item>
-      <title>{{ project.normalized_name }} added to PyPI</title>
+      <title>{{ project.name }} added to PyPI</title>
       <link>{{ request.route_url('packaging.project', name=project.normalized_name) }}</link>
       <guid>{{ request.route_url('packaging.project', name=project.normalized_name) }}</guid>
       <description>{{ project.releases[0].summary }}</description>
--- a/warehouse/templates/rss/packages.xml
+++ b/warehouse/templates/rss/packages.xml
@@ -1,14 +1,14 @@
 {% extends "base.xml" %}
 {% block title %}PyPI Newest Packages{% endblock %}
 {% block description %}Newest packages registered at the Python Package Index{% endblock %}
-{% block items %}
+{% block items -%}
   {% for project in newest_projects %}
-  <item>
-    <title>{{ project.normalized_name }} added to PyPI</title>
-    <link>{{ request.route_url('packaging.project', name=project.normalized_name) }}</link>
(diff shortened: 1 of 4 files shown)


warehouse/templates/rss/packages.xml
```diff
@@ -1,14 +1,14 @@
...
-    <title>{{ project.normalized_name }} added to PyPI</title>
-    <link>{{ request.route_url('packaging.project', name=project.normalized_name) }}</link>
-    <guid>{{ request.route_url('packaging.project', name=project.normalized_name) }}</guid>
-    <description>{{ project.releases[0].summary }}</description>
-    <pubDate>{{ project.created|format_rfc822_datetime() }}</pubDate>
-  </item>
-  {% endfor %}
-{% endblock %}
+    <item>
+      <title>{{ project.normalized_name }} added to PyPI</title>
```
dstufft: While we're messing around here, do we actually want to use the normalized name here? Seems like it'd be better using the actual name.

warehouse/templates/rss/updates.xml
```diff
@@ -1,13 +1,13 @@
...
-  <item>
-    <title>{{ release.project.normalized_name }} {{ release.version }}</title>
-    <link>{{ request.route_url('packaging.release', name=release.project.normalized_name, version=release.version) }}</link>
-    <description>{{ release.summary }}</description>
-    <pubDate>{{ release.project.created|format_rfc822_datetime() }}</pubDate>
-  </item>
-  {% endfor %}
-{% endblock %}
+    <item>
+      <title>{{ release.project.normalized_name }} {{ release.version }}</title>
```
dstufft: Same thing with normalized name here.

PR #3240, opened 2018-03-13
Add API changes to migration doc

brainwane:
Add API changes from legacy and more guidance to the "migrating to new PyPI" doc. Addresses the last bit of #3238.


brainwane:
Added @ncoghlan's observation in https://github.com/pypa/warehouse/issues/2935#issuecomment-369459540 about `Accept-Encoding: gzip`.

PR #3242, opened 2018-03-13
Add user assumptions to application overview

brainwane:
Help with #2794, and improve the directory listing a bit.

### DIFF
--- a/docs/application.rst
+++ b/docs/application.rst
@@ -80,28 +80,28 @@ The top-level directory of the Warehouse repo contains files including:
 
 Directories within the repository:
 
-::
+- bin/ - high-level scripts for Docker, Travis, and Makefile commands
+- dev/ - assets for developer environment
+- tests/ - tests
+- warehouse/ - code in modules
 
-    bin/ - high-level scripts for Docker, Travis, and Makefile commands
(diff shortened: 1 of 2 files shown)


docs/application.rst
```diff
@@ -28,6 +28,33 @@ pattern. This method of handling URLs may be unfamiliar to developers used to
...
+2. *Owners/maintainers of one or more projects.* This accounts for
+   almost all writes. A user must create and use a PyPI account to
+   maintain or own a project, and there is no particular functionality
+   available to a logged-in user other than to manage projects they
+   own/maintain. As of March 2018, PyPI had about 270,000 users, and
+   Test PyPI had about 30,000 users.
+3. *PyPI application administrators*, e.g., Ernest W. Durbin III,
+   Dustin Ingram, and Donald Stufft, who add classifiers, ban
+   spam/malware projects, help users with account recovery, and so
+   on. There are under ten such admins.
```
di: There are exactly three 😉

brainwane:
I'll keep the number of admins nonspecific; we may change the number in the next few months. Key point: there are so few admins, and they are so trusted, that we don't need to subdivide them into any further categories.

Thanks!

Issue #3223, opened 2018-03-11
Add `required` attribute to input element in Request Password Reset page


waseem18:
In [Request Password Reset](https://pypi.org/account/request-password-reset) page the input element doesn't have `required` attribute. Hitting `Request Password Reset` button with empty input sends a POST call and then adds `This field is required` message.
It's better to have a `required` attribute which marks the input as required.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

**Screenshot Required**: *If your pull request makes a visual change*, please include a screenshot of your update. This helps our team give you feedback faster.

nlhkabu:
Thanks @waseem18 :)

Issue #3197, opened 2018-03-09
Add JS validation when passwords do not match


nlhkabu:
It would be nice to use JavaScript to indicate whether or not the passwords match:

1. show a message
2. conditionally enable the button

----

![screenshot from 2018-03-09 06-36-40](https://user-images.githubusercontent.com/3323703/37194259-48d7d8da-2365-11e8-83b1-82bac0c1e220.png)

----

![screenshot from 2018-03-09 06-36-40-match](https://user-images.githubusercontent.com/3323703/37194328-94f916c0-2365-11e8-8cb5-bee346c806da.png)

@yeraydiazdiaz would this be something you might enjoy working on?

yeraydiazdiaz:
Sure, I'll take the opportunity to rewrite the feature in Stimulus.

Disabling the submit button was something [we explored before](https://github.com/pypa/warehouse/pull/2542#issuecomment-341894703) and decided against it. We are now using a pattern where feedback is given when clicking on the submit button in the form of tooltips.

Do we want to mix both feedback flows?

di:
@yeraydiazdiaz Great question. The submit button disabling was a problem on the registration form because there wasn't a way to provide feedback if the user was missing one of the required fields.

I think here it's OK, since it should be pretty clear to the user that they need to fill out the password field in order to be able to change their password, and we'll be able to conditionally show the "passwords match"/"don't match" message when the first field is filled out.

yeraydiazdiaz:
@nlhkabu which CSS classes are you using for the messages? The closest thing I'm finding are badges and don't quite look the same

nlhkabu:
@yeraydiazdiaz you could use `warehouse/static/sass/blocks/_form-errors.scss` to display the error.
For the green, you could create a `.form-error--valid` modifier on this component. Let me know if you need a hand with that - if you prefer you could push the PR unstyled and I can add a commit for this :)

PR #3243, opened 2018-03-13
Password strength gauge stimulus

yeraydiazdiaz:
This builds on #3219 addressing #3197 and #3223 

More importantly includes a refactor of the password gauge functionality to Stimulus.

There's a few decisions I made around the ability to compose components using JS "mixins" which I'm not sure it's "idiomatic Stimulus" if there is such a thing. Here's three captures of the end results in three areas:

- Reset password:
reset_password

- Register:
register

- Update password:
update_password

Hopefully it makes sense and we can move forwards with Stimulus using this pattern.


warehouse/static/js/warehouse/controllers/password_controller.js
```diff
@@ -11,10 +11,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+/* global zxcvbn */
+
```
di: Accidental comment?

di:
Also, thank you for tackling this!

warehouse/static/js/warehouse/controllers/password_controller.js
yeraydiazdiaz: `eslint` appeaser 😄

yeraydiazdiaz:
I've updated this to a much simpler approach of adding multiple controllers to the form.

I feel like [some lines](https://github.com/pypa/warehouse/pull/3243/files#diff-7ae9dcecd75b93cd37df3ca32da2c081R224) are getting a big too wide to read comfortably. I'd be in favor of something like:

```
{{ form.password_confirm(
    placeholder="Confirm password",
    required="required",
    class_="form-group__input",
    data_target="password.password password-match.passwordMatch",
    data_action="keyup->password-match#checkPasswordsMatch") }}
```

Would we consider disabling the template linting rule around single line echo statements to allow that?

di:
@yeraydiazdiaz I think that'd be fine.

yeraydiazdiaz:
Seems it's not something we can disable in the `html_lint` command line 😞

PR #3251, opened 2018-03-14
Move modal out of dropdown markup

### LINKED ISSUE #3152: Modal disappears when mouse leaves window
When a modal is up, like the "Delete Release" modal, the modal disappears when the user moves the mouse outside the browser window. It only reappears when the user hovers over the button:

mar-06-2018 19-04-26

yeraydiazdiaz:
Fixes #3152 

Turns out no Javascript was required, just refactoring the templates so the modal is independent of the dropdown. Here's the current behaviour for:

- Delete file
modal_disappears_fix

- Delete release:

modal_disappears_fix_2

I believe those are the only two instances affected by this bug?


brainwane:
Thanks for the PR, @yeraydiazdiaz! And allow me to say that I appreciate your consistent and thoughtful use of screenshots and animated screencast GIFs in your issues and PRs.

Heads-up @nlhkabu.

nlhkabu:
hi @yeraydiazdiaz thanks for this PR.  We have a similar problems with styles from the callout block bleeding into the modal.  

I fixed this by adding this CSS: https://github.com/pypa/warehouse/pull/3190

But maybe we could do the same thing with the HTML, which would allow us to simplify this CSS?

PR #3261, opened 2018-03-14
notification-bar: rename 'warning' style to 'danger'

alexwlchan:
Since the background-color is $danger-color, we should rename this class to make it consistent.

I gripped the codebase for "notification-bar", and fixed all the references I found.

Resolves #3164.

PR #3262, opened 2018-03-14
Document all the modifiers in the sass directory

alexwlchan:
While working on #3261, I noticed that not all of the Modifiers were documented, so I added appropriate comments for them.


nlhkabu:
Thanks for picking up on this and fixing it @alexwlchan !

brainwane:
Thanks @alexwlchan! And let me know if you'd like another issue suggested for you to look at.

PR #3288, opened 2018-03-16
set max_connections on redis clients

ewdurbin:


warehouse/config.py
```diff
@@ -159,6 +159,7 @@ def configure(settings=None):
     maybe_set(settings, "ratelimit.url", "REDIS_URL")
     maybe_set(settings, "recaptcha.site_key", "RECAPTCHA_SITE_KEY")
     maybe_set(settings, "recaptcha.secret_key", "RECAPTCHA_SECRET_KEY")
+    maybe_set(settings, "redis.max_connections", "REDIS_MAX_CONNECTIONS", int)
```
ewdurbin: @dstufft should we set a default here?

dstufft: If we want a default different than what Redis itself supplies, yes.

warehouse/rate_limiting/__init__.py
```diff
@@ -109,4 +109,5 @@ def __call__(self, context, request):
 def includeme(config):
     config.registry["ratelimiter.storage"] = storage_from_string(
         config.registry.settings["ratelimit.url"],
+        max_connections=config.registry.settings.get("redis.max_connections"),
```
dstufft: Because we're effectively making multiple pools, our max connections is actually going to be *double* what is specific as ``redis.max_connections``. We have a few options:

1. Live with it, and know that if we add another pool somewhere it will cause us to get another set of max connections.
2. Divide the number of max connections in half and give half to each pool.
3. Create a single connection pool and share it amongst all of the redis using things.

Another question if whether or not we want Redis to throw an error if it reaches this maximum number of connections, or if we want it to block for N seconds to wait for a connection.

dstufft:
Closing this, it's no longer important.

PR #3291, opened 2018-03-16
Move modal out of dropdown markup, ensure index matches slug

### LINKED ISSUE #3152: Modal disappears when mouse leaves window

yeraydiazdiaz:
Fixes #3152 again 😅 

The issue was the `id` attribute in the slug deriving from an empty `index` variable which made all modals have the same `id`.

- Releases

modal_disappears_fix_releases

- Release files

modal_disappears_fix_release_files


lgh2:
I tested this on Linux Mint 18.1 (Ubuntu 16.04) with Chrome (Version 65.0.3325.124 (Official Build) beta (64-bit)) and it worked, the modal remained on the screen when I moused away from it. 
It worked in Firefox 58.0.2 (64-bit) on the same OS.

PR #3292, opened 2018-03-16
Sort simple on Release.filename

### LINKED ISSUE #3284: Warehouse file order differs from legacy PyPI file list
Tonight, while load testing of pypi.org was ongoing, we saw some failures in automated systems that use `--require-hashes` with `pip install`, as ordering on the package file list page changed.

The specific package we saw break was `pandas` at version `0.12.0`. We had a single hash for `pandas-0.12.0.tar.gz`. A few of our hosts were served from the legacy PyPI service, which succeeded as normal. The Warehouse endpoint, however, failed, since `pandas-0.12.0.zip` now preceded `pandas-0.12.0.tar.gz` in the file list.

At the moment, you can see that https://pypi.org/simple/pandas/ and https://pypi.python.org/simple/pandas/ differ by searching for `pandas-0.12.0.tar.gz` and `pandas-0.12.0.zip` and comparing the position.

di:
Fixes #3284.

The order is still (intentionally) different from pypi-legacy, but `.tar` sdists will come before `.zip`s as before, and visually it makes more sense as links within the same version are just alphabetically sorted.

### DIFF
--- a/tests/unit/legacy/api/test_simple.py
+++ b/tests/unit/legacy/api/test_simple.py
@@ -202,7 +202,7 @@ class TestSimpleDetail:
 
         files = []
         for files_release in \
-                zip(egg_files, wheel_files, tar_files):
+                zip(egg_files, tar_files, wheel_files):
             files += files_release
 
         db_request.matchdict["name"] = project.normalized_name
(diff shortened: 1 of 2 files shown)


tests/unit/legacy/api/test_simple.py
```diff
@@ -212,9 +212,6 @@ def test_with_files_with_version_multi_digit(self, db_request):
         # Make sure that we get any changes made since the JournalEntry was
         # saved.
         db_request.db.refresh(project)
-        import pprint
```
ewdurbin: thanks for clearing this up as well.

PR #3302, opened 2018-03-16
allow legacy domain to call XMLRPC for load testing

ewdurbin:


ewdurbin:
not sure if this would have worked, and can do it in VCL since it's temporary anyway.

PR #3313, opened 2018-03-19
Improve owner/maintainer documentation

jMuzsik:
Hi, simple fix, only had to move some lines around/add text. Thanks!

### DIFF
--- a/warehouse/templates/manage/roles.html
+++ b/warehouse/templates/manage/roles.html
@@ -26,7 +26,7 @@
       <dt>Maintainer</dt>
       <dd>Can upload releases for a package. Cannot add collaborators. Cannot delete files, releases, or the project.</dd>
       <dt>Owner</dt>
-      <dd>Owns a package. Can upload releases. Can add other collaborators. Can delete files, releases, or the entire project.</dd>
+      <dd>Can upload releases. Can add other collaborators. Can delete files, releases, or the entire project.</dd>
     </dl>
   </div>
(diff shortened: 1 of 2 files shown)


warehouse/templates/manage/roles.html
```diff
@@ -23,10 +23,10 @@ <h2>Collaborators</h2>
   <div class="callout-block">
     <p>There are two possible roles for collaborators:</p>
     <dl>
+      <dt>Maintainer</dt>
+      <dd>May upload releases for a project. Cannot add collaborators, delete files, delete releases or delete the project.</dd>
       <dt>Owner</dt>
       <dd>Owns a package, may add other collaborators for that package, and upload releases for a package.</dd>
```
di: While you're here, could you update this line to indicate that owners _can_ delete releases and _can_ delete the project?

jMuzsik:
Ok, I edited the text a little. Replaced all may's with can's. Also, while reading closely I thought it to be preferable to reorder the order of ideas. Do you think it looks good?

One thing I am wondering is if I ought to delete the reference to `Owns a package` b/c maintainer has no reference to what he does. So perhaps I ought to write `Helps maintain a package.` or delete `Owns a package`?

di:
@jMuzsik I think deleting the "Owns a package" is fine, it's a bit redundant anyways 🙂

di:
Thanks @jMuzsik!

PR #3326, opened 2018-03-21
Internal Caching for XMLRPC methods

ewdurbin:
Caches specified XMLRPC methods via Redis.

Sets appropriate tags/keynames so that these caches can be invalidated via a similar mechanism to how we're performing CDN caches.


warehouse/legacy/api/xmlrpc.py
```diff
@@ -27,14 +27,24 @@
...
-
 pypi_xmlrpc = functools.partial(
     xmlrpc_method,
     endpoint="pypi",
     require_csrf=False,
     require_methods=["POST"],
 )
 
+pypi_xmlrpc_cache_by_project = functools.partial(
+    xmlrpc_method,
```
dstufft: You can partial a partial right? Feels like it'd be slightly nicer not to duplicate this stuff by swapping this from ``xmlrpc_method`` to ``pypi_xmlrpc`` and ditching the shared kwargs.

warehouse/xmlrpc_cache/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+        'xmlrpc_cache.url'
+    )
+    xmlrpc_cache_name = config.registry.settings.get(
+        'xmlrpc_cache.name', 'lru'
+    )
+    xmlrpc_cache_expires = config.registry.settings.get(
+        'xmlrpc_cache.expires', 3600
+    )
+
+    xmlrpc_cache_url_scheme = urlparse(xmlrpc_cache_url).scheme
```
dstufft: This is going to error if an xmlrpc cache isn't specified. Is that desired behavior? Or should we fall back to null in that case?

ewdurbin: adding a test to be sure, but my understanding was that `config.registry.settings.get` call would return `None` in that case.

```
>>> urlparse(None)
ParseResultBytes(scheme=b'', netloc=b'', path=b'', params=b'', query=b'', fragment=b'')
```

ewdurbin:
@dstufft @di ready for some more 👁 I'm gonna add/test purge functionality shortly!

ewdurbin:
would love some input on how we feel it would be best to invoke this. a separate task or all in the existing task:

https://github.com/pypa/warehouse/blob/088f648c62c6292369344e03eb75abbcf98b465f/warehouse/cache/origin/fastly.py#L27-L34

```diff
 def purge_key(task, request, key):
     cacher = request.find_service(IOriginCache)
     try:
+        xmlrpc_cache_service = request.find_service(IXMLRPCCache)
+    except ValueError:
+        xmlrpc_cache_service = None
+    try:
+        if xmlrpc_cache_service:
+            xmlrpc_cache_service.purge(key)
         cacher.purge_key(key)
     except (requests.ConnectionError, requests.HTTPError, requests.Timeout,
-            UnsuccessfulPurge) as exc:
+            UnsuccessfulPurge, CacheError) as exc:
         raise task.retry(exc=exc)
```

isn't the most elegant but it'd work.

ewdurbin:
@di @dstufft this is complete! let's cut it up!

warehouse/legacy/api/xmlrpc/cache/interfaces.py
```diff
@@ -0,0 +1,47 @@
...
+        expiration.
+        """
+
+    def purge(self, tag):
+        """
+        Issues a purge, clearing all cached objects associated with the tag
+        from the cache.
+        """
+
+    def purge_tags(self, tag):
```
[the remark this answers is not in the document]
ewdurbin: damn skippy it should be.

warehouse/legacy/api/xmlrpc/cache/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+@db.listens_for(db.Session, "after_commit")
+def execute_purge(config, session):
+    purges = session.info.pop(
+        "warehouse.legacy.api.xmlrpc.cache.purges", set()
+    )
+
+    try:
+        xmlrpc_cache_factory = config.find_service_factory(IXMLRPCCache)
+    except ValueError:
+        return
```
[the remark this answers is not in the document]
ewdurbin: if no IXMLRPCCache is available. this supports running w/o one. it's not the prettiest way of handling it but it's what we have.

PR #3327, opened 2018-03-21
Add default text to browse page

### LINKED ISSUE #3062: Display help text on browse page and null search page
Hi,

I think that it would make much better impression if the packages were sorted by e.g. download count/popularity/update rather alphabetically by default.

When you click on [Browse project](https://pypi.org/search/) and it takes you to the search page, it looks like this:
image

That's not really a very flattering choice of packages for Python, and more importantly, it's very probably not what a user wants to see. Clearly, what the user wants to see most likely is something more correlated with usage/download/popularity/visited rather than alphabet.

nlhkabu:
Closes #3062 

screenshot from 2018-03-21 07-04-34

Includes:

- Small restyle of the filter buttons (to give them more room)
- Adjustment of results message to reflect applied filters, examples:

### Search only
screenshot from 2018-03-21 07-07-36

### Classifiers only
screenshot from 2018-03-21 07-07-00
screenshot from 2018-03-21 07-06-53

### Search and classifiers
screenshot from 2018-03-21 07-08-07

### DIFF
--- a/warehouse/templates/search/results.html
+++ b/warehouse/templates/search/results.html
@@ -133,9 +133,21 @@
         <p>Enter a search query above, or select a filter from the list of classifiers on the left.</p>
         <p>You can combine searches and classifier filters. Examples:</p>
         <ul class="no-bottom-margin">
-          <li><a href="/search/?q=&o=&c=Programming+Language+%3A%3A+Python+%3A%3A+3">New Python 3 project releases</a></li>
-          <li><a href="/search/?q=&o=&c=Framework+%3A%3A+Sphinx+%3A%3A+Extension&c=Development+Status+%3A%3A+5+-+Production%2FStable">Sphinx extensions that are stable/production status</a></li>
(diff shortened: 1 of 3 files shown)


warehouse/templates/search/results.html
```diff
@@ -128,81 +128,94 @@ <h2 class="no-top-padding">Filter Projects</h2>
     </div>
 
     <div class="left-layout__main">
+      {% if not term and not applied_filters %}
+      <div class="callout-block no-top-margin">
+        <p>Enter a search query above, or select a filter from the list of classifiers on the left.</p>
+        <p>You can combine searches and classifier filters. Examples:</p>
+        <ul class="no-bottom-margin">
+          <li><a href="/search/?q=&o=&c=Programming+Language+%3A%3A+Python+%3A%3A+3">New Python 3 project releases</a></li>
```
nlhkabu: I wasn't sure the best way to make these examples - so I've just hardcoded them for now :/
could someone please advise?

di: Figured it was easier to "show" than "tell" so I fixed this in 16423df, take a look. 🙂

PR #3333, opened 2018-03-21
Document API policies

### LINKED ISSUE #2931: Document API policies
Add short explanations to `docs/api-reference/` for:

* request and rate limits
* API etiquette and ways to be a good neighbor (e.g., consider running your own organizational mirror instead, use a descriptive User-Agent header with your contact info in it, normalize your queries (how?) to optimize cache hits)
* what sort of caching to expect

di:
Fixes #2931.

### DIFF
--- a/docs/api-reference/index.rst
+++ b/docs/api-reference/index.rst
@@ -4,6 +4,50 @@ API Reference
 Warehouse has several API endpoints. See :doc:`../application` for the
 parts of Warehouse that generate them.
 
+.. contents:: :local:
+
+API Policies
+------------
+
+Please be aware of the following PyPI API Policies:
+
+Caching
+~~~~~~~
+
+All API requests are cached. Requests to the JSON, RSS or Legacy APIs are
+cached by our CDN provider. You can determine if you've hit the cache based on
+the ``X-Cache`` and ``X-Cache-Hist`` headers in the response.
+
+All API requests also provide an ``ETag`` header. If you're making a lot of
+repeated requests, please ensure your API consumer will respect this header to
+determine whether to actually repeat a request or not.
+
+Rate Limiting
+~~~~~~~~~~~~~
+
+Due to the heavy caching and CDN use, there is currently no rate limiting of
+PyPI APIs.
+
+However, if you plan to make a lot of requests to a PyPI API, please adhere to
+the following suggestions:
+
+* Set your consumer's ``User-Agent`` header to uniquely identify your requests
+* Try not to make a lot of requests (thousands) in a short amount of time
+  (minutes). Generally PyPI can handle it, but it's preferred to make requests
+  in serial over a longer amount of time if possible.
+
+API Preference
+~~~~~~~~~~~~~~
+
+For periodically checking for new packages or updates to existing packages,
+please use our RSS feeds.
+
+If at all possible, it is recommended to use the JSON/RSS/Legacy APIs over
+XML-RPC.
+
+Available APIs
+--------------
+
 .. toctree::
     :maxdepth: 2
 
--- a/docs/api-reference/index.rst
+++ b/docs/api-reference/index.rst
@@ -42,6 +42,10 @@ the following suggestions:
 * Try not to make a lot of requests (thousands) in a short amount of time
   (minutes). Generally PyPI can handle it, but it's preferred to make requests
   in serial over a longer amount of time if possible.
+* If your consumer is actually an organization or service that will be
+  downloading a lot of packages from PyPI, consider `using your own index
+  mirror or cache
+  <https://packaging.python.org/guides/index-mirrors-and-caches/>`_.
 
 API Preference
 ~~~~~~~~~~~~~~
--- a/docs/api-reference/index.rst
+++ b/docs/api-reference/index.rst
@@ -16,19 +16,25 @@ Caching
 
 All API requests are cached. Requests to the JSON, RSS or Legacy APIs are
 cached by our CDN provider. You can determine if you've hit the cache based on
-the ``X-Cache`` and ``X-Cache-Hist`` headers in the response.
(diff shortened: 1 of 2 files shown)


docs/api-reference/index.rst
```diff
@@ -4,6 +4,50 @@ API Reference
...
+
+All API requests also provide an ``ETag`` header. If you're making a lot of
+repeated requests, please ensure your API consumer will respect this header to
+determine whether to actually repeat a request or not.
+
+Rate Limiting
+~~~~~~~~~~~~~
+
+Due to the heavy caching and CDN use, there is currently no rate limiting of
+PyPI APIs.
```
ewdurbin: *At the edge. Requests that hit our backends (uploads, XMLRPC) may be rate limited.

docs/api-reference/index.rst
```diff
@@ -4,6 +4,50 @@ API Reference
...
+  (minutes). Generally PyPI can handle it, but it's preferred to make requests
+  in serial over a longer amount of time if possible.
+
+API Preference
+~~~~~~~~~~~~~~
+
+For periodically checking for new packages or updates to existing packages,
+please use our RSS feeds.
+
+If at all possible, it is recommended to use the JSON/RSS/Legacy APIs over
```
ewdurbin: Stronger language, may be worth stating that XMLRPC will be going away as soon as we find a way to.

>>> No new integrations should use the XMLRPC interfaces that PyPI provides, it is planned for deprecation.

docs/api-reference/index.rst
ewdurbin: *At the edge. Requests that hit our backends (uploads, XMLRPC) may be rate limited (automatically or manually) if they are causing degradation of service.

docs/api-reference/index.rst
```diff
@@ -4,6 +4,50 @@ API Reference
...
+~~~~~~~
+
+All API requests are cached. Requests to the JSON, RSS or Legacy APIs are
+cached by our CDN provider. You can determine if you've hit the cache based on
+the ``X-Cache`` and ``X-Cache-Hist`` headers in the response.
+
+All API requests also provide an ``ETag`` header. If you're making a lot of
+repeated requests, please ensure your API consumer will respect this header to
+determine whether to actually repeat a request or not.
+
```
ewdurbin: Need exception for XMLRPC here. At this point it provides no information about if the response was cached internally.

docs/api-reference/index.rst
```diff
@@ -4,6 +4,50 @@ API Reference
...
+------------
+
+Please be aware of the following PyPI API Policies:
+
+Caching
+~~~~~~~
+
+All API requests are cached. Requests to the JSON, RSS or Legacy APIs are
+cached by our CDN provider. You can determine if you've hit the cache based on
+the ``X-Cache`` and ``X-Cache-Hist`` headers in the response.
```
ewdurbin: *Hits

docs/api-reference/index.rst
ewdurbin: Additionally we should reserve the right to temporarily prohibit a client based on activity.

di:
@ewdurbin Might it be preferable to be _less_ explicit about how we can prohibit malicious clients? I think for the average user, it's enough to just know "we can drop the 🔨 on you".

docs/api-reference/index.rst
```diff
@@ -4,6 +4,56 @@ API Reference
...
+In addition, PyPI reserves the right to temporarily or permanently prohibit a
+consumer based on irresponsible activity.
+
+If you plan to make a lot of requests to a PyPI API, please adhere to
+the following suggestions:
+
+* Set your consumer's ``User-Agent`` header to uniquely identify your requests
+* Try not to make a lot of requests (thousands) in a short amount of time
+  (minutes). Generally PyPI can handle it, but it's preferred to make requests
+  in serial over a longer amount of time if possible.
```
brainwane: I suggest we add "consider running your own organizational mirror instead/in addition" & point to bandersnatch.

docs/api-reference/index.rst
```diff
@@ -4,6 +4,56 @@ API Reference
...
+
+Rate Limiting
+~~~~~~~~~~~~~
+
+Due to the heavy caching and CDN use, there is currently no rate limiting of
+PyPI APIs at the edge. The XML-RPC API may be rate limited if usage is causing
+degradation of service.
+
+In addition, PyPI reserves the right to temporarily or permanently prohibit a
+consumer based on irresponsible activity.
```
brainwane: At a past org we suggested: use a descriptive User-Agent header with your contact info in it, so we can at least potentially ping you to suggest alternate approaches if you're causing a ruckus.

docs/api-reference/index.rst
di: I'm not sure that's good advice for folks that need consume our APIs. Does bandersnatch provide all the exact same APIs as PyPI?

Generally I think "use bandersnatch" is good advice for users that either can't use `pypi.org` or are concerned about it's availability. I'm not sure that using it could be considered a "best practice" for consuming our API.

(With the exception of using a mirror for the Simple/download API, which could be helpful in reducing load on our CDN, but I think we don't really have the assumption that this is necessary or recommended for big consumers -- we just handle the traffic.)

docs/api-reference/index.rst
di: Great point.

docs/api-reference/index.rst
brainwane: This document tells API users how to be good neighbors, so I suggest we add a line advising orgs that it would be friendly for them to use internal mirrors for Simple/downloads. But my opinion is mild here.

di: I found a happy medium 🙂

PR #3339, opened 2018-03-21
Registration Honeypot

### LINKED ISSUE #3174: reCAPTCHA blocked in China
A pypi ID is necessary while Iwould like to share and upload a package on pypi. Unfortunnatly, there was something wrong with the website of "https://pypi.org/account/register/".  After writing the all information,  an ERROR was arise named "Recaptcha error.".
I'm not sure if that is caused by China Internet?

di:
Fixes #3174.


warehouse/templates/accounts/register.html
```diff
@@ -58,6 +56,14 @@ <h1 class="page-title">Create an account on PyPI</h1>
           {% endif %}
         </div>
 
+        {# The following is a honeypot field and is hidden from the user #}
+        <div class="form-group confirm-email" aria-hidden="true">
```
ewdurbin: not sure if it'd conflict but is `aria-placeholder` valid in addition? I'm guessing there's some ability in screen readers to unhide elements, and in case someone does that perhaps we should let them know what's up?

ewdurbin: https://www.w3.org/TR/wai-aria-1.1/#aria-placeholder

di: My understanding is that screen readers just ignore anything `aria-hidden`. I didn't want to put too much in the source here indicating what this is, adding a placeholder might do that.

dstufft:
Rather than one-off add some HTML, can we make this a proper, reusable form field?

warehouse/templates/accounts/register.html
ewdurbin: @tvance @Kataract any tips?

di:
@dstufft We could, kinda wanted it to blend into the source for this form a little bit so it wouldn't have an obvious name like "honeypot", do you think that's not a big deal?

dstufft:
Could always give it a non-obvious name, but still be reusable yea?

warehouse/templates/accounts/register.html
Kataract: `aria-hidden` isn't going to do anything for content that isn't generally interactable (like `<div>`s).  You can do this two ways:

* `display: none;` or `visibility: hidden` the `<div>`, and have the `<label>` text indicate that this field is a honeypot and not to fill it out
* apply `aria-hidden="true"` to the input and visually hide the `<label>`

see: https://www.w3.org/WAI/GL/wiki/Captcha_Alternatives_and_thoughts 

there's a specific post I have in mind that I used for an internal discussion on reCAPTCHA that I can't find quickly in google - if you want I can link it here later, if that isn't enough info.

di: Ok great! This is already hiding the `div` with `display: none` so it seems like I should move the `aria-hidden` directly to the input. Thanks very much!

Issue #3336, opened 2018-03-21
Unnecessary purges on `User` model change


di:
Currently right now any time the `User` model changes, we purge all the cache keys for that user's project.

This includes attribute changes that don't actually affect the project pages, like `last_login`, `password` etc.

We should filter out "purge-able" attribute changes and only issue purges when necessary. Said attributes include:
* `username`
* `name`
* `emails`

ewdurbin:
it also issues an `all-projects` which purges `/simple`

di:
Hmm, shouldn't be...

ewdurbin:
my bad, sneaky edited my original comment after checking logs more closely.

brainwane:
Ernest mentioned we need this before launching beta, because otherwise it's so costly to serve users who have many projects each.

PR #3346, opened 2018-03-22
WIP Beta FAQ update

### LINKED ISSUE #2518: Remove Pre-Production Warning
Currently there is a pre-production warning on the header of PyPI, we need to remove that (once the infrastructure is done) before we start redirecting pypi.python.org to pypi.org.

brainwane:
Marking as `blocked` since I know this depends on updating the banner in `pypi-theme`. I suggest `#beta` and `beta()` for the `help/` section label, which would require the banner changing its link.

@ewdurbin @dstufft @nlhkabu @di I am a thousand percent open to critique on this FAQ change; what promises should we be making about infra & features? What should we encourage people to do or avoid?

Closes #2518.


docs/api-reference/integration-guide.rst
```diff
@@ -45,7 +45,8 @@ Here are some tips.
 * JSON API: ``https://pypi.org/pypi/{name}/json`` returns the
   expected JSON response directly. See :doc:`json`.
 
-* XML-RPC API: see :ref:`changes-to-legacy-api`.
+* XML-RPC API: see :ref:`changes-to-legacy-api`. Will be deprecated in
+  the future; switch to the RSS or JSON APIs.
```
dstufft: Should we call to action to file issues for things that they use that only exists in the XMLRPC API?

warehouse/templates/pages/help.html
```diff
@@ -281,23 +281,24 @@ <h3 id="availability">{{ availability() }}</h3>
...
         <ul>
           <li>The UI may return unusual or erroneous results</li>
           <li>Performance may not yet be optimised</li>
-          <li>Some features that you use on the old interface may be missing</li>
+          <li>Some features that you used in <a href="https://pypi.python.org">the old interface (pypi.python.org)</a> may be missing</li>
+	  <li>We <a href="#feedback">need your feedback</a> to get the site production ready</li>
         </ul>
         <p>
-          While we are in pre-production, our infrastructure cannot yet support all of the traffic generated by our users running <code>pip install</code>. We therefore ask that you do not use this version of PyPI for production codebases.
+          While we are in beta, the pypi.org infrastructure cannot yet support all of the API traffic generated by our users running <code>pip install</code>, so don't explicitly point to it in your production setups. But we're occasionally redirecting portions of that traffic in load tests we announce <a href="http://status.python.org/">on our status page</a>.
```
dstufft: I think we can say we're generally ready for traffic? Or will be once @ewdurbin gets the XMLRPC caching sorted out? Are we ready to tell people they can start shifting their own traffic to Warehouse now? I'm not sure how to word it really, but like "Early Adopter" traffic is probably not going to knock us over?

brainwane: I'd like to encourage early adopter traffic, and can find a way to word it if @ewdurbin says it's ok.

docs/api-reference/integration-guide.rst
brainwane: Good call! Fixed.

warehouse/templates/pages/help.html
ewdurbin: Once we get #3336 cleared up, we'll be all set to start calling it beta. Right now it's just too costly if a user with lots of projects logs in.

brainwane: @ewdurbin Good info, thanks. So, once we're in beta, what level of traffic are you ok with encouraging? (this is kind of "what does 'beta' mean to you?")

ewdurbin: Anything that arrives organically is probably fine. We may be migrating our backend database as well now... so perhaps we should wait until after that?

brainwane:
Ready for re-review with new traffic wording.

PR #3352, opened 2018-03-22
Attribute Event purging for user changes

### LINKED ISSUE #3336: Unnecessary purges on `User` model change

di:
Fixes #3336.


dstufft:
I just woke up, but this seems to simply be adding more cases where things will get purged, but to actually fix #3336 we need to remove cases?

di:
@dstufft The diff is tricky but this removes all the `purge_keys` for the `User` model:

```diff
config.register_origin_cache_keys(
    User,
    cache_keys=["user/{obj.username}"],
-     purge_keys=[
-         key_factory("user/{obj.username}"),
-         key_factory(
-             "project/{itr.normalized_name}",
-             iterate_on='projects',
-         ),
-     ],
- ),
```

And adds them back only on specific attributes:

```diff
+    config.register_origin_cache_keys(
+        User.name,
+        purge_keys=[
+            key_factory("user/{obj.user.username}"),
+            key_factory(
+                "project/{itr.normalized_name}",
+                iterate_on='user.projects',
+            )
+        ],
+    )
+    config.register_origin_cache_keys(
+        Email.primary,
+        purge_keys=[
+            key_factory("user/{obj.user.username}"),
+            key_factory(
+                "project/{itr.normalized_name}",
+                iterate_on='user.projects',
+            )
+        ],
+    )
```

dstufft:
@di did you forget to push something? That's not what I see in the changes tab.

dstufft:
Oh wait I see it, fucking git.

warehouse/packaging/__init__.py
```diff
@@ -49,11 +60,24 @@ def includeme(config):
     config.register_origin_cache_keys(
         User,
         cache_keys=["user/{obj.username}"],
+    )
+    config.register_origin_cache_keys(
+        User.name,
```
di: @dstufft Adding these lines are removing the purge keys from `User` and adding the same ones for `User.name`. Sorry I didn't split this into two commits, that may have been more clear.

di:
@ewdurbin On it.

PR #3355, opened 2018-03-22
begin testing against postgresql 10.1

ewdurbin:


di:
If we want Postgres 10 on travis looks like we'll need to do this for now: 

```yaml
env:
  global:
  - PGPORT=5433
addons:
  postgresql: 10
  apt:
    packages:
    - postgresql-10
    - postgresql-client-10
```

Issue #3361, opened 2018-03-22
Dismissing pre-production banner does nothing


dstufft:
Going to pypi.org and hitting the "x" to dismiss the pre-production banner appears to do nothing. The banner is still there and nothing seems to have happened.

alex:
In firefox I see a `TypeError: t is null` in the console.

di:
I think that's unrelated, it's been doing that for while and I think there's an open PR to fix it?

PR #3370, opened 2018-03-22
Warn if notification target is missing id attribute

di:
Use a version of `-1` by default, and give a warning if we try to dismiss a notification that has no `id`.

(Towards fixing #3361)

### DIFF
--- a/warehouse/static/js/warehouse/controllers/notification_controller.js
+++ b/warehouse/static/js/warehouse/controllers/notification_controller.js
@@ -27,8 +27,8 @@ export default class extends Controller {
    */
   _getNotificationId() {
     /** Get data from `data-notification-version` attribute */
-    const version = this.data.get("version");
-    if (this.notificationTarget.id && version) {
+    const version = this.data.get("version") || "-1";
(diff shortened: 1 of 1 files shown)


warehouse/static/js/warehouse/controllers/notification_controller.js
```diff
@@ -46,7 +46,9 @@ export default class extends Controller {
     const notificationId = this._getNotificationId();
     if (notificationId) {
       localStorage.setItem(notificationId, 1);
+      this.notificationTarget.classList.remove("notification-bar--visible");
+    } else {
+      console.warn(`${this.notificationTarget} is dismissable but has no id attribute`); // eslint-disable-line no-console
```
dstufft: We support dismissable items without an ID I think? Those were the "emphereal" ones from my originally comment, and are things like flash messages which are going to go away on the next page load anyways (and so they don't need stored).

di: Yeah, I forgot about that, fixing...

PR #3389, opened 2018-03-23
log purges

ewdurbin:


tests/unit/cache/origin/test_fastly.py
```diff
@@ -28,6 +28,9 @@ def test_purges_successfully(self, monkeypatch):
         cacher = pretend.stub(purge_key=pretend.call_recorder(lambda k: None))
         request = pretend.stub(
             find_service=pretend.call_recorder(lambda iface: cacher),
+            log=pretend.stub(
+                info=pretend.call_recorder(lambda *args, **kwargs: None),
```
di: No sense in making these `call_recorder`s if we're not making assertions on `request.log.info.calls`.

ewdurbin: resolved in 18929d7f5afb62ec75869590a13baca1c94421d5

Issue #3248, opened 2018-03-14
Pressing Esc key should close modals


waseem18:
Right now clicking on `x` icon present on the top-right of the modal is the only way to close it.

Ideally the modal would also be closed when the user presses their `Esc` key.

di:
Revised this to be more clear that this is about closing the modal when the user presses the `ESC` key on their keyboard.

jMuzsik:
Hi, it's cool I work on this? 

I can continue working on first issues as well. Open to whatever. Any issues you'd recommend?

But as for this one, are there keywords that I can use to assure that I find each modal? As in, do they all have a class of `modal` or will this demand going through the site and finding each instance of a modal? If that is so, do you mind pointing me to a few to start? 

Thanks!

di:
@jMuzsik Absolutely! If you feel comfortable moving on from the "good first issues" then feel free. This would be a good one.

All the modals are created with this macro, so you should only have to change it once: https://github.com/pypa/warehouse/blob/fe7d676886a58a3500afe40181595ec4d50f0f51/warehouse/templates/manage/manage_base.html#L61-L93

jMuzsik:
I can work on #3198 at the same time because it is in the exact same place, sounds good?

di:
Fixed in #3392.

PR #3392, opened 2018-03-23
Exit modals by pressing escape button

jMuzsik:
In reference to #3248.

I created both files, one to escape and one to automatically focus because I could not find a way to escape while focusing on different DOM elements then the input field. I think it is because the modal takes up the entirety of the page by default (hidden or visible), so the user does not automatically focus on any particular element that can be escaped from. 

I had to setTimeout because... I suppose the DOM is not ready immediately and I did a click event so the controllers work occurs and the text resets to an empty string.

Though, most of these frameworks which I worked with are quite new to me so let me know if I should change/add/delete anything. Thanks!


jMuzsik:
Ok @di will change it asap. Though I initially did bind an event listener and soon realized that I could not find a way to call the stimulus controller so as to reset the input string to be empty. That is why I decided to click the close button, so that function automatically gets called.

If you know a way I can call the stimulus function in the event listener, please let me know, I spent some time researching and did not find a way that was not convoluted.

di:
Ah right, the controller. I think you can get ahold of it with something like `application.getControllerForElementAndIdentifier(element, id)`

jMuzsik:
I attempted to do this but for one reason or another I could not, I used this resource: https://github.com/stimulusjs/stimulus/issues/35, as it was the only one I could find but the `this` context was not being agreeable in the slightest no matter what I tried. `this.application` was defined as an error message, nothing else. Let me know if I should keep working on doing it or not.

And a screencast:
mar-24-2018 12-28-32 pypa2

Though, the pointer isn't displayed so it doesn't have an exactly ideal view

jMuzsik:
I moved some of the logic into the controller but had to keep the focus outside of the controller as the button that triggers the modal is not a part of the controller. Thanks!

di:
There's still some issues here:

* This doesn't seem to work on pages that have multiple modals (probably due to the `const modal = modals[0];` line)
* If it did work on pages that had multiple modals, this would be binding a new event listener for every modal, which seems unnecessary
* Trying to focus the individual element we've bound the event handler to seems somewhat hacky and brittle

I think it'd be better if we bound a single event listener to `document`, that changes the window location, and cancels all the modal controllers that are present. Something like:

```js
docReady(() => {
  document.addEventListener("keydown", event => {
    if (event.keyCode === 27) {
      window.location.href = "#modal-close";
      for (let element of document.querySelectorAll(".modal")) {
        application
          .getControllerForElementAndIdentifier(element, "confirm")
          .cancel();
      }
    }
  });
});
```

jMuzsik:
@di you are completely right, that is all that is needed, apologise for being completely off. I doubt I ought to do this commit seeing that the code you wrote is the exact solution and all that I wrote is, in a way, not useable. So I can close this or something along those lines if that is preferable. 

Either way, thanks for the explanation, I do not know why I did not think to add the event listener to the entire document.

PR #3393, opened 2018-03-23
Read only AdminFlag

### LINKED ISSUE #164: Implement a Read Only Flag to prevent any write operations
There are times when we need to prevent any state modification on pypi.python.org but we do not need to prevent reading from the database or other storage areas. It would be great if we could prevent all write operations using a toggle somehow.

di:
Fixes #164.

* Adds a `read-only` admin flag which dooms the transaction
* Adds request methods for admin flags (`request.flags.enabled` and `request.flags.all`)
* Adds a `AdminFlag.notify` attribute to determine if the flag should be displayed as a notification bar
* Shows a notification bar for any notifiable admin flags
* Admin UI for editing flags


warehouse/db.py
```diff
@@ -162,6 +162,12 @@ def cleanup(request):
         session.close()
         connection.close()
 
+    # Check if we're in read-only mode
+    from warehouse.admin.flags import AdminFlag
+    flag = session.query(AdminFlag).get('read-only')
+    if flag and flag.enabled:
```
ewdurbin: can we get an overall admin user short circuit here as well? that would alleviate the need to except for this in `edit_flag` and avoid having to except this for each and every operation an admin may need to perform during read-only situations.

warehouse/admin/views/flags.py
```diff
@@ -0,0 +1,55 @@
...
+    uses_session=True,
+    require_methods=False,
+    require_csrf=True,
+)
+def edit_flag(request):
+    flag = request.db.query(AdminFlag).get(request.POST['id'])
+    flag.description = request.POST['description']
+    flag.enabled = bool(request.POST.get('enabled'))
+
+    if flag.id == 'read-only' and request.tm.isDoomed():
```
ewdurbin: not sure having to add this clause for every admin action we add down the line is a a great approach. suggest we do this once in warehouse.db

dstufft:
What's the behavior that users are going to see if they attempt to do something that writes to the DB when we're in read only mode?

di:
@dstufft It will look like the write has succeeded, but it won't, and there will be a big red notification banner saying "Read Only Mode: Any write operations will have no effect"

We probably need to add a special case to fail upload since twine users won't see the notification banner.

PR #3396, opened 2018-03-23
Purge projects and user pages on role changes

### LINKED ISSUE #3383: Missing Purges
Noticed this while setting up new mirror. We don't seem to be purging `project/<normalized_name>` key when projects are deleted.

This leads bandersnatch to get confused and fall behind until the key is purged so the JSON api returns a 404

di:
Fixes #3383.

Also confirmed that project/release deletes _do_ purge the correct caches.

### DIFF
--- a/tests/unit/packaging/test_init.py
+++ b/tests/unit/packaging/test_init.py
@@ -18,7 +18,7 @@ from celery.schedules import crontab
 from warehouse import packaging
 from warehouse.accounts.models import Email, User
 from warehouse.packaging.interfaces import IFileStorage
-from warehouse.packaging.models import Project, Release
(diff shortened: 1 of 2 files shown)


ewdurbin:
🙏

ewdurbin:
hmmmm, doesn't seem to have stuck. 

`repokid` had an owner added in serial 3701601, but JSON api doesn't seem to have been purged.

```
DEBUG: Expected PyPI serial 3701601 for request https://pypi.org/pypi/repokid/json but got 3701586
```

ewdurbin:
further analysis shows the tasks _are_ being created and run. investigating if the purges are failing internally with our CDN provider.

PR #3399, opened 2018-03-24
Add mirror html template

nlhkabu:
ping @ewdurbin


ewdurbin:
@nlhkabu all yours now :)

nlhkabu:
Just reviewed the page, it looks fine as is :)
@ewdurbin I notice that the timezone isn't loading again?

ewdurbin:
Time stamp will only load on a mirror, as it requires the last-updated file to exist.

nlhkabu:
@ewdurbin I assume this needs to be a full stand-alone HTML page (not using the base template)?  The CI failures all seem to relate to the fact that this is not a standard Jinja template (inline CSS, no blocks, etc.). Do you have any advice on how to get around this?

ewdurbin:
@nlhkabu could it be as simple as moving the static page into `warehouse/static/html/mirror.html`?

Issue #991, opened 2016-02-29
Project Rating 


williamjmorenor:
Looking at the warehouse interface shuold be posible to show some rating, maybe a stars based system.

demianbrecht:
+1 to this feature. The only current measure of any sort is the number of downloads which is pretty much meaningless. This would at least give us a mechanism to have *some* user-driven feedback.

nlhkabu:
I agree that it would be good to have some kind of public measure of popularity (particularly for search ordering - see #702), but I think we need to be very careful here.  We need to particularly think about the impact a poor rating might have on a package author.  Is the review fair and accurate?  What recourse do they have to respond?

Importantly, is a rating system going to dissuade newcomers from publishing their packages because of fear of public criticism?

My alternative idea is to mimic the star system provided by GitHub.  This would give each package some kind of popularity count, while also allowing package consumers to keep track of their favourite packages.  It also carries no 'negative' feedback mechanism, which is good for package authors :)

demianbrecht:
Good point about negative feedback. However, I'd argue that the number stars would be akin to download count and doesn't *really* provide any real benefit.

Thinking about this a little more, the problem with reviews is that they're subjective by nature, so accuracy will always be arguable. Perhaps rather than stars, we could have a well defined quality matrix that can be represented visually in a variety of ways. Simple items could be marked off, for example:

Does the project have:
* Some form of CI setup (i.e. Travis-CI et al), although this one can be tricky for pull CI systems
* Some form of static analysis
* Unit tests

I'm not entirely sold on this yet, just thinking out loud really..

xavfernandez:
This could also be based on the fact that a bunch of other well-rated packages depends (via `install_requires`) on this particular package.

dstufft:
One thing to think about here is that in the past, PyPI did have a review and star rating system. Package authors got *really* upset about it and fought tooth and nail to get it removed (which it ultimately did get removed). That was before I was involved in the Python community at all, so I don't know specifics about why people were really against it. 

I know that @jacobian was one of the vocal opponents to the system as it existed back then, I'm not entirely sure if the main problem with it was the *implementation* or the *entire idea*. He might be willing to come in here and talk a bit about it, but there was a fairly long thread on catalog-sig about it [here](https://mail.python.org/pipermail/catalog-sig/2011-April/003542.html). Talking to him briefly, he said:

> my argument at the time -- which I still agree with -- is that ratings come down to a popularity contest, which is unfun

[...]

PR #3400, opened 2018-03-24
Add user help on PyPI statistics

brainwane:
Addresses the FAQ half of #991.


brainwane:
Ready for re-review.

I've updated other external links (to sites we do not host) as well.

PR #3401, opened 2018-03-24
purge projects on file removal

ewdurbin:


warehouse/packaging/__init__.py
```diff
@@ -42,6 +42,13 @@ def includeme(config):
     config.register_service_factory(storage_class.create_service, IFileStorage)
 
     # Register our origin cache keys
+    config.register_origin_cache_keys(
+        File,
+        cache_keys=["project/{obj.release.project.normalized_name}"],
```
dstufft: While this won't hurt, I also don't think it's required either. I don't think the ``File`` object is ever the context for a request. Just as an FYI.

ewdurbin: if it ever does, we covered :-D

PR #3403, opened 2018-03-25
Don't access the session on every view

dstufft:


warehouse/sessions.py
```diff
@@ -279,22 +279,33 @@ def session_view(view, info):
         # with a wrapper that just ensures that the session cannot be used.
         @functools.wraps(view)
         def wrapped(context, request):
+            # This whole method is a little bit of an odd duck, we want to make
+            # sure that we don't actually *access* request.session, because
+            # doing so triggers the machinaery to create a new session. So
```
ewdurbin: nit: typo `machinery`

ewdurbin:
should be incredibly beneficial as an interim solution for our non-session'd api routes.

PR #3413, opened 2018-03-25
Destroy documentation

### LINKED ISSUE #582: Add the ability to delete "legacy" documentation
If a project has existing legacy documentation hosted on pythonhosted.org, we should allow them to delete this documentation. This should come with appropriate warnings that this is a one time, destructive operation that cannot be undone, that new uploads of documentation are no longer allowed and that they should not do it until they've moved their documentation hosting elsewhere.

We should probably include at least a suggestion of using ReadTheDocs here.

This should also provide people with the option to add a redirect when deleting their documentation so that they can redirect to their new location.

ewdurbin:
Resolves #582 

Only displays admin tab for projects which have documentation on pythonhosted.org

I'm going to spot check the bucket and our DB and make sure it's in sync before we deploy

~~Requires new configuration~~ New configuration added:

## test.pypi.org
```
DOCS_BACKEND=warehouse.packaging.services.S3FileStorage bucket=pypi-docs-staging
```

## pypi.org
```
DOCS_BACKEND=warehouse.packaging.services.S3FileStorage bucket=pypi-docs
```

### DIFF
--- a/warehouse/utils/project.py
+++ b/warehouse/utils/project.py
@@ -13,7 +13,19 @@
 from packaging.utils import canonicalize_name
 from pyramid.httpexceptions import HTTPSeeOther
 
+from warehouse.tasks import task
+from warehouse.packaging.interfaces import IFileStorage
 from warehouse.packaging.models import JournalEntry
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
+def remove_documentation(task, request, project_name):
+    request.log.info("Removing documentation for %s", project_name)
+    storage = request.find_service(IFileStorage, name='docs')
+    try:
+        storage.remove_by_prefix(project_name)
+    except Exception as exc:
+        task.retry(exc=exc)
 
 
 def confirm_project(project, request, fail_route):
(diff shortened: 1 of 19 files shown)


ewdurbin:
screen shot 2018-03-25 at 12 54 42 pm
screen shot 2018-03-25 at 12 54 45 pm
screen shot 2018-03-25 at 12 54 47 pm
screen shot 2018-03-25 at 12 54 57 pm
screen shot 2018-03-25 at 12 55 00 pm
screen shot 2018-03-25 at 12 55 09 pm

warehouse/utils/project.py
```diff
@@ -61,3 +73,26 @@ def remove_project(project, request, flash=True):
...
+            submitted_by=request.user,
+            submitted_from=request.remote_addr,
+        )
+    )
+
+    request.task(remove_documentation).delay(project.name)
+
+    project.has_docs = False
+    # Flush so we can repeat this multiple times if necessary
+    request.db.flush()
```
dstufft: Do we really need a flush here? That really surprises me that it's required.

ewdurbin: eh, probably not, definitely just copy pasta :)

Issue #3198, opened 2018-03-09
"Confirm the username" should be "Confirm your username"


nlhkabu:
On the modal when a user wants to delete their account.   Currently we show this message:

![screenshot from 2018-03-09 06-49-17](https://user-images.githubusercontent.com/3323703/37194447-29f9f37a-2366-11e8-94c7-9c4dbc916626.png)

"Confirm the username" should be "Confirm your username"

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

di:
This might be a little tricky, because in order to make the modal reusable, I made this just say "Confirm the ___ to continue" where the value is the same as the label.

nlhkabu:
Thanks for the feedback @di - given this, should we remove the `good first issue` tag and description?

di:
Nah, I think it's still doable, just leaving some clues. 🙂

AnneTheAgile:
It looks like this should be closed or marked duplicate and closed, based on #3415

di:
Thanks @AnneTheAgile!

PR #3415, opened 2018-03-25
Confirm your username instead of the username

jMuzsik:
In reference to #3198:

mar-25-2018 16-53-38 your_the

Let me know if there is anything other then those instances to change, other then only the or your


nlhkabu:
Thank you @jMuzsik !

PR #3420, opened 2018-03-26
Add timeout for placing warning message on top (#3380)

### LINKED ISSUE #3380: notification bars no longer stack on load
When the page loads, the 'sticky' notification bar completely covers the donation bar:

screenshot from 2018-03-23 06-41-37

If you resize the browser, the display reverts back to what it should be:

screenshot from 2018-03-23 06-41-30

IIRC, the sticky bar works by calculating the height of all sticky notifications and then adding this padding to the body:

screenshot from 2018-03-23 06-45-08

This seems to be working fine when the page is resized, but not on load

yeraydiazdiaz:
Fixes #3380 

preprod_warning

Disclaimer: I didn't hack the private repo, just copied and pasted the content into `warning-banner.html` 😅

### DIFF
--- a/warehouse/static/js/warehouse/controllers/notification_controller.js
+++ b/warehouse/static/js/warehouse/controllers/notification_controller.js
@@ -12,6 +12,7 @@
  * limitations under the License.
  */
 import { Controller } from "stimulus";
+import positionWarning from "../utils/position-warning";
 
 export default class extends Controller {
   static targets = ["notification", "notificationDismiss"];
@@ -33,10 +34,6 @@ export default class extends Controller {
     }
   }
 
-  get isSticky() {
-    return this.notificationTarget.parentNode.classList.contains("js-stick-to-top");
-  }
-
   initialize() {
     const notificationId = this._getNotificationId();
     const isDismissable = this.notificationTarget.classList.contains("notification-bar--dismissable");
@@ -55,9 +52,6 @@ export default class extends Controller {
       localStorage.setItem(notificationId, 1);
     }
     this.notificationTarget.classList.remove("notification-bar--visible");
-    if (this.isSticky) {
-      let bodyElement = document.querySelector("body");
-      bodyElement.style.paddingTop = "0px";
-    }
+    positionWarning();
   }
 }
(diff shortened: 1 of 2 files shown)


brainwane:
@yeraydiazdiaz sorry for the wait on this. Dustin is working on another project this week and so Warehouse reviews have slowed down a bit; next week should be faster!

nlhkabu:
@lgh2 could you possibly test this, including on a mobile device?  Thanks

warehouse/static/js/warehouse/controllers/notification_controller.js
```diff
@@ -51,5 +55,9 @@ export default class extends Controller {
       localStorage.setItem(notificationId, 1);
     }
     this.notificationTarget.classList.remove("notification-bar--visible");
+    if (this.isSticky) {
+      let bodyElement = document.querySelector("body");
+      bodyElement.style.paddingTop = "0px";
+    }
```
di: Can you explain why this is necessary? This PR seems to solve the issue fine without the changes here, and I don't think we really want to be adding style attributes in the stimulus controllers.

yeraydiazdiaz: without this when dismissing the warning banner the `padding-top` on the body will stay at the previous value, effectively not collapsing

preprod_warning_collapse

I agree it's not very elegant to change the `body` style from the Stimulus controller, but given the dismissal logic lives there I couldn't think of a better way to collapse the body on dismissal.

nlhkabu: Is this possibly related to https://github.com/pypa/warehouse/issues/3519 ?

di: @nlhkabu Yes, seems likely.

nlhkabu:
@yeraydiazdiaz would you like me to test this on different devices/browsers?

di:
@nlhkabu Let's just merge this, I'm pretty sure it will work everywhere (and this banner won't be around for much longer anyways).

Issue #1303, opened 2016-06-30
Implement pagination on the homepage


nlhkabu:
The original design specifies that there should be 'show more' buttons at the bottom of the package lists on the homepage.  These buttons will be removed for launch (see #1302), but it would be nice to implement this feature in the future.

![screenshot from 2016-06-30 06-49-24](https://cloud.githubusercontent.com/assets/3323703/16478045/ebcd6d92-3e8e-11e6-8305-8c35873a653b.png)

nlhkabu:
As discussed on https://github.com/pypa/warehouse/issues/3447 the optimal solution for this is to add a "new release" and "trending" sort orders in the search results interface, and then link to this from these buttons.

nlhkabu:
Closing as no longer needed.

PR #3424, opened 2018-03-26
Show more button on home page implementation

jMuzsik:
Hi, in reference to #1303. This is what I did:

pypi_issue_1303

Some things to mention: 
- If you look at the video you'll notice that there is sometimes this gimicky behaviour when I reload quickly. I tried a handful of ways to fix this but think the only real way is inline styling, setting `display:none` automatically when the package box is created but that is not allowed. This or something can be done in python in which case I have no clue where to even begin.
- I added this little bit of logic: `{% set summary = release.summary if release.summary else "No description provided" %}` to get rid of the creation of an empty paragraph element that caused the boxes being even which you'll notice exists in the issue image.
- Last, basically all these frameworks are new to me so I am not sure if I am doing best practices.

Thanks!

### DIFF
--- a/warehouse/static/js/warehouse/controllers/increment_controller.js
+++ b/warehouse/static/js/warehouse/controllers/increment_controller.js
@@ -1,0 +1,61 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Controller } from "stimulus";
+import $ from "jquery";
+
+export default class extends Controller {
+  static targets = ["button"];
+
+  constructor() {
+    super();
+    this.trending_projects_visible = 5;
+    this.latest_releases_visible = 5;
+  }
+  addFiveTrending() {
+    const trendingElements = document.getElementsByClassName(
+      "hide-by-index-trending"
+    );
+    let j = this.trending_projects_visible;
+    const end = (this.trending_projects_visible += 5);
+    if (this.trending_projects_visible <= 20) {
+      for (; j < end; j++) {
+        if (trendingElements[j]) {
+          trendingElements[j].style.display = "block";
+        }
+      }
+      console.log(end);
+      if (end === 20) {
+        document.getElementById("increment-button-trending").style.display = "none";
+      }
+    }
+  }
+  addFiveLatest() {
+    const latestElements = document.getElementsByClassName(
+      "hide-by-index-latest"
+    );
+    let j = this.latest_releases_visible;
+    const end = (this.latest_releases_visible += 5);
+    if (this.latest_releases_visible <= 20) {
+      for (; j < end; j++) {
+        if (latestElements[j]) {
+          latestElements[j].style.display = "block";
+        }
+      }
+    }
+    console.log(end);
+    if (end === 20) {
+      document.getElementById("increment-button-latest").style.display = "none";
+    }
+  }
+}
--- a/warehouse/static/js/warehouse/controllers/increment_controller.js
+++ b/warehouse/static/js/warehouse/controllers/increment_controller.js
@@ -12,50 +12,36 @@
  * limitations under the License.
  */
 import { Controller } from "stimulus";
-import $ from "jquery";
 
 export default class extends Controller {
   static targets = ["button"];
 
+  // initially set default to 5 projects visible
   constructor() {
     super();
-    this.trending_projects_visible = 5;
-    this.latest_releases_visible = 5;
+    this.trending_visible = 5;
+    this.latest_visible = 5;
   }
-  addFiveTrending() {
-    const trendingElements = document.getElementsByClassName(
-      "hide-by-index-trending"
+  // Increase amount visible during each click event of Show More button
+  addFive(e) {
+    const eventType = e.target.value;
+    const elements = document.getElementsByClassName(
+      `hide-by-index-${eventType}`
     );
-    let j = this.trending_projects_visible;
-    const end = (this.trending_projects_visible += 5);
-    if (this.trending_projects_visible <= 20) {
+    // j is previous amount visible, end is how many will be
+    let j = this[`${eventType}_visible`];
+    const end = (this[`${eventType}_visible`] += 5);
+    if (end <= 20) {
       for (; j < end; j++) {
-        if (trendingElements[j]) {
-          trendingElements[j].style.display = "block";
+        if (elements[j]) {
+          elements[j].style.display = "block";
         }
       }
-      console.log(end);
+      // Hide Show More button when max amount is visible
       if (end === 20) {
-        document.getElementById("increment-button-trending").style.display = "none";
+        document.getElementById(`increment-button-${eventType}`).style.display =
+          "none";
       }
-    }
-  }
-  addFiveLatest() {
-    const latestElements = document.getElementsByClassName(
-      "hide-by-index-latest"
-    );
-    let j = this.latest_releases_visible;
-    const end = (this.latest_releases_visible += 5);
-    if (this.latest_releases_visible <= 20) {
-      for (; j < end; j++) {
-        if (latestElements[j]) {
-          latestElements[j].style.display = "block";
-        }
-      }
-    }
-    console.log(end);
-    if (end === 20) {
-      document.getElementById("increment-button-latest").style.display = "none";
     }
   }
 }
--- a/warehouse/static/js/warehouse/utils/show-more-initial.js
+++ b/warehouse/static/js/warehouse/utils/show-more-initial.js
@@ -1,0 +1,31 @@
+/* Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export default () => {
+  const trendingElements = document.getElementsByClassName(
+    "hide-by-index-trending"
+  );
+  const latestElements = document.getElementsByClassName(
+    "hide-by-index-latest"
+  );
+  // Hide all elements beyond 5 initially
+  for (let i = 5; i <= 20; i++) {
+    // ATM, with mock data? There are someimtes only 19 trending projects
+    if (trendingElements[i]) {
+      trendingElements[i].style.display = "none";
+    }
+    if (latestElements[i]) {
+      latestElements[i].style.display = "none";
+    }
+  }
+};
--- a/warehouse/templates/index.html
+++ b/warehouse/templates/index.html
@@ -15,13 +15,28 @@
 
 {% block title %}PyPI - the Python Package Index{% endblock %}
 
-{% macro project_snippet(release) -%}
-<div class="package-snippet">
+{% set trending_projects_visible = namespace(amount=20) %}
+{% set latest_releases_visible = namespace(amount=20) %}
+
+{% macro project_snippet_latest(release, summary, index) -%}
+<div class='package-snippet hide-by-index-latest'>
   <h3 class="package-snippet__title">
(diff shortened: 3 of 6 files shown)


brainwane:
Thank you for the PR, @jMuzsik ! @yeraydiazdiaz , could you review?

warehouse/templates/index.html
```diff
@@ -84,35 +97,39 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
...
       </div>
-      <div class="package-list">
+      <div class="package-list-timeout">
+        {% set current = namespace(amount_to_show=0) %}
         {% for release in trending_projects %}
-          {{ project_snippet(release) }}
+          {% if current.amount_to_show < trending_projects_visible.total %}
+            {{ project_snippet_trending(release, release.summary, current.amount_to_show) }}
+            {% set current.amount_to_show = current.amount_to_show + 1 %}
+          {% endif %}
```
yeraydiazdiaz: this loop might be clearer simply iterating over an index? something like

```
{% for i in range([trending_projects_visible.total, trending_projects|length]|min) %}
  {{ project_snippet_trending(trending_projects[i], trending_projects[i].summary, i) }}
{% endfor %}
```

(the range argument is admittedly awkward due to [`min`](http://jinja.pocoo.org/docs/dev/templates/#min) and [`length`](http://jinja.pocoo.org/docs/dev/templates/#length) being Jinja2 template filters)

warehouse/templates/index.html
```diff
@@ -15,16 +15,29 @@
...
 
-{% macro project_snippet(release) -%}
-<div class="package-snippet">
-  <h3 class="package-snippet__title">
-    <a href="{{ request.route_path('packaging.project', name=release.project.normalized_name) }}">{{ release.project.name }}</a>
-    <span class="package-snippet__version">{{ release.version }}</span>
-  </h3>
-  <p class="package-snippet__description">{{ release.summary }}</p>
+{% set trending_projects_visible = namespace(total=20) %}
+{% set latest_releases_visible = namespace(total=20) %}
```
yeraydiazdiaz: do you need the `namespace`? seems like it could simply be `latest_releases_visible_total = 20`

warehouse/templates/index.html
```diff
@@ -15,16 +15,29 @@
...
+
+{% macro project_snippet_latest(release, summary, index) -%}
+<div class="package-snippet hide-by-index-latest">
+  {{ project_snippet(release, summary, index)}}
+</div>
+{%- endmacro %}
+{% macro project_snippet_trending(release, summary, index) -%}
```
yeraydiazdiaz: these two macros differ only on the `class`, it might be cleaner to move the class to be an extra argument on the `project_snippet` macro below and just use that one?

warehouse/static/js/warehouse/controllers/increment_controller.js
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Controller } from "stimulus";
+
+export default class extends Controller {
+  static targets = ["button"];
+
+  // Initially set default to 5 projects visible
+  constructor() {
```
yeraydiazdiaz: rather than `constructor` [Stimulus encourages other methods](https://stimulusjs.org/handbook/managing-state#lifecycle-callbacks-explained), in this case I'd opt for `connect`

warehouse/static/js/warehouse/controllers/increment_controller.js
```diff
@@ -0,0 +1,47 @@
...
+    super();
+    this.trending_visible = 5;
+    this.latest_visible = 5;
```
yeraydiazdiaz: I think it might be neater and simpler if instead of having the controller wrapping both columns, having one instance of a controller per column, i.e. the `<div class="col-half">`.

That would make remove all the complexity below on parsing and retrieving the `eventType` and the separate `_visible` variables.

warehouse/templates/index.html
```diff
@@ -15,16 +15,29 @@
...
+<div class="package-snippet hide-by-index-trending">
+  {{ project_snippet(release, summary, index)}}
 </div>
 {%- endmacro %}
 
+{% macro project_snippet(release, summary, index) -%}
```
yeraydiazdiaz: do you need `summary` as an argument? seems you're retrieving it from `release` or defaulting anyway

warehouse/static/js/warehouse/utils/show-more-initial.js
```diff
@@ -0,0 +1,29 @@
...
+  );
+  for (let i = 5; i <= 20; i++) {
+    if (trendingElements[i]) {
+      trendingElements[i].style.display = "none";
+    }
+    if (latestElements[i]) {
+      latestElements[i].style.display = "none";
+    }
+  }
+};
```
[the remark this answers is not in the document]
jMuzsik: Ok I figured out how to use Stimulus properly (I believe). I got rid of any manipulation of the DOM without using Stimulus

warehouse/templates/index.html
```diff
@@ -84,35 +97,39 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
   </div>
 </section>
 
-<section class="horizontal-section horizontal-section--grey">
-  <div class="site-container">
+<section class="horizontal-section horizontal-section--grey" data-controller="increment">
```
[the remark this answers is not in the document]
jMuzsik: changed

warehouse/templates/index.html
```diff
@@ -84,35 +97,39 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
...
+  <div class="site-container" >
     <div class="col-half">
       <div class="heading-wsubtitle">
         <h2 class="heading-wsubtitle__heading">Trending Projects</h2>
         <p class="heading-wsubtitle__subtitle">Trending projects as downloaded by the community</p>
       </div>
-      <div class="package-list">
+      <div class="package-list-timeout">
```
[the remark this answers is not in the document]
jMuzsik: you're right, unnecessary

warehouse/templates/index.html
jMuzsik: made it much more clear now, actually how it was originally, I did not need all those extra jinga variables

warehouse/templates/index.html
jMuzsik: didn't need it at all, logic can be done in Stimulus

warehouse/templates/index.html
jMuzsik: Just noticed this, fixed it

warehouse/static/js/warehouse/controllers/increment_controller.js
jMuzsik: You're right, it was hard to read

warehouse/templates/index.html
jMuzsik: Nope, unnecessary, you are right

jMuzsik:
Thanks @yeraydiazdiaz! Believe that I have everything as you requested it to be. Definitely appreciated all the advice, definitely improved my ability with Stimulus and Jinga. 

Let me know if you'd like anything else to change

warehouse/static/js/warehouse/controllers/show_trending_projects_controller.js
```diff
@@ -0,0 +1,42 @@
+/**
```
yeraydiazdiaz: You don't need two controllers with the same code, Stimulus makes separate instances of the same class.

warehouse/templates/index.html
```diff
@@ -15,13 +15,18 @@
 
 {% block title %}PyPI - the Python Package Index{% endblock %}
 
-{% macro project_snippet(release) -%}
-<div class="package-snippet">
+{% macro project_snippet(release, class) -%}
+{% if class == "latest" %}
+<div class="package-snippet" data-target="show-latest-releases.latestRelease">
+{% else %}
+<div class="package-snippet" data-target="show-trending-projects.trendingProject">
```
yeraydiazdiaz: if you use a single controller, you can avoid this `if` completely and use something like `data-target="show-more.release"`

warehouse/templates/index.html
```diff
@@ -85,34 +90,30 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
 </section>
 
 <section class="horizontal-section horizontal-section--grey">
-  <div class="site-container">
-    <div class="col-half">
+  <div class="site-container" >
+    <div class="col-half" data-controller="show-trending-projects">
```
yeraydiazdiaz: you could simply use `data-controller="show-more"` here and below, Stimulus will create separate instances

warehouse/templates/index.html
```diff
@@ -85,34 +90,30 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
...
       <div class="package-list">
         {% for release in trending_projects %}
-          {{ project_snippet(release) }}
+          {{ project_snippet(release, "trending") }}
         {% endfor %}
       </div>
-      {#
-      <a class="button" href="TODO">Show More</a>
-      #}
+      <button class="button" type="button" data-action="click->show-trending-projects#addFive" data-target="show-trending-projects.addFiveButton">Show More</button>
```
yeraydiazdiaz: remember to change the names on these, there won't be clashing as they're separate DOM elements and instances

warehouse/templates/index.html
```diff
@@ -85,34 +90,30 @@ <h1 class="homepage-banner__title">Find, install and publish Python packages wit
...
-    <div class="col-half">
+    <div class="col-half" data-controller="show-latest-releases">
       <div class="heading-wsubtitle">
         <h2 class="heading-wsubtitle__heading">New Releases</h2>
         <p class="heading-wsubtitle__subtitle">Hot off the press: the newest project releases</p>
       </div>
       <div class="package-list">
         {% for release in latest_releases %}
-          {{ project_snippet(release) }}
+          {{ project_snippet(release, "latest") }}
```
yeraydiazdiaz: I think you can even remove the class parameter?

warehouse/static/js/warehouse/controllers/show-latest-releases-controller.js
```diff
@@ -0,0 +1,42 @@
...
+import { Controller } from "stimulus";
+
+export default class extends Controller {
+  static targets = ["latestRelease", "addFiveButton"];
+
+  // Initially set 5 projects to be visible
+  connect() {
+    this.amountVisible = 5;
+    let i = 0;
+    for (let element of this.latestReleaseTargets) {
```
yeraydiazdiaz: it might be more compact to use some [`Array` constructs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array):

```
this.releaseTargets.slice(this.amountVisible).forEach(element => {
  element.style.display = "none";
});
```

warehouse/templates/index.html
jMuzsik: I don't see how I can do this, `show-more.latestReleases` and `show-more.trendingProjects` are entirely different arrays, perhaps there is something clever I can do I don't think there is an explicit way I can do it

warehouse/templates/index.html
jMuzsik: not sure if that is possible, as I mentioned above

yeraydiazdiaz:
Sorry, what I meant was something like https://github.com/pypa/warehouse/commit/e7befb855449fbf2a9eb7a399d8f398c09d0b672

The key is adding two `data-controller` instances, one per column, instead of a general one for both.

It might sound odd because we're using the same names for the targets but they're local to each `data-controller` so no clashing occurs. The benefit is that the controller is much simpler and will work on any number of columns.

jMuzsik:
You are definitely right, I did not know that I could do this, but you basically wrote the entire PR based upon that code you wrote there. So perhaps you ought to send the PR? I mean if you want to I can close this and you can send it instead.

Either way, I'll put the working copy up. Thanks so much @yeraydiazdiaz!

yeraydiazdiaz:
@jMuzsik up for resolving the conflictsand bringing this one home? 😄

jMuzsik:
I didn't realise there were any! Thanks!

nlhkabu:
hi @jMuzsik thanks for this PR and thanks also to @yeraydiazdiaz for your reviews :)
Just to let you know - this is on my list of things to pull down and review this week - sorry for the delay and thanks for your patience!

di:
I think generally I'm in agreement with @pradyunsg. This also depends heavily on JS and won't work at all for users without JS or not in a modern browser, whereas linking to an ordered search result page would. I'll let have @nlhkabu have the final say here though.

nlhkabu:
I agree with @pradyunsg and @di on this.  If we were instead to direct the user to an ordered search results page, they could also apply search filters, which would give them more options. They would also benefit from any improvements we make to the search UI in the future.

@jMuzsik Is this something that you'd like to take on?

I have updated #1303 and #3447 accordingly.

jMuzsik:
It makes complete sense to do it as you guys have spoken of, though I cannot do it right now, I would say in a few weeks I can begin to but I really am not sure. But if anyone else would like to take it on, by all means, maybe @yeraydiazdiaz. Thanks!

nlhkabu:
ok great @jMuzsik :) Are you happy to close this PR in the meantime?

brainwane:
@jMuzsik Thank you for your work on this. Please do let us know if you want thoughts on stuff to take on when you're more ready, in a few weeks!

PR #3429, opened 2018-03-26
Add project.last_serial to simple.detail and legacy.api.json routes

ewdurbin:
This adds the project's latest serial to `/simple/<project_name>/` and `/pypi/<project_name>/json`.

As Warehouse uses Fastly Soft Purges, we need to tweak the ETag in order to notify the CDN that the page has changed. Currently the CDN performs a conditional GET in order to decide if it needs to pull refreshed content, and since the contents of these pages don't change for maintainer and role updates it skips purging them.

This is primarily a concern for clients like bandersnatch which use the X-PyPI-Last-Serial header to determine if they're getting the latest goods.

### DIFF
--- a/warehouse/legacy/api/json.py
+++ b/warehouse/legacy/api/json.py
@@ -186,6 +186,7 @@ def json_release(release, request):
             "bugtrack_url": project.bugtrack_url,
             "home_page": release.home_page,
             "download_url": release.download_url,
+            "last_serial": project.last_serial,
         },
         "urls": releases[release.version],
         "releases": releases,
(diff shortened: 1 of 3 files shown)


warehouse/legacy/api/json.py
```diff
@@ -186,6 +186,7 @@ def json_release(release, request):
             "bugtrack_url": project.bugtrack_url,
             "home_page": release.home_page,
             "download_url": release.download_url,
+            "last_serial": project.last_serial,
```
dstufft: This should probably be a top level key, it's not part of the release info.

ewdurbin: d'oh

PR #3457, opened 2018-03-28
fix bigquery integration for compute_trending task

### LINKED ISSUE #3455: Trending Projects are not updated
I think in the past month that I've been looking at pypi.org (and even before that), the "Trending Projects" haven't changed.

ewdurbin:
closes #3455, thanks @pradyunsg for report


warehouse/packaging/tasks.py
```diff
@@ -56,21 +56,11 @@ def compute_trending(request):
...
-        rows, total_rows, page_token = query.fetch_data(
-            max_results=1000,
-            page_token=page_token,
-        )
-
-        zscores.update(dict(rows))
-
-        if not page_token:
-            break
+    for row in query.result():
```
ewdurbin: this handles pagination for us now!

PR #3459, opened 2018-03-28
Reindex as celery task

ewdurbin:
this runs in ~1minute now... may as well make it a celery task and update it every 3 hours instead of nightly :)


warehouse/cli/search.py
```diff
@@ -10,11 +10,24 @@
...
+
+
+@search.command()
+@click.pass_obj
+def reindex(config):
+    """
+    Recreate the Search Index.
+    """
+
+    config.task(_reindex).apply_async()
```
ewdurbin: @dstufft I couldn't figure out how to get this to run "in the foreground".

I tried wrapping `apply()` in a way similar to `warehouse.tasks:WarehouseTask.apply_async()` but kept getting `transaction.interfaces.NoTransaction`

ewdurbin: runs great on the worker though!

dstufft: You probably have to do some setup to get the request in a reasonable state.

Issue #3163, opened 2018-03-07
Change delete file modal confirmation process


nlhkabu:
**From user testing:**

Currently, to delete a file, you must confirm the entire file name:

![screenshot from 2018-03-07 20-44-40](https://user-images.githubusercontent.com/3323703/37118147-2977e46e-224b-11e8-8d85-c379284e6cc2.png)

This is a little bit painful for some users, because the file name can get pretty long.

Instead, we should change this modal so that we require the `project name` for confirmation.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

brainwane:
@aalmazan could I ask for you to help with this?

aalmazan:
@brainwane Sure, however, I can't seem to find a good way to test uploading packages. Are there some docs you can point me to? Or at least a high-level of how I can create a test package? 

Would I need to follow [the packaging guide](https://packaging.python.org/tutorials/distributing-packages/) while having pypi.org point to localhost?

brainwane:
@aalmazan As a first step, if you haven't uploaded a package to PyPI before, I suggest you follow [the packaging guide](https://packaging.python.org/tutorials/distributing-packages/) and then [upload your package to Test PyPI](https://packaging.python.org/guides/using-testpypi/), just to see how the whole process works. [Here's a toy project I have set up reasonably well](https://gitlab.com/brainwane/form990s) in case you want to clone and use that (you'll have to modify the package name in `setup.py` so Test PyPI will let you upload it).

I am realizing now that [our Getting Started guide](https://warehouse.readthedocs.io/development/getting-started/) lacks information on uploading packages locally. @waseem18 and @lgh2 I think you've dealt with this -- how did you manage it? Whatever you advise, we should write it up and add it to the docs!

brainwane:
[...]

PR #3378, opened 2018-03-23
Change delete file modal confirmation process

aalmazan:
Fixes #3163. 

* Adds an additional optional parameter for the `confirm_modal` macro which will display after the warning box.
* Project name will be used instead of filename to confirm delete.

PR #3467, opened 2018-03-29
Remove placeholder text for deletion confirmation model

venthur:
Fixes the placeholder issues and the case of "Project Name" the confirmation of the file deletion via project name will be fixed through #3163


brainwane:
Thanks for the PR, @venthur! @aalmazan could I ask you to look at this?

aalmazan:
@brainwane Took a look and the first commit looks good. The second, however, may be introducing something unintended? I'm not sure what's expected, but that change will also change behavior in, for example, the delete file confirmation modal.

`fix/3162` now matches @nlhkabu 's [suggested example](https://github.com/pypa/warehouse/issues/3163#issuecomment-375195865), but I'm not sure if the removal of the placeholder was intentional. 

selection_030

vs. the current version on #3378 

selection_029

nlhkabu:
Yes, @aalmazan the removal of the placeholder is intentional.  We had feedback from user testing that this was confusing (some users didn't realise it was a placeholder - they thought it was real text).  I am happy to merge this if you are :)

aalmazan:
@nlhkabu Sorry about the delay (and I see it's already been merged haha), but no objections from me :+1:

brainwane:
@venthur Thanks for the fix! Let me know if you'd like a suggestion for another issue to work on?

PR #3475, opened 2018-03-30
deterministicly order journal entrys by id

### LINKED ISSUE #3474: Simultaneous events are not ordered consistently on project history page
Sometimes the "new release" event ends up after (above) the first upload of that release, other times it's the other way around:

image

(as reported in #pypa-dev)

ewdurbin:
Previously we sorted by submitted_date, which was troublesome as many JournalEntry objects may be committed in a single transaction, leading to identical timestamps.

SQLAlchemy retains the order of jounal creation, and flushes in order so ordering by id is deterministic and represents the logical order of operations.

Closes #3474, thanks to @anowlcalledjosh for reporting

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -1255,7 +1255,7 @@ class TestFileUpload:
         # Ensure that all of our journal entries have been created
         journals = (
             db_request.db.query(JournalEntry)
-                         .order_by("submitted_date")
+                         .order_by("id")
                          .all()
         )
         assert [
@@ -2160,7 +2160,7 @@ class TestFileUpload:
         # Ensure that all of our journal entries have been created
         journals = (
             db_request.db.query(JournalEntry)
-                         .order_by("submitted_date")
+                         .order_by("id")
                          .all()
         )
         assert [
@@ -2270,7 +2270,7 @@ class TestFileUpload:
         # Ensure that all of our journal entries have been created
         journals = (
             db_request.db.query(JournalEntry)
-                         .order_by("submitted_date")
+                         .order_by("id")
                          .all()
         )
         assert [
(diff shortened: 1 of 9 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -1255,7 +1255,7 @@ def storage_service_store(path, file_path, *, meta):
         # Ensure that all of our journal entries have been created
         journals = (
             db_request.db.query(JournalEntry)
-                         .order_by("submitted_date")
+                         .order_by("id")
```
dstufft: I think this is wrong-ish, if someone is doing something concurrently in two different transactions, the Ids might not  100% sense although I'm not sure that this can actually ever get exposed since IIRC we're using serializable transactions. I think we'd be better off sorting first by ``submitted_date``, and then falling back to ``id`` to tie break things with the same ``submitted_date``.

ewdurbin: If all the journal entries are created in the same request, they'll be created in the same transaction, so if we're using serializable transaction isolation we'll always have sequential ids.

however it is also unclear to me if this is the case, as our call to set the isolation level is inside a conditional:

https://github.com/pypa/warehouse/blob/a6d34187a91d4a7f872079d04d26c19eacbcec48/warehouse/db.py#L142-L150

ewdurbin: I had assumed we were set to SERIALIZABLE for all connections, so thanks for raising this.

ewdurbin: even still, the ids should be in the same order that the JournalEntry objects were added to the session, so I believe this test case is valid.

dstufft: I mean concurrent requests, so two different transactions, commiting at ~the same time. I don't know enough about PostgreSQL internals to know how it would resolve the conflict.

dstufft: IOW, could we get slightly different dates (because the two commits don't occur at the *exact* same time* with the IDs in the "wrong" order.

PR #3477, opened 2018-03-30
Fix text about classifier position

### LINKED ISSUE #3454: change filter reference to respond to search page width
pypi-narrow-no-classifiers

Per [this python-list post](https://mail.python.org/pipermail/python-list/2018-March/732168.html) -- a person seeing nothing on the left-hand side of https://pypi.org/search/ says,

> For me, that's a window width issue. The sidebar with the filters only shows when the window is wide enough. Unfortunately, the text mentioning it doesn't change, so this should be fixed.

nlhkabu:
Closes #3454 

screenshot from 2018-03-30 12-49-02
screenshot from 2018-03-30 12-48-49


nlhkabu:
Mobile will work the same as desktop - as it is based off resolution rather than device :)

Issue #3335, opened 2018-03-21
question: add "last released" to last-released timestamp?


brainwane:
Followup to #821 and breaking out of a feedback list in #2417.

The "About 5 days ago." text on [a sample project](https://pypi.org/project/sabyenc/) is under a "Latest version" box, but for those of us who see that as a glowing box and don't really look at the text it doesn't quite seem connected. Could/should we add a "last released" label, or visually connect it more to that "latest version" box?

di:
👍, this change is probably good for new contributors as well.

cheungnj:
I'd like to work on this!

cheungnj:
This is what I have so far. Is this what you're looking for? <img width="1154" alt="screen shot 2018-03-30 at 12 13 57 am" src="https://user-images.githubusercontent.com/6911370/38128446-7802ee38-33af-11e8-8f28-dffa99083fcc.png">

brainwane:
Thanks, @cheungnj -- I'll ask our designer @nlhkabu to take a look! A work-in-progress pull request would be great to see so we can also take a look at the code side.

cheungnj:
@brainwane I'll submit a PR so we can get the ball rolling on this! Thanks!

Issue #1988, opened 2017-05-13
Create 'condensed' design version


nlhkabu:
We have had a lot of positive feedback about the new design. However, there are a number of power users who have made it clear that they prefer a more condensed, data-rich layout.

One possible solution is to have different viewing modes available - a bit like gmail:

![screenshot from 2017-05-13 22-03-41](https://cloud.githubusercontent.com/assets/3323703/26029215/0fffcc62-3828-11e7-86a3-00743cac5c73.png)

To start with we could have "Standard mode" and "compact mode".

Standard mode would display the current layout. Compact mode would significantly reduce the page whitespace.

On the CSS side, this would be a matter of adding a class to the body and changes each style accordingly.

@dstufft what do you think?  How could we do this on your side?

dstufft:
I think the way of handling this in terms of implementation would be to have a separate cookie (that we filter out in the VCL so it doesn't affect caching) and implement the body class switching using javascript client side reading that cookie. If the user is logged in we can persist that setting server side using an ajax call to some internal endpoint and restore that persisted setting whenever the user logs in.

One thing I'd maybe push back on though is whether this is something that's worth spending time on. It seems like it's something that's going to increase the amount of effort of making any UI change since you're going to effectively need to design each change twice (or N times for each "mode") and since we don't have great UI level tests going, whatever the non-default mode is, is more likely to get broken unless we manually test each possible iteration.

To be clear, I'm not specifically against this, Mostly I'm questioning whether it's something we actually need. Is there in fact two distinct groups of users who need different UIs in order to effectively interact with the site, or can we satisfy both groups with a shared UI?

nlhkabu:
Yes, I agree, it does significantly increase the workload.

> Is there in fact two distinct groups of users who need different UIs in order to effectively interact with the site, or can we satisfy both groups with a shared UI?

The only way to verify this is to do some direct user testing. I have a list of volunteers for this, but not a lot of time to conduct them. In the short term, let's see what feedback we get on this issue.

rsyring:
FWIW, I think having two UIs is a waste of manpower.  One good UI should be sufficient IMO.  If the power users mentioned in the OP have convinced you that the current layout could be improved, then make those improvements.  If you aren't convinced, then leave the current layout.  But having two UI layouts to maintain/improve, just seems like a lot of work for very little benefit.

nkanaev:
[...]

PR #3481, opened 2018-03-30
update last released timestamp to visually connect it more to the latest version box

cheungnj:
This is a WIP to get quicker feedback regarding #3335 

This is what I have so far 
screen shot 2018-03-30 at 12 13 57 am


warehouse/filters.py
```diff
@@ -177,3 +177,8 @@ def contains_valid_uris(items):
 
 def parse_version(version_str):
     return packaging.version.parse(version_str)
+
+
+def uncapitalize(text):
```
cheungnj: jinja has the built-in capitalize filter so I could also:

1. remove this filter
2. update [convertToReadableText](https://github.com/cheungnj/warehouse/blob/add-last-released-to-timestamp/warehouse/static/js/warehouse/utils/timeago.js#L27) to return all lowercase values
3. then invoke the capitalize filter on any existing references to humanize

warehouse/templates/packaging/detail.html
```diff
@@ -91,7 +91,12 @@ <h1 class="package-header__name">
       {% else %}
       <a class="status-badge status-badge--good" href="{{ request.route_path('packaging.project', name=release.project.name) }}">Latest Version</a>
       {% endif %}
-      <p class="package-header__date">{{ humanize(release.created) }}</p>
+      <p class="package-header__date">
+        <span class="status-badge">
```
cheungnj: I wasn't sure if this was acceptable use of the `status-badge` class but I could also try using the `badge` class.

warehouse/filters.py
cheungnj: I could also use the `lower` filter. There's a lot of options...

brainwane:
@yeraydiazdiaz if you have any time to give your thoughts here, I'd appreciate it.

warehouse/templates/packaging/detail.html
yeraydiazdiaz: I think it's the appropriate class

warehouse/filters.py
yeraydiazdiaz: `lower` would be the easiest approach, yet it reads wrong when displaying a date, i.e `Released apr 19 2018` vs `Released Apr 19 2018` (ideally `Released on April 19 2018`).

I think we should separate the Jinja side from the Javascript side. If you're up for it we could:
- Extend the function `timeago.js` to read an optional suffix from a second attribute and deal with the capitalization if present.
- Remove `Released` text and the `uncapitalized` filter so it simply displays the date as before.

warehouse/static/sass/blocks/_package-header.scss
```diff
@@ -131,8 +135,18 @@
...
-    font-style: italic;
-    margin-top: 20px;
-    padding-bottom: 5px;
+    color: $text-color;
+
+    span {
+      margin-top: 10px;
+
+      &:before {
+        border-color: rgba(0, 0, 0, 0.4);
```
yeraydiazdiaz: maybe best to use the [border color defined in colours.scss](https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/settings/_colours.scss#L31)

warehouse/filters.py
cheungnj: Definitely up for it but if I'm reverting my changes so the date is displayed as before, then do we really need to extend `timeago.js`?

cheungnj:
This is what I have now. I addressed all of the comments besides extending the functionality of `timeago.js` because I wanted some clarification. 
screen shot 2018-04-01 at 12 47 16 pm

warehouse/filters.py
yeraydiazdiaz: I think prefixing the `Released` would be a nice touch, but it might not justify the change? thoughts @brainwane @nlhkabu?

warehouse/templates/packaging/detail.html
```diff
@@ -91,7 +91,12 @@ <h1 class="package-header__name">
...
+          {{ humanize(release.created) }}
```
di: I think prefixing this with a `Released: ` would be appropriate (not sure I totally follow the discussion in this PR about this).

cheungnj: Maybe @yeraydiazdiaz meant prefix instead of suffix and the presence of that would cause the human readable date to be lowercase?

di: I don't see how adding a prefix would make the date lowercase? To be clear, I'm talking about changing this line to something like:

```
Released on: {{ humanize(release.created) }}
```

cheungnj: I get that but doesn't `Released on: Just now` or `Released on: About 2 hours ago` sound kinda weird? I think the discussion earlier was talking about passing the optional prefix and if it was a truthy value then we'd make the return value of `convertToReadableText` lowercase. In that case we'd need to pass the prefix only when we're converting the value for the time element in `.package-header__date`.  Maybe just doing `Released:` is the way to go.

di: Yeah, I'll leave it to you to decide what sounds best, but I think the simplest approach possible here is probably ideal. `Last released:` might also work.

cheungnj: That sounds like a winner to me

nlhkabu:
hi @cheungnj thanks for this PR.  Just to let you know that this is on my list of things to pull down and review this week, so I'll have feedback for you soon :) Thanks for your patience.

nlhkabu:
hi @cheungnj thanks again for this PR :)

Right now, we have some feedback asking us to reduce the whitespace in our UIs (see #1988), so I think my preference here is to keep this style simple (and taking up less space), and just add the 'Last updated' text for now.

It is likely that we will redesign this header altogether based on user feedback, so I will keep this design in mind for then.

cheungnj:
Here's what I have now: 
screen shot 2018-04-28 at 1 58 36 pm

PR #3493, opened 2018-04-01
Move email handling into a service

dstufft:
This PR moves email handling into a service, this is done to prepare for eventual SES support (which will not go through SMTP, but rather through the SES API). Functionally speaking this doesn't change anything, but codewise a few APIs have changed:

* ``warehouse.email:send_email()`` no longer takes a ``recipients`` and a ``bcc`` kwarg, instead it takes a *single* ``recipient`` kwarg. 
  * By having a single send_email() call per email address, we eliminate issues where an address might be blocked at the server side, causing all addresses to fail to send.
  * Instead of using BCC to handle sending the same email to multiple people, this relies on a simple loop that calls ``warehouse.email:send_mail()`` multiple times. This is better because when end users receive the email, it will still have their email in the ``To:`` field, instead of no address.
* The order of subject and body being passed to ``warehouse.email:send_email()`` has changed and subject must be passed first.
  * This makes code match the traditional UI for sending email better, where the subject is first, and then the body follows. A minor UX improvement to writing email sending code.

This change does not introduce anything like configurable services for sending email, future PRs will include that, but this moves us closer to being able to swap out the email handling at runtime.


dstufft:
Note that this PR conflicts with https://github.com/pypa/warehouse/pull/3306 and whichever one lands second will have to be updated to handle the other.

PR #3503, opened 2018-04-02
remove rel attribute in api docs

lgh2:
Removes the `rel` attribute from the legacy /simple/ api docs, since this api no longer provides a `rel` attribute with the links it returns. (See [this discussion ](http://kafka.dcpython.org/day/pypa-dev/2018-04-02#14.03.54.sumanah)on IRC for why the `rel` was removed.)


docs/api-reference/legacy.rst
```diff
@@ -84,31 +63,26 @@ The Simple API implements the HTML-based package index API as specified in `PEP
...
         Accept: text/html
 
     **Example response**:
 
     .. code:: http
 
-        HTTP/1.0 200 OK
-        Content-Type: text/html; charset=utf-8
-        X-PyPI-Last-Serial: 867465
-
```
di: I think these lines should probably stay? You'll want to update the serial number to match the one in the comment below.

lgh2: Thank you, @di, I put those lines back. I am not sure if I put in the correct serial number.

di: The serial number can change at any time, so for this example it really can be anything as long as it's a) present and b) consistent.

PR #3555, opened 2018-04-05
Make color scheme WCAG 2.0 AA compliant

nlhkabu:
Closes #3334 

Includes code highlighting scheme:

screenshot from 2018-04-04 20-59-39

Note: this updates makes the Warehouse color scheme AA compliant, not *AAA* compliant.

PR #3556, opened 2018-04-05
Style focus states, use buttons where needed

nlhkabu:
1. Restyle focus states

screenshot from 2018-04-05 07-18-39
screenshot from 2018-04-05 07-18-56

2. Use buttons instead of `<a>` with no `href` for actions. This allows the button to be focused with a keyboard.

### DIFF
--- a/warehouse/static/sass/base/_forms.scss
+++ b/warehouse/static/sass/base/_forms.scss
@@ -34,7 +34,10 @@ select {
 #{$all-text-inputs-active},
 select:focus,
 select:hover,
-select:active {
+select:active,
+input[type="checkbox"]:focus,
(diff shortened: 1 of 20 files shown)


nlhkabu:
hi @di - we already have hover underline styles on most links :)  The footer links are the exception to this, but this is being changed with the footer update in #3555

di:
@nlhkabu I was suggesting that we add the same hover under underline when the element is focused via tabbing, right now this isn't the case.

nlhkabu:
ah, sorry, I misunderstood.  I'll give it a try and see what it looks like.  If it's good, I'll make a new PR :)

PR #3561, opened 2018-04-05
Add surrogate keys for HTML templates

### LINKED ISSUE #778: Document how to purge the HTTP cache when deploying the design
It'd be useful to document that you can do `curl -XPURGE https://warehouse.python.org/.../` when the design has been deployed to see the changes if they have been cached by Fastly.

di:
Fixes #778.

This adds a surrogate key `all-html` which can be used to purge all views rendered from an HTML template.

It also adds per-template surrogate keys as well, corresponding to the HTML template name (e.g. `accounts/request-password-reset.html`), so we can purge individual pages if they change.

### DIFF
--- a/docs/development/frontend.rst
+++ b/docs/development/frontend.rst
@@ -17,6 +17,27 @@ however, you can trigger a manual build of them by installing all of the
 dependencies using ``npm install`` and then running ``gulp dist``.
 
 
+Deploying
+---------
+
+When deploying changes to the frontend, to see the changes immediately it is
+necessary to purge the cache.
+
+Individual pages can be purged from the command line by issuing a ``XPURGE``
+request, e.g.::
+
+    curl -XPURGE https://pypi.org/the/page/to/purge
+
+All HTML pages can be dropped from the cache by purging the `all-html`
+surrogate key via our CDN provider.
+
+The entire cache can be purged by issuing a "Purge All" via our CDN provider.
+
+Purging the cache is not usually necesasry when making frontend changes, unless
+it would be unacceptable for the site to simultaneously have an "old" version
+of some pages, but the "new" version of others.
+
+
 Browser Support
 ---------------
 
--- a/docs/development/frontend.rst
+++ b/docs/development/frontend.rst
@@ -28,7 +28,7 @@ request, e.g.::
 
     curl -XPURGE https://pypi.org/the/page/to/purge
 
-All HTML pages can be dropped from the cache by purging the `all-html`
+All HTML pages can be dropped from the cache by purging the ``all-html``
 surrogate key via our CDN provider.
 
 The entire cache can be purged by issuing a "Purge All" via our CDN provider.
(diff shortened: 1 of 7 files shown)


warehouse/cache/origin/derivers.py
```diff
@@ -0,0 +1,38 @@
...
+            try:
+                cacher = request.find_service(IOriginCache)
+            except ValueError:
+                pass
+            else:
+                request.add_response_callback(
+                    functools.partial(
+                        cacher.cache,
+                        sorted(['all-html', renderer.name]),
+                        seconds=1 * 24 * 60 * 60,                 # 1 day
```
dstufft: This is probably wrong? We likely don't want to add a cache like this for every single rendered view automatically (and I'm not sure offhand what the behavior would be if something already declared a cache, which one would "win").

di: OK, I can see how this would break views that already have an `origin_cache` decorator. I'll think about that.

Maybe I'm misinterpreting #778 -- isn't the goal to have a surrogate key for every page rendered from a given template? Maybe this isn't really necessary if we can easily handle a  purge of `all-html` instead? 

Are you saying that automatically adding `all-html` doesn't make sense here either?

di: Or are you saying that we _only_ need to add the `Surrogate-Key` header, and shouldn't add these `Surrogate-Control` values?

dstufft: @di Adding the surrogate key is fine, it's the automatic addition of a ``CacheControl`` header to every response that I think is the wrong thing to do here. If it doesn't support it already, you might have to make the cache control stuff handle ``seconds=None`` etc.

As you mentioned, you'll also need to figure out a solution for things that already have a ``origin_cache`` header as well.

di:
@dstufft Ready for a re-review.

docs/development/frontend.rst
```diff
@@ -17,6 +17,27 @@ however, you can trigger a manual build of them by installing all of the
...
+
+When deploying changes to the frontend, to see the changes immediately it is
+necessary to purge the cache.
+
+Individual pages can be purged from the command line by issuing a ``XPURGE``
+request, e.g.::
+
+    curl -XPURGE https://pypi.org/the/page/to/purge
+
+All HTML pages can be dropped from the cache by purging the `all-html`
```
alex: Double backticks are needed

warehouse/cache/origin/fastly.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, *, api_key, service_id, purger):
         self.api_key = api_key
         self.service_id = service_id
         self._purger = purger
+        self.keys = set()
```
dstufft: I think this violates the API of the ``IOriginCache`` a bit, since technically a single ``IOriginCache`` can live through multiple requests (although the current implementation does not do that). This is alluded to by the fact that the ``cache()`` method requires passing in the ``request`` object.

I think this is along the right path, but I think probably you should just hang this list off of the request object somewhere. Another option is to store a dictionary of ``request: list(keys)``. This would likely need to be a ``weakref.WeakKeyDictionary`` to prevent having a memory leak.

PR #3580, opened 2018-04-07
Implement email sending via SES

dstufft:
This implements sending emails via Amazon SES. It has the following features:

- [x] Implements a SESEmailSender service that sends emails and logs them to the database.
- [x] Implements a Web hook that SES via SNS will call with various delivery related events.
- [x] Will Unflag emails as verified when they bounce or are marked as complaint, requiring users to re-verify their emails.
- [x] Provides an admin panel that will show the emails sent (omitting their body, incase the body has sensitive information in it) as well as the email status (accepted, delivered, bounced, complained) and any delivery events that were received about the email.
- [x] Provides a flag on the email to indicate it has been having delivery issues.
  - Long term goal of this flag is to provide better UX in the user's account management page to indicate that the reason this email is no longer verified is because of delivery issues. This PR does *not* implement that though.
- [x] Cleans up logged emails after 14 days. This is to keep the table from growing unbounded (see the journals table) and also to provide a finite lifespan on any email addresses or similar that get logged into this table.
- [x] Allow searching the email logs for specific addresses, etc.
- [x] Allows configuring the email sending backend similarly to the way the file storage backend is configured.
- [x] Handles flagging an account as having delivery problems after N transient delivery failures.
- [x] Authenticate/Validate that the web hook was actually called by SNS for the desired topic and it wasn't an attacker attempting to DoS the email handling.

Overall it still needs some cleanups as well (tests need written, some hardcoded region values, error handling is broken, etc). However it has all the basic functionality!

Here's some screenshots of the admin interface:

download 2

download 1

download

download 3


dstufft:
Note: Before we deploy this we will need to change the environment variables in our deploys, the new mechanism uses ``MAIL_BACKEND=<backend> foo=bar`` format that the storage and docs and such use.

dstufft:
I've hardcoded the number of transient bounces that an email address can have before it gets flagged for re-verification to 5. I'm not sure if that's a good number or not or if that should be generally configurable or not.

-  Whenever we get a soft bounce that *doesn't* occur after a delivery (i.e. not an OOTO auto-responder) increment the number of transient bounces by 1.
- Whenever we get a successful email delivery, reset the number of transient bounces to 0.
- Whenever the number of transient bounces is greater than 5, treat it as if the email had hard bounced and un-verify the email address.
- Whenever an email gets verified (either the first time, or any subsequent times) we reset the number of transient bounces back to 0.

warehouse/utils/sns.py
```diff
@@ -0,0 +1,136 @@
...
+        try:
+            pubkey.verify(signature, data, PKCS1v15(), SHA1())
+        except _InvalidSignature:
+            raise InvalidMessage("Invalid Signature") from None
+
+    def _get_pubkey(self, cert_url):
+        # Before we do anything, we need to verify that the URL for the
+        # signature matches what we expect.
+        cert_host = urllib.parse.urlparse(cert_url).netloc
+        if _signing_url_host_re.search(cert_host) is None:
```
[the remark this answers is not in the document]
dstufft: I'm torn! I generally always use ``.search()`` because ``.match()`` messes with my mental model of how regular expressions work. In this case, ``.match()`` is probably better because I don't know that it ever makes sense to have this *not* anchored at the beginning.

dstufft:
To switch to SES, you need a configuration like:

```
MAIL_BACKEND=warehouse.email.services.SESEmailSender sender=noreply@pypi.org region=us-west-2 topic=...
```

That assumes you have the other environment variables setup to configure authentication for AWS.

dstufft:
Note: After this merges and deploys, https://github.com/python/pypi-infra/pull/5 will need to be merged and applied to wire the SNS topic up to Warehouse's web hook handler. It will likely take two runs of terraform for this to work, since we need to know the expected TopicArn before we can subscribe to it.

warehouse/email/ses/tasks.py
```diff
@@ -0,0 +1,27 @@
...
+
+from warehouse import tasks
+from warehouse.email.ses.models import EmailMessage
+
+
+CLEANUP_AFTER = datetime.timedelta(days=14)
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def cleanup(request):
```
ewdurbin: I think it might behoove us to cleanup unsuccessful deliveries on a slower pace than successful deliveries. I'd suggest we keep 30-90days of failures for debugging as it's quite possible for someone to not notice that things have gone awry in the short term. Particularly for email notifications for things that they did not initiate.

The alternative to holding onto logs for longer period is to persist some kind of summary as to _why_ the email address was unverified, rather than just that it was.

Either option works for me, but the latter actually has the benefit of persisting for a much longer period than even 30-90 days which is nice.

When a user finally reaches out, we can at least tell them what the "last straw" for their previously verified address was.

warehouse/routes.py
```diff
@@ -209,6 +209,9 @@ def includeme(config):
     )
     config.add_route("packaging.file", files_url)
 
+    # SES Webhooks
+    config.add_route("ses.hook", "/_/ses-hook/", domain=warehouse)
```
ewdurbin: I've not used SES -> SNS -> webhook for this in the past, can we ensure we have the appropriate [DeliveryPolicies](https://docs.aws.amazon.com/sns/latest/dg/DeliveryPolicies.html) in place to give these webhooks a fighting chance?

warehouse/email/ses/views.py
```diff
@@ -0,0 +1,125 @@
...
+    return Response()
+
+
+@view_config(
+    route_name="ses.hook",
+    require_methods=["POST"],
+    require_csrf=False,
+    header="x-amz-sns-message-type:Notification",
+)
+def notification(request):
```
ewdurbin: This route needs to lean towards failing hard to make Retry logic simple with SNS webhooks.

SNS webhook retry conditions:
- HTTP status in the range 500-599.
- HTTP status outside the range 200-599.

Looks like we should be OK, as the explicit `HTTPBadRequest` responses are indeed for requests that are successfully ignored, but something to keep in mind.

warehouse/migrations/versions/7f0d1b5af8c7_add_tables_for_storing_ses_information.py
```diff
@@ -0,0 +1,154 @@
...
+        ),
+        sa.Column(
+            "status",
+            SESEmailStatuses,
+            nullable=False,
+            server_default="Accepted",
+        ),
+        sa.Column("message_id", sa.Text(), nullable=False),
+        sa.Column("from", sa.Text(), nullable=False),
+        sa.Column("to", sa.Text(), nullable=False),
```
ewdurbin: should we index this to help with admin search?

warehouse/email/ses/tasks.py
ewdurbin: To be clear, the boolean fields stored on accounts_email are probably enough to continue forward but may not capture all the necessary information to help a user get things back up and running. Also we may need more insight for certain spam greylisting/blacklisting situations.

warehouse/migrations/versions/7f0d1b5af8c7_add_tables_for_storing_ses_information.py
dstufft: I've added an index now!

warehouse/email/ses/tasks.py
dstufft: Set it up so failures are kept for 90 days, and success are kept for 14 days.

PR #3592, opened 2018-04-09
Changed anchor tag to button to enable tabbing on package detail & release page

### LINKED ISSUE #3576: Cannot tab to 'copy to clipboard'
On the project detail page, we have a link where the user can copy the pip install command to their clipboard:

screenshot from 2018-04-07 07-48-51

The code for this is:

```html
<a class="-js-copy-pip-command tooltipped tooltipped-s" data-clipboard-target="#pip-command" aria-label="Copy to clipboard" data-original-label="Copy to clipboard">
  <i class="fa fa-copy" aria-hidden="true"></i>
  <span class="sr-only">Copy PIP instructions<span>
</a>
```

We need to ensure that a user using a keyboard *only* can access this functionality.

Right now we are using a `<a>` with no `href` for this element.  When tabbing through the page using a keyboard, this "empty" `<a>` is *ignored*.  Instead we should use a `<button>` element, which will be recognised in the tabindex.

Making this change _may_ mean that we have to update the associated JavaScript.

---

Note: when making this change, we should also update the 'copy to clipboard' links on `/templates/manage/release.html`:

- https://github.com/pypa/warehouse/blob/master/warehouse/templates/manage/release.html#L152
- https://github.com/pypa/warehouse/blob/master/warehouse/templates/manage/release.html#L161
- https://github.com/pypa/warehouse/blob/master/warehouse/templates/manage/release.html#L170

waseem18:
Resolves #3576

### DIFF
--- a/warehouse/static/sass/blocks/_package-header.scss
+++ b/warehouse/static/sass/blocks/_package-header.scss
@@ -23,7 +23,7 @@
       <h1 class="package-header__name"></h1>
       <p class="package-header__pip-instructions">
         <span>// pip instructions</span>
-        <a><i></i></a> // copy icon
+        <button><i></i></button> // copy icon
       </p>
     </div>
     <div class="package-header__right">
@@ -37,7 +37,7 @@
   @include site-container;
   display: flex;
 
-  a {
+  button {
     color: $white;
     text-decoration: underline;
     text-decoration-color: $transparent-white;
@@ -104,7 +104,7 @@
       }
     }
 
-    a {
+    button {
       color: $white;
       line-height: $spacing-unit;
       float: left;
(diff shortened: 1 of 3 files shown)


warehouse/static/sass/blocks/_package-header.scss
```diff
@@ -114,6 +114,7 @@
       border-right: 1px dotted $transparent-white;
       border-top: 1px dotted $transparent-white;
       border-bottom: 1px dotted $transparent-white;
+      border-left: 1px;
```
waseem18: Had to add this as changing `a` to `button` was creating extra white border to left as shown below
clipboard

waseem18:
@nlhkabu Could you please review this? Thanks

nlhkabu:
Thank you @waseem18 :tada:

Issue #3436, opened 2018-03-27
Search API to search over keywords


mitar:
With planned deprecation of XML-RPC API it seems there will no way to search packages by their keywords. So I would like to do a feature request for this for the new API.

brainwane:
@mitar Thank you for bringing this up! We will only remove the XML-RPC API when its functionality is covered by other new APIs, so I have put this in a future milestone, and I've tagged it so people will see it when looking at API issues. Thanks again.

yuvalreches:
Hey @brainwane 
Looking at your roadmap and this thread, do I understand correctly that the old XML-RPC API will remain operational in the near future under pypi.org as well?

Currently when performing `pip search <packageName>` under pypi.python.org we get _HTTP 302_ and being redirected to pypi.python.org/pypi

However when performing the same (pip search) under pypi.org we get _HTTP error 404 while getting http://pypi.org/RPC2_

Isn't it supposed to be redirected to the old API for now?
Thanks

di:
@yuvalreches:

> Looking at your roadmap and this thread, do I understand correctly that the old XML-RPC API will remain operational in the near future under pypi.org as well?

Yes, the XML-RPC API will remain for now.

> Currently when performing `pip search <packageName>` under pypi.python.org we get HTTP 302 and being redirected to pypi.python.org/pypi

Correct, the 302 redirect from `https://pypi.python.org/` to `https://pypi.python.org/pypi` doesn't exist for `pypi.org`.

> However when performing the same (pip search) under pypi.org we get HTTP error 404 while getting http://pypi.org/RPC2

I'm not sure exactly what index URL you're using here, but you should be using `https://pypi.org/pypi`:
```
$ pip search foobar -vvv --index https://pypi.org/pypi
Starting new HTTPS connection (1): pypi.org
https://pypi.org:443 "POST /pypi HTTP/1.1" 200 272
foobar (1.1)         - This is the FooBar  project. (foo is taken at PyPI -
                       hahaha)
django-foobar (1.0)  - Super awesome portable module for Django
```

yuvalreches:
Hey @di 

I get your point, but why redirect to an endpoint that doesn't exist?

Would it be possible to have the same redirect as `https://pypi.python.org/` has for the time being (until XML-RPC v1 is deprecated)?

Meaning redirect from `https://pypi.org/` to `https://pypi.org/pypi` instead of `https://pypi.org/RPC2`

It would help us a lot

[...]

PR #3594, opened 2018-04-09
Add XML-RPC /RPC2 endpoint

di:
Per #3436, add a `/RPC2` endpoint for XML-RPC clients that will use this endpoint when the root does not support XML-RPC.

### DIFF
--- a/warehouse/legacy/api/xmlrpc/views.py
+++ b/warehouse/legacy/api/xmlrpc/views.py
@@ -30,22 +30,18 @@ from warehouse.packaging.models import (
     Role, Project, Release, File, JournalEntry, release_classifiers,
 )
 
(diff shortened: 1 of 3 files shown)


warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -28,15 +30,29 @@
     Role, Project, Release, File, JournalEntry, release_classifiers,
 )
 
-pypi_xmlrpc = functools.partial(
-    xmlrpc_method,
-    endpoint="pypi",
+_xmlrpc_method_partial = functools.partial(
```
dstufft: This might be cleaner by just eliminating the partial and inlining these options into the one method now.

PR #3602, opened 2018-04-10
Add 'switch to desktop' button

### LINKED ISSUE #3588: Add "Desktop version" link to mobile page
The mobile pypi.org page does not provide the same capabilities as the desktop page. For example, no GitHub statistic in the nav bar.
p
That's fine though. Sometimes, it may be useful to access the full information on the desktop page even when on mobile. For this, GitHub adds a "Desktop version" button on the bottom of each page.
dv
Perhaps that's something that warehouse could adopt.

nlhkabu:
Closes #3588 

Implements the HTML and CSS for this feature.  I will need someone else to actually hook up the JS as described here: https://github.com/pypa/warehouse/issues/3588#issuecomment-379986099

screenshot from 2018-04-10 07-53-36

### DIFF
--- a/warehouse/static/js/warehouse/controllers/viewport_toggle_controller.js
+++ b/warehouse/static/js/warehouse/controllers/viewport_toggle_controller.js
@@ -1,0 +1,55 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Controller } from "stimulus";
+
+// A constant value with represents a usual desktop viewport width
+const DESKTOP_WIDTH = 1280;
+
+export default class extends Controller {
+  static targets = ["switchToDesktop", "switchToMobile"];
+
+  connect() {
+    // Check if localStorage has already been set
+    const originalWidth = localStorage.getItem("originalWidth");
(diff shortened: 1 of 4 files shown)


pradyunsg:
I think a cleaner approach would be the click handler to just run:

```js
document.querySelector('meta[name="viewport"]').content = "width=1280, initial-scale=" + window.innerWidth / 1280;
```

Also, I think there should be some way to go back to the mobile view and for the desktop view to persist across pages. Sorry for the noise if already happens.

di:
> will this support the "Request Desktop Version" button common to many mobile browsers?

The way this works is just by resizing the viewport for that window (and sometimes changing the `User-Agent`). So `pypi.org` already supports this:

| screenshot_20180412-232105 |  screenshot_20180412-232057 |
| - | - |

...we just need to be able to do it manually when this button is pressed for browsers that don't already support it. 

> Also, I think there should be some way to go back to the mobile view and for the desktop view to persist across pages.

Agreed, we'll probably have to set something in `localStorage` to perisist this, which should also trigger a big "Switch to mobile version" banner when present.

di:
@nlhkabu I've hooked up the JS to handle this. I've also:
* made the button not visible if JS is disabled, since it would do nothing
* added a "Switch to mobile version" button which is shown only when the user has explicitly switched to the desktop version

I think the only thing left to do here is style the "Switch to mobile version" button so it's _much_ larger. Here's what Github's looks like for comparison:

github com_pypa_warehouse_pull_3602 nexus 5

nlhkabu:
Updated @di:
screenshot from 2018-04-17 07-27-27

I've tested this on browserstack and it seems to be working well.  Just need to test on my own mobile :)

di:
Still need to sort a small issue out here, don't merge.

Hultner:
@nlhkabu Try making the entire bar blue and clickable instead of a button inside a white container (like they do in the github version), it will look less crowded that way.

di:
@nlhkabu I think this is good to go on my end, could you test it one last time and merge if it looks good to you?

nlhkabu:
Looking and working fine with a sticky notification:

screenshot from 2018-04-18 06-39-05

PR #3605, opened 2018-04-10
extend XMLRPC cache TTL

ewdurbin:
1 hour was good during testing, moving this out to 24 hours will yield better cache coverage


warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -50,7 +50,7 @@ def decorator(f):
 xmlrpc_cache_by_project = functools.partial(
     xmlrpc_method,
     xmlrpc_cache=True,
-    xmlrpc_cache_expires=3600,
+    xmlrpc_cache_expires=86400,
```
di: This is a tremendous nit but would you mind doing the math here:

```python
xmlrpc_cache_expires=24 * 60 * 60,  # 24 Hours
```

Issue #3558, opened 2018-04-05
Keep tabs inside modal


nlhkabu:
## How to reproduce

- Log in
- Use your keyboard only
- Open a modal with your keyboard (either delete release, delete project, delete account)
- Tab through the modal

## What happens

- After you open the modal, you can tab through the content of the modal \o/
- However, once you have tabbed through each button/field, the tab index then moves to the content *under* the modal :(

## What should happen

Once a modal is opened, the user can only tab within the modal.  E.g. there is a tab "loop".

Souce:
https://www.marcozehe.de/2015/02/05/advanced-aria-tip-2-accessible-modal-dialogs/

> On the very last element of your dialog, add a special keyPress handler that traps the forward tab key to redirect it to the first element of the dialog. Otherwise, tab would jump outside the dialog into the surrounding document or to the browser UI, and we don’t want that.

> Likewise, add a keyPress handler to the very first focusable element inside the dialog that traps the shift+tab key combination so the backwards tab order also doesn’t leave the dialog.

di:
Fixed by #3619.

PR #3619, opened 2018-04-11
Keep tab cycling inside modal when modal is active

aalmazan:
Addresses #3558 with some ideas taken from https://stackoverflow.com/a/12444641/6553469.

Works well for me on both Chrome and Firefox. Have not had the chance to test IE11/Safari yet. I do wonder, though, if this should be within `confirm_controller.js` or in `static/warehouse/index.js`. 

Anyway, let me know if there are any issues -- any feedback appreciated.


aalmazan:
Interesting. This does work completely fine on my end even after rebasing upstream master and purging/rebuilding. Either way, indeed I glossed over that implementation detail of only catching the keypress on the first/last item and I'll update that.

> Also it looks like we can't tab to the "Cancel" button, but that might be a separate issue.

While I can tab to "Cancel" on my end, using spacebar doesn't work properly (since it's an `a` and not a `button` element). Don't know if that's also an issue worth looking into.

di:
Ah, realizing now this might be because of the way my Firefox browser is configured. I'll re-test this a bit more on Monday.

aalmazan:
@di Great to know it works on your end too! 

Also, I did some more fiddling around to bind the handlers directly on the first/second last/last elements, but think the approach above is the better one. Reason being: if we add handlers to the close button (the first item), SHIFT+TAB combinations when the button is selected will work, however, if you're on the text `input` and: 

1. (while on `input` field) hold SHIFT, press TAB
2. (on close button) release TAB
3. (still on close button) press TAB. This should be a SHIFT+TAB, but the SHIFT on the `input` field wasn't handled.

So it seems like we would have to add listeners to the entire modal or modal fields/buttons anyway?

di:
Yes I agree, I think the current approach is fine.

nlhkabu:
Thank you for this @aalmazan - this is working great for me :)

aalmazan:
Updated with the requested changes:

* Cancel now a button and its cancel action moved into `confirm_controller.js`
* Moved tabbing handling into the main `index.js` file

I also added e73cd74 since I noticed the ESC handler was firing when modal elements exist on the page, but aren't active (yet). This was causing a null exception with my changes.

PR #3620, opened 2018-04-11
Include Site Name in generated emails

dstufft:
Instead of having emails only come from a plain email address and without having any identifier what site they're for in the subject, we will instead inject the site name in both locations.

This ends up looking like:

```
From: PyPI <noreply@pypi.org>
Subject: [PyPI] PyPI email verification
```

We probably want to refactor the subjects of the emails we're sending currently since they all include the hardcoded string "PyPI" at the front of them (which is confusing when those emails come from TestPyPI instead!). We can probably just drop that string entirely from them and let the subjects focus on what is actually happening.

### DIFF
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -37,6 +37,7 @@ class TestSendEmail:
         task = pretend.stub()
         request = pretend.stub(
             find_service=pretend.call_recorder(lambda *a, **kw: sender),
+            registry=pretend.stub(settings={"site.name": "DevPyPI"}),
         )
 
         email.send_email(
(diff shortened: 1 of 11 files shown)


tests/unit/email/test_init.py
```diff
@@ -67,7 +72,10 @@ def retry(exc):
                 raise celery.exceptions.Retry
 
         sender, task = FakeMailSender(), Task()
-        request = pretend.stub(find_service=lambda *a, **kw: sender)
+        request = pretend.stub(
+            find_service=lambda *a, **kw: sender,
+            registry=pretend.stub(settings={"site.name": "DevPyPI"}),
```
pradyunsg: doubt: do we need this here?

di: We do, the subject is still going to be constructed before celery fails.

PR #3635, opened 2018-04-12
Ask user for regular (not normalized) project name when confirming deletion

### LINKED ISSUE #3491: Warehouse prompts for munged project name when deleting project with period character
Today, I went to delete the [jaraco.persona](https://pypi.org/project/jaraco.persona), but when I did, it did not accept the name of the project as confirmation:

image

Instead, it wanted a munged name with the period replaced by a dash.

image

Granted, it's just an annoyance, but given that it's prompting for the project to ensure the correct project is being deleted, it's quite surprising that the munged name is being demanded.

I'd like for warehouse not to munge names at all (I'm not sure what requirement drove the desire to munge names), but if it must, it should at least hide that detail from the user consistently.

saxenanurag:
Fixes #3491.


saxenanurag:
@brainwane Please review.

saxenanurag:
#3491

saxenanurag:
Will do.

warehouse/templates/manage/settings.html
```diff
@@ -51,7 +51,7 @@ <h3>Delete Project</h3>
       Deleting will irreversibly delete this project.
     {% endif %}
     </p>
-    {% set action = request.route_path('manage.project.delete_project', project_name=project.normalized_name) %}
-    {{ confirm_button("Delete project", "Project Name", project.normalized_name, action=action) }}
+    {% set action = request.route_path('manage.project.delete_project', project_name=project.name) %}
```
di: We actually don't need to change this line, we should use the normalized name for generating the path.

saxenanurag: @di Done

PR #3671, opened 2018-04-14
accept xmlrpc requests on /pypi/

ewdurbin:
seems devpi calls xmlrpc with the url /pypi/... 😕


ewdurbin:
@dstufft is there a better way to do this?

ewdurbin:
Happy to revisit if we can find a better solution, but need to roll with this for now.

PR #3677, opened 2018-04-15
Need to use pip < 10 until pip-tools supports pip 10

dstufft:


ewdurbin:
it is the shoemakers shoes going shoeless :-p

dstufft:
pip-tools isn't something I wrote!

ewdurbin:
all of the cool kids are using pip10, Donald.

alex:
I think this can be reverted now that pip-tools 2.0 is out.

PR #3681, opened 2018-04-15
Revert "Need to use pip < 10 until pip-tools supports pip 10"

alex:
Reverts pypa/warehouse#3677


dstufft:
huh, those commits are weird, diff looks fine though 🤷‍♂️

alex:
@dstufft yeah idk what's up with that. should be good to review though

PR #3687, opened 2018-04-15
Remove the temporary hack to purge legacy PyPI

dstufft:


ewdurbin:
@dstufft do we need to wait for legacy to be shut down? actually thinking about it... are these purges even working right now?

dstufft:
They aren’t working right now unless we’ve changed the legacy domain in the env vars to the new legacy url.

PR #3691, opened 2018-04-15
style nit in template

alex:


dstufft:
Errors should be fixed by https://github.com/pypa/warehouse/pull/3692.

PR #3693, opened 2018-04-15
Partial search updates

ewdurbin:
A fools attempt at #701 

Incrementally reindexes new documents or unindexes whenever a Project or Release is created, updated, or deleted.

### DIFF
--- a/warehouse/search/__init__.py
+++ b/warehouse/search/__init__.py
@@ -27,23 +27,40 @@ from warehouse.packaging.models import Project, Release
 def store_projects_for_project_reindex(config, session, flush_context):
     # We'll (ab)use the session.info dictionary to store a list of pending
     # Project updates to the session.
-    purges = session.info.setdefault("warehouse.search.project_updates", set())
+    projects_to_update = session.info.setdefault(
+        "warehouse.search.project_updates", set())
+    projects_to_delete = session.info.setdefault(
+        "warehouse.search.project_deletes", set())
 
     # Go through each new, changed, and deleted object and attempt to store
     # a Project to reindex for when the session has been committed.
-    for obj in (session.new | session.dirty | session.deleted):
+    for obj in (session.new | session.dirty):
         if obj.__class__ == Project:
-            purges.add(obj)
+            projects_to_update.add(obj)
         if obj.__class__ == Release:
-            purges.add(obj.project)
+            projects_to_update.add(obj.project)
+
+    for obj in (session.deleted):
+        if obj.__class__ == Project:
+            projects_to_delete.add(obj)
+        if obj.__class__ == Release:
+            projects_to_update.add(obj.project)
 
 
 @db.listens_for(db.Session, "after_commit")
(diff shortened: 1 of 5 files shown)


warehouse/search/__init__.py
```diff
@@ -18,7 +18,32 @@
...
+from warehouse import db
 from warehouse.search.utils import get_index
+from warehouse.packaging.models import Project, Release
+
+
+@db.listens_for(db.Session, "after_flush")
+def store_projects_for_project_reindex(config, session, flush_context):
+    # We'll (ab)use the session.info dictionary to store a list of pending
+    # Project updates to the session.
+    purges = session.info.setdefault("warehouse.search.project_updates", set())
```
di: Probably want to call this something other than `purges`... `reindexable`? `changed_projects`? `updated_projects`?

warehouse/search/tasks.py
```diff
@@ -87,6 +87,9 @@ def _project_docs(db):
         .order_by(Release.name)
     )
 
+    if project_name:
+        release_data = release_data.filter(Release.name == project_name)
```
di: Probably want to add this filter earlier on `releases_list` instead of here? Maybe it makes no difference...

warehouse/search/__init__.py
```diff
@@ -18,7 +18,32 @@
...
+
+    # Go through each new, changed, and deleted object and attempt to store
+    # a Project to reindex for when the session has been committed.
+    for obj in (session.new | session.dirty | session.deleted):
+        if obj.__class__ == Project:
+            purges.add(obj)
```
di: If the project has been deleted, I'm not sure it makes sense to add it to the set here, because the `.filter` below will not return any releases at all. (I guess a deleted project can just fall out of the index when we reindex nightly instead of trying to do something special for it here.)

However if a release has been deleted, it _does_ make sense.

warehouse/search/__init__.py
```diff
@@ -18,7 +18,32 @@
...
+        if obj.__class__ == Release:
+            purges.add(obj.project)
+
+
+@db.listens_for(db.Session, "after_commit")
+def execute_project_reindex(config, session):
+    purges = session.info.pop("warehouse.search.project_updates", set())
```
di: Ditto here.

warehouse/search/tasks.py
ewdurbin: _smart_

ewdurbin: it's damn fast either way, but you're right!

ewdurbin: so, since `releases_list` is a subquery, it seems it is not easily modified after the fact. probably best to let this sleeping dog lay.

di: You could just `.subquery` it later, and do `releases_list = releases_list.filter(...)` earlier. Again, might not really matter.

ewdurbin: easy enough.

Issue #3632, opened 2018-04-11
Get ~all users to verify email addresses


dstufft:
We have a problem with a bit of our data, namely that due to historical reasons we have a fair amount of users in the database that do not have a verified primary email address. The side effect of this is that we're currently sending emails to email addresses that we have not had verified. This is a bad situation to be in, because in order to keep our bounce/spam rate low, we should be confirming all email addresses before sending email to them. In addition the way our bounce handling code works is it un-verifies the email address, which the intent was to stop sending email to it until the user has reverified their email address.

In total there are about 193k user accounts with a unverified email address for their primary address, and 44k that do have a verified email address for their primary account.

So we need to come up with a strategy to resolve this, because it's pretty important that we don't send email to unverified addresses.

Here's what I've come up with, but I'd like to see what other people think as well.

For background, the way activation worked on legacy PyPI was that when you registered, it added a One time token (OTK) to a separate table that stored (username, OTK, datetime). When you verified your email with PyPI it would delete the entry from this other table, so effectively this table acts as a list of user accounts that legacy PyPI registered, but whom never activated their account via legacy PyPI.

So that means we have accounts in 3 possible states:

- They have a primary email address that is verified.
- They have a primary email address that is unverified, and they exist in the OTK table.
- They have a primary email address that is unverified, and they do not exist in the OTK table.

The first state is the happy state, and we currently have 44k accounts in that state. Looking at the OTK table, there are currently ~135k rows, if we assume that 100% of them are for accounts that did not end up verifying via Warehouse instead, that means that we have 135k accounts in the second state, and ~58k accounts in the third state. Just to correlate this, we also have ~135k users who are not in the is_active state.

Thus my plan of action is:

- [x] 1) Start displaying a flash-message like warning at the top of every page load for logged in users without a verified primary email address with a call to action to get a verified email address as their primary email address.
- [x] 2) Expand the limitations of not having a verified, primary address so that you cannot do much in the ways of project management without it. What exactly should be limited is on the table, but I think uploads in general should require a valid, verified email, and likely so should other actions like deletions, managing contributors, etc.
- [x] 3) Start a campaign of blogs, tweets, mailing list posts, etc to ask users to verify their email addresses with PyPI.
- [x] 4) Assume the ~135k are drive by accounts that have never been activated, and leave them marked unverified and inactive (if they haven't verified on Warehouse).
[...]

PR #3695, opened 2018-04-16
Unverified primary notification

di:
Towards #3632.


warehouse/templates/manage/manage_base.html
```diff
@@ -18,6 +18,20 @@
 {# Hide mobile search on manager pages #}
 {% block mobile_search %}{% endblock %}
 
+{% block extra_notifications %}
```
ewdurbin: should this be `session_notifications`?

di: Yup, good catch.

ewdurbin:
worked great locally with that change, 👍

Issue #3673, opened 2018-04-14
Open issues count on PyPI is open issues + open pull requests on GitHub


jrbourbeau:
The "Open issues" number currently quoted on PyPI for a project is the number of open issues plus the number of open pull requests for the project GitHub repo. 

For instance (at the time of submitting this issue) the dask project has 405 open issues on PyPI (https://pypi.org/project/dask/), while on the corresponding GitHub repo (https://github.com/dask/dask/) there are 364 open issues and 41 open pull requests. 

This may not be an error as I could be misinterpreting what "Open issues" should mean on PyPI. But my initial interpretation is that the number of open issues on PyPI should match the number of open issues on GitHub.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

di:
Github considers pull requests to be "issues" and doesn't provide a way in the API we're using to discern between the two without making multiple requests. From [here](https://developer.github.com/v3/issues/):

> As far as the API is concerned, every pull request is an issue, but not every issue is a pull request.

We might want to change "Open Issues" to something like "Issues/PRs" instead.

PR #3697, opened 2018-04-16
Correctly label open issues/PRs

glasnt:
Works around #3673

Opted to change the label vs changing the data collection type

The GitHub REST API doesn't nicely give the separation of Issues/PRs without having to do a deep count

There are ways in the GraphQL API, but that would require a complete change to how we collect data. A solution like https://github.com/isaacs/github/issues/536 would work for this particular datapoint. 

Before/After: 
after before 

Mockup of a project with a lot of issues
Mockup

I was worried that the length of the label would mean a limit to how much data we could display, but it looks like there's still plenty of space. 

While the labels no longer have a consistent length, it now displays accurate data.

([ding contributor 100?](https://twitter.com/EWDurbin/status/985686173879332865))

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -161,7 +161,7 @@
           </a>
           <a class="vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed github-repo-info__item" data-key="html_url" data-attr="href" data-supplement="/issues" rel="noopener" target="_blank">
             <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
-            <strong>Open issues: </strong>
+            <strong>Open Issues/PRs: </strong>
             <span class="github-repo-info__item" data-key="open_issues_count"></span>
           </a>
         </div>


warehouse/templates/packaging/detail.html
```diff
@@ -161,7 +161,7 @@ <h3 class="sidebar-section__title">GitHub statistics</h3>
           </a>
           <a class="vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed github-repo-info__item" data-key="html_url" data-attr="href" data-supplement="/issues" rel="noopener" target="_blank">
             <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
-            <strong>Open issues: </strong>
+            <strong>Open Issues/PRs: </strong>
```
di: This is a tremendous nitpick, but per our style guide (which is just the Material Design style guide), we'll want to use sentence-style capitalization here, so:

> Open issues/PRs

(https://material.io/guidelines/style/writing.html#writing-capitalization-punctuation)

glasnt: Do you prefer separate commits or squashes for such minor changes?

glasnt: It's a separate commit for you to do with as you choose. Screenshots updated to suit.

di: > Do you prefer separate commits or squashes for such minor changes?

Separate commits are preferred because it's easier to track what's been changed, and because we squash all PRs down to a single commit when merging anyways. Thanks for asking!

PR #3740, opened 2018-04-17
Update migration guide, fix legacy PyPI links

brainwane:


docs/api-reference/integration-guide.rst
```diff
@@ -18,6 +18,12 @@ Warehouse has now replaced `the legacy site pypi.python.org
 upload to pypi.python.org, it may continue to work due to redirects,
 but you should use pypi.org instead.
 
+A temporary installation of the legacy PyPI codebase is `available at
+legacy.pypi.org <http://legacy.pypi.org/>`_ until 30 April 2018, when
```
hugovk: Please use https here

brainwane: Thanks, fixing this plus some other non-HTTPS links.

PR #3755, opened 2018-04-18
add an admin only route for getting a tracemalloc

ewdurbin:
screen shot 2018-04-18 at 3 55 51 pm


ewdurbin:
trying to hunt down a steady memory consumer!

PR #3764, opened 2018-04-19
2nd go: add browser warning for IE 10

contrepoint:
* For [this issue](https://github.com/pypa/warehouse/issues/3517)
* Previous PR (merged and reverted) [here](https://github.com/pypa/warehouse/pull/3569)
* The warning should show for IE 10 and below
* Thanks in advance for your help testing, @nlhkabu!

fairly sure @di's comment [here](https://github.com/pypa/warehouse/pull/3569/files#r181210563) is right. it was because the `unsupportedBrowser` function was not being called - `unsupportedBrowser` rather than `unsupportedBrowser()` so it would always return something truthy, which is why the warning shows up for non IE-10 browsers.


nlhkabu:
Thanks for the update @contrepoint.

Tested with BrowserStack:

### Banner is showing as expected

- ie10

### JS is working, banner is not showing

- Mac high sierra - Chome 65
- Mac high sierra - Safari 11
- Mac high sierra - Opera 52
- Ubuntu 16.04 - Chrome 63
- Ubuntu 16.04 - Firefox 59
- Windows 10  - Edge 16
- Windows 10  - IE 11

I also checked on a range of phones (iphone, windows phone, android) via browserstack.

@di can you please confirm all is ok on your side? Thx

PR #3789, opened 2018-04-23
Customise confirmation message when verifying the user's primary email address

browniebroke:
Congratulations on last week migration, it's amazing to see the new PyPI website! 🎉 

I just verified my email address from my existing account and the confirmation message prompted a small confusing moment when it was saying _"You can now set this email as your primary address"_  although it was already my primary (I only had one email address). 

I would suggest tweaking a bit the message in that case, so here is my tiny contribution to make this happen.

PS: The new message is only my own suggestion, and english isn't my first language. Suggestions for improvements are welcome!

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -411,11 +411,15 @@ def verify_email(request):
     email.unverify_reason = None
     email.transient_bounces = 0
 
+    if not email.primary:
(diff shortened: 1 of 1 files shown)


brainwane:
Thank you very much for the contribution and the kind words! We'll review this soon!

PR #3820, opened 2018-04-25
Adding description_content_type to json API

### LINKED ISSUE #3798: Expose long_description_content_type in JSON API
Feature request: Add  'long_description_content_type' to JSON API in 'info': section, alongside the existing 'description'

Now that PyPI supports Markdown and more than one kind of content type for long descriptions, it would be helpful to be able to see what the content type is with the JSON API. Otherwise it is more difficult to correctly interpret the 'description', since it isn't immediately obvious if it's reStructuredText or Markdown with which variant.

I'm willing to try this if I'm pointed in the right general direction :) If someone else can get it done trivially that's fine too.

nixjdm:
Fixes #3798

I have tested this locally, and extended the existing unit tests for the JSON and XMLRPC APIs. I also extended the API documentation to show these changes.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-trove-classifier.md
+++ b/.github/ISSUE_TEMPLATE/new-trove-classifier.md
@@ -1,11 +1,14 @@
 Request to add a new Trove classifier.
 
 ## The name of the classifier you would like to add
-<!-- Classifier name -->
+
+<!-- Replace the following with the name of your classifier -->
+* `Your :: Trove :: Classifier`
 
 ## Why do you want to add this classifier?
 <!--
     Please include a brief explanation to justify your request.
     Why do the current classifiers not meet your need?
     How many projects do you expect to use this new classifier?
+    If you are requesting multiple classifiers, why do you need more than one?
 -->
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -48,6 +48,7 @@ Project
                     "Topic :: Software Development :: Build Tools"
                 ],
                 "description": "...",
+                "description_content_type": null,
                 "docs_url": null,
                 "download_url": "",
                 "downloads": {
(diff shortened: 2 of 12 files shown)


ewdurbin:
@nixjdm thank you for your contribution!

The XMLRPC endpoints are planned for deprecation. As such, we generally do not add features to them except to improve compatibility with the previous implementation.

You may choose to either split this into two Pull Requests to be considered/discussed individually (one for XMLRPC and one for JSON) or remove the XMLRPC change altogether.

nixjdm:
10-4 @ewdurbin. I have removed the xmlrpc changes from the branch / PR.

di:
Sorry for the mixed signals about the XML-RPC stuff, @nixjdm. I'll defer to @ewdurbin here.

I'm a little surprised to see that the test for this endpoint just has a bunch of `None`s for the fields... I agree that it'd be nice to see this value set on the release and assert that the same value is in the result from the view.

nixjdm:
I was too. I'd be happy to take a crack at expanding the tests to be more
useful for this and other fields in the json api tests. As you said, it
isn't just this field that's testing what is basically the trivial case.

If you make a follow-up ticket, you can assign it to me.

If it's a blocker for this PR, I suppose I should keep the scope narrow,
and expand the tests further for the rest of the json api in a follow-up.

di:
> If it's a blocker for this PR, I suppose I should keep the scope narrow, and expand the tests further for the rest of the json api in a follow-up.

Yep, I think this is probably ideal.

nixjdm:
@di I updated the tests. A quick explanation of them: I tried to add a non-trivial test case for `description_content_type`, leave most everything else untouched, and prep the other values for easily modifying their test values later for the `test_detail_renders`.

There was previously 3 releases for to the test project, with the 2nd being the most detailed. I felt adding to the complexity of the second release meant awkward looking code, so I instead made the final release the detailed one. I made the first the empty release. So the order of the releases tested is switched around (in a way I think is more intuitive), but all of the other tests except the one with a `description_content_type` are otherwise unchanged. I also kept testing of an `info` section with mainly empty field values. To accomplish that I made it a separate dedicated test. The diff here in github looks a bit weird in regards to the `url` sections of the two defs. The second def is entirely new, and so the first `urls` section should be the one that looks partially changed.

How does this look?

Issue #3819, opened 2018-04-25
pip search not working if pkg name contains '-'


Ayub-Khan:
* Pip version:
9.0.1 and 10.0.1
* Python version:
python 2 and python 3
* Operating system:
Ubuntu

### Description:

pip search edx-i18n-tools
Expected Behavior : 
It should show the version and summery of latest version
Actual Behavior : 
shows Nothing

pip search edx-ace
Expected Behavior : 
It should show the version and summery of latest version
Actual Behavior : 
shows Nothing

pip search edxval
Expected Behavior : 
It should show the version and summery of latest version
Actual Behavior : 
shows correctly

###Another issue
search also not works if the word is not in lower case in pkg description
ie.
pip search internationalization
It should show all the libraries which have Internationalization/internationalization in their pkg description
only shows one with name internationalization. 

###Another issue
Search also does not work if query string is not in lower case
ie. 
pip search edx
works shows a list

pip search edX
shows nothing
it should show the same list as 'pip search edx'
pip search EDX
it should show the same list as 'pip search edx'

https://github.com/pypa/pip/issues/5332

ewdurbin:
I think the results for all provided examples make more sense in the "main" search.

Looks like we should be able to make results for `pip search` more sensible by applying the same search logic/boosting. I'm gonna approach this shortly.

ewdurbin:
all reported issue here were resolved in #3827

Ayub-Khan:
@ewdurbin thanks a lot for fixing the issue. I ran the following command to verify the behavior 
pip search edx-i18n
pip search django-filter
It is working fine but it seems it is showing all results for 'edx' combined with all results of 'i18n'. it is treating it as two different strings django and filter and showing the union of result sets for both. it should only show the results if whole query string is in name or summry right ?

ewdurbin:
@Ayub-Khan I believe your theory is correct, though I'm not sure how to improve that.

Issue #3717, opened 2018-04-16
Behavior Difference in Search


dstufft:
The previous search let you OR multiple things together to get the results for both, you could see this in the behavior of ``pip search pip initools``, however the new search is not returning both. We need to decide if this is now expected behavior, or fix it.

ewdurbin:
@dstufft I believe this is resolved in #3827 can you confirm?

wimglenn:
This seems resolved in warehouse (using xmlrpc client directly to search for "pip initools" is working ok).  But still would need to be addressed in pip cli.  

    pip search pip initools  # crash, unhandled error in xml rpc call time
    pip search "pip initools"  # does not return results

jamadden:
This certainly seems fixed to me:

```console
$ pip --version
pip 18.1 from //lib/python3.7/site-packages/pip (python 3.7)
$ pip search pip | grep -i initools
$ pip search initools | grep -i initools
INITools (0.3.1)  - Tools for parsing and using INI-style files
$ pip search pip initools | grep -i initools
INITools (0.3.1)                     - Tools for parsing and using INI-style files
```

PR #3827, opened 2018-04-26
xmlprc search improvements

ewdurbin:
harmonizes xmlrpc search to our main search. this dramatically improves it's value.

### DIFF
--- a/warehouse/legacy/api/xmlrpc/views.py
+++ b/warehouse/legacy/api/xmlrpc/views.py
@@ -30,6 +30,7 @@ from warehouse.classifiers.models import Classifier
 from warehouse.packaging.models import (
     Role, Project, Release, File, JournalEntry, release_classifiers,
 )
+from warehouse.search.queries import SEARCH_BOOSTS
 
 
 _MAX_MULTICALLS = 20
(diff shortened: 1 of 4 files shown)


ewdurbin:
addresses #3819 #3717

warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -107,18 +108,21 @@ def search(request, spec, operator="and"):
...
+                q |= Q("match", **{field: kw})
         queries.append(q)
 
     if operator == "and":
         query = request.es.query("bool", must=queries)
     else:
         query = request.es.query("bool", should=queries)
 
-    results = query[:1000].execute()
+    results = query[:100].execute()
```
di: 👍

wimglenn: This could do with some documentation. Expected results are mysteriously missing from "pip search" if you are searching a common term, and it's not clear why unless digging through the warehouse source code.

ewdurbin: https://github.com/pypa/warehouse/pull/4281

PR #3829, opened 2018-04-26
Add data to mobile view

### LINKED ISSUE #3546: Homepage link missing on mobile
1. Open https://pypi.org/project/cmarkgfm/ on mobile (or a desktop, with a sufficiently narrow window, e.g. 683px)
2. Try to find the Homepage project link
3. It's not there!

nlhkabu:
Closes #3546

Behaviour:

- 800px and under, adds a 'data' item into the mobile tabs menu.
- 801px and above, the data tab is removed, and the data is displayed in the sidebar

This PR only adds the HTML needed to make this change.  We also need to adjust the JS so that:

1. The tabs work on desktop (currently broken)
2. Users who are on the 'data' tab when <800px, *then* resize their browser to be > 800px, are "switched" to the project description tab

screenshot from 2018-04-26 08-31-34

@yeraydiazdiaz would this be something you might be able to help me with?

### DIFF
--- a/warehouse/templates/includes/packaging/project-data.html
+++ b/warehouse/templates/includes/packaging/project-data.html
@@ -1,0 +1,107 @@
+{% if release.urls.values() | contains_valid_uris %}
+<div class="sidebar-section">
+  <h3 class="sidebar-section__title">Project links</h3>
+  {% for name, url in release.urls.items() %}
+  {% if is_valid_uri(url) %}
+  <a class="vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed" href="{{ url }}" rel="nofollow">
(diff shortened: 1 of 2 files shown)


yeraydiazdiaz:
@nlhkabu sure, I can have a look 👍

yeraydiazdiaz:
@nlhkabu the logic was getting a bit awkward to handle on vanilla JS so I ported it to Stimulus, here's a capture of the result:

detail_tabs

@di could you review the JS code, please?

nlhkabu:
Thanks @yeraydiazdiaz - it looks great to me!

tomfbiz:
I tested this in chrome using _Dev Tools > Device Mode_ to emulate mobile browsers, and it works as expected.

PR #3837, opened 2018-04-27
Updated profile pages to show username if not viewing own profile

### LINKED ISSUE #3591: Change 'statistics' text based on whether it is the logged in user
Right now, we have the following text on the user profile page:

screenshot from 2018-04-09 12-49-03

This text shows whether the user being viewed is logged in or not.

What would be better:

1. *If* the logged in user is viewing their *own* public profile, show:

```html
View statistics for your projects via Libraries.io, or by using Google BigQuery.
```

2. *Else* (the user is not logged in, or is viewing the profile of _another_ user), show:

```
View statistics for <username>'s projects via Libraries.io, or by using Google BigQuery.
```
---

**Screenshot Required**: This issue will require an update to the visual design of the site. To help our team give you faster feedback, please include a screenshot in your Pull Request.

coreygirard:
Closes #3591

### When signed in and viewing own profile
screen shot 2018-04-26 at 8 20 28 pm

---

### When signed out or viewing someone else's profile
screen shot 2018-04-26 at 8 20 50 pm

### DIFF
--- a/warehouse/templates/includes/accounts/edit-profile-button.html
+++ b/warehouse/templates/includes/accounts/edit-profile-button.html
@@ -14,4 +14,16 @@
 
 {% if request.user == user %}
   <a href="{{ request.route_path('manage.account') }}" class="button button--primary author-profile__edit-button">Edit profile</a>
+
+  <div class="sidebar-section">
+    <div style="height:20px"></div>
+    <h3 class="sidebar-section__title">Statistics</h3>
+    <p>View statistics for your projects via <a href="https://libraries.io/">Libraries.io</a>, or by using <a href="https://packaging.python.org/guides/analyzing-pypi-package-downloads/">Google BigQuery</a></p>
+  </div>
+{% else %}
+<div class="sidebar-section">
+  <div style="height:20px"></div>
+  <h3 class="sidebar-section__title">Statistics</h3>
+  <p>View statistics for {{ user.username }}'s projects via <a href="https://libraries.io/">Libraries.io</a>, or by using <a href="https://packaging.python.org/guides/analyzing-pypi-package-downloads/">Google BigQuery</a></p>
+</div>
 {% endif %}
--- a/warehouse/templates/includes/accounts/edit-profile-button.html
+++ b/warehouse/templates/includes/accounts/edit-profile-button.html
@@ -16,13 +16,13 @@
   <a href="{{ request.route_path('manage.account') }}" class="button button--primary author-profile__edit-button">Edit profile</a>
 
   <div class="sidebar-section">
-    <div style="height:20px"></div>
+    <div class="sidebar-section__spacer"></div>
     <h3 class="sidebar-section__title">Statistics</h3>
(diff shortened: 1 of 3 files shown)


warehouse/templates/includes/accounts/edit-profile-button.html
```diff
@@ -14,4 +14,16 @@
...
   <a href="{{ request.route_path('manage.account') }}" class="button button--primary author-profile__edit-button">Edit profile</a>
+
+  <div class="sidebar-section">
+    <div style="height:20px"></div>
+    <h3 class="sidebar-section__title">Statistics</h3>
+    <p>View statistics for your projects via <a href="https://libraries.io/">Libraries.io</a>, or by using <a href="https://packaging.python.org/guides/analyzing-pypi-package-downloads/">Google BigQuery</a></p>
+  </div>
+{% else %}
+<div class="sidebar-section">
+  <div style="height:20px"></div>
```
ewdurbin: The lint violation is due to the fact that our Content Security Policies do not allow inline style. You'll need to update the SCSS in https://github.com/pypa/warehouse/blob/053c890c8b305cb5ca03de718866ba600915ab40/warehouse/static/sass/blocks/_sidebar-section.scss#L27 or find an appropriate class to apply.

ewdurbin: Also note that the linter can be run locally with `make lint`

brainwane:
Thanks for the quick update! Corey did this at an NYC Python sprint night tonight - glad to meet you!

coreygirard:
My pleasure! And likewise!

warehouse/templates/includes/accounts/edit-profile-button.html
```diff
@@ -14,4 +14,16 @@
 
 {% if request.user == user %}
   <a href="{{ request.route_path('manage.account') }}" class="button button--primary author-profile__edit-button">Edit profile</a>
+
+  <div class="sidebar-section">
+    <div class="sidebar-section__spacer"></div>
```
nlhkabu: Hi @coreygirard, thanks for this PR !

I think we can avoid adding an empty div here, instead we can add a modifier to `sidebar-section`.

e.g. `sidebar-section sidebar-section--spaced`. 

Then we can add a `margin-top: 20px` style in the css. 

Typically, we add modifiers at the bottom of the codeblock like this `&--spaced`, and add a small comment in the documentation at the top of the page.  In this case, there seems to also be some out of date comments there too.

The button group CSS file is a good place to look for a good example :)

Let me know if this isn't clear, or if you'd like me to make these changes for you :)

coreygirard: So I had originally tried to do basically exactly that, but for some reason, it didn't end up showing the space. I just tried again, with no luck. But my CSS skills are quite limited, so if you wouldn't mind making the edit, that would be much appreciated :)

nlhkabu:
Hi @coreygirard , I've made an updated PR [here](https://github.com/pypa/warehouse/pull/3852) that includes your commits, so I'm going to close this in favour of that PR :)

PR #3848, opened 2018-04-28
WIP: cache base image to speed up frontend tests

yeraydiazdiaz:
Branched from #3450 Travis has a restriction whereby it will remove environment variables from PRs from forked repositories.

Testing a base image creation and push to try and speed up end to end tests [as commented in #3450](https://github.com/pypa/warehouse/pull/3450#issuecomment-385181779)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ warehouse/static/dist
 warehouse/admin/static/dist
 
 tags
+selenium-report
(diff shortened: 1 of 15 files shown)


yeraydiazdiaz:
Giving up on this attempt, just realized that even if it did speed up the build Travis's restriction would prevent any contributors PRs from running frontend tests anyway. 😞

PR #3856, opened 2018-04-30
implement legaacy files and display actions

### LINKED ISSUE #3845: Redirect files &/or display action URLs from legacy
I see some users using legacy URLs like:

* https://pypi.python.org/pypi?:action=files&name=limix&version=1.0.14 (on legacy: https://legacy.pypi.org/pypi?:action=files&name=limix&version=1.0.14 )
* https://pypi.python.org/pypi?:action=display&name=craft-ai (on legacy: https://legacy.pypi.org/pypi?:action=display&name=craft-ai )
* https://pypi.python.org/pypi?:action=display&name=text_progress_bar (on legacy: https://legacy.pypi.org/pypi?:action=display&name=text_progress_bar )

It'd be nice to redirect the display URLs and, potentially, the files URLs to Warehouse.

Followup to #3799.

ewdurbin:
resolves #3845

### DIFF
--- a/tests/unit/legacy/api/test_pypi.py
+++ b/tests/unit/legacy/api/test_pypi.py
@@ -131,3 +131,105 @@ class TestBrowse:
 
         with pytest.raises(HTTPNotFound):
             pypi.browse(db_request)
+
+
+class TestFiles:
+
+    def test_files(self, db_request):
+        name = "pip"
+        version = "10.0.0"
+
+        db_request.params = {"name": name, "version": version}
+        db_request.route_path = pretend.call_recorder(
+            lambda *a, **kw: f'/project/{name}/{version}/#files'
+        )
+
+        result = pypi.files(db_request)
+
+        assert isinstance(result, HTTPMovedPermanently)
+        assert result.headers['Location'] == (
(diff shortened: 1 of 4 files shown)


di:
Isn't the `files` action just for logged-in users? E.g. when I visit https://legacy.pypi.org/pypi?:action=files&name=pip&version=9.0.1, it doesn't actually show me a list of files for that release.
The equivalent redirect would be to `/manage/project/pip/release/9.0.1/` instead.

However, if people are making links to this action on legacy, they probably actually want the public file list and not the logged-in view, so maybe we really do want to change the behavior?

ewdurbin:
@di heh, you're right! @brainwane can you detail where you found the `:action=files` links?

ewdurbin:
oh hey @di my bad. I broke this in https://github.com/pypa/pypi-legacy/commit/c62f721df5b14426cb06141599063cf4ede73bfd#diff-49574d26fae10c594eae7f8ba60452e4

the behavior before was a list!

brainwane:
Belatedly: the `:action=files` links were in GitHub READMEs and comments.

PR #3865, opened 2018-04-30
Remove twisted and dependencies

di:
Closes #3855.


requirements/deploy.txt
```diff
@@ -65,60 +22,3 @@ cryptography==2.2.2 \
...
-    --hash=sha256:21506674d30c009271fe68a242d330c83b1b9d76d62d03d87e1e9528c61beea6 \
-    --hash=sha256:f41037260deaacb875db250021fe883bf536bf6414a4fd25b25059b02e31b120 \
-    --hash=sha256:9ada83f4384bbb12dedc152bcdd46a3ac9f5f7720d43ac3ce3e8e8b91d733c10 \
-    --hash=sha256:e881ef610ff48aece2f4ee2af03d2db1a146dc7c705561bd6089b2356f61641f \
-    --hash=sha256:e6b77ae84f2b8502d99a7855fa33334a1eb6159de45626905cb3e454c023f339 \
-    --hash=sha256:a1daf9c5120f3cc6f2b5fef8e1d2a3fb7bbbb20ed4bfdc25bc8364bc62dcf54b \
-    --hash=sha256:57c38470d9f57e37afb460c399eb254e7193ac7fb8042bd09bdc001981a9c74c
-
-# The following packages are considered to be unsafe in a requirements file:
-setuptools==39.1.0 \
```
ewdurbin: I think we need to keep this.

ewdurbin: indeed, looks like this is what is catching you up in CI

di: I don't think we need to keep this? What hung up the CI was that I forgot to put `cryptography` as a dependency in `main.in`.

PR #3895, opened 2018-05-04
Access issues template

ewdurbin:
Followup from #3885, If users are having network access issues, it is critically helpful to have any/all of this up front.


.github/ISSUE_TEMPLATE/access-issues.md
```diff
@@ -0,0 +1,166 @@
...
+    issues for PyPI.
+-->
+
+<!--
+### Fastly Debug
+    Note that this output is optional, as it encodes identifying information
+    about your browser, public IP, and location.
+
+    We'll remove this output from the issue before closing, but do not feel
+    comfortable making it requried for all users.
```
jamadden: requried -> required

.github/ISSUE_TEMPLATE/access-issues.md
```diff
@@ -0,0 +1,166 @@
...
+
+    If you have a browser that reproduces your access issue, go to
+    https://fastly-debug.com, when the page has loaded copy and paste the
+    base64 encoded blob at the top of the page into the box below, and
+    uncomment this section.
```
jamadden: When I was following along with the instructions on the issue yesterday, I found that I had to disable all my browser's ad-blockers and privacy protectors to get the scripts there to run. I don't know if that's worth mentioning.

.github/ISSUE_TEMPLATE/access-issues.md
```diff
@@ -0,0 +1,166 @@
...
+### DNS Resolution
+
+<!--
+    Provide the servers you're querying for DNS here if you know. They can be
+    found in /etc/resolv.conf on most linux servers, or your network settings
+    for MacOS and Windows.
+-->
+
+```shell
+dig pypi.org A
```
jamadden: Is it worth mentioning what dig is and how to get it (like is done for traceroute below), or is that pretty standard on modern *ix-like OSs? (It is installed by default on macOS 10.13 as is `host`.)

ewdurbin:
@jamadden thank you for the swift review, I've touched on all of the notes.

.github/ISSUE_TEMPLATE/access-issues.md
```diff
@@ -0,0 +1,173 @@
...
+curl -vvv -I --ipv6 https://files.pythonhosted.org/packages/ae/e8/2340d46ecadb1692a1e455f13f75e596d4eab3d11a57446f08259dee8f02/pip-10.0.1.tar.gz
+<Replace with your output>
+```
+
+### TLS Debug
+
+#### IPv4
+
+```shell
+openssl s_client -4 -connect pypi.org:443
```
di: At least for me, these commands will hang waiting for a command to be input, which might be confusing. Something like `echo -n | openssl s_client -4 -connect pypi.org:443` might be better.

ewdurbin: good call.

PR #3907, opened 2018-05-05
Add Dependabot to list of release monitoring services

greysteil:
Hi team,

I noticed libraries.io and pyup.io listed as ways to receive notifications of new releases, and wondered if you'd be happy to list [Dependabot](https://dependabot.com), too. It offers Pipfile and requirements.txt support (although it's better at Pipfile), is free for open source and personal accounts, and we try to be good open source citizens by contributing back whenever we can. (We've helped fixed a few bugs in Pipenv recently, and will continue doing so whenever we encounter any.)

### DIFF
--- a/warehouse/templates/pages/help.html
+++ b/warehouse/templates/pages/help.html
@@ -171,7 +171,7 @@
         </p>
         <h3 id="project-release-notifications">{{ project_release_notifications() }}</h3>
         <p>
-          PyPI itself does not offer a way to get notified when a project uploads new releases. However, there are several third-party services that offer comprehensive monitoring and notifications for project releases and vulnerabilities such as <a href="https://pyup.io" target="_blank" rel="noopener">pyup.io</a> and <a href="https://libraries.io/" target="_blank" rel="noopener">libraries.io</a>.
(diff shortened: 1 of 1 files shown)


greysteil:
Of course - have a fantastic time!

di:
Hi @greysteil, thanks for this. I agree that these types of projects should all be listed somewhere but I'm not sure that this PyPI help page should be the place where it should be maintained. Do you have any ideas about what would be a better place for this to live?

I think I would prefer if this FAQ were less opinionated (i.e., no service was explicitly named), and instead linked elsewhere to a list of 3rd party services that perform this function.

(see also #3915).

greysteil:
Yeah, I agree with that - seems odd for you to be curating this list.

The only place with an organisation curating this that I'm aware of is the dependency management section of the GitHub Marketplace, [filtered for Python](https://github.com/marketplace/category/dependency-management?utf8=%E2%9C%93&query=python). The downside is that that doesn't (yet) have pyup.io on it, but the upside is it would shift these requests to GitHub and away from PyPI.

**Update:** Even better might be [this link](https://github.com/works-with/category/dependency-management?utf8=%E2%9C%93&query=python), which has all the apps that work with GitHub on it (a lower bar than getting into the marketplace) so includes pyup 🎉

PR #3913, opened 2018-05-06
Stimulus unit tests

yeraydiazdiaz:
Related to #3450 I found through the [Stimulus discourse](https://discourse.stimulusjs.org/t/testing-stimulus/90) a repo that implements Stimulus unit tests. Here's a simple proof of concept testing a single Stimulus controller.

I'm having some issues getting them to run on containers ~and it's likely I'll have to tweak the Travis conf to get them to run properly~, regardless I think it's a good alternative/complement to full end to end tests.

If this is deemed interesting we should probably introduce documentation on the matter, as there's a few pitfalls in the way tests are set up.

/cc @di


yeraydiazdiaz:
Travis seems to have our back, I'm sorry I doubted you 😄

I found a way to run these tests in the `static` docker container though it requires installing Firefox and Xvfb which takes some time. Seems to me it's easier to simply use the dev's environment rather than mimiking the way `make tests` runs. Let me know if there's a strong reason to run them in the container.

yeraydiazdiaz:
I decided to go with running the tests in the `static` container for consistency since the extra package installation will be cached anyway and you can always manually run `npm test` if needed.

I did however modified the `Dockerfile.static` separate from the main `Dockerfile` to prevent installing the new test-only dependencies.

yeraydiazdiaz:
I believe this is now ready to merge.

I've added tests for all Stimulus controllers and while doing so I found some behaviour inconsistent to the description in the notification controller. [I changed](https://github.com/pypa/warehouse/pull/3913/files#diff-96ccc3f8e782e14808361d2b283f390bR44) and tested it but @di, could you make sure the logic still holds, please?

yeraydiazdiaz:
Rebased and added linting and licenses.

warehouse/static/js/warehouse/controllers/notification_controller.js
```diff
@@ -41,7 +41,7 @@ export default class extends Controller {
     // Check if the target is dismissable, and if so:
     // * whether it has no notificationId (it's ephemeral)
     // * or it's not in localStorage (it hasn't been dismissed yet)
-    if (isDismissable && (!notificationId || !localStorage.getItem(notificationId))) {
+    if (!notificationId || (isDismissable && !localStorage.getItem(notificationId))) {
```
di: I _think_ this was right as-is? But it's been a while since I've taken a look at this, could you go into a bit more detail why you think this should be changed?

Also, if it does need changing, we should update the comment as well.

Dockerfile.static
```diff
@@ -8,12 +8,12 @@ WORKDIR /opt/warehouse/src/
...
     && apt-get update \
     && apt-get install --no-install-recommends -y \
-        libjpeg-dev nasm
+        libjpeg-dev xvfb firefox-esr nasm
 
 # However, we do want to trigger a reinstall of our node.js dependencies anytime
 # our package.json changes, so we'll ensure that we're copying that into our
 # static container prior to actually installing the npm dependencies.
-COPY package.json package-lock.json .babelrc /opt/warehouse/src/
+COPY package.json package-lock.json /opt/warehouse/src/
```
di: Why are we removing `.babelrc` here?

di:
This is really tremendous work @yeraydiazdiaz! Just some small comments but overall I think this is great.

Dockerfile.static
yeraydiazdiaz: I don't think that was intentional, though it's been a while 😅

warehouse/static/js/warehouse/controllers/notification_controller.js
yeraydiazdiaz: the reason is that ephemeral notifications should always be visible, regardless of the dismissability of the notifications.

The logic I'm trying to follow, and is not fully covered by code and tests atm is to show the notification if:

- the notification is ephemeral, i.e. it has no notification ID
- it's not ephemeral and is not dismissable
- it's not ephemeral, is dismissable and the user has not dismissed it yet

If that makes sense I can add the necessary test cases to cover all those scenarios

di: OK, that makes sense, and I can see the edge case that this change is covering. I don't think additional tests are needed but it would be good to update the comment.

nlhkabu:
Hi @yeraydiazdiaz thank you for his PR :grin: 

Would it be better to include the new tests in `make tests`?  Maybe we could [use arguments](https://stackoverflow.com/questions/15500204/makefile-passing-command-line-arguments-to-file-inside-makefile), e.g.

`make tests` - runs all tests
`make tests TYPE="front"` - front only
`make tests TYPE="back"` back only

My concern is that I'm going to forget to run them :( 

@di what do you think?

Either way, I think we should document this [in the docs](https://github.com/pypa/warehouse/blob/master/docs/development/getting-started.rst#running-tests-and-linters)

yeraydiazdiaz:
Hey @nlhkabu, I've created a `tests_backend` recipe and had `tests` execute it first and, if successful, execute `tests_frontend`. Does that sound reasonable? Hopefully it won't annoy people that are used to just typing `make tests` for backend tests.

.travis.yml
```diff
@@ -39,6 +39,7 @@ script:
   - node -v
   - npm -v
   - bin/tests --postgresql-host 127.0.0.1
+  - npm test
```
di: This should probably be `bin/tests_frontend` instead, for consistency.

Makefile
```diff
@@ -95,11 +95,18 @@ serve: .state/docker-build
 debug: .state/docker-build
 	docker-compose run --rm --service-ports web
 
-tests:
+tests_backend:
 	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
 								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
 								  bin/tests --postgresql-host db $(T) $(TESTARGS)
 
+tests_frontend:
```
di: Perhaps these would be better as `web_tests` and `static_tests`? Since the differentiation is which container they're being run in.

di:
@nlhkabu I think your concerns have been addressed, can you re-review?

nlhkabu:
Thanks for the updates @yeraydiazdiaz :)

Can you please document the new make commands in the documentation? The relevant file is here: https://github.com/pypa/warehouse/blob/master/docs/development/getting-started.rst#running-tests-and-linters

nlhkabu:
Thanks for the documentation @yeraydiazdiaz !

di:
@yeraydiazdiaz If you can bring this up to date, I'd like to merge it.

yeraydiazdiaz:
@di sure, however I found [a simpler way of testing Stimulus controllers](https://shime.sh/testing-stimulus) so it may take some time.

yeraydiazdiaz:
Closed in favor of https://github.com/pypa/warehouse/pull/6617

PR #3928, opened 2018-05-08
redirect /pypi to /

### LINKED ISSUE #3914: Missing legacy redirection from pypi.python.org/pypi/
**Describe the bug**
Redirections from `https://pypi.python.org/pypi/` are not handled (only redirected to `https://pypi.org/pypi/` by varnish (fastly)).

As https://pypi.org/pypi/ does not exists, it creates some broken links.

**Expected behavior**
A 301 to `https://pypi.org/`, simply.

**To Reproduce**
```$ curl -sI https://pypi.python.org/pypi/
HTTP/2 301 
server: Varnish
retry-after: 0
location: https://pypi.org/pypi/
[...redacted for readability...]
```

stloma:
Fixes #3914 

Simple redirect fix, thanks!


di:
@dstufft @ewdurbin I think we need to do this at the edge in VCL instead, right?

tomfbiz:
@di 
In a related issue, I noticed that on Pypi.org (but not locally) https://pypi.org/pypi/project/testinfra/ redirects incorrectly to https://pypi.org/project/project/testinfra/ . (project is doubled up)

di:
@tomfbiz I think that's operating as expected. That first URL with `/pypi/project` was not a valid URL on legacy, so I don't think anyone will be caught off guard by this redirect. Thanks for raising it though.

dstufft:
We don't need to do this in VCL (although doing it there would be slightly more performant in some cases), but doing it here is fine and is easier to test locally.

brainwane:
Thanks for the fix, @stloma! Let us know if you'd like suggestions for more Warehouse stuff to work on. :)

PR #3930, opened 2018-05-08
More descriptive journal actions

di:


dstufft:
This might break some mirroring software that attempts to draw meaning from the action. I’m fine with the change, but if something pops up it’ll be good to keep this change in mind.

Issue #3167, opened 2018-03-07
Audit copywriting against styleguide


nlhkabu:
As per [our documentation](https://warehouse.readthedocs.io/ui-principles/#write-clearly-with-consistent-style-and-terminology), Warehouse uses the [material design writing styleguide](https://material.google.com/style/writing.html).

We should cross check our UIs to ensure that we are meeting these guidelines - in particular in relation to [capitalization and punctuation](https://material.io/guidelines/style/writing.html#writing-capitalization-punctuation).

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

nlhkabu:
To audit:

- [ ] Templates (`/warehouse/templates`)
- [ ] PyPI theme (I will need to make this PR as it is on a separate repository)
- [ ] All "flash" messages, e.g. `request.session.flash` and corresponding tests
- [ ] All error messages, e.g. https://github.com/pypa/warehouse/blob/master/warehouse/manage/forms.py#L30

nlhkabu:
A few places in the templates we seem to have missed:

![screenshot from 2018-04-10 13-40-16](https://user-images.githubusercontent.com/3323703/38581523-0dc24b22-3d05-11e8-9ccb-d0f09b392629.png)
![screenshot from 2018-04-10 13-40-20](https://user-images.githubusercontent.com/3323703/38581524-0de3777a-3d05-11e8-9b20-61dd710b9af2.png)
![screenshot from 2018-04-10 13-45-45](https://user-images.githubusercontent.com/3323703/38581526-0e3132bc-3d05-11e8-896c-1b7813bccc69.png)
![screenshot from 2018-04-10 13-46-05](https://user-images.githubusercontent.com/3323703/38581528-0e4bc2a8-3d05-11e8-9377-ed64a3f6cbc8.png)
[...]

PR #3950, opened 2018-05-12
Style: be succinct, remove "successfully"

hugovk:
For #3167.

The Material Design styleguide says to [be succinct](https://material.io/design/usability/accessibility.html#writing):

> Keep content and accessibility text short and to the point. Screen reader users hear every UI element read aloud. The shorter the text, the faster the screen reader users can navigate it.

"Successfully" can generally be removed. https://ux.stackexchange.com/q/61403/59318

This also removes the CI branch restriction so that I could test this feature branch beforehand, otherwise I would have submitted a broken PR the first time!


.travis.yml
```diff
@@ -60,10 +60,6 @@ matrix:
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 
-branches:
```
pradyunsg: Unrelated change?

hugovk: Not entirely:

> This also removes the CI branch restriction so that I could test this feature branch beforehand, otherwise I would have submitted a broken PR the first time!

I also had to make a few more updates to make the tests pass, before cleaning up the commits, and I don't think anyone needs to see all that going on in here. Better to run tests before proposing.

di: I think we want to leave this as it was, this change will cause the `#pypa-dev` channel to get notified about build status for everyone's forks as well.

Are you not able to run the tests locally?

hugovk: > I think we want to leave this as it was, this change will cause the #pypa-dev channel to get notified about build status for everyone's forks as well.

Oh, sorry for the spam I've already triggered!

How about setting it up so forks don't spam the channel? 

One solution is to put the channel name as a secure variable: https://github.com/travis-ci/travis-ci/issues/5063.

---

> Are you not able to run the tests locally?

No, I've tried running tests locally, but it's erroring. It took a very long time to set up and I also don't think I have the disk space for the whole Docker image and all the downloads.

I only want to change some strings, it's quite heavy to do  https://warehouse.readthedocs.io/development/getting-started/#running-tests-and-linters and it downloaded several GB and taken about half an hour doing `make tests`. And then `make serve` is needed because `make tests` isn't working at the moment, which has downloaded several more GB before erroring:

```
ERROR: for warehouse_web_1  Cannot start service web: driver failed programming external connectivity on endpoint warehouse_web_1 (6704063cda248388329eecac5542769d636f0cb3dcf8bdCreating warehouse_worker_1        ... done

ERROR: for web  Cannot start service web: driver failed programming external connectivity on endpoint warehouse_web_1 (6704063cda248388329eecac5542769d636f0cb3dcf8bd3f5a5b3e3e21bda6bc): Error starting userland proxy: Bind for 0.0.0.0:80: unexpected error (Failure EADDRINUSE)
ERROR: Encountered errors while bringing up the project.
make: *** [serve] Error 1
```

An important point of a CI (for me) is to run all the tests in a known clean environment. Much effort has been put into setting up the CI to do all the tests, I'd much rather leave all that work to the CI server. It's also quicker.

It would be great if we could configure the CI so it can run for feature branches in forks without spamming IRC.

Thanks!

pradyunsg: > One solution is to put the channel name as a secure variable: travis-ci/travis-ci#5063.

I was going to suggest the same -- pip does this.

nlhkabu:
Thanks for the text updates @hugovk.

I am happy for this to be merged on my side, once you have sorted the CI updates with @di and @pradyunsg :)

di:
Could someone at the packaging sprints review the changes in this PR?

.travis.yml
malikoth: Tests succeed on my local copy, FWIW.

malikoth:
@di I have reviewed this.  Doesn't seem that there's a resolution as to whether the master-branch-only thing should be removed from the Travis config that I can see, so pending a final answer on that, this looks good to me.  In addition to conforming to the style guide, these changes read much more naturally.

.travis.yml
di: @hugovk Can we remove this for now and address it in a separate PR?

hugovk: Sure, will do when I'm next at a keyboard.

hugovk: Done, please see PR https://github.com/pypa/warehouse/pull/3983.

di:
Thanks @hugovk and @malikoth!

PR #3955, opened 2018-05-12
Update JS dependencies solving some npm audit issues

yeraydiazdiaz:
[NPM 6 now ships with automatic auditing on install](https://blog.npmjs.org/post/173719309445/npm-audit-identify-and-fix-insecure), this PR fixes quite a few by simply following the automated suggestions.

[`npm audit`](https://docs.npmjs.com/getting-started/running-a-security-audit) still shows quite a few but generally derived from dev dependencies like gulp or webpack and would require more a more manual review.


package.json
```diff
@@ -5,11 +5,11 @@
     "node": "8.9.4"
   },
   "dependencies": {
-    "babel-core": "6.26.0",
+    "babel-core": "^6.26.3",
```
di: I think I'd prefer to pin these dependencies more explicitly, without the `^` specifier.

yeraydiazdiaz: sure, didn't notice that, good catch

PR #3972, opened 2018-05-14
Don't fail on `xz` when unzipped file exists

malikoth:
This would be a bad thing if anybody ever puts local modifications in their local `dev/<db>.sql` file before running `initdb`.  Is that a thing that anybody does?


matt-land:
I just opened a PR for the same issue.

PR #3973, opened 2018-05-14
DX Improvement: recover from failed DB

matt-land:
Hello maintainers.

I followed the guide in `development/getting-started.rst` but had some issues.
I did not have enough resources allocated, and followed the section:
`make initdb is slow or appears to make no progress`
without success.

This patch fixes three issues in recovery from a failed DB setup:
1. if xz has decompressed a file already, it fails to work on a second attempt, and stops make initdb
2. `make clean` does not remove files previously created by `make initdb`
3. The documentation states to run `make serve` and then kick off `make initdb`. This will not work if `make serve` has gotten far enough to open a web session. I changed the documentation to something that worked for me.

Please accept these changes.


di:
Thanks @matt-land!

PR #3975, opened 2018-05-14
Update link to Material design writing guide

GTmmiller:
The current link: https://material.google.com/style/writing.html redirects to https://material.io/design/. I believe the information it was pointing to is now present in this link: https://material.io/design/usability/accessibility.html#writing


di:
Thanks @GTmmiller!

Issue #3838, opened 2018-04-27
Audit SCSS comments


nlhkabu:
I've noticed that some of our comments are out of date.  In the `blocks` folder, we should:

1. Check that the code example at the top of each file is up to date
2. Check that each modifier is documented

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, please work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, please feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

joeygrasso:
I'm taking a look. 👍

nlhkabu:
This has been assigned in the EuroPython sprints.

leoz0610:
Can I work on this issue at Bloomberg PyPA Sprint?

nlhkabu:
Closing in favour of #6482

PR #3976, opened 2018-05-14
Update comments and code example

eevelweezel:
Update css comments and code example, per issue #3838


di:
Thanks @eevelweezel!

di:
Should this close #3838 or is there still more to do?

PR #3978, opened 2018-05-14
Bugfix: tests run without building assets first

matt-land:
Dear Maintainers:

This is a fix for issue 
https://github.com/pypa/warehouse/issues/1536

~~I will discuss with @di  on the remaining test,  "TestWhitenoiseTween.test_serves"~~

This should remove the dependency between building static assets and tests.

I verified this with:
- make clean
- make tests

All tests passed locally

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 script:
   - node -v
   - npm -v
-  - ./node_modules/.bin/gulp dist
+  #- ./node_modules/.bin/gulp dist
   - bin/tests --postgresql-host 127.0.0.1
 
 matrix:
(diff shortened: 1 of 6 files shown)


Bjwebb:
This branch doesn't work for me, but it does if I merge master. (Here's a [merge commit that fixes the conflicts](https://github.com/Bjwebb/warehouse/commit/56f2598cc9ecde5fef90a683cf9ac13d38677511))

This PR creates the file in the bash script. @di I'm not sure if this counts as "properly mocking" the tests as described in see https://github.com/pypa/warehouse/issues/1536#issuecomment-388918026.

PR #3979, opened 2018-05-14
Revise CSP headers

### LINKED ISSUE #3857: use CSP: sandbox on /simple/ pages
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox + https://www.youtube.com/watch?v=fbhW37JZtSA&feature=youtu.be

I believe this is a great fit for /simple/, which don't need any ability to do anthing but have a simple HTML structure.

I _think_ we can replace the whole current header with `Content-Security-Policy: sandbox allow-top-navigations`.

eevelweezel:
Fixes #3857, adds sandbox and allow-top-navigation on /simple/ pages.


eevelweezel:
Alright, revised this to replace the headers on /simple/ pages.

di:
Thanks @eevelweezel!

PR #3988, opened 2018-05-15
Elasticsearch 6

di:
We'll need to spin up a new cluster and set `ELASTICSEARCH_SIX_URL` before this can be merged.

(cc @HonzaKral @ewdurbin)

Closes #2749, closes #3319.


docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
     image: rabbitmq:latest
 
   elasticsearch:
-    image: elasticsearch:5
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.2.4
```
HonzaKral: this might have to be `docker.elastic.co/elasticsearch/elasticsearch-oss:6.2.4` otherwise `x-pack` would get installed which can interfere with the tests a bit. Just if you run into unexpected issues.

di: It didn't affect the tests locally, but this is good to know. 👍

PR #3989, opened 2018-05-15
Fix reindex

### LINKED ISSUE #3746: Search reindex task leaves empty index.
The 'Search projects' function does not work for me on [https://pypi.org](https://pypi.org). Irrespective of the query, the search does not return any results. (Example: [https://pypi.org/search/?q=numpy](https://pypi.org/search/?q=numpy))

HonzaKral:
Fixes #3746 

see https://github.com/pypa/warehouse/issues/3746#issuecomment-389255637 for details:

The problem is that `Project` doesn't have the new `index` associated with it when running in celery - the `Index.doc_type` method only sets the `index` name on the `DocType` *if no index is set*. That would explain why it runs fine from CLI but not from celery - CLI starts with fresh objects that don't have an index associated with it while celery is a long running process where it might happen that the previous `reindex` run has already registered an index name with the `Project` doc type.

The fix is to issue the document dicts without `_index` information and then pass it in explicitly when calling `parallel_bulk`.


HonzaKral:
relevant code: https://github.com/elastic/elasticsearch-dsl-py/blob/master/elasticsearch_dsl/index.py#L90-L91

PR #3994, opened 2018-05-15
Forgot password rephrased 

Adarsh12AD:

PR #4010, opened 2018-05-16
Spaces removed from forgot password

Adarsh12AD:
Fixes #3994
Removed extra spaces and rephrased the procedure


di:
@Adarsh12AD Can you address the rest of the feedback on #3994 as well?

PR #4015, opened 2018-05-16
Revert "Warehouse API currently redirects pypi.org/pypi/{project}/json/ (#3971)"

### LINKED ISSUE #4012: Redirect loop when requesting missing package
**Describe the bug**
When I `curl -v https://pypi.org/pypi/bogus/1.0.0/json` with an invalid package/version combination, I get put in an infinite redirect loop. If the package/version combination is correct, then the data is served as expected.

I noticed this because the `license_finder` gem just now started having issues when looking for my team's internal (non-PyPi) packages. Previously, it hadn't complained.

**Expected behavior**
For an invalid package/version combination, I'd expect a 404 (or whatever the previous behavior was).

**To Reproduce**
`curl -v https://pypi.org/pypi/bogus/1.0.0/json`, note that the `Location` header leads you to the same URL.

**My Platform**
Pretty sure my platform isn't relevant. I've reproduced the problem both from my personal internet connection as well as within AWS.

**Additional context**
Adding @dstufft @di as suggested on IRC.

di:
This reverts commit 464271b9676819e97348dd72646bad5dfe41d873.

Fixes #4012.

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -288,16 +288,6 @@ def test_routes(warehouse):
         pretend.call("/p/{name}/", "/project/{name}/", domain=warehouse),
         pretend.call("/pypi/{name}/", "/project/{name}/", domain=warehouse),
         pretend.call(
-            "/pypi/{name}/{version}/json/",
-            "/pypi/{name}/{version}/json",
-            domain=warehouse,
-        ),
-        pretend.call(
-            "/pypi/{name}/json/",
-            "/pypi/{name}/json",
-            domain=warehouse,
-        ),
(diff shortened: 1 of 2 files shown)


dstufft:
Let's open an issue to re-fix the original issue now that it's reverted.

di:
New issue for the original issue is here: #4016.

jbergknoff-rival:
It seems like the Travis build is stalled. Any idea know when this change will land?

di:
It might take a bit for this to get deployed to PyPI because there's currently a pretty long backlog in Travis due to all the PyPA development happening during the sprints.

PR #4017, opened 2018-05-17
Provide example of how to use db object in interactive shell

mbacchi:


docs/development/getting-started.rst
```diff
@@ -368,6 +368,24 @@ db     The SQLAlchemy ORM ``Session`` object which has already been configured
...
+
+    $ make shell
+    docker-compose run --rm web python -m warehouse shell
+    Starting warehouse_redis_1 ...
+    ...
+    (InteractiveConsole)
+    >>>
+    >>> from warehouse.accounts.models import User
+    >>> from warehouse.db import Session
+    >>> session = Session(bind=config.registry["sqlalchemy.engine"])
```
dstufft: You actually don't need this bit, you can do:

```python-console
>>> from warehouse.accounts.models import User
>>> db.query(User).filter_by(username='test').all()
[User(username='test')]
```

In other words, the ``db`` object is a global in the shell, and it's already an instance of the Session.

mbacchi: Thanks @dstufft ! I updated that section (and tested it, which works perfectly,) I didn't realize the db object did that.

PR #4025, opened 2018-05-17
Do not push tab state to history. #3997

### LINKED ISSUE #3997: [ui] url hijacking on page load
**Describe the bug**
On page load there is some JS that hijacks the url to focus on the `#description` anchor.

anchor

**Expected behavior**
Do nothing with users' urls, as there is important info above that should be immediately visible:

correct

**Additional context**
I think this caused it https://github.com/pypa/warehouse/pull/3829 cc @nlhkabu @di

yeraydiazdiaz:
Fixes #3997


di:
I think we probably still want this behavior, we just don't want it to happen on page load when there is no fragment.

yeraydiazdiaz:
Updated, here's a capture of the current behaviour

project_tabs

Issue #890, opened 2016-01-04
Create a credits page


nlhkabu:
It would be great to create a page where we can list our major contributors, with a small profile for each person.  This should be linked to from the footer.

dstufft:
One concern is that it has a pretty high chance of getting stale data for the small profiles depending on what exactly is in the profiles. I'm +1 on giving credit of some kind though. Another option would be http://humanstxt.org/

Perhaps give credit to the libraries we're using too?

dstufft:
Going to drop this from Become PyPI, we don't have it on PyPI already and it seems like an enhancement that can happen any time.

nlhkabu:
Sounds reasonable to me :P

nlhkabu:
It would be great to also credit OSS dependencies on this page.  See #https://github.com/pypa/warehouse/issues/1099 for further info.

Thanks to @stevepiercy for this suggestion.

mbacchi:
I'm working on the solution to this issue, and wanted to describe what I'm
planning for discussion and to ask a couple questions. (FYI @di)

Contributor credits

This doesn't fully address the initial request to have a 'developer profile' on
the warehouse page. But I thought the easiest way to do this was to simply query
Github and list the names of the people with links to their Github profile.

1. I initially tried to do this with a live query to Github at page load time,
but it took over 20 seconds thus making the page load way too long. So I
mentioned my plan to use a celery task on #pypa-dev and dstufft agreed it would
be the right way to go.

2. The major parts of this solution will be:
  a. new table in the database (warehouse_contributors) (work complete)

  b. celery task to query the Github api (this will require a Github API access
  token owned by the PyPA and set via an environment variable) (work mostly
  complete, still need to make it to run)

  c. credits.html template which will display the people who have contributed
  (planning to sort alphabetically by name, but could sort by commit count, open
  for discussion) (work mostly complete)

  d. Updated views.py. (work partially complete)


OSS libraries/dependencies used

1. I think this could be addressed by listing the requirements.txt dependencies
and linking to their PyPi project pages. We could start with a simple list
collected from requirements.txt, but then how would they be updated when
requirements are added or removed? Should we do a database table with this list
and a celery task to update these on a regular basis too? I don't think this
will enable displaying project icons such as Pyramid or SQLAlchemy, but that
could be a version 2.0 issue if enough people think its important.

My biggest question so far is about where to put this code. I'm currently
putting the celery task, view and model in warehouse/utils but only because it
[...]

PR #4026, opened 2018-05-17
Remove the bulk of the unused tables

dstufft:
These are all left over tables from Legacy PyPI that we no longer are using or need to worry about. Removing them makes the database cleaner and easier to interactively explore.

I'm going to leave this for @ewdurbin to merge, because we should probably take a database dump right before this happens, just in case we ever need to go back and pull data out of these tables (although I believe that very unlikely).

PR #4038, opened 2018-05-18
WIP: #890 create a credits page

mbacchi:
This is a very early stage WIP PR. I need to add the following before it should be considered merge fodder:

1. tests
2. a section on the credits page for packages we rely on in warehouse
3. enhanced error handling in warehouse/credits/tasks.py

Also, after speaking with @dstufft and @di this will require a GitHub personal access token for the API queries in this code. We make approximately 100 API queries as of now, and that will increase as more users contribute to Warehouse. The limit imposed by Github is 60 per hour. So in order to work around that and using a personal access token, GitHub recommends a way to setup a 'machine user' which is a GitHub account you could create with read only access to the repository, and use that key in the API query: https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users


dstufft:
Getting a token shouldn't be an issue, just make sure it's configurable and we'll make sure it's in the environment before we deploy this.

mbacchi:
I'm getting a test failure now because there are multiple heads in the migrations: https://travis-ci.org/pypa/warehouse/jobs/388056080#L1364

It looks like PR #3367 and #4026 changed a bunch of migrations, and now the version that I was revising is revised by that PR:

https://github.com/mbacchi/warehouse/blob/issue890/warehouse/migrations/versions/7a9ceafce0e8_add_contributor_table_for_issue_890.py#L16

https://github.com/pypa/warehouse/blob/master/warehouse/migrations/versions/f2a453c96ded_remove_no_longer_needed_tables.py#L16

Any ideas how I can deal with this? Should I rebase, or remove the table and recreate the migration, or start the PR again? Thanks.

dstufft:
If your migrations don't conflict, you can just edit your migration and change the ``down_revision`` to point to the latest revision. You can also do ``python -m warehouse db merge heads`` and it'll create a merge revision.

mbacchi:
Wow, thanks for the seriously quick response @dstufft. Which do you prefer, is there a 'preferred' way out of the two you suggested?

dstufft:
I would probably edit the ``down_revision`` personally, but either one is fine really. One results in a more linear history, and one is less manual tweaking.

nlhkabu:
Note: 

warehouse/templates/pages/credits.html
```diff
@@ -0,0 +1,42 @@
...
+{% block title %}{% trans %}Development Credits{% endtrans %}{% endblock %}
+
+{% set core_contributors = ['ewdurbin', 'di', 'dstufft', 'nlhkabu',
+                        'yeraydiazdiaz'] %}
+
+{% block content %}
+  <section class="horizontal-section">
+  <div class="narrow-container">
+    <h2>{% trans %}Development Credits{% endtrans %}</h2>
+      <p>{% trans %}We'd like to thank the following people who have helped build the{% endtrans %} <a href="https://pypi.org/">{% trans %}PyPI Warehouse{% endtrans %}</a> {% trans %} application for the Python community{% endtrans %}:</p>
```
merwok, line 24: Splitting text in parts like this will not work for many languages.
This could be one text with a little markup in the middle (trusting translator to keep it as is), or you could move the markup string to a variable.

warehouse/credits/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# limitations under the License.
+
+from celery.schedules import crontab
+
+from warehouse.credits.tasks import get_contributors
+
+
+def includeme(config):
+    # Add a periodic task to get contributors every 24 hours
+    if config.get_settings().get("warehouse.github_access_token"):
```
di, line 20:
```suggestion
    if config.get_settings().get("github.token"):
```

warehouse/config.py
```diff
@@ -175,6 +175,9 @@ def configure(settings=None):
         default="https://api.github.com/meta/public_keys/token_scanning",
     )
     maybe_set(settings, "warehouse.trending_table", "WAREHOUSE_TRENDING_TABLE")
+    maybe_set(
+        settings, "warehouse.github_access_token", "WAREHOUSE_GITHUB_ACCESS_TOKEN"
+    )
```
di, line 180:
```suggestion
```

warehouse/credits/tasks.py
```diff
@@ -0,0 +1,92 @@
...
+        "dependabot-preview[bot]",
+        "github-actions[bot]",
+        "pyup-bot",
+        "requires",
+        "weblate",
+    ]
+
+    contributors = {}
+
+    access_token = request.registry.settings["warehouse.github_access_token"]
```
di, line 42:
```suggestion
    access_token = request.registry.settings.get("github.token")
```

tests/unit/credits/test_init.py
```diff
@@ -0,0 +1,39 @@
...
+from warehouse import credits
+from warehouse.credits.tasks import get_contributors
+
+
+@pytest.mark.parametrize("with_github_access_token", [True, False])
+def test_includeme(with_github_access_token):
+
+    config = pretend.stub(
+        get_settings=lambda: (
+            {"warehouse.github_access_token": "foobar"}
```
di, line 27:
```suggestion
            {"github.token": "foobar"}
```

mbacchi:
Hey @di, I got some of the `test_tasks.py` working but I am stuck on how to provide a pretend stub or fixture for the database query. Any input you could give would be appreciated.

di:
@mbacchi You've got the `db_request` test fixture in your test, so this means that the tests will run against an actual isolated instance of the database. Generally that's preferable here to mocking out the database calls. That means that depending on what you're testing, you might want to do some setup in advance (e.g. creating new `Contributors` and adding them to the database).

For example, see this test that's creating projects, users, etc.

https://github.com/pypa/warehouse/blob/e7fc79b3445dc1367e9a67b0e27d2e30cc5ef218/tests/unit/forklift/test_legacy.py#L642-L646

mbacchi:
@di asked me to upload an image of what this looks like for folks to look at. Here it is.

Screenshot from 2022-03-22 15-14-51

warehouse/credits/contributors.py
```diff
@@ -0,0 +1,29 @@
...
+from warehouse import db
+
+
+class Contributor(db.Model):
+
+    __tablename__ = "warehouse_contributors"
+
+    contributor_login = Column(Text, primary_key=True, unique=True, nullable=False)
+    contributor_name = Column(Text, nullable=False)
+    contributor_url = Column(Text, nullable=False)
```
merwok, line 24: I think the file should be named `models.py`, and the fields should not start with `contributor`.
(Taking sponsors as an example)

miketheman:
#Triage

It's been quite some time for this PR, and in conversation today we've decided:

- this PR has a bunch of conflicts with modern code, due to aging
- there's not a huge appetite for this feature, considering that https://github.com/pypi/warehouse/graphs/contributors shows who's contributed

Closing for now, but if someone is passionate about re-adding this, please open an issue to discuss the potential implementation prior to writing more code.

Issue #4018, opened 2018-05-17
Test PyPI should include the `--index-url` argument for the pip install command.


theacodes:
**What's the problem this feature will solve?**
Copy-pasting the pip install command on Test PyPI won't actually install from Test PyPI

**Describe the solution you'd like**
instead of `pip install package_name` we should show `pip install --index-url https://test.pypi.org/simple/ package_name`

**Additional context**
It would also be super awesome if pip just had an alias for this, e.g. `pip install --with-test-pypi package_name`.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

fitodic:
@nlhkabu Can I try this one?

nlhkabu:
Great @fitodic

nlhkabu:
@fitodic GitHub wont let me assign this to you - so I'll assign it to me so no one else takes it :)

nlhkabu:
Removing the sprint label to reserve :)

fitodic:
Thanks :+1:

fitodic:
@nlhkabu I've just noticed there's already a pull request for this issue: https://github.com/pypa/warehouse/pull/4044

Should I move on?

nlhkabu:
ah, yes sorry.  I will remove this from the sprint list.  my mistake!

di:
Fixed in #4941.

PR #4044, opened 2018-05-19
Include '--index-url' argument for the pip install command for Test PyPI

kfr2:
The template logic for a package's pip installation command has been
moved to a jinja macro called `pip_command`. This macro furthermore
determines whether or not to add the Test PyPI's `--index-url` based on
the existence of the *testPyPI* variable.

This PR implements #4018.


di:
Fixed in #4941, thanks @kfr2!

Issue #4079, opened 2018-05-25
`long_description_content_type` flag in setup.py seems to be ignored on upload.


fake-name:
**Describe the bug**
I have a python package on PyPi that just (as of the last ~4-5 days) now fails to upload, with a new error:

```
durr@rwpscrape /m/S/S/WebRequest> python3 setup.py sdist upload

[snip build process]

Submitting dist/WebRequest-0.0.28.tar.gz to https://upload.pypi.org/legacy/
Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
error: Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

Ok, that's clear enough. My readme is in markdown, let's see how to fix this.

Going to https://packaging.python.org/tutorials/packaging-projects/#description, it states you can add `long_description_content_type="text/markdown",` to your `setup.py` file, and it'll be processed as markdown.

```
<does edits>....
<submits to PyPi again
```

```
Submitting dist/WebRequest-0.0.28.tar.gz to https://upload.pypi.org/legacy/

[snip build process]

Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
error: Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

What?

**Expected behavior**
I expect if I specify the readme is markdown, as the documentation says to do so, the readme should be treated as markdown.

**To Reproduce**
Attempt to upload a package with a readme that is valid markdown, but not valid reStructuredText.

**My Platform**
 - Python 3.5.2 x64
 - Ubuntu 16.04 LTS
 - setuptools-39.2.0 (I specifically checked that it's the most current version)
 - readme_renderer 20.0

Actual project in question is [here](https://github.com/fake-name/WebRequest). The readme.md in question is [here](https://github.com/fake-name/WebRequest/blob/master/README.md), and the setup.py is [here](https://github.com/fake-name/WebRequest/blob/master/setup.py)

fake-name:
Note. https://pypi.org/help/#description-content-type also claims you can use other formats, but after installing `readme_renderer`, local checking appears to fail (?):

```
durr@rwpscrape /m/S/S/WebRequest> python3 setup.py check -r -s
running check
warning: Check: Not checking long description content type 'text/markdown', this command only checks 'text/x-rst'.

[...]

PR #4099, opened 2018-05-29
Revert erroring when readme content does not render

theacodes:
This temporarily disables the validation added #3980 which is preventing several users from uploading packages (#4079).

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -919,6 +919,7 @@ class TestFileUpload:
             "for more information."
         ).format(name)
 
+    @pytest.mark.xfail(reason='https://github.com/pypa/warehouse/issues/4079')
     @pytest.mark.parametrize(
         ("description_content_type", "description", "message"),
         [
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -919,7 +919,7 @@ class TestFileUpload:
             "for more information."
         ).format(name)
 
-    @pytest.mark.xfail(reason='https://github.com/pypa/warehouse/issues/4079')
+    @pytest.mark.xfail(reason="https://github.com/pypa/warehouse/issues/4079")
     @pytest.mark.parametrize(
         ("description_content_type", "description", "message"),
         [
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -899,24 +899,27 @@ def file_upload(request):
         rendered = readme.render(
             form.description.data, description_content_type, use_fallback=False
         )
-        if rendered is None:
-            if form.description_content_type.data:
-                message = (
-                    "The description failed to render "
-                    "for '{description_content_type}'."
-                ).format(description_content_type=description_content_type)
-            else:
-                message = (
-                    "The description failed to render "
-                    "in the default format of reStructuredText."
-                )
-            raise _exc_with_message(
-                HTTPBadRequest,
-                "{message} See {projecthelp} for more information.".format(
-                    message=message,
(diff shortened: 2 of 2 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -919,6 +919,7 @@ def test_fails_with_invalid_names(self, pyramid_config, db_request, name):
             "for more information."
         ).format(name)
 
+    @pytest.mark.xfail(reason='https://github.com/pypa/warehouse/issues/4079')
```
ewdurbin: `black` will want these to be `"` instead of `'`, though that landed while I was away :-D not sure if we enforce it in linter yet.

warehouse/forklift/legacy.py
```diff
@@ -892,31 +892,34 @@ def file_upload(request):
         )
 
     # Uploading should prevent broken rendered descriptions.
-    if form.description.data:
-        description_content_type = form.description_content_type.data
-        if not description_content_type:
-            description_content_type = "text/x-rst"
-        rendered = readme.render(
```
ewdurbin: import of `warehouse.utils.readme` on line 55 no longer used.

ewdurbin:
Looks like the original PR no longer cleanly reverts which is a bummer, so the comment out approach is OK to keep from losing too much context but isn't the standard approach.

Do we know what the long term fix is for the issue?

theacodes:
> Do we know what the long term fix is for the issue?

Not sure, I'll continue investigating.

ewdurbin:
@theacodes perfect. if you can push the lint issues  we'll get this merged and deployed to unblock publications.

theacodes:
delinted

di:
It seems to me like this is working as expected, and that the issue lies with either twine, wheel or setuptools versions being old and not supporting Metadata 2.1, or users using distutils instead of twine to upload.

theacodes:
Yes, it's working as expected but having unintended consequences. I'd
prefer to err on the side of allowing uploads while we sort this out.

ewdurbin:
gonna merge up to unblock users. any opposition @di?

ewdurbin:
On closer inspection I agree with @di that this is a metadata/client issue. I'm not sure about the revert given that, but do see at least one scenario [where the wheel upload worked but sdist did not](https://github.com/pypa/warehouse/issues/4079#issuecomment-392743793). This still seems like it could be a client issue but is confusing enough that it probably requires further investigation before shipping the validation.

ocefpaf:
> It seems to me like this is working as expected, and that the issue lies with either twine, wheel or setuptools versions being old and not supporting Metadata 2.1, or users using distutils instead of twine to upload.

I'm experiencing this using Travis-CI auto deployment feature. According to the logs Travis-CI is using

- `setuptools 39.2.0`
- `twine 1.11.0`
- `wheel 0.31.1`

those look new enough but I'm not sure if Travis-CI is calling `twine` or `setup.py` to upload.

See https://travis-ci.org/pyoceans/erddapy/jobs/385360404#L1591

Thanks to this PR the deployment worked as expected: https://travis-ci.org/pyoceans/erddapy/jobs/385409165

di:
> (But the Markdown description is not rendering as expected 😬)

Right, this is what we were trying to prevent. After taking a look at this, I'm fairly sure the problem is either out of date dependencies or using  `setup.py upload` instead of twine. For any environment for which these things are out of the user's control, using a Markdown description is essentially unsupported.

ocefpaf:
> Right, this is what we were trying to prevent.

I get the same if I upload manually with all tools updated and twine :-/

> using a Markdown description is essentially unsupported.

That is fine IMO as long as the upload is not blocked.

Thanks everybody for the info and the quick action!

PR #4120, opened 2018-06-03
[Bugfix] fix wrong docker Compose file format version

gangh:
Error message when I use ``` docker-compose build```:
```
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. 
```
specify a version of  "2",solved it

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '2'
 
 services:
   db:


di:
Hi @gangh, what version of `docker-compose` are you using? (You can determine this by running `docker-compose --version`)

I don't think we want to downgrade the version here, instead you should upgrade your version of `docker-compose` to one that's compatible with more versions (see https://docs.docker.com/compose/compose-file/compose-versioning/).

ewdurbin:
Indeed, upgrading Docker/Docker-Compose is the correct way forward.

PR #4167, opened 2018-06-13
email.ses: Handle SoftBounce => Delivered transition

ewdurbin:
Reference:

https://sentry.io/python-software-foundation/warehouse-production/issues/579601360/?query=is:unresolved
https://sentry.io/python-software-foundation/warehouse-production/issues/579601394/?query=is:unresolved

Looks like SoftBounces can occur _before_ Delivery.


ewdurbin:
@dstufft I haven't played with this state machine before and would appreciate your review!

warehouse/email/ses/models.py
```diff
@@ -155,6 +155,12 @@ def _reset_transient_bounce(self):
         collector=lambda iterable: list(iterable)[-1],
     )
 
+    soft_bounced.upon(
+        deliver,
+        enter=delivered,
+        outputs=[],
```
dstufft: We probably want to reset our transient bounces in this case, because we've successfully delivered it. We can do that by adding ``_incr_transient_bounce`` to the outputs, and we'll need to set a collector too.

ewdurbin:
@dstufft ready for another look!

PR #4184, opened 2018-06-15
Adding project_urls to json api and docs.

### LINKED ISSUE #3831: Expose project_urls in JSON API
https://packaging.python.org/tutorials/distributing-packages/#project-urls

Related to #3798 / #3820

I realized project_urls is not currently exposed by the JSON API. I propose adding it.

Though the keys in the project_urls dict can be anything, they're fairly standardized, enough to be useful when querying for them over and API. For example, [Flask's API response](https://pypi.org/pypi/Flask/json) lists its home_page as https://www.palletsprojects.com/p/flask/ (not it's github account which is fairly typical), and puts it's GitHub link in `project_urls['Code']`, which is not currently in the API response.

nixjdm:
Fixes #3831 

This adds the `project_urls` to the json api and its docs. I added to the trivial test case (None) and the detailed test case. The construction of the `project_urls` is in the same form as in https://github.com/pypa/warehouse/blob/master/tests/unit/packaging/test_models.py#L227-L255 , starting as a list.

The detailed test case has a number of different URIs tested which I think are more comprehensive than other tests for that. The test cases come from `tests/unit/packaging/test_models.py::test_urls`, https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Examples, and https://stackoverflow.com/questions/695438/safe-characters-for-friendly-url.

The only URIs this project can't handle in `project_urls` are ones with commas, which are allowed generally in URIs, though they aren't common. This is because of the way `project_urls` are ingested, as lists as in the form you see in the test, then to construct the dict each element is split on a comma. There's probably a better way of doing this for a potential improvement.


dstufft:
Sorry this took so long! This looks great to me, thanks!

nixjdm:
No problem @dstufft! Thanks!

PR #4204, opened 2018-06-20
Google Analytics: Respect common Do Not Track settings.

ewdurbin:


warehouse/static/js/warehouse/utils/analytics.js
```diff
@@ -21,7 +21,7 @@ export default () => {
   if (element) {
     var dnt = navigator.doNotTrack || window.doNotTrack || navigator.msDoNotTrack;
     if (dnt == "1" || dnt == "yes") {
-      window['ga-disable-' + element.dataset.gaId] = true;
+      window["ga-disable-" + element.dataset.gaId] = true;
     }
```
di: We don't need to do anything else in this script if this is true, so we can probably wrap everything after this conditional in an `else`.

PR #4209, opened 2018-06-21
make LocalFileStorage work end-to-end for dev environment

### LINKED ISSUE #4197: LocalFileStorage is broken for dev environment
We have a `LocalFileStorage` class for uploading distributions locally: https://github.com/pypa/warehouse/blob/c3a2393481ab0aa461486664a2ab3137c0bd5e0b/warehouse/packaging/services.py#L24-L36

And we have it configured to be used in development: https://github.com/pypa/warehouse/blob/d1e4b7baad46b4cd2acdf379f0ce361da5ec1b92/dev/environment#L23

However since it's configured to use the `example.com` host, it won't actually serve an uploaded file as intended.

We should fix the configuration so it works out-of-the-box in local development, or nix it entirely.

ewdurbin:
fixes #4197


di:
One edge case for this that I didn't consider: if a developer uploads a file, then makes some change which causes the containers to be rebuilt, there will still be an entry for the filename in the database, but the `files` service will 404. I could see this being pretty confusing to someone doing local testing.

Maybe the `packages` volume should be shared with the host OS?

ewdurbin:
@di the volume should persist across container builds, since it is defined independently of a single container.

ewdurbin:
confirmed this behavior! the volume will persist until the volume is deleted directly, `make purge`, or `make stop` is called.

docker-compose creates the volume as a docker volume, which will hang around until explicitly deleted.

di:
Ah, I think I was getting confused by an upload I did prior to switching to this branch, my bad!

I think the fact that this would happen with a `make stop` is probably less of an issue.

ewdurbin:
`make stop` will also clear the DB volume, so at least the state will be consistent.

PR #4217, opened 2018-06-22
Add Sentry for frontend JS

di:
This PR adds Sentry error collection for our frontend JS.

It requires setting the `SENTRY_FRONTEND_DSN` environment variable to a valid Sentry DSN.


warehouse/templates/base.html
```diff
@@ -78,6 +78,14 @@
 
     <link rel="search" type="application/opensearchdescription+xml" title="PyPI" href="{{ request.route_path('opensearch.xml') }}">
 
+    {# This needs to be included first (in order to capture errors when
+     # executing other JS scripts. It also can't be async, because then order
+     # of execution is not guaranteed, and it can't be deferred, because most
+     # of our JS would then execute before it, and we might miss exceptions.
+     #}
+    {% if request.registry.settings.get("sentry.frontend_dsn") -%}
+    <script src="https://cdn.ravenjs.com/3.26.2/raven.min.js" crossorigin="anonymous"></script>
```
ewdurbin: can we add a [`integrity`](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) attribute here with the expected checksum? this is a lot more feasible with a versioned url, unlike our Google Analytics and Fastly Insights loads.

ewdurbin:
kinda gross looking, but they do appear to have a solution for async loading documented: https://docs.sentry.io/clients/javascript/install/#async-loading

di:
That would require either `script-src: 'unsafe-inline'` or a bit of rearchitecting of how we serve our JS bundle. The script is pretty small, so I'm not too concerned about how long it will block the DOM.

ewdurbin:
Sounds good to me :)

ewdurbin:
`SENTRY_FRONTEND_DSN` for test.pypi.org and pypi.org! let's roll!

PR #4263, opened 2018-07-06
fix #4137

### LINKED ISSUE #4137: font-weight 300 is too light to read
https://github.com/pypa/warehouse/blob/e1f61c6bb1f84bd3ec35aed765765fc28caaf069/warehouse/static/sass/settings/_fonts.scss#L35
https://github.com/pypa/warehouse/blob/e1f61c6bb1f84bd3ec35aed765765fc28caaf069/warehouse/static/sass/settings/_fonts.scss#L17

The `$base-font-weight` is set to 300. For "Source Sans Pro", it means "Source Sans Pro ExtraLight", which is too light to read the text comfortably:
screenshot

This situation happens when webfont is not loaded properly, e.g. lose connection to fonts.googleapis.com

TitanSnow:
fix #4137 

set $base-font-weight to 400

### DIFF
--- a/warehouse/static/sass/settings/_fonts.scss
+++ b/warehouse/static/sass/settings/_fonts.scss
@@ -32,5 +32,5 @@ $button-font-size: $base-font-size;
 $input-font-size:  $base-font-size * 1.1;
 
 $base-line-height: 1.4;
-$base-font-weight: 300;
+$base-font-weight: 400;
 $bold-font-weight: 600;


di:
Thanks @TitanSnow! Can you change this to `500` as @nlhkabu suggested here? https://github.com/pypa/warehouse/issues/4137#issuecomment-402778101

TitanSnow:
https://github.com/pypa/warehouse/blob/dc120b9c0f8fba9dc98809e68e1b56a1d1bda5e6/warehouse/templates/base.html#L57

But it loads Source Sans Pro with weight 400, 600, 700, no 500. If $base-font-weight is set to 500, the font weight will be 500 before the web font is loaded and be 400 after the web font is loaded, which makes a blink. In addition, other common-used fonts just not have a weight 400 so there's no difference between 400 and 500 for them.

di:
Thanks for pointing that out. It also looks like Source Sans Pro is not available from Google Fonts at a 500 weight. I'll let @nlhkabu review, but this looks good to me.

TitanSnow:
Oh, well, sorry that I was massed. 400 is the normal weight, while 500 is the weight between normal and bold. I guess few fonts have weight 500.

nlhkabu:
thanks @TitanSnow :)

PR #4274, opened 2018-07-09
Send raw email to allow for Unicode in names

### LINKED ISSUE #4273: PyPI sends emails with unicode in header
**Describe the bug**

I created a new account on pypi.org, but I am not receiving any email verification link by email - not after accouunt creation and also not after resending the verification link (even if pypi.org claims "Verification email for nicolas@seita.nl resent").

test.pypi.org works for me, using the same account name and email address. I deleted and recreated the account on both pypi.org and test.pypi.org - same behaviour (well, the initial email did sometimes not arrive on test.pypi.org, but the the verification resend always arrived right away).

**Expected behavior**

I should get an email token after registering, and also when I resend the verification mail from the account page. I know it might take a while, but I am experiencing this since Friday.

**To Reproduce**

Register on pypi.org, wait for email. Resend the verification email from the account settings page, wait again.

**My Platform**

Firefox 52

di:
This PR switches our SES email service to use `send_raw_email` instead of `send_email`.

Fixes #4273.

### DIFF
--- a/tests/unit/email/test_services.py
+++ b/tests/unit/email/test_services.py
@@ -101,10 +101,18 @@ class TestSESEmailSender:
         assert sender._sender == "DevPyPI <noreply@example.com>"
         assert sender._db is request.db
 
-    def test_send(self, db_session):
+    def test_send(self, db_session, monkeypatch):
+        # Determine what the random boundary token will be
+        import random
+        import sys
+
+        random.seed(42)
+        token = random.randrange(sys.maxsize)
+        random.seed(42)
+
         resp = {"MessageId": str(uuid.uuid4()) + "-ses"}
         aws_client = pretend.stub(
-            send_email=pretend.call_recorder(lambda *a, **kw: resp)
+            send_raw_email=pretend.call_recorder(lambda *a, **kw: resp)
         )
         sender = SESEmailSender(
             aws_client, sender="DevPyPI <noreply@example.com>", db=db_session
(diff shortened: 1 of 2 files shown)


tests/unit/email/test_services.py
```diff
@@ -101,10 +101,18 @@ def test_creates_service(self):
         assert sender._sender == "DevPyPI <noreply@example.com>"
         assert sender._db is request.db
 
-    def test_send(self, db_session):
+    def test_send(self, db_session, monkeypatch):
```
ewdurbin: it's not clear to me what the inclusion of the `monkeypatch` fixture is doing here, resetting the touch of `random.seed`?

di: Oops, nope, just accidently leftover.

tests/unit/email/test_services.py
```diff
@@ -101,10 +101,18 @@ def test_creates_service(self):
...
+        # Determine what the random boundary token will be
+        import random
+        import sys
+
+        random.seed(42)
+        token = random.randrange(sys.maxsize)
+        random.seed(42)
```
ewdurbin: should we rescramble the seed after the test? `random` is bad, but I assume other things are using it in the codebase and we probably want to either prefer deterministic pseudorandom behavior... or pure pseudorandom behavior for our suite.

ewdurbin: not a blocker, just a curio.

di: I don't think it's necessary!

Issue #4154, opened 2018-06-11
GitHub statistics: separate thousands


hugovk:
**Describe the bug**
Numbers in GitHub statistics over 999 should have thousands separated.

**To Reproduce**
1. Visit a project with GH stats over 999.
Examples:
* https://pypi.org/project/colorama/
* https://pypi.org/project/cryptography/
* https://pypi.org/project/pytest/

**Actual behavior**

Numbers over 999 don't have separated thousands.

![image](https://user-images.githubusercontent.com/1324225/41219581-65875dce-6d67-11e8-936b-6ff93bc3b424.png)

**Expected behavior**
Numbers over 999 should have thousands separated, for example "Stars: 2,680".

Should apply to "Forks" and "Open issues/PRs" in case are they're big enough.

GitHub comma separates thousands:

![image](https://user-images.githubusercontent.com/1324225/41219794-1f8b522a-6d68-11e8-85b8-d1c19a3d77c9.png)

**Additional context**
* https://github.com/pypa/warehouse/blob/master/warehouse/templates/includes/packaging/project-data.html#L33-L53
* See also https://github.com/pypa/warehouse/pull/1362.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

PR #4287, opened 2018-07-12
4154 GitHub stats thousands format

nicholasserra:
Refs #4154 Numbers in GitHub statistics over 999 should have thousands separated.

- Switch `forks` key to `forks_count` key in template. Both keys have the same data.
- Check for `_count` in key name and format number to get commas in thousands place. Uses `toLocaleString` which has high browser support:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

screen shot 2018-07-12 at 2 59 23 pm


ewdurbin:
@nlhkabu or @di could you take a look?

di:
Thanks @nicholasserra!

PR #4325, opened 2018-07-19
Expose requires_python in each release in JSON API

### LINKED ISSUE #3024: Expose requires_python for each release in JSON API
The following JSON:

https://pypi.python.org/pypi/astropy/json

or:

https://pypi.io/pypi/astropy/json

includes the following for the latest release:

```
        "requires_python": ">=3.5", 
```

but it would be really helpful to have this for all releases listed in the ``releases`` section, to avoid having to hit the server for each release to find out the ``requires_python``.

uranusjr:
Forgive me if this is not an intended change. I was working on something related (at the client side), and this strikes me as a very useful entry. It is a little awkward to expose this key with file information (it is logically tied with the release), but since `releases` is a list, it has nowhere else to go :(

Fix #3024.


ewdurbin:
Thanks @uranusjr!

`requires_python` is already exposed for a given release in the JSON API in the top level `info` key when requesting `https://pypi.org/pypi/<project>/<release>/json` for a specific release or `https://pypi.org/pypi/<project>/json` for the latest release.

https://github.com/pypa/warehouse/blob/4ed82418721cbfa444191cccb8d3817bf70c6079/warehouse/legacy/api/json.py#L161

```
$ curl -s https://pypi.org/pypi/Django/json | jq '.info.requires_python'
">=3.4"
```

```
$ curl -s https://pypi.org/pypi/Django/1.11.14/json | jq '.info.requires_python'
""
```

Can you expand on your use case for exposing this value for each release file? The value is specific to a given release and can't vary for each file, but as you say there's no better place to expose this value for all releases at once.

uranusjr:
@ewdurbin Hi, thanks for the response! My use case is be able to get the answer “what is the best (e.g. latest) version of a given package that matches my Python version” without needing to query multiple endpoints. In the current form, if the latest version is not installable, I need to query each version one by one, resulting in a lot of  HTTP queries.

As mentioned in #3024, the simple API currently carries this information in `data-requires-python`, and it is exactly what I am using right now to get this information in one shot. This is, however, far from optimal, since the simple API lacks much more information available in JSON, and I need maintain two independent parsers for the job.

I don’t particularly love this solution, as mentioned above, since it generates duplicate information (every URL in a given release, instead of one per release), and is hierarchically incorrect. and would happily accept if my need can be addressed in another way.

dstufft:
Thanks for this PR, while I agree that this is sort of an awkward place to have it, it seems like a reasonable change to me and I can't think of a better way to handle it that doesn't involve even more drastic changes. Given that, I'm going to go ahead and merge this.

Thanks again!

PR #4328, opened 2018-07-20
Incremental search index updates 2.0

ewdurbin:
Attempt 2 at #701, restores changes from #3693.

With #3989 showing good results over the past few months, this is worth another shot :)


dstufft:
We probably want to run a full index less often if we have incremental indexing?

PR #4358, opened 2018-07-27
Upgrade js

dstufft:


nanuxbe:
I had the same problem with javascript.

Restarting `make serve` didn't seem to work but launching a shell on the `static` container and launching gulp by hand worked

PR #4368, opened 2018-07-28
Update troubleshooting documentation for first build

acangiani:


acangiani:
Hi!
 
Yes, I'm actually using Ubuntu, I could not test this issue on another SO, at least here on the sprint this error was not reported on another SO. Should I add that disclaimer?

di:
@acangiani Yes, something like "if you're using Ubuntu and you experience ... you should do ...", so that macOS/Windows users don't try to tun this command.

acangiani:
@di I had added the specific reference to the SO, if there is any other feedback please do,

thanks!

PR #4369, opened 2018-07-28
Migrate friction log template from google docs

mferi:
Migrate friction log template from google docs
https://docs.google.com/document/d/1kcEv9c_diZbqdXk2VkVu6miHz62hUoxxzfIo5ILnESM/edit#


nlhkabu:
fyi @brainwane.  We've added this here, to be able to track these within the issue tracker :)  Maybe it's worth adding some documentation at the beginning to explain when/why they should be used?

brainwane:
Thanks! Indeed it might be -- I don't have time to put in the issue or add that documentation, though!

PR #4370, opened 2018-07-28
Document memory requirements for Linux in the Getting started guide 

fitodic:
Connected to issues:
    
- https://github.com/pypa/warehouse/issues/1511
- https://github.com/pypa/warehouse/issues/2811

### DIFF
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -143,6 +143,18 @@ increase the memory allocated to Docker in
 or `Docker Settings <https://docs.docker.com/docker-for-windows/#advanced>`_
 (on Windows) by moving the slider to 4 GB in the GUI.
 
+If you are using Linux, you may need to configure the maximum map count to get
+the `elasticsearch` up and running. According to the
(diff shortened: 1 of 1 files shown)


fitodic:
@nlhkabu As stated by @pradyunsg in https://github.com/pypa/warehouse/issues/1511#issuecomment-358700810, the maximum map count needs to be configured to get the `elasticsearch` service up and running on Linux. Since the documentation only mentions Windows and Mac workarounds, I thought it would be useful to add a Linux workaround as well.

rivol:
I ran into the same problem and can confirm this is useful.

Maybe change the docs link to point to the specific version used by Warehouse though (ATM it points to 2.x docs which are old (though the instructions themselves are the same)): https://www.elastic.co/guide/en/elasticsearch/reference/6.2/vm-max-map-count.html

nlhkabu:
@rivol could you add your suggested changes and make a new PR?

rivol:
@nlhkabu I could, but it is a very small change...

I guess either @fitodic could change it within this PR, or the current PR could be merged and then I can make another one for the link change, to avoid conflicts. Let me know which approach you prefer.

PR #4371, opened 2018-07-28
Update docker-compose.yml to use concrete versions of redis and rabbitmq.

hbrylkowski:
In docker-compose file two images were pulled from tag `latest`. As this may result in potential undetermined behavior when there are introduced breaking changes in given images I've changed it to corresponding minor version tag. 

There is also `warehouse-camo` but it only has `latest` tag


rivol:
FWIW: LGTM (no Warehouse experience but Docker experience)

Issue #996, opened 2016-03-02
Add support for two-factor authentication


edmorley:
Account compromise of the owner of a popular package on PyPI is something that would have pretty dire consequences. 2FA/MFA would be a big step to helping avoid this.

This would need to be implemented in both:
- Website login
- API used for package upload/register by setuptools/pip/... 

The 1st of these is presumably the easiest, and I believe will still add some value even if the 2nd isn't implemented initially - since it will still prevent things like package owner/maintainer role changes or modification of a user's listed GPG key or login details (presuming these cannot be made via the API).

Harder is what to do with the API, given it requires client changes and is also used by some in a non-interactive manner in automation. I see a few options:
1. Add 2FA support to setuptools/pip, and just document that people uploading packages in automation will just need to disable 2FA for their whole account (or ideally create a separate account just for automation).
2. Add support for API keys to both Warehouse and clients (#994) and allow those to circumvent 2FA. For the people that really must upload packages non-interactively this would still protect against a few risks (credentials re-use, weak credentials), and the API keys could also be made package-specific or tied to a particular IP range. For everyone else, they would just not create an API key and still have full 2FA protection.
3. All of the above.

In terms of implementation, there are a number of packages that support TOTP (used by Google Authenticator):
https://pypi.python.org/pypi/onetimepass/1.0.1
https://pypi.python.org/pypi/pyotp
https://pypi.python.org/pypi/oath
https://pypi.python.org/pypi/otpauth

(I've filed this against warehouse since I'm presuming this is beyond the scope of maintenance-only changes being made to the old PyPI codebase)

dstufft:
I do plan on adding 2FA to Warehouse, though it will be a post-launch task and will likely require deprecating the current upload API (which I want to do anyways). As far as what technology we use to handle it, TOTP is an obvious shoe-in. I'd like to explore some options for services that handle 2FA authentication though and see if any of them would donate service (assuming the export story to allow us to go away from their service in the future is sane). I haven't done a ton of exploring in this idea yet, but it would be nice if we could support more than just TOTP (particularly, something like Authy's OneTouch would be really nice).

ryanhiebert:
FWIW, I've been impressed with U2F, which is supported by GitHub and Google currently, of the sites I use regularly. Built-in support for Chrome, and an add-on to support it in Firefox, and MS has registered their intent to implement it as well.

TOTP is still good, since it's much more available, but I'd love to see U2F supported as well.

edmorley:
[...]

PR #4373, opened 2018-07-28
[WIP] #996 Add two-factor authentication

asfaltboy:
EuroPython 2018 collaborators working on #996


di:
Thanks for the PR! I think we'll probably want to hold off on merging this until we have the rest of #996 implemented as well.

Sparkycz:
Printscreen of the verification page http://prntscr.com/kc4x8e

Sparkycz:
@asfaltboy Could you please edit the title of the PR? This one is a little bit confusing.. Maybe to `#996 add two-factor authentication`

asfaltboy:
Good point @Sparkycz done that!

Sparkycz:
I think we should add one column more into the table AccountUser to indicate that the user has enabled two-factor auth.

Something like..
```
op.add_column(
    "accounts_user", sa.Column("2fa_enable", sa.Boolean(), nullable=False, default=False)
)
```

However, i'm still working on the login presenters..

sergeykolosov:
@Sparkycz I'd like to make a number of clean-ups, commit squashes, etc, which basically involves rewriting branch history. Please notify me when it's an appropriate moment to do so, so that it won't conflict with your work on presenters.

sergeykolosov:
Updated the branch as follows:
* fixed requirements ordering;
* renamed OTP secret to TOTP secret (to avoid collisions as soon as other OTP methods are added);
* fixed migration not being regenerated after field length change;
* reordered and squashed commits into logically consistent groups;
* implemented TOTP primitives and documented design decisions around those (see docstings);
* covered TOTP primitives with basic unit-tests.

@nlhkabu suggested teaming up with others involved in #996 to avoid putting effort into duplicate artifacts.

sergeykolosov:
@Sparkycz please address unittest & linting issues.

pradyunsg:
Could someone add the EuroPython label on this? :)

sergeykolosov:
Those who had already applied DB migrations, before pulling the updated branch, make sure to downgrade first:
```
docker-compose run web python -m warehouse db downgrade 263b37556f74-1
```

dstufft:
FWIW, we do ``git merge --squash``, so you don't have to worry at all about keeping a clean history.

warehouse/utils/otp.py
```diff
@@ -0,0 +1,59 @@
...
+    https://tools.ietf.org/html/rfc4226#section-4
+
+    Given the base32 encoding, it is represented as a 32-character string, i.e.:
+    5 bits per character * 32 characters = 160 bits.
+
+    TODO:
+    consider using os.urandom(); this, however, is likely to cause
+    compatibility issues unless base32 vocabulary is used.
+    """
+    return pyotp.random_base32(length=32)
```
dstufft: Looking at this function, it's not actually being base32 encoded, it's just generating a 16 length random string which just happens to also be valid base32. If this just needs a be a base32 encoded string, then I'd prefer to do something like:

```python
import base64
import os

def generate_totp_secret():
    return base64.b32encode(os.urandom(32)).decode("ascii")
```

This matches how we've done things in ``warehouse/utils/crypto.py``.

warehouse/utils/otp.py
```diff
@@ -0,0 +1,59 @@
...
+        * 30-second interval
+    """
+    return pyotp.TOTP(secret).provisioning_uri(username, issuer_name=issuer_name)
+
+
+def verify_totp(secret, value, valid_window=1):
+    """
+    Verifies a given TOTP-secret and value.
+
+    The *valid_window* argument value is intentionally chosen to be non-zero
```
dstufft: Can we add a note here as to what exactly the valid window translate into? Is it an extra second? 6?, 30?

Sparkycz:
I amended fix of the routes test into the commit with template. However there is still output from `black` linter (would reformat .. ) but i wouldn't like to reformat before the end of the changes in acounts/views.py

Meanwhile i'm finishing first version of the presenters (1-2 unit-tests left) despite We'll have to discuss the behavior of sessions etc..

Sparkycz:
i guess you'll have a lot of requests for changes but somewhere we've gotta start ;-) 

pls check todos:
1. add enable/disable option of 2FA to profile settings
2. implement calling of utils/otp functions to user_service

woodruffw:
I've restarted work on this in https://github.com/pypa/warehouse/compare/master...trailofbits:tob/996-add-2fa-support. Will open a PR shortly so that others can review progress in real time.

di:
Closing in favor of #5567, but thank you for your contribution!

PR #4374, opened 2018-07-28
Change the colors of keyword, keyword name and function name

### LINKED ISSUE #4047: Visual Design: swap syntax highlighting colors
Not a biggie, but keywords being highlighted in blue is more common. If swapped with the integer color, keywords could be blue and numbers green.

Believe that would match the site colors better as well, as the keywords stand out a lot.

shot

Thanks, awesome job.

nividonis:
Closes #4047

### DIFF
--- a/warehouse/static/sass/blocks/_project-description.scss
+++ b/warehouse/static/sass/blocks/_project-description.scss
@@ -194,7 +194,7 @@ $black:         #000;
   }
 
   .k { /* Keyword */
-    color: $green;
+    color: $darkest-blue;
     font-weight: bold;
   }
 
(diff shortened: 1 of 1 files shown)


nividonis:
@nlhkabu Is this ok? 
screenshot from 2018-07-28 14-35-53

nlhkabu:
Thanks again for this PR @nividonis 

If you'd like to continue on this area, there are two more issues open associated with improving the code highlighting theme:

https://github.com/pypa/warehouse/issues/4400
https://github.com/pypa/warehouse/issues/3939

Let me know if you'd like to work on either of these!

nividonis:
@nlhkabu Sorry for the late response. I see that someone has already taken those two issues, but I will definitely find something I like when my schedule clears up

PR #4380, opened 2018-07-28
Without JS, show default navbar and notification that some things might be broken

cpulvermacher:
This PR does includes two things:
- Showing the logged-out version of the navbar in case CSI fails, so users can log in without JavaScript. 
- Show a notification at the top if JS is disabled, and disable the fixed positioning to avoid hiding the menu bar.

Let me know if this can be improved, in particular I'm not entirely sure if the way I've imported the macro is OK.

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -150,6 +150,15 @@
         <span class="notification-bar__message">You are using TestPyPI – a separate instance of the Python Package Index that allows you to try distribution tools and processes without affecting the real index.</span>
       </div>
       {% endif %}
+      <noscript>
+      <div class="notification-bar notification-bar--danger">
+        <span class="notification-bar__icon">
+          <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+          <span class="sr-only">Warning:</span>
+        </span>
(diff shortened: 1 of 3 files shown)


warehouse/templates/base.html
```diff
@@ -141,6 +150,15 @@
...
       {% endif %}
+      <noscript>
+      <div class="notification-bar notification-bar--danger">
+        <span class="notification-bar__icon">
+          <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
+          <span class="sr-only">Warning:</span>
+        </span>
+        <span class="notification-bar__message">Some features may not work without JavaScript. Please try enabling it if you encounter problems.</span>
+      </div>
+      </noscript>
```
di: If we add this, can you add a TODO comment to remove it once https://github.com/pypa/warehouse/issues/3828 is resolved (and leave a comment on that issue about it)? Everything _should_ work without JavaScript enabled.

cpulvermacher: Will do. Most things work already, but I suspect there are some more subtle bits that might take a bit longer.

di: Indeed, I left a comment on https://github.com/pypa/warehouse/issues/3828 about one in particular.

PR #4383, opened 2018-07-28
Add a badge to pre-release releases

### LINKED ISSUE #4118: Visually mark pre-release versions in "Release history" view

Identifying whether the version in list is going to be installed by `pip` without `--pre` CLI arg.

Some label/badge next to version title in different color with text like `pre-release` or `pre`.

**Additional context**

When you look at https://pypi.org/project/aiohttp/3.3.0a0/#history it's impossible to visually distinguish between stable and pre-release versions. Of course if one knows how it works (`a0` is for alpha and isn't installed unless `--pre` arg is used in `pip install`), but I believe that the average user doesn't know this.
Even if they know it is still much more friendly to have some visual signal for this.

rixx:
Closes #4118. As pre-releases aren't installed via pip unless explicitly
included, distinguishing them visually is helpful to avoid confusion for
everybody who is not familiar with pre-release naming schemes.

screenshot_2018-07-28 pyownet 1


warehouse/templates/packaging/detail.html
```diff
@@ -210,7 +210,10 @@ <h2 class="page-title split-layout">
               </div>
 
               <a class="card release__card" href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}">
-                <p class="release__version">{{ hversion.version }}</p>
+                <p class="release__version">
+                  {{ hversion.version }}
+                  {% if hversion.is_prerelease %}<span class="badge badge--warning">pre-release</span>{% endif %}
```
nlhkabu: To follow BEM convention, we should add a class here called `release__version-badge` and add it into the SCSS.

warehouse/static/sass/blocks/_release.scss
```diff
@@ -101,6 +101,10 @@
     padding-bottom: 0;
     font-size: 20px;
     font-weight: $bold-font-weight;
+
+    .badge {
+      margin-left: 4px;
+    }
```
nlhkabu: As below, we should instead add a class called `release__version-badge` and add it as `&__version-badge` just above `&__version-date`

rixx:
Thank you for your detailed actionable feedback, I just pushed the requested changes.

nlhkabu:
@rixx looks like we have a linting error :( Could you take a look?

PR #4384, opened 2018-07-28
Add GitLab Statistics

am-on:
Added star and fork statistics on detailed project view for projects with an
valid GitLab URL.

Closes https://github.com/pypa/warehouse/issues/4301
Ref: https://github.com/pypa/warehouse/issues/4301


am-on:
image

di:
We could also probably make that section more generic, since it's unlikely that a project has both GitHub and gitlab links.

Also, we should probably prefer one over the other here so we don't show approximately the same section twice (I'd lean towards preferring GitHub).

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -54,6 +54,26 @@ <h3 class="sidebar-section__title">Statistics</h3>
     </a>
   </div>
   {% endif %}
+
+  {% if release.gitlab_repo_info_url %}
+  <div class="gitlab-repo-info hidden" data-url="{{release.gitlab_repo_info_url}}">
+    <em>GitLab statistics:</em>
+    <a class="vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed gitlab-repo-info__item"
+       data-key="html_url" data-attr="href" data-supplement="/stargazers" rel="noopener"
```
am-on: change/remove `data-supplement="/stargazers"`

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -54,6 +54,26 @@ <h3 class="sidebar-section__title">Statistics</h3>
...
+       target="_blank">
+      <i class="fa fa-star" aria-hidden="true"></i>
+      <strong>Stars: </strong>
+      <span class="gitlab-repo-info__item" data-key="star_count"></span>
+    </a>
+    <a class="vertical-tabs__tab vertical-tabs__tab--with-icon vertical-tabs__tab--condensed gitlab-repo-info__item"
+       data-key="html_url" data-attr="href" data-supplement="/network" rel="noopener"
```
am-on: change/remove `data-supplement`

warehouse/static/sass/blocks/_vertical-tabs.scss
```diff
@@ -54,12 +54,13 @@
...
-    &:hover,
-    &:focus {
-      text-decoration: none;
-      color: darken($primary-color, 10)
+    @at-root {
+      a#{&}:hover,
+      a#{&}:focus {
+        text-decoration: none;
+        color: darken($primary-color, 10)
+      }
```
di: It's not clear to me what these changes are for, could you explain?

TheCodersHangout:
Cool! Waiting for it!

warehouse/static/sass/blocks/_vertical-tabs.scss
am-on: Statistic items don't have a link in all cases (GitLab star view is not implemented yet). To make it clear `span` is used instead of `a`, but css was still changing span element on hover. The code above apply css only for `a`.

The code above is compiled to `a.<class>:hover, ...`

nlhkabu: Could we do this instead by simply removing this line: 
https://github.com/pypa/warehouse/blob/master/warehouse/static/sass/blocks/_vertical-tabs.scss#L57

?

am-on: No, in that case hovering over statistic item without link - `span` would still get same hover styling as links. 

If you look at screenshot https://github.com/pypa/warehouse/pull/4384#issuecomment-408613727 hovering over `Stars` make no difference to its styling, without that addition, hovering over `Star` would change the text color to blue.

nlhkabu:
@am-on - what project were you testing this with in development?

am-on:
@nlhkabu not sure which is gitlab and which github

pymagnitude
krangpower

paternal:
This is nice. :smile: 

However, Gitlab being free, there are several instances (for instance, [my projects](https://pypi.org/project/PdfAutoNup/) are hosted at [Framagit](http://framagit.org/spalax/pdfautonup)). While having warehouse check a list of every single public gitlab instance is a bad idea, I wonder if it would be possible to:
1. Check if the project homepage URL is a gitlab instance;
2. If so, fetch and display its statistics.

Thanks for your work!

nlhkabu:
Thanks again for this PR @am-on - and sorry for the delay on getting back to it!

I've just pushed a PR that separates out the CSS classes from the JS class (the JS hook is now `js-repo-data`).

I've also updated the template to be (what I think is) more readable, and adapted the SCSS accordingly.

screenshot from 2018-11-18 13-55-58

screenshot from 2018-11-18 13-55-08

@yeraydiazdiaz would you be able to re-review the JS for this PR?  There is an empty `catch` function at the bottom of `repository-info.js` that I am not sure about :)

warehouse/packaging/models.py
```diff
@@ -384,6 +384,17 @@ def github_repo_info_url(self):
...
+    @property
+    def gitlab_repo_info_url(self):
+        for parsed in [urlparse(url) for url in self.urls.values()]:
+            segments = parsed.path.strip("/").rstrip("/").split("/")
+
+            if (
+                parsed.netloc == "gitlab.com" or parsed.netloc == "www.gitlab.com"
+            ) and len(segments) >= 2:
+                user_name, repo_name = segments[:2]
+                return f"https://gitlab.com/api/v4/projects/{user_name}%2F{repo_name}"
```
yeraydiazdiaz: this property has a lot of common code with the above. It's not a blocker for the PR, but it'd be nice if we could refactor the common logic into a helper function.

brainwane:
@am-on would you be interested in finishing this up?

Also, some of us will be at the PyCon North America sprints again this year, May 6th-9th, in case you would like to join us then!

brainwane:
@am-on Is this ready to be re-reviewed?

nlhkabu:
I believe this is waiting on another HTML and SCSS review on my side (and possible updates).

nlhkabu:
Note: 

yeraydiazdiaz:
@am-on would you be interested in continuing work on this? Looks like it was pretty close to merging.

miketheman:
This is still. a good idea, but due to age, too many conflicts, closing this PR. If there's still interest, please revive.

Issue #3828, opened 2018-04-26
Audit templates - what does not work without JavaScript


nlhkabu:
Following an email from a user who prefers to browse the web with JS off - can we please audit the codebase to find those places where the user cannot acheive their goals?

The user mentioned checksums in particular.

di:
This is due to the client-side include failing (which requires Javascript).

It should be possible to add some "default" HTML for the logged-out user which would be replaced if the CSI is successful here: https://github.com/pypa/warehouse/blob/master/warehouse/templates/base.html#L184-L185

Probably worth mentioning that this same HTML is currently inside the include as well, so this shouldn't be duplicated twice and instead should become a macro probably: https://github.com/pypa/warehouse/blob/master/warehouse/templates/includes/current-user-indicator.html#L77-L82

di:
Another thing to audit here is the click-to-copy feature for file hashes on the "Files" page, I suspect it doesn't work at all without JS enabled.

PR #4390, opened 2018-07-28
WIP: Show popup with hash instead of copying directly to clipboard

cpulvermacher:
Also for #3828 .

Note that closing the popup via Esc is broken. 
On the manage release page, it works by accident as there are two other possible popups that set the data-controller attribute and thus work with https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/index.js#L129 (it still throws some TypeErrors). 

Any ideas on how to properly fix the key handler? Or should everything use the data-controller?

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -247,11 +247,44 @@
                       {{ file.filename }}
                     </a>
                     {% if file.size %} ({{ file.size|filesizeformat() }}){% endif %}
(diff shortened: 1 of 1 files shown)


di:
I think anything that is "closable" could use that same controller, and the controller should just close anything open on the page, so that's the approach I'd recommend.

Once you get this working, can you post a screenshot (or better yet, a GIF is possible)?

di:
Closing due to inactivity & merge conflicts -- @cpulvermacher, if you're still interested in fixing this, feel free to make a new PR.

PR #4396, opened 2018-07-28
Add sole owner badge on lists of projects

### LINKED ISSUE #3127: Add 'sole owner' badge on list of projects
We've had a lot of positive feedback on the bus factor feature on the account settings page.

It would be nice to also expose this information on `/manage/projects/` so that users don't have to hunt for it :)

gabrielgradinaru:
closes #3127

### DIFF
--- a/warehouse/manage/views.py
+++ b/warehouse/manage/views.py
@@ -279,19 +279,41 @@ def manage_projects(request):
             return project.releases[0].created
         return project.created
 
-    projects_owned = set(
-        project.name
-        for project in (
-            request.db.query(Project.name)
-            .join(Role.project)
-            .filter(Role.role_name == "Owner", Role.user == request.user)
-            .all()
-        )
+    projects_owned = (
+        request.db.query(Project)
+        .join(Role.project)
+        .filter(Role.role_name == "Owner", Role.user == request.user)
+        .subquery()
     )
+
+    with_owner_count = (
+        request.db.query(
+            Role.package_name, func.count(Role.package_name).label('owner_count')
+        )
+        .join(projects_owned)
+        .filter(Role.role_name == "Owner")
+        .group_by(Role.package_name)
+        .subquery()
+    )
+
+    query_result = (
+        request.db.query(Project.name, with_owner_count.c.owner_count)
+        .join(with_owner_count)
+        .all()
+    )
+
+    projects_owned = set()
+    projects_sole_owned = set()
+
+    for project, owner_count in query_result:
+        projects_owned.add(project)
+        if owner_count == 1:
+            projects_sole_owned.add(project)
 
     return {
         "projects": sorted(request.user.projects, key=_key, reverse=True),
         "projects_owned": projects_owned,
+        "projects_sole_owned": projects_sole_owned,
     }
 
 
--- a/warehouse/templates/manage/projects.html
+++ b/warehouse/templates/manage/projects.html
@@ -27,7 +27,11 @@
     <div class="package-snippet">
       <div class="split-layout split-layout--table split-layout--wrap-on-tablet">
         <div>
-          <h3 class="package-snippet__title">{{ project.name }}</h3>
+          <h3 class="package-snippet__title">{{ project.name }}
+            {% if project.name in projects_sole_owned %}
+            <span class="badge">SOLE OWNER</span>
+            {% endif %}
+          </h3>
           {% if release %}
           <p class="package-snippet__meta">
             <em>Last released on {{ release.created|format_date() }}</em>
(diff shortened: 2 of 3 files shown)


di:
Thanks @gabrielgradinaru! Could you add a screenshot?

gabrielgradinaru:
This is what it looks like:
[
screen shot 2018-07-29 at 11 50 14
](url)
And I will soon try to fix the CI.

gabrielgradinaru:
CI fixed.
@di , @nlhkabu can you take a look at this pull request?

warehouse/templates/manage/projects.html
```diff
@@ -27,7 +27,11 @@ <h1 class="page-title">Your projects ({{ projects|length }})</h1>
     <div class="package-snippet">
       <div class="split-layout split-layout--table split-layout--wrap-on-tablet">
         <div>
-          <h3 class="package-snippet__title">{{ project.name }}</h3>
+          <h3 class="package-snippet__title">{{ project.name }}
+            {% if project.name in projects_sole_owned %}
+            <span class="badge">SOLE OWNER</span>
```
nlhkabu: I don't think we need this to be "SOLE OWNER" as we have a CSS style that will make it uppercase.  It should instead be "Sole owner".

It would also be great, if we could put a 4px left margin on this.  If you could follow the same pattern as shown in this pr: https://github.com/pypa/warehouse/pull/4383/files

... that would be great :)

Let me know if this isn't clear!

gabrielgradinaru: Done, it now looks like this:
screen shot 2018-07-29 at 13 35 08

nlhkabu:
thanks @gabrielgradinaru. Could you please provide an updated screenshot?

pradyunsg:
@nlhkabu https://github.com/pypa/warehouse/pull/4396#discussion_r205968612

warehouse/templates/manage/projects.html
nlhkabu: great :) could you also take a screenshot on a smaller resolution (mobile) ?

gabrielgradinaru: Sure, on mobile resolution it's like this:
screen shot 2018-07-29 at 14 03 05

nlhkabu:
Ok, great. Looks good on my side.  The badge is not wrapping ideally on mobile, but we can open a separate ticket to address this, as there are several other situations where this is happening.

@di can you please review the Python code here?

warehouse/manage/views.py
```diff
@@ -279,19 +279,41 @@ def _key(project):
...
-            request.db.query(Project.name)
-            .join(Role.project)
-            .filter(Role.role_name == "Owner", Role.user == request.user)
-            .all()
+    projects_owned = (
+        request.db.query(Project)
+        .join(Role.project)
+        .filter(Role.role_name == "Owner", Role.user == request.user)
+        .subquery()
+    )
```
di: I'd prefer if the `projects_owned` variable wasn't being overwritten later on in this function, perhaps this could be called "user_projects_query" or something like that?

warehouse/manage/views.py
```diff
@@ -279,19 +279,41 @@ def _key(project):
...
+        .filter(Role.role_name == "Owner")
+        .group_by(Role.package_name)
+        .subquery()
     )
 
+    query_result = (
+        request.db.query(Project.name, with_owner_count.c.owner_count)
+        .join(with_owner_count)
+        .all()
+    )
```
di: I think we could probably add to this query to remove the need for the `for` loop below, using a `groupby` and something like `.having(func.count() == 1)` then calling `.all()` and just setting it to `projects_sole_owned` directly. (Disclaimer, that might not work)

warehouse/manage/views.py
gabrielgradinaru: Changed `projects_owned ` to `user_projects_query `

warehouse/manage/views.py
gabrielgradinaru: The for loop can surely be removed easily but I would have to perform 2 queries (which I'm trying to avoid).
Right now I'm performing a single query to get info for both "projects_owned" and "projects_sole_owned".
Instead of the for loop I could make 2 set comprehensions instead, something like this:
```
projects_owned = set(project for project, _ in query_result)
projects_sole_owned = set(
    project for project, owner_count in query_result if owner_count == 1
)
```
But in my opinion it's not so readable.

From what I've researched, I did not find a way to make your suggestion work :(
Let me know what you this is the best way to proceed.

gabrielgradinaru: @di just a friendly reminder for this pull request :)

di: Thanks, I will try to investigate this a little more to figure out if the query I'm suggesting is actually possible or not.

Issue #3626, opened 2018-04-11
RFE: display date in project list


larsks:
The project list in the Warehouse allows sorting projects by "Date last updated". Unfortunately, that date isn't displayed in the list itself, which reduces the utility of the sorting. It would be nice to differentiate between "these projects have all been updated recently" vs "the most recent project was udpated last year and the rest more than three years ago".  This helps filter out stale results.

brainwane:
Thanks for the suggestion, @larsks!

ewdurbin:
Is this resolved now?

nlhkabu:
Yes :)

PR #4402, opened 2018-07-29
RFE: display date in project list #3626 (pypa/warehouse#3626)

rok:
Proposed solution for [#3626](https://github.com/pypa/warehouse/issues/3626), adding human readable time since last release to package search view.

Would appreciate guidance on how to read time as datetime (currently string) from the elasticsearch in [warehouse/views.py](https://github.com/pypa/warehouse/blob/master/warehouse/views.py#L311).

Collaborators: @AyalaO, @MiHess


nlhkabu:
Thanks @rok, @AyalaO and @MiHess!

Could you please post some screenshots (desktop, tablet, mobile) of this?

nlhkabu:
morning @pradyunsg!

Would you be able to provide guidance on "how to read time as datetime (currently string) from the elasticsearch in warehouse/views.py." ?

pradyunsg:
Hi @nlhkabu!

I'm not familiar enough with the Warehouse codebase to help out here and I'm a little short on time currently due to some academic workload to take a deeper look here. Sorry! :(

nlhkabu:
ok, no problem @pradyunsg, just saw you were about :P I've also assigned @di so we should be covered here :)

pradyunsg:
Thanks! ^>^

All the best for the sprints! Based on the activity here, it seems to be going well. :)

rok:
@nlhkabu Thanks! :smile: Attaching some screenshots.
display_date_desktop
display_date_mobile
display_date_tablet

warehouse/templates/search/results.html
```diff
@@ -21,6 +21,7 @@
     <h3 class="package-snippet__title">
       <span class="package-snippet__name">{{ item.name }}</span>
       <span class="package-snippet__version">{{ item.latest_version }}</span>
+      <span class="package-snippet__released">{{ humanize(item.created_timestamp) }}</span>
```
di: Can we not just call `humanize` on `item.created`? E.g.:

```
<span class="package-snippet__released">{{ humanize(item.created) }}</span>
```

di:
I think it would be nice to only show the date if the user is sorting by date, otherwise the timestamp seems a little meaningless to the user (i.e., it could be the date the project was created, the last release, etc.)

rok:
@di - thanks for the feedback! Removed the item.last_updated and used item.created. Timestamps should now only when sorting on date last update.
A question about item.created: elasticsearch query returns a timestamp string and we must cast it in order for humanize to work. Is there a way to read it as timestamp directly to avoid this casting?

di:
> A question about item.created: elasticsearch query returns a timestamp string and we must cast it in order for humanize to work. Is there a way to read it as timestamp directly to avoid this casting?

I don't think so, we are probably also expecting this to be a string in other places. One thing we could do to make this a little cleaner is to add the cast as a property function on [this class](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/packaging/search.py#L34) instead.

rok:
> I don't think so, we are probably also expecting this to be a string in other places. One thing we could do to make this a little cleaner is to add the cast as a property function on this class instead.

I've tried adding a property to [Project](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/packaging/search.py#L34) class like so:
```
@property
def created_timestamp(self):
    return datetime.strptime(self.created, "%Y-%m-%dT%H:%M:%S.%f")
```
but that didn't seem to add a new attribute to instances of <class 'elasticsearch_dsl.response.hit.Hit'>. Is this what you suggested?

chfw:
this feature is useful for me as a project owner. sometimes, I forget when I have released my package since I have 40+ packages. So I resorted to look at pypi for more info. However, github releases and pypi at moment both love moment.js and also give some time ago. Still I could not find the exact date when a version is released.

rok:
@di - It seems to me that the [search](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/views.py#L244) is not using the [Project](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/packaging/search.py#L34) model, but rather uses [elasticsearch_dsl](https://github.com/elastic/elasticsearch-dsl-py) to search directly. I would guess that is either for flexibility reasons or using the model has simply not been implemented yet. Elasticsearch_dsl is returning the created field as a string. It would be interesting to know why this was made this way.

I think we have two options here:

1. Cast the created field to datetime on the fly as we originally proposed.
2. Refactor [search](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/views.py#L244) to use the [Project](https://github.com/pypa/warehouse/blob/aca70dbc88bf00234777b7afb324e42b160d8fdd/warehouse/packaging/search.py#L34) model.

What do you think? How would such a decision normally be made?

Thanks for your input :)

di:
Hi @rok, sorry for the delay here. I think casting it on the fly is probably fine!

rok:
Hey @di! No worries. Shall we merge it then?
I suppose if more cases like this appear a more general mechanism should be made.

nlhkabu:
Thanks @rok, @AyalaO and @MiHess!

ewdurbin:
Reverting this for the moment, something is going awry with the strptime being applied to results from elastic search:

```ValueError: time data '2008-12-14T10:22:11' does not match format '%Y-%m-%dT%H:%M:%S.%f'```

full traceback:

```
HTTPNotFound: The resource could not be found.
  File "pyramid/tweens.py", line 13, in _error_handler
    response = request.invoke_exception_view(exc_info)
  File "pyramid/view.py", line 769, in invoke_exception_view
    raise HTTPNotFound
ValueError: time data '2008-12-14T10:22:11' does not match format '%Y-%m-%dT%H:%M:%S.%f'
  File "warehouse/raven.py", line 40, in raven_tween
    return handler(request)
  File "pyramid/tweens.py", line 43, in excview_tween
    response = _error_handler(request, exc)
  File "pyramid/tweens.py", line 17, in _error_handler
    reraise(*exc_info)
  File "pyramid/compat.py", line 179, in reraise
    raise value
  File "pyramid/tweens.py", line 41, in excview_tween
    response = handler(request)
  File "warehouse/cache/http.py", line 75, in conditional_http_tween
    response = handler(request)
  File "warehouse/sanity.py", line 76, in sanity_tween_egress
    return unicode_redirects(handler(request))
  File "pyramid/router.py", line 148, in handle_request
    registry, request, context, context_iface, view_name
  File "pyramid/view.py", line 657, in _call_view
    response = view_callable(context, request)
  File "warehouse/sessions.py", line 299, in wrapped
    return view(context, request)
  File "warehouse/csrf.py", line 39, in wrapped
    return view(context, request)
  File "pyramid/viewderivers.py", line 514, in csrf_view
    return view(context, request)
  File "warehouse/cache/origin/__init__.py", line 85, in wrapped
    return view(context, request)
  File "warehouse/cache/origin/derivers.py", line 31, in wrapper_view
    return view(context, request)
  File "warehouse/metrics/views.py", line 34, in wrapper_view
    return view(context, request)
  File "pyramid/viewderivers.py", line 436, in rendered_view
    result = view(context, request)
  File "pyramid/viewderivers.py", line 144, in _requestonly_view
    response = view(request)
  File "warehouse/views.py", line 357, in search
    item.created = datetime.strptime(item.created, "%Y-%m-%dT%H:%M:%S.%f")
  File "python3.6/_strptime.py", line 565, in _strptime_datetime
    tt, fraction = _strptime(data_string, format)
  File "python3.6/_strptime.py", line 362, in _strptime
    (data_string, format))
```

nlhkabu:
Thanks @ewdurbin - is this error something we can add a unit test for?

ewdurbin:
Turns out some of our oldest records are with timestamps formatted as `%Y-%m-%dT%H:%M:%S` rather than `%Y-%m-%dT%H:%M:%S.%f`, so if a search result has releases before that changed we get this exception.

The reason this didn't show up locally is that our dev db is based on TestPyPI which was reset most recently _after_ the change that began storing the `.%f` part.

As far as tests for this go within warehouse, it's a fairly straightforward set of cases for the return of values from this view. The data translation being done should probably fail safe anyway rather than blow up.

PR #4403, opened 2018-07-29
Several style guideline related issues #3167

mferi:
https://github.com/pypa/warehouse/issues/3167

Rename "Security policy" to "Report security issue"
Remove all ampersands in the codebase
Change filter list titles to 'By licence', 'By environment', etc.
Change the link in the documentation to https://web.archive.org/web/20180410101124/https://material.io/guidelines/style/writing.html

### DIFF
--- a/docs/ui-principles.rst
+++ b/docs/ui-principles.rst
@@ -61,7 +61,7 @@ help content).
 -------------------------------------------------------
 
 Warehouse follows the `Material design writing style guide
-<https://material.io/design/usability/accessibility.html#writing>`_.
+<https://web.archive.org/web/20180410101124/https://material.io/guidelines/style/writing.html#>`_.
 
 When writing interfaces use direct, clear and simple language. This is
 especially important as Warehouse caters to an international audience with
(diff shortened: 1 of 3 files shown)


docs/ui-principles.rst
```diff
@@ -61,7 +61,7 @@ help content).
 -------------------------------------------------------
 
 Warehouse follows the `Material design writing style guide
-<https://material.io/design/usability/accessibility.html#writing>`_.
+<https://web.archive.org/web/20180410101124/https://material.io/guidelines/style/writing.html>`_.
```
di: Why did this link change? Looks like https://material.io/design/usability/accessibility.html#writing still exists to me.

pradyunsg: IIRC, this was the link originally and then this was changed when material design got an update, which broke the links.

Maybe we should follow the latest version of Material on this; but that's a moving target so, choosing an archived version also makes sense.

nlhkabu: As far as I can tell, they have reduced the guidelines in the newer version, which is why I asked to change to the archived link.  I agreewith @pradyunsg that it makes sense to pin to a particular version.  Another option would be to add this documentation into our own docs, so that we were not affected by future changes, or could make changes to the guidelines as/when we needed.

nlhkabu: ticket opened: https://github.com/pypa/warehouse/issues/4422

PR #4410, opened 2018-07-29
Make the xz decompress to stdout and psql to eat the SQL from stdin, so that no unnecessary temporary disk space is required.

ztane:
I ran out of disk space in the middle of `make initdb` in my home directory. `psql` should be able to eat the source straight from standard input and this works for *me* with my docker-compose and Linux; I am not so sure about other platforms.

This change avoids writing to a temporary file; it would also make sure that no uncompressed file would be left to linger on disk, should `psql` fail for some reason.


ewdurbin:
Thanks for your contribution @ztane! Confirmed this works well on OS X as well.

PR #4412, opened 2018-07-29
Add warning about docker version for linux users

nanuxbe:
This simple warning would have saved me a lot of time, I think it's worth adding.


di:
Hmm, https://github.com/pypa/warehouse/pull/4368 just added a recommendation to install from a snap, perhaps this is bad advice? I'm not really familiar with the best way to get Docker on Linux.

nlhkabu:
@dstufft or @ewdurbin could you please advise on this PR? Thx :)

ewdurbin:
I don’t have enough experience using Docker on Linux Desktop/Workstation to assess.

di:
Closing due to lack of response here. If more users experience this we can revisit.

PR #4421, opened 2018-07-31
Support using SQS instead of RabbitMQ for a Celery Broker

dstufft:
This will let Warehouse *optionally* use SQS as a Celery broker in addition to RabbitMQ. This changes the broker configuration from ``AMQP_URL`` to ``BROKER_URL`` to reflect the fact it supports either option now.

Before merging this, PyPI and TestPyPI should have the ``AMQP_URL`` renamed to ``BROKER_URL``.

To actually switch PyPI (not TestPyPI) to SQS, do the following (in order):

1. [x] Merge https://github.com/python/pypi-infra/pull/29 and ``terraform apply`` it.
2. [x] Configure the PyPI AWS IAM user with the ``PyPIWorkerSQS`` policy.
3. [x] Update ``BROKER_URL`` to ``sqs:///pypi-worker?region=us-east-2`` (Note the triple slash).


ewdurbin:
`AMQP_URL` settings for test-pypi and pypi have been copied into `BROKER_URL` for the time being.

As this PR is backwards compatible, let's merge it and test migrating to SQS in testpypi first.

ewdurbin (approved):

PR #4428, opened 2018-07-31
use localstack for celery SQS broker locally

ewdurbin:


ewdurbin:
SQS has been reliable and we're sticking with it. Going to merge this!

PR #4439, opened 2018-08-02
Add link to PSF in footer

### LINKED ISSUE #4432: Link to Python.org
There's no link to python.org or the PSF on the main page. It seems logical that there should be one. 

There seems to be plenty of space in `header.site-header` to add a link. I would even like to see the whole top header from python.org replicated (but that is a separate issue I suppose).

image

**Additional context**
There is this:
image
I believe that even making this a link would be helpful.

azeemansar:
Fixes #4432.


di:
Thanks @azeemansar!

PR #4462, opened 2018-08-03
Remove the rego_otk table

dstufft:
Assign to @ewdurbin so he can merge this one.

Removes the no longer needed ``rego_otk`` table and related indexes. We might want to take a backup prior to removing this incase we ever need it again, but we shouldn't require it.


ewdurbin:
I think all good here, but we should remove https://github.com/pypa/warehouse/blob/745183f039df30ef507d2376733064b06e8312f6/dev/clean.sql#L16 at the same time.

ewdurbin:
Cool, I'll take care of merging this once I get a chance to capture and store a snapshot of the table.

ewdurbin:
`s3://pypi-infra-terraform/backup/warehouse/rego_otk-2018-08-03.sql.gz.gpg`

encrypted with my GPG public key with id `22ADAE0FC41E160AD1596B118F8E423168928420` and @dstufft's GPG public key with id `7C6B7C5D5E2B6356A926F04F6E3CBCE93372DCFA`

PR #4468, opened 2018-08-05
Prevent compromised/breached passwords

dstufft:
This uses the HaveIBeenPwned password API to determine if a given password has ever appeared in a breached list of passwords or not. If a password appears in the list of breached/compromised passwords, it will reject the password during registration, password reset, and changing passwords.

This does *NOT* send the password to HIBP, instead of does a sha1 hash of the password, and sends the first 5 characters of the hex digest to HIBP and HIBP will return a list of all of the sha1 hashes that begin with that prefix. This is not enough for HIBP to determine the password, or whether the user's password existed in the returned set. For more information see https://www.troyhunt.com/ive-just-launched-pwned-passwords-version-2/.

### DIFF
--- a/tests/unit/accounts/test_core.py
+++ b/tests/unit/accounts/test_core.py
@@ -17,8 +17,16 @@ import pretend
 import pytest
 
 from warehouse import accounts
-from warehouse.accounts.interfaces import IUserService, ITokenService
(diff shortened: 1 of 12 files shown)


brainwane:
Donald's blog post on this: https://caremad.io/posts/2018/08/pypi-compromised-passwords/

PR #4474, opened 2018-08-06
Add JSON /stats endpoint + Human Friendly Units

cooperlees:
- With the help of @dstufft get a JSON API + HTML stats endpoint
-- Tested JSON with `curl -v -H "Accept: application/json" -H 'Content-Type: application/json' http://localhost/stats/`
- Import humanfriendly to use for human friendly units
- Adjust tests to pass

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -343,11 +343,12 @@ def stats(request):
         .limit(100)
         .all()
     )
-
-    return {
-        "total_packages_size": total_size_query[0][0],
-        "top_packages": top_100_packages,
+    # Move top packages into a dict to make JSON more self describing
+    top_packages = {
+        pkg_name: {"size": pkg_bytes} for pkg_name, pkg_bytes in top_100_packages
     }
+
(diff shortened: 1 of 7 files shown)


warehouse/views.py
```diff
@@ -325,19 +326,34 @@ def filter_key(item):
...
         .limit(100)
         .all()
     )
+    top_100_packages_human = {}
+    for pkg_name, pkg_bytes in top_100_packages:
+        top_100_packages_human[pkg_name] = format_size(parse_size(str(pkg_bytes)))
 
     return {
         "total_packages_size": total_size_query[0][0],
+        "total_packages_size_human": total_size_human,
```
dstufft: Instead of adding a parameter here, we should just return the bytes in the dictionary. In the template we can convert to using a human friendly display by doing ``total_packages_size|filesizeformat()``. This is what we do eslewhere:

https://github.com/pypa/warehouse/blob/6e8aebc1bc9fc01def59c877387b21a54b0c62b7/warehouse/templates/packaging/detail.html#L251

PR #4475, opened 2018-08-06
Add Resilience to HIBP API calls

dstufft:


warehouse/http.py
```diff
@@ -26,7 +28,12 @@ def __call__(self, request):
             return session
         except AttributeError:
             request.log.debug("creating new session")
+
+            adapter = HTTPAdapter(max_retries=1)
```
ewdurbin: Will this affect any other remote calls as well? Seems like a not terrible idea overall, just curious what the impact may be.

warehouse/accounts/services.py
```diff
@@ -246,9 +248,16 @@ def __eq__(self, other):
...
 class HaveIBeenPwnedPasswordBreachedService:
-    def __init__(self, *, session, api_base="https://api.pwnedpasswords.com"):
+    def __init__(
+        self, *, session, api_base="https://api.pwnedpasswords.com", metrics=None
+    ):
         self._http = session
         self._api_base = api_base
+        self._metrics = metrics
+
+    def _metrics_increment(self, *args, **kwargs):
```
ewdurbin: I think is clearly a smell that I should get to work refactoring datadog into a proper interface/service.

dstufft: Yea, I thought of that as I was writing this too :) A nice ``IMetricsService`` with an inmemory provider could be cool. Might even make it possible to get rid of the ``notdatadog`` script, since the in memory one (or maybe "local" is a better name) could just print out directly instead of sending to ``notdatadog``.

warehouse/http.py
dstufft: It will affect anything that uses ``request.http``, which is currently just the bit of code that fetches the PEM from SNS to verify SNS messages from SES.

PR #4487, opened 2018-08-08
Travis CI: Upgrade testing from Python 3.6 --> 3.7

cclauss:
The [Travis CI worker continues to be Trusty](https://travis-ci.org/pypa/warehouse/jobs/413453666#L20) instead of __Xenial__ as we requested in this PR's changes the __.travis.yml__ file.

According to travis-ci/travis-ci#9069, if the Travis worker is less than Xenial then the [download of Python 3.7 will fail](https://travis-ci.org/pypa/warehouse/jobs/413453666#L431)

See #4488 for possible next steps...


ewdurbin:
Celery needs Python 3.7 support before we can move forward. It appears this will come with Celery version 4.3, which the maintainers indicate should occur by October.

pradyunsg:
This also moves to a full blown VM on their service though. That'll cause a slight slowdown.

ewdurbin:
Kombu released an update... that may help? worth a shot!

ewdurbin:
Kombu is python3.7 compatible as of #5402... let's see how this goes!

ewdurbin:
tested locally, looks like we still need a celery release.

PR #4489, opened 2018-08-08
add FAQ entry for compromised passwords

ewdurbin:


warehouse/accounts/forms.py
```diff
@@ -14,6 +14,8 @@
 import wtforms
 import wtforms.fields.html5
 
+from pyramid.threadlocal import get_current_request
```
ewdurbin: not optimal but I'm not sure how else to get the link in front of the user in the validation error

warehouse/accounts/forms.py
```diff
@@ -115,8 +117,15 @@ def validate_new_password(self, field):
             field.data, tags=["method:new_password"]
         ):
             raise wtforms.validators.ValidationError(
-                "This password has appeared in a breach or has otherwise been "
-                "compromised."
+                (
+                    "This password has appeared in a breach or has otherwise been "
+                    "compromised and cannot be used. See {project_help} for more "
```
ewdurbin: tried wrapping in an `a` tag, but we don't render html in our validation errors... which is pretty wise.

warehouse/accounts/forms.py
dstufft: Could pass the ``request`` into the form like we do the ``breach_service``, or pass in the link directly.

Alternatively you could add a method to the breach service that returns a failure message, and the factory for the breach service could then be modified to accept a request (or again, the url directly).

I kind of like the idea of putting the failure message method (or even just a property. Zope interface supports properties too:

```python
from zope.interface import Attribute, Interface

class IFoo(Interface):
    failure_message = Attribute("The failure message")


@implementer(IFoo)
class Foo:

    def __init__(self, url):
        self._url = url

    @property
    def failure_message(self):
        return f"Bad things happened, see: {self._url}"
```

dstufft: Nice thing about making the failure message part of the service, is it will let us tailor the message to different services, if we end up using a different API than HIBP or something.

dstufft: Although if it's completely generic like it is now, maybe it would be better on the form? I dunno.

warehouse/accounts/forms.py
dstufft: Hrm. Maybe It would be better to raise a sigil here, like "<REPLACE WITH BREACHED PASSWORD HELP>", and add a conditional to the template looking for that string in specific? It's kind of gross but it lets us put real HTML in the error message?

Actually, what happens if you try to put html in this? I wonder if ``raise ValidationError(jinja2.Markup("..."))`` works?

warehouse/accounts/interfaces.py
```diff
@@ -97,6 +97,8 @@ def loads(token):
 
 
 class IPasswordBreachedService(Interface):
+    failure_message = Attribute("Password compromised")
```
dstufft: This should document what this attribute is, so something like ``Attribute("The message to describe the failure that occured")`` or something.

warehouse/accounts/forms.py
ewdurbin: neat, your suggestion works! updating.

ewdurbin:
@di or @brainwane any input on the copy in the FAQ entry here?

warehouse/templates/pages/help.html
```diff
@@ -25,6 +25,7 @@
 {% macro publishing() %}How do I package and publish my code for PyPI?{% endmacro %}
 {% macro trove_classifier() %}What's a trove classifier?{% endmacro %}
 {% macro verified_email() %}Why do I need a verified email address?{% endmacro %}
+{% macro compromised_password() %}Why won't PyPI accept my password?{% endmacro %}
```
brainwane: Sometimes PyPI doesn't accept a password because it's inaccurate (e.g., the caps lock key is on), so I suggest rewording this to include a key phrase from the error message "This password has appeared in a breach or has otherwise been compromised and cannot be used."

warehouse/templates/pages/help.html
```diff
@@ -158,6 +160,20 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
         <p>
           You can manage your account's email addresses in your <a href="{{ request.route_path('manage.account') }}">Profile</a>. This also allows for sending a new confirmation email for users who signed up in the past, before we began enforcing this policy.
         </p>
+
+
+        <h3 id="compromised-password">{{ compromised_password() }}</h3>
+          <p>
+            PyPI securely checks if passwords users supply when registering, authenticating, or updating their password have appeared in known data breaches.
```
brainwane: This sentence is a little tangled. How about:

> Each time a user supplies a password -- while registering an account, authenticating, or updating their password -- PyPI securely checks whether that password has appeared in known data breaches.

warehouse/templates/pages/help.html
```diff
@@ -158,6 +160,20 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+          <p>
+            During each of these processes, PyPI generates a SHA-1 hash of the supplied password and uses the first five (5) characters of the hash to check the <a href="https://haveibeenpwned.com/API/v2#SearchingPwnedPasswordsByRange" target="_blank" rel="noopener">Have I Been Pwned API</a> and determine if the password has been previously compromised. The plaintext password is never stored by PyPI or submitted to the Have I Been Pwned API.
+          </p>
+          <p>
+            PyPI will not allow such passwords to be used when setting a password at registration or updating your password.
+          </p>
+          <p>
+            If the password you supplied is noted as having appeared in known compromises, you should change it all other places that you use it as soon as possible.
```
brainwane: I suggest we include the text of the error message here:

>  This password has appeared in a breach or has otherwise been compromised and cannot be used.

so people will find this page when searching the web for the error message, and will find this answer when using Control-F on the page.

warehouse/templates/pages/help.html
ewdurbin: "Why is PyPI telling me my password is compromised?"

warehouse/templates/pages/help.html
ewdurbin: I'm not sure of a good way to incorporate the full error message, but whenever it is shown, it displays a link to this FAQ entry. Is that sufficient?

warehouse/templates/pages/help.html
```diff
@@ -158,6 +160,21 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
         <p>
           You can manage your account's email addresses in your <a href="{{ request.route_path('manage.account') }}">Profile</a>. This also allows for sending a new confirmation email for users who signed up in the past, before we began enforcing this policy.
         </p>
+
+
+        <h3 id="compromised-password">{{ compromised_password() }}</h3>
+          <p>
+            Each time a user supplies a password &mdash; while registering, authenticating, or updating their password &mdash; PyPI securely checks whether that password has appeard in public data breaches.
```
dstufft: Would it be a good idea to lead off with the fact that this does not mean PyPI itself was breached?

ewdurbin: "PyPI itself has not suffered a breach, this is a protective measure to reduce the risk of [credential stuffing](https://www.owasp.org/index.php/Credential_stuffing) attacks against PyPI and it's users."

warehouse/templates/pages/help.html
```diff
@@ -25,6 +25,7 @@
 {% macro publishing() %}How do I package and publish my code for PyPI?{% endmacro %}
 {% macro trove_classifier() %}What's a trove classifier?{% endmacro %}
 {% macro verified_email() %}Why do I need a verified email address?{% endmacro %}
+{% macro compromised_password() %}Why is PyPI telling me my password is compromised?{% endmacro %}
```
brainwane: :+1:

warehouse/templates/pages/help.html
brainwane: Typo in "whether that password has appeard" -- should be "appeared".

ewdurbin: something like that? not sure if there's a better way to word "PyPI itself has not suffered a breach"

warehouse/templates/pages/help.html
brainwane: It's a good step and it helps, but I'm sorry to say no, I don't think it's sufficient, because it doesn't help people who do a control-F. Here's a suggested phrasing:

> PyPI will not allow such passwords to be used when setting a password at registration or updating your password. 
> If you receive an error message saying that "This password has appeared in a breach or has otherwise been compromised and cannot be used", you should change it all other places that you use it as soon as possible.

warehouse/templates/pages/help.html
brainwane: That sounds like a good idea to me.

warehouse/templates/pages/help.html
```diff
@@ -158,6 +160,24 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
         <p>
           You can manage your account's email addresses in your <a href="{{ request.route_path('manage.account') }}">Profile</a>. This also allows for sending a new confirmation email for users who signed up in the past, before we began enforcing this policy.
         </p>
+
+
+        <h3 id="compromised-password">{{ compromised_password() }}</h3>
+          <p>
+            PyPI itself has not suffered a breach, this is a protective measure to reduce the risk of <a href="https://www.owasp.org/index.php/Credential_stuffing" target="_blank" rel="noopener">credential stuffing</a> attacks against PyPI and it's users.
```
brainwane: Typo - should be:

>             PyPI itself has not suffered a breach. This is a protective measure to reduce the risk of <a href="https://www.owasp.org/index.php/Credential_stuffing" target="_blank" rel="noopener">credential stuffing</a> attacks against PyPI and its users.

PR #4506, opened 2018-08-09
Refactor metrics to use pyramid_services

dstufft:
This provides indirection for our metrics integration, that allows us to do things like have a no-op provider when other units under test require a metrics provider.

This also comes with improvements to testing pyramid services in our test framework.

Still to do is to convert the pyramid_datadog metrics into using the service instead of the hardcoded datadog client.

Will need to add a setting like ``METRICS_BACKEND=warehouse.metrics.DataDogMetrics``.


dstufft:
I've added the configuration to Cabotage, will confirm with @ewdurbin if there are any ``DATADOG_*`` headers that wouldn't have been reflected in the cabotage UI.

Issue #4481, opened 2018-08-07
Users with no primary email set cannot set a new primary email.


ewdurbin:
**Describe the bug**

If a user has no primary email set, the [`ManageAccountViews.change_primary_email`](https://github.com/pypa/warehouse/blob/472110dc94d5bd17b60e350bbaa42dc4855555a7/warehouse/manage/views.py#L153-L183) view errors when attempting to [send the notification email](https://github.com/pypa/warehouse/blob/472110dc94d5bd17b60e350bbaa42dc4855555a7/warehouse/email/__init__.py#L137-L149) as [`accounts.models.User.email`](https://github.com/pypa/warehouse/blob/472110dc94d5bd17b60e350bbaa42dc4855555a7/warehouse/accounts/models.py#L82-L93) returns `None`.

**Expected behavior**

User is able to set a primary email.

**To Reproduce**

Manually create a user with no emails marked as Primary, attempt to set a primary address.

**Additional context**

Needs investigation to see how widespread the initial state is.

Ref: https://sentry.io/python-software-foundation/warehouse-production/issues/633645579

ewdurbin:
```sql
select
  count(*)
from
  accounts_user as u
	full outer join (
	  select user_id, email from accounts_email where accounts_email.primary
	) as e
	on u.id = e.user_id
where e.email is NULL;
```

Currently returns 34833 records.

ewdurbin:
Users with _no_ email address at all:

```sql
select
  u.username
from
  accounts_user as u
	full outer join (
	  select user_id, email from accounts_email
	) as e
	on u.id = e.user_id
where e.email is NULL;
```

Currently returns 34827 records.

dstufft:
In https://github.com/pypa/warehouse/pull/4515 I make it so these users can add a primary email address to their account and generally improve the work flow for these users.

However one thing I want to possibly raise here, is we have 34792 users who do not have a primary email address on file (of which 2 have a non primary email address on file) who are not attached to a project in any way. Should we purge those 34.7k users? If I query them and look down through them, a lot of the usernames look auto generated (20.6k of them are exactly 10 characters long and the patterns look auto generated).

Even if they're real users.. they're real users who almost all of them haven't logged in at all in *years* with no email to contact them at, and no projects on PyPI. It seems to me like the only risk here is that someone might lose a username that they registered for themselves and then never used (or haven't used in years), which seems pretty low risk to me.

[...]

PR #4515, opened 2018-08-10
Improve the experience of a user without a primary email

dstufft:
* Gives a better warning in the header of the manage pages, that indicate there is no primary email, rather than that the primary email is not verified.
* Show the drop down beside a non primary email if there is only one email.
* If a new email is added to a user who does not currently have a primary email address, default to adding that new email as their primary address.
* Skip sending the warning email to the old primary email address, if there is no old primary address.

Refs #4481

We should probably decide if this is the approach we want to take. The basic idea behind these changes are to function as optimally as possible when we don't have a primary email address, but to try and get something marked as a primary address as quickly as possible.

There are.. 5? users (according to #4481) that have an email address, but do not have a primary email address. We could just email those 5 users and then move that email to primary. However I don't personally think it's worth it.

One possible alternative is to attempt to manually reach out to the ~40 users who do not have a primary email address who are attached to a project in some way (see https://github.com/pypa/warehouse/issues/4481#issuecomment-411955541) and manually fix those users. However this is a fairly low cost change so I again don't think it's worth it.

Fixes https://sentry.io/python-software-foundation/warehouse-production/issues/633645579/.


ewdurbin:
This approach seems most reasonable, allowing users to resolve the issue themselves. We can always circle back "in time" and reach out if necessary.

PR #4518, opened 2018-08-10
Handle errors connecting to PostgreSQL better

dstufft:
Instead of just exploding when we can't connect to PostgreSQL, we will instead raise a different error (``DatabaseNotAvailable``) so we can differentiate it from other causes of this error. Then we'll add an exception view that will turn the ``DatabaseNotAvailable`` error into a generic 503 error.

This will keep these errors from getting caught by sentry, but to provide insight into if they're happening or not metrics and logging have been added.

Fixes: https://sentry.io/python-software-foundation/warehouse-production/issues/534930428/.
Fixes: https://sentry.io/python-software-foundation/warehouse-production/issues/570776802/.

### DIFF
--- a/warehouse/db.py
+++ b/warehouse/db.py
@@ -11,8 +11,10 @@
 # limitations under the License.
 
 import functools
+import logging
 
 import alembic.config
+import psycopg2
 import psycopg2.extensions
 import sqlalchemy
 import venusian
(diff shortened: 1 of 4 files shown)


warehouse/db.py
```diff
@@ -23,15 +25,25 @@
...
+logger = logging.getLogger(__name__)
+
+
 DEFAULT_ISOLATION = "READ COMMITTED"
 
 
+# A generic wrapper exception that we'll raise when the database isn't available, we
+# use this so we can catch it later and turn it into a generic 5xx error.
+class DatabaseNotAvailable(Exception):
+    ...
```
ewdurbin: huh, never seen an ellipse in place of `pass`. is there any actual difference?

dstufft: It's functionally the same. I think it might be a Python3-ism. It's actually slightly wrong to use it here (not sure why I did!), it's generally recommended to use it when the body has been omitted for some reason (like in typing stubs or similar).

Doesn't effect the outcome at all though.

Issue #2877, opened 2018-02-01
Search does not prioritize exact match when package name contains numbers


di:
For example, [a search for `python-miio`](https://pypi.org/search/?q=python-miio) has the `python2-miio` as the first result, likely because the `2` is being tokenized out of the query.

gsb-eng:
@di Any progress on this?  If not, Can I pick this?

brainwane:
@gsb-eng Yes, please go ahead and pick it up! At least within Warehouse, our rule is: If there is not a corresponding pull request for this issue (visibly linked by issue number/PR number), it is up for grabs. So in cases like this, you don't even need to ask. :)

As always, if you have questions along the way as you work on this, please feel free to ask them here, in [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev). Thank you!

waseem18:
@di Could you please help me in providing me more information about this issue?

di:
@waseem18 Unfortunately I don't know much more than you here. We got a report that packages with a number in them could be prioritized higher in search than a package with the exact name as the search query.

However looking at https://pypi.org/search/?q=python-miio now, it seems like `python-miio` has the highest rank now. We will likely need to do some debugging of the elasticsearch query and the results to see why it (at some point) was possible.

waseem18:
Yeah - I was feeling the same. Because now the search queries seems to work as expected.

Will do some debugging and check.

di:
Here's another current example of this happening:

https://pypi.org/search/?q=psycopg2

waseem18:
@di I presume the issue is because of usage of `lowercase` filter in [normalized_name](https://github.com/pypa/warehouse/blob/b66a37d0f5dd5d77730a1bfa427548aa554c40b5/warehouse/packaging/search.py#L29) which is tokenizing `psycopg2` to `psycopg` and `2`

As per Elastic search, 

>  tokenizer breaks text into terms whenever it encounters a character which is not a letter

di:
Sounds like the same thing would happen with package names containing other valid non-alphanumeric characters like `-`, `_`, and `.`. Perhaps we need to tweak that tokenizer a bit.

waseem18:
Yeah - We can have a custom analyzer so that we can avoid tokenizing numbers and non-alphanumeric characters.

di:
[...]

PR #4519, opened 2018-08-10
Fix for #2877 - add nonanalyzed (i.e. keyword) 'name' field to package search

selotape:
To whom it may concern,

I fixed the bug to the best of my understanding, but I still have a few issues. I'm "prematurely" opening this PR with the purpose of receiving some review, feedback, and help.

See some code comments about the fix inline.

About my status:

1. I managed to set up and run warehouse in a container (😄)
2. I couldn't get the tests to execute :\. When I run "make tests" I get the following error:
```
    env: ‘bin/tests’: No such file or directory
    Makefile:99: recipe for target 'tests' failed
```
3. I couldn't replicate the bug on my PC, hence I couldn't make sure the bug is solved. Finding "bad" examples is hard even in production, and the test ES probably has much fewer packages. The bug can probably be fabricated by adding carefully named mock packages to the index, but I've no idea how to do that.

4. The "tests" I did run were making sure that 1) search still works, 2) `make initdb` still succeeds and 3) the ES search query contains the new search field, `name.keyword`.

Any help and guidance would be appreciated.
Thanks!

### DIFF
--- a/warehouse/packaging/search.py
+++ b/warehouse/packaging/search.py
@@ -32,6 +32,7 @@ NameAnalyzer = analyzer(
 
 @doc_type
 class Project(Document):
+
     name = Text(fields={'keyword': Keyword()})
     normalized_name = Text(analyzer=NameAnalyzer)
     version = Keyword(multi=True)
--- a/warehouse/packaging/search.py
+++ b/warehouse/packaging/search.py
@@ -32,8 +32,7 @@ NameAnalyzer = analyzer(
 
 @doc_type
 class Project(Document):
-
-    name = Text()
+    name = Text(fields={'keyword': Keyword()})
     normalized_name = Text(analyzer=NameAnalyzer)
     version = Keyword(multi=True)
     latest_version = Keyword()
--- a/warehouse/search/queries.py
+++ b/warehouse/search/queries.py
@@ -10,27 +10,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SEARCH_FIELDS = [
-    "author",
-    "author_email",
-    "description",
-    "download_url",
-    "home_page",
-    "keywords",
-    "license",
-    "maintainer",
-    "maintainer_email",
-    "normalized_name",
-    "platform",
-    "summary",
-]
 SEARCH_BOOSTS = {
     "name": 10,
     "normalized_name": 10,
     "description": 5,
     "keywords": 5,
     "summary": 5,
+    "author": 1,
+    "author_email": 1,
+    "download_url": 1,
+    "home_page": 1,
+    "license": 1,
+    "maintainer": 1,
+    "maintainer_email": 1,
+    "platform": 1,
 }
+
 SEARCH_FILTER_ORDER = (
     "Framework",
     "Topic",
--- a/warehouse/search/queries.py
+++ b/warehouse/search/queries.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 SEARCH_BOOSTS = {
+    "name.keyword": 20,
     "name": 10,
     "normalized_name": 10,
     "description": 5,
--- a/warehouse/search/queries.py
+++ b/warehouse/search/queries.py
@@ -26,7 +26,6 @@ SEARCH_BOOSTS = {
     "maintainer_email": 1,
     "platform": 1,
 }
-
 SEARCH_FILTER_ORDER = (
     "Framework",
     "Topic",
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -41,7 +41,7 @@ from warehouse.cache.origin import origin_cache
 from warehouse.cache.http import add_vary, cache_control
 from warehouse.classifiers.models import Classifier
 from warehouse.packaging.models import Project, Release, File, release_classifiers
-from warehouse.search.queries import SEARCH_BOOSTS, SEARCH_FIELDS, SEARCH_FILTER_ORDER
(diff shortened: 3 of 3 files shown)


warehouse/packaging/search.py
```diff
@@ -33,7 +33,7 @@
 @doc_type
 class Project(Document):
 
-    name = Text()
+    name = Text(fields={'keyword': Keyword()})
```
selotape: Here is where I add the inner field `name.keyword`, with the keyword type.
This is following @HonzaKral s suggestion.

warehouse/search/queries.py
```diff
@@ -10,26 +10,21 @@
...
     "summary": 5,
+    "author": 1,
+    "author_email": 1,
+    "download_url": 1,
+    "home_page": 1,
+    "license": 1,
+    "maintainer": 1,
+    "maintainer_email": 1,
+    "platform": 1,
 }
```
selotape: This is a small refactoring - I removed the duplicate definition of search fields. Now you only need to state search fields once and you must set a boost for it.

warehouse/views.py
```diff
@@ -413,8 +413,8 @@ def form_query(query_type, query):
     Returns a multi match query
     """
     fields = [
-        field + "^" + str(SEARCH_BOOSTS[field]) if field in SEARCH_BOOSTS else field
-        for field in SEARCH_FIELDS
+        field + "^" + str(boost) if boost != 1 else field
+        for field, boost in SEARCH_BOOSTS.items()
     ]
```
selotape: This is the same refactoring as above

warehouse/packaging/search.py
selotape: FYI - I saw no need to add a "deserialization" of `name.keyword` in the `from_db` class method. So I didn't do it :)

di:
> I couldn't get the tests to execute :. When I run "make tests" I get the following error:

Is there more to this error? I would imagine there would be a stack trace or something similar above what you included.

> I couldn't replicate the bug on my PC, hence I couldn't make sure the bug is solved. Finding "bad" examples is hard even in production, and the test ES probably has much fewer packages. The bug can probably be fabricated by adding carefully named mock packages to the index, but I've no idea how to do that.

You can upload new distributions directly to your development instance of Warehouse with `twine`, see http://warehouse.readthedocs.io/development/reviewing-patches/#testing-with-twine for instructions. I like to use https://github.com/pypa/sampleproject for this purpose, and modify it as needed.

You should be able to reproduce this by adding two new packages to your development instance.

You'll need to run `make reindex` afterwards to rebuild the elasticsearch index to include the new packages.

ewdurbin:
> You'll need to run make reindex afterwards to rebuild the elasticsearch index to include the new packages.

With #4328 that _shouldn't_ be the case as long as the workers are running!

ewdurbin:
@selotape it sounds like you're using a bit of a non-standard environment to develop warehouse.

Could you help us by describing how you're running warehouse? Our expected development environment is documented at https://warehouse.readthedocs.io/development/getting-started/#detailed-installation-instructions, if you're diverging from that too much it can be difficult for contributors to assist you.

If you're running into snags with the documented setup, that's a bug and we encourage you to file an issue!

di:
Oh yeah, forgot that we fixed that!

selotape:
@ewdurbin , 

about my setup, I did follow the "detailed installation instructions", but I'm not familiar with docker so I may have gotten some things wrong.

My main OS is windows 10 so I installed ubuntu on the WSL. I installed docker CE in windows and docker and docker-compose on the Ubuntu. I also followed the "quirks" section. After all this I ran `make serve` and `make initdb` and things seemed to run well, including all the sanity tests (except `make tests`...).

Is there anything specific I can check and show you to make sure I'm set up properly?



And about the tests, the following is all I get when I run `make tests` from the project root - 
image


I find the "no such file" error strange, because the `bin/tests` file does exist - 
image

...maybe it's somehow related to the fact that I'm user `root` on the Ubuntu... and the docker can't see the tests...?
Any other suggestions for how to run the tests?

Thanks!

selotape:
btw, `make lint` also doesn't work for me. It looks like a python3.6 issue - 
image

Is the python3.6 supposed to be installed on the container or on the Ubuntu?

warehouse/packaging/search.py
HonzaKral: That's not needed - the `name.keyword` does not acctually appear on the document, it is just an alternative index that can be used for searching and/or aggregations/sorting.

brainwane:
@selotape I'm so sorry for the delay in response! Short answer: [we're in IRC most weekdays around 10am-5pm US Eastern Time](https://webchat.freenode.net/?channels=%23pypa-dev,pypa), come talk with us?

Longer answer: [From December 2017 till the end of April 2018, PyPI had funding to get the new site up and running and perform the switchover.](https://wiki.python.org/psf/PackagingWG) Then the grant money ran out. Then the Packaging Working Group, seeking [donations](http://donate.pypi.org/) and further grants to fund more work, got [some new funding](https://pyfound.blogspot.com/2018/12/upcoming-pypi-improvements-for-2019.html), and we're getting our momentum back on Warehouse work and being able to spend more time helping contributors. So if you come talk with us in IRC sometime soon we should be able to help you get your setup fixed! Or if you'll be at PyCon North America in May, [talk with us during the sprints](https://wiki.python.org/psf/PackagingSprints)!

Thanks and sorry again.

PR #4532, opened 2018-08-11
Purge the database of inactive users with no email or projects

dstufft:
Deletes any user who matches the follow constraints:

* Does not have any emails associated with their account.
* Has not logged in for over a year (or last login is null, to deal with *really* old data).
* Does not have a role on any projects in PyPI.

I'm making this a migration because I figure it's the easiest way to execute some SQL and get it reviewed etc.

I'm going to leave this to @ewdurbin to merge, because as is usual, we probably want to snapshot the database before executing just in case. I was not able to really test this locally because the dev database doesn't contain any users that match, though there is a single user who matches if we drop the last login restriction. I did test it with that single user and it worked fine, but it may be a smart idea to grab a copy of the database and make sure that it only deletes 34790 users, and there are no roles, emails, or projects gone once those users are deleted.

Refs #4481.


ewdurbin:
snapshot at `s3://pypi-infra-terraform/backup/warehouse/warehouse-2018-08-11-user-purge.sql.gz.gpg`

PR #4533, opened 2018-08-11
Revert "Purge the database of inactive users with no email or projects"

ewdurbin:
Reverts pypa/warehouse#4532


ewdurbin:
Looks like we need to exclude the sentinel user `deleted-user`

```
ERROR:  update or delete on table "accounts_user" violates foreign key constraint "journals_submitted_by_fkey" on table "journals"
DETAIL:  Key (username)=(deleted-user) is still referenced from table "journals".
```

ewdurbin:
Will also need to handle users from this purge that have actions in the `journals` table.

ewdurbin:
cc @dstufft

PR #4534, opened 2018-08-11
Email verification feedback

### LINKED ISSUE #3590: Provide feedback to users about email delivery issues
Once https://github.com/pypa/warehouse/pull/3580 is merged, Warehouse will begin tracking delivery issues for email addresses we have on file, and can ultimately un-verify a particular email address due to those reasons. We'll want to expose those reasons to end users so that they understand what has occurred to their email addresses on file.

This gives us the following states an email address can be in:

1. Unverified because the email has had a hard failure during delivery.
2. Unverified because the email had too many transient delivery failures, and we've given up sending to that address and we're treating it as a hard failure.
3.  Unverified because the user has marked an email from this address as spam, and their ISP or email provider has forwarded that complaint on to us.
4. Unverified because it is brand new and has yet to be verified:
5. Verified, but we've been having transient issues sending email to this address, if they don't go away this will be upgraded to the (2) case.
6. Verified, no issues.

I can handle wiring this up in the code once we have the design side sorted out for how it should be implemented.

yeraydiazdiaz:
Fixes #3590 , here are some captures:

email_verification_verified_warning
email_verification_unverified_none
email_verification_unverified_soft_bounce
email_verification_unverified_hard_bounce
email_verification_unverified_spam_complaint

Copy suggestions are welcomed and encouraged.

I couldn't find any tests for the actual rendering of the views, have I missed something or is this by design?


nlhkabu:
Hi @yeraydiazdiaz thanks for this PR :grin: 

From an accessibility perspective, we should give another indicator *other* than color to indicate status.  This is for color blind users.

To do this, I think we should add a small icon inside each verified / unverified badges.

I'm thinking:

| Badge class | Badge text  | Badge icon  |  tooltip text / message |
|---|---|---|---|
| badge--success | Verified | fa-check | - |
| badge--warning | Verified | fa-exclamation-circle | Intermittent delivery failures may lead to verification loss |
| badge--danger | Unverified | fa-spinner | Email verification in progress |
| badge--danger | Unverified | fa-exclamation-circle | Too many delivery failures |
| badge--danger | Unverified | fa-exclamation-circle | Hard failure during delivery |
| badge--danger | Unverified | fa-exclamation-circle | Email from PyPI being treated as spam |

Maybe we could also add a FAQ under https://pypi.org/help/#verified-email that describes each of these messages in detail?

We could link to the FAQ either from the tooltip, or by adding a link above the table, something like:

> For more information on verification status', visit the [FAQ]

What do you think?

yeraydiazdiaz:
Adding a badge makes total sense.

I don't think you can add a link in the tooltip though as the text is added in the `arial-label` attribute and putting and `a` tag would break the HTML.

nlhkabu:
Ok, I thought that might be the case.  In this case, we can add a separate link.  Maybe, though, we can do this in a separate PR to keep this granular?

yeraydiazdiaz:
Updated with icons and copy tweaks
screen shot 2018-08-11 at 19 46 36
screen shot 2018-08-11 at 19 46 07
screen shot 2018-08-11 at 19 44 49
screen shot 2018-08-11 at 19 43 58
screen shot 2018-08-11 at 19 43 32
screen shot 2018-08-11 at 19 43 04

nlhkabu:
Looking good @yeraydiazdiaz! One thing - could we move the icons to the left of the text?
I'm also thinking that using `exclamation-triangle` might be better?

@dstufft are you happy with this text?

dstufft:
The text looks fine to me.

yeraydiazdiaz:
Updated location and `exclamation-triangle`:

screen shot 2018-08-12 at 09 45 16
screen shot 2018-08-12 at 09 47 22
screen shot 2018-08-12 at 09 47 51
screen shot 2018-08-12 at 09 49 11
screen shot 2018-08-12 at 09 49 41
screen shot 2018-08-12 at 09 50 07

pradyunsg:
A super minor point -- we can probably drop the tick in the verified case, since it's redundant.

nlhkabu:
I disagree @pradyunsg - I we should be explicit over implicit :) Also, visually, it is more consistent.

dstufft:
Thanks! ✨

pradyunsg:
I feel that having text only badge would be consistent with the badge beside it. ;) 

As I said, it's a minor point anyway; I don't have a problem going either way and slightly prefer the one I'm suggesting. I'm not going to push for it. :)

PR #4537, opened 2018-08-11
Purge the database of inactive users with no email or projects.. Again!

### LINKED ISSUE #4481: Users with no primary email set cannot set a new primary email.

dstufft:
Second try at this, this time we're additionally filtering out any user that has ever appeared in the journals entry. That keeps ``deleted-user`` from being deleted, and only filters out 7 other users.

Like last time, leaving this for @ewdurbin to merge.

Fixes #4481.

### DIFF
--- a/warehouse/migrations/versions/f71770e7d9b3_remove_users_without_an_email_or_a_.py
+++ b/warehouse/migrations/versions/f71770e7d9b3_remove_users_without_an_email_or_a_.py
@@ -36,8 +36,12 @@ def upgrade():
                 OR accounts_user.last_login IS NULL)
           AND NOT EXISTS
             (SELECT 1
-             FROM ROLES
+             FROM roles
              WHERE roles.user_name = accounts_user.username)
+          AND NOT EXISTS
+            (SELECT 1
+             FROM journals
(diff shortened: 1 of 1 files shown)


pradyunsg:
Linking to previous iteration of this: #4532

PR #4538, opened 2018-08-11
Project tabs back button

### LINKED ISSUE #4461: Browser "Back" button not working for first click on project page
**Describe the bug**

After loading a project page, if I click on "Release history" or "Download files" and then click on my browser's "back" button, nothing happens.

**Expected behavior**

The previous page (a.k.a. the project description page) should be displayed

**To Reproduce**

- go to https://pypi.org/project/pyler/ (but any project would do)
- click on "Project history"
- click on you browser's "back" button
- if you're still on the history page, the bug is reproduced.

**My Platform**
- Reproduced both on Firefox (61.0.1) OSX (high sierra) and Chromium Ubuntu (bionic)

**Additional context**

The problem is when the previous URL doesn't contain an anchor (https://pypi.org/project/pyler/ instead of https://pypi.org/project/pyler/#description).

I'm guessing there's a handler on `popstate` event and this handler does nothing if there's no state information, but instead, no state info should be an indication that it needs to display the project description page.

yeraydiazdiaz:
Fixes #4461 

back_button_project_tabs

@ewjoachim thanks reporting and researching possible fixes, [I did try your approach first](https://github.com/pypa/warehouse/commit/db3f30ab62c1c725068a889382a717944e2ab801) and while it did fix the Back button pushing the first tab's hash broke the forward button!

This solution simply toggles the first tab without pushing the state similar to what happens on `connect`.


dstufft:
Thanks! Will merge once tests pass again.

dstufft:
Merged! Thanks ✨

ewjoachim:
Ah, perfect, thank you very much !

PR #4541, opened 2018-08-12
Disable users with breached passwords

### LINKED ISSUE #4471: Notify & restrict users with breached/compromised passwords
As a follow up to https://github.com/pypa/warehouse/pull/4468, we should figure out the best way to get people who have an *existing* password that has been breached/compromised to change their password to something that hasn't been reused. This is somewhat complicated by the fact that we (rightfully) cannot access the plaintext password *except* when the user is logging into the site.

A few ideas for what we could possibly do here:

* Just check on user login and send a flash message asking them to change their password.
* Causing an error to occur in the basic authentication scheme whenever a user tries to login with breached credentials.
  * This would effectively block uploads until the user changes their password to a good one.
* Trigger an email anytime the user authenticates with breached credentials, telling them their credentials have been breached, and asking them to change them.
* Flag the user account as having a compromised/breached password, and display a persistent warning in the UI (similar to unverified email).
  * One potential issue with this is that it leaks additional persistent data about the user's password into the database. If a database gets leaked, an attacker would have a list of users to focus on to try and attempt to break because they have passwords that likely exist in their password dictionary.
* Store a flag in the user's session indicating they have a compromised/breached password, and prevent accessing any other features until they rectify this situation.
  * This has similar problems to storing a flag in the database, except the storage is time limited to how long the user's session can live (I think 12 hours?), which drastically decreases the potential risk. It also lives in a separate system (currently, though there is some desire to move sessions into the database) so it would require compromising two systems instead of one.

One thing to keep in mind, is that for a lot of users, they very rarely log into the UI, and their primary interaction is using an upload client like ``twine`` so a strategy that relies entirely on signals in the UI is not likely to reach a large portion of users. Another thing to keep in mind is that in the future, we are almost certainly going to move to some form of an API key (https://github.com/pypa/warehouse/issues/994).

Given that, I think the right answer here is likely:

- Store a flag in the user's session (or perhaps the DB?) and prevent them from doing anything else in the UI, until they've changed their password.
   - If we have a flag stored in the database, maybe we want to periodically send out an email telling them to change their password? This could get spammy though.
- Block basic authentication when a compromised password is being used.
   - This effectively blocks uploads for now, however in the future when we add API keys, if someone is using an API key, it will *not* block their authentication then.
   - If we're storing the flag for the first item in the DB, then trying to authenticate with a compromised password in the database should also set that flag.

This means that immediately, having a compromised password more or less blocks your ability to do anything until you change it. However once we get API keys, you'll still be able to upload if your account has a compromised password (since presumably the API key isn't also compromised). If an attacker finds a compromised password and logs into the user's account with it, they'll be forced to change the password before they can use the account for anything else, which will trigger an email to the primary address on file indicating that the password has been changed. If they didn't change their password they'll ideally freak out appropriately and start reseting passwords and contacting us.

dstufft:
This pull request effectively disables a user if their password appears in the HIBP data until they reset their password. Whenever the user attempts to authenticate, we will check their password using the HIBP breached password service, and if it comes back as a positive hit we will disable their password, preventing any future attempts to authenticate, send them an email with information, and finally return an error ultimately preventing them from authenticating.

While we can't iterate over our database and look for any users that have compromised credentials, this will effectively do the same thing since once a password has appeared in the HIBP data, it will no longer be usable to log into PyPI.

This will be in effect both for users authenticating using the log in form in the UI and for users uploading files to PyPI.

This is a fairly abrupt interruption to the user's flow-- they are attempting to do something and we forcibly prevent them and require them to do something else first. However this is considered to be an acceptable trade off in this case, because these password are known to the general public, and can be used in [credential stuffing](https://www.owasp.org/index.php/Credential_stuffing) attacks on PyPI. This was recently the cause of a fairly large security issue on NPM, where reused credentials by one of the users for a popular package were used to upload malicious packages ([more information](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes)).

This branch is currently working, however it still needs:

- [x] A plaintext (non HTML) error message when authenticating via Basic Auth.
- [x] Tests
- [x] A solution for local development. We currently have all users set to have the password, ``password``, which doesn't work because it has been compromised. Likely we should just have a no-op breached password service in local development.
- [ ] ~~Maybe give a different error when we're failing during authentication versus when changing the password? Consistency in error message makes it easier to google the problem, but the error message is more generic and harder to know exactly what just happened.~~

I'm submitting this PR now, so that people can start to review the messaging that we're sending to users. Particularly the emails that i've written (plain text and HTML) and the additional information I've added to the FAQ.

Fixes #4471.

### DIFF
--- a/tests/unit/accounts/test_services.py
+++ b/tests/unit/accounts/test_services.py
@@ -299,6 +299,17 @@ class TestDatabaseUserService:
 
         assert found_user is None
 
+    def test_disable_password(self, user_service):
+        user = UserFactory.create()
+
+        # Need to give the user a good password first.
+        user_service.update_user(user.id, password="foo")
+        assert user.password != "!"
+
+        # Now we'll actually test our disble function.
+        user_service.disable_password(user.id)
+        assert user.password == "!"
+
 
 class TestTokenService:
     def test_verify_service(self):
--- a/tests/unit/accounts/test_services.py
+++ b/tests/unit/accounts/test_services.py
@@ -597,3 +597,36 @@ class TestHaveIBeenPwnedPasswordBreachedService:
         )
         svc = services.hibp_password_breach_factory(context, request)
         assert svc.failure_message == expected
+
+    @pytest.mark.parametrize(
+        ("help_url", "expected"),
+        [
+            (
+                None,
+                (
+                    "This password has appeared in a breach or has otherwise "
+                    "been compromised and cannot be used."
+                ),
+            ),
+            (
+                "http://localhost/help/#compromised-password",
+                (
+                    "This password has appeared in a breach or has otherwise "
+                    "been compromised and cannot be used. See "
+                    "the FAQ entry at http://localhost/help/#compromised-password for "
+                    "more information."
+                ),
+            ),
+        ],
+    )
+    def test_failure_message_plain(self, help_url, expected):
+        context = pretend.stub()
+        request = pretend.stub(
+            http=pretend.stub(),
+            find_service=lambda iface, context: {
+                (IMetricsService, None): NullMetrics()
+            }[(iface, context)],
+            help_url=lambda _anchor=None: help_url,
+        )
+        svc = services.hibp_password_breach_factory(context, request)
+        assert svc.failure_message_plain == expected
--- a/tests/unit/accounts/test_services.py
+++ b/tests/unit/accounts/test_services.py
@@ -25,6 +25,7 @@ from warehouse.accounts import services
 from warehouse.accounts.interfaces import (
     IUserService,
     ITokenService,
+    IPasswordBreachedService,
     TokenExpired,
     TokenInvalid,
     TokenMissing,
@@ -454,6 +455,11 @@ def test_token_service_factory_eq():
 
 
 class TestHaveIBeenPwnedPasswordBreachedService:
+    def test_verify_service(self):
+        assert verifyClass(
+            IPasswordBreachedService, services.HaveIBeenPwnedPasswordBreachedService
+        )
+
     @pytest.mark.parametrize(
         ("password", "prefix", "expected", "dataset"),
         [
@@ -559,7 +565,9 @@ class TestHaveIBeenPwnedPasswordBreachedService:
             }[(iface, context)],
             help_url=lambda _anchor=None: f"http://localhost/help/#{_anchor}",
         )
-        svc = services.hibp_password_breach_factory(context, request)
+        svc = services.HaveIBeenPwnedPasswordBreachedService.create_service(
+            context, request
+        )
 
         assert svc._http is request.http
         assert isinstance(svc._metrics, NullMetrics)
@@ -595,7 +603,9 @@ class TestHaveIBeenPwnedPasswordBreachedService:
             }[(iface, context)],
             help_url=lambda _anchor=None: help_url,
         )
-        svc = services.hibp_password_breach_factory(context, request)
+        svc = services.HaveIBeenPwnedPasswordBreachedService.create_service(
+            context, request
+        )
         assert svc.failure_message == expected
 
     @pytest.mark.parametrize(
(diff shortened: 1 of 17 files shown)


dstufft:
One additional thing, we probably want to announce this change to pypi-announce and to distutils-sig whenever we merge it. We probably want to get some messaging sorted out for that as well.

brainwane:
Thanks. Will take a look tomorrow.

warehouse/templates/pages/help.html
```diff
@@ -178,6 +178,9 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
           <p>
             If you receive an error message saying that "This password has appeared in a breach or has otherwise been compromised and cannot be used", you should change it all other places that you use it as soon as possible.
           </p>
+          <p>
+            If you've received this error while attempting to log in to or upload to PyPI, then your password has been reset and you will no longer be able to log into PyPI until you <a href="{{ request.route_url('accounts.reset-password') }}">reset your password</a>.
```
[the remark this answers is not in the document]
dstufft: I didn't like the way this sentence read changing you've to you, but I expanded it out to "you have".

dstufft:
Addressed the feedback by @hugovk.

tests/unit/accounts/test_services.py
```diff
@@ -299,6 +300,17 @@ def test_get_user_by_email_failure(self, user_service):
...
         assert found_user is None
 
+    def test_disable_password(self, user_service):
+        user = UserFactory.create()
+
+        # Need to give the user a good password first.
+        user_service.update_user(user.id, password="foo")
+        assert user.password != "!"
+
+        # Now we'll actually test our disble function.
```
brainwane: Should be "disable"

warehouse/accounts/services.py
```diff
@@ -363,9 +390,16 @@ def check_password(self, password, *, tags=None):
...
+@implementer(IPasswordBreachedService)
+class NullPasswordBreachedService:
+    failure_message = "This password appears in a breach."
+    failure_message_plain = "This password appears in a breach."
+
+    @classmethod
+    def create_service(cls, context, request):
+        return cls()
+
+    def check_password(self, password, *, tags=None):
```
brainwane: I think I don't quite understand under what circumstances we want to use this?

warehouse/templates/pages/help.html
```diff
@@ -178,6 +178,9 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
           <p>
             If you receive an error message saying that "This password has appeared in a breach or has otherwise been compromised and cannot be used", you should change it all other places that you use it as soon as possible.
```
brainwane: Do we need to update the wording of the quoted error message here?

warehouse/accounts/services.py
dstufft: We use it in local development, because we set all users to have the password "password", which appears in breaches. Thus without the ``NullPasswordBreachedService`` in local development, we would never be able to log into those users, and would have to generate secure password for them.

warehouse/templates/pages/help.html
dstufft: Yup, good catch.

dstufft:
Feedback from @brainwane addressed and responded to!

warehouse/templates/email/password-compromised/body.html
```diff
@@ -0,0 +1,54 @@
...
+  PyPI itself has not suffered a breach. This is a protective measure to reduce the
+  risk of <a href="https://www.owasp.org/index.php/Credential_stuffing">credential stuffing</a>
+  attacks against PyPI and its users.
+</p>
+
+
+<h3>What should I do?</h3>
+<p>
+  To regain access to your account,
+  <a href="{{ request.route_url('accounts.reset-password') }}">reset your password</a>
```
ewdurbin: I think this should be `request.route_url('accounts.request-password-reset')`

warehouse/templates/email/password-compromised/body.txt
```diff
@@ -0,0 +1,46 @@
...
+
+PyPI itself has not suffered a breach. This is a protective measure to reduce the risk
+of credential stuffing (https://www.owasp.org/index.php/Credential_stuffing) attacks
+against PyPI and its users.
+
+
+# What should I do?
+
+To regain access to your account, reset your password on PyPI
+({{ request.route_url('accounts.reset-password') }}). We also recommend that you go to
```
ewdurbin: I think this should be `request.route_url('accounts.request-password-reset')`

warehouse/templates/pages/help.html
```diff
@@ -176,7 +176,10 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
             PyPI will not allow such passwords to be used when setting a password at registration or updating your password.
           </p>
           <p>
-            If you receive an error message saying that "This password has appeared in a breach or has otherwise been compromised and cannot be used", you should change it all other places that you use it as soon as possible.
+            If you receive an error message saying that "This password appears in a breach or has been compromised and cannot be used", you should change it all other places that you use it as soon as possible.
+          </p>
+          <p>
+            If you have received this error while attempting to log in or upload to PyPI, then your password has been reset and you cannot log in to PyPI until you <a href="{{ request.route_url('accounts.reset-password') }}">reset your password</a>.
```
ewdurbin: I think this should be `request.route_url('accounts.request-password-reset')`

ewdurbin:
Noticed that we're sending folks to the wrong URL in the emails/FAQ. This route flashes "Invalid token: no token supplied", as indeed no token is supplied.

dstufft:
Going to go ahead and merge this once tests pass. I believe that all of the feedback was addressed. If there are any further changes we can address that in a future PR.

PR #4553, opened 2018-08-14
xmlrpc ratelimiting

ewdurbin:
rate limit the xmlrpc endpoint per IP

### DIFF
--- a/warehouse/legacy/api/xmlrpc/views.py
+++ b/warehouse/legacy/api/xmlrpc/views.py
@@ -23,6 +23,7 @@ import typeguard
 
 from elasticsearch_dsl import Q
 from packaging.utils import canonicalize_name
+from pyramid.httpexceptions import HTTPTooManyRequests
 from pyramid.view import view_config
 from pyramid_rpc.mapper import MapplyViewMapper
 from pyramid_rpc.xmlrpc import (
@@ -45,6 +46,7 @@ from warehouse.packaging.models import (
     Role,
     release_classifiers,
 )
+from warehouse.rate_limiting import IRateLimiter
 from warehouse.search.queries import SEARCH_BOOSTS
 
 # From https://stackoverflow.com/a/22273639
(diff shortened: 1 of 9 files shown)


warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -62,6 +64,28 @@ def wrapped(context, request):
...
+        def wrapped(context, request):
+            ratelimiter = request.find_service(
+                IRateLimiter, name="xmlrpc.client", context=None
+            )
+            if not ratelimiter.test(request.remote_addr):
+                raise XMLRPCWrappedError(
+                    HTTPTooManyRequests(
+                        "The action could not be performed because there were too "
+                        "many requests by the client. Limit resets in "
+                        f"{ratelimiter.resets_in(request.remote_addr)}"
```
dstufft: If I remember correctly, this is when the limit *might* reset. A `RateLimiter` can contain multiple identifiers, and thus this value is meant to be advisory only. It also might be ``None`` if there is a problem talking to redis for some reason.

warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -108,6 +110,33 @@ def wrapped(context, request):
...
+            if not ratelimiter.test(request.remote_addr):
+                metrics.increment("warehouse.xmlrpc.ratelimiter.exceeded", tags=[])
+                message = (
+                    "The action could not be performed because there were too "
+                    "many requests by the client."
+                )
+                _resets_in = ratelimiter.resets_in(request.remote_addr)
+                if _resets_in is not None:
+                    _resets_in = int(_resets_in.total_seconds())
+                    message += f" Limit may reset in {_resets_in} seconds."
```
di, line 130:
```suggestion
                    message += f" Limit will reset in {_resets_in} seconds."
```
Is there a case where the limit won't reset in that time?

di: D'oh, seeing Donald's comment now. Still seems like a weird error message though.

ewdurbin: The `may` is for the case where we configure a compound rate limit like "1 per second;30 per minute;900 per hour" as supported by our limit library. The resets_in function returns the first limit exceeded (or None) as it is currently implemented.

di: Since we're only using the one identifier (IP address) I think it's probably OK to say "will reset" here.

ewdurbin: it's not a matter of IP, but the multiple limits that may be enforced.

ewdurbin: see https://limits.readthedocs.io/en/stable/string-notation.html

di: Got it, didn't know we could compound the limit notation like that. 👍

pfmoore:
Can the actual limits be documented somewhere?

coogor:
Except that the limit has changed to 1 second, still he same issue....

PR #4581, opened 2018-08-17
Cleanup a lot of legacy cruft from the database schema

dstufft:
Goes through and kills a lot of unused columns that we've accumulated over a decade. It does *not* remove them from the database (that will be a follow up PR) because that would cause the currently running code to explode when these migrations run.


warehouse/migrations/versions/62f6becc7653_set_user_is_staff_to_nullable.py
```diff
@@ -0,0 +1,32 @@
...
+
+from alembic import op
+
+
+revision = "62f6becc7653"
+down_revision = "522918187b73"
+
+
+def upgrade():
+    op.alter_column("accounts_user", "is_staff", nullable=True)
```
ewdurbin: any reason we aren't including this drop in the second "pass"?

dstufft: What drop? This doesn't drop the column, it's changing it to allow ``NULL`` values. Without this inserting a user breaks because there's no default value on this column, and SQLAlchemy will no longer be inserting a value here with this PR.

PR #4585, opened 2018-08-17
[WIP] Periodically prune unused users

dstufft:
This is just an idea, basically we get a fair amount of spam user signups who never go on to do anything meaningful with their account. I'm just throwing this PR up here for some discussion on whether this is a good idea or not.

This will currently delete about 102889 users on the production database, again to reiterate these are users who never activated, uploaded, or were added to a project.

If we did this, we would probably want to adjust the language of the verification email (at least for the first one) to say that if the user doesn't verify their email, they will be deleted after some period of time.

/cc @pypa/warehouse-committers

### DIFF
--- a/warehouse/accounts/__init__.py
+++ b/warehouse/accounts/__init__.py
@@ -12,6 +12,7 @@
 
 import datetime
 
+from celery.schedules import crontab
 from pyramid.authorization import ACLAuthorizationPolicy
 from pyramid_multiauth import MultiAuthenticationPolicy
 
@@ -31,6 +32,7 @@ from warehouse.accounts.auth_policy import (
     BasicAuthAuthenticationPolicy,
     SessionAuthenticationPolicy,
 )
+from warehouse.accounts.tasks import prune_unused_users
 from warehouse.errors import BasicAuthBreachedPassword
 from warehouse.email import send_password_compromised_email
 from warehouse.rate_limiting import RateLimit, IRateLimiter
@@ -156,3 +158,6 @@ def includeme(config):
     config.register_service_factory(
         RateLimit("1000 per 5 minutes"), IRateLimiter, name="global.login"
     )
+
+    # Add our periodic user prune task
+    config.add_periodic_task(crontab(minute=0, hour=0), prune_unused_users)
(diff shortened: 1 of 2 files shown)


brainwane:
Help me understand why we would do this, and what your suggested age threshold would be?

dstufft:
So the basic reason for it is because we have a bunch of spam accounts getting registered to the site. These accounts thus far are basically not doing anything, but they show up on user pages and such. A common pattern for these is that they never activate their account or verify their email address.

As far as what the benefit is, when we're looking at making changes to the site, we often try to gauge impact. These accounts make it harder to accurately gauge impact, because you have to carefully construct your queries to exclude them. They also do take up some minor amount of resources since they take up disk space and the like. It also frees up the user name for someone else who might want it, and will actually do something with it.

As far as why we *wouldn't* do it... the main reasons I can think of is if someone registered an account to "claim" the name, and then never logged in or did anything with it, but they still want that name claimed. It could also be a less optimal experience for someone who comes back some time after having tried to create an account but not finishing the process, to find their account is no longer there.

As far as time frame, this PR has it set to 90 days. I suspect the number of people who come back after 90 days to try and activate an account they created > 90 days ago is approaching zero... but I'm not really married to that time frame either. I just sort of arbitrarily picked it.

miketheman:
Resurfacing this as a conversation topic.
Depending on the future of PyPI user management, I can see a benefit in here to either remove these accounts completely, or even disable them from logging in without a recovery process and (hopefully) setting up 2FA.

Some questions I thought of that might facilitate the overall outcome:

- Would make sense to reevaluate "the purge" on a 12-month history, and add a disable for the 90-day period?

- Should we add a notice during account creation/email verification of "if you don't verify your email in 90 days, the account may get recycled" to set the expectations for new folks?

- Should we perform one mass mail of the purge candidates to warn them of impending removal, despite not having verified their account, and give them 30/60/90 days to accomplish that?

miketheman:
For prior art, I dug up the npm TOS blog change from back in July 2017: https://blog.npmjs.org/post/163187838935/upcoming-change-verified-email-required.html

miketheman:
#Triage 

In conversation with Admins today, this is probably still aa good idea, but should not proceed in the current implementation.
Closing for now, but reserving the right to pursue this again in the future.

PR #4627, opened 2018-08-27
redirect trailing slash on JSON API (fixes #4016)

### LINKED ISSUE #4016: Add trailing slash redirects to JSON API
The JSON API currently resolves URLs like `https://pypi.org/pypi/pip/json`, however adding a trailing slash (e.g. `https://pypi.org/pypi/pip/json/`) results in a 404, even though trailing slashes were valid in legacy PyPI.

PR https://github.com/pypa/warehouse/pull/3971 was an attempt to fix this, but because we have Pyramid configured to automatically add a trailing slash in the event of a 404, this resulted in a redirect loop.

stloma:
Fixes #4016 

My thought with fixing this issue was creating a route for the URLs with trailing slashes that redirect back to the non-slash URL.

Obviously just creating a redirect would be ideal. However, as noted in #4012, if a package does not exist, a redirect loop would occur since `@notfound_view_config(append_slash=HTTPMovedPermanently)` appends slashes to URLs when packages do not exist.

Here is what the request-response cycle looks like now (it works the same for release versions):

User requests `https://pypi.org/pypi/pip/json`, receives `200`.

User requests `https://pypi.org/pypi/pip/json/`, receives `301 https://pypi.org/pypi/pip/json`.

User requests `https://pypi.org/pypi/badpackage/json`, receives `301 https://pypi.org/pypi/badpackage/json/`.
User requests `https://pypi.org/pypi/badpackage/json/`, receives `404`

Note that the last `404` comes directly from `https://pypi.org/pypi/badpackage/json/`, not `https://pypi.org/project/badpackage/json` as it used to. I could not find any reason that would cause an issue, but thought I'd mention it in case I'm not aware of something.


warehouse/legacy/api/json.py
```diff
@@ -75,18 +77,24 @@ def json_project(project, request):
...
 
+@view_config(
+    route_name="legacy.api.json.project_slash",
+    context=Project,
+    decorator=_CACHE_DECORATOR,
+)
+def json_project_slash(project, request):
+    return HTTPMovedPermanently(
+        # Respond with redirect to url without trailing slash
+        request.current_route_path(name=project.name)[:-1],
```
di: I think this should probably be something like this instead:

```python
request.route_path("legacy.api.json.project", name=project.name)
```

warehouse/legacy/api/json.py
```diff
@@ -180,3 +188,16 @@ def json_release(release, request):
...
+
+@view_config(
+    route_name="legacy.api.json.release_slash",
+    context=Release,
+    decorator=_CACHE_DECORATOR,
+)
+def json_release_slash(project, request):
+    return HTTPMovedPermanently(
+        # Respond with redirect to url without trailing slash
+        request.current_route_path(name=project.name)[:-1],
```
di: Same here, but with `"legacy.api.json.release"` instead.

warehouse/legacy/api/json.py
stloma: Good call

stloma:
The redirects now rely on existing routes.

Thanks for the review and please let me know if there's anything else!

tests/unit/legacy/api/test_json.py
```diff
@@ -118,6 +118,22 @@ def test_only_prereleases(self, monkeypatch, db_request):
         assert json_release.calls == [pretend.call(release, db_request)]
 
 
+class TestJSONProjectSlash:
+    def test_normalizing_redirects(self, db_request):
+        project = ProjectFactory.create()
+
+        name = project.name
+
+        db_request.matchdict = {"name": name}
```
di: Is this actually required for the test to pass? I don't think it is.

tests/unit/legacy/api/test_json.py
```diff
@@ -118,6 +118,22 @@ def test_only_prereleases(self, monkeypatch, db_request):
...
+        db_request.route_path = pretend.call_recorder(
+            lambda *a, **kw: "/project/the-redirect"
+        )
+
+        resp = json.json_project_slash(project, db_request)
+
+        assert isinstance(resp, HTTPMovedPermanently)
```
di: Can you add some more assertions here? Something like:

```python
assert db_request.route_path.calls == [
    call("legacy.api.json.project", name=project.name),
]
```

tests/unit/legacy/api/test_json.py
```diff
@@ -428,3 +444,19 @@ def test_minimal_renders(self, pyramid_config, db_request):
...
         }
+
+
+class TestJSONReleaseSlash:
+    def test_normalizing_redirects(self, db_request):
+        release = ReleaseFactory.create()
+
+        name = release.name
+
+        db_request.matchdict = {"name": name}
```
di: Same here.

tests/unit/legacy/api/test_json.py
```diff
@@ -428,3 +444,19 @@ def test_minimal_renders(self, pyramid_config, db_request):
...
+        db_request.route_path = pretend.call_recorder(
+            lambda *a, **kw: "/project/the-redirect"
+        )
+
+        resp = json.json_release_slash(release, db_request)
+
+        assert isinstance(resp, HTTPMovedPermanently)
```
di: Same here.

tests/unit/legacy/api/test_json.py
di: @stloma Can you add the assertion on `db_request.route_path.calls` as well?

stloma: Ah, shoot, sorry I missed that :(

di:
I'm a little concerned about the change in existing behavior here, namely that `https://pypi.org/pypi/badpackage/json` will return a 301 instead of a 404 as it does now.

@stloma Do you think we change the behavior to be:

* `https://pypi.org/pypi/pip/json` -> `200`
* `https://pypi.org/pypi/pip/json/` ->  `301 https://pypi.org/pypi/pip/json`
* `https://pypi.org/pypi/badpackage/json` -> 404
* `https://pypi.org/pypi/badpackage/json/` -> 404

stloma:
I think there's a slight misunderstanding in my original comment. `https://pypi.org/pypi/badpackage/json` has always returned a 301 due to this code:

https://github.com/pypa/warehouse/blob/da5ca10a8f56fd33685ef3c0b46aec89154412f2/warehouse/views.py#L58

The difference I noted from the original behavior is that `https://pypi.org/pypi/badpackage/json/` returns a `404` now, instead of a `301` to `https://pypi.org/project/badpackage/json/`

Here's the current behavior on `master`:

`https://pypi.org/pypi/badpackage/json` -> `301 https://pypi.org/pypi/badpackage/json/`
`https://pypi.org/pypi/badpackage/json/` -> `301 https://pypi.org/project/badpackage/json/`
`https://pypi.org/project/badpackage/json/` -> `404`

Here's what this PR does:

`https://pypi.org/pypi/badpackage/json` -> `301 https://pypi.org/pypi/badpackage/json/`
`https://pypi.org/pypi/badpackage/json/` -> `404`

So the last 404 comes directly from `https://pypi.org/pypi/badpackage/json/`, not `https://pypi.org/project/badpackage/json/`.

di:
Thanks @stloma!

PR #4671, opened 2018-09-05
Raise 404 instead of return

di:
Raise a `HTTPNotFound` instead of returning it so it can get caught by our exception handler and a nice 404 page will be rendered instead.

### DIFF
--- a/tests/unit/packaging/test_views.py
+++ b/tests/unit/packaging/test_views.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import pretend
+import pytest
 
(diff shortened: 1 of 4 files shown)


dstufft:
Not really required for this, but I wonder if we can lint for returning a 404 instead of raising one somehow?

di:
Considered it, there's [at least one place where we really do want to return a `HTTPNotFound`](https://github.com/pypa/warehouse/blob/b66d92985d42de2b2554ff7872ce4723860e1bd5/warehouse/legacy/api/json.py#L73) so I'm not sure it's worth the effort.

Issue #3836, opened 2018-04-26
Dev env setup guide should document forking and cloning as the preferred workflow.


kchawla-pi:
In the guide to setting up a development environment for new contributors, the suggested workflow is to start by cloning the repo locally. 
https://warehouse.readthedocs.io/development/getting-started/#getting-the-warehouse-source-code

It should be  changed to-
1. Fork the repo.
2. Clone the fork locally.
3. Add the original repo as a second remote with a different name (upstream).
4. Pull from upstream master regularly to keep the forked master current with the project master.

This is a better workflow as it gives contributors a place to push their code to remote while staying current with the upstream master.

@brainwane concurs.

kchawla-pi:
I will work on this and submit a PR.

kchawla-pi:
I may not be able to attend to this. Please feel free to reassign it.

PR #4702, opened 2018-09-11
Add remote and sync in docs (#3836)

stloma:
Fixes: #3836 

Add documentation for keeping local repository synced with upstream.


ewdurbin:
Thank you @stloma!

PR #4752, opened 2018-09-22
Align JSON API and XML-RPC timestamps

### LINKED ISSUE #4751: JSON API omits timezone and microseconds from upload_time
When a request is made to `https://pypi.org/pypi/$PROJECT/json` or `https://pypi.org/pypi/$PROJECT/$VERSION/json`, the `"upload_time"` fields in the response contain no timezone information and has no sub-second precision (e.g., `"2018-09-22T19:16:05"`).  Compare this to the XML-RPC API, in which the `"upload_time"` fields in the data returned by the `release_urls` method all end with `Z` and have microsecond precision (e.g., `"2018-09-22T19:16:05.067954Z"`).

jwodder:
Fixes #4751 (I think).


dstufft:
The change looks generally OK to me, though I am concerned that this is going to break people who had manually written parse strings for the JSON date.

PR #4782, opened 2018-09-28
Match Codecov and Coveralls for the check icon

hugovk:
Follow on from https://github.com/pypa/warehouse/pull/4780.

# CI

* In addition to code build sites, add code coverage sites
* There's no icons for these, but I think the check/tick is a good match as it's used for example on GitHub for code checks.

image

image

https://fontawesome.com/icons/check?style=solid


ewdurbin:
Thanks again @hugovk

PR #4792, opened 2018-10-01
Change CI icon to tasks list

hugovk:
Further to https://github.com/pypa/warehouse/pull/4780#issuecomment-425795784, the check mark looked a bit too much like PyPI is reporting on the success of CI runs, which it isn't.

Instead, use the "tasks" icon, which would look like this:

image

https://pypi.org/project/aiohttp/

This also adds Shippable to the list of CIs, as found by chance on that list.


ewdurbin:
Thanks again @hugovk

PR #4795, opened 2018-10-02
Make container wide enough to no wrap buttons

Carreau:
At least for me on OSX/Chrome, this leads to
https://pypi.org/manage/projects/ view with button next to each other
instead of below each other. Which seem to look more in line with the
design.

Otherwise I guess "Manage" and "View" should have the same width.

After

screen shot 2018-10-02 at 14 41 02

--- 

Before
screen shot 2018-10-02 at 14 40 51


nlhkabu:
Closing in favour of https://github.com/pypa/warehouse/pull/4778
Thanks for the PR @Carreau :)

PR #4838, opened 2018-10-10
Fix mistakes in XML-RPC docs

jwodder:
This PR fixes two mistakes in [the docs on the XML-RPC API](https://warehouse.readthedocs.io/api-reference/xml-rpc/):

* The docs claim that the `show_hidden` flag of `package_releases()` is now ignored.  In actuality, it now controls whether to return the latest version or all versions.

* The docs for `changelog_since_serial()` claim that the argument is a Unix timestamp.  It is not; it's a serial ID.


ewdurbin:
Thank you @jwodder! I'll review and merge once I finish handling @pyup-bot PRs!

ewdurbin:
@jwodder one question: Is your use case of XMLRPC a new implementation? We desperately want to deprecate XMLRPC and I'd love to hear about your use case if it is new! ernest@python.org

jwodder:
Yes, it's a new implementation.  I'm using the XML-RPC API to:

* Get a list of all project names from PyPI (not done often)
* Read the changelog to keep a local list of wheels in sync with PyPI (because the RSS feed doesn't give details on removals or wheels added to releases late, and the feeds only go back a few hours).

ewdurbin:
@jwodder thank you for the details. currently the changelog is only available via XMLRPC so we can't really offer anything different.

Would hitting `/simple` to get the list of projects be acceptable? The only caveat is that projects with no releases will not be listed, though for your use case that doesn't seem problematic. That API is well specified in [PEP 503](https://www.python.org/dev/peps/pep-0503/) and can be cached by our CDN, while the XMLRPC request is one of the more costly ones and must be handled by our backends.

jwodder:
My experience has been that fetching & parsing `/simple` is generally slower than using the XML-RPC API.  I will consider switching, though.

ewdurbin:
Thank you @jwodder! Regarding parsing the Simple API vs just using the XMLRPC response, you're certainly right, if you're doing it infrequently it might be a decent bit of future proofing until a new PyPI API comes along and it would avoid making the costly backend request. It's certainly your call, but we'd appreciate you hitting our CDN (simple) instead of our backend (XMLRPC).

PR #4845, opened 2018-10-11
Tweak feedback FAQ

di:
Revised per https://github.com/pypa/warehouse/issues/4844#issuecomment-429101267.

We also don't need the specific bug report items because that's all in the issue template now.


ewdurbin:
@ghjwp7 thanks for the excellent feedback that led to this!

PR #4877, opened 2018-10-17
Don't pin to prereleases

di:
Pinning to a prerelease causes pip-compile to think that we always will
accept a prerelease for a given package, which will potentially cause
the dependency checker to fail if subdependencies change between a
latest release and prerelease.

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -30,7 +30,7 @@ passlib>=1.6.4
 premailer
 psycopg2
 pycurl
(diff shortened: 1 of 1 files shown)


di:
Even after changing this pin to `pyramid>=1.9` this failed, something about the `pyramid` package is causing `pip-compile` to use the prerelease, I had to explicitly pin this to `pyramid>=1.9,<=1.9.2`

PR #4880, opened 2018-10-17
Render more project links as docs

nijel:
Consider domains starting with docs. as documentation as well.


nlhkabu:
Thanks @nijel :tada:

PR #4941, opened 2018-10-27
Include '--index-url' argument for the pip install command for Test PyPI

yeraydiazdiaz:
@krf2 thanks for [your PR](https://github.com/pypa/warehouse/pull/4044), as part of the Bloomberg sprint I picked it up, but is a little out of date and it felt easier to just cherry pick your commit on this PR, I hope you don't mind.

This is what it looks like atm:

screen shot 2018-10-27 at 11 50 23

Looks a bit weird imho maybe we should reduce the font size? @nlhkabu what do you think?


yeraydiazdiaz:
Builds are failing due to missing itsdangerous==1.0.0, fixed in https://github.com/pypa/warehouse/pull/4940

di:
> Looks a bit weird imho maybe we should reduce the font size?

I agree it's a bit weird, but I think I'd rather have it look a bit weird than have a smaller font.

Also since this is only on TestPyPI and [eventually TestPyPI will go away](https://github.com/pypa/warehouse/issues/918), I think it's OK!

yeraydiazdiaz:
It's not that bad with `-i`:

screen shot 2018-10-27 at 15 48 24

might depend on the name of the package though

PR #4957, opened 2018-10-27
Add comment about CSS used in pypi-theme

di:


di:
@nlhkabu This style was confusing for a contributor working on SCSS auditing because it seemed like it wasn't being used in the codebase (because they couldn't see the `pypi-theme` codebase), they were proposing removing this style and it took me a while to realize why it seemed like it was unused.

nlhkabu:
Ok, makes sense. I just re-read it - even if we use this style again in the codebase, this comment will still apply - thanks for the PR :)

Issue #4772, opened 2018-09-27
Migrate away from using natural primary keys


di:
A number of our tables currently have natural primary keys, such as:

* The `packages` table (used by the `Project` model)
* The `releases` table (used by the `Release` model)
* (...this may be all of them)

We should migrate away from natural keys and use surrogate keys instead where appropriate, e.g.:

```python
id = Column(Integer, primary_key=True, nullable=False)
```

This will require a DB migration and possibly a maintenance period.

steiza:
I'm going to start working on this as part of the PyPA sprints. If anyone is interested in pairing, let me know!

ewdurbin:
Ah, this is already pretty far along in my court @steiza. Given that it will require an admin to roll out (potential downtime, db backups, etc). I'll continue on it!

yeraydiazdiaz:
@ewdurbin is there anything else pending for this issue or should can it be closed?

PR #4958, opened 2018-10-27
Major Database Refactoring

ewdurbin:
working on resolving #4772


ewdurbin:
Something goofy is happening with the ProjectFactory! @steiza @di if you want to take a look I'd love any input you might have:

```
================================================================================ FAILURES ================================================================================
__________________________________________________________________________ TestIndex.test_index __________________________________________________________________________

self = <sqlalchemy.orm.attributes.ScalarAttributeImpl object at 0x7f18f98232a8>, state = <sqlalchemy.orm.state.InstanceState object at 0x7f18fa324a90>
dict_ = {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x7f18fa324a90>, 'id': None, 'name': 'MyEkoMlCQjXW'}, passive = symbol('PASSIVE_OFF')

    def get(self, state, dict_, passive=PASSIVE_OFF):
        """Retrieve a value from the given object.
        If a callable is assembled on this object's attribute, and
        passive is False, the callable will be executed and the
        resulting value will be set as the new value for this attribute.
        """
        if self.key in dict_:
            return dict_[self.key]
        else:
            # if history present, don't load
            key = self.key
            if key not in state.committed_state or \
                    state.committed_state[key] is NEVER_SET:
                if not passive & CALLABLES_OK:
                    return PASSIVE_NO_RESULT
    
                if key in state.expired_attributes:
                    value = state._load_expired(state, passive)
                elif key in state.callables:
                    callable_ = state.callables[key]
                    value = callable_(state, passive)
                elif self.callable_:
                    value = self.callable_(state, passive)
                else:
                    value = ATTR_EMPTY
    
                if value is PASSIVE_NO_RESULT or value is NEVER_SET:
                    return value
                elif value is ATTR_WAS_SET:
                    try:
>                       return dict_[key]
E                       KeyError: 'normalized_name'

../lib/python3.6/site-packages/sqlalchemy/orm/attributes.py:609: KeyError

During handling of the above exception, another exception occurred:

self = <tests.unit.test_views.TestIndex object at 0x7f18fa28a320>, db_request = <pyramid.testing.DummyRequest object at 0x7f18fa33e7b8>

    def test_index(self, db_request):
    
>       project = ProjectFactory.create()

tests/unit/test_views.py:181: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../lib/python3.6/site-packages/factory/base.py:563: in create
    return cls._generate(enums.CREATE_STRATEGY, kwargs)
../lib/python3.6/site-packages/factory/base.py:500: in _generate
    return step.build()
../lib/python3.6/site-packages/factory/builder.py:279: in build
    kwargs=kwargs,
../lib/python3.6/site-packages/factory/base.py:314: in instantiate
    return self.factory._create(model, *args, **kwargs)
tests/common/db/base.py:35: in _create
    session.flush()
../lib/python3.6/site-packages/sqlalchemy/orm/scoping.py:153: in do
    return getattr(self.registry(), name)(*args, **kwargs)
../lib/python3.6/site-packages/sqlalchemy/orm/session.py:2254: in flush
    self._flush(objects)
../lib/python3.6/site-packages/sqlalchemy/orm/session.py:2380: in _flush
    transaction.rollback(_capture_exception=True)
../lib/python3.6/site-packages/sqlalchemy/util/langhelpers.py:66: in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
../lib/python3.6/site-packages/sqlalchemy/util/compat.py:249: in reraise
    raise value
../lib/python3.6/site-packages/sqlalchemy/orm/session.py:2348: in _flush
    self.dispatch.after_flush(self, flush_context)
../lib/python3.6/site-packages/sqlalchemy/event/attr.py:246: in __call__
    fn(*args, **kw)
warehouse/cache/origin/__init__.py:41: in store_purge_keys
    purges.update(key_maker(obj).purge)
warehouse/cache/origin/__init__.py:121: in key_maker
    cache=[k.format(obj=obj) for k in cache_keys],
warehouse/cache/origin/__init__.py:121: in <listcomp>
    cache=[k.format(obj=obj) for k in cache_keys],
../lib/python3.6/site-packages/sqlalchemy/orm/attributes.py:242: in __get__
    return self.impl.get(instance_state(instance), dict_)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sqlalchemy.orm.attributes.ScalarAttributeImpl object at 0x7f18f98232a8>, state = <sqlalchemy.orm.state.InstanceState object at 0x7f18fa324a90>
dict_ = {'_sa_instance_state': <sqlalchemy.orm.state.InstanceState object at 0x7f18fa324a90>, 'id': None, 'name': 'MyEkoMlCQjXW'}, passive = symbol('PASSIVE_OFF')

    def get(self, state, dict_, passive=PASSIVE_OFF):
        """Retrieve a value from the given object.
        If a callable is assembled on this object's attribute, and
        passive is False, the callable will be executed and the
        resulting value will be set as the new value for this attribute.
        """
        if self.key in dict_:
            return dict_[self.key]
        else:
            # if history present, don't load
            key = self.key
            if key not in state.committed_state or \
                    state.committed_state[key] is NEVER_SET:
                if not passive & CALLABLES_OK:
                    return PASSIVE_NO_RESULT
    
                if key in state.expired_attributes:
                    value = state._load_expired(state, passive)
                elif key in state.callables:
                    callable_ = state.callables[key]
                    value = callable_(state, passive)
                elif self.callable_:
                    value = self.callable_(state, passive)
                else:
                    value = ATTR_EMPTY
    
                if value is PASSIVE_NO_RESULT or value is NEVER_SET:
                    return value
                elif value is ATTR_WAS_SET:
                    try:
                        return dict_[key]
                    except KeyError:
                        # TODO: no test coverage here.
                        raise KeyError(
                            "Deferred loader for attribute "
                            "%r failed to populate "
>                           "correctly" % key)
E                       KeyError: "Deferred loader for attribute 'normalized_name' failed to populate correctly"

../lib/python3.6/site-packages/sqlalchemy/orm/attributes.py:615: KeyError
```

ewdurbin:
cc @dstufft

ewdurbin:
If we can figure out what's going on with `normalized_name` this is not far from done. Need to wrap up `Role` relation to `Project`.

steiza:
Ah, I think the mocks in `tests/` need to be updated. I did something like this:

```
$ git diff
diff --git a/tests/common/db/packaging.py b/tests/common/db/packaging.py
index cd72f6d5..84ab5a55 100644
--- a/tests/common/db/packaging.py
+++ b/tests/common/db/packaging.py
@@ -12,6 +12,7 @@
 
 import datetime
 import hashlib
+import uuid
 
 import factory
 import factory.fuzzy
@@ -36,6 +37,7 @@ class ProjectFactory(WarehouseFactory):
     class Meta:
         model = Project
 
+    id = uuid.uuid4()
     name = factory.fuzzy.FuzzyText(length=12)
```

.. and of course then it choked on the mock `ReleaseFactory`. But I think that's the right path.

ewdurbin:
Dang! Good Catch @steiza! Thank you!

di:
@ewdurbin Are you auto-generating the migration from the models? Looks to me like there's a bit of disagreement between the two.

ewdurbin:
@di not completely no. there is currently some disagreement because I wanted to get what I had up for review before departing for a bit! fixing tests and syncing now!

ewdurbin:
I think that resolves most of the schema and model concerns. Now it's a matter of cleaning up the newly goof'd references. I plan to pick this up tomorrow morning if no-one runs with it.

Contributors: Feel free to pick branch from this and add commits in a separate PR! Just note that you've done so and reference this pull request. I'll happily close and pick up from your work in the morning :)

ewdurbin:
Could use some help pushing this along. There's something I don't seem to understand in the relationship between `Dependency` and `Release`. I'm not sure if that's a modeling issue or an issue with our test factories.

warehouse/packaging/models.py
```diff
@@ -65,13 +62,18 @@ class Role(db.Model):
...
+    project_id = Column(
+        ForeignKey("packages.id", onupdate="CASCADE", ondelete="CASCADE"),
+        nullable=False,
     )
 
     user = orm.relationship(User, lazy=False)
     project = orm.relationship("Project", lazy=False)
 
+    @property
+    def package_name(self):
```
dstufft: This is a problem, because we are using ``Role.project_name`` in queries as part of an ``ORDER BY``, see this traceback for ``TestUserDetail.test_gets_user``.

The correct solution depends on why we're ordering by ``Role.project_name``.. if the goal is just to provide a consistent ordering, we can change it to order by ``Role.project_id`` instead. If the goal is to explicitly order it by the project name, then you'll probably want to turn this into a [Hybrid Property](https://docs.sqlalchemy.org/en/latest/orm/extensions/hybrid.html) or a [Column Property](https://docs.sqlalchemy.org/en/latest/orm/mapped_sql_expr.html#mapper-column-property-sql-expressions). Another option would be to adjust the query to do a join manually, and then order by ``Project.name`` instead of ``Role.package_name``.

warehouse/packaging/models.py
```diff
@@ -65,13 +62,18 @@ class Role(db.Model):
...
+        return project.name
```
dstufft: This looks just wrong, ``project`` isn't defined here. Presumably you meant ``self.project.name``?

dstufft:
I've fixed a few more of the tests. I think the remaining tests exist because we still have ``Release.name``, ``File.name``, ``File.version``, but nothing is populating them. I see three possible options for fixing it:

1. Adjust all of the queries we make, to pass those values in.
2. Add a trigger in the database to automatically populate those columns from the referenced parent whenever the value is ``null``.
3. Remove ``Release.name``, ``File.name``, and ``File.version``, and update all of our queries to use joins to get that information from the normalized location.

dstufft:
Oh, and (1) and (2) will need something to make sure that if we update ``Project.name`` or ``Release.version``, it gets updated in the references ``Release`` and/or ``File`` objects.

(3) is maybe the right answer, but we'd have to see how it performed, doing those joins might increase the query duration by a non-trivial amount. I have no idea.

dstufft:
Error here is that this PR needs updated to deal with https://github.com/pypa/warehouse/blob/b5be03eb54f2f48ef9d26a2aaf9a9a687aa32e91/warehouse/migrations/versions/eeb23d9b4d00_add_squats_table.py.

dstufft:
I'm hijacking this PR to include a number of other changes that I think we should do, but I've avoided doing due to downtime they would likely require. Since this PR is going to require downtime already, merging all of these changes into a singe PR allows us to just take the downtime once.

dstufft:
Ok, so the effect of this PR:

* Add a ``Project.id`` and ``Release.id`` column.
* Replaces all of the FKs to ``Project`` and ``Release`` to use that new ``id`` column.
* Deletes the ``Admin`` role, which is no longer being used.
* Updates ``Role.user``, ``Release.uploader``, and ~~``JournalEntry.submitted_by``~~ to FK using ``User.id`` instead of ``User.username``.
  * In the case of ``Release.uploader``, this includes removing a gnarly join and just setting the uploader at upload time.
* Renames ``packages`` to ``projects``.
* Renames ``accounts_user`` to ``users``.
* Renames ``accounts_email`` to ``user_emails``.
* Renames ``warehouse_admin_*`` to ``admin_*``.

Replacing all uses of things like ``Release.name`` to be ``Release.project.name`` means we're now triggering joins (or extra queries) when we weren't previously. I've tried to go through and spot check to see if we've introduced any of those, and fixed any I found. However it's possible that some will still have been missed. We'll want to monitor performance after this to make sure we're not degrading our perf.

I expect this set of migrations will take a bit to run. On the dev db, my ``make initdb`` is taking about 10 minutes, and the production database is quite a bit larger.

dstufft:
Maybe, there's something on my calendar, but I'm not sure what it is or if I have to go to it. Will have to check and get back.

di:
I'll be at PyCon CA that weekend -- Sunday would be better for me.

ewdurbin:
Sunday is toast for me. Committed all afternoon.

@dstufft if something a bit later works for you Saturday I'm good all afternoon (12PM-5PM Eastern)

ewdurbin:
Scheduled.

https://status.python.org/incidents/vhyxtc7cvf9g

ewdurbin:
Deployment to prod disabled temporarily. Merging to send to test.pypi.org.

Issue #3939, opened 2018-05-10
Remove bold from markdown code sections


cprogrammer1994:
Hi!

I think bold looks bad in the rendered version of the long description.
It appears in the code section for imports and numbers.

Thanks!

**Describe the bug**

bold looks bad, especially for numbers

![image](https://user-images.githubusercontent.com/11232402/39877243-1cdee440-547e-11e8-9130-0a470d49ce52.png)

**Expected behavior**

- do not use bold at all

---

> ### Here is how github renders it:

## Sample usage

```py
>>> import moderngl
>>> ctx = moderngl.create_standalone_context()
>>> buf = ctx.buffer(b'Hello World!')  # allocated on the GPU
>>> buf.read()
b'Hello World!'
```

For complete examples please visit the [Examples](https://github.com/cprogrammer1994/ModernGL/tree/master/examples/#readme).

## Easy to use with Pillow and Numpy

```py
>>> img = Image.open('texture.jpg')
>>> ctx.texture(img.size, 3, img.tobytes())
<Texture: 1>
```

```py
>>> ctx.buffer(np.array([0.0, 0.0, 1.0, 1.0], dtype='f4'))
<Buffer: 1>
```

nlhkabu:
It might be worth doing #4400 before starting on this ticket.

mixmastamyk:
Hmm, I tend to prefer bold for keywords, though it does seem a excessive for numbers.

Believe part of the problem is because of the difference in weights between normal and bold of this fonts.  The bold is a bit *too* bold. 

If you are using a custom font with several weights available, one can be selected with a number in css: `font-weight: 600` for example.

gforcada:
Is there anything missing here after #4959 was merged? :thinking:

di:
Nope!

PR #4959, opened 2018-10-27
Fix bold in numbers

vaguiar:
Fixing bold in numbers which was reported in #3939 

cutelog
abysmal


di:
@vaguiar and I spoke and we're going to remove bold fonts from all the syntax highlighting, not just numbers.

vaguiar:
as per the comment above, I've removed the bold fonts everywhere.

cutelog2
abysmal2

di:
Thanks @vaguiar!

PR #4960, opened 2018-10-27
accessibility fix: remove italic styles from html and css

jpurviance:
Fixes https://github.com/pypa/warehouse/issues/4035

I have removed italic rules form css and the `<em>` tag (I noticed that client css sheets) apply the italic rule to these elements as well.


warehouse/static/sass/resets/_normalize.scss
```diff
@@ -134,14 +134,6 @@ strong {
   font-weight: bold;
 }
 
-/**
- * Address styling not present in Safari and Chrome.
- */
-
-dfn {
-  font-style: italic;
```
di: We probably want to explicitly set this to `font-style: normal;` if the default is for this to become italic in some browsers.

warehouse/templates/pages/help.html
```diff
@@ -112,16 +112,16 @@ <h2 class="faq-title" id="Basics">Basics</h2>
 
         <h3 id="packages">{{ packages() }}</h3>
         <p>
-          We use a number of terms to describe software available on PyPI, like <em>project</em>, <em>release</em>, <em>file</em>, and <em>package</em>. Sometimes those terms are confusing because they're used to describe different things in other contexts. Here's how we use them on PyPI:
+          We use a number of terms to describe software available on PyPI, like project, release, file, and package. Sometimes those terms are confusing because they're used to describe different things in other contexts. Here's how we use them on PyPI:
```
di: Can we put these in quotes instead? I'd like to maintain some sort of distinction that these are specific terms. (Same for the lines below)

warehouse/templates/pages/help.html
```diff
@@ -379,10 +379,10 @@ <h3 id="contributing">{{ contributing() }}</h3>
           We have a huge amount of work to do to continue to maintain and improve PyPI (also known as <a href="https://warehouse.readthedocs.io/">the Warehouse project</a>).
         </p>
         <p>
-        <em>Financial</em>: We would deeply appreciate <a href="https://donate.pypi.org/">your donations to fund development and maintenance</a>.
+          Financial We would deeply appreciate <a href="https://donate.pypi.org/">your donations to fund development and maintenance</a>.
```
di: Let's make this bold instead. Also, you've lost the colon here.

jpurviance:
@di ready for review

warehouse/static/sass/resets/_normalize.scss
di: @jpurviance Can you resolve this?

di:
Thanks @jpurviance!

Issue #4781, opened 2018-09-28
Add support to rename project


filips123:
Please add support to transfer project to another name without loosing project history and previous releases.

ewdurbin:
Thank you for filing an issue @filips123! This is already being discussed and tracked in #1919.

PR #4974, opened 2018-10-29
Remove De-normalization Of Name and Version

dstufft:
When we were using natural primary keys, a lot of tables had name and version keys on them, since that was how the FKs worked. With the move to surrogate keys for all of our primary keys, these existing name and version fields are just leftover bits of denormalization that needs to be handled.

In https://github.com/pypa/warehouse/pull/4958#issuecomment-433734700 I laid out a few options for solving this. This PR represents choice (3).

There may be performance implications here, I've tried to keep any index that still seemed like it made sense, but this is sort of drastically altering the performance characteristics of some of these queries, so it's possible that there will be a performance regression. However, if there is we can, at that time, either solve it with additional indexes or perhaps choose to denormalize our data then.

### DIFF
--- a/tests/common/db/packaging.py
+++ b/tests/common/db/packaging.py
@@ -46,7 +46,6 @@ class ReleaseFactory(WarehouseFactory):
         model = Release
 
     id = factory.LazyFunction(uuid.uuid4)
-    name = factory.LazyAttribute(lambda o: o.project.name)
     project = factory.SubFactory(ProjectFactory)
     version = factory.Sequence(lambda n: str(n) + ".0")
     canonical_version = factory.LazyAttribute(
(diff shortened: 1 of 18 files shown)


ewdurbin:
I think the only concern here is changing the name of a project per #4781 would destroy the old name for a given release.

dstufft:
I'm not sure what you mean by destroy the original name for the given release?

dstufft:
Ah I see what you mean.

I don't think that matters, as long as we don't allow renames that change the normalized name all tools should treat them equivalent anyways. The current state is that the name in the dist file and the name on the ``Release`` metadata aren't always going to line up, because ``Release.name`` uses whatever ``Project.name`` was at the time of registration. However people can make changes in their dist file, and as long as the name normalizes to the same, we'll look up the correct ``Project`` instance, and use ``Project.name`` from that, even if that differers from what is in the dist file.

Issue #5031, opened 2018-11-10
Paginate History Page


dstufft:
The history pages in the management UI (``/manage/project/{foo}/history/``) does not paginate the results, and it should. The longer a project exists for, the longer this page is going to get.

roberson-io:
This is a really new issue, but it seems like it might be good for a first-timer like me.  If so, I would be interested in working on this.

nlhkabu:
thank you @roberson-io

PR #5038, opened 2018-11-10
Paginate project history

roberson-io:
This paginates project history (#5031 ).  I moved the pagination macro from the search results template to a separate template that the project history page also uses.

### DIFF
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -16,7 +16,8 @@ import uuid
 import pretend
 import pytest
 
-from pyramid.httpexceptions import HTTPBadRequest, HTTPSeeOther
+from paginate_sqlalchemy import SqlalchemyOrmPage as SQLAlchemyORMPage
+from pyramid.httpexceptions import HTTPBadRequest, HTTPNotFound, HTTPSeeOther
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 from webob.multidict import MultiDict
@@ -24,6 +25,7 @@ from webob.multidict import MultiDict
 from warehouse.manage import views
 from warehouse.accounts.interfaces import IUserService, IPasswordBreachedService
 from warehouse.packaging.models import JournalEntry, Project, File, Role, User
+from warehouse.utils.paginate import paginate_url_factory
 from warehouse.utils.project import remove_documentation
 
 from ...common.db.accounts import EmailFactory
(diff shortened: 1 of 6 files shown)


warehouse/templates/search/results.html
```diff
@@ -44,50 +46,9 @@ <h3 class="package-snippet__title">
...
-    <a href="{{ map.next_page.href }}" class="button button-group__button">Next</a>
-    {% endif %}
-  </div>
-{% endif %}
-{%- endmacro %}
-
 {% block search_form_extra_inputs -%}
   {%- for classifier in applied_filters %}
-    <input name="c" type="hidden" value="{{ classifier|e }}">
+  <input name="c" type="hidden" value="{{ classifier|e }}">
```
pradyunsg: An unnecessary change.

roberson-io: I thought so, too, but the linter failed on this.  I changed it back, and it looks like the Travis lint job fails, too.  Thanks for taking a look!

yeraydiazdiaz:
Thank for your PR @roberson-io, here are two screenshots of the functionality for reviewer convenience:

screen shot 2018-11-12 at 14 17 03
screen shot 2018-11-12 at 14 18 22

warehouse/manage/views.py
```diff
@@ -729,13 +731,25 @@ def delete_project_role(project, request):
     permission="manage:project",
 )
 def manage_project_history(project, request):
-    journals = (
+    try:
+        page_num = int(request.params.get("page", 1))
+    except ValueError:
+        raise HTTPBadRequest("'page' must be an integer.") from None
```
yeraydiazdiaz: I believe you don't need `from None` here

tests/unit/manage/test_views.py
```diff
@@ -1657,3 +1657,26 @@ def test_get(self, db_request):
...
+
+        url_maker = pretend.stub()
+        url_maker_factory = pretend.call_recorder(lambda request: url_maker)
+        monkeypatch.setattr(views, "paginate_url_factory", url_maker_factory)
+
+        project = ProjectFactory.create()
+        with pytest.raises(HTTPBadRequest):
+            views.manage_project_history(project, db_request)
+
+        assert page_cls.calls == []
```
yeraydiazdiaz: Would you mind adding two additional positive tests?

Ideally creating >25 JournalEntry objects and asserting the first 25 entries are returned when retrieving the first page and asserting that the remaining entry is returned when requesting the second page.

Or less entries if you decide to move the 25 to a setting

warehouse/manage/views.py
```diff
@@ -729,13 +731,25 @@ def delete_project_role(project, request):
...
         .options(joinedload("submitted_by"))
         .filter(JournalEntry.name == project.name)
         .order_by(JournalEntry.submitted_date.desc(), JournalEntry.id.desc())
-        .all()
     )
+
+    journals = SQLAlchemyORMPage(
+        journals_query,
+        page=page_num,
+        items_per_page=25,
```
yeraydiazdiaz: It may be convenient for testing to have this as a setting, possibly opening the door for a nice dropdown in the pagination controls in the future

warehouse/manage/views.py
roberson-io: I removed `from None` here.

tests/unit/manage/test_views.py
roberson-io: I added the two tests you suggested plus a test for the case where someone requests a page out of range.  For example, requesting page 3 when there are only 2 pages.

warehouse/manage/views.py
roberson-io: I would be happy to make `items_per_page` a setting, but I think I need some guidance on how I should go about doing that.  Thanks for reviewing and for your feedback!

yeraydiazdiaz: On second thought it might work better as a request paramenter read by the view, but only makes sense if we were to implement the feature in the frontend so for now it's fine as it is. 👍

ewdurbin:
thank you @roberson-io!

PR #5039, opened 2018-11-11
Optimize a number of queries

dstufft:


warehouse/packaging/models.py
```diff
@@ -498,6 +501,7 @@ def __table_args__(cls):  # noqa
             Index("journals_name_idx", "name"),
             Index("journals_version_idx", "version"),
             Index("journals_submitted_by_idx", "submitted_by"),
+            Index("journakls_submitted_date_id_idx", cls.submitted_date, cls.id),
```
pradyunsg: There's a typo here.

PR #5057, opened 2018-11-12
[WIP] Switch to using PyPy3-6.0.0

dstufft:
A PR to discuss if this seems like something we should do.

The basic motivation behind this PR is I was trying to look at what was making our XMLRPC APIs so slow, and it appears that fundamentally the issue is just that the XMLRPC APIs are poorly designed (duh) and rendering the huge XML structure that some of the APIs return is taking up the bulk of the time spent.

I'm pretty sure that PyPy is likely to speed up the performance of the entire code base, but I haven't actually went through view by view to get a sense of whether it did or not.

So the biggest downsides to switching to PyPy:

* Big change to our runtime, which comes with it risk of something exploding in Production that we didn't find in testing.
* PyPy3 is still on Python 3.5 (although it has some features, like f-strings). This wasn't a big deal in our code base (see the diff) but 3.5 is getting old enough it feels like people might stop dropping support soon?
  * I believe PyPy is actively working on 3.6 support, and ``#pypy`` suggests it is nearly done.
* Some C extensions just don't work (although cpyext is getting better compatibility all the time), and when they do work they are often times slow.
  * Using Pure Python instead of C is often times just as fast, and when it's not there is typically a CFFI version of the library that can be used instead.
* CPython has the bulk of the real world use, so PyPy might have more edge cases?
* Performance is going to be less consistent due to JIT warmup.
  * I don't honestly think our performance is that consistent right now as it is.
* Tests are likely slower?
  * PyPy's JIT works best when you're running the same code over and over.. tests do the exact opposite.
* PyPI running on PyPy, double the confusion, double the fun. It's PYPYI!

The big upside is a pretty healthy performance boost (10x on XMLRPC ``list_packages()`` API when running locally, without any effort besides porting to PyPy itself).

This isn't actually fully working as of now, I see an error being generated from celery when running locally, and there are ~74 test failures (A lot of which look like fallout from going from 3.6 -> 3.5). Throwing this up to give folks a chance to comment, and to think about what a possible migration to PyPy might look like.


dstufft:
@pypa/warehouse-committers

alex:
PyPy is a great computer program.

pradyunsg:
> PyPI running on PyPy, double the confusion, double the fun.

I'm probably more excited about seeing this happen, than is reasonable.

dstufft:
This is currently blocked on [pypy/pypy#2912](https://bitbucket.org/pypy/pypy/issues/2912/typeerror-decoding-unicode-is-not) either that being fixed in PyPy itself or in one of {PyCurl, Kombu}, or worked around in Warehouse.

pradyunsg:
ISTM installing PyPy 3 6.0.0 via pyenv when the Travis Job starts should work (but I haven't tested that).

dstufft:
Alright, I've updated this so that we're using a nightly version of PyPy which has the above bug fixed, and I've verified that Celery is working with it.

I've already gotten all of the tests working fine, so that means the main thing left is figuring out how to work in CI. We can just try to install PyPy3 as part of the Travis job, but that is yet another place we have to update what version of Python we're using.

Since our real deployment now uses Docker Containers, and I believe (@ewdurbin can confirm) that it uses a docker container as built by our ``Dockerfile``, that it may make the most sense to just build our docker container in CI and run our tests inside of that. That will give us far more consistent behavior between local, CI, and production.

blueyed:
Just a side note: it might make sense to use CircleCI instead of Travis, given that it has better support for testing via Docker containers right away.

dstufft:
Oh, another solution to speeding this up would be to move some of the stuff we do into a "base image" of sorts that we manage as an independent repository. This wouldn't be able to be any of the Python level dependencies, but we could potentially have a "warehouse base" image that looked something like:

```Dockerfile
FROM debian:jessie-slim as build

RUN set -x \
    && apt-get update \
        bzip2 build-essential libexpat1-dev libffi-dev libxml2-dev libxslt-dev libpq-dev libcurl4-openssl-dev libssl-dev 

ADD http://buildbot.pypy.org/nightly/py3.5/pypy-c-jit-95312-3bb86a88e1d3-linux64.tar.bz2 /tmp/pypy.tar.bz2
RUN set -x \
    && echo "9d6e7c11645833794f5f207051331f1835f1ba10aec3c9aaa0ae0612e8829b4e  /tmp/pypy.tar.bz2" | sha256sum -c - \
    && mkdir -p /opt/pypy \
    && tar xjf /tmp/pypy.tar.bz2 --strip-components=1 -C /opt/pypy
ENV PATH="/opt/pypy/bin:${PATH}"

```

This would increase our final image size a little bit (since we'd have the dev headers installed in production, instead of just the libs themselves) but I don't think by a meaningful amount. It would make things like bumping the version of Python or adding a new C dependency harder to do, since we'd have to manage that inside of another repository, but maybe it would be worth it?

dstufft:
This is currently blocked on https://github.com/ambv/black/issues/613, where our use of f-strings (which PyPy3 supports even though it is a 3.6 feature and PyPy3 is 3.5) is causing black to flag the file as Python 3.6+, which causes it to use features which are *not* Python 3.5 compatible, which then causes PyPy to fail to parse that file because it doesn't support *all* of the Syntax from Python 3.6.

There are only a few ways around this:

* Remove all of our use of f-strings so we are "Pure" 3.5 compatible.
  * I really don't want to do this, f-strings are a really nice improvement and PyPy supports them fine. Also a PyPy3 that supports 3.6 is hopefully in not the super far future, where it would suck to have to go through and undo all of the undoing of f-strings.
* Patch black to add a ``--no-py36`` flag that we can use to force < Python 3.6 (I assume it would still handle the fact that there are f-strings, just wouldn't format around them).
  * I'm not sure if black will accept this configuration option or not.
* Wait until a PyPy3 that supports 3.6 comes out.
  * We can do this today with nightlies, and I think an alpha will be out soon?
* Disable the numeric literals rule in black (one of two rules PyPy3 fails on, and the only one the current code base actually triggers) and hope that we don't trigger the other rule before a 3.6 supporting PyPy3 comes out.

Personally the only options that feel reasonable to me are patching black and waiting until a 3.6 PyPy3 comes out. I'm going to play with these two options and see what I can make happen.

.travis.yml
```diff
@@ -1,61 +1,45 @@
...
 addons:
-  hosts:
-    - warehouse.local
-  postgresql: 10
   apt:
     packages:
-    - postgresql-10
-    - postgresql-client-10
-    - nasm
+      - docker-ce
```
webknjaz: You don't need this. Docker is installed in Travis CI by default.

.travis.yml
```diff
@@ -1,61 +1,45 @@
...
+  - make build
+  - make tests
 
 matrix:
   include:
     - env:
       - SUITE=Tests
-    - script: make lint BINDIR="$(dirname $(which python))"
-      env:
+    - env:
```
webknjaz: It looks like you use this only to "label" jobs in the UI of Travis CI. There's a new feature, which allows you to do this natively instead:
```suggestion
    - name: Linter checks
```

.travis.yml
```diff
@@ -1,61 +1,45 @@
...
         - SUITE=Lint
```
webknjaz: And then it won't be needed:
```suggestion
```

.travis.yml
```diff
@@ -1,61 +1,45 @@
...
+  - docker-compose version
 
 script:
-  - node -v
-  - npm -v
-  - bin/tests --postgresql-host 127.0.0.1
+  - make build
+  - make tests
 
 matrix:
```
webknjaz: `jobs` is a more intuitive alias name for `matrix`. And you can also make Travis CI report failures faster (once any of jobs fails):
```suggestion
 matrix:
 jobs:
   fast_finish: true
```

.travis.yml
```diff
@@ -1,61 +1,45 @@
...
-    - postgresql-10
-    - postgresql-client-10
-    - nasm
+      - docker-ce
 
 services:
-  - postgresql
+  - docker
 
 cache:
```
webknjaz: Travis CI is more aware of what they suggest to cache for pip:
```suggestion
 cache:
   pip: true
```

.travis.yml
dstufft: This updates docker to the latest version.

webknjaz:
PyPy3.5 v6.0 is now supported in Xenial boxes: https://github.com/cherrypy/cheroot/commit/cfee0e13e660f4f67a37c1928a737dd9ccc05fec

.travis.yml
```diff
@@ -1,62 +1,41 @@
 language: python
 python: 3.6.6
```
webknjaz:
```suggestion
 python: pypy3.5-6.0

 dist: xenial
```
can enable you run PyPy v6.0 in Travis

dstufft:
Closing this, it's bit rotted at this point and I don't think we're actually going to end up doing it. We're on CPython 3.7 now and will likely switch to CPython 3.8 when it comes out as well.

At this point we're likely to get more traction on getting the new APIs out which don't have the fundamental problems that the XMLRPC APIs do, and then maybe we will make a translation layer for the, and *that* can run in PyPy.

PR #5072, opened 2018-11-15
Document /stats/ JSON API

cooperlees:
- Add new .rst documenting how the `/stats/` JSON API can be queried

Finishes https://github.com/pypa/warehouse/issues/4288


ewdurbin:
ready to 🚢 when CI passes :)

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,50 @@
+STATS API
```
di:
```suggestion
Stats API
```

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,50 @@
...
+=========
+
+PyPI offers a JSON stats endpoint via a `Content-Type: application/json` GET
```
di:
```suggestion
PyPI offers a JSON stats endpoint via a ``Content-Type: application/json`` GET
```

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,50 @@
...
+request. Without the Content-Type header set, you will be returned a HTML page.
```
di:
```suggestion
request. Without the ``Content-Type`` header set, you will be returned a HTML page.
```

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,50 @@
...
+        "Fiona": {
+          "size": 6209535709
+        },
+        "FlexGet": {
+          "size": 4387002448
+        },
+        ...
+      },
+      "total_packages_size": 23965450269
+    }
```
di: This whole block needs indented.

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,50 @@
...
+    .. code:: http
+
+        GET /stats HTTP/1.1
+        Host: pypi.org
+        Accept: application/json
+        Content-Type: application/json
+
+    **Example response**:
+
+    .. code:: http
```
di:
```suggestion
    .. code:: json
```

docs/api-reference/stats.rst
cooperlees: JSON API uses all caps so I did. But will change.

di: Right, because JSON is an acronym -- "Stats" is not (see also the "Legacy API")

cooperlees:
Anything else required from me here guys?

docs/api-reference/stats.rst
```diff
@@ -0,0 +1,49 @@
+Stats API
+=========
+
+PyPI offers a JSON stats endpoint via a ```Content-Type: application/json`` GET
```
di:
```suggestion
PyPI offers a JSON stats endpoint via a ``Content-Type: application/json`` GET
```

One too many backticks here now.

PR #5084, opened 2018-11-18
Manifest prevents running tests

### LINKED ISSUE #1536: Tests depend on `manifest.json` existing

yeraydiazdiaz:
Fixes #1536 in a different way than #3978 using only monkeypatching in the test code. Closes #3978.

I had to resort to using `unittest.mock` since `app_config` is a session fixture and `monkeypatch` is a function one. It's not pretty but it works.

I'm testing Travis builds without `gulp dist` which should also speed up the test build.


.travis.yml
```diff
@@ -38,7 +38,7 @@ install:
 script:
   - node -v
   - npm -v
-  - ./node_modules/.bin/gulp dist
+  # - ./node_modules/.bin/gulp dist
```
di:
```suggestion
```

Issue #3440, opened 2018-03-27
Main page should auto-focus the "search" box


pfmoore:
On pypi.python.org, the search box is automatically focused, allowing you to immediately type in a search when you load the page. On pypi.org, the search box isn't focused, so you need an extra click to do a search.

ewdurbin:
Discussed previously in #854

pfmoore:
Ah thanks. I did a search but didn't find that issue, sorry.

ewdurbin:
no need to apologize, it may be worth discussing again. i'm gonna leave this open and tag it for UX.

nlhkabu:
The challenge with this is twofold:

1. it's annoying on mobile (as per #854)
2. It disturbs the tab index of the page. This may make it painful for users who use a keyboard to access the top menu.  

Note that we *are* doing this on the login page - so we probably need to check this page for both these factors.

If we can find a workaround for both mobile and keyboard-only users, then I will be very happy to implement this.

pohmelie:
Vote for this both hands.
@nlhkabu 
* Now it is painful for users who use a keyboard to access search box. 11 (eleven!) tabs required!
* Google have top menu, but focus still on search. 
* Do you have research about how  much users go from front page with search box and how much users go via top menu? (I think 99% go to search box)
* Is it too hard to detect mobile client? (I don't think so)

nlhkabu:
> Google have top menu, but focus still on search.

To get back to the top menu on Google, you only have to tab through a handful of options.  On warehouse, you would have to tab through a **lot** of options.  To address this we could look at changing the tab index of the page to jump *from* the search, to the top menu.

> Do you have research about how much users go from front page with search box and how much users go via top menu? (I think 99% go to search box)

We currently do not have this data, but it's safe to assume that *most* people use the search. However, the login link is in the top menu, so a significant number of package maintainers may also want easy access to this link.

As I mentioned earlier, I am not opposed to this (quite the opposite).  I just think we need to address these two issues.  As @pohmelie mentioned, it's not hard to detect a mobile client (or a small resolution), so I don't think either issue is insurmountable.  Any help with implementing this feature would be appreciated!

pohmelie:
@nlhkabu 
> On warehouse, you would have to tab through a lot of options

Use `shift-tab` to go from search box to top menu. One hit and you are on "Register" menu entry. One more hit and you are on "Log in" entry. Do not see any pain here.

jamadden:
[...]

PR #5102, opened 2018-11-21
Autofocus search field

techtonik:


di:
Lots of discussion about this in #854 and #3440.

techtonik:
@nlhkabu can you confirm that `autofocus` on mobile causes keyboard popup? I would appreciate a prooflink, because it sounds like a bug rather than expected behavior for HTML5 element. Also @di previous patch didn't use HTML5, but JavaScipt - https://github.com/nlhkabu/warehouse/commit/0d45556a2ed10c27d4ed677cfaf14fb9f3bf42c5 so I think it should be safe.

@dstufft already merged HTML5 fix in https://github.com/pypa/warehouse/pull/985 but I don't see why it was reverted.

techtonik:
The only argument that is left is accessibility of the top menu. While accessibility touches the focus issue, I think that Accessibility should be addressed site-wide with PSF funding, which could result in proper coverage and development and guidelines for contributors.

di:
> @dstufft already merged HTML5 fix in #985 but I don't see why it was reverted.

@nlhkabu removed it in https://github.com/pypa/warehouse/pull/2822, though I'm not sure why.

di:
@nlhkabu Can you merge/close this issue as you see fit?

nlhkabu:
Hi @techtonik thanks for this PR

Thanks to the recent OTF funding for accessibility, I've had time to look into the issue of *whether* or not this is a good idea. I've opened a ticket with my findings here: #6088

TLDR: Based on the available accessibility recommendations, and the data that shows how users are using the page, we should not focus the search box.

Thanks again for your PR.

techtonik:
@nlhkabu I can not comment on your research in #6088, because comments are locked.

Am I reading your results right?

 - 52.9% use search after landing
 - 8.5% do something else
 - 38.6% people bounce and leave the site

If that is so, that means that 84.5% of site users are doing search, which also means its the primary site function for the 84.5% of users. Do they do extra clicking? Or sole keyboard navigation?

PR #5104, opened 2018-11-21
remove ipaddress from requirements/main.txt

ewdurbin:
elasticsearch-dsl has a conditional dependency on this for Python 3.3 and below.


ewdurbin:
#5100 took care of this :)

PR #5107, opened 2018-11-21
Enforce 100% coverage using coverage.py instead of codecov

dstufft:


dstufft:
This is failing because coverage on `master` has dropped below 100%, but once `master` is fixed, we should be good to go to merge this and drop the requirement on codecov.

PR #5132, opened 2018-11-28
Add isort as a companion to black

dstufft:
Much like black, this will sort our imports automatically and provide a check that will ensure that future PRs correctly sort imports. It has been added to ``make reformat`` and of course, any editors that support actions can be configured to automatically run these as well.

The biggest differences are:

* We now require alphabetical imports.
  * Previously we kind of half way alphabetized them, but without any real enforcement.
* We now require seperation of stdlib, thridparty, and our own imports.
  * Previously we mostly did this, but a few imports got stuck in the wrong place.
* ~~``import`` and ``from <foo> import <bar>`` are now grouped together, with ``import`` being first.~~
  * ~~Previously we did basically this, but we had the ``import`` and ``from import`` styles separated as separate sections, but otherwise did the same.~~
* ~~Separate "complex" imports moved to one per import statement (``from foo import bar as frob, thing as other``).~~
  * ~~We used to have multiple of these per statement, but isort moves them to be one per import statement.~~


hynek:
Since this is probably my only chance until I'll have to forever hold my peace: **please** add `lines_between_types=1` 🐶  I'll get you a beer at PyCon…a fancy one! :)

dstufft:
Thanks to @hynek I fixed a few of the issues that were different than our previous ad hoc standard. Now the isort configuration is basically what we've always done, just automatically and enforced.

PR #5182, opened 2018-12-11
Improve error handling about duplicated email within current account

### LINKED ISSUE #5124: Error message is incorrect when attempting to add existing email
If I try to add an email address I've already added to my account, the error message says:

> This email address is already being used by another account. Use a different email.

Which isn't technically true, since the email address is already in use by the _current_ account. We should differentiate it and give a better message when this happens.

---

dlax:
Previously, when trying to add an email to one's account that was
already registered we'd get the generic error message about email being
already used by *another* account from NewEmailMixin (in
warehouse/accounts/forms.py).

We improve this feedback by adding a check in AddEmailForm to validate
that specified email does not match with emails already used by this
account. As we need the current user to perform this check, we now store
the "request" as an attribute in AddEmailForm.

Fixes #5124.

### DIFF
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -114,7 +114,7 @@ class TestAddEmailForm:
         assert not form.validate()
         assert (
             form.email.errors.pop()
-            == "You can't add an email address from this domain. "
+            == "You can't use an email address from this domain. "
             "Use a different email."
         )
 
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -104,6 +104,20 @@ class TestAddEmailForm:
             "Use a different email."
         )
 
+    def test_blacklisted_email_error(self):
+        form = forms.AddEmailForm(
+            data={"email": "foo@bearsarefuzzy.com"},
+            user_service=pretend.stub(find_userid_by_email=lambda _: None),
+            user_id=pretend.stub(),
+        )
+
+        assert not form.validate()
+        assert (
+            form.email.errors.pop()
+            == "You can't add an email address from this domain. "
+            "Use a different email."
+        )
+
 
 class TestChangePasswordForm:
     def test_creation(self):
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -71,9 +71,38 @@ class TestCreateRoleForm:
 class TestAddEmailForm:
     def test_creation(self):
         user_service = pretend.stub()
-        form = forms.AddEmailForm(user_service=user_service)
+        form = forms.AddEmailForm(user_service=user_service, user_id=pretend.stub())
 
         assert form.user_service is user_service
+
+    def test_email_exists_error(self):
+        user_id = pretend.stub()
+        form = forms.AddEmailForm(
+            data={"email": "foo@bar.com"},
+            user_id=user_id,
+            user_service=pretend.stub(find_userid_by_email=lambda _: user_id),
+        )
+
+        assert not form.validate()
+        assert (
+            form.email.errors.pop()
+            == "This email address is already being used by this account. "
+            "Use a different email."
+        )
+
+    def test_email_exists_other_account_error(self):
+        form = forms.AddEmailForm(
+            data={"email": "foo@bar.com"},
+            user_id=pretend.stub(),
+            user_service=pretend.stub(find_userid_by_email=lambda _: pretend.stub()),
+        )
+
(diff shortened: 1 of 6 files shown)


warehouse/manage/forms.py
```diff
@@ -61,10 +61,20 @@ class AddEmailForm(NewEmailMixin, forms.Form):
 
     __params__ = ["email"]
 
-    def __init__(self, *args, user_service, **kwargs):
+    def __init__(self, *args, request, user_service, **kwargs):
```
di: We don't need to pass the entire `request` in to the form. Instead we can just pass in `request.user.id` as `user_id`, and have it default to `None` if it's not provided.

warehouse/manage/forms.py
```diff
@@ -61,10 +61,20 @@ class AddEmailForm(NewEmailMixin, forms.Form):
...
         self.user_service = user_service
 
+    def validate_email(self, field):
+        userid = self.user_service.find_userid_by_email(field.data)
+        if userid is not None and userid == self.request.user.id:
+            raise wtforms.validators.ValidationError(
+                "This email address is already being used by this account. "
+                "Use a different email."
+            )
+        return super().validate_email(field)
```
di: I think this check should be part of the `validate_email` function of the `NewEmailMixin`, rather than being overridden here.

dlax:
@di, thanks for your review. I've taken your points into account.
While I was there, I added another commit to change the message when adding a email with a blacklisted domain to an existing account.

brainwane:
@alanbato do you potentially have time to help out by taking a look at this PR? Thanks!

tests/unit/manage/test_forms.py
```diff
@@ -69,10 +69,53 @@ def test_validate_role_name_fails(self, value, expected):
...
+        form = forms.AddEmailForm(
+            data={"email": "foo@bar.com"},
+            user_id=user_id,
+            user_service=pretend.stub(find_userid_by_email=lambda _: user_id),
+        )
+
+        assert not form.validate()
+        assert (
+            form.email.errors.pop()
+            == "This email address is already being used by this account. "
```
alanbato: Maybe we could assign the error message in a variable i.e. `expected_error` and then do the assertion like this
```python
assert from.email.errors.pop() == expected_error
```
I think that'd improve readability a bit :)

tests/unit/manage/test_forms.py
```diff
@@ -69,10 +69,53 @@ def test_validate_role_name_fails(self, value, expected):
...
+        form = forms.AddEmailForm(
+            data={"email": "foo@bar.com"},
+            user_id=pretend.stub(),
+            user_service=pretend.stub(find_userid_by_email=lambda _: pretend.stub()),
+        )
+
+        assert not form.validate()
+        assert (
+            form.email.errors.pop()
+            == "This email address is already being used by another account. "
```
alanbato: See previous comment

tests/unit/manage/test_forms.py
```diff
@@ -69,10 +69,53 @@ def test_validate_role_name_fails(self, value, expected):
...
+        form = forms.AddEmailForm(
+            data={"email": "foo@bearsarefuzzy.com"},
+            user_service=pretend.stub(find_userid_by_email=lambda _: None),
+            user_id=pretend.stub(),
+        )
+
+        assert not form.validate()
+        assert (
+            form.email.errors.pop()
+            == "You can't add an email address from this domain. "
```
alanbato: See previous previous comment

tests/unit/manage/test_forms.py
dlax: I essentially tried to write something consistent with other tests, especially with those in `./tests/unit/accounts/test_forms.py`; see, e.g.: https://github.com/pypa/warehouse/blob/3db5abc1c80a98658fa97b6953c545bc6aacf740/tests/unit/accounts/test_forms.py#L311

I can change that if needed, though.

brainwane:
@alanbato could you respond to @dlax's question/comment? Thanks!

tests/unit/manage/test_forms.py
alanbato: Oh, thanks for noticing that. 
And you're right, we should prioritize consistency around the codebase!

LGTM 👍

brainwane:
@alanbato if you approve this PR can you leave an approval/:+1:  review? Thanks!

@di please review. Thanks!

brainwane:
Thanks, @dlax! If you'd like to work on another issue next, let us know by replying, and I'll be happy to give you a suggestion!

PR #5201, opened 2018-12-15
Remove invalid chars from XML responses

di:
Fixes https://github.com/pypa/pip/issues/5760.

We're using [`pyramid_rpc.xmlrpc`](https://github.com/Pylons/pyramid_rpc/blob/master/pyramid_rpc/xmlrpc.py) which uses [`xmlrpc.client.dumps`](https://docs.python.org/3/library/xmlrpc.client.html#xmlrpc.client.dumps), however according to https://docs.python.org/3/library/xmlrpc.client:

> When passing strings, characters special to XML such as <, >, and & will be automatically escaped. However, it’s the caller’s responsibility to ensure that the string is free of characters that aren’t allowed in XML

All of our user-supplied fields accept strings that could be invalid XML, so we should be:
* replacing invalid characters with their equivalent XML character reference where we can;
* removing any characters that don't have an XML character reference.

I believe I got all the places where user-supplied content could make it into XML (aside from things like package name and version, where these characters are already invalid).

This definitely has the potential to slow down the XML-RPC endpoints, so we'll probably want to keep an eye on things after deploying this.


ewdurbin:
XMLRPC calls over last 7 days:

screen shot 2018-12-17 at 11 30 27 am

So indeed we'll need to keep an eye on this, but I'm willing to ship/revert after measuring impact.

ewdurbin:
no discernable impact on performance of xmlrpc search method:

### average call time:

screen shot 2018-12-17 at 1 03 02 pm

### max call time:

screen shot 2018-12-17 at 1 03 10 pm

Issue #4011, opened 2018-05-16
Create "Moderator" level User and corresponding permissions


di:
Currently on PyPI, we have two types of users:
* regular users
* administrators (for which `User.is_superuser` is true)
* (There is also a field `User.is_staff` but I believe it's unused).

It'd be great to add an intermediate level of user, "Moderator", which has the ability to do the following:
* view all the existing `/admin` views (but read-only)
* set upload limits
* add classifiers
* etc

The "Moderator" user should not be able to:
* change emails (including verified or active status)
* delete projects
* nuke users

(cc @99)

di:
Agreed. I think the steps to address this, roughly broken up by PRs, would be:
* Add a new field/migration to the `User` model, as well as corresponding UI in `/admin` for administrators to toggle this flag for users;
* Add ACLs for various `/admin` POST endpoints, as well as conditionals in the HTML templates to hide actions that are not enabled for moderators, and to make fields not appear editable;
* Add back these actions as "request/recommendation" actions that get raised to administrators. This could piggyback on the system we use for #3896.

dstufft:
Yea, that seems like a great path forward.

crwilcox:
I am going to start working on this. I have the first part complete (add new field and migration as well as adding is_moderator as settable in admin UIs) https://github.com/pypa/warehouse/pull/5249/files

crwilcox:
@di @dstufft I am moving on to `Add ACLs for various /admin POST endpoints, as well as conditionals in the HTML templates to hide actions that are not enabled for moderators, and to make fields not appear editable;`

Do you have a starting list of activities you would like to allow moderators to do? For instance, should moderators be able to mark other users as having a verified email?

di:
@crwilcox See the issue body:

> view all the existing /admin views (but read-only)
> set upload limits
> add classifiers

di:
#5249 is merged, we can now set users as moderators.

@jamadden @yeraydiazdiaz Would you like to be moderators?

yeraydiazdiaz:
That’s a yes for me, thanks ☺️

crwilcox:
@theacodes volunteered if you want additional moderators.

pradyunsg:
I volunteer too, if you need more eyes and hands on deck. :)

jamadden:
@ewdurbin I got your Slack invite, thank you. Would it be possible to switch it to the alternate address for my PyPI account?

ewdurbin:
[...]

PR #5249, opened 2019-01-04
Allow Users to be moderators and for field to be set via Admin dashboard

crwilcox:
As part of address #4011 one task is:
Add a new field/migration to the User model, as well as corresponding UI in /admin for administrators to toggle this flag for users;


crwilcox:
I opened this early, my mistake, I am currently working on tests.

warehouse/accounts/__init__.py
```diff
@@ -98,6 +98,8 @@ def _authenticate(userid, request):
 
     if user.is_superuser:
         principals.append("group:admins")
+    if user.is_moderator or user.is_superuser:
```
crwilcox: With the thought that filtering things to one principal, considering superuser to inherit moderator privileges would mean things could be flagged as moderators.

If someone else wanted to weigh in, it may make sense to not do this in this way though.

crwilcox:
Tests have been run. Also verified the admin portal and setting/unsetting admin.

warehouse/accounts/__init__.py
di: I think we just need to add moderators to the moderators group. I think we can just set up the ACLs based on whether the user is in the admin group or not.

warehouse/migrations/versions/3db69c05dd11_add_field_on_user_model_for_is_moderator.py
```diff
@@ -0,0 +1,49 @@
...
+down_revision = "2d6390eebe90"
+
+# Note: It is VERY important to ensure that a migration does not lock for a
+#       long period of time and to ensure that each individual migration does
+#       not break compatibility with the *previous* version of the code base.
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
```
di: You can remove this comment and the comments below in this file.

crwilcox:
At this point the current state is that you can be a moderator and attempts to post to most all endpoints will fail. Yet to be done is to make the buttons disabled to make it clear what moderators can/can't do.

di:
@crwilcox Mind saving that last commit for a separate PR? I was just about to leave some comments on the current status of this.

warehouse/accounts/interfaces.py
```diff
@@ -66,7 +66,14 @@ def check_password(user_id, password, *, tags=None):
...
         """
 
-    def create_user(username, name, password, is_active=False, is_superuser=False):
+    def create_user(
+        username,
+        name,
+        password,
+        is_active=False,
+        is_superuser=False,
+        is_moderator=False,
```
di: I think we don't need this argument here -- we'll only be upgrading already existing users to moderators.

Come to think of it, I doubt we need the `is_superuser` argument either. Might be best to remove that as well.

warehouse/accounts/__init__.py
di: Scratch that, after seeing the implementation this makes sense as-is.

docs/application.rst
```diff
@@ -56,6 +56,9 @@ Warehouse serves three main classes of users:
    Dustin Ingram, and Donald Stufft, who add classifiers, ban
    spam/malware projects, help users with account recovery, and so
    on. There are under ten such admins.
+4. *PyPI application moderators*. These users have a subset of the
+   permissions of *PyPI application administrators* to assist in some
+   routine administration tasks.
```
di: Not sure which commit you want to add this in, but the paragraph above this list needs updated as well ("three main classes of users"). Probably makes sense to move the things moderators can do, like "who add classifiers" to here instead, and be more explicit about what "routine administration tasks" entails.

crwilcox: I reordered these, fixed the count before this, and expanded on what moderators do.

warehouse/accounts/interfaces.py
crwilcox: I removed this from the interface.

warehouse/accounts/services.py
```diff
@@ -172,15 +172,20 @@ def check_password(self, userid, password, *, tags=None):
...
+        is_active=False,
     ):
 
         user = User(
             username=username,
             name=name,
             password=self.hasher.hash(password),
             is_active=is_active,
-            is_superuser=is_superuser,
+            is_superuser=False,
```
crwilcox: After removing the defaults from the interface, I noticed that the default isn't respected with db.Model. This is possibly a bug. This is what you will see if you don't have these lines

```
self = <sqlalchemy.dialects.postgresql.psycopg2.PGDialect_psycopg2 object at 0x7f5f8020e668>, cursor = <cursor object at 0x7f5f7f5ec618; closed: -1>
statement = 'INSERT INTO users (username, name, password, is_active, disabled_for) VALUES (%(username)s, %(name)s, %(password)s, %(is_active)s, %(disabled_for)s) RETURNING users.id'
parameters = {'disabled_for': None, 'is_active': False, 'name': 'test_name', 'password': '$argon2i$v=19$m=1024,t=6,p=6$dG5NSWltzfn//3/vnbP23g$+7Alw6BCiPsC38K1JMkDmw', ...}
context = <sqlalchemy.dialects.postgresql.psycopg2.PGExecutionContext_psycopg2 object at 0x7f5f7ea9bd30>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) null value in column "is_superuser" violates not-null constraint
E       DETAIL:  Failing row contains ($argon2i$v=19$m=1024,t=6,p=6$dG5NSWltzfn//3/vnbP23g$+7Alw6BCiPsC..., 2019-01-09 17:31:33.509108, null, test_user, test_name, f, 2019-01-09 17:31:33.509108, d, 2019-01-09 17:31:33.509108, b4bc1a89-aaf5-47a9-8b25-84a4f6c0d734, null, f).
E        [SQL: 'INSERT INTO users (username, name, password, is_active, disabled_for) VALUES (%(username)s, %(name)s, %(password)s, %(is_active)s, %(disabled_for)s) RETURNING users.id'] [parameters: {'username': 'test_user', 'name': 'test_name', 'password': '$argon2i$v=19$m=1024,t=6,p=6$dG5NSWltzfn//3/vnbP23g$+7Alw6BCiPsC38K1JMkDmw', 'is_active': False, 'disabled_for': None}] (Background on this error at: http://sqlalche.me/e/gkpj)

../lib/python3.6/site-packages/sqlalchemy/engine/default.py:509: IntegrityError
```

di: The issue is that `compare_server_default` is off by default, so an auto-migration for this change wouldn't actually add the defaults. I fixed this in #5275.

crwilcox: Thanks @di I will rebase on my end and correct :)

di:
@crwilcox Looks like the "nuke user" button on the User page needs to be conditionally disabled (or really, the whole "Actions" box can go away):

screen shot 2019-01-15 at 6 04 28 pm

crwilcox:
@di , latest push fixes that button to disable.

PR #5202, opened 2018-12-16
Add "trending" sort order to search page

yeraydiazdiaz:
Fixes #3447

Add zscore to Project document mapping
Add zscore to `_project_docs` query to populate on reindex
Add "Trending" sort order to search ordering dropdown

screenshot 2018-12-16 at 13 19 32

The results in the screenshot are not particularly relevant since I populated the `zscore` with random values on 100 project. It'd be nice to test it with proper values if possible to guage how useful the sort order is.

PR #5252, opened 2019-01-05
Add zscore to Project document mapping

yeraydiazdiaz:
Split from #5202 as suggested


ewdurbin:
manual complete reindex complete.

PR #5488, opened 2019-02-26
Add a note that you need to use Twine to take advantage of metadata fields.

### LINKED ISSUE #3797: Description of how to set a project description doesn't mention that you have to use "twine"
I looked at the page for ConceptNet on the new PyPI, and saw that it didn't have much of a description.

I went to the "manage" page and saw this message: https://imgur.com/a/pSlvUl6

I read the documentation it linked, and read about `long_description` and `long_description_content_type`, so I put those in setup.py, incremented the version number, and ran `python setup.py sdist upload`, like I always have.

Whoops. It turns out that you can't use `long_description_content_type` unless you're using Twine and have a bunch of other dependencies up to date, so the formatting is still garbage: https://pypi.org/project/ConceptNet/

I eventually learned that Twine is required by Googling a lot and finding it several paragraphs down on someone's blog post: https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi . Yes, the packaging documentation now says to use Twine, but very far from the place that PyPI linked me to -- and I had no reason to know that the command had changed from `python setup.py sdist upload` or that there was imperfect compatibility with the old command. I don't just re-read the entire packaging guide every week for fun.

The message about how to set your project description should mention Twine, or ideally, PyPI should be able to fix releases created with `python setup.py sdist upload`.

theacodes:
Closes #3797

### DIFF
--- a/warehouse/templates/manage/settings.html
+++ b/warehouse/templates/manage/settings.html
@@ -25,11 +25,13 @@
     <p>
       To set the '{{ project.name }}' description, author, links,
       classifiers, and other details for your next release, use
-      the <a href="https://packaging.python.org/tutorials/distributing-packages/#setup-args"><tt>setup()</tt>
+      the <a href="https://packaging.python.org/tutorials/distributing-packages/#setup-args" rel="noopener" target="_blank"><tt>setup()</tt>
       arguments in your <tt>setup.py</tt> file</a>. Updating these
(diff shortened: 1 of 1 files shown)


theacodes:
(I also added `target=blank` and `rel=noopener` to those links, as there's no reason to link externally without those attributes in this case)

Issue #5535, opened 2019-03-07
Predictable wheel URL redirect broken for tbvaccine package


flying-sheep:
The only documentation about the predictable URLs I found is [here](https://warehouse.pypa.io/api-reference/integration-guide/#if-you-so-choose), combined with the PEP 0491 section about the [wheel file name convention](https://www.python.org/dev/peps/pep-0491/#file-name-convention). Predictable wheel URLs can therefore be constructed like this:

```py
info = dict(
    distribution='tbvaccine',
    version='0.3.1',
    python='py2.py3',
    abi='none',
    platform='any',
)

info_wheel_file_name = {
    tag: re.sub("[^\w\d.]+", "_", part, re.UNICODE)
    for tag, part in info.items()
}

wheel_file_name = '{distribution}-{version}-{python}-{abi}-{platform}.whl'.format_map(info_wheel_file_name)

url = f'https://files.pythonhosted.org/packages/{info["python"]}/{info["distribution"][0]}/{info["distribution"]}/{wheel_file_name}'
```

This results in the URL: https://files.pythonhosted.org/packages/py36/t/tbvaccine/tbvaccine-0.3.1-py2.py3-none-any.whl

**Expected behavior**
I expect this URL to redirect to the correct file, but it only gives me 404:

**To Reproduce**
```console
$ curl -I 'https://files.pythonhosted.org/packages/py2.py3/t/tbvaccine/tbvaccine-0.3.1-py2.py3-none-any.whl'
HTTP/2 404 
content-type: application/octet-stream
server: nginx/1.13.9
accept-ranges: bytes
date: Thu, 07 Mar 2019 11:52:05 GMT
age: 0
content-length: 0
...
```

** Additional Info **
Weirdly, “Python version” is “3.6” on the [file list](https://pypi.org/project/tbvaccine/#files):

Filename, size & hash | File type | Python version | Upload date
-- | -- | -- | --
tbvaccine-0.3.1-py2.py3-none-any.whl (8.2 kB) | Wheel | 3.6 | Dec 14, 2018

jamadden:
The reference you referred to also ends with:

> In general this is only a good idea for `source` as a `python_version` to fetch tar and zip files.

And that's the only example it shows. Given that, I wouldn't expect this to work for .whl files.

The official guidance is to use the [provided API](
[...]

PR #5536, opened 2019-03-07
Reworded the section about predictable URLs

flying-sheep:
In #5535 it became clear that the current state of the documentation misleads people into thinking the predictable URL API might me some kind of barely maintained step child instead of the valuable asset for Linux package maintainers that it is.

# [Rendered](https://warehouse--5536.org.readthedocs.build/api-reference/integration-guide.html#predictable-urls)

### DIFF
--- a/docs/dev/api-reference/integration-guide.rst
+++ b/docs/dev/api-reference/integration-guide.rst
@@ -96,12 +96,34 @@ Query PyPI’s `JSON
 API <https://warehouse.pypa.io/api-reference/json/>`__ to
 determine where to download files from.
 
-If you so choose
+Predictable URLs
 ----------------
 
-You can use our conveyor service to fetch this file, but this
-realistically exists primarily to support that historical predictable
-URLs still work.
+You can use our conveyor service to fetch this file, which exists for
+cases where using the API is impractical or impossible. This is for
+example the case for Linux package maintainers, as package build scripts
+or package metadata expect static URLs in some cases.
+
+URLs can be constructed as follows, with wheel file names following
+:pep:`491#file-name-convention`.
+
+.. code:: python
+
+   host = 'https://files.pythonhosted.org'
+
+   def source_url(name, version):
+       return f'{host}/packages/source/{name[0]}/{name}/{name}-{version}.tar.gz'
+
+   def wheel_url(name, version, build_tag, python_tag, abi_tag, platform_tag):
+       # https://www.python.org/dev/peps/pep-0491/#file-name-convention
+       wheel_parts = {
+           tag: re.sub('[^\w\d.]+', '_', part, re.UNICODE)
+           for tag, part in locals().items()
+       }
+       wheel_parts['optional_build_tag'] = f'-{wheel_parts["build_tag"]}' if build_tag else ''
+       filename = '{name}-{version}{optional_build_tag}-{python_tag}-{abi_tag}-{platform_tag}.whl'\
+                  .format_map(wheel_parts)
+       return f'{host}/packages/{python_tag}/{name[0]}/{name}/{filename}'
 
 Example:
 ~~~~~~~~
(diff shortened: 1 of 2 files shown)


flying-sheep:
I’m confused about `python_tag` (the specified part in wheels) vs `python_version` (the part in the URLs after `/packages/` that’s mostly “source”, or the `python_tag`, but sometimes something else entirely)

Is there a spec for this? Why do some wheels have some freeform stuff in there even though their `python_tag` is correct?

- `coverage-config-reload-plugin` has `python_tag=py2.py3` and `python_version=any`
- `numericalunits` and `tbvaccine` have `python_tag=py2.py3` and `python_version=3.6`

Is that a bug in the consistency check of PyPI? Or is it allowed to have differing versions there?

There needs to be one of two things:

1. Either I need an explanation for `python_version`
2. Or we need to patch the redirection so it also makes wheels available under `/packages/{python_tag}/...` instead of having to figure out their bogus `python_version`.

flying-sheep:
PS: the travis failure is a bug in the validation: ``:pep:`number#anchor` `` is valid syntax.

pradyunsg:
Sounds like a good bugfix PR for doc8. :)

miketheman:
#Triage 
Surfacing this for reevaluation. Generally speaking, the docs update seems reasonable, but we've since changed CI providers since this was authored, and there's some conflicts that need to be resolved.

@flying-sheep are you still interested in making this contribution?

flying-sheep:
Sure, why not!

docs/api-reference/integration-guide.rst
```diff
@@ -96,12 +96,34 @@ Query PyPI’s `JSON
...
-You can use our conveyor service to fetch this file, but this
-realistically exists primarily to support that historical predictable
-URLs still work.
+You can use our conveyor service to fetch this file, which exists for
+cases where using the API is impractical or impossible. This is for
+example the case for Linux package maintainers, as package build scripts
+or package metadata expect static URLs in some cases.
+
+URLs can be constructed as follows, with wheel file names following
+:pep:`491#file-name-convention`.
```
di, line 108: Looks like the linter doesn't like this line, I'm not sure if we can use the `:pep:` directive with anchor tags like this.

flying-sheep: I answered this question 3 years ago :wink: https://github.com/pypi/warehouse/pull/5536#issuecomment-470542743

> PS: the travis failure is a bug in the validation: ``:pep:`number#anchor``` is valid syntax.

di: Indeed you did!

Since this seems like a docutils bug (that hasn't been fixed in the last 3 years) we should find a way to either override this or just make this a regular link.

di: We can make a PR to try removing the pins, but I don't see a recent release at https://pypi.org/project/doc8/ or https://pypi.org/project/docutils/, am I missing something?

di: Found the issue: https://github.com/PyCQA/doc8/issues/124. I made a PR to unconstrain these in https://github.com/pypi/warehouse/pull/12711, will update that with the latest version when the release is made.

flying-sheep: we can leave the lower bound (>=1.1.0) to this PR, that version is not needed outside of it, right?

di: I don't think it's strictly necessary, because we lock our requirements files, but it doesn't hurt.

flying-sheep: OK, I merged the main branch. It should work now!

di: The version of `doc8` with your changes still hasn't been released.

flying-sheep: oh, whoops!

di (approved):

PR #4618, opened 2018-08-24
Added multi-factor authentication for account login

mschwager:
Fixes https://github.com/pypa/warehouse/issues/996.

Hey all,

I'm hoping to get some early feedback on this MFA work. Here's the current workflow:

Visit the account management page and configure MFA

2018-08-24-092346_1162x733_scrot

Then, when logging in, you're presented with the same primary authentication page

2018-08-24-092404_982x516_scrot

If MFA is enabled for that user then you'll be asked to input an authentication token

2018-08-24-092417_1076x613_scrot

This workflow is similar to what we were discussing in IRC. There's undoubtedly some UI/UX improvements to be made, but I was hoping to get some eyes (@dstufft, @di, @sergeykolosov, others?) on this as soon as possible. Addtionally, there's some things that need to be done, and probably some that _should_ be done, before this is ready for prime time:

- Pass username/password along to MFA form when performing secondary authentication
- Allow disabling of MFA in account management
- ~Add additional MFA tests for the accounts views page~
- Spot check UI/UX/verbiage

I'm about to take off for the weekend and was hoping to get this up before then :)

PR #5567, opened 2019-03-13
2FA (TOTP) support

woodruffw:
See #4373
Issue #996

CC: @nlhkabu


woodruffw:
@ewdurbin @di @brainwane I think we're ready to do an initial review here!

We should be at 100% coverage, and the basic flow for TOTP-based 2FA is present: users can add and remove a TOTP secret from their account, provision a device, and are prompted to enter a TOTP code upon login. Recovery codes haven't been added yet; my current plan is to add them in a separate PR, but I'm open to changes to that.

Here are some things I'd like to also handle in this PR:
* QR code generation. Right now only a provisioning URL is shown; I was going to take a look at @mschwager's work in https://github.com/pypa/warehouse/pull/4618 and adapt it this implementation.
* Integrating @nlhkabu's proposed authentication UI in #5587 and additional management UI (no issue yet).

This changeset is getting pretty big, though, so I can look at splitting it apart to make reviewing easier. Let me know what you think.

warehouse/accounts/models.py
```diff
@@ -80,6 +81,9 @@ class User(SitemapMixin, db.Model):
         nullable=True,
     )
 
+    totp_secret = Column(Binary(length=20), nullable=True)
+    totp_provisioned = Column(Boolean, nullable=False, server_default=sql.false())
```
woodruffw: NTS: I don't really like this double state here (`totp_secret` holds the user's TOTP secret key, while `totp_provisioned` is only set to true once the user verifies that they've added their device correctly by entering a code), but I'm not sure about better ways to do it:

* We could pass `totp_secret` as a cookie or hidden form parameter in `manage.account.totp-secret`, but this would allow the user/page javascript to modify the TOTP secret and potentially compromise it. 
* We could regenerate `totp_secret` each time the user refreshes `manage.account.totp-secret`, but this produces a really bad UX: the user will have to scan a new QR code each time they fail to validate their device.

Is there another place we can sequester the TOTP secret state without actually updating the user model? Alternatively, is there a way to ensure cookie integrity (i.e., that the user/page hasn't modified the cookie we sent)?

ewdurbin: @woodruffw you could use [this utility module](https://github.com/pypa/warehouse/blob/1fbb4ac752e68b5840b9e09b68e44a165569bfa6/warehouse/utils/crypto.py) to create a signed payload if we choose to go the cookie route.

Alternatively Redis is readily available for stashing this state.

woodruffw:
> I am curious if there is a case to be made for separating this from the User model except by relation.

Yeah, I think so. Looks like the current `email` relation will be a good reference for this 😄

warehouse/manage/views.py
```diff
@@ -288,6 +326,61 @@ def delete_account(self):
...
+    def validate_totp_provision(self):
+        if self.request.user.totp_provisioned:
+            self.request.session.flash("TOTP already provisioned.", queue="error")
+            return HTTPSeeOther(self.request.route_path("manage.account"))
+
+        form = ProvisionTOTPForm(**self.request.POST, user_service=self.user_service)
+        totp_uri = self.user_service.totp_provisioning_uri(self.request.user.id)
+
+        if form.validate():
+            totp_value = form.totp_value.data.encode("ascii")
```
mossberg: Just wondering why the `ascii` codec is explicitly used here rather than the default `utf-8`? (I'm assuming `form.totp_value.data` is a `str`). Does `form.validate()` guarantee that the data here will be ascii?

woodruffw: No particular reason! `encode()` with the default UTF-8 encoding would suffice. I'll make this change tomorrow.

`form.validate()` does indeed ensure that the data is ASCII, by virtue of only accepting `/[0-9]{6}/`.

woodruffw:
Alright! We've got QR generation:

Screen Shot 2019-03-22 at 6 51 26 PM

The provisioning URI is also placed in the `<canvas>`'s `aria-label` attribute, for accessibility purposes.

warehouse/templates/manage/account.html
```diff
@@ -276,41 +276,46 @@ <h2 class="sub-title">Change password</h2>
...
+          <td>{% if user.totp_provisioned %}
+              <span class="badge badge--success">Enabled</span>
+              {% else %}
+              <span class="badge badge--danger">Disabled</span>
+              {% endif %}
+          </td>
+          <td class="table__action">
+            {% if user.totp_provisioned %}
+            <a href="#disable-totp" class="button button--primary">Disable</a>
+            {{ confirm_modal(
```
woodruffw: This tag and below are causing a linter assertion:

```
/home/travis/virtualenv/python3.6.6/bin/html_lint.py --printfilename --disable=optional_tag,names,protocol,extra_whitespace,concerns_separation `find ./warehouse/templates -path ./warehouse/templates/legacy -prune -o -name '*.html' -print`
  File "/home/travis/virtualenv/python3.6.6/bin/html_lint.py", line 66, in <module>
    sys.exit(html_linter.main(options))
  File "/home/travis/virtualenv/python3.6.6/lib/python3.6/site-packages/html_linter.py", line 1094, in main
    clean_html = template_remover.clean(io.open(filename).read())
  File "/home/travis/virtualenv/python3.6.6/lib/python3.6/site-packages/template_remover.py", line 258, in clean
    return _TemplateRemover(html_content, pattern).get_clean_content()
  File "/home/travis/virtualenv/python3.6.6/lib/python3.6/site-packages/template_remover.py", line 199, in get_clean_content
    assert False, 'Echo tags should be in just one line.'
AssertionError: Echo tags should be in just one line.
make: *** [lint] Error 1
The command "make lint BINDIR="$(dirname $(which python))"" exited with 2.
```

Maybe this is worth a behavior change? I find these easier to read than their single line equivalents.

Or perhaps https://github.com/deezer/template-remover isn't capable of removing multi-line tags, hence the assertion?

ewdurbin: You could do the following to collapse the width somewhat.

```
{% set title="Disable 2FA by TOTP application" %}
{% set confirm_button_label="Disable TOTP application" %}
{{ confirm_modal(title=title, confirm_name="Username", confirm_string=user.username, confirm_button_label=confirm_button_label, slug="disable-top", action=None, warning=False, confirm_string_in_title=False) }}
```

warehouse/templates/manage/account/totp-provision.html
```diff
@@ -0,0 +1,67 @@
...
+  <ol>
+    <li>When you sign into PyPI, you will be asked to provide your username and password, as usual</li>
+    <li>You will then see a screen asking for an additional authentication code</li>
+    <li>Use the application of your choice (provisioned below) to generate a code</li>
+    <li>Enter the code to access your PyPI account</li>
+  </ol>
+
+  <h2>Provisioning your application</h2>
+
+  <p>You can provision any compatible TOTP application to generate two factor authentication codes for PyPI. We recommend <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Google Authenticator for Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a>, <a href="https://support.1password.com/one-time-passwords/">1Password</a>, <a href="https://authy.com/">Authy</a> or <a href="https://lastpass.com/auth/">LastPass Authenticator</a>.</p>
```
mschwager: Perhaps we could call out [Duo Mobile](https://duo.com/product/trusted-users/two-factor-authentication/authentication-methods/totp) here.

Also, perhaps we could whittle this down to one or two suggestions. A laundry list of options will potentially confuse and overwhelm users.

mschwager: Actually, this would be a better page to link to: https://guide.duo.com/third-party-accounts

mossberg: I like this idea @mschwager, I agree that too many options could confuse some users. Google Authenticator and Duo seem like a good, minimal set of recommendations to me.

mschwager: I agree!

mossberg: @nlhkabu you originally wrote this part of the page -- any thoughts?

warehouse/templates/manage/account.html
mossberg: I ended up going with a slight variation on @ewdurbin's suggestion.

warehouse/accounts/models.py
mossberg: Summarizing a discussion @woodruffw and I had last week:

While having double state enables a user to potentially "get stuck" in the intermediate state of having a secret, but not being provisioned, any alternative where a secret key is generated and displayed, but not fully persisted could result in a bad UX scenario where a user has broken Pypi entries in their 2FA app.

For example, consider the case where:
- we generate and display a secret
- the user adds it to their device, but, for whatever reason, does not provision
- we expire that secret after some amount of time (since we do not fully persist)
- the user attempts to enable 2FA again
- we generate and display a new secret
- the user adds it to their device

The user will then have two Pypi entries in their device, of which only one is correct and usable. 

Depending on our expectations for the type of user, this might be a rare case, but in order to fully ensure good UX related to their 2FA app, we probably want to go with updating the user model when we generate the secret. Especially given that there are no real security drawbacks of having this intermediate state, it seems like a totally adequate solution to me. The only thing is that future development needs to be aware of the double totp state.

mossberg:
Hi all, after getting CI passing yesterday, I began designing a solution for one of the outstanding features to be implemented: password reverification before TOTP can be disabled. The idea is that when a user disables TOTP, they are decreasing their security, so we want to ensure they are actually the owner of the account before doing this. The initial solution I've come up with involves the following flow:

- User clicks the "disable" button, submits the form on the account settings page
- That submits to `/manage/account/totp-provision`, which does the username check (current behavior)
- User is redirected to normal login flow. They enter their username and password and reauthenticate.
- (Optional/undecided) They are redirected to TOTP reauthentication. They enter a valid TOTP code.*
- After valid reauthentication, server generates a signed "reauthentication  token" with the user id and timestamp, sends it to the client. Server stores no state.
- User is then redirected to an endpoint that actually handles the TOTP disabling. Endpoint verifies there is a valid reauthentication token, the user id within matches the current session, and the timestamp is not too old. If those pass, it then finally disables TOTP for the user id.
- User is redirected to the account management page.

feedback welcome!

*the question of whether to require 2FA during the process of disabling 2FA is an open one, and is in many ways a UX question. If a user is disabling 2FA because they lost their device, we may not want to require a code. Requiring this really only seems necessary if we anticipate a threat model where an adversary has access to an active session _and_ knows the user's password.

woodruffw:
Thanks @mossberg! I'm going to start merging your changes in https://github.com/trailofbits/warehouse/tree/tob/otp-model-refactor into this branch 😄

woodruffw:
NTS: 3ca936d will probably break a few of the services unit tests due to changes in the way we access the TOTP secret and provisioning status.

woodruffw:
Summarizing the discussion @ewdurbin and I had about requiring 2FA to disable 2FA:

* We don't want the user to use the second factor that they're trying to remove, since it's common enough for people to lose TOTP applications and hardware tokens. We'd actually be decreasing the security of an account by refusing to let a user remove a second factor that they know is lost/compromised.

* If a user has multiple second factors, we could choose one of the ones not selected for removal for confirmation. This has a (slight) security benefit, in that it requires the user to prove that they have the authority to log in fully before they can strictly decrease the security of the account. A quick survey of websites shows that most don't do this, since they treat a fully authenticated user as knowing what's best for their 2FA settings. 

* Regardless of the above, we should ensure that the user must *affirmatively* request that a second factor be removed. GitHub's current MFA UI/UX, while otherwise excellent, is a perfect example of how we *don't* want to do this: they allow the user to click a single red trashcan button to remove a second factor.

woodruffw:
Replaced the dual state (`totp_secret` + `totp_provisioned`) with a combination of a short-expiry signed cookie and just a `totp_secret` column. Extant unit tests should be passing (and should be simplified in many places), although I'll need to add a few more to handle new cases.

warehouse/templates/manage/account/totp-provision.html
webknjaz: May I ask adding FreeOTP/FreeOTP+ to the list? The reason for this is that they are open source and others are AFAIK proprietary which makes them questionable from the security recommendations PoV.

webknjaz: May I ask adding FreeOTP/FreeOTP+ to the list? The reason for this is that they are open source and others are AFAIK proprietary which makes them questionable from the security recommendations PoV.

woodruffw: FreeOTP sounds good to me! I personally used FreeOTP+ back when I was on Android, but in retrospect I'm not actually sure about recommending it -- it provides a TOTP export functionality that compromises the application's ability to behave as a true second factor.

webknjaz: Well, you can as well feed the same seed to several apps/devices and you'll get the same issue. I don't consider export a truly bad thing as it provides a backup way other than recovery codes.
If you know anything other which is open source, that would probably work as well.
But my point is that trusting proprietary software is probably more of a security risk vs. having an export feature.
BTW FreeOTP+ also uses more of nice Android APIs to protect data, like hiding previews on the app switch screen unlike Google Authenticator and others.

woodruffw: > Well, you can as well feed the same seed to several apps/devices and you'll get the same issue.

Yep, that's why we settled on the warning right below the recommended application list:

```
For security reasons, you should only provision one device/application at a time.
```

TOTP's symmetricity makes it hard for us to *enforce* good client security decisions, so I think we should make an effort to *suggest* the strongest defaults. That, IMO, includes not suggesting TOTP applications that weaken their status as a second factor by allowing users to extract TOTP secrets. It might also be a good reason to remove 1Password from that list, since also violates the second factor property.

FWIW, I *completely* agree that it's not a truly bad thing -- for expert users who understand that exports violate the second factor property. That's why I used it, back when I did! But I don't think we should encourage that as one of the default recommendations, for essentially the same reason that we shouldn't require recovery codes.

FreeOTP is the open source TOTP application that I'm most familiar with. I see that [andOTP](https://github.com/andOTP/andOTP) appears to be popular, although I don't know anything about it in terms of history/behavior. Another search on F-Droid brings up [otp-authenticator](https://github.com/0xbb/otp-authenticator). I agree that trusting a proprietary TOTP application is a security hazard, although I think applications that encourage lay-users to back up their TOTP secrets are more of a direct security risk.

webknjaz: So maybe just add another warning about that? I mean like "hey, there's other apps which you can actually verify if you're truly concerned about security"

woodruffw: Yeah that sounds good to me. I think we could integrate it directly into the recommendations, e.g. "We recommend FreeOTP, ... as open source TOTP applications. Google Authenticator, ..., are also available as proprietary alternatives." Thoughts @nlhkabu?

nlhkabu:
```suggestion
  <p>You can provision any compatible TOTP application to generate two factor authentication codes for PyPI. We recommend <a href="https://freeotp.github.io/">FreeOTP</a>, <a href="https://play.google.com/store/apps/details?id=org.liberty.android.freeotpplus&hl=en">FreeOTP+</a> (both open source), Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a>, or <a href="https://authy.com/">Authy</a> (proprietary).</p>
```

nlhkabu: ping @woodruffw - as discussed :)

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+
+        <h3 id="2fa">{{ twoFA() }}</h3>
+        <p>Users who have chosen to set up two factor authentication (2FA) on their PyPI account must provide a second method of identity verification (other than their username and password) to log in.</p>
+        <p>PyPI currently supports a single 2FA method: <a href="#totp">Generating a code through a TOTP application</a>.
+
+        <h3 id="totp">{{ totp() }}</h3>
+        <p>When enabling two factor authentication (2FA) via <a href="https://en.wikipedia.org/wiki/Time-based_One-time_Password_algorithm">TOTP</a> in your account admin, you were asked to provision an application (usually a mobile phone app) in order to generate authentication codes. Popular applications include:</p>
+        <ul>
+          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a></li>
+          <li><a href="https://support.1password.com/one-time-passwords/">1Password</a></li>
```
nlhkabu:
```suggestion
```

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+          <li><a href="https://authy.com/">Authy</a></li>
+          <li><a href="https://lastpass.com/auth/">LastPass Authenticator</a></li>
```
nlhkabu:
```suggestion
```

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+        <h3 id="2fa">{{ twoFA() }}</h3>
+        <p>Users who have chosen to set up two factor authentication (2FA) on their PyPI account must provide a second method of identity verification (other than their username and password) to log in.</p>
+        <p>PyPI currently supports a single 2FA method: <a href="#totp">Generating a code through a TOTP application</a>.
+
+        <h3 id="totp">{{ totp() }}</h3>
+        <p>When enabling two factor authentication (2FA) via <a href="https://en.wikipedia.org/wiki/Time-based_One-time_Password_algorithm">TOTP</a> in your account admin, you were asked to provision an application (usually a mobile phone app) in order to generate authentication codes. Popular applications include:</p>
+        <ul>
+          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a></li>
+          <li><a href="https://support.1password.com/one-time-passwords/">1Password</a></li>
+          <li><a href="https://authy.com/">Authy</a></li>
```
nlhkabu:
```suggestion
          <li><a href="https://authy.com/">Authy</a> (proprietary)</li>
```

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+
+        <h3 id="2fa">{{ twoFA() }}</h3>
+        <p>Users who have chosen to set up two factor authentication (2FA) on their PyPI account must provide a second method of identity verification (other than their username and password) to log in.</p>
+        <p>PyPI currently supports a single 2FA method: <a href="#totp">Generating a code through a TOTP application</a>.
+
+        <h3 id="totp">{{ totp() }}</h3>
+        <p>When enabling two factor authentication (2FA) via <a href="https://en.wikipedia.org/wiki/Time-based_One-time_Password_algorithm">TOTP</a> in your account admin, you were asked to provision an application (usually a mobile phone app) in order to generate authentication codes. Popular applications include:</p>
+        <ul>
+          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a></li>
```
nlhkabu:
```suggestion
          <li><a href="https://freeotp.github.io/">FreeOTP</a> (open source)</li>
          <li><a href="https://play.google.com/store/apps/details?id=org.liberty.android.freeotpplus&hl=en">FreeOTP+</a> (open source)</li>
          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a> (proprietary)</li>
```

warehouse/templates/pages/help.html
nlhkabu: ping @woodruffw - as discussed.

warehouse/templates/manage/account/totp-provision.html
mschwager: Per above, I believe we're planning to include [Duo Mobile](https://guide.duo.com/third-party-accounts) here. Maybe we can swap it with Authy?

warehouse/templates/pages/help.html
woodruffw: Thanks @nlhkabu! Applying now 😄

woodruffw:
Ping @ewdurbin @di @dstufft -- I think this is ready for your review!

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+
+        <h3 id="2fa">{{ twoFA() }}</h3>
+        <p>Users who have chosen to set up two factor authentication (2FA) on their PyPI account must provide a second method of identity verification (other than their username and password) to log in.</p>
+        <p>PyPI currently supports a single 2FA method: <a href="#totp">Generating a code through a TOTP application</a>.
+
+        <h3 id="totp">{{ totp() }}</h3>
+        <p>When enabling two factor authentication (2FA) via <a href="https://en.wikipedia.org/wiki/Time-based_One-time_Password_algorithm">TOTP</a> in your account admin, you were asked to provision an application (usually a mobile phone app) in order to generate authentication codes. Popular applications include:</p>
+        <ul>
+          <li><a href="https://freeotp.github.io/">FreeOTP</a> (open source)</li>
+          <li><a href="https://play.google.com/store/apps/details?id=org.liberty.android.freeotpplus&hl=en">FreeOTP+</a> (open source)</li>
```
webknjaz: I believe it's better to not include language query param in the link.
```suggestion
          <li><a href="https://play.google.com/store/apps/details?id=org.liberty.android.freeotpplus">FreeOTP+</a> (open source)</li>
```

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,26 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a> (proprietary)</li>
```
webknjaz:
```suggestion
          <li>Google Authenticator for <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2">Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a> (proprietary)</li>
```

webknjaz: Maybe, iTunes link can be changed but I don't know anything about that store so sending only Google link patch for now.

woodruffw:
Thanks @webknjaz!

warehouse/manage/views.py
```diff
@@ -288,6 +300,119 @@ def delete_account(self):
...
+        except TokenException:
+            token_data = {
+                "totp_secret": urlsafe_b64encode(otp.generate_totp_secret()).decode()
+            }
+            self.request.response.set_cookie(
+                TWO_FACTOR_TOTP_PROVISION_COOKIE, self.token_service.dumps(token_data)
+            )
+
+        totp_secret = urlsafe_b64decode(token_data["totp_secret"])
+        totp_uri = otp.generate_totp_provisioning_uri(
```
ewdurbin: We should use the `site.name` configuration variable here for the `issuer_name`.

woodruffw: Got it! Will fix tomorrow morning.

woodruffw:
Alright, `issue_name` change done. I've also added a `<noscript>` block to the TOTP provisioning page, which just displays the provisioning URI for users whose clients won't render the QR in Javascript.

warehouse/accounts/models.py
```diff
@@ -111,6 +120,22 @@ class UnverifyReasons(enum.Enum):
     SoftBounce = "soft bounce"
 
 
+class TwoFactor(db.ModelBase):
+    __tablename__ = "two_factor"
+    __table_args__ = (
+        UniqueConstraint("user_id", name="two_factor_user_key"),
+        Index("two_factor_user_id", "user_id"),
```
dstufft: If I remember correctly, this is going to create a duplicate index to the ``UniqueConstraint`` above, just one with and without the unique flag set. We should double check that, but I think we can drop the ``Index`` here on PostgreSQL.

warehouse/accounts/models.py
```diff
@@ -111,6 +120,22 @@ class UnverifyReasons(enum.Enum):
...
+    )
+
+    id = Column(Integer, primary_key=True, nullable=False)
```
dstufft: Small nit, I know we use integer IDs elsewhere, but I generally prefer using UUIDs for a primary key. You can actually just delete this line, and inherit from ``db.Model`` instead to have the primary key taken care of for you automatically. We generally only use ``db.ModelBase`` for things that have primary keys that pre-date Warehouse.

warehouse/accounts/services.py
```diff
@@ -77,6 +79,20 @@ def get_user(self, userid):
...
+
+        if result:
+            return result
+        else:
+            user = self.get_user(user_id)
+            if not user:
+                return None
+            two_factor = TwoFactor(user=user)
+            self.db.add(two_factor)
+            self.db.flush()
```
dstufft: Is this flush really required? It might be, but normally SQLAlchemy should handle flushing as needed behind the scenes.

warehouse/accounts/services.py
```diff
@@ -211,6 +227,14 @@ def update_user(self, user_id, **changes):
 
         return user
 
+    def update_two_factor(self, user_id, **changes):
+        two_factor = self.get_two_factor(user_id)
```
dstufft: This can return ``None``, which is going to generate an error like ``AttributeError: "NoneType" object has no attribute "whatever"``. We should probably just make that an explicit error with a meaningful message instead.

warehouse/accounts/services.py
```diff
@@ -226,6 +250,30 @@ def is_disabled(self, user_id):
         else:
             return (True, user.disabled_for)
 
+    def has_two_factor(self, user_id):
+        """
+        Returns True if the user has any form of two factor
+        authentication.
+        """
+        two_factor = self.get_two_factor(user_id)
```
dstufft: Likewise to the above, ``two_factor`` can be ``None`` here, which will generate a less than ideal exception below.

warehouse/accounts/services.py
```diff
@@ -226,6 +250,30 @@ def is_disabled(self, user_id):
...
+        # will also go.
+        return two_factor.totp_secret is not None
+
+    def check_totp_value(self, user_id, totp_value):
+        """
+        Returns True if the given TOTP is valid against the user's secret.
+
+        If the user doesn't have a TOTP secret, returns False.
+        """
+        two_factor = self.get_two_factor(user_id)
```
dstufft: Again with the ``two_factor`` can be ``None`` thing. Might make sense to just fix the ``get_two_factor`` function to raise an error so these can stay more focused.

warehouse/accounts/views.py
```diff
@@ -122,38 +126,123 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+        return HTTPSeeOther(request.route_path("accounts.login"))
+
+    redirect_to = token_data.get("redirect_to")
+
+    user_service = request.find_service(IUserService, context=None)
+
+    form = _form_class(
+        request.POST,
+        user_service=user_service,
+        check_password_metrics_tags=["auth_method:two_factor_form"],
```
dstufft: I think we want these metric tags to match what is on the normal login form: ``["method:auth", "auth_method:login_form"]``.

warehouse/accounts/views.py
```diff
@@ -122,38 +126,123 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+    )
+
+    if request.method == "POST":
+        if form.validate():
+            totp_value = form.totp_value.data.encode()
+            if not user_service.check_totp_value(userid, totp_value):
```
dstufft: Is there a reason to validate this here instead of making it part of the form validation?

warehouse/accounts/views.py
```diff
@@ -122,38 +126,123 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+def two_factor(request, _form_class=TwoFactorForm):
+    if request.authenticated_userid is not None:
+        return HTTPSeeOther(request.route_path("manage.projects"))
+
+    token_service = request.find_service(ITokenService, name="two_factor")
+
+    # Load token data from cookies or
+    # redirect the user to login page again if any token doesn't exist.
+    try:
+        token_data = token_service.loads(request.cookies.get(TWO_FACTOR_COOKIE_KEY))
```
dstufft: Is there a reason to pass this via a cookie (which will be sent to every request made to PyPI until it gets successfully deleted) instead of encoding it in the URL? If I understand correctly, the only data contained here is the userid and the redirect_to, neither of which are sensitive and can't be stored in the URL.

Modifying "global" state like a cookie to handle a multi step transaction always makes me a bit nervous versus putting it in the URL if possible.

warehouse/accounts/views.py
```diff
@@ -456,7 +547,6 @@ def _login_user(request, userid):
 
     # Whenever we log in the user, we want to update their user so that it
     # records when the last login was.
-    user_service = request.find_service(IUserService, context=None)
```
dstufft: Is there a point to this change? It seems kind of churny for churn sake, but I might just be missing it.

warehouse/manage/views.py
```diff
@@ -288,6 +300,121 @@ def delete_account(self):
...
+    def __init__(self, request):
+        self.request = request
+        self.user_service = request.find_service(IUserService, context=None)
+        self.token_service = request.find_service(ITokenService, name="totp_provision")
+
+    @property
+    def default_response(self):
+        return HTTPSeeOther(self.request.route_path("manage.account"))
+
+    @view_config(request_method="GET")
```
dstufft: We probably want to set this page to no caching ever. e.g. ``Cache-Control: no-store``. I don't think we have a mechanism for doing that currently though, so it might make sense to just raise is as an issue.

warehouse/manage/views.py
```diff
@@ -288,6 +300,121 @@ def delete_account(self):
...
+
+        try:
+            token_data = self.token_service.loads(
+                self.request.cookies.get(TWO_FACTOR_TOTP_PROVISION_COOKIE)
+            )
+        except TokenException:
+            token_data = {
+                "totp_secret": urlsafe_b64encode(otp.generate_totp_secret()).decode()
+            }
+            self.request.response.set_cookie(
```
dstufft: Isn't this putting the TOTP secret in the cookie store of the browser in cleartext? What's the rationale for doing that versus storing it in the server side session data? If it needs to be in a cookie should we be encrypting it at least?

I know there was some discussion related to this so I am probably just missing context here :)

warehouse/manage/views.py
```diff
@@ -288,6 +300,121 @@ def delete_account(self):
...
+                self.request.cookies.get(TWO_FACTOR_TOTP_PROVISION_COOKIE)
+            )
+        except TokenException:
+            self.request.session.flash("Invalid or expired TOTP cookie.", queue="error")
+            return self.default_response
+
+        totp_secret = urlsafe_b64decode(token_data["totp_secret"])
+        totp_value = form.totp_value.data.encode()
+
+        if not otp.verify_totp(totp_secret, totp_value):
```
dstufft: Any reason not to do this validation inside the form itself so that errors get handled like form errors?

warehouse/migrations/versions/42f0409bb702_add_two_factor_and_update_users.py
```diff
@@ -0,0 +1,76 @@
...
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
dstufft: Is this table correct? I don't see the unqiue constraint being added. It looks like you can have multiple rows all referring to the same user, but your model above has a unqiue constraint. Did you forget to update the migration after making a code change?

warehouse/migrations/versions/c0c0544354e7_add_totp_secret_field.py
```diff
@@ -0,0 +1,58 @@
...
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
```
dstufft: This migration doesn't seem like it matches the current state of the models at all, and we then immediately drop these columns in the next migration. Doesn't seem worthwhile to keep it in this PR?

warehouse/accounts/models.py
```diff
@@ -111,6 +120,22 @@ class UnverifyReasons(enum.Enum):
     SoftBounce = "soft bounce"
 
 
+class TwoFactor(db.ModelBase):
```
dstufft: Is there a reason to have a second table (with ``lazy=False``, so a join will be generated by default in every query) and an index to store a single 20 byte binary? It seems like it should be just as fine to add the ``totp_secret`` column to the ``UserModel`` itself. The only reason I can think of for making it a separate table is if we plan on allowing multiple TOTP secrets to be registered per user, but currently there is a unique constraint preventing that (and the code base is not setup to handle multiple).

dstufft:
This might be a silly question, but is there a reason we're generating the QRCode in Javascript, which requires the user to have JS enabled, versus doing it in Python using something like the [qrcode](https://pypi.org/project/qrcode/) library? Our JS testing story is not great in general, so putting as much work into Python seems like the better overall play to me.

webknjaz:
I just want to add that there's even services where you can put data right into the URI and put that into an `` tag. Like Google Charts. So this can be done even without any backend code if needed.

ewdurbin:
> I just want to add that there's even services where you can put data right into the URI and put that into an `` tag. Like Google Charts. So this can be done even without any backend code if needed.

I'd prefer if we generated the QR codes ourselves rather than sending TOTP secrets to a third-party service.

warehouse/accounts/services.py
woodruffw: I added that `flush` based on `create_user` and `add_email`, although both of those perform the explicit flush to ensure that the `id` attribute is immediately available. I'll try removing it here and seeing if anything explodes 😄

warehouse/accounts/services.py
woodruffw: Sounds good to me! Is this just a matter of throwing a better exception, or is there a different way to propagate an error from here? Many of the other service calls (e.g. `add_email`, `update_user`) appear to ignore the case where `get_user` returns `None`, but maybe I'm missing some implicit checking?

warehouse/accounts/services.py
woodruffw: Ah, I guess this answers my question in https://github.com/pypa/warehouse/pull/5567#discussion_r273292255 😄. I'll make `get_two_factor` raise an error.

warehouse/accounts/views.py
woodruffw: Nope, no reason that I can think of. This code comes from the original branch that I began from, so there might be a (no longer relevant) historical cause. I'll stuff the redirect and user ID into the URL.

warehouse/manage/views.py
woodruffw: I'm not extremely familiar with Pyramid, but I *think* we can do this with `@view_config(..., http_cache=0)`. I'll test this tomorrow and see if it sets the headers correctly.

Relevant docs: 

* https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_view

* https://docs.pylonsproject.org/projects/pyramid/en/latest/api/view.html#pyramid.view.view_config

warehouse/manage/views.py
woodruffw: Yup! The context here is the dual-state discussion + my commit message in https://github.com/pypa/warehouse/pull/5567/commits/37ed90dce638d180079bcf653a7b6e0acbe1ce09.

In sum: I used a cookie here because I didn't like my workaround of having two separate TOTP columns + a potentially stale TOTP secret, and because I wasn't familiar with the Warehouse helpers/idioms for storing session data on the server-side. If you have any pointers on that, I'd greatly appreciate them!

As for encrypting: TOTP is fully symmetric, so we need to expose the TOTP secret entirely anyways: it's in the provisioning URI as a base32-encoded string. The cookie store is definitely a wider security scope though (especially if the cookie somehow outlives the provisioning flow), so this is another good reason to switch to server-side state.

warehouse/migrations/versions/42f0409bb702_add_two_factor_and_update_users.py
woodruffw: Yeah, I think I dropped the migration changes while merging the DB changes last week. Thanks!

warehouse/migrations/versions/c0c0544354e7_add_totp_secret_field.py
woodruffw: Yep, this is old. I'll remove it.

warehouse/accounts/models.py
woodruffw: The idea here was to provide breathing space for the additional column(s) that U2F/Webauthn will require (and therefore to minimize the number of `User` migrations`).

In retrospect, though, that's a little shaky: there's no real semantic connection between TOTP and U2F/Webauthn data, and having them share the same table would make configurations like "one TOTP, many security keys" difficult to model. In light of that, we could drop this back down to a single column on the `User` model and add a separate `SecurityKey` model a la `Email` when the time comes. Thoughts?

woodruffw:
> This might be a silly question, but is there a reason we're generating the QRCode in Javascript, which requires the user to have JS enabled, versus doing it in Python using something like the [qrcode](https://pypi.org/project/qrcode/) library?

Not a silly question! I cribbed the QR code generation from @mschwager's work in #4618, but two reasons stand out to me:

1. No need to generate an image file, serve it, and remember to destroy it on the server side. As such, we minimize the chances of accidentally storing/caching the image (which contains the TOTP secret) somewhere unintended.
    * Maybe Warehouse/Pyramid already has primitives for exactly this purpose? If so, we could use those.
    * To avoid the generate-save-destroy cycle, we could use an inlined base64-encoded image. I don't know of any *particular* downsides to this, other than finding it a little clunky compared to a `<canvas>`. Data URIs are also apparently not a strong-point for consistent behavior across browsers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

2. I generally trust client-side Javascript rendering into a `<canvas>` more than I trust Python + potentially native code (Pillow) rendering into a PNG/SVG. This is a little bit irrational on my part, since the input in question (the provisioning URL) is almost entirely server controlled and Pillow's security record is nowhere near as bad as ImageMagick's.

There's also a (minor) accessibility perk: we currently get [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) compatibility for free because we *need* to send the provisioning URI for QR generation on the client side. If we generate the QR on the server side, we'll need to special-case that sending. But that's not hard to do at all.

warehouse/manage/views.py
woodruffw: Nope, none that I can think of. I'll make this change.

warehouse/accounts/views.py
woodruffw: Nope, same as TOTP checking during provisioning 👍

warehouse/accounts/views.py
woodruffw: I don't think there is; it was part of the original branch's changeset. I'll test and see what, if anything, breaks.

brainwane:
Sounds like we're waiting on a decision from @ewdurbin and @di on the QR code generation? Is that something we could decide easily to change later, so we can get this merged and start running tests on test.pypi? (Ok if the answer is no!)

warehouse/manage/views.py
woodruffw: Confirmed that `http_cache=0` works and adds the relevant response headers. I'll commit this in a moment.

ewdurbin:
Discussed with @di and QR codes should be generated within the Python codebase and returned to the user.

I'd recommend https://pypi.org/project/PyQRCode/ with https://pypi.org/project/pypng/ for generation rather than something based on PIL/Pillow.

As far as the machinery goes, It might make sense to have a single endpoint that generates the image based on the contents of the signed/timestamped cookie if they are valid. This absolves the need for us to store any images, they can just be rendered on the fly.

Thoughts @woodruffw?

ewdurbin:
As far as accessibility goes, we can still include the aria elements on the rendered image, right?

woodruffw:
> I'd recommend https://pypi.org/project/PyQRCode/ with https://pypi.org/project/pypng/ for generation rather than something based on PIL/Pillow.

👍 


> As far as the machinery goes, It might make sense to have a single endpoint that generates the image based on the contents of the signed/timestamped cookie if they are valid. This absolves the need for us to store any images, they can just be rendered on the fly.

Just to make sure I understand here: are you thinking something like a `manage/account/totp-provision/image` endpoint that, when hit via the ``, generates and responds with the PNG data? If so, that sounds good to me.

warehouse/accounts/forms.py
```diff
@@ -32,6 +33,18 @@ def validate_username(self, field):
             raise wtforms.validators.ValidationError("No user found with that username")
 
 
+class TOTPValueMixin:
+
+    totp_value = wtforms.StringField(
+        validators=[
+            wtforms.validators.DataRequired(),
+            wtforms.validators.Regexp(
+                rf"^[0-9]{{{TOTP_LENGTH}}}$", message="TOTP code must be 6 digits."
```
reaperhulk: This is a nitpick (especially since this length is unlikely to ever change), but the message should probably be `message=f"TOTP code must be {TOTP_LENGTH} digits."`

warehouse/utils/otp.py
```diff
@@ -0,0 +1,82 @@
...
+    Verifies an OTP value and time against the given TOTP object.
+    """
+    try:
+        totp.verify(value, time)
+        return True
+    except InvalidToken:
+        return False
+
+
+def verify_totp(secret, value):
```
reaperhulk: I could probably be convinced that this logic should be hoisted up into `cryptography` for a `fuzzy_verify`, but it's fine here for now.

tests/unit/utils/test_otp.py
```diff
@@ -0,0 +1,87 @@
...
+    query = parse_qsl(parsed.query)
+
+    assert ("digits", "6") in query
+    assert ("secret", b32encode(secret).decode()) in query
+    assert ("algorithm", "SHA1") in query
+    assert ("issuer", issuer_name) in query
+    assert ("period", "30") in query
+
+
+def test_verify_totp_success():
```
reaperhulk: this test can be parametrized to cover success, positive, and negative skew using `@pytest.mark.parametrize` to reduce duplication in the tests.

tests/unit/utils/test_otp.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_verify_totp_success_positive_skew():
+    secret = generate_totp_secret()
+    totp = TOTP(secret, TOTP_LENGTH, SHA1(), TOTP_INTERVAL, backend=default_backend())
+    value = totp.generate(time.time() + 20)
+    assert verify_totp(secret, value)
+
+
+def test_verify_totp_failure():
```
reaperhulk: I'd like to see this parametrized with multiple positive/negative skews that should cause failures.

warehouse/accounts/__init__.py
```diff
@@ -124,6 +124,12 @@ def includeme(config):
     config.register_service_factory(
         TokenServiceFactory(name="email"), ITokenService, name="email"
     )
+    config.register_service_factory(
```
woodruffw: NTS: These can be removed now that we're no longer stashing any 2FA/TOTP state in cookies.

warehouse/utils/otp.py
woodruffw: Yeah, it'd definitely be nice to have this in `cryptography`!

Additional support: Section 6 in the RFC suggests exactly this behavior (current interval + previous and next): https://tools.ietf.org/html/rfc6238#section-6

woodruffw:
> Maybe land the OTP utility module + base tests as a separate PR and rebase this one?

I'm all for this, provided that it helps speed up the review & merge process. Will this help on your side @ewdurbin / @di / @dstufft?

warehouse/accounts/views.py
```diff
@@ -122,38 +123,102 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+    renderer="accounts/two-factor.html",
+    uses_session=True,
+    require_csrf=True,
+    require_methods=False,
+)
+def two_factor(request, _form_class=TwoFactorForm):
+    if request.authenticated_userid is not None:
+        return HTTPSeeOther(request.route_path("manage.projects"))
+
+    userid = request.params.get("userid")
```
dstufft: I *think* we need to sign this data (or stash it in the server side session data, but signing is probably good enough) because right now I think this allows people to degrade 2fa auth into a single factor (the TOTP token). Basically this currently lets someone just hit this URL directly and completely bypass the need to type in a password at all (unless I'm misinterpreting this?).

Perhaps It would be better to roll this page into the normal login page, and make the login page do something like:

1. If Get, render the 1fa log in form.
2. If Post, validate the 1fa login form.
  1. If Invalid, render 1fa login form with error.
  2. If valid, check if user has 2fa, if so render 2fa login form (subclass of 1fa login form) with username/pw already filled out.

Does that make sense? That removes the need to shuffle data around at all, and means that we always check if the username, password and totp values are all valid.

One possible riff on this idea is to check if the user has 2fa enabled before we check the password at all, and if they do render the 2fa form first, and skip validating the pw in that case. That means we would leak the information if a person has TOTP enabled or not, but it also means you can't brute force password validation I think? 

Thoughts?

warehouse/accounts/models.py
```diff
@@ -111,6 +120,18 @@ class UnverifyReasons(enum.Enum):
     SoftBounce = "soft bounce"
 
 
+class TwoFactor(db.Model):
```
dstufft: I'm still not a huge fan of this extra model to store 20 bytes. I liked the idea you posted somewhere else of just moving ``totp_secret`` to the ``User`` model, and then when security keys get added, adding a dedicated model for those ala ``Email``.

warehouse/accounts/views.py
woodruffw: > Basically this currently lets someone just hit this URL directly and completely bypass the need to type in a password at all (unless I'm misinterpreting this?).

Yeah, this looks correct. Major oversight on my part -- I thought that `uses_session` on a `view_config` provided implicit integrity, but that's clearly wrong 🤐

Any pointers on signing query parameters (is this something we can do?), or are you thinking we should just revert to the earlier signed cookie version?

CCing @nlhkabu for thoughts on merging this with the normal login page: my big concern there is that things will get complicated quickly as we add multiple second factor types.

Re: Checking 2FA first: I can't think of any security reason (other than the information leakage) why this wouldn't be fine, although I've never seen it done before. Password bruteforcing should already be capped by your rate limiter, so I would err on the conservative side here and suggest that we only test the second factor after confirming the user's password.

warehouse/accounts/models.py
woodruffw: Alright, sounds good. I'll revert back to having `totp_secret` on the `User` model directly.

warehouse/accounts/views.py
dstufft: Ah yea, ``uses_session`` just controls whether you can access ``request.session`` or not. Anything stored inside of ``request.session`` is securely tied to that specific session, but beyond that, ``uses_session`` doesn't have any real security implications. It exists primarily for caching (anything that uses the session can't be cached globally).

We've never put any signed content into the URL before, but you can use ``warehouse.utils.crypto`` to get a signer that will let you sign an opaque string.

The other option is to just stick the user id in ``request.session`` in the login view before we redirect to the 2fa page. That allows us not to have any need for crypto in the login flow, but it does mean that our login flow becomes stateful server side which is maybe kind of weird? I'm not sure which option I prefer, but I think I prefer the crypto solution.

nlhkabu: > One possible riff on this idea is to check if the user has 2fa enabled before we check the password at all, and if they do render the 2fa form first, and skip validating the pw in that case. That means we would leak the information if a person has TOTP enabled or not, but it also means you can't brute force password validation I think?

If we don't ask for a password, would this not just be a single factor login?  

> thoughts on merging this with the normal login page: my big concern there is that things will get complicated quickly as we add multiple second factor types

My strong preference is to keep these pages separate, as this is the typical work flow on other 2fa interfaces.

warehouse/templates/manage/manage_base.html
```diff
@@ -87,23 +102,25 @@ <h3 class="sidebar-section__title">Your account</h3>
...
-            {% endif %}
-          </p>
-        </div>
-        {% set context = 'your' if confirm_name.lower() == 'username' else 'the' %}
-        {{ extra_description if extra_description else '' }}
-        <p>Confirm {{ context }} {{ confirm_name|lower }} to continue.</p>
-        {% set name = "confirm_" + confirm_name.lower().replace(' ', '_') %}
-        <label for="{{ name }}">{{ confirm_name }}</label>
-        <input name="{{ name }}" data-action="input->confirm#check" data-target="confirm.input" type="text" autocomplete="off" autocorrect="off" autocapitalize="off">
+          <h3 class="modal__title">{{ title }} {% if confirm_string_in_title %}{{ confirm_string }}{% endif %}?</h3>
```
di:
```suggestion
          <h3 class="modal__title">{{ title }}{% if confirm_string_in_title %} {{ confirm_string }}{% endif %}?</h3>
```
This is adding an extra space before the question mark when `confirm_string_in_title` is false.

warehouse/accounts/services.py
```diff
@@ -226,6 +228,56 @@ def is_disabled(self, user_id):
...
+
+    def check_totp_value(self, user_id, totp_value, *, tags=None):
+        tags = tags if tags is not None else []
+
+        self._metrics.increment("warehouse.authentication.two_factor.start", tags=tags)
+        """
+        Returns True if the given TOTP is valid against the user's secret.
+
+        If the user doesn't have a TOTP secret, returns False.
+        """
```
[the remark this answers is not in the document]
woodruffw: Whoops, yeah.

warehouse/templates/manage/account.html
```diff
@@ -189,7 +189,7 @@ <h2 class="sub-title">Account details</h2>
   {{ form_error_anchor(add_email_form) }}
   <section id="account-emails">
     <h2 class="sub-title">Account emails</h2>
-    <p>You can associate several emails with your account. You can use any <span class="badge badge--success"><i class="fa fa-check"></i> Verified</span> email to recover your account, but only your <span class="badge">Primary</span> email will receive notifications.</p>
+    <p>You can associate several emails with your account. You can use any <strong>verified</strong> email to recover your account, but only your <strong>primary</strong> email will receive notifications.</p>
```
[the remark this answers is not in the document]
woodruffw: Nope, probably just got caught up while integrating the help text. I'll fix it.

warehouse/accounts/views.py
```diff
@@ -122,38 +124,113 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+    require_methods=False,
+)
+def two_factor(request, _form_class=TwoFactorForm):
+    if request.authenticated_userid is not None:
+        return HTTPSeeOther(request.route_path("manage.projects"))
+
+    token_service = request.find_service(ITokenService, name="two_factor")
+
+    try:
+        two_factor_data = token_service.loads(request.query_string)
```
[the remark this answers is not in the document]
woodruffw: Got it. Anything I need to do in the codebase to make that happen?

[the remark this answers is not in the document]
di: Nope, this is something we need to configure before this gets merged.

warehouse/templates/manage/manage_base.html
```diff
@@ -34,6 +34,8 @@
     </span>
   </div>
   {% endif %}
+
+  <!-- TODO: Missing two factor warning -->
```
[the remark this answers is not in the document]
woodruffw: Yeah, there should be a warning for missing second factors. I actually had this stashed, will commit in a moment.

warehouse/templates/manage/manage_base.html
```diff
@@ -34,6 +34,18 @@
...
   </div>
   {% endif %}
+
+  {% if request.user.two_factor_allowed and not request.user.has_two_factor %}
+  <div class="notification-bar notification-bar--danger">
+    <span class="notification-bar__message">
+      <i class="fa fa-exclamation-triangle" aria-hidden="true"><span class="sr-only">Warning</span></i>
+      Two factor authentication is disabled.
+      <a href="{{ request.route_path('manage.account', _anchor="two-factor") }}">
+        Add a second factor.
```
di: I think this warning might be a little too scary/prominent. I'm concerned about a few things:
* users who don't understand what two factor authentication is being confused that something went wrong when logging in, or when they're not on the account management page;
* users who don't want to enable 2FA being unable to dismiss this;
* general fatigue/disregard for these warnings.

I do think we should "announce" this to users somehow, though. I think if we changed this to a dismissable yellow warning that said something like "Two-factor authentication is available, [enable it now for your account](url)." I'd feel better about this.

warehouse/templates/manage/account/totp-provision.html
```diff
@@ -35,11 +35,9 @@ <h2>How 2FA with TOTP works</h2>
 
   <h2>Provisioning your application</h2>
 
-  <p>You can provision any compatible TOTP application to generate two factor authentication codes for PyPI. We recommend <a href="https://play.google.com/store/apps/details?id=com.google.android.apps.authenticator2&hl=en_GB">Google Authenticator for Android</a> or <a href="https://itunes.apple.com/gb/app/google-authenticator/id388497605?mt=8">iPhone</a>, <a href="https://support.1password.com/one-time-passwords/">1Password</a>, <a href="https://authy.com/">Authy</a> or <a href="https://lastpass.com/auth/">LastPass Authenticator</a>.</p>
-
   <p>For security reasons, you should only provision one device/application at a time.</p>
 
-  <p><strong>Scan the QR code with your application of choice to provision the application:</strong></p>
+  <p><strong>Scan the QR code with your application of choice:</strong></p>
```
di: Let's link to the question on the help page here w/ the links to the various applications.

warehouse/accounts/views.py
```diff
@@ -122,38 +124,113 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
+                # We'll use this cookie so that client side javascript can
+                # Determine the actual user ID (not username, user ID). This is
+                # *not* a security sensitive context and it *MUST* not be used
+                # where security matters.
+                #
+                # We'll also hash this value just to avoid leaking the actual User
+                # IDs here, even though it really shouldn't matter.
+                resp.set_cookie(
+                    USER_ID_INSECURE_COOKIE,
+                    hashlib.blake2b(str(userid).encode(), person=b"warehouse.userid")
```
di: The diff is a little hard to read here, but this was changed from being explicitly ASCII-encoded to using the default encoding. It's not clear to me why?

warehouse/templates/manage/manage_base.html
woodruffw: > I do think we should "announce" this to users somehow, though. I think if we changed this to a dismissable yellow warning that said something like "Two-factor authentication is available, [enable it now for your account](url)." I'd feel better about this.

This sounds good to me! I'll move it from `manage_base.html` to `account.html` as well, so that users will only see it during account management.

warehouse/accounts/views.py
reaperhulk: This PR mostly uses `encode()` since the default arg is `utf8` (which works fine here of course). However, a quick grep of the warehouse codebase appears to show a preference for explicit encoding arguments so it should probably be changed across this whole PR. What do you think @woodruffw?

woodruffw: Looks like this got caught up in some churn during the TOTP cookie changes. It shouldn't affect the correctness of the `set_cookie`, but I'll roll it back since it's vestigial.

woodruffw: That's what I get for commenting without refreshing! I agree with @reaperhulk, will also add explicit encodings to the rest of this changeset.

warehouse/templates/manage/account/totp-provision.html
woodruffw: Done!

warehouse/templates/manage/manage_base.html
woodruffw: Kept it in `manage_base.html` per off-thread discussion.

nlhkabu:
Looks good to me - thank you for your work on this @woodruffw.  There are a couple of UI changes I'd like to propose, but I think it will be much, much easier to make a separate PR for these later :)

warehouse/accounts/views.py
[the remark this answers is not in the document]
dstufft: https://github.com/python/pypi-infra/pull/42

pradyunsg:
Yay!

woodruffw:
#5795 is the successor to this PR.

warehouse/templates/pages/help.html
```diff
@@ -145,12 +164,27 @@ <h3 id="verified-email">{{ verified_email() }}</h3>
...
+        <p>During each of these processes, PyPI generates a SHA-1 hash of the supplied password and uses the first five (5) characters of the hash to check the <a href="https://haveibeenpwned.com/API/v2#SearchingPwnedPasswordsByRange" target="_blank" rel="noopener">Have I Been Pwned API</a> and determine if the password has been previously compromised. The plaintext password is never stored by PyPI or submitted to the Have I Been Pwned API.</p>
+        <p>PyPI will not allow such passwords to be used when setting a password at registration or updating your password.</p>
+        <p>If you receive an error message saying that "This password appears in a breach or has been compromised and cannot be used", you should change it all other places that you use it as soon as possible.</p>
+        <p>If you have received this error while attempting to log in or upload to PyPI, then your password has been reset and you cannot log in to PyPI until you <a href="{{ request.route_url('accounts.request-password-reset') }}">reset your password</a>.</p>
+
+        <h3 id="2fa">{{ twoFA() }}</h3>
```
merwok: Unless HTML5 changed this, ID values should not start with a number.

brainwane: Yup, fixed in #6110.

merwok: Ha nice!  Of course, I didn’t find a clear spec before I posted my message, and soon after I posted I found out that requirements were relaxed to «at least 1 char, no whitespace».

PR #5609, opened 2019-03-22
Validate uploaded .tar.gz files.

gpshead:
This adds basic validation of uploaded tar files.  It checks that they are a .tar file and contain a PKG-INFO file.  The same check as performed on zip/egg files.  Plain tar, gz, and bz2 are supported.

For https://github.com/pypa/warehouse/issues/5584.


gpshead:
Huh, neat, no other tests run by make tests are failing.  Color me surprised.

warehouse/forklift/legacy.py
```diff
@@ -569,7 +572,27 @@ def _is_valid_dist_file(filename, filetype):
...
+            with tarfile.open(filename, f"r:{z_type}") as tar:
+                # This decompresses the entire stream until we find the tar
+                # file member we want.  Easy to DoS attack our CPU. :(
+                bad_tar = True
+                member = tar.next()
+                while member:
+                    parts = os.path.split(member.name)
+                    if len(parts) == 2 and parts[1] == "PKG-INFO":
+                        bad_tar = False
+                        break
```
gpshead: Breaking out here is optimizing for performance at the cost of potentially not detecting truncated or corrupted tar files that fail to parse beyond this point.

I'll leave it to the project owners to decide which behavior they want.

gpshead: i removed the break, it would allow invalid tars which defeats the point of validation.  cpu consumption is going to happen here no matter what and can be forced by any upload regardless of it.

warehouse/forklift/legacy.py
```diff
@@ -569,7 +572,26 @@ def _is_valid_dist_file(filename, filetype):
...
+        try:
+            with tarfile.open(filename, f"r:{z_type}") as tar:
+                # This decompresses the entire stream until we find the tar
+                # file member we want.  Easy to DoS attack our CPU. :(
+                bad_tar = True
+                member = tar.next()
+                while member:
+                    parts = os.path.split(member.name)
+                    if len(parts) == 2 and parts[1] == "PKG-INFO":
+                        bad_tar = False
```
cjerdonek: Your code comment above suggests you meant to break here. Is there a reason not to?

cjerdonek: I see your "review" comment now which was hidden / collapsed. I would suggest updating your code comment with an explanation of why you're not breaking (basically what is in your review comments).

gpshead: hah yes, thanks for noticing.  that's what i get for using github's inline editor to quickly remove one line.  updated! :)

cjerdonek: I like how you also updated `:(` to `:/`. Is that an upgrade or downgrade in the severity level? :P

gpshead: Who knows? Denial of service is always easy. The only winning move is not to serve. ;)

tests/unit/forklift/test_legacy.py
```diff
@@ -429,6 +453,41 @@ def test_bails_with_invalid_zipfile(self, tmpdir, filename, filetype):
...
+            fp.write(b"Dummy data file.")
+
+        with tarfile.open(tar_fn, "w:" + compression) as tar:
+            tar.add(data_file, arcname="package/module.py")
+
+        assert not legacy._is_valid_dist_file(
+            tar_fn, "sdist"
+        ), "no PKG-INFO; should fail"
+
+        os.unlink(tar_fn)
```
berkerpeksag: Instead of unlinking `tar_fn` here, wouldn't it be better to make this a separate test?

tests/unit/forklift/test_legacy.py
```diff
@@ -56,6 +58,28 @@
 )
 
 
+def _get_tar_testdata(compression_type=""):
+    temp_f = io.BytesIO()
+    with tarfile.open(fileobj=temp_f, mode="w:" + compression_type) as tar:
```
berkerpeksag: Style nit: I saw you've used an f-string in a similar code. Perhaps we could change `"w:" + compression_type` to `f"w:{compression_type}"`?

brainwane:
@gpshead hi there -- ping re: resolving @berkerpeksag's comments. Once you do that I'll ask a maintainer to look it over and, I hope, merge. Thanks!

tests/unit/forklift/test_legacy.py
gpshead: good idea. done.

gpshead:
sure, just let me know when to rebase and i'll dive into git heck for a bit to remember out how to do that :)

dstufft:
FWIW you don't have to rebase, you can just merge master into it too, we squash merge PRs anyways so you don't have to worry about keeping history clean. Whatever is easiest for you.

ewdurbin:
I guess for some reason I had in my head that we’d get more info if we ship metrics first, but realize now, we wouldn’t. Was curious how often we saw invalid tars. But without this validation we wouldn’t know!

brainwane:
@ewdurbin I suppose we could spot-check some sample of recently uploaded tars to check what proportion of them are broken, to get an initial number to compare to?

brainwane:
Hi @gpshead - looks like once you fix the failing lint test here we'll be good to merge! Go ahead and please do that. Thanks!

gpshead:
fixed, though now there appears to be an infrastructure problem in the CI setup.

ewdurbin:
thanks for sticking that through, @gpshead

brainwane:
@gpshead Thanks for your persistence! Please let us know if you're interested in helping out with another Warehouse issue!

PR #5731, opened 2019-04-22
Allow classifier words to contain octothorp and plus.

### LINKED ISSUE #5729: `#` not allowed in classifiers
**Describe the bug**
The admin UI doesn't allow creating top-level classifiers that contain a `#`, such as #5727. The regex in https://github.com/pypa/warehouse/blob/master/warehouse/admin/static/js/controllers/parent_classifier_controller.js#L27 doesn't seem to allow them.

**Expected behavior**
Octothorps should be allowed? I can't find a spec for the trove classifiers anywhere, so I'm not sure.

Note that there seems to be *no* validation for child classifiers, and `F#` was allowed as a new child.

(Also, it would be nice if the feedback about a bad classifier value was more prominent that the very subtle shift of blue that the `Add` button undergoes, but that's a whole different issue 😄 )

**To Reproduce**

Use the admin UI to create a top-level classifier such as the one from #5727 and observe that you cannot add it from the UI.

**My Platform**

Safari 12.1, macOS 10.14.4

jamadden:
Like PL :: C++ and PL :: C#, though we don't enforce them to be at the end, they can be anywhere.

Fixes #5729


jamadden:
Thanks! (I don't have the commit bit, so someone that does will need to merge when ready.)

PR #5766, opened 2019-04-30
Remove /stats/ route.

ewdurbin:
The query has become too pathological to serve live. It likely needs to be refactored somehow.

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -111,10 +111,6 @@ def test_routes(warehouse):
         ),
         pretend.call("classifiers", "/classifiers/", domain=warehouse),
         pretend.call("search", "/search/", domain=warehouse),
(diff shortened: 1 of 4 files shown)


ewdurbin:
cc @cooperlees

ewdurbin:
Will be opening an inverse "restore..." PR and filing an issue after this merges.

cooperlees:
A once an hour or even daily run would be fine. I did suggest this originally. Happy to knock this out at the sprints.

Also, wouldn't it be nicer to just send back a static response stating we're looking to scale it rather than 404? But I guess the 404 temporarily works. This isn't an important API.

ewdurbin:
@cooperlees I agree, but I don't have the bandwidth at the moment to make a pretty response. could you PR one after this lands?

cooperlees:
See if I get time - I'm at F8 today. Tomorrow when flying to Cleveland I'll try.

PR #5767, opened 2019-04-30
Restore /stats route.

ewdurbin:
Reverts: #5766

This will need some improvement before going live. Just putting PR up for reference.


brainwane:
@cooperlees suggests we close this in favor of #5801 -- will defer to you @ewdurbin & @di.

PR #5784, opened 2019-05-04
Guard against anonymous users

dstufft:


warehouse/templates/includes/session-notifications.html
```diff
@@ -28,7 +28,7 @@
 </div>
 {% endif %}
 
-{% if request.user.two_factor_allowed and not request.user.has_two_factor %}
+{% if request.usere and request.user.two_factor_allowed and not request.user.has_two_factor %}
```
di: @dstufft Typo here, `usere`

PR #5786, opened 2019-05-04
Added date joined field to the account settings page

theb10n707:
Fixes #5786.

Screen Shot 2019-05-04 at 5 29 14 PM


warehouse/templates/manage/account.html
```diff
@@ -168,6 +168,8 @@ <h2 class="sub-title">Account details</h2>
     <div class="form-group">
       <span class="form-group__label">Username</span>
       <p class="form-group__text">{{ user.username }}</p>
+      <span class="form-group__label">Date Joined</span>
+      <p class="form-group__text"><i class="fa fa-calendar-alt" aria-hidden="true"></i> {{ user.date_joined|format_date() }}</p>
```
di: `user.date_joined` might be `None` here, we need to guard against this similar to:

https://github.com/pypa/warehouse/blob/aaea30e44e6fe0b2ac6d948d91f965d03f98d885/warehouse/templates/accounts/profile.html#L31-L33

theb10n707: Added the check in. I thought it was best to remove the icon since no other fields on the page had icons. Let me know what you think.

Screen Shot 2019-05-06 at 3 53 34 PM

brainwane:
Thanks @jreese for the idea!

PR #5787, opened 2019-05-04
Fix spacing in delete prompts

ofek:
Note: I have no idea if this works 😅 

Capture

Capture


dstufft:
This looks great! Thanks ✨

PR #5795, opened 2019-05-06
2FA (Webauthn)

woodruffw:
Begins work on Webauthn as a supported 2FA method.

TODO:

- [x]  Fill out the `utils.webauthn` helper methods
- [x]  Link the helpers up to the `Webauthn` model via `IUserService`
- [x]  Frontend integration (`navigator.credentials`)
- [x] Provisioning, removing, authentication with security keys
- [x] Allow multiple security keys
- [x] Unit tests

cc @brainwane @nlhkabu @di @ewdurbin @dstufft


warehouse/accounts/models.py
```diff
@@ -83,6 +83,13 @@ class User(SitemapMixin, db.Model):
     two_factor_allowed = Column(Boolean, nullable=False, server_default=sql.false())
     totp_secret = Column(Binary(length=20), nullable=True)
 
+    webauthn = orm.relationship(
+        "Webauthn",
+        backref="user",
+        uselist=False,
```
dstufft: Don't we want this to be a list? Ideally we're supporting any number of u2f tokens right?

woodruffw: Yep, that was just copied over and modified from the `TwoFactor` table that we didn't end up using in the TOTP PR, just so I could get something working quickly. I'll be changing it to a one-to-many relationship pretty soon.

warehouse/accounts/views.py
```diff
@@ -232,7 +228,99 @@ def two_factor(request, _form_class=TwoFactorForm):
...
+@view_config(
+    require_csrf=True,
+    require_methods=False,
+    uses_session=True,
+    request_method="POST",
+    request_param=WebAuthnAuthenticationForm.__params__,
+    route_name="accounts.webauthn-authenticate.validate",
+    renderer="json"
+)
+def webauthn_authentication_validate(request):
```
woodruffw: Pinging @di @dstufft @ewdurbin for feedback here:

The WebAuthn authentication flow requires a separate form and renderer, since we `POST` from client-side JS. However, this leads to some confusing naming above: `login` calls `two_factor`, which renders its template with the TOTP form and/or the WebAuthn button depending on what the user has.

The user then either `POST`s back to `two_factor` (for TOTP), or `POST`s to `webauthn_authentication_validate` (for WebAuthn). This is a little confusing, since `two_factor` is **only** the TOTP `POST`-back.

Some ideas:
* Split `two_factor` into two methods: one `GET` that renders the TOTP form + WebAuthn button, and one `POST` (named `totp_authentication_validate` or similar) that validates the TOTP `POST`-back. This makes it more obvious as to what is doing what, but it also means that we have to juggle the TOTP authentication form in two methods instead of one.
* Rename `two_factor` to `two_factor_and_totp_authenticate` or something similar. Not very pleasant to read.
* Document the current behavior in comments. Involves the fewest code changes, but retains the slightly confusing naming.

I'm not completely happy with any of these options, so I'd appreciate your insight.

tests/unit/packaging/test_models.py
```diff
@@ -102,6 +102,7 @@ def test_doc_url(self, pyramid_config, db_request):
             pretend.call("legacy.docs", project=project.name)
         ]
 
+    @pytest.mark.xfail(strict=False)
```
woodruffw: I'm at a loss to explain why this test is failing -- AFAICT nothing in the changeset touches ACLs, especially in such a way that would only affect packaging models.

The `user.id` comparisons are what's failing, in particular.

tests/unit/packaging/test_models.py
```diff
@@ -270,6 +271,7 @@ def test_urls(self, db_session, home_page, download_url, project_urls, expected)
         # TODO: It'd be nice to test for the actual ordering here.
         assert dict(release.urls) == dict(expected)
 
+    @pytest.mark.xfail(strict=False)
```
woodruffw: Ditto above.

woodruffw:
Hey @brainwane @di @dstufft @ewdurbin -- this should be good for a review!

We're still waiting on https://github.com/duo-labs/py_webauthn/issues/41 for multiple WebAuthn credential support; we can either push forwards with this PR and begin a new one for multiple credentials, or wait on that (and begin another task) while they cut a release. Thoughts?

di:
I think I'd prefer to merge this for support w/ multiple credentials if the Duo team's timeline is compatible with ours.

@mschwager, is there anything you can do to help us resolve https://github.com/duo-labs/py_webauthn/issues/41?

warehouse/accounts/views.py
ewdurbin: I don't have a preference, though the second option seems most reasonable to me. Even if it isn't pleasant, it is explicit!

ewdurbin:
This seems functionally excellent.

Looks like we still need to sort out:
  - Two skipped tests
  - Design concern on the TOTP vs WebAuthn forms/routes
  - UX/UI @nlhkabu will you have time to review?

brainwane:
Now that we have a release of `webauthn` https://pypi.org/project/webauthn/0.4.5/ that includes multiple credential support, I think all we are awaiting is @nlhkabu's review?

warehouse/accounts/models.py
gbdlin: Wouldn't it be better to create some generic 2Factor relation, so any new 2Factor method can use it?

woodruffw: Probably not, since different 2FA methods have different relational properties: TOTPs are 1-1 (one user, one TOTP secret + application), while WebAuthn is 1-N (1 user, N potential WebAuthn credentials). The current WebAuthn relation is 1-1, but that's only because I haven't yet updated it to take advantage of the underlying `webauthn` module's (new) multiple credential support.

Another challenge with using a single relation: we'd have to add additional type/kind checks to the 2FA method being retrieved outside of the invariants that the DB provides implicitly.

nlhkabu:
@woodruffw I've pushed some changes to this branch:

## 1. Added help text

Screenshot from 2019-06-06 07-15-53


## 2. Styled 2fa login page

In the end, I didn't go for the tabbed design, as it would require extra work for users to indicate their preferred method:

Screenshot from 2019-06-06 06-41-33

This design reduces to one column on tablet / mobile

alex:
Would it make sense to put the security key option first, for users with one configured, since it's the more secure option and we want to encourage it.

woodruffw:
Thanks a ton @nlhkabu! I'll take a look today.

Re: @alex's suggestion: yeah, I think so. I'll make that change as well.

warehouse/accounts/views.py
woodruffw: Sounds good! I'll make this change while adding support for multiple WebAuthn credentials.

nlhkabu:
## Updated order (as requested above)

Screenshot from 2019-06-07 07-30-07

## Provisioning screen

Screenshot from 2019-06-06 15-57-16

nlhkabu:
FYI @woodruffw I'd also like to review the flash messages with you, but I think we should make a separate PR for this, as this would include some of the messages from TOTP.

woodruffw:
Multiple credentials in progress:

Screen Shot 2019-06-08 at 3 07 47 PM

warehouse/manage/forms.py
```diff
@@ -104,3 +108,68 @@ def validate_totp_value(self, field):
         totp_value = field.data.encode("utf8")
         if not otp.verify_totp(self.totp_secret, totp_value):
             raise wtforms.validators.ValidationError("Invalid TOTP code. Try again?")
+
+
+class DeleteWebAuthnForm(forms.Form):
+    __params__ = ["confirm_webauthn_credential_label"]
+
+    label = wtforms.StringField(
```
woodruffw: NTS: Could limit the length of `label` to something reasonable here, like `32` or `64`.

warehouse/manage/forms.py
```diff
@@ -104,3 +108,68 @@ def validate_totp_value(self, field):
...
+        webauthn = self.user_service.get_webauthn_by_label(self.user_id, label)
+        if webauthn is None:
+            raise wtforms.validators.ValidationError("No WebAuthn key with given label")
+        self.webauthn = webauthn
+
+
+class ProvisionWebAuthnForm(WebAuthnCredentialMixin, forms.Form):
+    __params__ = ["label", "credential"]
+
+    label = wtforms.StringField(
```
woodruffw: Same as above.

nlhkabu:
@woodruffw I've now updated the UI:

## No 2fa
Screenshot from 2019-06-13 06-33-12

## 2fa by app only
Screenshot from 2019-06-13 06-34-06

## Single security key
Screenshot from 2019-06-13 06-35-12

## Multiple security keys
Screenshot from 2019-06-13 06-56-46

## Add key page
Screenshot from 2019-06-13 06-44-38

## Remove key modal
Screenshot from 2019-06-13 06-38-01

nlhkabu:
@woodruffw it seems I've broken the remove modal - when trying to remove a key, I am getting a 404 error.  I can't work out how to fix it - could you please look into this?

woodruffw:
Sure, I'll take a look in a moment.

warehouse/templates/pages/help.html
```diff
@@ -83,6 +84,7 @@ <h2><a href="#my-account">My Account</a></h2>
           <li><a href="#compromised-password">{{ compromised_password() }}</a></li>
           <li><a href="#2fa">{{ twoFA() }}</a></li>
           <li><a href="#totp">{{ totp() }}</a></li>
+          <li><a href="#utfkey">{{ utfkey() }}</a></li>
```
merwok: I did a double-take at `utf` :slightly_smiling_face: 
IIUC this is about U2F keys, not Unicode’s UTF, right?

brainwane: It is about U2F keys and not Unicode's UTF; the T stands for "two". I'm ok with this since it's just an anchor tag, but if you want to change it I figure a PR would be pretty easy!

PR #5801, opened 2019-05-07
#5769 - restore /stats route

### LINKED ISSUE #5769: 503 on /stats route
Previous implementation can be viewed in #5767

The resulting query was causing havoc in the DB, spilling over to disk and taking _a long time_ (minutes).

Suggested approaches:
  - Pre calculate results and update on daily basis, serve from some cache
  - Add a column on Project model, run a "first pass", and tally as files come in.

DavidBord:
trigger calculation on every new release ; upgrade script for existing data
Fixes #5769


brainwane:
@DavidBord I presume that this PR would supersede, and thus learns from, #5767?

ewdurbin:
I like the approach as a trigger. I'd be interested to see if we can do away with the manual `dev/calc_totals.py` step by including the following in the migration:

```sql
WITH project_totals AS (
    SELECT
        p.id as project_id,
        sum(size) as project_total
    FROM
        release_files rf
        JOIN releases r on rf.release_id = r.id
        JOIN projects p on r.project_id = p.id
    GROUP BY
        p.id
)
UPDATE projects AS p
SET total_size = project_totals.project_total
FROM project_totals
WHERE project_totals.project_id = p.id;
```

This would include the initial data load in the migration's transaction and maintain consistency/sanity.

ewdurbin:
In reviewing the trigger, I'm curious if there's a way to do incremental addition/subtraction from the total_size rather than recalculating. I need to do a quick bit of benchmarking but am worried the subquery in the trigger might be a bit costly on that hot table.

ewdurbin:
please add reversion of #5868 to this PR

brainwane:
@DavidBord is this something you could get to next week?

DavidBord:
@ewdurbin we were advised by @dstufft  to separate updating the existing content to `dev/calc_totals.py` was for it to happen in the background and not block because it might take a while. If you still think it is better to move this to the upgrade script - I'll do it gladly.

About the trigger: I preferred simplicity here over optimizing for a use case that I am not 100% sure that needs to be optimized. Since you have access to the real data and can check the impact, if you think it is substantial, I'll happily change the trigger.

Don't need to re-version just yet because it wasn't merged yet

pradyunsg:
(@ewdurbin: you seem to have linked to an as-yet not merged PR)

ewdurbin:
thank you @pradyunsg! I'll merge that now.

cooperlees:
Can you advise David which way you want him to go se we can just get this merged please. I'd really like the work he did to be counted and for the endpoint to return rather than land these documentation changes.

ewdurbin:
@cooperlees was re-reviewing now. I missed the earlier message.

ewdurbin:
> @ewdurbin we were advised by @dstufft to separate updating the existing content to `dev/calc_totals.py` was for it to happen in the background and not block because it might take a while. If you still think it is better to move this to the upgrade script - I'll do it gladly.

On the production dataset the select from the example SQL I provided runs in under 10 seconds, so I'm in favor of doing that in the migration itself.

<details>
  <summary>`EXPLAIN ANALYZE` output, click to expand</summary>

```
warehouse=> EXPLAIN ANALYZE                                                     warehouse-> SELECT
warehouse->         p.id as project_id,
warehouse->         sum(size) as project_total
warehouse->     FROM
warehouse->         release_files rf
warehouse->         JOIN releases r on rf.release_id = r.id
warehouse->         JOIN projects p on r.project_id = p.id
warehouse->     GROUP BY
warehouse->         p.id;
                                                                               QUERY PLAN                                                                                
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=670092.36..723963.64 rows=182956 width=24) (actual time=6675.364..8040.582 rows=171873 loops=1)
   Group Key: p.id
   ->  Gather Merge  (cost=670092.36..720304.52 rows=365912 width=24) (actual time=6675.344..7785.011 rows=377895 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial GroupAggregate  (cost=669092.33..677069.18 rows=182956 width=24) (actual time=6524.119..7341.238 rows=125965 loops=3)
               Group Key: p.id
               ->  Sort  (cost=669092.33..671141.43 rows=819638 width=20) (actual time=6524.111..6919.320 rows=649594 loops=3)
                     Sort Key: p.id
                     Sort Method: external merge  Disk: 20064kB
                     ->  Hash Join  (cost=287346.88..571774.42 rows=819638 width=20) (actual time=3213.585..5863.774 rows=649594 loops=3)
                           Hash Cond: (r.project_id = p.id)
                           ->  Hash Join  (cost=279787.37..551563.31 rows=819638 width=20) (actual time=2994.974..4797.402 rows=649594 loops=3)
                                 Hash Cond: (rf.release_id = r.id)
                                 ->  Parallel Seq Scan on release_files rf  (cost=0.00..250819.38 rows=819638 width=20) (actual time=0.018..766.341 rows=649594 loops=3)
                                 ->  Hash  (cost=253767.61..253767.61 rows=1345661 width=32) (actual time=2994.218..2994.218 rows=1348252 loops=3)
                                       Buckets: 65536  Batches: 32  Memory Usage: 3159kB
                                       ->  Seq Scan on releases r  (cost=0.00..253767.61 rows=1345661 width=32) (actual time=0.019..1926.834 rows=1348252 loops=3)
                           ->  Hash  (cost=4378.56..4378.56 rows=182956 width=16) (actual time=217.671..217.672 rows=183007 loops=3)
                                 Buckets: 131072  Batches: 4  Memory Usage: 3185kB
                                 ->  Seq Scan on projects p  (cost=0.00..4378.56 rows=182956 width=16) (actual time=0.012..105.694 rows=183007 loops=3)
 Planning time: 0.523 ms
 Execution time: 8100.557 ms
(23 rows)
```
</details>
 
> About the trigger: I preferred simplicity here over optimizing for a use case that I am not 100% sure that needs to be optimized. Since you have access to the real data and can check the impact, if you think it is substantial, I'll happily change the trigger.

In testing, there doesn't seem to be much of a performance impact for the full recalculation as the schema exists currently. I tested with a few projects and most are 1-2ms for the following variants:

```sql
SELECT SUM(rf.size) FROM release_files rf join releases r on rf.release_id = r.id WHERE r.project_id = _project_id;

SELECT SUM(rf.size) FROM release_files rf join releases r on rf.release_id = r.id WHERE r.project_id = _project_id;
```

The project we have with the most releases was 25-30ms for the same query.

The project we have with the most release_files was a similar 25-30ms.

So it doesn't appear to need any optkmizaiotn.

> Don't need to re-version just yet because it wasn't merged yet

Forgot to merge that up, it is merged now :)

DavidBord:
So, to conclude:
- I'll remove the warning message on this route
- I'll move the code from calc_totals.py to the upgrade script
@ewdurbin is it okay to keep my current implementation with the alembic syntax or you want me to change to your sql?

ewdurbin:
>   @ewdurbin is it okay to keep my current implementation with the alembic syntax or you want me to change to your sql?

Either is fine, but if you stick with the alembic syntax I'd like to review and performance test the SQL it generates if possible.

brainwane:
@DavidBord I think you have some tests to fix here?

ewdurbin:
Thank you @DavidBord

PR #5838, opened 2019-05-12
Implement PEP 592: Yanked Releases

dstufft:
- [x] Update the ``Release`` model to support marking a release as yanked.
- [x] Update the simple API so that it exposes the ``data-yanked`` attribute.
- [x] ~Provide an API for yanking/unyanking a release.~
- [x] Tests

### DIFF
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -227,7 +227,7 @@ Release
                         "python_version": "2.7",
                         "size": 3795,
                         "upload_time_iso_8601": "2015-06-14T14:38:05.869374Z",
-                        "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl"
+                        "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl",
                         "yanked": false
                     },
                     {
@@ -244,7 +244,7 @@ Release
                         "python_version": "source",
                         "size": 3148,
                         "upload_time_iso_8601": "2015-06-14T14:37:56.394783Z",
-                        "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz"
+                        "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz",
                         "yanked": false
                     }
                 ]
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -72,7 +72,8 @@ Project
                 "requires_dist": null,
                 "requires_python": null,
                 "summary": "A sample Python project",
-                "version": "1.2.0"
+                "version": "1.2.0",
+                "yanked": false
             },
             "last_serial": 1591652,
             "releases": {
@@ -92,7 +93,8 @@ Project
                         "python_version": "2.7",
                         "size": 3795,
                         "upload_time_iso_8601": "2015-06-14T14:38:05.093750Z",
-                        "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl"
+                        "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl",
+                        "yanked": false
                     },
                     {
                         "comment_text": "",
@@ -108,7 +110,8 @@ Project
                         "python_version": "source",
                         "size": 3148,
                         "upload_time_iso_8601": "2015-06-14T14:37:56Z",
-                        "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz"
+                        "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz",
+                        "yanked": false
                     }
                 ]
             },
@@ -127,7 +130,8 @@ Project
                     "python_version": "2.7",
                     "size": 3795,
                     "upload_time_iso_8601": "2015-06-14T14:38:05.234526",
-                    "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl"
+                    "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl",
+                    "yanked": false
                 },
                 {
                     "comment_text": "",
@@ -143,7 +147,8 @@ Project
                     "python_version": "source",
                     "size": 3148,
                     "upload_time_iso_8601": "2015-06-14T14:37:56.000001Z",
-                    "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz"
+                    "url": "https://files.pythonhosted.org/packages/eb/45/79be82bdeafcecb9dca474cad4003e32ef8e4a0dec6abbd4145ccb02abe1/sampleproject-1.2.0.tar.gz",
+                    "yanked": false
                 }
             ]
         }
@@ -201,7 +206,8 @@ Release
                 "requires_dist": null,
                 "requires_python": null,
                 "summary": "",
-                "version": "1.0"
+                "version": "1.0",
+                "yanked": false
             },
             "last_serial": 1591652,
             "releases": {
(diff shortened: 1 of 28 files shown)


nlhkabu:
Note:

brainwane:
This feature would be helpful in [improving pip's test automation](https://wiki.python.org/psf/PackagingWG/2020-01-23-pip#Automation_snags_and_permissions) so it would be great if you could prioritize finishing this WIP.

dstufft:
I am unlikely to find time in the interim to finish this, though I think we could deploy this without an API and just have it be a UI only feature for now.

brainwane:
@ewdurbin @di @dstufft hey - as Ernest and @pradyunsg  and I [talked about in a meeting about the pip resolver work](https://wiki.python.org/psf/PackagingWG/2020-01-23-pip), getting this implemented might help us smooth the second-order effects of the resolver rollout, so if you could help get this finished and merged, we'd appreciate that!

di:
@nlhkabu, could you take a look at https://github.com/pypa/warehouse/pull/5838/commits/1d7824d969762d09a5222847b8fb7db50c2b0ac8 if you have a second? Without this change, the callout blocks were appearing above the modals.

ewdurbin:
The value of `yanked` be added to the JSON documents for projects/releases, rather than hiding them.

warehouse/packaging/models.py
```diff
@@ -142,8 +142,9 @@ def __getitem__(self, version):
             return (
                 session.query(Release)
                 .filter(
-                    (Release.project == self)
-                    & (Release.canonical_version == canonical_version)
+                    Release.project == self,
+                    Release.yanked.is_(False),
```
dstufft, line 146: Do we want to do this? This will make it so yanked versions can't be viewed online at all right?

dstufft (commented):
I think my biggest thing is I'm not sure that we want yanked to function as hiding releases from the UI (but maybe we do, I'm not sure?).

My gut instinct tells me we probably want this behavior:

- On the unversioned URL for a project, we won't pick a yanked version, and would instead act as if there were no versions.
  - One could argue that it's better to just display a yanked version as a last ditch effort to display *something*, I'm not sure though?
- On a versioned URL, yanked releases will show up normally, except there is some marker on the page to indict this release has been yanked (likely some sort of red warning banner?).
- On the list of all versions, the yanked releases still show up, but are marked in some fashion to show that they have been yanked (greyed out to demphasize? red? idk).

Fundamentally, I don't think we want to reimplement the hidden release option from legacy PyPI, this is largely about making it so that people can prevent a version from being installed unless absolutely needed, it isn't about hiding it like it never existed and I think that being able to introspect it in the UI still is still useful.

di:
I see where you're coming from @dstufft. My thought is that if we want people to use yanks instead of deleting releases, we need to make yanks behave just like deleting releases, except you can still pin to them. People delete releases for a lot of reasons: bad `long_description`, typos, etc. Ideally in those cases, maintainers can still "remove" a release from the public eye, but anyone already using it can continue to pin to it.

Another issue this helps avoid is requiring average users to understand what a yanked release even is. If a user is looking at project page for `some-project` and they see that the latest release, `1.0.0` is marked as "yanked" with a big red box, but it's still there, they're probably not going to understand what's happening when they `pip install some-project` and get version `0.9.9` instead of `1.0.0`. If the project page just shows the un-yanked releases though, nothing changes here, and they never need to know that `1.0.0` even exists.

di:
Hmm, I see your point, and I think you're probably right. I hadn't considered the difference in behaviors between clients that do/don't support yanked releases.

di:
@dstufft I think I addressed all your feedback here. Yanked releases will show up everywhere except:

* `Project.latest_version` will not return a yanked release as the latest
* The project detail page will not show a yanked release as the latest
* `/pypi/<project_name>/json` will not use a yanked release as the latest
* Search reindexing will not include yanked releases

warehouse/templates/legacy/api/simple/detail.html
```diff
@@ -19,7 +19,7 @@
   <body>
     <h1>Links for {{ project.name }}</h1>
     {% for file in files -%}
-    <a href="{{ request.route_url('packaging.file', path=file.path) }}#sha256={{ file.sha256_digest }}"{% if file.release.requires_python %} data-requires-python="{{ file.release.requires_python }}"{% endif %}>{{ file.filename }}</a><br/>
+    <a href="{{ request.route_url('packaging.file', path=file.path) }}#sha256={{ file.sha256_digest }}"{% if file.release.requires_python %} data-requires-python="{{ file.release.requires_python }}"{% endif %}{% if file.release.yanked %}data-yanked{%if file.release.yanked %}="file.release.yanked"{% endif %}{% endif %}>{{ file.filename }}</a><br/>
```
dstufft, line 22: This conditional looks weird, why is there two identical conditionals nested inside each other? Why are we setting the ``data-yanked`` attribute equal to the string ``file.release.yanked``?

warehouse/templates/pages/help.html
```diff
@@ -209,6 +211,13 @@ <h3 id="trove-classifier">{{ trove_classifier() }}</h3>
           Classifiers are used to categorize projects on PyPI. See <a href="{{ href }}">the classifiers page</a> for more information, as well as a list of valid classifiers.
           {% endtrans %}
         </p>
+
+        <h3 id="yanked">{{ yanked() }}</h3>
+        <p>
+          {% trans trimmed href='https://www.python.org/dev/peps/pep-0592/', title=gettext('External link') %}
+          A yanked release is a release that is always ignored by an installer, unless it is the only release that matches a version specifier (using either <code>==</code> or <code>===</code>. See <a href="{{ href }}" title="{{ title }}" target="_blank" rel="noopener"><abbr title="Python enhancement proposal">PEP</abbr> 541</a> for more information.
```
dstufft, line 218: The PEP number is wrong here in the copy.

warehouse/templates/legacy/api/simple/detail.html
di: Looks like that was a holdover from https://github.com/pypa/warehouse/pull/5838/commits/f976d0abf1156990a5c34db8b8130fcde1796f02, where we might have had a comment or something? I'm not sure what your original intention was, but I updated this just to show the attribute when the release is yanked.

warehouse/templates/email/unyanked-project-release/body.html
```diff
@@ -0,0 +1,35 @@
...
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+{% block content %}
+<p>
+  <ul class="removed-project-release">
+    <li>The {{ project }} release {{ release }} released on {{ release_date }} has been yanked.</li>
```
pradyunsg, line 19:
```suggestion
    <li>The {{ project }} release {{ release }} released on {{ release_date }} has been un-yanked.</li>
```

Because the email is for "unyanked-project-release".

warehouse/templates/email/unyanked-project-release/body.html
```diff
@@ -0,0 +1,35 @@
...
+    <li>
+      <strong>Yanked by:</strong> {{ submitter }} with a role: {{ submitter_role }}.
```
pradyunsg, line 21:
```suggestion
      <strong>Un-yanked by:</strong> {{ submitter }} with a role: {{ submitter_role }}.
```

warehouse/templates/email/unyanked-project-release/body.txt
```diff
@@ -0,0 +1,27 @@
...
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/body.txt" %}
+
+{% block content %}
+  The {{ project }} release {{ release }} released on {{ release_date }} has been yanked.
+
+  Yanked by: {{ submitter }} with a role: {{ submitter_role }}.
```
pradyunsg, line 20:
```suggestion
  Un-yanked by: {{ submitter }} with a role: {{ submitter_role }}.
```

warehouse/templates/email/unyanked-project-release/body.txt
```diff
@@ -0,0 +1,27 @@
...
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/body.txt" %}
+
+{% block content %}
+  The {{ project }} release {{ release }} released on {{ release_date }} has been yanked.
```
pradyunsg, line 18:
```suggestion
  The {{ project }} release {{ release }} released on {{ release_date }} has been un-yanked.
```

pradyunsg (commented):
I feel like it'd be nice to require users to put in something as "reason for yanking" (picking from a bunch of fixed choices, for example) since installers can present that information to the user (pip prints a warning containing the message), and having it be compulsory from the start would help avoid a potential transition step if we add it later.

Then again, it's much easier to do a smoother transition on web application than a command line application, so it's perfectly fine if y'all think it's OK to not start with that... :)

warehouse/templates/manage/release.html
```diff
@@ -121,6 +121,30 @@ <h3>{% trans %}No files found{% endtrans %}</h3>
...
+    {% else %}
+      {% trans %}
+        Yanking will mark this release to be ignored when installing in most common scenarios.
+      {% endtrans %}
+    {% endif %}
+    {% trans project_name=project.name, version=release.version %}
+      This release will still be installable for users pinning to this exact version, e.g. when using <code>{{ project_name }}=={{ version }}</code>.
+    {% endtrans %}
+    {% trans href="https://www.python.org/dev/peps/pep-0592/" %}
+      For more information, see <a href="{{ href }}">PEP 592</a>.
```
dstufft, line 142: This doesn't need to block this PR, but I've wondered if our propensity for linking to PEPs is the best user experience we can provide here, or if we should figure out a real documentation pipeline for Warehouse (not really counting what exists now since it's primarily developer facing documentation, not user facing) and start moving to that.

dstufft (commented):
I can't approve this because it's my own PR, but we'll just pretend I did.

warehouse/templates/manage/release.html
di: I think I agree, we are sort of hitting the limits of what we can document inline and in /help.

brainwane: I filed #7817 - thanks for the idea, @dstufft

ewdurbin (approved):

docs/api-reference/json.rst
```diff
@@ -222,6 +228,7 @@ Release
                         "size": 3795,
                         "upload_time_iso_8601": "2015-06-14T14:38:05.869374Z",
                         "url": "https://files.pythonhosted.org/packages/30/52/547eb3719d0e872bdd6fe3ab60cef92596f95262e925e1943f68f840df88/sampleproject-1.2.0-py2.py3-none-any.whl"
+                        "yanked": false
```
merwok, line 231: Comma missing on the line before (json fun)
Same for line 248

pradyunsg:
YAY! Thanks for all of your work on this @nlhkabu @di @dstufft @ewdurbin! ^>^

PR #5872, opened 2019-05-17
error `make initdb` if xz isn't installed

ewdurbin:
This threw me for a loop getting warehouse setup on my new laptop! `make initdb` was silently failing to load the dev db assets.


brainwane:
@di I feel ok going ahead and merging this; how do you feel?

di:
I had talked to @ewdurbin elsewhere about moving this step into a container to sidestep this problem entirely. I think he was attempting to do that but probably got sidetracked by more important stuff.

brainwane:
If this is a small positive step then how about we merge it and then work on the bigger improvement later?

ewdurbin:
Sorry I didn't close the loop! Moving it into the docker container was non-trivial. Let's merge this up. 

Might make sense to look through all remaining local dev tasks and file an issue with a punch list to move them all into containers. This will likely make for a good contribution from someone with more docker-compose experience!

Issue #5661, opened 2019-04-03
2FA/API tokens: staging/testing rollout


brainwane:
**What's the problem this feature will solve?**
To finish #996 (see #5567), we need to test MFA with real users on real packages; asking them to spin up dev environments is too hard and won't help multi-maintainer projects reason well about what MFA policies they want to set up.

**Describe the solution you'd like**
My tentative suggestion is:

* We first roll out MFA on test.pypi.org and publicize it to project maintainers for a designated 2-week testing period, and we allot extra time for dealing with support requests during that time
* We then roll it out to pypi.org and announce it on the PyPI announcement email list, PSF blog, etc.

**Additional context**
* If things go awry during initial testing, is there any chance we will need to wipe tokens from users' accounts?
* Are there particular categories of user we need to make sure we get in the beta test? Ideas that come to mind for me: Windows, mobile, users of particular TOTP implementations and U2F implementations, people on very slow connections, people who habitually block a lot of cookies/JS, maintainers who maintain 20+ projects, people with very old and weird PyPI accounts (e.g., we do not have verified email for them, their passwords do not adhere to current policy), users with upwards of 10 MFA methods they want to add to their accounts, multi-maintainer projects, multi-owner projects, organizations where users share an auth token within a group.
* Is test.pypi.org the right place for this? Should we spin up some other instance for this particular kind of testing, since maintainers do use Test PyPI for real uploads (of packages that need testing)?

Cc @ewdurbin .

di:
If we want to roll out to TestPyPI first, we'll probably want to feature-flag this feature, which would be a bit of work.

It might be preferable to do a true "beta" on PyPI instead, with a limited group of users. This could just be a boolean field on the `User` model that enables the feature, which we set in the admin UI. And we could just skip this check and enable it for all TestPyPI users, and eventually remove it once we've rolled everything out.

This could allow us to do a timeline like:

* 1 week period on TestPyPI for everyone
* 2 week beta on PyPI for select users
* Open to all users

> If things go awry during initial testing, is there any chance we will need to wipe tokens from users' accounts?

Possibly? I think we should be explicit in our "beta" announcement that this might happen, so users aren't surprised.

> Are there particular categories of user we need to make sure we get in the beta test? 

I think all the groups you listed would be ideal.

> people who habitually block a lot of cookies/JS

[...]

PR #5889, opened 2019-05-20
warehouse: Require a verified email for 2FA modifications

### LINKED ISSUE #5866: Require verified email before allowing user to enable 2FA
Sometimes people enable two-factor auth and then run into some kind of problem like losing access to their provisioned TOTP-generating app. We want to make it easier for PyPI admins to check the legitimacy of users' requests for account recovery in such circumstances.

Prohibit users from enabling any 2FA until the user has verified at least one email address on their account.

**Additional context**
(followup to #5661, filed on @di's behalf)

woodruffw:
Prevents users from modifying their 2FA methods without a primary, verified email. Already enabled 2FA methods will continue to work, even if the user's email lapses into an unverified state.

Inverts the `two_factor_allowed` column into `two_factor_prohibited`, with a new property on `User` testing `two_factor_prohibited` plus whether the user has both a primary email *and* whether that email is verified.

Closes #5866.


woodruffw:
> Its not clear to me under what condition we would ever flag an account as being prohibited from using 2FA, but I suppose it's nice to have the option.

Yeah, that state/option is vestigial from the TOTP rollout; I can take it out in this PR or a future one, if desired 🙂 

> My only concern is that we send blind 403s without explanation. I suppose we could flash a notification to help people understand why they're getting a 403?

Yeah, I can add this. I'm also about to push some changes that ensure that we only reject 2FA provisioning requests and not authentication (to prevent the de-verified email scenario discussed out-of-band).

woodruffw:
Alright, warning flashes added.

brainwane:
> @brainwane when this merges TOTP goes live, do we need to coordinate launch?

To be clear: I'd like active approval by @di before we can merge this PR and thus launch TOTP as a feature for all PyPI.org and test.pypi.org users, but I'm ok with going forward without @dstufft's +1. If Dustin, Ernest, or Donald objects and thinks Donald's +1 is a necessary prerequisite, please override me. :-)
  
Re: launch coordination (also, followup to #5661): I will write up a blog post for the PSF blog to be published within the next few days, and I think we should link to it on https://status.python.org/ . Therefore, my request is that we merge after Wednesday May 29th 2pm Eastern Time, to give me time to draft it.

brainwane:
@di is reviewing this now!

@ewdurbin https://pad.sfconservancy.org/p/2fa-pypi-launch-post is what I think could be posted at https://pyfound.blogspot.com/ , https://blog.python.org/ , and status.python.org. Take a look, and please feel free to edit. Also, I need to step away from my laptop for the afternoon and evening, so please feel free to go ahead and publish that post once we have merged this PR.

pradyunsg:
@brainwane It'd be best to also link to the pypi-announce mailing list at the end, when mentioning it: https://mail.python.org/mailman3/lists/pypi-announce.python.org/

warehouse/manage/views.py
```diff
@@ -325,6 +325,12 @@ def default_response(self):
 
     @view_config(route_name="manage.account.totp-provision.image", request_method="GET")
     def generate_totp_qr(self):
+        if not self.request.user.two_factor_provisioning_allowed:
+            self.request.session.flash(
+                "Modifying 2FA requires a verified email.", queue="error"
```
di: This message (and other instances of it) aren't quite right: As long as we have `two_factor_prohibited`, this might fail even if the user has a valid email.

woodruffw: Yep, I'll just remove `two_factor_prohibited` from the PR entirely. It was there to allow us to feature-gate on test.pypi.org, but since we're doing the full rollout it doesn't serve a purpose anymore 😄

PR #5938, opened 2019-05-31
Require login after a successful password reset

woodruffw:
Instead of treating a successful reset as a login and bypassing the user's second factors, redirect to the login page.

cc @di @dstufft @ewdurbin


di:
I think it make more sense here to keep the behavior the same for users without 2FA, and just redirect 2FA users to the 2FA page after reset instead of the login page.

Admittedly, this might take a little more effort, but would be nicer and more consistent UX while still fixing this issue.

woodruffw:
This is ultimately your call, but my 0.02c: I think that requiring login after a password reset makes sense in all scenarios, and I believe that it's (semi) standard practice. The reasoning: resetting your password is a management action based on a single-use (and permission/time scoped) token, not an act of authentication. Combining the two makes things convenient for users and doesn't harm security *too* much in the non-2FA case, but I'd consider it surprising behavior from an authentication perspective.

In terms of consistency: I agree that it's good not to break current expectations. OTOH, we get a different kind of consistency by always requiring re-login on password reset.

PR #5946, opened 2019-05-31
Add dependency management pattern

di:


brainwane:
I should mention - I also read this for content and it makes sense to me. :)

PR #5965, opened 2019-06-05
Fix "nuke user" view

di:
Use joinedload to `avoid raise_on_sql` exception. Fixes https://sentry.io/organizations/python-software-foundation/issues/851747167/


di:
(This was missed in #5039)

PR #6001, opened 2019-06-12
Autofocus the main search bar

Ivoz:


brainwane:
But: see #5102, #854 and #3440 for discussion on whether we _should_ do this.

nlhkabu:
Hi @Ivoz thanks for this PR.

Thanks to the recent OTF funding for accessibility, I've had time to look into the issue of *whether* or not this is a good idea. I've opened a ticket with my findings here: #6088

TLDR: Based on the available accessibility recommendations, and the data that shows how users are using the page, we should not focus the search box.

Thanks again for your PR.

PR #6013, opened 2019-06-13
reduce email validation strictness

robindboer:
Hi 👋, I took a shot at https://github.com/pypa/warehouse/issues/3733

The email validation is replaced by a simpler regex validation as suggested in the issue.

Please let me know if I can improve anything as this is my first contribution to the project.


di:
Thanks @robindboer. Looks good, can you add a test which shows this is fixed? The original issue was with an email address like `foo@n--tree.net`.

robindboer:
Hi @di , Thank you for your feedback. I have updated the pull request to include a unit test that shows the original issue is fixed.

PR #6035, opened 2019-06-17
Add browser support info to WebAuthn FAQ item

### LINKED ISSUE #6034: Document Safari WebAuthn issue
Our WebAuthn support [requires browser support for `PublicKeyCredential`](https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/utils/webauthn.js#L41), which means [it doesn't work with any version of IE, Safari, or Opera](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential).

If we're ok with requiring that our users use Chrome, Firefox, or Edge to use WebAuthn, then we should document that in the FAQ and give a clearer error message than ["Your browser doesn't support WebAuthn."](https://usercontent.irccloud-cdn.com/file/grfXo1ef/Screen%20Shot%202019-06-17%20at%2012.38.11%20PM.png) (It supports _part_ of WebAuthn!) Or, if we want to support those browsers, then this is a bug.

Thanks @cooperlees for the report.

brainwane:
Closes #6034.

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -62,6 +62,21 @@ docReady(() => {
   }
 });
 
+// Hide webauthn option for unsupported browsers, or show errors
+docReady(() => {
+  if (!window.PublicKeyCredential) {
+    let webauthn_button = document.getElementById("webauthn-button");
+    if (webauthn_button) {
+      webauthn_button.style.display = "none"
+    }
+
+    let webauthn_error = document.getElementById("webauthn-browser-support");
+    if (webauthn_error) {
+      webauthn_error.style.display = "block"
+    }
+  }
+});
+
 // Human-readable timestamps for project histories
 docReady(() => {
   timeAgo();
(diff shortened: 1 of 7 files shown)


nlhkabu:
## Updates

### 1. Show errors on login and provisioning page
Screenshot from 2019-06-18 06-24-52
Screenshot from 2019-06-18 06-24-03

### 2. Hide provisioning button 
Screenshot from 2019-06-18 06-47-50

Note: if TOTP is enabled, the webauthn button will also not show at the bottom of the table

### 3. Update help text
Screenshot from 2019-06-18 06-25-12

@brainwane I made a small update to the help text, so it starts with "To use a security key with WebAuthn, you'll need..."

@woodruffw - please feel free to update the JS if there is a more efficient way of doing this.

warehouse/static/js/warehouse/index.js
```diff
@@ -62,6 +62,21 @@ docReady(() => {
   }
 });
 
+// Hide webauthn option for unsupported browsers, or show errors
```
ewdurbin: Is fully hiding it the right move? I think graying it out would be more reasonable for discoverability of the feature.

nlhkabu:
Updated as per feedback:

Screenshot from 2019-06-18 15-46-14

PR #6054, opened 2019-06-19
Add Collaborators by invitation

### LINKED ISSUE #5790: Send invitations when adding owner/maintainer roles
Currently as a PyPI package maintainer, it's possible to give any other user an owner/maintainer role for your package without their consent.

We should change the way role assignment currently works, from just allowing owners to assign other users roles on a given project, to allowing owners to _request_ that other users accept roles for a given project.

The way we should do this is by sending an email to the new owner/maintainer with a validation link, and allowing them to accept/reject the request before the new role is added.

**Additional context**
Originally described in #4729.

ewdurbin:
resolves #5790, carrying on from work by @DavidBord


di:
I think this should probably be good to go. @ewdurbin can you review?

di:
Hmm, perhaps not, looks like there's a number of places where we're not filtering on the invitation status and an invited user would have the role before accepting the invitation.

This makes me think we should break invitations out entirely separate from the `Roles` table and only actually add the `Role` once the invitation has been accepted.

nlhkabu:
Note:

PR #6072, opened 2019-06-22
Fix typo

minho42:


tests/unit/forklift/test_legacy.py
```diff
@@ -2858,13 +2858,13 @@ def test_upload_succeeds_creates_squats(self, pyramid_config, db_request):
         user = UserFactory.create()
         EmailFactory.create(user=user)
 
-        filename = "{}-{}.tar.gz".format("exmaple", "1.0")
+        filename = "{}-{}.tar.gz".format("example", "1.0")
```
pradyunsg: This typo is intentional -- the test checks for typo-squatting.

IMO it's be a good idea to add a comment above this line explaining that the typo is intentional. :)

PR #6084, opened 2019-06-24
Macaroon-based API keys

woodruffw:
Based on https://github.com/dstufft/warehouse/pull/2.

TODO:
- [x] Package-scoped tokens
- [x] Unit tests


warehouse/macaroons/services.py
```diff
@@ -0,0 +1,81 @@
...
+                .filter(Macaroon.id == macaroon_id)
+                .one()
+            )
+        except NoResultFound:
+            return
+
+        return dm
+
+    def find_userid(self, macaroon):
+        m = pymacaroons.Macaroon.deserialize(macaroon)
```
moshez: This will fail if the `macaroon` input is `None`.

Note that the extraction function "Returns either a `None` if no macaroon could be found".

warehouse/macaroons/services.py
```diff
@@ -0,0 +1,81 @@
...
+
+        try:
+            dm = (
+                self.db.query(Macaroon)
+                .options(joinedload("user"))
+                .filter(Macaroon.id == macaroon_id)
+                .one()
+            )
+        except NoResultFound:
+            return
```
webknjaz: I don't really like using this technique of returning a special value in Python. It makes the function return different data types in different scenarios.
It's basically what we're forced to do in C or Go: requiring the user of this function to not forget to check the return value.
I think it's better to communicate errors via exceptions.

warehouse/macaroons/services.py
```diff
@@ -0,0 +1,81 @@
...
+    def verify(self, macaroon):
+        m = pymacaroons.Macaroon.deserialize(macaroon)
+        dm = self._get_record_from_db(m)
+
+        if dm is None:
+            raise InvalidMacaroon
+
+        verifier = pymacaroons.Verifier()
+
+        if not verifier.verify(m, dm.key):
```
webknjaz: I think it would read better without the negation:
```suggestion
        if verifier.verify(m, dm.key):
        	return True

        raise InvalidMacaroon('The key is not valid')
```

warehouse/macaroons/services.py
```diff
@@ -0,0 +1,81 @@
...
+        dm = self._get_record_from_db(m)
+
+        return None if dm is None else dm.user.id
+
+    def verify(self, macaroon):
+        m = pymacaroons.Macaroon.deserialize(macaroon)
+        dm = self._get_record_from_db(m)
+
+        if dm is None:
+            raise InvalidMacaroon
```
webknjaz: It's beneficial for debugging to have a reason string attached to the exception objects
```suggestion
            raise InvalidMacaroon('There is no such key in the database')
```

warehouse/macaroons/auth_policy.py
```diff
@@ -0,0 +1,120 @@
...
+    authorization = request.headers.get("Authorization")
+    if not authorization:
+        return None
+
+    try:
+        auth_method, auth = authorization.split(" ", 1)
+    except ValueError:
+        return None
+
+    if auth_method.lower() != "macaroon":
```
dstufft: Just noting that this will have to change if we want to make it capable of being used with existing clients without modifications to pull out the password from the client. Not sure what we should expect for username in that case, we could either expect the username to match the macaroon or ignore the username? Not sure if there's a better option.

Another option is to support *both* this header and the basic auth shim, and try to guide people over to using this header instead. Not sure if that is worthwhile or not.

warehouse/macaroons/auth_policy.py
```diff
@@ -0,0 +1,120 @@
...
+        macaroon = extract_http_macaroon(request)
+
+        # This logic will only happen on requests that are being authenticated with
+        # Macaroons. Any other request will just fall back to the standard Authorization
+        # policy.
+        if macaroon is not None:
+            macaroon_service = request.find_service(IMacaroonService, context=None)
+
+            try:
+                macaroon_service.verify(macaroon)
```
dstufft: I know we talked about how to scope these to just uploads right now, and I wanted to just mention that one way of doing that would be to just one basic caveat (for right now), and basically do this:

```macaroon_service.verify(macaroon, context, principals, permission)```

By passing those three values into the verify function, we could then do something like this to verify:


```python
def verify_caveat(predicate, *, permission):
    # This function has to deal with the entire dictionary that could be possible
    # which means it has to validate the keys we've gotten.
    try:
        data = json.loads(predicate)
    except Exception:  # TODO: Better exception
        return False

    # Check for any unknown predicate keys, we always fail if there are any.
    if data.keys() - {"permissions"}:  # TODO: Might be better to use jsonschema or something eventually.
        return False

    # Go through the caveats and ensure that they match.
    if "permissions" in data and permission not in data["permissions"]:
        # If we have a caveat for permissions, and the permission we're checking for is NOT in that, we fail
        return False

    # Finally, we return True
    return True

v = Verifier()
v.satisfy_general(verify_caveat)
v.verify(...)
```

My suggestion here is basically if we can quickly get some agreement on the general *shape* we want caveats to take (json dict, msgpack dict, whatever) *and* that we want to bake permissions into those caveats, we can quickly  whip up a quick macaaroon based mechanism to handling it.

I think we should still bake version numbers into the ID, just to allow us to completely change the caveat language if we want to, but I think doing this means when we expand, we're already setup to handle the next type of permission we want to add. Upload tokens aren't special, they're just ones that got created with only one permission scope. 

Thoughts?

dstufft: Also in the above code, we could obviously use a class or something instead of a partial, that was just a quick and dirty example of how easy I think it would be to push the verification of "the currently generated tokens are only good for upload" off to the caveat language.

This *would* mean that as implemented an upload token is still not limited to what *projects* it can support and it's scoped the same as the user account that created it in terms of blast radius. That should be pretty easily to do similarly, add a "projects" array to the caveats dictionary and check that the current context is one of those project, basically the same as permissions are done. Everything else should be done by the AuthZ/N framework.

woodruffw:
Thanks @moshez and @webknjaz! I'll be addressing your feedback, but could you hold off on providing more until the "[WIP]" is removed from the title?

warehouse/macaroons/auth_policy.py
woodruffw: Hmm. I'm not sure about worthwhileness either, but my intuition is that this should *eventually* become its own header entirely, rather than piggybacking on `Authentication: basic` (which it'll have to do for this rollout).

Here's one idea: we could support both `Authorization: macaroon <macaroon>` and `Authorization: basic <token>`, where `<token>` is `base64("macaroon:" . <macaroon>)`. That results in a pretty chunky header since we end up encoding the macaroon twice, but it preserves compatibility and gives us a "username" that clients could enter without making any changes. Thoughts?

warehouse/macaroons/auth_policy.py
woodruffw: > I think we should still bake version numbers into the ID, just to allow us to completely change the caveat language if we want to, but I think doing this means when we expand, we're already setup to handle the next type of permission we want to add. Upload tokens aren't special, they're just ones that got created with only one permission scope.
> 
> Thoughts?

I like this idea! I'll start on it.

warehouse/macaroons/auth_policy.py
```diff
@@ -0,0 +1,150 @@
...
+    if not authorization:
+        return None
+
+    try:
+        auth_method, auth = authorization.split(" ", 1)
+    except ValueError:
+        return None
+
+    if auth_method.lower() == "basic":
+        return _extract_basic_macaroon(auth)
```
webknjaz: It looks like this function operates bytes internally but you pass unicode...

warehouse/macaroons/auth_policy.py
```diff
@@ -0,0 +1,150 @@
...
+    Returns ``None`` if the header doesn't contain a structurally
+    valid macaroon, or the candidate (not yet verified) macaroon.
+    """
+    try:
+        authorization = base64.b64decode(auth)
+    except ValueError:
+        return None
+
+    try:
+        auth_method, auth = authorization.split(b":", 1)
```
webknjaz: It's probably cleaner to use `str.partition()`

di:
@webknjaz, did you see @woodruffw's request above?

> I'll be addressing your feedback, but could you hold off on providing more until the "[WIP]" is removed from the title?

webknjaz:
Yeah, I just got a notification and was tempted to promptly leave small comments... I'll try to not do that while it's wip :)

warehouse/macaroons/auth_policy.py
woodruffw: `_extract_basic_macaroon` can take either bytes or a string, since it shoves its input into `b64decode` and the latter is happy to take either. The raw macaroon in both cases ends up being passed to `Macaroon.deserialize` (which can also take either bytes or a string), but for the sake of consistency I'll make sure that `extract_http_macaroon` only ever returns a string.

warehouse/macaroons/services.py
woodruffw: This is consistent with the account service methods, which also return `None` for most operations. The exception is verification methods, which (generally) raise to indicate a failure. I don't have a strong opinion here, but I think it's fine in the context: these methods get called within a larger set of checks, whose failure as a whole is usually reported via either an exception or some dedicated type like `Denied`.

warehouse/macaroons/services.py
woodruffw: Yep, this is a good idea. What I intend to do here is actually have it trickle in from `verifier`, since the internal logic there will be able to produce more useful errors.

warehouse/macaroons/auth_policy.py
woodruffw: Ended up going with the nested `basic` idea, with the "username" as `@token` to distinguish it from a real username.

warehouse/macaroons/services.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        m = pymacaroons.Macaroon.deserialize(raw_macaroon)
+        dm = self.find_macaroon(m.identifier.decode())
+
+        if dm is None:
+            raise InvalidMacaroon
+
+        verifier = Verifier(m, context, principals, permission)
+        if verifier.verify(dm.key):
+            dm.last_used = datetime.datetime.now()
```
merwok: This will be saved because there is a call to commit the DB session somewhere in the request lifecycle?

woodruffw: Yes, that's correct. We use the same style for some changes to WebAuthn models as well (IIRC).

dstufft: For completeness sake: We're using pyramid_tm to tie the database transaction to the lifecycle of the HTTP request.

woodruffw:
This should be ready for initial testing, for the adventurous.

Some quick notes:
* The UI is functional, but needs to be updated to conform to PyPI's design
* Our current auth-policy is drop-in compatible with Twine and distutils. When using a token, your "username" will be `@token` and your "password" will be the token itself

warehouse/templates/manage/token.html
```diff
@@ -0,0 +1,96 @@
...
+  {% if user.macaroons|length > 0 %}
+  <section id="active-tokens">
+    <h2>Active API tokens</h2>
+    <table class="table table--light">
+      <thead>
+        <th>Description</th>
+        <th>Identifier</th>
+        <th>Created</th>
+        <th colspan="2">Last used</th>
+      </thead>
```
brainwane: table-and-api-key-for-github-review

Is it possible for one of the headings to be **Scope**? If not, could we add some text (next to the "copy it now") also reminding the user to make a note of the scope?

warehouse/templates/manage/token.html
```diff
@@ -0,0 +1,96 @@
...
+        Scope:
+        <select name="token_scope">
+          <option value="scope:user">User (all projects)</option>
+        {% for project in project_names %}
+          <option value="scope:project:{{ project }}">{{ project }}</option>
+        {% endfor %}
+        </select>
+      </div>
+      <div>
+        <input value="Create token" class="button button--primary" type="submit">
```
brainwane: Right now, if I leave the token description empty, or put in a too-long (like, 40 characters) description, and click "Create token", I get taken back to the API token management page, but I don't get an error message. I figure we should probably give the user an error message.

warehouse/templates/manage/token.html
```diff
@@ -0,0 +1,96 @@
...
+            {% set action=request.route_path('manage.account.token.delete') %}
+            {% set extra_fields %}
+              <input type="text" name="macaroon_id" value="{{ macaroon.id }}" hidden>
+            {% endset %}
+            {{ confirm_modal(title=title, confirm_name="Username", confirm_string=user.username, confirm_button_label=confirm_button_label, slug="delete-token", extra_fields=extra_fields, action=action, warning=False, confirm_string_in_title=False) }}
+          </td>
+        </tr>
+        {% endfor %}
+      </tbody>
+    </table>
```
brainwane: wordwrap-issue-api-keys

Two items:
* should we allow the user to make a second key with an identical description to an existing key? I think we shouldn't, to help the user avoid confusion.
* word wrap issue

warehouse/templates/manage/token.html
```diff
@@ -0,0 +1,96 @@
...
+{% block title %}{{ title }}{% endblock %}
+
+{% block main %}
+  <h1 class="page-title">{{ title }}</h1>
+
+  {% if serialized_macaroon %}
+  <section id="provisioned-key">
+    <div class="callout-block callout-block--warning">
+      <p>
+        This token will only appear once. Copy it now.
```
brainwane: :+1:

warehouse/templates/manage/token.html
brainwane: I'm also curious -- maybe I've missed something -- about what the "Identifier" (hash?) is for. It's clearly a unique identifier for the API key, but is the user ever going to need to refer to it, either in her own systems or when communicating with PyPI tech support or something like that?

warehouse/templates/manage/account.html
```diff
@@ -364,6 +364,12 @@ <h2 class="sub-title">Two factor authentication (2FA)</h2>
 
   <hr>
 
+  <section id="api-keys">
+    <a href="{{ request.route_path('manage.account.token') }}">Manage API tokens</a>
```
brainwane: Should this be prefaced with `{% if email.verified %}`?

warehouse/templates/manage/token.html
woodruffw: Yep, I was giving some thought to how best to represent the token's scope -- it can currently be a single user or a set of packages. That's relatively easy to present, but future changes to the caveat language might require more complex presentations.

We can punt that down the road though and just handle the user + package set cases for now, though.

warehouse/templates/manage/token.html
woodruffw: Yep, I need to expose the form errors. Will do that now, thanks!

warehouse/templates/manage/token.html
woodruffw: > * should we allow the user to make a second key with an identical description to an existing key? I think we shouldn't, to help the user avoid confusion.

I agree; this was part of the original schema that I adopted but it makes sense to enforce uniqueness here to me.



> * word wrap issue

Yep; I might need @nlhkabu's help fixing this one 😅

warehouse/templates/manage/account.html
woodruffw: Hmm, possibly. Users are currently able to add tokens even without a verified email, but the underlying policy *should* boot them out when they try to use them.

If we do add this, we'll probably need to do it in the forms themselves (to prevent users from navigating directly to the view).

woodruffw:
> This is not technically accurate -- this user account **is** allowed to upload to that project, but the **token** isn't. And maybe we don't want to leak the username in this error message. If we see that it's a token being used, can/should we change the error message in this case to something like "This token isn't allowed to upload to project 'name'"?

Yep, this needs to be changed here:

https://github.com/pypa/warehouse/blob/0d9c726153394cf6aca178c9015c467707fbaa2e/warehouse/forklift/legacy.py#L944-L954

It might be difficult to parametrize the error message in this context, so I could replace "user" with "credential owned by user" -- thoughts?

Re: leaking the username: this check happens after an authorization check, so we already have a valid credential (just not one for this package). As such it isn't an infoleak *per se* (and might be useful to users doing complicated deployments, since their error messages will tell them which underlying user identity failed).

warehouse/templates/manage/token.html
woodruffw: Oh, and re: the identifier: macaroons require a unique identifier, but users definitely don't need to see it. It was there for my debugging purposes; I'll remove it 🙂

dstufft: One possible benefit of exposing it is to make it easier for a user to figure out what row in their tokens corresponds with a token they happen to have possession of.

Of course, there are other ways of doing this, like embedding the description into the macaroon identifier as well.

woodruffw: Yep, that's a good point. We don't currently embed the description into the macaroon at all (and, at 100 characters, doing so might increase the length quite a bit), so I'll re-add the identifier.

woodruffw: Here's what the scope looks like with just the dictionary dumped into it:

Screen Shot 2019-07-18 at 1 03 57 PM

brainwane: > One possible benefit of exposing it is to make it easier for a user to figure out what row in their tokens corresponds with a token they happen to have possession of.

So the imagined user here is storing this structured data (description, identifier, and token) locally, in a config file? Just checking!

woodruffw: > So the imagined user here is storing this structured data (description, identifier, and token) locally, in a config file?

The token has the identifier embedded in it, so the user would probably only store the token. The description is only stored in the DB for presentation in the web interface -- it isn't tightly coupled to the token itself. We could change that and embed it into the token directly, but at the cost of longer tokens and (IMO) relatively little benefit.

warehouse/templates/manage/account.html
brainwane: @woodruffw Yes - when I tried to use the token with a no-verified-email user account, I did get stopped, with a useful error message! :+1: 

@nlhkabu I'll defer to you here since this feels like a UX issue. My assumption is that, if a token's unusable until/unless the user has a verified email address, then we should stop a no-verified-email user from creating a token in the first place, thus avoiding an unpleasant surprise down the line.

woodruffw: Cool -- good to know the underlying policy is working 😄 

I'm on board with restricting token creation to only users with verified emails -- it definitely is an unpleasant surprise to be able to create unusable tokens.

tests/unit/macaroons/test_models.py
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from warehouse.macaroons import models
+
+
+def test_generate_key():
+    key = models._generate_key()
```
ewjoachim: Is it worth checking that generating keys twice lead to 2 different keys ?

warehouse/macaroons/auth_policy.py
```diff
@@ -0,0 +1,147 @@
...
+    if auth_method != "@token":
+        return None
+
+    return auth
+
+
+def _extract_http_macaroon(request):
+    """
+    A helper function for the extraction of HTTP Macaroon from a given request.
+    Returns either a None if no macaroon could be found, or the byte string
```
ewjoachim: not 100% sure, but this looks like a string, not byte string. We're splitting with `" "` and not `b" "`, and matching with `"basic"` and `"token"`, not their bytes counterparts.

ewjoachim: not 100% sure, but this looks like a string, not byte string. We're splitting with `" "` and not `b" "`, and matching with "basic" and "token", not their bytes counterparts.

tests/unit/macaroons/test_models.py
ewjoachim: Is it worth checking that generating keys twice lead to 2 different keys ?

woodruffw: `_generate_key` is a straight-line call to `os.urandom`, so I don't think this test would add much (at most it would tell us that `os.urandom` on whatever machine is running the tests isn't really random, but that isn't likely to be a production machine).

warehouse/macaroons/auth_policy.py
woodruffw: Yep, that comment is out of date. Will fix now, thanks!

nlhkabu:
Hi @woodruffw - I have pushed changes as discussed.

I'll need your help with the following:

- [ ] It should be possible to delete an API token from the token management page, _in addition_ to on the manage account page (using the 'remove' button)

- [ ] API token delete confirmation message should include the API token name.  E.g. `API token deleted` should become `Deleted API token "my token"`

- [ ] API token delete confirmation message should appear on the same page the deletion was initated from (i.e. either the manage account page, or the manage token page)

Thanks :)

nlhkabu:
> It might be difficult to parametrize the error message in this context, so I could replace "user" with "credential owned by user" -- thoughts?

Instead of `The user 'brainwane' isn't allowed to upload to project 'Forms990-analysis'` could we do something like `The API token you are using is not valid for project 'Forms990-analysis'` ?

woodruffw:
> As per @brainwane's comment above, it would also be nice to hide the "create token" UI if the user does not have a verified email address. For 2fa, we have a permission called `user.two_factor_provisioning_allowed`. Could we please do the same thing for API tokens?

Yep -- I'm going to rename `two_factor_provisioning_allowed` to `has_primary_verified_email` and re-use it.

> Instead of `The user 'brainwane' isn't allowed to upload to project 'Forms990-analysis'` could we do something like `The API token you are using is not valid for project 'Forms990-analysis'` ?

This might be a little difficult with the current permission system: from within the view we can only see that `has_permission` returned `False`, not *why* it did so. I updated the language to be a bit more generic, but we'd probably have to consider adding a mechanism for side-channeling the actual reason.

nlhkabu:
> I updated the language to be a bit more generic, but we'd probably have to consider adding a mechanism for side-channeling the actual reason.

I don't think this is MVP. Could we open a separate ticket for this?

woodruffw:
> I don't think this is MVP. Could we open a separate ticket for this?

Yep! I'll do that now.

woodruffw:
Weird CI failures here:

For `lint`:

```
/home/travis/virtualenv/python3.7.3/bin/html_lint.py --printfilename --disable=optional_tag,names,protocol,extra_whitespace,concerns_separation `find ./warehouse/templates -path ./warehouse/templates/legacy -prune -o -name '*.html' -print`
  File "/home/travis/virtualenv/python3.7.3/bin/html_lint.py", line 66, in <module>
    sys.exit(html_linter.main(options))
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/html_linter.py", line 1094, in main
    clean_html = template_remover.clean(io.open(filename).read())
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/template_remover.py", line 258, in clean
    return _TemplateRemover(html_content, pattern).get_clean_content()
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/template_remover.py", line 199, in get_clean_content
    assert False, 'Echo tags should be in just one line.'
AssertionError: Echo tags should be in just one line.
```

and `dependencies`:

```
# install ipython if enabled
/home/travis/virtualenv/python3.7.3/bin/pip-compile --no-annotate --no-header --upgrade --allow-unsafe -o /tmp/tmp.JBxyonQDnV/deploy.txt requirements/deploy.in > /dev/null
  File "/home/travis/virtualenv/python3.7.3/bin/pip-compile", line 10, in <module>
    sys.exit(cli())
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/piptools/scripts/compile.py", line 256, in cli
    repository = PyPIRepository(pip_options, session, build_isolation)
  File "/home/travis/virtualenv/python3.7.3/lib/python3.7/site-packages/piptools/repositories/pypi.py", line 95, in __init__
    self.finder = PackageFinder(**finder_kwargs)
TypeError: __init__() got an unexpected keyword argument 'find_links'
Makefile:130: recipe for target 'deps' failed
```

Any ideas @dstufft @ewdurbin?

warehouse/templates/manage/token.html
```diff
@@ -0,0 +1,95 @@
...
+
+      {{ confirm_modal(
+        title=title,
+        confirm_name="Username",
+        confirm_string=user.username,
+        confirm_button_label=confirm_button_label,
+        slug=slug,
+        extra_fields=extra_fields,
+        custom_warning_text=token_warning_text,
+        confirm_string_in_title=False) }}
```
di: @woodruffw This is what's causing the `lint` failure, this should all be a single line (FYI, `{{` and `}}` are called 'echo' tags)

di:
@woodruffw The `dependencies` failure is due to https://github.com/jazzband/pip-tools/issues/856, we'll either have to wait until they release a new version, or pin back the version of `pip` we're using in that job.

warehouse/templates/pages/help.html
```diff
@@ -269,6 +271,29 @@ <h3 id="utfdevices">{{ utfdevices() }}</h3>
...
+
+        <p>API tokens provide an alternative way (instead of username and password) to authenticate when uploading packages to PyPI.</p>
+        <p>You can create a token for an entire PyPI account, in which case, the token will work for all projects associated with that account. Alternatively, you can limit a token's scope to a specific project.</p>
+        <p><strong>We strongly recommend you authenticate with an API token where possible.</strong></p>
+
+        <p>To make an API token:</p>
+
+        <ul>
+          <li><a href="#verified-email">Verify your email address</a> (check your <a href="{{ request.route_path('manage.account') }}">account settings</a>)</li>
+          <li>In your <a href="{{ request.route_path('manage.account') }}">account settings</a>), go to the API tokens section and select "Add API token"</li>
```
brainwane: Dangling paren. I made an amend to this commit https://github.com/pypa/warehouse/commit/d9b42b0450bf88e02d6d01ba581cfd7e08714306 to fix it, but I'm hesitant about force-pushing to your branch. Please feel free to rewrite history here or just add a new commit to fix the paren, your call. Sorry.

warehouse/templates/manage/token.html
woodruffw: Thanks!

warehouse/templates/pages/help.html
woodruffw: No sweat, just pushed the change.

ewdurbin:
preemptively tested what we'll need to change to disable password uploads for users with some form of two-factor enabled, and it seems as straightforward as.

```diff
diff --git a/warehouse/accounts/__init__.py b/warehouse/accounts/__init__.py
index 8c4c6961..e618755b 100644
--- a/warehouse/accounts/__init__.py
+++ b/warehouse/accounts/__init__.py
@@ -32,7 +32,7 @@ from warehouse.accounts.services import (
     database_login_factory,
 )
 from warehouse.email import send_password_compromised_email_hibp
-from warehouse.errors import BasicAuthBreachedPassword
+from warehouse.errors import BasicAuthBreachedPassword, BasicAuthTwoFactor
 from warehouse.macaroons.auth_policy import (
     MacaroonAuthenticationPolicy,
     MacaroonAuthorizationPolicy,
@@ -70,6 +70,11 @@ def _basic_auth_login(username, password, request):
             raise _format_exc_status(
                 BasicAuthBreachedPassword(), breach_service.failure_message_plain
             )
+        elif user.has_two_factor:
+            raise _format_exc_status(
+                BasicAuthTwoFactor(),
+                f"{user.username} has enabled two-factor auth, provision an API Token!",
+            )
         elif login_service.check_password(
             user.id, password, tags=["method:auth", "auth_method:basic"]
         ):
diff --git a/warehouse/errors.py b/warehouse/errors.py
index 73b75209..347367cc 100644
--- a/warehouse/errors.py
+++ b/warehouse/errors.py
@@ -15,3 +15,7 @@ from pyramid.httpexceptions import HTTPUnauthorized
 
 class BasicAuthBreachedPassword(HTTPUnauthorized):
     pass
+
+
+class BasicAuthTwoFactor(HTTPUnauthorized):
+    pass
```

PR #6118, opened 2019-06-28
Refactor search logic out of views.py

yeraydiazdiaz:
Split from #5311


yeraydiazdiaz:
The motivation was to avoid having the actual fields in the querystring and in the template, as this exposes internal details and couples the template and the searching logic.

I didn't really think about ordering to be honest. Maybe we should officially introduce it as a second dropdown? Though I'm not sure if they would apply to the score-based modes if you find "desc created" useful I'm sure more people would.

nlhkabu:
Note:

yeraydiazdiaz:
I've updated this MR taking @di's comments into account and is ready for review.

I verified the search results are the same both in master and this branch.

yeraydiazdiaz:
I forgot to mention re @nlhkabu's comment above, there are no template changes anymore so I believe no translation changes are required.

warehouse/templates/search/results.html
```diff
@@ -178,9 +178,9 @@ <h2 class="sr-only">{% trans %}Search results{% endtrans %}</h2>
             <input id="search" type="hidden" name="q" value="{{ term }}">
             <label for="order">{% trans %}Order by{% endtrans %} &nbsp;</label>
             <select class="-js-form-submit-trigger" id="order" name="o">
-              {{ search_option(gettext("Relevance"), "") }}
-              {{ search_option(gettext("Date last updated"), "-created") }}
-              {{ search_option(gettext("Trending"), "-zscore") }}
+              {{ search_option("Relevance", "") }}
```
ewdurbin: will removing the `gettext` "untranslated" these strings?

yeraydiazdiaz: Woops, yeah, that wasn't intentional, good catch 👍

PR #6128, opened 2019-07-01
Add upload success metrics

di:


warehouse/forklift/legacy.py
```diff
@@ -1315,6 +1316,9 @@ def file_upload(request):
                 },
             )
 
+    metrics = request.find_service(IMetricsService)
```
ewdurbin: load this at open, emit a `warehouse.upload.attempt` metric. this will allow us to look for failure count before more fine grained metrics are implemented.

PR #6142, opened 2019-07-03
Add password breach controller checking password on register

### LINKED ISSUE #4470: Add javascript/frontend validation of breached passwords
In https://github.com/pypa/warehouse/pull/4468 we've started blocking new passwords (via registration, reset, or password change) that have previously been in a data breach of a compromised website. However, that validation currently only occurs server side, it would be nice to also have it occur client side.

This can happen in two ways:

* Have the client JS hit the HIBP API directly, effectively re-implementing the logic that exists in the Python code.
* Have the client JS hit a private API endpoint on Warehouse, that calls into the logic that exists there.
  * If we do this, what should this API look like? Do we want to just send the password over the wire (since that's going to happen on submit anyways?), do we want to prehash it with sha1 and send that over the wire? Or do we want to reimplement the HIBP api in Warehouse? (this would only prevent leaking our users PII via additional http requests to an external service).

Of course, the other question is whether we want to do this or not. It's hard to do it without either reimplementing logic, or sending the password to Warehouse as the user is typing it, neither of which are ideal. Maybe having this validation happen only server side is ok?

yeraydiazdiaz:
Fixes #4470 

This PR implements the first of the options described by @dstufft: reimplementing the logic in the backend in JavaScript. Here's a capture of the new behaviour

hibp_frontend

Note the submit button is *not* affected by this new controller since it's managed by the `password_match_controller` which presents a technical difficulty and I figured the warning was discouraging enough. However if it's something we'd like to add I'm happy to figure out a way to combine the results of both controllers.


brainwane:
@yeraydiazdiaz there's a merge conflict - could you take a look?

yeraydiazdiaz:
Done, it'd be good if @nlhkabu could double check though.

nlhkabu:
Hi @yeraydiazdiaz - thanks for this PR.  I'm thinking that it could be useful to link to the help documentation from the error message.  We have an FAQ question [here](https://pypi.org/help/#compromised-password)

If we do this, I wonder, if it would be enough just to say "This password has been compromised and cannot be used"?  Is there a specific reason you have put "appears in a breach" in the error text?

yeraydiazdiaz:
Hey @nlhkabu, the text in the error message is the same as in the [service](https://github.com/pypa/warehouse/blob/master/warehouse/accounts/services.py#L503), I thought it best to keep it consistent. I'm happy to update both to include the link to the FAQ.

yeraydiazdiaz:
Would also you like to reword "appears in a breach"? The error message appears verbatim on the FAQ so we'll need to change it there too.

nlhkabu:
I think that depends @yeraydiazdiaz - do you think the meaning of the term "breach" is clear for the majority of users?  

Maybe "security breach" would be clearer? 

Note this faq: https://haveibeenpwned.com/FAQs

yeraydiazdiaz:
Yeah, "security breach" sounds more specific, I'll change it to that. 👍

nlhkabu:
Thank you!

nlhkabu:
Note:

yeraydiazdiaz:
I've rebased and updated this branch including frontend tests. It is now ready for review.

warehouse/accounts/services.py
```diff
@@ -516,9 +516,7 @@ def __eq__(self, other):
 @implementer(IPasswordBreachedService)
 class HaveIBeenPwnedPasswordBreachedService:
 
-    _failure_message_preamble = (
-        "This password appears in a breach or has been compromised and cannot be used."
-    )
+    _failure_message_preamble = "This password appears in a security breach or has been compromised and cannot be used."
```
yeraydiazdiaz: I tried adding `localize` to this string but when running `make build-mos` I ended up with a lot of conflicts with the locale files. Is there something else I need to do?

warehouse/static/js/warehouse/controllers/password_breach_controller.js
```diff
@@ -0,0 +1,110 @@
...
+      let text = await response.text();
+      this.parseResponse(text, hex);
+    }
+  }
+
+  async digestMessage(message) {
+    // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
+    const encoder = new TextEncoder();
+    const data = encoder.encode(message);
+    return window.crypto.subtle.digest("SHA-1", data);
```
di: According to [the spec](https://github.com/w3c/webcrypto/issues/28) `window.crypto.subtle` is `undefined` in insecure  contexts, and as a result I get a `TypeError` on this line when trying to run it locally over HTTP.

I'm curious a) how you were able to work around this when testing and b) how you suggest we handle this for local development.

warehouse/accounts/services.py
di: You don't need to run `make build-mos`, that's what the translation service bot will do. You just need to run `make compile-pot` to add the new translatable string.

warehouse/static/js/warehouse/controllers/password_breach_controller.js
yeraydiazdiaz: That's news to me 😅

The capture shows the controller working locally and I just tested it successfully using `http://localhost/account/register/` in Chrome, Firefox and Safari on my Mac.

How exactly are you getting the `TypeError`?

di: Huh, weird. I use a custom host mapping to 127.0.0.1, and `window.crypto.subtle` is `undefined` when I use that host, but not when I use `localhost`/`127.0.0.1`, even though they're all HTTP.

Guess I'll just have to ignore this! 😅

PR #6155, opened 2019-07-06
Make form errors work better for screen readers

nlhkabu:
To merge after #https://github.com/pypa/warehouse/pull/6150

To test:

- [x] Login page
- [x] Request password reset page
- [x] Complete password reset page
- [x] Registration page
- [x] My account page
- [x] TOTP login page
- [x] TOTP provisioning page

Note: webauthn is out of scope for this PR, we will need to make a separate PR for this

### DIFF
--- a/warehouse/templates/accounts/two-factor.html
+++ b/warehouse/templates/accounts/two-factor.html
@@ -13,7 +13,10 @@
 -#}
 {% extends "base.html" %}
 
-{% block title %}Two-factor authentication{% endblock %}
+{% block title %}
+  {% if totp_form and totp_form.errors %}Error processing form &ndash;{% endif %}
+  Two-factor authentication
+{% endblock %}
 
 {% block content %}
 {% if testPyPI %}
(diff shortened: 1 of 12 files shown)


nlhkabu:
TODO:
- password matching JS broken

nlhkabu:
How to test:

- Go to each form page using a screen reader
- Force form errors on each form
- Check that the errors are read out by the screen reader
- On page reload, the title of the page should also change to indicate there are errors on the page
- When errors are shown, they should be associated with the field using 'aria-describedby'

warehouse/templates/accounts/two-factor.html
```diff
@@ -28,7 +31,7 @@ <h1 class="page-title">Two-factor authentication</h1>
 
     <div class="twofa-login">
       {% if has_webauthn %}
-      <div class="twofa-login__method {% if totp_form %}twofa-login__method--padded{% endif %}">
+      <div class="twofa-login__method {% if form %}twofa-login__method--padded{% endif %}">
```
yeraydiazdiaz: Is this change intentional? Looking at [the view code](https://github.com/pypa/warehouse/blob/master/warehouse/accounts/views.py#L207) seems `topt_form` was correct.

nlhkabu: thanks @yeraydiazdiaz - this was left over from some refactoring that it decided not to press ahead with.  I've removed it now.

PR #6193, opened 2019-07-15
warehouse: Allow anonymous WebAuthn attestation

### LINKED ISSUE #6183: 2FA: Enrolling a TouchID sensor as a webauthn security key fails (Chrome, Mac)
**Describe the bug**
I'm trying to enroll a TouchID sensor as a webauthn device. PyPI and Chrome do let me select the sensor, and I do get prompted for a touch, but then PyPI throws an error: "Registration rejected. Error: Self attestation is not permitted.."

**Expected behavior**
I expect to be able to enroll a TouchID sensor. 

**To Reproduce**
- PyPI --> Account Settings
- Click "Add 2FA With Security Key"
- Type a key name, click "Provision Key"
- Chrome prompts to choose between a USB security key and a built-in sensor. Choose "Built-in sensor"
- MacOS prompts to hit the TouchID sensor. Do so.
- Chrome prompts, "Allow this site to see your security key?" Click "Allow"
- PyPI displays an error: "Registration rejected. Error: Self attestation is not permitted.."

**My Platform**
- MacOS 10.14.5
- MacBook Air (2018 edition, with TouchID)
- Chrome "75.0.3770.100 (Official Build) (64-bit)"

woodruffw:
This allows users to use TouchID and other methods
that don't use separate public keys.

cc @di @jacobian

Fixes #6183

### DIFF
--- a/warehouse/utils/webauthn.py
+++ b/warehouse/utils/webauthn.py
@@ -111,7 +111,11 @@ def verify_registration_response(response, challenge, *, rp_id, origin):
     # first for the entire clientData payload, and then again
     # for the individual challenge.
     response = pywebauthn.WebAuthnRegistrationResponse(
-        rp_id, origin, response, _webauthn_b64encode(challenge.encode()).decode()
+        rp_id,
+        origin,
+        response,
+        _webauthn_b64encode(challenge.encode()).decode(),
+        self_attestation_permitted=True,
     )
     try:
         return response.verify()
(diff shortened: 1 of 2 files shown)


warehouse/utils/webauthn.py
```diff
@@ -111,7 +111,11 @@ def verify_registration_response(response, challenge, *, rp_id, origin):
     # first for the entire clientData payload, and then again
     # for the individual challenge.
     response = pywebauthn.WebAuthnRegistrationResponse(
-        rp_id, origin, response, _webauthn_b64encode(challenge.encode()).decode()
+        rp_id,
+        origin,
+        response,
+        _webauthn_b64encode(challenge.encode()).decode(),
```
pradyunsg: nit: it might be nice to move this expression out, into a dedicated variable.

PR #6200, opened 2019-07-17
Update 2fa instruction text

### LINKED ISSUE #6171: Improve 2fa instructions
### On the webauthn provisioning page

- "Webauthn" is not commonly understood. Instead, refer to webauthn as "phyiscal security key (e.g. USB)".  Include examples of key brands (e.g. "Yubikey") in instructional text
- Avoid using 'provisioning'.  Instead use 'set up'
- "Name key" is not a clear enough input label. Add an instruction above, or rename the label to "Enter key name to begin"
- Remove 'How it works' session and replace with set up guidelines
- Add 'troubleshooting' callout block - something like "Not working? You may be using an old USB security key that's not FIDO compatible (the standard supported by PyPI)."
- Add/fix link to help page
- Consider adding image of USB to page to give visual instruction

### On the TOTP provisioning page

- "TOTP application" should be referred to as "Authentication application"
- Avoid using 'provisioning'.  Instead use 'set up'
- Remove 'How it works' session and replace with set up guidelines
- Add list of application names to page (e.g. "Authy", "Google Auth" etc.)
- Consider adding image of phone/app to page to give visual instruction
- Add/fix link to help page

nlhkabu:
Closes #6171

Note that I've renamed "physical security key" to "security device (e.g. USB key)".  I think this is clearer, but happy to discuss @woodruffw 

Screenshot from 2019-07-17 07-01-56

---------------------------

Screenshot from 2019-07-17 07-01-26

### DIFF
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -1017,7 +1017,7 @@ class TestProvisionTOTP:
         ]
         assert request.session.flash.calls == [
             pretend.call(
-                "Authentication application deleted from PyPI. "
+                "Authentication application removed from PyPI. "
                 "Remember to remove PyPI from your application.",
                 queue="success",
             )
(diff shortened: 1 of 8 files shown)


woodruffw:
> Note that I've renamed "physical security key" to "security device (e.g. USB key)". I think this is clearer, but happy to discuss

I'm happy with that!

PR #6207, opened 2019-07-18
WebAuthn: Better errors

### LINKED ISSUE #6157: Improve webauthn errors
Problems:

- We currently have two pieces of JS that control the display of webauthn errors - some in `index.js`, some in `webauthn.js`
- The errors are not announced to the screenreader (via `role=alert`)
- The errors are not associated with the webauthn label field (on the provisioning page) - we should use `aria-describedby` for this
- The user is able to put text into the label field on the provisioning page - it should be disabled

woodruffw:
Also fixes deprovisioning and removes a merge conflict.

Closes #6157.

### DIFF
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -293,7 +293,7 @@ class TestDeleteWebAuthnForm:
         )
 
         assert not form.validate()
-        assert form.label.errors.pop() == "Specify a device label"
+        assert form.label.errors.pop() == "Specify a device name"
 
     def test_validate_label_not_in_use(self):
         user_service = pretend.stub(
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -293,7 +293,7 @@ class TestDeleteWebAuthnForm:
         )
 
         assert not form.validate()
-        assert form.label.errors.pop() == "Specify a label"
+        assert form.label.errors.pop() == "Specify a device label"
 
     def test_validate_label_not_in_use(self):
         user_service = pretend.stub(
--- a/warehouse/manage/forms.py
+++ b/warehouse/manage/forms.py
@@ -111,11 +111,11 @@ class ProvisionTOTPForm(TOTPValueMixin, forms.Form):
 
 
 class DeleteWebAuthnForm(forms.Form):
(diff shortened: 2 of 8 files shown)


nlhkabu:
Thanks for catching these errors @woodruffw :)  A few items:

## Consolidating JS code

We currently have two places where we check for webauthn browser compatibility:

- https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/utils/webauthn.js#L41
- https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/index.js#L67

Would it make sense to move all of this into one place?

## Consistent error messages

Can we update [this line](https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/utils/webauthn.js#L42) to have the following message:

`<a href="https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential#Browser_compatibility" title="External link" target="_blank" rel="noopener">Upgrade your browser</a>`

This would be consistent with the other webauthn errors in the template.

## Disabling the input

Can we please also update the JS so that the input is disabled on the provisioning page, e.g.

### Before

Screenshot from 2019-07-23 07-33-11

### After

Screenshot from 2019-07-23 07-39-14

woodruffw:
> Would it make sense to move all of this into one place?

Yup! Done.



> Can we update [this line](https://github.com/pypa/warehouse/blob/master/warehouse/static/js/warehouse/utils/webauthn.js#L42) to have the following message:
> 
> `<a href="https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential#Browser_compatibility" title="External link" target="_blank" rel="noopener">Upgrade your browser</a>`
> 
> This would be consistent with the other webauthn errors in the template.

This should be fixed now; I've made it so that `populateWebAuthnErrorList` doesn't clobber the existing error, so that should always appear.


> Can we please also update the JS so that the input is disabled on the provisioning page, e.g.

Done!

nlhkabu:
Thanks @woodruffw.

I've improved the display of the errors:

Screenshot from 2019-07-23 20-35-23
Screenshot from 2019-07-23 20-34-54

I think we should be ready to merge this one now.

woodruffw:
Yep, everything's good on my end. Thanks for fixing that!

PR #6245, opened 2019-07-25
Improve 2fa UI on mobile

nlhkabu:
## Ensure long buttons wrap

Screenshot from 2019-07-24 07-18-34

## Wrap table

Screenshot from 2019-07-25 06-37-48


woodruffw:
Looks like this'll need to be rebased now that #6084 is merged, but otherwise LGTM!

Issue #6256, opened 2019-07-25
Always display API token management panel


ewdurbin:
Noticed after deploy of #6084, that the API token management panel is not at all visible if any single email address is not verified.

I suggest that we should:
  - Always display the panel, but grey out or disable it
  - Allow for management of API Tokens when at least one verified primary email is present.

brainwane:
I agree that we should allow for API token management if there's _at least one_ verified primary email address. @woodruffw please look into this, and also please check whether this problem also affects our 2FA methods as a whole.

I'll defer to @nlhkabu on the accessibility and UX implications of the greyed out/disabled view.

nlhkabu:
@ewdurbin I am not able to reproduce this in dev:

User with one verified email:

![Screenshot from 2019-07-26 07-16-42](https://user-images.githubusercontent.com/3323703/61930370-6b6f4700-af75-11e9-933d-3b52f7bd9742.png)

Can see and set up API tokens (and 2fa):

![Screenshot from 2019-07-26 07-17-32](https://user-images.githubusercontent.com/3323703/61930399-7d50ea00-af75-11e9-8444-1417d3b23f41.png)

nlhkabu:
Opened #6269 to address comment:

> Always display the panel, but grey out or disable it

brainwane:
I also can't reproduce the issue where a single unverified email address stops the user from being able to create an API token. @ewdurbin can you still reproduce this on production?

brainwane:
I'm going to call this addressed. Thanks!

PR #6269, opened 2019-07-26
Ask users to verify email to set up API keys

nlhkabu:
Refs #6256 
> Always display the panel, but grey out or disable it

- Updates 2fa and API section to specifically instruct users to verify their primary email (this is more explicit than just greying it out)
- Links to email section on account admin page, so users can resend verification email.

Screenshot from 2019-07-26 07-13-20

### DIFF
--- a/warehouse/templates/manage/account.html
+++ b/warehouse/templates/manage/account.html
@@ -71,7 +71,7 @@
       </div>
     </td>
     <td class="table__action">
-      {% if not email.verified or not email.primary%}
+      {% if not email.verified or not email.primary %}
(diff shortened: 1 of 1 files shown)


nlhkabu:
@ewdurbin @woodruffw  what would be even nicer here:

- Change message to `To add two factor authentication to your account, first verify your primary email.  [Resend verification email](link that sends verification email).`

WDYT?

ewdurbin:
This is good as is, though your suggested improvement could be reasonable! I'm not sure if it is crucial that we put the link right in the message though, seems best to leave the behavior in a single place.

PR #6271, opened 2019-07-26
Show warning when user scopes API key to account

nlhkabu:
Closes https://github.com/pypa/warehouse/issues/6266

Screenshot from 2019-07-26 07-49-19

@woodruffw would you be able to write a small snippet of JS that toggles the visibility of the warning, depending on whether or not "Entire account (all projects)" is selected above?

### DIFF
--- a/warehouse/static/sass/blocks/_callout-block.scss
+++ b/warehouse/static/sass/blocks/_callout-block.scss
@@ -24,6 +24,7 @@
 
   Modifiers:
     - danger: Makes border red
+    - warning: Makes border brown
     - success: Makes border green
     - bottom-margin: Adds extra margin below the callout
 */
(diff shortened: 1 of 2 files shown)


nlhkabu:
@woodruffw what do you think of the copy here?

"have the same power as your account password" isn't _technically_ true, as it will only work for package uploads...

nlhkabu:
In addition to the warning, we could force users to explicitly select their scope, e.g.

Screenshot from 2019-07-26 08-22-58

woodruffw:
> "have the same power as your account password" isn't _technically_ true, as it will only work for package uploads...

Yeah, that's true. How about "will be able to modify all of your projects"?

> In addition to the warning, we could force users to explicitly select their scope, e.g.

Great idea! I totally forgot this was possible 😅

woodruffw:
I'll take a look at adding the JS snippet.

woodruffw:
Closing in favor of #6274.

PR #6272, opened 2019-07-26
Add fullstop

hugovk:
Now consistent with:

"Two factor authentication adds an additional layer of security to your account. Learn more."

# Before

image


nlhkabu:
thank you @hugovk

PR #6274, opened 2019-07-26
Force user to explicitly select token scope, show warning if user scopes to account

### LINKED ISSUE #6266: API token default scope: user or project?
Followup to #994: 

@brettcannon [asks](https://discuss.python.org/t/pypi-security-work-multifactor-auth-progress-help-needed/1042/33):

> > @ewdurbin said: By default, newly created tokens will have “user” scope, meaning that they’ll behave exactly like your password.

> Are there plans to change this default so that using such a strong token is not the default so that people have to opt into it? (I’m no security expert so this is more inquisitive.)

As far as I know there are no such plans but I'd like @woodruffw and @nlhkabu to weigh in.

woodruffw:
Closes #6266 
Builds on #6271.

cc @nlhkabu


nlhkabu:
## Updated errors (to be consistent)
Screenshot from 2019-07-27 10-43-56

## Updated warning text
Only shown when "whole account" is selected:
Screenshot from 2019-07-27 10-44-24

nlhkabu:
ready for review @di or @ewdurbin:

Functional updates:
- Force users to explicitly select scope (and show form error if no scope selected)
- Show "warning" message if "entire account" scope is selected

warehouse/templates/manage/token.html
```diff
@@ -70,15 +70,20 @@ <h2>Token for "{{ macaroon.description }}"</h2>
         <p id="description-help-text" class="form-group__help-text">What is this token for?</p>
       </div>
       <div class="form-group">
-        <label for="scope" class="form-group__label">Scope</label>
-        <select name="token_scope" id="scope" class="form-group__input">
+        <label for="token_scope" class="form-group__label">Scope</label>
+        <select name="token_scope" id="token_scope" class="form-group__input" aria-describedby="token_scope-errors">
+          <option hidden selected value="scope:unspecified">Select scope...</option>
```
di:
```suggestion
          <option disabled selected value="scope:unspecified">Select scope...</option>
```

I think it's better here if this is visible, but disabled. As is, when the user opens the select box, "Entire account" already appears to be selected, but it's not _actually_ selected until the user clicks on the option.

Before:
Screen Shot 2019-07-29 at 10 43 40 AM

After:
Screen Shot 2019-07-29 at 10 43 24 AM

PR #6288, opened 2019-07-29
Improve API UI

### LINKED ISSUE #6257: Add API Token form confusion
noticed that the form elements for adding an API Token still display after creating a token

Screen Shot 2019-07-25 at 11 04 22 AM

This is a bit confusing, and I'm not sure if intentional

nlhkabu:
Closes #6257 

Screenshot from 2019-07-29 06-48-50

- Makes permissions explicit in token creation form
- Lists permission and scope above API token display
- Adds "Add another token" heading, when displaying API token


Carreau:
Thanks ! That Looks Great !

PR #6301, opened 2019-07-29
warehouse: Allow maintainers to craft project tokens

### LINKED ISSUE #6262: [Project-scoped API tokens] aren't available to maintainers
**Describe the bug**

When I use a "bot" account with "Maintainer" level access to projects, there are no projects to select from in the form for the token creation.

**Expected behavior**

Since this "bot" can upload dists using user/password auth, it should also have similar privileges set when using tokens.

**To Reproduce**

Go to https://pypi.org/manage/account/token and try selecting a project where you have only "Maintainer"-level access, not "Owner".

**My Platform**

N/A

**Additional context**

N/A

woodruffw:
Closes #6262.

cc @di / @dstufft / @brainwane


woodruffw:
> Perhaps it should link to `project/[project name]/` if the user is a maintainer. Or if that's inelegant, it should link to that page for all users.

Done: I've linked to `/project/{name}` unconditionally, since retrieving the user's role for each project would be a little inelegant in this context (and I think consistency is best, since we're filling out a single row within a table).

warehouse/manage/views.py
```diff
@@ -559,7 +559,24 @@ def __init__(self, request):
...
+        projects = (
+            self.request.db.query(Project)
+            .join(
+                projects_owned_or_maintained,
+                Project.id == projects_owned_or_maintained.c.id,
+            )
+            .order_by(Project.name)
+            .all()
+        )
         return [project.name for project in projects]
```
di: Looks like this all could probably just become:

```python
sorted(project.name for project in request.user.projects)
```

unless I'm missing something.

woodruffw:
@di should be good for another pass!

PR #6305, opened 2019-07-29
warehouse: Catch WebAuthn exceptions, populate list

woodruffw:
Populates `webauthn-errors` with any errors we receive from the Credentials API, and uses a more idiomatic Promises structure to avoid fall-through exceptions.

Also forces `fetch` to send credentials on the same origin, which was causing 303s with Chrome 67.


brainwane:
@di mind merging?

warehouse/static/js/warehouse/utils/webauthn.js
```diff
@@ -185,23 +187,28 @@ export const ProvisionWebAuthn = () => {
...
-    }
+      const status = await postCredential(label, transformedCredential, csrfToken);
+      if (status.fail) {
+        populateWebAuthnErrorList(status.fail.errors);
+        return;
+      }
 
-    window.location.replace("/manage/account");
+      window.location.replace("/manage/account");
+    }).catch((error) => {
```
webknjaz: why not use try/catch style instead?

woodruffw: I was under the impression that await-then-catch was the idiomatic way to do this kind of pattern with async code in ES6. But N.B. that I'm not much of a JS programmer.

PR #6319, opened 2019-07-30
Warn users about the consequences of deleting a project

### LINKED ISSUE #6318: Warn maintainers that deleting a project allows the name to be reclaimed
Maintainers deleting their project may not realise that doing so frees up the name completely, allowing it to be claimed by malware authors, without any managed transfer of maintenance responsibility.

* an explicit warning that full deletion allows immediate name takeover
* an easy way to yank every released version instead (preferably as one of the options in the warning notice)

**Additional context**
* Victor Stinner just did this for several of his projects, and may need help reclaiming the names

di:
Fixes #6318.

Screen Shot 2019-07-30 at 12 01 37 PM


brettcannon:
What does "with 1 release" mean? Is the latest release taken out but the files for all other releases left behind? (Basically I'm wondering if project deletion leads to a situation as bad as left-pad or just more of an index deletion than a complete nuking of the project).

di:
The "with `N` release(s)` line is already present, it links to all the existing releases (and the number changed depending on the number of releases).

From a user's perspective, project deletion deletes everything: the project, all releases, all files.

Behind the scenes: we delete everything _except_ the actual distribution file in our storage bucket, but the only way a user would be able to continue installing a deleted distribution would be to have the URL to that file and do `pip install https://files.pythonhosted.org/../............`. We also maintain a reference to what distributions once existed, which allows us to continue preventing filename reuse.

nlhkabu:
Updated as discussed @di 

Screenshot from 2019-07-31 07-24-28

PR #6342, opened 2019-08-01
warehouse: Friendlier token "username", prefix

### LINKED ISSUE #6287: Change username & API token prefix, to make Travis auth easier
Just wanted to share a potential pain point for folks using the new PyPI API tokens for authentication via a travis.yaml. The @token username needs to be wrapped in quotes and the `:` after `pypi` needs to be escaped to work, otherwise you hit a 403 error (thanks for the help tracking this down @ewdurbin).

If you're using the environment variables through Travis' UI, the following works:
```
deploy:
  provider: pypi
  user: "@token"  # quotes required for travis
  # server: https://test.pypi.org/legacy/  # uncomment to do a test deploy
  password:
    secure: $PASSWORD  # stored in travis env var, with `:` after `pypi` escaped (pypi\:)
  on:
    branch: master
  skip_existing: true
```

If you're encrypting the token via the command line, you can just wrap your token in quotes:
```travis encrypt "<your-pypi_token>" --add deploy.password [--com]```

### PR DESCRIPTION
**Pull request by woodruffw**
Changes the token username to `__token__` and the prefix to `pypi-`.

Preserves the original `@token` username and `pypi:` token prefix for backwards compatibility during the beta.

Closes #6287.

### DIFF
--- a/tests/unit/macaroons/test_auth_policy.py
+++ b/tests/unit/macaroons/test_auth_policy.py
@@ -32,7 +32,7 @@ from warehouse.macaroons.services import InvalidMacaroon
         ("maybeafuturemethod foobar", None),
         ("token foobar", "foobar"),
         ("basic QHRva2VuOmZvb2Jhcg==", "foobar"),  # "@token:foobar"
-        ("basic JHRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
+        ("basic XnRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
     ],
 )
 def test_extract_http_macaroon(auth, result):
@@ -50,7 +50,7 @@ def test_extract_http_macaroon(auth, result):
         ("bm90YXJlYWx0b2tlbg==", None),  # "notarealtoken"
         ("QGJhZHVzZXI6Zm9vYmFy", None),  # "@baduser:foobar"
         ("QHRva2VuOmZvb2Jhcg==", "foobar"),  # "@token:foobar"
-        ("JHRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
+        ("XnRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
     ],
 )
 def test_extract_basic_macaroon(auth, result):
--- a/tests/unit/macaroons/test_auth_policy.py
+++ b/tests/unit/macaroons/test_auth_policy.py
@@ -32,7 +32,7 @@ from warehouse.macaroons.services import InvalidMacaroon
         ("maybeafuturemethod foobar", None),
         ("token foobar", "foobar"),
         ("basic QHRva2VuOmZvb2Jhcg==", "foobar"),  # "@token:foobar"
-        ("basic XnRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
+        ("basic XnRva2VuOmZvb2Jhcg==", "foobar"),  # "^token:foobar"
     ],
 )
 def test_extract_http_macaroon(auth, result):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by woodruffw** [comment]
cc @di @ewdurbin

**Comment by woodruffw** [comment]
> Isn't `$` as a username prefix going to cause headaches for some shells?

Yeah, hadn't thought of that. I'll change it to `^`; that shouldn't interfere with most things.



> Aside from that I'd just like to see us file an issue preemptively of merge and reference it for removal of backward compatibility for beta "grammar" of these.

👍

**Comment by woodruffw** [comment]
Alright, opened #6345 to track removal.

File: tests/unit/macaroons/test_auth_policy.py [hunk 1]
```diff
@@ -32,6 +32,7 @@
         ("maybeafuturemethod foobar", None),
         ("token foobar", "foobar"),
         ("basic QHRva2VuOmZvb2Jhcg==", "foobar"),  # "@token:foobar"
+        ("basic XnRva2VuOmZvb2Jhcg==", "foobar"),  # "$token:foobar"
```
ewdurbin: comment needs "repaired"

woodruffw: Blah, thanks

warehouse/macaroons/auth_policy.py
```diff
@@ -38,7 +38,8 @@ def _extract_basic_macaroon(auth):
     except ValueError:
         return None
 
-    if auth_method != "@token":
+    # TODO: Remove @token as an acceptable token username.
```
ewdurbin: can you add GH-6345 to the comment?

warehouse/macaroons/services.py
```diff
@@ -31,21 +31,21 @@ class DatabaseMacaroonService:
...
-        if raw_macaroon is None:
+        if prefixed_macaroon is None:
             return None
 
-        try:
-            prefix, raw_macaroon = raw_macaroon.split(":", 1)
-        except ValueError:
-            return None
+        prefix, split, raw_macaroon = prefixed_macaroon.partition("-")
+        # TODO: Remove ':' as an acceptable delimiter for tokens.
```
ewdurbin: can you add GH-6345 to the comment?

warehouse/templates/pages/help.html
```diff
@@ -289,7 +289,7 @@ <h3 id="apitoken">{{ apitoken() }}</h3>
         <p>To use an API token:</p>
 
         <ul>
-          <li>Set your username to <code>@token</code></li>
+          <li>Set your username to <code>$token</code></li>
```
graingert:
```suggestion
          <li>Set your username to <code>^token</code></li>
```

pradyunsg:
Caret (^) is an escape character in Windows batch scripts.

https://stackoverflow.com/questions/20342828/what-does-symbol-mean-in-batch-script

Is there a reason that we aren't using the "token" username here?

woodruffw:
> Caret (^) is an escape character in Windows batch scripts.

:disappointed: 

Maybe `'token` or `+token`?

> Is there a reason that we aren't using the "token" username here?

I listed some reasons in this comment: https://github.com/pypa/warehouse/issues/6345#issuecomment-517714693

graingert:
I had a look at the regex used in warehouse for validating usernames and came up with some potentially nicer options:

`_token`, `-token`, `.token`
`token_`, `token-`, `token.`

and my personal favorite:
`__token__`

di:
FYI, `'token` and `-token` would both require quoting in YAML as well.

woodruffw:
I like `__token__`! Will update the PR to reflect it while we continue to discuss.

ewdurbin:
Looks good to me, any concerns with the selected prefix or username of `__token__` @di @dstufft ?

ZaxR:
Just made the change, and `__token__` is working well with Travis, but just an FYI that the hyphen in the token still needs to be escaped for Travis.

PR #6348, opened 2019-08-02
Fix minor accessibility issues on project detail page

### LINKED ISSUE #6012: a11y: Project view
Accessibility improvements can be made to `/project/{project}/`:

In the history tab:

* `<div class="release__graphic">`  has child `<div>`s that could either be `aria-hidden=true` **or** should have descriptive `alt` texts (current text is "History Node" for both current and past releases)

In the files tab:

* The file `<table>` itself should have a `<caption>`, as the first child node. This is partially mitigated by the `<h2>` above
* The  `<table>` should include `scope=col` or another appropriate scope for each `<th>` node

### LINKED ISSUE #6081: a11y: Package download table
Minor accessibility improvements can be made to `/project/{project_name}/#files`:

* The `<table>` itself should have a `<caption>`, as the first child node. This is partially mitigated by the `<h2>` above
* The `<table>` should include `scope=col` or another appropriate scope for each `<th>` node

nlhkabu:
Closes #6012 
Closes #6081

> `<div class="release__graphic">` has child `<div>`s that could either be aria-hidden=true or should have descriptive alt texts (current text is "History Node" for both current and past releases)

Instead, I have just emptied the `alt` tags in the images, which is also WCAG compliant

> The file `<table>` itself should have a `<caption>`, as the first child node. This is partially mitigated by the `<h2>` above.  The `<table>` should include scope=col or another appropriate scope for each `<th>` node

Done (mentioned in #6012 *and* #6081)

> The pip installation suggestions lack an aria-label

@woodruffw I didn't understand this one -> what are you advising here?

> The button to copy the pip invocation into the clipboard lacks type=button

Fixed in https://github.com/pypa/warehouse/pull/6313

> For the maintainers section (see screenshot 1), the user icon and username links can be combined into a single link, avoiding content duplication with a screen reader.

Done

> For the "Last released" element (see screenshot 2), use Jinja2's localize instead of humanize to insert proper ARIA tags.

Will address this globally with https://github.com/pypa/warehouse/issues/6312

> For improved accessibility and consistency with other instances of clipboard actions, the link should be turned into a button and made larger and/or placed into its own column on the download table.

Done:

Screenshot from 2019-08-02 07-13-26

I've made this a button which opens a modal to reveal all hashes - this mirrors the functionality in the admin

> The font size on the table can also be increased from 14px, potentially to 17px (for consistency with other tables in the project view).

Will review with https://github.com/pypa/warehouse/issues/6349


woodruffw:
> I didn't understand this one -> what are you advising here?

Sorry for the ambiguity -- that was a reference to the `<span id="pip-command">` where we provide an example `pip install` invocation. Looking at it again the text is pretty self-descriptive, so I think that `aria-label` suggestion was wrong on my part.

PR #6353, opened 2019-08-02
Approach for resolving #6232

### LINKED ISSUE #6232: Improved error reporting for token permissions
As a follow on to #6084, we should improve the message returned to users when their provided token isn't acceptable in the given context (e.g., if it doesn't match the package or doesn't have the right permissions at all).

Current `master` provides this message, which is misleading:

> `The user 'brainwane' isn't allowed to upload to project 'Forms990-analysis'`

#6084 changes that message to the more generic:

> `The credential associated with user 'brainwane' isn't allowed to upload to project 'Forms990-analysis'`

But ideally we'd have something like this:

> `This API token is not valid for project 'Forms990-analysis'`

cc @nlhkabu @brainwane

ewdurbin:
Fixes #6232.

Resulting error messages:

Invalid Username/Password:

```
HTTPError: 403 Client Error: Invalid or non-existent authentication information. for url: http://localhost/legacy/
```

Invalid API Token:

```
HTTPError: 403 Client Error: Invalid or non-existent authentication information. for url: http://localhost/legacy/
```

Valid Username/Password that does not have permission:

```
HTTPError: 403 Client Error: The user '<USER_NAME>' isn't allowed to upload to project '< PROJECT_NAME >'. See http://localhost/help/#project-name for more information. for url: http://localhost/legacy/
```

Valid API Token missing scope for project:

```
HTTPError: 403 Client Error: Invalid API Token: project-scoped token is not valid for project '<PROJECT_NAME>' for url: http://localhost/legacy/
```

### DIFF
--- a/warehouse/macaroons/auth_policy.py
+++ b/warehouse/macaroons/auth_policy.py
@@ -139,7 +139,7 @@ class MacaroonAuthorizationPolicy:
                 macaroon_service.verify(macaroon, context, principals, permission)
             except InvalidMacaroon as exc:
                 return WarehouseDenied(
-                    "Invalid API Token: %s", str(exc), reason="invalid_api_token"
+                    f"Invalid API Token: {exc}!r", reason="invalid_api_token"
                 )
 
             # If our Macaroon is verified, and for a valid permission then we'll pass
@@ -149,7 +149,8 @@ class MacaroonAuthorizationPolicy:
                 return self.policy.permits(context, principals, permission)
             else:
                 return WarehouseDenied(
-                    f"API tokens are not valid for permission: {permission}!"
+                    f"API tokens are not valid for permission: {permission}!",
+                    reason="invalid_permission",
                 )
 
         else:
--- a/warehouse/macaroons/auth_policy.py
+++ b/warehouse/macaroons/auth_policy.py
@@ -18,7 +18,7 @@ from pyramid.threadlocal import get_current_request
 from zope.interface import implementer
 
 from warehouse.cache.http import add_vary_callback
-from warehouse.errors import DeniedMacaroonInvalid, DeniedMacaroonNoRequest
+from warehouse.errors import WarehouseDenied
 from warehouse.macaroons.interfaces import IMacaroonService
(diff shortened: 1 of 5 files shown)


ewdurbin:
@dstufft do you have a preference for which approach to take here?

ewdurbin:
Offered another approach based on feedback from @dstufft, though I'm still feeling really icky about how much we're having to bubble through.

brainwane:
This item is a beta blocker https://github.com/pypa/warehouse/milestone/17 so it'd be nice to address this month so we can send out a pypi-announce email around Halloween (security is spooky!).

brainwane:
I would like to wrap up the beta of the API token feature and email out an announcement to get more people using the feature. Thus: @ewdurbin could you please rebase this, and then @dstufft could you review it, so we can finish getting this merged and then get more people uploading with API tokens?

Thanks.

brainwane:
IMO fixing this is still a blocker for coming out of beta https://github.com/pypa/warehouse/issues/5661#issuecomment-515802931 -- @ewdurbin @di @woodruffw @yeraydiazdiaz could we finish this PR within the next few weeks so we can announce the feature on pypi-announce, increase adoption, and better secure PyPI packages?

ewdurbin:
I think this is ready for review.

warehouse/macaroons/auth_policy.py
```diff
@@ -136,15 +138,19 @@ def permits(self, context, principals, permission):
...
 
             # If our Macaroon is verified, and for a valid permission then we'll pass
             # this request to our underlying Authorization policy, so it can handle its
             # own authorization logic on the prinicpal.
             if permission in valid_permissions:
                 return self.policy.permits(context, principals, permission)
             else:
-                return Denied(f"API tokens are not valid for permission: {permission}!")
+                return WarehouseDenied(
+                    f"API tokens are not valid for permission: {permission}!"
```
di: Does this need a `reason`?

ewdurbin: indeed, it should have one :)

Issue #6293, opened 2019-07-29
Persist data on "Add API token" form when validation fails


di:
Currently when using the "Add API token" form at `/manage/account/token/`, if there is an error in the form (either there is no token name or scope selected), when the page is loaded, any existing form data is removed. It'd be nice if anything the user has already entered is persisted, in addition to the errors being shown.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

Sterbic:
Attempting this during the PyCon AU sprints :)

PR #6356, opened 2019-08-05
Persist create token form params on error

Sterbic:
Previously the form would clear if an error was detected on submit.

Video capture: https://www.dropbox.com/s/6t6fwefi0gtjdf5/token.mov?dl=0

Closes: #6293

### DIFF
--- a/warehouse/manage/views.py
+++ b/warehouse/manage/views.py
@@ -566,6 +566,7 @@ class ProvisionMacaroonViews:
                 user_id=self.request.user.id,
                 macaroon_service=self.macaroon_service,
                 project_names=self.project_names,
+                description="",
             ),
             "delete_macaroon_form": DeleteMacaroonForm(
                 macaroon_service=self.macaroon_service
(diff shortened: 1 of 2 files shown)


warehouse/templates/manage/token.html
```diff
@@ -90,7 +90,8 @@ <h2>Add another token</h2>
           <option disabled selected value="scope:unspecified">Select scope...</option>
```
di: We should probably make the `selected` attribute here conditional based on what `create_macaroon_form.token_scope.data` is.

warehouse/manage/views.py
```diff
@@ -566,6 +566,7 @@ def default_response(self):
                 user_id=self.request.user.id,
                 macaroon_service=self.macaroon_service,
                 project_names=self.project_names,
+                description="",
```
di: Might be better to do `{{ create_macaroon_form.description.data or ""}}` in the template instead of this.

PR #6358, opened 2019-08-05
Only show 2fa banner to users with a verified email

nlhkabu:
Closes https://github.com/pypa/warehouse/issues/6239


warehouse/templates/includes/session-notifications.html
```diff
@@ -31,7 +31,7 @@
 </div>
 {% endif %}
 
-{% if request.user and not request.user.has_two_factor %}
+{% if request.user.primary_email.verified and not request.user.has_two_factor %}
```
di:
```suggestion
{% if request.user and request.user.primary_email and request.user.primary_email.verified and not request.user.has_two_factor %}
```

di:
```suggestion
{% if request.user and request.user.primary_email.verified and not request.user.has_two_factor %}
```

PR #6363, opened 2019-08-05
warehouse: Use normalized project names for API tokens

woodruffw:
cc @di @ewdurbin @dstufft


ewdurbin:
That certainly makes sense! We might need to do a quick verification after merge and notify users who hit this bug.

PR #6370, opened 2019-08-06
Mark required fields

### LINKED ISSUE #6281: Update form labels to reflect required status and format requirements
WCAG 3.3.1:

> Required form elements or form elements that require a specific format, value, or
length provide this information within the element's label.

nlhkabu:
Closes #6281 
Screenshot from 2019-08-06 06-50-30

### DIFF
--- a/warehouse/static/sass/blocks/_form-group.scss
+++ b/warehouse/static/sass/blocks/_form-group.scss
@@ -16,7 +16,10 @@
   A form field, with label, input, and help text.
 
   <div class="form-group">
-    <label class="form-group__label" for="myinput">Label here</label>
+    <label class="form-group__label" for="myinput">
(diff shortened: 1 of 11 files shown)


di:
I think it might be better to add something like:

```
{% if field.flags.required %}(Required){% endif %}
```

to every field, so this happens automatically instead of having to be done manually.

Might make sense to use a macro to render a field in a consistent way so we only have to do this in one place, but that would also require figuring out where to set the human-readable field name, and doing it for all the fields.

nlhkabu:
Hi @di, @woodruffw - I've updated this PR to use `field.flags.required` as suggested, and opened a ticket for creating a field macro here: https://github.com/pypa/warehouse/issues/6393

PR #6373, opened 2019-08-06
Multi project token

### LINKED ISSUE #6292: Allow scoped tokens with rights for multiple projects
I have a project (PyObjC) that generates multiple PyPI distributions/projects from a single source base. This is both for historical reasons and because a single repo makes development easier.

API tokens are currently suboptimal for this, I need to create either:
- A user-scoped token for my normal user
- A user-scoped token for a new user with limited rights
- Multiple project-scoped tokens

It would be nice if it were possible to create a project-scoped token that's scoped to multiple projects.

**Additional context**
See above.

Note that this should IMHO have low priority, the work around with a new user with limited rights is workable for me.

Sterbic:
This PR adds support for multi project tokens. The backend was mostly ready for this it seems, I just had to make the forms and other UI components compatible wit it.

Thanks to @ip4368 for the help with the CSS so that the multi select doesn't look as if it were from the 90s.

Screen capture: https://www.dropbox.com/s/ukyqp5ag0ocrbd2/multi-proj-token.mov?dl=0

Closes: #6292


Sterbic:
CI seems to have some transient issues with `apt get` but can't find a way of retrying it myself :/

warehouse/manage/forms.py
```diff
@@ -216,34 +221,41 @@ def validate_description(self, field):
...
-
-        if scope_kind != "project":
-            raise wtforms.ValidationError(f"Unknown token scope: {scope}")
-        if scope_value not in self.project_names:
-            raise wtforms.ValidationError(
-                f"Unknown or invalid project name: {scope_value}"
-            )
+            if scope_kind == "user":
+                if len(scopes) != 1:
+                    raise wtforms.ValidationError(f"Do not mix user and project scopes")
```
woodruffw: Nit: Maybe drop the imperative?

```python
raise wtforms.ValidationError("Mixed user and project scopes")
```

woodruffw: In addition to this, we could probably also add some JS to prevent the user from selecting both project scopes and the user scope on the client side.

Sterbic: That sounds like a good next step but I can barely read JS so would need some guidance on how to make that happen :P

nlhkabu:
HI @Sterbic - thank you for this PR!

I have a suggestion regarding the user interface.  Instead of throwing an error when a user makes a mistake with the token scope, we can prevent it by not allowing users to select their entire account *and* a project at the same time.

I was thinking something like this would work well:

### For accounts with fewer than 10 projects

- No radio selected by default
- Checkboxes disabled *unless* "by project" is selected.

Screenshot from 2019-08-08 08-58-56


### For accounts with more than 10 projects

- No radio selected by default
- Select disabled *unless* "by project" is selected

more-than-10

Sterbic:
Thanks @nlhkabu, that design looks pretty neat! Would you be ok with it going into a separate pull request since it would be touching only the UI? This PR has all the backend logic and tests so I'd rather merge it as is so it's easier to review the UI changes afterwards.

nlhkabu:
Hi @Sterbic - my strong preference is to include the UI improvements in this PR, rather than ship as it and fix it later. My reason: there is no guarantee we'll actually "fix it later".

Did you want some help with the HTML/CSS/JS here?

Sterbic:
Fair enough, I'll amend this PR. Is there a similar example in the codebase for such a UI? I have zero JS experience tbh.

di:
Agreed w/ @nlhkabu.

As an alternative to adding the commits here, someone else could create a PR against your branch.

We use [Stimulus](https://stimulusjs.org) for our JS framework, this should be a pretty straightforward controller.

woodruffw:
>  Instead of throwing an error when a user makes a mistake with the token scope, we can prevent it by not allowing users to select their entire account _and_ a project at the same time.

Just a nitpick: we still need to perform this validation on the backend, since a user could use `curl` or another client to bypass the client-side restriction on what they're allowed to select. But it'll be good to *also* have it on the frontend :smile:

brainwane:
@yeraydiazdiaz do you think you could help @Sterbic here?

warehouse/templates/manage/token.html
```diff
@@ -35,7 +35,9 @@ <h2>Token for "{{ macaroon.description }}"</h2>
       {% if macaroon.caveats.permissions == "user" %}
       <strong>Scope:</strong> Entire account (all projects)
       {% else %}
-      <strong>Scope:</strong> Project "{{ macaroon.caveats.permissions.projects[0] }}"
+      {% for project in macaroon.caveats.permissions.projects %}
+        <strong>Scope:</strong> Project "{{ project }}"<br>
```
pradyunsg: nit-pick: We probably shouldn't add a "Scope:" for every project and instead have it just once, with the project names in a comma-separated-with-and-for-final-join manner.

yeraydiazdiaz:
@Sterbic do you want to have a go at the JavaScript for this or should I tackle it myself?

Sterbic:
Have been a bit swamped at work since I started working on this at PuCon AU but definitely want to finish it.

yeraydiazdiaz:
That's fine, let me know if you need help 🙂

nlhkabu:
Note:

Sterbic:
@yeraydiazdiaz, if you want to take a stab at the JS part go for it. I have a fire at work and won't have any cycles to spend learning JS any time soon 😭 I can rebase this PR on master if it helps.

yeraydiazdiaz:
Hey @Sterbic no worries, I can rebase myself and take it from here. Thanks so much for starting this PR 🌟

yeraydiazdiaz:
@Sterbic, I created a [PR on your fork](https://github.com/Sterbic/warehouse/pull/1) merging `master` onto your branch to bring this PR up to date. I was planning on creating an additional one with the JavaScripts changes [as suggested by @di](https://github.com/pypa/warehouse/pull/6373#issuecomment-520013690).

If you don't have time or don't want the noise let me know. 🙂

yeraydiazdiaz:
I've implemented the changes suggested by @nlhkabu, here's the capture using checkboxes:

token-multi-project-checkboxes

And another using multi-select, note this should only happen if the number of projects it greater than 10, I changed the code in the template to use it in this example:

token-multi-project-multiselect

Once the PR in @Sterbic's repo is merged this should be ready for review.

warehouse/templates/manage/token.html
```diff
@@ -133,23 +135,42 @@ <h2>{% trans %}Add another token{% endtrans %}</h2>
...
-          <option value="scope:user">{% trans %}Entire account (all projects){% endtrans %}</option>
+        {% set token_scopes = create_macaroon_form.token_scopes.data %}
+        {% set all_projects = not token_scopes or token_scopes == ["scope:user"] %}
+        <div>
+          <input type="radio" name="token_scopes" value="scope:user" id="scope:user" data-action="token-scopes#user" {{ "checked" if all_projects }}>
+          <label for="scope:user">{% trans %}Entire account (all projects){% endtrans %}</label>
+        </div>
+        <div>
+          <input type="radio" name="token_scopes" value="scope:by_project" id="scope:by_project" data-action="token-scopes#project" {{ "checked" if not all_projects }}>
+          <label for="scope:by_project">{% trans %}By project{% endtrans %}</label>
```
woodruffw: Could you explain the distinction between `scope:project` and `scope:by_project` here?

If I'm understanding correctly `scope:project` gets used for a single project selection and `scope:by_project` gets used for a multi-selection, but IMO it might make sense to treat either `scope:` as a list and just standardize on one.

yeraydiazdiaz: > If I'm understanding correctly scope:project gets used for a single project selection and scope:by_project gets used for a multi-selection

Yes. Although I'm not sure we actually `scope:by_project` at all 🤔 

> but IMO it might make sense to treat either scope: as a list and just standardize on one

Not sure what you mean, sorry.

To clarify a bit what's happening, all the token related inputs in the form use `token_scopes` as their name. The backend's form validation reads the list of values and produces the list as required.

yeraydiazdiaz: > Although I'm not sure we actually need scope:by_project at all 🤔

Given it's a radio button both inputs need to have the same name. We could change the name of the per project token inputs, but we still have to pass the "by_project" value, whatever it may be, and ignore it if it's active in the form validation logic so we wouldn't be gaining a lot.

di: I'm also not following what the distinction is here. I think what @woodruffw is saying is: why can't we use the same scope for both of them, and handle `token_scopes` as being a list with one scope (per-project) or multiple scopes (multi-project)?

yeraydiazdiaz: Right, I think I understand.

The only valid values for tokens scopes are `scope:user` and a list of one or more `scope:project:<PROJECT_NAME>`, i.e. there is no `scope:project` alone without a `:<PROJECT_NAME>` suffix. The form validation ensures the user cannot mix `scope:user` and `scope:project:<PROJECT_NAME>`.

`scope:by_project` is only there to allow the radio buttons to flick via standard HTML, it is explicitly ignored by the form validation. I could really be called anything else and probably should.

We could also not rely on HTML for the radio button flick, that might make things easier to understand but it would add complexity to the Stimulus controller.

Does that make sense or am I still missing something?

yeraydiazdiaz: Gentle ping on this comment, was I still missing something? Would removing `scope:` from `scope:by_project` suffice?

woodruffw: Yes, I think dropping the `scope:` prefix is appropriate given that it isn't parsed during form validation.

yeraydiazdiaz: Pshed the changes to https://github.com/Sterbic/warehouse/pull/2. I'll ask for a re-review once @Sterbic merges it.

yeraydiazdiaz:
@Sterbic has merged https://github.com/Sterbic/warehouse/pull/2 so this is now ready for review.

miketheman:
#Triage 
This PR currently has merge conflicts that would need to be addressed before proceeding.

@Sterbic are you still interested in pursuing this effort? It's okay if not!

firefly9966 (approved):

ttodua:
is this still planed or not? would be very useful.

woodruffw:
> is this still planed or not? would be very useful.

It's still planned in the sense that there's desire for it, but it needs an owner unless @Sterbic comes back to it.

In the mean time, you can use a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/) to perform multi-project publishing -- that's built natively into the TP flow.

PR #6384, opened 2019-08-06
warehouse: Guard macaroon deserialization

woodruffw:
Guards calls to `Macaroon.deserialize`, preventing a malformed macaroon from causing an unexpected exception.


woodruffw:
cc @di

PR #6389, opened 2019-08-07
warehouse: Avoid invalid button attribute for WebAuthn

### LINKED ISSUE #6387: Remove csrf-token from button (?)
https://github.com/pypa/warehouse/blob/master/warehouse/templates/manage/account/webauthn-provision.html#L47

This line is throwing a W3C error:
`ERROR: Attribute “csrf-token” not allowed on element “button” at this point`

@woodruffw how can we refactor this functionality to use valid HTML5?

woodruffw:
Uses the `value` attribute to stash our CSRF token instead of the custom (and invalid) `csrf-token` attribute.

cc @nlhkabu 

closes #6387

### DIFF
--- a/warehouse/static/js/warehouse/utils/webauthn.js
+++ b/warehouse/static/js/warehouse/utils/webauthn.js
@@ -40,13 +40,8 @@ const doWebAuthn = (buttonId, func) => {
     return null;
   }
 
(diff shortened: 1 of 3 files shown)


nlhkabu:
Thank you @woodruffw

PR #6408, opened 2019-08-10
2fa disabling ask password

### LINKED ISSUE #5825: 2FA Ask for password not username when disabling 2FA
**Describe the bug**
When disabling 2FA for an account, it would make sense to ask for the user's password and not the username for confirmation as that would make it much harder to do drive-by disabling on unattended computers.

**Expected behavior**

Ask for password when disabling 2FA.

**To Reproduce**

Try disabling 2FA on https://pypi.org/manage/account/.

yeraydiazdiaz:
Fixes #5825

Here's a capture of the new functionality:

2fa_remove_ask_password

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -95,7 +95,7 @@ class PasswordMixin:
         super().__init__(*args, **kwargs)
 
     def validate_password(self, field):
-        userid = self.user_service.find_userid(self._get_username())
+        userid = self.user_service.find_userid(self.username.data)
         if userid is not None:
             try:
                 if not self.user_service.check_password(
@@ -109,13 +109,6 @@ class PasswordMixin:
                     "There have been too many unsuccessful login attempts, "
                     "try again later."
                 ) from None
-
-    def _get_username(self):
-        # the username may be set directly in the instance or as a field
-        if isinstance(self.username, str):
-            return self.username
-        else:
-            return self.username.data
 
 
 class NewPasswordMixin:
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -95,7 +95,7 @@ class PasswordMixin:
         super().__init__(*args, **kwargs)
 
     def validate_password(self, field):
-        userid = self.user_service.find_userid(self.username.data)
+        userid = self.user_service.find_userid(self._get_username())
         if userid is not None:
             try:
                 if not self.user_service.check_password(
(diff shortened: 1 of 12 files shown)


warehouse/accounts/forms.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, *args, check_password_metrics_tags=None, **kwargs):
         super().__init__(*args, **kwargs)
 
     def validate_password(self, field):
-        userid = self.user_service.find_userid(self.username.data)
+        userid = self.user_service.find_userid(self._get_username())
```
woodruffw: You could probably use the `field` parameter directly here to avoid this `_get_username` helper.

yeraydiazdiaz: I believe in this case `field` is the `password` field not the username. This mixin expects the instance to have a `username` attribute of type field, which it does most of the time as it's usually paired with the `UsernameMixin`, but in this case it's not, so I set it as a [string instance variable](https://github.com/pypa/warehouse/pull/6408/files#diff-19b7c365d44b94a6782decdcdfeee79aR96).

woodruffw: Ah, sorry, misinterpreted the context.

nlhkabu:
Hi @yeraydiazdiaz, thanks for this PR :grin: 

A couple of items:

Could we please update the error text (when a user enters an incorrect password) to be more meaningful than 'invalid credentials'?

Screenshot from 2019-08-12 08-48-36

Maybe something like: 
"Incorrect password. Try again." ??

Can we also please use this functionality for the 'remove API token' modal as well?

Thanks!

brainwane:
@yeraydiazdiaz is this ready for re-review?

yeraydiazdiaz:
Yes, it is. Here is a capture of the changes requested by @nlhkabu 

token_remove_ask_password

warehouse/manage/forms.py
```diff
@@ -87,12 +87,13 @@ def __init__(self, *args, user_service, **kwargs):
...
-class DeleteTOTPForm(UsernameMixin, forms.Form):
+class DeleteTOTPForm(PasswordMixin, forms.Form):
 
-    __params__ = ["confirm_username"]
+    __params__ = ["confirm_password"]
 
-    def __init__(self, *args, user_service, **kwargs):
+    def __init__(self, *args, username, user_service, **kwargs):
         super().__init__(*args, **kwargs)
+        self.username = username
```
di: Why not just make this a regular form field?

yeraydiazdiaz: I didn't think of that to be honest, probably because the actual HTML form only has a password, but it does make sense to have both mixins and manually add the username in the views. Thanks 👍

yeraydiazdiaz:
I've implemented the change suggested by @di and rebased on master, this ready for re-review.

yeraydiazdiaz:
I merged master onto this branch and had to update the some of the tests, I believe everything's correct but @woodruffw could you review again just in case?

nlhkabu:
Thank you @yeraydiazdiaz ! :tada:

PR #6409, opened 2019-08-10
#5769 - handle null values both in trigger and in route

### LINKED ISSUE #5769: 503 on /stats route

DavidBord:
Fixes #5769


ewdurbin:
Thank you @DavidBord!

PR #6419, opened 2019-08-13
Normalize classifier IDs for HTML4 compatibility

### LINKED ISSUE #6380: Invalid classifier IDs on search results page
Each classifier on the search results page displays in the sidebar like so:

```html
<li>
    <input name="c" type="checkbox" id="Framework :: AsyncIO" class="-js-form-submit-trigger checkbox-tree__checkbox" value="Framework :: AsyncIO">
    <label class="checkbox-tree__label" for="Framework :: AsyncIO">AsyncIO</label>
 </li>
```

The ID here e.g. (`"Framework :: AsyncIO"`) is invalid HTML as it contains spaces.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

We should also check that we are following MDNs recommendation:

> Note: Using characters except ASCII letters, digits, '_', '-' and '.' may cause compatibility problems, as they weren't allowed in HTML 4. Though this restriction has been lifted in HTML5, an ID should start with a letter for compatibility.

woodruffw:
Needs discussion on a few items:

* This normalizes spaces and `::`, but leaves in slashes, apostrophes and other potentially HTML4-invalid characters. I can't find a formal definition of the classifier format anywhere (PEP 301 mentions some basic constraints), so I didn't include them in this changeset (yet).
* If the classifier format isn't well specified or sufficiently constrained (e.g., allows for Unicode), then we might want to consider using (urlsafe) base64-encoded IDs. That would solve the normalization issues *and* avoid any potential duplication caused by the normalization process.
* HTML5 doesn't these constraints on the content of identifiers. If the plan is to make PyPI strictly HTML5, then it might not make sense to apply these changes at all.

Closes #6380.

cc @nlhkabu

### DIFF
--- a/tests/functional/test_templates.py
+++ b/tests/functional/test_templates.py
@@ -40,6 +40,7 @@ def test_templates_for_empty_titles():
             "format_rfc822_datetime": "warehouse.i18n.filters:format_rfc822_datetime",
             "format_number": "warehouse.i18n.filters:format_number",
             "format_classifiers": "warehouse.filters:format_classifiers",
+            "classifier_id": "warehouse.filters:classifier_id",
             "format_tags": "warehouse.filters:format_tags",
             "json": "warehouse.filters:tojson",
             "camoify": "warehouse.filters:camoify",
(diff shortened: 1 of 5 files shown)


nlhkabu:
Thanks @woodruffw - I think removing the spaces is enough to close the issue.  I don't think we need to spend any more time worrying about HTML4.

Issue #3218, opened 2018-03-10
Implement an AdminFlag for halting project/release modifications


di:
Similar to #2967, we should implement an `AdminFlag` that allows us to:
* halt all new uploads
* prevent release/project deletion

brainwane:
As with the other spam-related features, I'm leaving this out of the milestone bucketing since it isn't pertinent specifically to [the MOSS grant](https://blog.mozilla.org/blog/2018/01/23/moss-q4-supporting-python-ecosystem/).

calvin:
3 new admin flags:
 - Disallow uploads
 - Disallow release
 - Disallow project deletion

calvin:
@di It lookse like disallowing uploads is already implemented in #3393. Can you confirm the requirements?

di:
@calvin #3393 prevents any database writes at all (e.g., you can't change your display name, etc.) where here a "halt all new uploads" flag means that just new uploads are prevented.

Issue #5837, opened 2019-05-12
Implement PEP 592 -- Yanking of Releases/Files


dstufft:
The way we should implement this is that an entire release can be yanked or unyanked at will by a maintainer of a project. Yanking should set an attribute on the simple repository link for any files for that release.

dstufft:
Here's a PR that starts implementing this https://github.com/pypa/warehouse/pull/5838.

brainwane:
Hey @dstufft -- I think I may be misunderstanding something. In [our discussion in March](https://wiki.python.org/psf/PackagingWG/2019-03-22-Warehouse) we discussed wanting a system for generic flags and statuses on projects and releases, in order to address #345 "Ability to mark a version of a package as deprecated or unsupported" and #3709 "Offer a discouraged/deprecated releases option", and to help with future moderation work on stuff like #6062 re: malware/spam detection.

Will the yanking feature address either of those "mark package as deprecated" issues?

And: are you planning on implementing the "yanked" attribute within a generic flags/statuses system? If not then I should file an issue to start speccing that out.

Also, I figure yanking will constitute an event that should be listed in #5863, so, linking to that issue.

brainwane:
And is #4962 potentially relevant?

cjerdonek:
For reference purposes, I just wanted to note on this issue that PEP 592 support has been added to pip and will be in its next release: https://github.com/pypa/pip/issues/6633

di:
In #6318, @ncoghlan asked that the "delete project/release" dialog to include

> an easy way to yank every released version instead (preferably as one of the options in the warning notice)

Ideally implementing this PR would include such a link/notice/feature.

pradyunsg:
> Will the yanking feature address either of those "mark package as deprecated" issues?

Yep, for both issues. #4962 doesn't seem relevant to me.

brainwane:
@ewdurbin See https://github.com/pypa/pip/pull/7621#issuecomment-576662585 for why this might be a feature to prioritize in Warehouse.

brainwane:
Per https://github.com/pypa/warehouse/pull/5838#issuecomment-578863942 , we'd like help with this. Donald wrote part of a pull request to implement this, and said,

> I am unlikely to find time in the interim to finish this, though I think we could deploy this without an API and just have it be a UI only feature for now.

This feature would be helpful in improving pip's release process, so it would be great if a volunteer could step up and finish this WIP within the next couple of months.

[...]

PR #6518, opened 2019-08-23
Add disallow deletion AdminFlag

calvin:
Add new AdminFlag which prevents projects/release from deletion when enabled.

See #3218


calvin:
Testing upload:

```
python3 -m twine upload --repository-url http://warehouse/legacy/ dist/*
Enter your username: user
Enter your password:
Uploading distributions to http://warehouse/legacy/
Uploading itsdangerous-1.1.1-py2.py3-none-any.whl
100%|██████████████████████████████████████████████████████████████████████████████████████████████| 22.6k/22.6k [00:00<00:00, 83.0kB/s]
NOTE: Try --verbose to see response content.
HTTPError: 403 Client Error: New uploads are temporarily disabled. See http://warehouse/help/#admin-intervention for details for url: http://warehouse/legacy/
```

calvin:
@di This PR is ready to review.

brainwane:
@calvin Thank you for this PR! Would you mind fixing the merge conflict and then say here when you've done that, and I'll try to find some reviewers to help move this along? Thanks!

calvin:
@brainwane resolved conflicts.

warehouse/templates/manage/release.html
```diff
@@ -91,6 +91,7 @@ <h2 class="heading-wsubtitle__heading">Release version {{ release.version }}</h2
                 </a>
               </li>
 
+              {% if not disallow_deletion %}
```
ewdurbin: This should probably be inverted to match the behavior below for "Delete release":

```
{% if disallow_deletion %}
    <p>....message about release deletion being temporarily disabled....</p>
{% else %}
    normal behavior
{% endif %}

File: warehouse/templates/manage/releases.html [hunk 2]
```diff
@@ -78,6 +78,7 @@ <h2>Releases ({{ project.releases|length }})</h2>
                     View
                   </a>
                 </li>
+                {% if not disallow_deletion %}
```
ewdurbin: Similarly, invert to `{% if disallow_deletion %}` and provide a message visible to user.

warehouse/admin/flags.py
```diff
@@ -15,6 +15,14 @@
 from warehouse import db
 
 
+class AdminFlagValue:
```
di:
```suggestion
class AdminFlagValue(enum.Enum):
```
(plus the necessary import)

warehouse/templates/manage/release.html
di: I think instead it'd be better to leave the templates unchanged and instead, and just let the request happen and block it in the "delete" view instead (with a flash indicating that it failed), similar to how we do for user registration and read-only mode.

warehouse/admin/flags.py
calvin: I'm hesitant to make the suggested change. This will make the values as a type of `enum` object. In order to access the string value, you have to call  `AdminFlagValue.XXX.value`. This will be a big change.

calvin: (Accidentally deleted the comment I left earlier)

I'm hesitant to make the suggested change. In order to access the value, you need to use call `AdminFlagValue.XXX.value`. There are many places which just expect a string, e.g tests/unit/admin/test_flags.py. However, it's a good idea and I'm happy to make the changes with another PR.

warehouse/templates/manage/release.html
calvin: Make changes as per @di

warehouse/templates/manage/releases.html
calvin: Using a flash instead.

ewdurbin:
that was a really solid suggestion @di!

warehouse/admin/flags.py
di: I still think this should be an `Enum` to be consistent with other enum-like patterns we have, and it's fine to just include it in this PR as it's own commit.

brainwane:
@calvin I just want to note that this was great for a first PR not just in this repo but in PyPA as a whole. Thank you! If you're interested in helping more, we would welcome it, and I could suggest a couple things if you like!

calvin:
@brainwane I do want to make more contributions. Please tell me more.

brainwane:
@calvin Would you take a look at #5837 ?

brainwane:
@calvin Or, if you want to do some more UI-only stuff, #6606?

Issue #6377, opened 2019-08-06
Fix all tables on mobile


nlhkabu:
- Review all tables on mobile
- Ensure no information is lost on smaller screens

Ref WCAG 1.4.4

PR #6520, opened 2019-08-23
Show files in release summary table

### LINKED ISSUE #3200: Redesign release table
Right now we list the 'summary' of a release in the table:

screenshot from 2018-03-09 07-04-08

Feedback from user testing is that this value *so rarely* changes that it's not worth exposing here.

Instead we should:

1. Add information about what files are in the release
2. Try to expose the buttons directly (move them out of the dropdown if there is enough space)

nlhkabu:
Closes #3200 
Prerequisite for #6377 

Screenshot from 2019-08-23 09-03-25


warehouse/templates/manage/releases.html
```diff
@@ -40,12 +40,18 @@ <h2>Releases ({{ project.releases|length }})</h2>
           </th>
           <td class="table__date">{{ humanize(release.created) }}</td>
           <td class="table__summary">
-            {% if release.summary %}
-              {{ release.summary }}
+            {% set num_files = release.files.all()|length %}
```
nlhkabu: @di is this likely to create performance issues?  i.e. will it hit the db?

di: Yes. I added a commit here which will do this more efficiently, perhaps @woodruffw can add the necessary tests?

As a side effect, we're able to say how many of each packagetype for a given release, so I added that, e.g. instead of:

> 7 files (Wheel, Source)

it will say:

> 7 files (6 Wheel, 1 Source)

feel free to change as necessary, but I think this is more clear.

warehouse/templates/manage/releases.html
```diff
@@ -40,12 +40,22 @@ <h2>Releases ({{ project.releases|length }})</h2>
...
+              {{ count }} files
+            {% endtrans %}
+              ({% for packagetype, packagetype_count in version_to_file_counts[release.version].items() %}
+              {%- if packagetype != 'total' -%}
+              {{ packagetype_count }} {{ packagetype|format_package_type }}{{ ", " if not loop.last }}
+              {%- endif -%}
+              {% endfor %})
             {% else %}
-              —
+            No Files
```
di:
```suggestion
            No files
```

warehouse/templates/manage/releases.html
nlhkabu: Great, thanks Dustin :)

nlhkabu:
Patch applied and merged.  Thanks @di and @woodruffw !

PR #6527, opened 2019-08-25
2fa disabling ask password -- Take 2

### LINKED ISSUE #5825: 2FA Ask for password not username when disabling 2FA

yeraydiazdiaz:
Closes #5825 

This PR builds on top of #6408 and fixes the errors reported by @di

The errors originated on the refactor of the modal template code:

- The `name` attribute on the `input` tag was accidentally deleted which caused the all other "confirm" modal forms to fail.
-  A call with an `index` kwarg was left behind which caused the traceback @di posted.

Here's a capture of the reported broken funcionality working with this PR:

deleting_releases_fix

I'd appreciate a further round of tests before merging if at all possible.


nlhkabu:
I confirm that the other modals are now working again.

nlhkabu:
@yeraydiazdiaz I'd like to ask for the user's password instead of their username when deleting an account. Do you think it's something we could add to this PR?

yeraydiazdiaz:
Thanks for verifying @nlhkabu. Yes, I can add that as well.

yeraydiazdiaz:
@nlhkabu I've added confirming the password on account deletion. Here's a capture:

delete_account_password

brainwane:
@di can you review?

warehouse/templates/manage/manage_base.html
```diff
@@ -137,7 +186,20 @@ <h3 class="modal__title">{{ title }}{% if confirm_string_in_title %} {{ confirm_
...
+  {{ confirm_modal(title, confirm_name, confirm_string, slug, extra_fields=extra_fields, action=action) }}
+{% endmacro %}
+
+{% macro confirm_password_button(
+  title,
+  extra_fields=None,
+  action=None
+)
+%}
+  {% set slug = title.lower().replace(' ', '-') + '-modal' + ('-{}'.format(index) if index else '') %}
```
di: It seems like there will never be an `index` here?

nlhkabu:
@yeraydiazdiaz could you please ensure that the modal content is translatable, e.g.
https://github.com/pypa/warehouse/pull/6598/files#diff-fdc9c6c54285180f385c5f8c0609ccc1R111

This will ensure we don't revert changes from #6598 

Thanks

warehouse/templates/manage/manage_base.html
yeraydiazdiaz: Yeah, good catch. The reason I went for removing it is because none of the calls to `confirm_button` actually use it, they simply the title which seems to be enough. I'll reinstate it anyway.

yeraydiazdiaz: Agh, got it the other way around 😄 I'll remove it

warehouse/templates/manage/manage_base.html
```diff
@@ -79,44 +79,43 @@ <h3 class="sidebar-section__title">Your account</h3>
...
+              {% trans %}This action cannot be undone!{% endtrans %}
             </p>
             {% if custom_warning_text %}{{ custom_warning_text }}{% endif %}
           </div>
           {% endif %}
-          {% set context = 'your' if confirm_name.lower() == 'username' else 'the' %}
+          {% set context = gettext('your') if confirm_name.lower() == gettext('username') else gettext('the') %}
           {{ extra_description if extra_description else '' }}
-          <p>Confirm {{ context }} {{ confirm_name|lower }} to continue.</p>
+          <p>{{ gettext("Confirm %(context) %(confirm_name) to continue.")|format(context=context, confirm_name=confirm_name.lower()) }}</p>
```
yeraydiazdiaz: @nlhkabu I wasn't sure if this is the right way to get translations on a parametrized message. I'm following https://jinja.palletsprojects.com/en/2.10.x/extensions/#id2 but I think [in the PR you mentioned](https://github.com/pypa/warehouse/pull/6598/files#diff-be3f1c8f146e973b06e40074eb56265cR23) there's a slightly different way? I'm not sure if they're equivalent.

yeraydiazdiaz: Nevermind that doesn't actually work.

nlhkabu:
@di are we ready to go here?

@yeraydiazdiaz since we have now merged the translations PR (see #6598) we will need to rebase this, making sure that all `{% trans %}` blocks remain in place.

pradyunsg:
A minor nit: after deleting the account, there's no indication that the action was indeed successful. I'm not sure how difficult it might be to add, but that was something I found myself surprized to not see in the GIF above.

di:
@pradyunsg In the GIF @yeraydiazdiaz is using the incorrect password, so the account doesn't actually delete -- there's a warning, and then the GIF loops.

If the password was correct, the user would have been logged out, which is probably indication enough that it was successful. And since the user is logged out, we can't flash a message. The only other thing we could do would be to redirect to a page that is specifically for confirming that account deletion was successful, but I'm not sure that's worth doing. If you do though, please feel free to create an issue!

pradyunsg:
Thanks for the quick response @di! Reading your description, I don't think there's any change needed here IMO. :)

PR #6535, opened 2019-08-27
Warehouse: l10n skeleton

### LINKED ISSUE #6455: Add language switcher to PyPI UI
For the upcoming localisation work -> we will need to offer a way for users to switch their language.

Probably the easiest way to support this is to add a small section in the footer with text links to the different supported languages.

- We should *not* use flags (a country is not a language)
- The text links themselves should be translated

References:
- https://usersnap.com/blog/design-language-switch/

woodruffw:
TODO:

- [x] Changes to the `.mo` files should cause the web container/service to restart in development mode
- [x] `Accept-Language` is not working, for unclear reasons
- [x] Figure out how to defer `request` so that we can use `_()` in form fields

Closes #6455

### DIFF
--- a/tests/unit/utils/test_msgpack.py
+++ b/tests/unit/utils/test_msgpack.py
@@ -1,0 +1,27 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from warehouse.i18n import LazyString
+from warehouse.utils.msgpack import object_encode
+
+
+def test_object_encode_passes_through():
+    assert object_encode("foo") == "foo"
+
+
+def test_object_encode_converts_lazystring():
+    def stringify(*args, **kwargs):
+        return "foo"
+
+    ls = LazyString(stringify, "foobar")
+    assert object_encode(ls) == "foo"


warehouse/locale/de/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,41 @@
...
+"Language: de\n"
+"Language-Team: de <LL@li.org>\n"
+"Plural-Forms: nplurals=2; plural=(n != 1)\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 2.7.0\n"
+
+#: warehouse/accounts/views.py:59
+msgid "There have been too many unsuccessful login attempts. Try again later."
```
yeraydiazdiaz: Just a quick note, feel free to disregard.

In the past I've been bitten by having the `msgid` being the actual english copy. While more intuitive when reading it in the code it made simple english copy changes much harder.

Is it worth considering a coded `msgid` + an american english de facto translation?

ewdurbin: I'd lean towards using IDs for the reasons that @yeraydiazdiaz has pointed out.

ewdurbin: Decision has been made to use the English copy as the msgid, Conversations in https://github.com/pypa/warehouse/pull/6598 supported this and ultimately it seems that as far as translator experience goes it should serve us best.

woodruffw:
To follow up on the "deferred `request`" issue:

This is our current localization method:

```python
def localize(message, **kwargs):
    request = get_current_request()
    return request.localizer.translate(_translation_factory(message, **kwargs))
```

which we expose as the traditional `_()`-style identifier for string tagging:

```python
from warehouse.i18n import localize as _

# ...

request.session.flash(
    _(
        "Email address ${email_address} verified. ${confirm_message}.",
        mapping={"email_address": email.email, "confirm_message": confirm_message},
    ),
    queue="success",
)
```

This works well for the bodies of views and forms, since there's an active `request` pending that `get_current_request` can yank from the threadlocal state.

Unfortunately, this **doesn't** work in form field declarations, like so:

```python
class TOTPValueMixin:

    totp_value = wtforms.StringField(
        validators=[
            wtforms.validators.DataRequired(),
            wtforms.validators.Regexp(
                rf"^[0-9]{{{TOTP_LENGTH}}}$",
                message=_(f"TOTP code must be {TOTP_LENGTH} digits."),
            ),
        ]
    )
```

...since `_()` is called before a `request` is active.

mmerickel:
Yeah this is the exact issue I was expecting to see - makes sense.

The workflow with the `translationstring` package (you probably know this, but just to frame it):

1) Tag every translatable string as a `TranslationString`.

2) Eventually (you determine when) run those strings through the localizer. Usually it's clear when you should do this because it's the same time that you need to interpolate the values into the string. It's not as clear when the string has no interpolation placeholders (like most strings in this pull request so far).

Your current approach combines both steps into 1 via the `localize(foo, args)` function. This approach falls apart when you need to mark strings as translatable at module scope (with no active request).

Ok cool, problem stated - my solution: I would suggest defining `_ = TranslationStringFactory('messages')` and using `warehouse.i18n.localize` only in functions (as it won't work at module scope). A purist could even remove the usage of `_translation_factory` from `localize`. This basically doesn't change anything you have done so far except renaming `_` to `localize` and wrapping the arg to `localize` in `_`. For example `localize(_('You did evil things.'))`.

For module-scope you need to find a good location to take a translationstring marked with `_` and run it through the localizer. This is the biggest wart in Pyramid's approach to localization imo, but it kinda is what it is unless you define a magic version of translationstring that uses the localizer in `__str__` and just assuming that you will never use `__str__` outside of request scope.

For wtforms (I don't have much experience with this library so much suggestion is likely to be vague here) it probably means at the spot where you handle errors, you receive the error messages and you then run them through `localize(error_msg)`. If the error messages need interpolated values then it needs to be done in the actual error functions (similar to what you are already doing).

I hope this helps and I'm happy to spitball on some other options.

ewdurbin:
The only outstanding question I have is how we will handle setting the locale when rendering messages in email templates.

I don't know that it must be addressed immediately, but given that once we launch translation services the strings will be translatable we might want to consider that.

docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
         # working on pypi-theme, this is a private repository due to the fact
         # that other people's IP is contained in it.
         #THEME_REPO:
-    command: hupper -m gunicorn.app.wsgiapp -b 0.0.0.0:8000 -c gunicorn.conf warehouse.wsgi:application
+    command: hupper -w 'warehouse/locale/**/*.mo' -m gunicorn.app.wsgiapp -b 0.0.0.0:8000 -c gunicorn.conf warehouse.wsgi:application
```
ewdurbin: should this be watching .po files as well?

ewdurbin: d'oh, I missed building `.mo` files. this is fine :)

woodruffw:
> The only outstanding question I have is how we will handle setting the locale when rendering messages in email templates.

Hmm, do we have access to the request that prompted the email? If so, we could use our normal locale negotiation. Otherwise, we could probably add a `locale` field to the user model and set it as the highest priority in the negotiator, when available.

Issue #6472, opened 2019-08-17
form on webauthn page is not a <form>


nlhkabu:
On `warehouse/templates/manage/account/webauthn-provision.html` we have a form where users can provision their device.

This is *not* wrapped in `<form>` tags, which means it is not semantic.

nlhkabu:
@woodruffw I have reassigned this one to you - do you think there is a workaround we can find that will not break the functionality?  If not, let's close this ticket.

Thanks :)

woodruffw:
Thanks!

Yeah, I *think* there should be a workaround that includes both the `<form>` and a bit of JS (similar to the bit we reverted in the last PR) to perform the proper handling. I'll take a look today.

PR #6546, opened 2019-08-29
warehouse: Remove <form> on WebAuth provision

woodruffw:
`<form>` was introduced in #6524, but was causing some weird routing when submitted via the return key rather than a direct button press (probably because `<form>` defaults to `GET`). It's also unfortunately still nonfunctional with `method="POST"`, since we still need to perform the relevant WebAuthn operations in JavaScript.

### DIFF
--- a/warehouse/static/js/warehouse/utils/webauthn.js
+++ b/warehouse/static/js/warehouse/utils/webauthn.js
@@ -205,6 +205,19 @@ export const ProvisionWebAuthn = () => {
       return;
     });
   });
+
+  const labelInput = document.getElementById("webauthn-provision-label");
+
+  if (labelInput === null) {
+    return;
+  }
+
+  labelInput.addEventListener("keyup", async (event) => {
+    if (event.keyCode === 13) {
+      event.preventDefault();
+      document.getElementById("webauthn-provision-begin").click();
+    }
+  });
 };
 
 export const AuthenticateWebAuthn = () => {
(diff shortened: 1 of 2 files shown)


woodruffw:
cc @di @nlhkabu

di:
Is it possible to use `onsubmit` here instead?

woodruffw:
> Is it possible to use `onsubmit` here instead?

I was under the impression that the CSP policy would block it, based on this `script-src` definition:

```
                "script-src": [
                    SELF,
                    "www.googletagmanager.com",
                    "www.google-analytics.com",
                    "*.fastly-insights.com",
                    "https://cdn.ravenjs.com",
                ],
```

But I'll give it a try locally!

woodruffw:
Yep, I get a CSP error when activating the form via `onsubmit`.

nlhkabu:
Why do we need the new JS here?  Can we just revert adding the `<form>` tag for now?

woodruffw:
> Why do we need the new JS here?

The JS handles the case where the user hits the return key on the input label. With `<form>` that caused an (incorrect) `GET` to the endpoint; it now submits the `POST` correctly.

I could remove it and limit this PR to just the revert, but I think the non-`<form>` behavior without it is a little unintuitive (but also not strictly broken).

nlhkabu:
ok, thanks.  I'll defer to @di for reviewing the JS.  Once this is merged, I will reopen #6472

PR #6570, opened 2019-09-03
Use OrderedDict for sorting releases in JSON

techtonik:
Right now releases in JSON are sorted as strings and not as versions, and that might be because of unsorted dict usage.

https://pypi.org/pypi/boto3/1.9.220/json

### DIFF
--- a/warehouse/legacy/api/json.py
+++ b/warehouse/legacy/api/json.py
@@ -121,7 +121,7 @@ def json_release(release, request):
 
     # Map our releases + files into a dictionary that maps each release to a
     # list of all its files.
-    releases = {}
+    releases = OrderedDict()
     for r, file_ in release_files:
         files = releases.setdefault(r, [])
         if file_ is not None:


di:
I'm not sure this necessary. We don't imply any ordering here, and a JSON object is by definition an unordered set of name/value pairs.

Even if we use an `OrderedDict` here, there's no guarantee that whatever parser/serializer/deserializer between us and the user is going to maintain that ordering.

ewdurbin:
I agree with @di, if sort order is important it should be handled explicitly by the client. Are you considering this a regression or just a proposed feature @techtonik, I'm not certain how pypi-legacy responded.

techtonik:
Some languages [are sorting JSON](https://stackoverflow.com/questions/18668652/how-to-produce-json-with-sorted-keys-in-go) by default. It is not for client - it is for debugging. Specifically, if you try to locate entry for the latest release of https://pypi.org/pypi/boto3/json, it will be located somewhere in the middle of the list. Releases are already sorted there, but as strings, not versions. Sorting them doesn't break the JSON standard, and if somebody relies on the sorting behavior - that's not a problem with PyPI.

miketheman:
Reviewing this older issue.

Python 3.7 introduced dict ordering preserved on insertion order. We already order the `release_files` before adding them to `releases`, so the current state is that the results are ordered, unintentionally.

To support visual debugging, the latest release is (often) the last one, but as was pointed out, JSON object ordering is not explicit.

Recommend closing!

PR #6571, opened 2019-09-03
Toggle aria attributes on accordion component

### LINKED ISSUE #6479: Add 'aria-expanded' and 'aria-haspopup` to accordion component
When closed, the accordion triggers (`accordion__link`) should have:
- `aria-haspopup=true`
- `aria-expanded=false`

When open:
- `aria-haspopup=true`
- `aria-expanded=true`

nlhkabu:
Closes #6479 

@woodruffw note that I have reused your `|classifier_id` filter - do you think it would be worthwhile to rename this?

### DIFF
--- a/warehouse/static/js/warehouse/utils/search-filter-toggle.js
+++ b/warehouse/static/js/warehouse/utils/search-filter-toggle.js
@@ -30,8 +30,22 @@ export default () => {
   };
 
   const toggleAccordion = (event) => {
-    var element = event.currentTarget.parentElement;
-    element.classList.toggle("accordion--closed");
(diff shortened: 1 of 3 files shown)


woodruffw:
> note that I have reused your `|classifier_id` filter - do you think it would be worthwhile to rename this?

Potentially -- the name looks mostly appropriate here given that `top_level` is a component of a classifier ID, but the filter is generic enough to probably benefit from a rename if we subsequently reuse it anywhere else.

nlhkabu:
Ok, let's leave it for now and rename later if appropriate :)

PR #6580, opened 2019-09-04
warehouse: Use semantic forms for webauthn actions

### LINKED ISSUE #6472: form on webauthn page is not a <form>

woodruffw:
cc @di @nlhkabu 

Fixes #6472.

### DIFF
--- a/warehouse/static/js/warehouse/utils/webauthn.js
+++ b/warehouse/static/js/warehouse/utils/webauthn.js
@@ -30,18 +30,23 @@ const populateWebAuthnErrorList = (errors) => {
   });
 };
 
-const doWebAuthn = (buttonId, func) => {
+const doWebAuthn = (formId, func) => {
   if (!window.PublicKeyCredential) {
     return;
   }
 
-  const webAuthnButton = document.getElementById(buttonId);
-  if (webAuthnButton === null) {
+  const webAuthnForm = document.getElementById(formId);
+  if (webAuthnForm === null) {
     return null;
   }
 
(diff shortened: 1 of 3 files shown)


nlhkabu:
Thank you @woodruffw.  Ok on my side, but I will wait for @di to review for the JS :)

warehouse/static/js/warehouse/utils/webauthn.js
```diff
@@ -30,18 +30,23 @@ const populateWebAuthnErrorList = (errors) => {
...
+  if (webAuthnForm === null) {
     return null;
   }
 
+  const webAuthnButton = webAuthnForm.querySelector("button[type=submit]");
   webAuthnButton.disabled = false;
-  webAuthnButton.addEventListener("click", async () => { func(webAuthnButton.value); });
+
+  webAuthnForm.addEventListener("submit", async() => {
+    func(webAuthnButton.value);
```
webknjaz: This should probably be:
```suggestion
    await func(webAuthnButton.value);
```

nlhkabu: Thanks @webknjaz.
If you want to make a new PR with this update, please do so :)

webknjaz: Here you go: https://github.com/pypa/warehouse/pull/6608

nlhkabu: thanks!

PR #6585, opened 2019-09-05
6575 tab accessibility

### LINKED ISSUE #6575: Improve tab accessibility
For screenreader users:

Currently, on the **javascript** tabs component, there is no user feedback when a new tab is selected (and a new tab panel is displayed).

We should:

1. ~Change JS tabs to use buttons instead of links~*
1. Add `aria-selected=true` to the selected tab
1. Move the keyboard focus to the tab panel when a new tab is selected
1. Add an aria label to each tab explaining that the focus will be moved, e.g. `aria-label="focus will be moved to downloads table"

*Investigated: not currently possible, as we use the link `href` to define the tab destination

FYI @bmustill-rose

nlhkabu:
Closes #6575

### DIFF
--- a/warehouse/static/js/warehouse/controllers/project_tabs_controller.js
+++ b/warehouse/static/js/warehouse/controllers/project_tabs_controller.js
@@ -72,14 +72,20 @@ export default class extends Controller {
     content.style.display = "none";
     let contentId = content.getAttribute("id");
     this._getAllTabsForContentId(contentId)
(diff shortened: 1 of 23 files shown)


nlhkabu:
Need to fix up, as this commit should not be in this PR: https://github.com/pypa/warehouse/pull/6585/commits/46420fbe131c4925a92107daf780775764b84e2b

nlhkabu:
Closing in favour of https://github.com/pypa/warehouse/pull/6607

Issue #6551, opened 2019-08-30
Improve instructional text for API tokens


nlhkabu:
Based on feedback collected in user testing:

On the API token confirmation page, we should:

- Show the username/password instructions on the API token confirmation page
- Show an example .pypirc file, and explain where to put them for "entire account" tokens vs "scoped to project" tokens

On the `manage project > settings` page:

- Add a link back to the API token creation page

FYI @justinmayer

nlhkabu:
Example help text for "whole account" (user scoped) tokens:

![Screenshot from 2019-08-30 07-38-03](https://user-images.githubusercontent.com/3323703/63998523-3799bb00-caf9-11e9-8194-331bbc83b8e1.png)

For tokens scoped to a specific project, what would the `.pypirc` file look like and where would it be located?

I assume that users can create a `.pypirc` file in their project root directory, but I could not find this documented anywhere.

@theacodes @di @dstufft - any ideas?

woodruffw:
> For tokens scoped to a specific project, what would the `.pypirc` file look like and where would it be located?

Project-scoped tokens look and are configured identically to user-scoped tokens, so the `.pypirc` file should look and be placed identically. The major distinction between the two will probably be in whether people use `.pypirc` frequently for project-scoped tokens, given that they're mostly useful for CI/CD and platforms like Travis provide separate configuration mechanisms.

nlhkabu:
Just to be clear -> if I created a project-scoped token, then I would add the details of this token into a `.pypirc` file at `$HOME/.pypirc`.  Effectively, it means that I can only have one token on my local machine at any time? 

E.g. I *cannot* create `.pypirc` files that correspond to each project locally (either by naming them differently, or by placing them in the project root)?

woodruffw:
> Effectively, it means that I can only have one token on my local machine at any time?

Ah, not exactly. You would effectively have one "default" token. If you wanted to have multiple project-scoped tokens, you could do something like this:

```ini
# either your user-scoped token or your default project-scoped token
[pypi]
username = __token__
password = pypi-blah

# some other token
[whatever]
username = __token__
password = pypi-blarg
```

And then do `twine --repository whatever` to switch to the `whatever` token.

nlhkabu:
Thanks for your feedback @woodruffw.  I'll update the text based on your feedback and we can go from there :)

nlhkabu:
Ok, so second attempt:

## User scoped token instructions
[...]

PR #6586, opened 2019-09-05
Link to API token form from manage project page

nlhkabu:
Partially addresses #6551 

## With verified email
Screenshot from 2019-09-05 06-52-30

## Without verified email
Screenshot from 2019-09-05 06-54-11

### DIFF
--- a/warehouse/templates/manage/settings.html
+++ b/warehouse/templates/manage/settings.html
@@ -20,6 +20,19 @@
 {% block main %}
   <h2>Project settings</h2>
 
+  <h3>API tokens</h3>
+  <p>API tokens provide an alternative way to authenticate when uploading packages to PyPI.</p>
+  {% if user.has_primary_verified_email %}
+    <p class="no-bottom-padding">
+      <a href="{{ request.route_path('manage.account.token') }}" class="button button--primary">Create a token for {{ project.name }}</a>
+    </p>
+  {% else %}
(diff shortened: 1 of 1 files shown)


nlhkabu:
@woodruffw - is there a way that we can set up this URL so that when we land on the 'create token' page, the project is pre-selected in the form?  E.g. for this project ("clandestined") landing on the page, it would look like this:

Screenshot from 2019-09-05 07-00-06

woodruffw:
> is there a way that we can set up this URL so that when we land on the 'create token' page, the project is pre-selected in the form? E.g. for this project ("clandestined") landing on the page, it would look like this:

Hmm, I'm not sure that there is with the current form setup -- we handle defaulting to the user's previously selected scope on form error by checking `create_macaroon_form.token_scope.data`, but I don't believe there's a direct way to set that from `route_path`.

A more invasive option would be to add some kind of route parameter like `default_scope` that we could prioritize over the state of `create_macaroon_form`, but that adds complexity. Maybe @dstufft or @ewdurbin have opinions/insight here?

nlhkabu:
@woodruffw do you think we should merge this PR and open a separate ticket for pre-selecting the form?

woodruffw:
> do you think we should merge this PR and open a separate ticket for pre-selecting the form?

Yeah, I think that's a good solution for now.

PR #6608, opened 2019-09-10
Inject a missing await in webauthn.js

webknjaz:
Ref:
https://github.com/pypa/warehouse/pull/6580#discussion_r322854287

### DIFF
--- a/warehouse/static/js/warehouse/utils/webauthn.js
+++ b/warehouse/static/js/warehouse/utils/webauthn.js
@@ -44,7 +44,7 @@ const doWebAuthn = (formId, func) => {
   webAuthnButton.disabled = false;
 
   webAuthnForm.addEventListener("submit", async() => {
-    func(webAuthnButton.value);
+    await func(webAuthnButton.value);
     event.preventDefault();
   });
 };


woodruffw:
My JS isn't the strongest, but I don't think this is actually needed here -- we don't necessarily want to `await` on `func`, since `doWebAuthn` composes the totality of behavior in both `ProvisionWebAuthn` and `AuthenticateWebAuthn`.

It shouldn't hurt us behaviorally, but my understanding of `await` is that it shouldn't be used unless we actually *need* to capture the result of `func` in that scope.

di:
I think I agree w/ @woodruffw -- @webknjaz, could you add more details about why this necessary here?

webknjaz:
This may produce `Uncaught (in promise) Exception` if an error happens in that callback and that is hard to track if it's not awaited anywhere.
AFAIS all callbacks passed there are async functions.
I think it's a good practice to either always await all promises or attach `.catch()` error handlers.

webknjaz:
This is what I thought (at some point) was causing https://github.com/pypa/warehouse/issues/6050. Which is one example of "hard to trace"

woodruffw:
> This may produce `Uncaught (in promise) Exception` if an error happens in that callback and that is hard to track if it's not awaited anywhere.

That makes sense; I think we're now handling those cases explicitly with `await-then-catch` in each of the callbacks. So there *shouldn't* be any more uncaught exceptions in those promises and the current code *should* be fine, but it wouldn't really hurt to add the `await`.

webknjaz:
Only some places in those callbacks have error handlers, but they don't have their whole bodies wrapped with try/catch stuff. This issue will happen if for some reason (missing element) `document.getElementById("webauthn-provision-label").value` would raise an exception, for example. Or any other unprotected code statement.

woodruffw:
> This issue will happen if for some reason (missing element) `document.getElementById("webauthn-provision-label").value` would raise an exception, for example. Or any other unprotected code statement.

Right. That shouldn't happen given the element checks we perform, but that strikes me as reasonable and should help debugging if those IDs are ever changed. Adding `await` makes sense to me 👍

woodruffw:
This broke both provisioning and authentication; pushing a revert PR in a moment.

webknjaz:
Broke how?

di:
From private chats:

> webauthn login is not working for me: after completing username/password I click "authenticate with device" and it just dumps me back to the login page with a "Error: Invalid or expired two factor login."

webknjaz:
Weird

Issue #6255, opened 2019-07-25
Add caveats to macaroons for expiration (time) and version


alex:
**What's the problem this feature will solve?**

This will allow further attenuating the permissions granted by an API key

**Describe the solution you'd like**

Addition of two addition types of caveat, project version (for uploads) and time (expiry).

brainwane:
This is a fine idea to add onto future work for #994 -- thank you! I believe it's out of scope for [our current funding for security improvements to PyPI](https://pyfound.blogspot.com/2018/12/upcoming-pypi-improvements-for-2019.html), sorry to say.

woodruffw:
Another potentially useful caveat would be IP address/range, for future consideration.

brainwane:
@rcipkins -- how is this going?

rcipkins:
@brainwane I am almost finished, I just need to fix a couple things and add tests!

brainwane:
@rcipkins - Great to hear!

It's a good idea to push your branch to your GitHub fork and start a "work in progress" ("WIP") pull request. That way:

* in case your computer breaks or is lost/stolen, you haven't lost your progress
* other people can easily see that you're working on the issue, and avoid accidentally duplicating your effort
* it's easy to ask "am I on the right track?" questions about your general architectural approach, before spending time polishing stuff and fixing tests

To do this, push your branch to your fork, and create a pull request where the start of the PR title is "(WIP)".

Looking forward to seeing it!

rcipkins:
Done! Thank you so much!

woodruffw:
NB: This is addressed by #11122.

(We'll need separate UI work to make that caveat useable with user-minted macaroons, but that PR will add all of the backend logic needed.)

PR #6613, opened 2019-09-10
New Caveats #6255

rcipkins:
Added new caveats for new releases and expiration time of a macaroon.


woodruffw:
Whoops, requested the wrong reviewer -- that should be me.

woodruffw:
This is looking really good! I'll do another pass tomorrow.

dev/notdatadog.py
```diff
@@ -16,11 +16,10 @@
```
woodruffw: I think it'd be safe to revert this file.

dev/smtp.py
```diff
@@ -16,17 +16,15 @@
```
woodruffw: Ditto here.

docs/conf.py
```diff
@@ -28,11 +28,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
```
woodruffw: Ditto here.

tests/unit/manage/test_views.py
```diff
@@ -1465,9 +1465,11 @@ def test_default_response(self, monkeypatch):
         monkeypatch.setattr(views, "DeleteMacaroonForm", delete_macaroon_cls)
 
         project_names = [pretend.stub()]
```
woodruffw: I think this variable + associated monkeypatch (and all duplicates below) can be removed.

warehouse/macaroons/caveats.py
```diff
@@ -42,33 +48,75 @@ def verify_projects(self, projects):
             )
 
         project = self.verifier.context
-        if project.normalized_name in projects:
-            return True
 
+        for user_proj in projects:
```
woodruffw: I think we need to make this loop more flexible to handle API tokens that are already out there in the world: we should check if each element is a `str` -- if it is, then treat is as the normalized name. If it's a `dict`, then perform the `project-name` lookup.

warehouse/manage/forms.py
```diff
@@ -238,12 +251,59 @@ def validate_token_scope(self, field):
...
+        expiration = datetime.strftime(expiration, "%Y-%m-%dT%H:%M")
+
+    def validate(self):
+        res = super().validate()
+        if not isinstance(self.validated_scope, str):
+            self.validated_scope = {
+                "expiration": self.expiration.data,
+                "projects": [
+                    {
+                        "project-name": self.token_scope.data,
```
woodruffw: Could be misunderstanding here, but won't this embed the full `scope:` prefix into the `project-name`? I think we need to parse it out first.

woodruffw:
@ewdurbin @dstufft I believe this is good for an initial review!

Some notes:

* We should probably (better) figure out what the semantics of a release-limited token are -- the upload endpoint accepts one file at a time and creates the release on the first file, so creating a version-limited token will cause package upload to bomb out if the release is more than one file.

* Similarly, PyPI already imposes some release constraints: duplicate filenames can never be re-released. This partially obviates the general need for version-scoped tokens, although it doesn't cover the case where a malicious user uploads a new filename for the same release.

ewdurbin:
ref #6255

woodruffw:
#6935 supersedes this and includes additional caveat work; thanks @rcipkins!

PR #6665, opened 2019-09-19
warehouse: Add filter panel tab bindings

### LINKED ISSUE #6508: Make filter panel more accessible
On the search results page, we hide the classifier filters on smaller resolutions, revealing them in a panel.

We need to:

1. ensure that the keyboard focus is move to the panel, when it is opened by the user
2. ensure that the user cannot 'tab out' of the panel when it is open (e.g. users should be able to loop over the links in the panel

woodruffw:
cc @nlhkabu 

Closes #6508.

### DIFF
--- a/warehouse/static/js/warehouse/utils/bind-filter-keys.js
+++ b/warehouse/static/js/warehouse/utils/bind-filter-keys.js
@@ -19,15 +19,19 @@ export default () => {
     const filterPanel = document.querySelector(".filter-panel");
     keys[event.keyCode] = event.type === "keydown";
 
-    if (filterPanel && keys[9]) {
-      // `firstItem` assumes that the close button is the first tabbable element.
-      const firstItem = filterPanel.querySelector(".filter-panel__close");
+    if (filterPanel === null) {
+      return;
+    }
 
-      const getLastItem = () => filterPanel.querySelector(
-        ".accordion:last-child").classList.contains("accordion--closed")
-        ? filterPanel.querySelector(".accordion:last-child > .accordion__link")
-        : filterPanel.querySelector(".accordion:last-child li:last-child > input");
+    // `firstItem` assumes that the close button is the first tabbable element.
+    const firstItem = filterPanel.querySelector(".filter-panel__close");
 
+    const getLastItem = () => filterPanel.querySelector(
+      ".accordion:last-child").classList.contains("accordion--closed")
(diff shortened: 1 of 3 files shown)


warehouse/static/js/warehouse/utils/bind-filter-keys.js
```diff
@@ -0,0 +1,49 @@
...
+    keys[event.keyCode] = event.type === "keydown";
+
+    if (filterPanel && keys[9]) {
+      // `firstItem` assumes that the close button is the first tabbable element.
+      const firstItem = filterPanel.querySelector(".filter-panel__close");
+
+      const getLastItem = () => filterPanel.querySelector(
+        ".accordion:last-child").classList.contains("accordion--closed")
+        ? filterPanel.querySelector(".accordion:last-child > .accordion__link")
+        : filterPanel.querySelector(".accordion:last-child li:last-child > input");
```
woodruffw: @nlhkabu These selectors could use your 👀: they work for me, but I'm not sure if they're as efficient/semantic as they could be.

nlhkabu: Maybe a better way to do this would be to add a class `.filter-panel__last-item` to both the link and input?  my concern here is that if we add more content to the panel (e.g. anything that is not an accordion) we will mess up the script.

At least if we can keep the selectors specific to the `filter-panel` component, we can document this class in the SCSS file for future reference.  Wdyt?

nlhkabu:
Thanks @woodruffw!  Would you be able to add the ability to close the panel by pressing the `esc` key to this PR?

warehouse/static/js/warehouse/utils/bind-filter-keys.js
woodruffw: > Maybe a better way to do this would be to add a class `.filter-panel__last-item` to both the link and input? my concern here is that if we add more content to the panel (e.g. anything that is not an accordion) we will mess up the script.

Yep, I agree, but I'm not actually sure about the best way to do that here -- we generate the classifier tree with a recursive macro, and there might not be a clean way to determine the last child `input`.



> At least if we can keep the selectors specific to the `filter-panel` component, we can document this class in the SCSS file for future reference. Wdyt?

👍

woodruffw:
Okay, leaving the selectors as-is for now (see #6693). I've added support for `esc`.

PR #6684, opened 2019-09-22
Allow manylinux2014 wheel upload

mayeut:
manylinux2014 platform tag is official per PEP599

see also pypa/manylinux#338


mayeut:
@di,
I forced-pushed an updated commit with all architectures except `armhfp`.
Can you confirm the detection code in [PEP599](https://www.python.org/dev/peps/pep-0599/#platform-detection-for-installers) has been tested on all platforms ?
`linux-armhfp` won't be returned on some platforms (if not all, depending on previous answer)

Simple test I did:
```
docker pull arm32v7/centos:7
docker run --rm -it arm32v7/centos:7 /bin/bash
[root@36ef319dca55 /]# uname -a
Linux 36ef319dca55 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019 armv7l armv7l armv7l GNU/Linux
[root@36ef319dca55 /]# python
Python 2.7.5 (default, Oct 31 2018, 20:11:09) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.uname()
('Linux', '36ef319dca55', '4.9.184-linuxkit', '#1 SMP Tue Jul 2 22:58:16 UTC 2019', 'armv7l')
>>> from distutils.util import get_platform
>>> get_platform()
'linux-armv7l'
```

This also means that there is no difference between armhfp and armsfp (the later being uncommon these days), what shall I do ?

di:
@mayeut It was not tested it seems. I've made https://github.com/python/peps/pull/1181 to update the PEP.

di:
@auvipy, while I appreciate your review, if you don't see any issues would you mind reserving approvals for maintainers? Otherwise it just generates noise. Thanks!

mayeut:
@di, I updated the PR with `armv7l`. Given the PR you made for the PEP, I guess you've made up your mind on the `soft-float` vs `hard-float` ABI incompatibility and that we should consider `glibc 2.17+`  and `armv7l` to always be `hard-float` ABI then. Thanks.

PR #6711, opened 2019-09-25
Use the same message for "too many failed logins" error

atugushev:
These duplicated messages I've caught in Weblate and this PR aimed to reduce translators' work.


ewdurbin:
Thank you @atugushev!

atugushev:
Thanks for merging so quick!

Issue #6088, opened 2019-06-25
Should we autofocus the search bar on PyPI homepage?


nlhkabu:
# Should we autofocus the search bar on PyPI homepage?

We have had many requests (and some PRs) from PyPI users to autofocus the search bar on the PyPI homepage.

As part of the work funded by the recent [OTF grant](https://pyfound.blogspot.com/2019/03/commencing-security-accessibility-and.html), I have been researching whether or not this is a good decision to make from an accessibility perspective. [Skip to conclusion](#conclusion).

## Accessibility guidelines

To ensure that PyPI works for all users, our objective is to make PyPI [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) compliant.

Regarding autofocus, the relevant WCAG guideline is: 

> “Guideline 2.4 Navigable: Provide ways to help users navigate, find content, and determine where they are.”

and criterion:

> 2.4.3 Focus Order: If a Web page can be navigated sequentially and the navigation sequences affect meaning or operation, focusable components receive focus in an order that preserves meaning and operability. (Level A).  

The documentation for this criterion [can be found here](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html), while the [failure information is found here](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F44).

## Meeting the guidelines

The key point in the criterion is that information should be presented:

> in an order that is consistent with the meaning of the content

In the case of the PyPI homepage, it *could* be argued that as a primary task on the page, adding autofocus to the search is consistent with the meaning of the content.  However, this would come down to the interpretation of the accessibility auditor.

After more digging, I found a recommendation in the [WIA Aria best practices document](https://www.w3.org/TR/wai-aria-practices/#kbd_focus_discernable_predictable) that provides clearer guidelines:

> * Do not set initial focus when the page loads except in cases where:
>   * The page offers a single, primary function that nearly all users employ immediately after page load.
>   * Any given user is likely to use the page often.

## Assessing the recommendation

To assess this recommendation it is useful to look at the way that the PyPI homepage is currently used by our community.

Notes:
- **Data source**: Google Analytics
- **Dates studied**: 1 June 2018 - May 31 2019 (last 12 full months)

###  Does the page offers a single, primary function that nearly all users employ immediately after page load?

[...]

Issue #6040, opened 2019-06-18
a11y: Login view


woodruffw:
Issues in `/account/login`:

* Explicit positive `tabindex` values [should not be used](https://www.scottohara.me/blog/2019/05/25/tabindex.html) in the login form's inputs. Omitting `tabindex` entirely should be fine in this case.

anowlcalledjosh:
This means that the password field now takes two <kbd>Tab</kbd>-presses to reach from the username field rather than one, which breaks at least one popular password manager.

If the accessibility issue is more important than avoiding some manual configuration of password managers, then fine (I'm not qualified to make that decision), but the original change to add `tabindex`es was intentional – see #2824, #2917.

di:
This has actually been bugging me too. I think we can fix it without adding tab indexes by moving the "Show password" box under the password field.

@nlhkabu @woodruffw Thoughts?

woodruffw:
> I think we can fix it without adding tab indexes by moving the "Show password" box under the password field.

Yep, I think this would be a good solution.

PR #6721, opened 2019-09-27
Tabindex

saippuakauppias:
For navigation by keyboard ( https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex )

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -211,7 +211,7 @@
 
             <form class="search-form search-form--primary" action="{{ request.route_path('search') }}" role="search">
               <label for="search" class="sr-only">{% trans %}Search PyPI{% endtrans %}</label>
-              <input id="search" class="search-form__search" type="text" name="q" placeholder="{% trans %}Search projects{% endtrans %}" value="{{ term }}" autocomplete="off" autocapitalize="off" spellcheck="false">
+              <input id="search" class="search-form__search" type="text" name="q" placeholder="{% trans %}Search projects{% endtrans %}" value="{{ term }}" autocomplete="off" autocapitalize="off" spellcheck="false" tabindex="1">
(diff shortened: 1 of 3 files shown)


di:
I'll let @nlhkabu review, but I should point out that we explicitly removed `tabindex` attributes from the login form in #6040 due to accessibility issues.

saippuakauppias:
I think, previous issue with login form breaks using tab-autocompletion, right?
In this case autocomplete in search field already off'ed

saippuakauppias:
@nlhkabu, @di any info?..

saippuakauppias:
@di, @nlhkabu, @ewdurbin any info?..

saippuakauppias:
@di, @nlhkabu, @ewdurbin, any info?..

This is useful for navigating through the page by pressing the tab key. Now to get to the search bar - you need to press several times. Although on the main page of the main focus and idea in the search box - it comes first. And you do not want to use the mouse to click on it. 
This will not break the availability, because all the next clicks on the "tab" will go like this - from the beginning.

I think this is a very useful suggestion that will shorten a lot of time for everyone who uses the keyboard navigation.

Translated with www.DeepL.com/Translator (free version)

di:
I appreciate your contribution here @saippuakauppias, unfortunately both changing the tab index and auto-focusing the search field actually make PyPI less accessible overall based on WCAG and ARIA guidelines. There's lots more details in #6088 if you're interested.

Also, I should mention that PyPI implements the [OpenSearch description format](https://developer.mozilla.org/en-US/docs/Web/OpenSearch), so if you're accessing PyPI from a browser that supports this standard (like Chrome or Firefox) you can search PyPI without even visiting pypi.org first: https://twitter.com/di_codes/status/1058101995918319617

Issue #6722, opened 2019-09-27
Move "show password" checkbox on login form


di:
From #6040:

@anowlcalledjosh:
> This means that the password field now takes two <kbd>Tab</kbd>-presses to reach from the username field rather than one, which breaks at least one popular password manager.
>
> If the accessibility issue is more important than avoiding some manual configuration of password managers, then fine (I'm not qualified to make that decision), but the original change to add `tabindex`es was intentional – see #2824, #2917.

@di:
> This has actually been bugging me too. I think we can fix it without adding tab indexes by moving the "Show password" box under the password field.
> 
> @nlhkabu @woodruffw Thoughts?

@woodruffw:
> Yep, I think this would be a good solution.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

singleparadox:
Hello, I'd like to work on this. Can I? I'll start after I get home from work...

woodruffw:
Taking a stab at this.

di:
@woodruffw Looks like we might already have a PR: https://github.com/pypa/warehouse/pull/6755

woodruffw:
Whoops! Ignore me then 😅

di:
Although heads up @ddaan, that doesn't quite do what was described in this issue. We don't want to use `tabindex`.

woodruffw:
Dropping a non-`tabindex` diff for reference:

```diff
diff --git a/warehouse/templates/accounts/login.html b/warehouse/templates/accounts/login.html
index 56a5e327..ab676b3a 100644
--- a/warehouse/templates/accounts/login.html
+++ b/warehouse/templates/accounts/login.html
@@ -64,17 +64,13 @@
         </div>
 
         <div data-controller="password" class="form-group">
-          <div class="split-layout">
+          <div>
             <label for="password" class="form-group__label">
[...]

PR #6755, opened 2019-10-04
#6722 fix tab ordering

ddaan:


di:
Closing this since it's not quite what we were looking for in #6722, but thanks for the PR!

PR #6765, opened 2019-10-05
tests: assert jinja templates are all rendered correctly.

### LINKED ISSUE #6634: Missing test for project detail page (template failure)
## Problem
When updating the templates for the project detail page (e.g. https://pypi.org/project/python-ecobee-api/) I was getting a Jinja template exception from the project data include (`warehouse/templates/includes/packaging/project-data.html`)

This was *not* being picked up by running the PyPI unit tests.

## Proposed solution
We should add a unit test to check the appropriate rendering of this page

sk1u:
Hi all, I added a unit test that should address https://github.com/pypa/warehouse/issues/6634

I opted to try rendering all templates instead of just the specified one.

Fixes #6634.


tests/functional/test_templates.py
```diff
@@ -69,3 +69,55 @@ def test_templates_for_empty_titles():
...
+            "json": "warehouse.filters:tojson",
+            "camoify": "warehouse.filters:camoify",
+            "shorten_number": "warehouse.filters:shorten_number",
+            "urlparse": "warehouse.filters:urlparse",
+            "contains_valid_uris": "warehouse.filters:contains_valid_uris",
+            "format_package_type": "warehouse.filters:format_package_type",
+            "parse_version": "warehouse.filters:parse_version",
+            "localize_datetime": "warehouse.filters:localize_datetime",
+        }
+    )
```
di: Perhaps we can automatically generate this list?

sk1u: It makes sense, would you have any suggestions on how to do that?

di: Yeah, might be a bit much for this PR. At the very least, can we not duplicate this between the two tests in this file?

sk1u:
Is there anything else I can improve for this? :)

di:
Thanks!

PR #6767, opened 2019-10-05
Fixes #6691 - Typo 'Top Storage users' in Stats page

### LINKED ISSUE #6691: Typo "Top Storage users" in Stats page
The [Stats page, line 30](https://github.com/pypa/warehouse/blob/97c4a5124e86119eabe57c7b896a34d2732dc1d0/warehouse/templates/pages/stats.html#L30) (e.g. [PyPI's Stats page](https://pypi.org/stats/)) has the heading text **Top storage users** but it seems to be related to **uses**, rather than **users**. Maybe it is a typo?

raulgspan:
Fixes #6691 - Changed to 'Top projects by total package size', corrected possessive apostrophe, and added down arrow to indicate sort order


rffontenelle:
Can you edit the description adding a "Fixes #`<number of the issue>`"? This will 1) make a link from your PR to the issue in opened, and to 2) close that issue when/if the PR is accepted ;)

rffontenelle:
Also it seems Travis complained about your down arrow:
```./warehouse/templates/pages/stats.html:43:81: Error: Entity References: Use the unicode equivalent instead: Change "&darr;" to "↓".```

Can you edit your commit with the suggested change?

**Comment by brainwane** [comment]
And @raulgspan -- just so you know, you don't need to close the old PR and open a new one. You can [update your pull request](https://zulip.readthedocs.io/en/latest/git/pull-requests.html#update-a-pull-request) and we recommend you do that (instead of closing the old PR and opening a new one) because then it's easier to see the history of what you've done, to to check that you've responded to previous reviewers' comments, and to preserve links we've mentioned (in comments) with other issues and PRs.

**Comment by raulgspan** [comment]
Thanks @rffontenelle -- updated the PR description.

**Comment by rffontenelle** [comment]
> Thanks @rffontenelle -- updated the PR description.

Er.. nope. You updated the summary, but you should update the description with the "fixes" statement.

Also, consider fixing the commit as explained above. Might be a good exercise. Let us know if you get stuck.

**Comment by raulgspan** [comment]
> Er.. nope. You updated the summary, but you should update the description with the "fixes" statement.

Just want to make sure, by "description" you mean the first comment, right? Correct now?

> Also, consider fixing the commit as explained above. Might be a good exercise. Let us know if you get stuck.

Is this referring to fixing to use the unicode equivalent? See second commit in the PR.

Thanks--

**Comment by rffontenelle** [comment]
>
> Just want to make sure, by "description" you mean the first comment,
> right? Correct now?
>
>
Exactly. Now it is correct. Notice in the mentioned issue how your PR is
now mentioned as a proposal that fixes it.

Is this referring to fixing to use the unicode equivalent? See second
> commit in the PR.
>

Oh, I didn't see it. Awesome. Thanks. Now it is up to a
developer/maintainer to commit if accepted.

File: warehouse/templates/pages/stats.html [hunk 1]
```diff
@@ -40,7 +40,7 @@ <h2>{% trans %}Top storage users{% endtrans %}</h2>
       <thead>
         <tr>
           <th scope="col">{% trans %}Project name{% endtrans %}</th>
-          <th scope="col">{% trans %}Sum of release files (bytes){% endtrans %}</th>
+          <th scope="col">{% trans %}Sum of release files (bytes){% endtrans %} ↓</th>
```
yeraydiazdiaz: Nit: I'm not sure about this additional `↓`, it might convey to the user that you can click on it to reverse the order.

di: Agreed!

raulgspan: True, but I think it's preferable to no indicator of sort order... other ideas?

yeraydiazdiaz: Personally I think the copy above is quite clear:

> Top storage users
> Here is a list of the top 100 packages based on sum of their packages sizes (in bytes). 

If the table was presented on its own I would agree with you 🙂

raulgspan: Who reads copy? :) Will remove.

brainwane:
Thanks, @raulgspan! 

Are you interested in another small issue to fix?

PR #6768, opened 2019-10-06
Make confirm modal case-insensitive

### LINKED ISSUE #6676: "Confirm username" modal is case-sensitive
When I try to turn off 2FA on my pypi account, after I am prompted to enter my username, the option to "remove application" on the warehouse web interface is grayed out. Unfortunately since I am changing phones today that means that I'll be locked out of my pypi account once my old phone is gone. It would be great if one of the warehouse maintainers could work with me to restore access to my account and re-enable 2FA. My username is ngoldbaum on freenode and I'm happy to chat about this in #pypa or #pypa-dev.

calvin:
Close #6676


ngoldbaum:
Thanks for fixing this, I can confirm that this was very confusing for me :)

PR #6805, opened 2019-10-10
Added ISO 8601 compliant upload time field to JSON API

### LINKED ISSUE #4751: JSON API omits timezone and microseconds from upload_time

AWhetter:
Closes #4751

I'm not sure if `upload_time_iso_8601` is the best name for the field. It's descriptive at least...

This is an alternative to #4752, closes #4752.


di:
I think this is probably the best solution to #4751.

Could you also add a `upload_time_iso_8601` field to the `release_urls` method of the XML-RPC API as well, just for consistency? (even though it will just have the same value)

AWhetter:
Done! I've added the field to the XML-RPC API as well.

di:
@AWhetter One last thing, let's also remove `upload_time` from the documentation.

AWhetter:
Done!

PR #6806, opened 2019-10-10
Autolabel issues created from i18n template

webknjaz:


webknjaz:
@nlhkabu I noticed that you apply these labels manually but they actually can be pre-filled.

PR #6820, opened 2019-10-12
Enable es/fr locales

di:
* Adds French and Spanish as known locales, including `.mo` files.
* Updates `build-mos` to only consider locales in `KNOWN_LOCALES`
* Adds docs on adding new locales.


ewdurbin:
Looks like there is some churn generated in the `.po` files due to our `update-po` target configuration. Do we want to isolate `compile-po` from `update-po` (as currently called by `build-mos`)? My only concern is how Weblate will handle the changes.

It might be preemptive, so probably best to merge this and see. 👨‍🔬

PR #6829, opened 2019-10-13
Implement Google Cloud Storage FileStorage backend

di:
This PR allows us to target multiple file storage services by simply configuring `FILES_BACKEND` to use different file storage backend services. All new services are drop-in replacements for the existing `warehouse.packaging.services.S3FileStorage` service:

* `warehouse.packaging.services.GCSFileStorage`: A service which targets [Google Cloud Storage](https://cloud.google.com/storage/)
* ~~`warehouse.packaging.services.DualFileStorage`: A service which targets S3 as a primary storage backend, and GCS as a secondary backend. Calling `DualFileStorage.store()` will replicate a file across both the primary and secondary storage services.~~

~~For now, the secondary storage service for the `DualFileStorage` is allowed to silently fail: my thought is that this should never get in the way of a successful upload.~~

~~The idea is that this enables us to do the following~~:
~~1. Switch from `S3FileSorage` to `DualFileStorage`~~
2. Backfill our GCS bucket with everything from before this switch
2a. Update Fastly to check GCS first before falling back to S3 (https://github.com/python/pypi-infra/pull/57)
3. Configure Fastly to to use GCS instead to take advantage of the [integration discount](https://docs.fastly.com/en/guides/discounted-egress-from-google)
4. Switch from ~~`DualFileStorage`~~ `S3FileStorage` to `GCSFileStorage`~~
5. Perform a final sync to pull in files uploaded to `S3FileStorage` in the interim.

Note that I didn't implement `.get()` for `GCSFileStorage` ~~or `DualFileStorage`~~. This function is not actually used in production due to our Fastly configuration interacting directly with our storage backend.

~~Also, one caveat here: I didn't add a way to configure different bucket names for different services when using the `DualFileStorage`: right now it assumes that the S3 and GCS bucket names are the same. I figured that our bucket names are unique enough that it shouldn't be an issue to have the same bucket name in both backends.~~

Note: ninja edits by (EWDurbin)

### DIFF
--- a/warehouse/packaging/services.py
+++ b/warehouse/packaging/services.py
@@ -84,19 +84,10 @@ class LocalDocsStorage:
             pass
 
 
-@implementer(IFileStorage)
-class S3FileStorage:
+class GenericFileStorage:
     def __init__(self, bucket, *, prefix=None):
         self.bucket = bucket
         self.prefix = prefix
-
-    @classmethod
-    def create_service(cls, context, request):
-        session = request.find_service(name="aws.session")
-        s3 = session.resource("s3")
-        bucket = s3.Bucket(request.registry.settings["files.bucket"])
-        prefix = request.registry.settings.get("files.prefix")
-        return cls(bucket, prefix=prefix)
 
     def _get_path(self, path):
         # Legacy paths will have a first directory of something like 2.7, we
@@ -112,7 +103,21 @@ class S3FileStorage:
 
         return path
 
+
+@implementer(IFileStorage)
+class S3FileStorage(GenericFileStorage):
+    @classmethod
+    def create_service(cls, context, request):
+        session = request.find_service(name="aws.session")
+        s3 = session.resource("s3")
+        bucket = s3.Bucket(request.registry.settings["files.bucket"])
+        prefix = request.registry.settings.get("files.prefix")
+        return cls(bucket, prefix=prefix)
+
     def get(self, path):
+        # Note: this is not actually used in production, instead our CDN is
+        # configured to connect directly to our storage bucket. See:
+        # https://github.com/python/pypi-infra/blob/master/terraform/file-hosting/vcl/main.vcl
         try:
             return self.bucket.Object(self._get_path(path)).get()["Body"]
         except botocore.exceptions.ClientError as exc:
--- a/warehouse/packaging/services.py
+++ b/warehouse/packaging/services.py
@@ -16,6 +16,7 @@ import warnings
 
 import botocore.exceptions
 
+from pyramid_services import ProxyFactory
 from zope.interface import implementer
 
 from warehouse.packaging.interfaces import IDocsStorage, IFileStorage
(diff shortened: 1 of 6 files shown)


warehouse/packaging/services.py
```diff
@@ -161,3 +167,70 @@ def remove_by_prefix(self, prefix):
...
+        raise NotImplementedError
+
+    def store(self, path, file_path, *, meta=None):
+        path = self._get_path(path)
+        blob = self.bucket.blob(path)
+        blob.upload_from_filename(file_path)
+
+        if meta is not None:
+            blob.metadata = meta
+            blob.patch()
```
yeraydiazdiaz: Does this need to be a separate call or can it be added to the `blob` created above?

I can't tell from the docs but it might be worth avoiding a second request.

warehouse/packaging/services.py
```diff
@@ -161,3 +167,70 @@ def remove_by_prefix(self, prefix):
...
+        secondary_service = self.request.find_service(name="secondary-file-backend")
+
+        # This must succeed
+        primary_service.store(path, file_path, meta=meta)
+
+        # This can fail
+        try:
+            secondary_service.store(path, file_path, meta=meta)
+        except Exception:
+            pass
```
yeraydiazdiaz: Is it worth logging this exception for visibility?

warehouse/packaging/services.py
```diff
@@ -161,3 +167,70 @@ def remove_by_prefix(self, prefix):
...
+        bucket = storage_client.get_bucket(bucket_name)
+        prefix = request.registry.settings.get("files.prefix")
+
+        return cls(bucket, prefix=prefix)
+
+    def get(self, path):
+        # Note: this is not actually used in production, instead our CDN is
+        # configured to connect directly to our storage bucket. See:
+        # https://github.com/python/pypi-infra/blob/master/terraform/file-hosting/vcl/main.vcl
+        raise NotImplementedError
```
yeraydiazdiaz: Maybe it'd be worth adding a message similar to the note above? That way if it is raised we'd know why in the traceback without looking at the code

warehouse/packaging/services.py
di: As far as I can tell, this is the only way to do it.

warehouse/packaging/services.py
di: Yeah, likely so, I was expecting @ewdurbin to have some thoughts here as well.

warehouse/packaging/services.py
di: Eh, I'm not so sure it's worth it. This is the only `NotImplementedError` in the entire codebase, and it's pretty unlikely it will get raised.

yeraydiazdiaz: Fair enough :)

warehouse/packaging/services.py
ewdurbin: Probably best to send this to sentry and continue on. That should be doable by issuing at the ERROR level and letting the [logging config](https://github.com/pypa/warehouse/blob/dc120b9c0f8fba9dc98809e68e1b56a1d1bda5e6/warehouse/logging.py#L68) pick it up.

You may also be able to manually create an event by tapping into the registry and finding the sentry client there?

ewdurbin:
> Also, one caveat here: I didn't add a way to configure different bucket names for different services when using the `DualFileStorage`: right now it assumes that the S3 and GCS bucket names are the same. I figured that our bucket names are unique enough that it shouldn't be an issue to have the same bucket name in both backends.

Confirmed that we have the same bucket names on S3 and GCS

warehouse/packaging/services.py
ewdurbin: A quick test shows that

```python
blob = self.bucket.blob(path)
blob.metadata = meta
blob.upload_from_filename(file_path)
```

Correctly sets the metadata without having to make two calls!

di: Does it work when `meta` is `None` as well?

ewdurbin: in the patch I'm about to send, I still do a None check :-D

ewdurbin:
@di thoughts on dropping the DualFileStorage service given the migration path provided by https://github.com/python/pypi-infra/pull/57

di:
@ewdurbin Yep, don't see why we'd need it anymore. 👍

ewdurbin:
Thanks @di for the implementation and @yeraydiazdiaz for the reviews!

PR #6833, opened 2019-10-14
add option to display primary email on public profile

### LINKED ISSUE #5944: Add option to display primary email on public profile
Currently there's no way to publicly display an email address as a point of contact for a PyPI account. Instead, maintainer/author emails for a user's packages must be used instead.

Similar to Github, on the ["Account Settings"](https://pypi.org/manage/account/) page, we should have an option such as:

> ☑️**Keep my email address private**
> We won't show your primary email on your public PyPI profile page.

This should default to "True" for all new and existing users.

---

rascalking:
The basics are there, including tests, but i'm not sure if we want to wrap the email address in a mailto:, or how we want to handle emails that are too long.

Fixes #5944.


rascalking:
Sorry, this is for #5944.

rascalking:
Also, I think I'm supposed to run `make build-mos` to update everything after adding a new string that needs translation, but it looks like there's some other strings getting changed in the diffs as well.  Just wanted to make sure that's to be expected before committing and pushing that.

warehouse/templates/manage/account.html
```diff
@@ -300,6 +300,21 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
             Displayed on your <a href="{{ href }}">public profile</a>
           {% endtrans %}
         </p>
+        <label class="form-group__label" for="is_email_private">
+          {% trans %}️Keep my email address private{% endtrans %}
```
di:
```suggestion
          {% trans %}️Email privacy{% endtrans %}
```

warehouse/templates/manage/account.html
```diff
@@ -300,6 +300,21 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
...
+          {% if save_account_form.name.flags.required %}
+          <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+          {% endif %}
+          {{ save_account_form.is_email_private(class_="form-group__input", **{"aria-describedby":"is-email-private-errors"}) }}
```
di: Let's move this out of the label and into a `<p class="form-group__text">` with some text before it like "Keep my email address private".

warehouse/templates/manage/account.html
```diff
@@ -300,6 +300,21 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
...
+        </label>
+        <div id="is-email-private-errors">
+          {{ field_errors(save_account_form.is_email_private) }}
+        </div>
+        <p class="form-group__help-text">
+          {% trans trimmed href=request.route_path('accounts.profile', username=user.username) %}
+            We won't show your primary email on your <a href="{{ href }}">public profile</a>
```
di:
```suggestion
            If checked, your primary email won't be displayed on your <a href="{{ href }}">public profile</a>
```

warehouse/templates/manage/account.html
```diff
@@ -300,6 +300,21 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
             Displayed on your <a href="{{ href }}">public profile</a>
           {% endtrans %}
         </p>
+        <label class="form-group__label" for="is_email_private">
+          {% trans %}️Keep my email address private{% endtrans %}
+          {% if save_account_form.name.flags.required %}
```
di: This seems to be from the `name` field before this field.

di:
> or how we want to handle emails that are too long.

I don't think you need to handle this in this PR, I created #6880 to capture this.

warehouse/manage/views.py
```diff
@@ -136,7 +139,7 @@ def default_response(self):
     def manage_account(self):
         return self.default_response
 
-    @view_config(request_method="POST", request_param=SaveAccountForm.__params__)
+    @view_config(request_method="POST", request_param=["name"])
```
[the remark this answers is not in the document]
rascalking: The trick is that an unchecked checkbox means that the field isn't sent at all, leading to view routing failing when people try to make their emails public when we include `is_email_private` in the request params.

warehouse/migrations/versions/2db6fb5e2280_add_user_is_email_private.py
```diff
@@ -0,0 +1,54 @@
...
+down_revision = "ee4c59b2ef3a"
+
+# Note: It is VERY important to ensure that a migration does not lock for a
+#       long period of time and to ensure that each individual migration does
+#       not break compatibility with the *previous* version of the code base.
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
```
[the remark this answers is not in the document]
rascalking: Yup.

tests/unit/manage/test_views.py
```diff
@@ -147,7 +150,7 @@ def test_save_account(self, monkeypatch):
         update_user = pretend.call_recorder(lambda *a, **kw: None)
         user_service = pretend.stub(update_user=update_user)
         request = pretend.stub(
-            POST={"name": "new name"},
+            POST={"name": "new name", "is_private_email": "y"},
```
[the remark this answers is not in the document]
rascalking: That is the value that will be in the post request.  I'm not supplying a value to wtforms anywhere, so I'm guessing thats its default for checkboxes.

tests/unit/manage/test_views.py
```diff
@@ -172,7 +175,7 @@ def test_save_account_validation_fails(self, monkeypatch):
...
```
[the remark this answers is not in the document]
rascalking: Ditto.

warehouse/templates/accounts/profile.html
```diff
@@ -32,6 +32,13 @@ <h1 class="author-profile__name">{{ user.name }}</h1>
               <span class="sr-only">{% trans %}Username{% endtrans %}</span>
               &nbsp;&nbsp;{{ user.username }}
             </p>
+            {% if not user.is_email_private %}
+            <p>
+              <i class="fa fa-envelope" aria-hidden="true"></i>
+              <span class="sr-only">{% trans %}Email{% endtrans %}</span>
+              &nbsp;&nbsp;{{ user.primary_email.email }}
```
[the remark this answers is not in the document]
rascalking: Yup, fixed.

warehouse/templates/manage/account.html
rascalking: Fixed.

warehouse/templates/manage/account.html
rascalking: Done.

warehouse/templates/manage/account.html
rascalking: Done.

warehouse/templates/manage/account.html
rascalking: Yup, cut-n-paste mixup.  Thanks for the catch.

warehouse/manage/views.py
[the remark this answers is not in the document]
rascalking: Oh, just occurred to me, is the convention that `__params__` contains just the things we want to match with `request_param`?  In which case I should remove `is_email_private` from `__params__` instead, right?

rascalking:
As for more tests, I'm happy to, but I didn't see any existing tests that let the views run and mutate db entries, so I wasn't sure where it went.  The stuff in `tests/unit/manage/test_views.py` looks like it's going out of its way to avoid actually letting the view execute, so it seemed like the wrong place.  Maybe somewhere under `tests/functional`?

rascalking:
> > or how we want to handle emails that are too long.
> 
> I don't think you need to handle this in this PR, I created #6880 to capture this.

Thanks.

rascalking:
I took a stab at a test verifying that it makes the right changes to the `User` object.  This is my first time with pyramid...not sure if there's a cleaner way to call the view that would be preferrable to what I've hacked up so far.

warehouse/manage/views.py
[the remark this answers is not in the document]
di: It's just used for routing, so yes, we can just remove it from `__params__` as long as we don't have another route that uses the `name` parameter (which we don't).

warehouse/manage/forms.py
```diff
@@ -63,8 +63,9 @@ class ChangeRoleForm(RoleNameMixin, forms.Form):
 
 class SaveAccountForm(forms.Form):
 
-    __params__ = ["name"]
+    __params__ = ["is_email_private", "name"]
```
di:
```suggestion
    __params__ = ["name"]
```

warehouse/manage/views.py
di:
```suggestion
    @view_config(request_method="POST", request_param=SaveAccountForm.__params__)
```

ewdurbin:
Just a couple thoughts on this implementation:

- Rather than displaying the Primary email address, we should probably allow a user to select any Verified email address. I know for instance I would prefer to publicize my Verified @python.org email address, but prefer to keep a separate personal email address as my Primary.
- If possible, only displaying the email address to logged in users might be a good idea? I think the concern there would be for caching user pages.

warehouse/manage/views.py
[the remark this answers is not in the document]
rascalking: Fixed.

rascalking:
> I also decided in #6880 that the behavior for a long email should be different (it should truncate with an ellipsis), so we should address that in this PR.

I've got this set up, but I'm noticing two things:
* the trailing envelope icon that seems to get added to all mailto links gets ellipsized along with the rest of the email address
  - i hadn't noticed that gets added before...do we want to remove the leading icon on the left?
* when it needs to be ellipsized, the email address gets bumped down to the line below the leading envelope icon.  my css isn't all that good, i'm not sure how to fix that.

rascalking:
> Just a couple thoughts on this implementation:
> 
>     * Rather than displaying the Primary email address, we should probably allow a user to select any Verified email address. I know for instance I would prefer to publicize my Verified @python.org email address, but prefer to keep a separate personal email address as my Primary.

Sounds reasonable.  Default to Primary?

>     * If possible, only displaying the email address to logged in users might be a good idea? I think the concern there would be for caching user pages.

With the proviso that I don't know the codebase well, it doesn't look like there's a way to indicate that the user profile page should only be cached, and served from cache, for non-logged-in users.  If that's the case, is it worth giving up the cache in order to hide the email from non-logged-in folks?

di:
> Sounds reasonable. Default to Primary?

I think it'd be reasonable to just default to "not displayed" and force the choice when enabling.

> With the proviso that I don't know the codebase well, it doesn't look like there's a way to indicate that the user profile page should only be cached, and served from cache, for non-logged-in users. If that's the case, is it worth giving up the cache in order to hide the email from non-logged-in folks?

This would need to be done via a Client Side Include. An example would be the text on the profile page when the user has no packages, which changes for logged-in users (to do something different on their own profile):

Default, wrapped in a `csi` block:
https://github.com/pypa/warehouse/blob/9000661ff05e6d7f2cb139abb0ebe55844b3ae90/warehouse/templates/accounts/profile.html#L74-L78

The route:
https://github.com/pypa/warehouse/blob/92b28e69d2cb9af3456d2be047610ec955bc3bc5/warehouse/accounts/views.py#L686-L694

The template that replaces it:
https://github.com/pypa/warehouse/blob/92b28e69d2cb9af3456d2be047610ec955bc3bc5/warehouse/templates/includes/accounts/profile-callout.html#L15-L25

rascalking:
> > Sounds reasonable. Default to Primary?
> 
> I think it'd be reasonable to just default to "not displayed" and force the choice when enabling.

Oh, yeah, sorry if that was unclear.  I meant once you'd checked the "make my email public" checkbox.

Thanks for the explainer on client side includes!

di:
Looks like this is still missing the ability to set a non-primary verified email address as the public address.

I think the checkbox should probably become a dropdown with the ability to select either the default "Not displayed" or any primary address.

rascalking:
Yeah, I was still waffling about the UI and backend options.  Would you recommend adding a `public` column to the email table, or switching the `User.is_email_private` column to a nullable `User.public_email` column instead?

di:
I think it might make the most sense to handle it just like we do for primary email addresses now: a `public` column to the email table.

rascalking:
I fixed the alembic error, but the remaining errors seem to be something to do with the Travis setup, and not something related to my change.   Let me know if that's not the case?

di:
@rascalking Fixed in #7278 and CI is now passing here.

rascalking:
I think this is all set, let me know if I missed anything.

di:
Thanks for your patience here @rascalking!

rascalking:
Thanks for all the help!

PR #6852, opened 2019-10-16
Fix compile pot

### LINKED ISSUE #6850: Missing translation strings in Weblate
## Bad source in Weblate

There's parts of the website that aren't present in Weblate. I guess that's because they were marked translatable after the initial upload.
So we need to collect new strings and sync them there.

photo5278686657471687796

ewdurbin:
Loads our custom Jinja2 extension into pybabel's jinja2 environment to ensure all strings are marked for translation.

Resolves #6850.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ stop:
 	docker-compose down -v
 
 compile-pot: .state/env/pyvenv.cfg
-	$(BINDIR)/pybabel extract \
+	PYTHONPATH=$(PWD) $(BINDIR)/pybabel extract \
 		-F babel.cfg \
 		--copyright-holder="PyPA" \
 		--msgid-bugs-address="https://github.com/pypa/warehouse/issues/new" \
--- a/babel.cfg
+++ b/babel.cfg
@@ -1,5 +1,5 @@
 [python: **.py]
 [jinja2: **.html]
 encoding = utf-8
-extensions=jinja2.ext.autoescape,jinja2.ext.with_
+extensions=jinja2.ext.autoescape,jinja2.ext.with_,warehouse.utils.html:ClientSideIncludeExtension
 silent=False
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -175,7 +175,7 @@ stop:
 	docker-compose down -v
 
 compile-pot: .state/env/pyvenv.cfg
-	$(BINDIR)/pybabel extract \
+	PYTHONPATH=$(PWD) $(BINDIR)/pybabel extract \
```
webknjaz: I was about to send exactly the same patch except also adding this to targets `init-po`, `update-po` and `compile-po`.

babel.cfg
```diff
@@ -1,5 +1,5 @@
 [python: **.py]
 [jinja2: **.html]
 encoding = utf-8
-extensions=jinja2.ext.autoescape,jinja2.ext.with_
+extensions=jinja2.ext.autoescape,jinja2.ext.with_,warehouse.utils.html:ClientSideIncludeExtension
```
webknjaz: I didn't check out your branch but I came up with exactly the same solution so I guess it works. But I think you should do https://github.com/pypa/warehouse/pull/6852/files#r335553211 before merging.

Makefile
ewdurbin: The extension isn't necessary for those targets at this time.

ewdurbin:
The modification isn't required for the other targets, only `compile-pot`. If that changes for some reason #6851 will make noise :)

webknjaz:
@ewdurbin for some reason, the new strings still don't show up in Weblate. Any ideas why?

ewdurbin:
I think there is some delay after `pot` files are updated as Weblate processes them. If things haven't shown up in a few more hours I'll take a closer look.

webknjaz:
k, thanks

webknjaz:
FWIW on the info page (https://hosted.weblate.org/projects/pypa/warehouse/uk/#information) Weblate references the last commit from master already

webknjaz:
@ewdurbin looks like you also have to regenerate PO-files: I guess merging #6855 will fix this.

ewdurbin:
> @ewdurbin looks like you also have to regenerate PO-files: I guess merging #6855 will fix this.

I'm not _certain_ if that's the case, as the PO files in that PR are from Weblate itself, but worth a shot?

webknjaz:
@ewdurbin I can confirm that I've got the notification from Weblate about new strings to translate right after that PR got merged.

PR #6857, opened 2019-10-16
[WIP] Make i18n machinery opt-in on a per view basis

dstufft:
This is basically the exact same problem as with sessions, we need to know what responses will vary based on the locale or not-- so we'll use the same solution. We won't allow access to the ``request.localizer`` unless a view has been configured with ``has_translations=True``, and when it does, it will add a ``Vary: PyPI-Locale`` to the response.

This also removes translations from the ``/simple/`` index, as that's an API and not a UI page and thus shouldn't be translated.

TODO:
- [x] Revert the parts of https://github.com/pypa/warehouse/pull/6802 that add the ``PyPI-Locale`` header.
- [x] Audit the existing views and add ``has_translations=True`` to all of the ones that expect to be translated.
- [x] Edit the VCL to replace ``Vary: PyPI-Locale`` with ``Vary: Cookie``.
- [x] Write Tests.

### DIFF
--- a/tests/functional/test_caching.py
+++ b/tests/functional/test_caching.py
@@ -16,4 +16,7 @@ import pytest
 @pytest.mark.parametrize("path", ["/"])
 def test_basic_views_dont_vary(webtest, path):
     resp = webtest.get(path)
-    assert resp.headers["Vary"] == "Accept-Encoding"
+    assert resp.headers["Vary"] in [
+        "Accept-Encoding, PyPI-Locale",
+        "PyPI-Locale, Accept-Encoding",
+    ]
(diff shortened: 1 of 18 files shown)


dstufft:
We probably also want to normalize the value of PyPi-Locale to reduce the total number of variations of it that we allow. For instance, right now you can put a nonce in there and effectively bypass the cache. Ideally we should only allow very specific values and otherwise fall back to english.

ewdurbin:
Opened https://github.com/python/pypi-infra/pull/50 to address `Edit the VCL to replace Vary: PyPI-Locale with Vary: Cookie.`

ewdurbin:
@dstufft I think this is ready for review.

PR #6861, opened 2019-10-17
Unify the "Log out" spelling

switowski:
Hi,

while translating the project, I found out that there are 2 versions of `log out` string: "Log out" and "Log Out". This commit replaces the "Log Out" one with "Log out" to avoid having 2 translations.


di:
Thanks @switowski!

PR #6925, opened 2019-10-30
Change Translation to local language

### LINKED ISSUE #6924: Translation links in footer should display in local language
For example:
- "French" should be "Français"
- "Spanish" should be "Español"

Etc.

ebrahimahmadi:
For example:
"French" changes to"Français"

Fixes #6924.


ebrahimahmadi:
#6924 PR

di:
Thanks @ebrahimahmadi!

[There is a similar mapping from locale code to local name in the Django project](https://github.com/django/django/blob/4b45b6c8e4d7c9701a332e80d3b1c84209dc36e2/django/conf/locale/__init__.py). We should probably be consistent with what they use there, including capitalization.

Let's also add a comment with a link to that file for future reference.

warehouse/i18n/__init__.py
```diff
@@ -19,13 +19,16 @@
...
+# https://github.com/django/django/blob/master/django/conf/locale/__init__.py
+
 KNOWN_LOCALES = {
     "en": "English",
-    "es": "Spanish",
-    "fr": "French",
-    "ja": "Japanese",
-    "pt_BR": "Portuguese (Brazil)",
-    "uk": "Ukrainian",
+    "es": "Español",
```
di:
```suggestion
    "es": "español",
```

warehouse/i18n/__init__.py
```diff
@@ -19,13 +19,16 @@
...
+    "fr": "Français",
```
di:
```suggestion
    "fr": "français",
```

warehouse/i18n/__init__.py
```diff
@@ -19,13 +19,16 @@
...
+    "ja": "日本語",
+    "pt_BR": "Português (do Brasil)",
```
di:
```suggestion
    "pt_BR": "Português Brasileiro",
```

Issue #6378, opened 2019-08-06
Single-use API token to create new package


Zac-HD:
I'm currently at the PyCon Australia sprints, in the 2FA beta, and a big fan of narrowly scoped API tokens.  But when I went to make the first ever release of `hypothesmith` I had a problem: there's no way to scope an API token when the package doesn't exist yet!

My workaround was to create a token that works for all repos, upload the package, then revoke the token (and finally create a scoped token).

I and others here would prefer to have an option for a single-use and time limited token that can only be used to upload a new package, and is automatically revoked immediately on first use.

brainwane:
Thanks for the feature request, @Zac-HD! This is a great idea in my opinion. Unfortunately I believe it is out of scope for the current grant-funded project (for more on our current grants, see [the Packaging WG wiki page](https://wiki.python.org/psf/PackagingWG)), as are other caveat limitations (see #6255).

So I'll be talking with @ewdurbin and other maintainers to figure out whether (a) I'm wrong about it being a great idea :) and (b) whether we can find funding to design and implement this in the future.

Thanks again!

brainwane:
This use case is specific and makes a lot of sense! But. Project creation is a rare event in general for PyPI, so we want to be careful about what we engineer for and what common vs. uncommon cases we make easier in terms of UX.

A few different ways we could implement this:

* as a caveat, perhaps along with the other caveats suggested in #6255 -- a one-time user-scoped token
* via [package preview](https://wiki.python.org/psf/Fundable%20Packaging%20Improvements#Package_preview_feature_for_PyPI), #726, so the user can stage a project, then scope a token to it
* allowing users to create a token scoped arbitrarily - type any project name! (This would require that we ensure an auth check, so using such a project name doesn't actually allow the user to upload to an existing project they don't have permissions for.) And, ideally, make it a one-time token so that using it trashes the token along the way (maybe using caveats)

rcipkins:
This would probably be best to implement as a caveat. I can work on adding it.

brainwane:
@woodruffw mentions [some questions for discussion](https://github.com/pypa/warehouse/pull/6935#issuecomment-562618951) ("The semantics around one-time tokens and project version tokens") - @Zac-HD and @di you may be interested in commenting there.

webknjaz:
@brainwane may I suggest another idea:
* Create a token that has the only privilege: *create* a new project (i.e. for the first upload). It'd create a project under the user it's bound to.
[...]

PR #6935, opened 2019-10-31
One time use project upload token

rachelcipkins:
Created a one time user scoped token for project uploads. Implemented as a caveat. #6378


warehouse/legacy/api/pypi.py
```diff
@@ -145,3 +154,35 @@ def display(request):
...
+        macaroon_service.verify(
+            raw_macaroon=request.master_key,
+            context="",
+            principals="",
+            permission={"version": 2, "permissions": "user"},
+        )
+    except InvalidMacaroon:
+        raise HTTPUnauthorized()
+    user = macaroon_service.find_userid(
+        request.master_key
```
woodruffw: I see that this `master_key` attribute gets mocked in the tests, but I can't find where it actually appears in the (real) request object. Does it get set anywhere, or is this WIP?

woodruffw:
This PR now contains a few different features:

* One-time tokens (#6378)
* Expiration and project version caveats (#6255)
* An API endpoint for crafting new tokens (`/pypi/create_token/`)

woodruffw:
Some notes:

* The API endpoint *should* return an HTTP 400 (or 403, as appropriate) + JSON body on errors, but it currently returns HTML when authentication fails (e.g., when a macaroon fails caveat checks). This is probably happening somewhere further up in the auth policy and needs to be special cased; I need to look into this.

* The semantics around one-time tokens and project version tokens need some discussion: the project/file upload API is currently designed to take one upload at a time, meaning that a version release that includes multiple distributions (one `sdist`, a couple of `bdist-*`) will need `N` one-time tokens for `N` uploads. Similarly, a release is considered "created" once the first file associated with it is uploaded, meaning that macaroons bound to a single release will only work for the first uploaded file.

brainwane:
@rcipkins @woodruffw I suspect this is the kind of feature that could use a sentence of explanation in https://pypi.org/help/#apitoken .

merwok:
> The API endpoint _should_ return an HTTP 400 (or 403, as appropriate) + JSON body on errors, but it currently returns HTML when authentication fails […]. This is probably happening somewhere further up in the auth policy and needs to be special cased

Could be coming from the generic forbidden view:
https://github.com/pypa/warehouse/blob/master/warehouse/views.py#L94

A more specific view_config (using a specific exception as `context` or if not possible a custom predicate) could do what you want.

warehouse/legacy/api/json.py
```diff
@@ -205,3 +207,82 @@ def json_release_slash(release, request):
...
+        payload = request.json_body
+        if not isinstance(payload, dict):
+            raise ValueError
+    except Exception:
+        return _fail("invalid payload")
+
+    macaroon_service = request.find_service(IMacaroonService, context=None)
+
+    form = CreateMacaroonForm(
+        **payload,
```
alex, line 240: Passing `**` from attacker provided JSON is a risky move. It lets the attacker control which args are provided. If `CreateMacaroonForm` adds a new optional argument this can turn into a vulnerability. It ought to be passed as `data=payload` or explicitly unpacked.

woodruffw: Yep, excellent point. This PR needs a decent amount of reworking (and discussion, w/r/t to the semantics of "one time use" for uploads of multiple files per release).

Issue #4603, opened 2018-08-22
User RSS feeds?


glyph:
**What's the problem this feature will solve?**
I want to syndicate my activity on PyPI to other sites around the web; particularly, I'd like it if followers could see new releases of all projects that I maintain in a single feed which can be easily syndicated to Twitter and Facebook.

Users can already follow me on Github but this notifies them of when I do random garbage commits that they probably *don't* care about; following on PyPI makes more sense, so they can get notified of actual releases.

**Describe the solution you'd like**
Ideally, I'd like an RSS feed under https://pypi.org/user/glyph/.

(I actually want 2 feeds: one for "stuff I did directly" and one for "projects I am a maintainer on that did releases", to account for projects where the actual releases are done by a bot-user, or by a release manager.)

glyph:
Failing an RSS feeds, more robust API support for users would be nice; I probably wouldn't mind writing this code myself if it didn't have to download all of PyPI first.

asottile:
I've got some interest in implementing some of this -- tidelift currently has integration for rubygems and npm for determining "packages maintained by a user" but doesn't have a way to do this (they understandably do not want to scrape the `/u/<username>` page!)

if I were to attempt to solve this should I try and revive #6938 ? or should I start over just implementing the small api I'm interested (essentially a `/api/user/<username>/projects` or some such)

di:
> if I were to attempt to solve this should I try and revive https://github.com/pypi/warehouse/pull/6938 ? or should I start over just implementing the small api I'm interested (essentially a /api/user/<username>/projects or some such)

I think we're open to implementing this, it's just not very high priority. I think #6938 was mostly headed in the right direction and could be revived, but also I'm not sure what you actually want and if it's different from this issue, which is about a chronological syndication feed, and seems like it might not be what tidelift wants.

asottile:
> > if I were to attempt to solve this should I try and revive #6938 ? or should I start over just implementing the small api I'm interested (essentially a /api/user//projects or some such)
> 
> I think we're open to implementing this, it's just not very high priority. I think #6938 was mostly headed in the right direction and could be revived, but also I'm not sure what you actually want and if it's different from this issue, which is about a chronological syndication feed, and seems like it might not be what tidelift wants.

[...]

PR #6938, opened 2019-10-31
User updates RSS feed

rascalking:
This addresses issue #4603.


warehouse/templates/rss/user_updates.xml
```diff
@@ -0,0 +1,13 @@
+{% extends "base.xml" %}
+{% block title %}{{ user.name }} recent updates{% endblock %}
```
yeraydiazdiaz: This should be `username` it's currently not rendering anything resulting `<title> recent updates</title>`

```suggestion
{% block title %}{{ user.username }} recent updates{% endblock %}
```

Also happening in the line below.

warehouse/templates/accounts/profile.html
```diff
@@ -39,6 +39,12 @@ <h1 class="author-profile__name">{{ user.name }}</h1>
               &nbsp;&nbsp;{% trans start_date=humanize(user.date_joined) %}Joined on {{ start_date }}{% endtrans %}
             </p>
             {% endif %}
+            {% if projects %}
+            <p>
+              <i class="fa fa-rss" aria-hidden="true"></i>
+              <span class="sr-only">{% trans %}RSS Feed{% endtrans %}</span>
+              &nbsp;&nbsp;<a href="{{ request.route_path('rss.user_updates', username=user.username) }}">{% trans %}Recent updates{% endtrans %}</a></p>
+            {% endif %}
```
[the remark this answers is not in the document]
rascalking: I think adding in the extra `alternate` link probably makes sense, regardless.  I'll add it in with the bugfix.

warehouse/templates/rss/user_updates.xml
rascalking: Fixed.

warehouse/templates/accounts/profile.html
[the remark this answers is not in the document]
di: I think it makes sense to make the link visible in the profile somewhere, otherwise it would be challenging to find these feeds. The other site-wide ones are listed at https://warehouse.readthedocs.io/api-reference/feeds/, and this PR should probably update that documentation as well.

warehouse/rss/views.py
```diff
@@ -72,3 +73,34 @@ def rss_packages(request):
...
+@view_config(
+    route_name="rss.user_updates",
+    context=User,
+    renderer="rss/user_updates.xml",
+    decorator=[
+        origin_cache(
+            1 * 24 * 60 * 60,  # 1 day
+            stale_while_revalidate=1 * 24 * 60 * 60,  # 1 day
+            stale_if_error=5 * 24 * 60 * 60,  # 5 days
+            keys=["all-projects"],
```
di: This line would cause the cache for every user RSS feed to be purged whenever any new project was created. It's not necessary since we already have the `context=User` line above.

rascalking: Thanks for the catch.  I just copied and pasted this from the other RSS views, should have read through the code to understand it first.

warehouse/templates/accounts/profile.html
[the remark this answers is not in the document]
rascalking: Done.

warehouse/rss/views.py
```diff
@@ -72,3 +73,33 @@ def rss_packages(request):
...
+)
+def rss_user_updates(user, request):
+    request.response.content_type = "text/xml"
+
+    request.find_service(name="csp").merge(XML_CSP)
+
+    latest_releases = (
+        request.db.query(Release)
+        .join(Project)
+        .filter(Project.users.any(username=user.username))
```
di: Note that this doesn't mean that this user has made this release, but rather that this user has some role on a project that has made a release.

In #4603 the request is for both:

> (I actually want 2 feeds: one for "stuff I did directly" and one for "projects I am a maintainer on that did releases", to account for projects where the actual releases are done by a bot-user, or by a release manager.)

I think it's fine if we're only implementing half of #4603 here, but we might want to think a little more about what we're calling this feed since `updates` is pretty generic and we might have multiple feeds here eventually.

We also might want to make it more clear on the profile page what this feed is exactly.

rascalking: Hrm...maybe call them "My Releases" vs. "My Projects' Releases"?

tests/unit/test_routes.py
```diff
@@ -283,6 +283,13 @@ def add_policy(name, filename):
         pretend.call("ses.hook", "/_/ses-hook/", domain=warehouse),
         pretend.call("rss.updates", "/rss/updates.xml", domain=warehouse),
         pretend.call("rss.packages", "/rss/packages.xml", domain=warehouse),
+        pretend.call(
+            "rss.user_updates",
+            "/user/{username}/updates.xml",
```
di: Looking at #7013, I think this should probably be `/rss/user/{username}/something.xml`, where `something` is distinctly one of the two types of feeds discussed in #4603.

rascalking:
Okay, I implemented both feeds, named them `My Releases` and `My Projects' Releases`, and updated the urls as suggested.  I think this is all set, let me know if there's anything I missed.

di:
Hi @rascalking, can you update the merge conflicts here now that #7013 has been merged? You might also want to make some updates (docs, etc.) to align more with what was merged there.

Also, after giving it some thought I think the URLs should be:

* Uploaded Releases: `https://pypi.org/rss/user/<username>/uploaded_releases.xml`
* Maintained Releases: `https://pypi.org/rss/user/<username>/maintained_releases.xml`

rascalking:
Sorry, I don't think I'm going to be able to finish this.

PR #6957, opened 2019-11-04
add locale el (greek translations)

manikos:


manikos:
My apologies if this pull request misses something but when I `make build-mos` I got this output when it tries to compile the `messages.po` file for the `el` locale:

`catalog warehouse/locale/el/LC_MESSAGES/messages.po is marked as fuzzy, skipping`

I have no idea why this is happening! I'll be glad to give more info on this. BTW, the translations on weblate for the Greek language are 100% (https://hosted.weblate.org/projects/pypa/warehouse/el/).

ewdurbin:
@manikos thank you for your Pull Request. I'm not sure why you received the error message above. I was able to successfully compile the mo using the `build-mos` target.

Can you attempt to reproduce and file an issue for us to investigate further?

manikos:
> @manikos thank you for your Pull Request. I'm not sure why you received the error message above. I was able to successfully compile the mo using the `build-mos` target.
> 
> Can you attempt to reproduce and file an issue for us to investigate further?

I think it has to do with the setup of the cloned repo. In other words, I think it's my issue.

1. Forked the repo
2. Created a branch and checked out to this branch
3. Made the changes to the `__init__.py` file to add the locale string
4. Created a virtualenv using virtualenvwrapper and pyenv (python 3.7)
5. Edit the `Makefile` because at the top it has a variable `BINDIR = $(PWD)/.state/env/bin` which point to a bin dir which is empty. I changed it to point to `~/.virtualenvs/warehouse/bin`
6. Run `make build-mos`
7. Got this output

No docker-related programs installed.

So, I strongly believe it is because I have done something with the setup (edit the `Makefile` which is not supposed to be touched at all).

PR #6980, opened 2019-11-08
Add link to homepage on 404 template

RDIL:
Nice-to-have.

### DIFF
--- a/warehouse/templates/404.html
+++ b/warehouse/templates/404.html
@@ -15,7 +15,7 @@
 
 {% block title %}{% trans %}Page Not Found (404){% endtrans %}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


ewdurbin:
Hi @RDIL! The 404 page already contains two links back to the home page from the `error-base.html` template.

One in the logo, though without any prompt:

https://github.com/pypa/warehouse/blob/73f0325f62dd4e73c41963c4c0b324e153eccb44/warehouse/templates/error-base.html#L21-L23

Screen Shot 2019-11-08 at 9 24 46 AM

And one in the footer, which explicitly states "Back to the homepage":

https://github.com/pypa/warehouse/blob/73f0325f62dd4e73c41963c4c0b324e153eccb44/warehouse/templates/error-base.html#L33

Screen Shot 2019-11-08 at 9 24 56 AM

I'm not sure that a third is necessary.

PR #6988, opened 2019-11-10
clarify error on readme render failure

wesinator:
failure could be on the license, not just description 
https://github.com/pypa/setuptools/issues/1390#issuecomment-552220402

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -1015,13 +1015,13 @@ class TestFileUpload:
             (
                 "text/x-rst",
                 ".. invalid-directive::",
-                "400 The description failed to render for 'text/x-rst'. "
+                "400 The readme failed to render for 'text/x-rst'. "
                 "See /the/help/url/ for more information.",
             ),
             (
                 "",
                 ".. invalid-directive::",
-                "400 The description failed to render in the default format "
+                "400 The readme failed to render in the default format "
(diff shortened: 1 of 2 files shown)


ewdurbin:
Hi @wesinator, thanks for your pull request.

I'm not sure that we want to change these error messages as they are being generated only in the case where the `description` metadata contains something we can't render, so I'm not sure how it relates to the license metadata. Colloquially people interchange "Readme" and "Description" but the official name for the metadata field is [Description](https://packaging.python.org/specifications/core-metadata/#description).

wesinator:
> Hi @wesinator, thanks for your pull request.
> 
> I'm not sure that we want to change these error messages as they are being generated only in the case where the `description` metadata contains something we can't render, so I'm not sure how it relates to the license metadata. Colloquially people interchange "Readme" and "Description" but the official name for the metadata field is [Description](https://packaging.python.org/specifications/core-metadata/#description).

the error was displayed in the case where _license_ wasn't renderable - https://github.com/pypa/warehouse/issues/3664#issuecomment-552218625

ewdurbin:
@wesinator it looks like the bug in setuptools that was coming up in your case caused a malformed PKG-INFO to be created which in turn caused the client uploading to PyPI to send a malformed payload for "Description". So in this case PyPI was correct in reporting a problem with that field.

PyPI can't detect the specific issue that is being referenced, so the error message that's being returned here is correct.

PR #7003, opened 2019-11-14
add PEP 541 link

hallazzang:
As documentation says:
> Follow the "How to request a name transfer" section of PEP 541.

add link for "How to request a name transfer" section of PEP 541.


di:
Thanks!

PR #7010, opened 2019-11-17
Match RSS version with content, mark as 2.0

scop:
Ref http://www.rssboard.org/rss-specification (and elsewhere).

While the unmodified files pass various "RSS validators" out there (including the W3C one), those validators apparently fail to actually validate the content per the DTD -- they would not validate, because they're using the 2.0, not 0.91 structure. For 2.0 there's no DTD, http://www.rssboard.org/convert-rss-0-91-to-rss-2-0


scop:
Rebased to current master.

PR #7012, opened 2019-11-17
Add item authors to RSS feeds

scop:


tests/unit/rss/test_views.py
```diff
@@ -64,5 +67,40 @@ def test_rss_packages(db_request):
...
+    release.author_email = "noreply@pypi.org"
+    assert rss._format_author(release) == "Nön Äscii <noreply@pypi.org>"
+
+    release.author = "Author One, Author Two, Author Three"
+    release.author_email = "noreply@pypi.org, noreply@pypi.org, noreply@pypi.org"
+    assert rss._format_author(release) == release.author_email
+
+    release.author = "John Doe"
+    release.author_email = None
+    assert rss._format_author(release) is None
```
yeraydiazdiaz: Nit: is there a reason to not include the author at all if their email is `None`? Seems like we should include that information if we do have it.

tests/unit/rss/test_views.py
```diff
@@ -64,5 +67,40 @@ def test_rss_packages(db_request):
...
+    release.author_email = "noreply@pypi.org"
+    assert rss._format_author(release) == "No Reply <noreply@pypi.org>"
+
+    release.author = "Nön Äscii"
+    release.author_email = "noreply@pypi.org"
+    assert rss._format_author(release) == "Nön Äscii <noreply@pypi.org>"
+
+    release.author = "Author One, Author Two, Author Three"
+    release.author_email = "noreply@pypi.org, noreply@pypi.org, noreply@pypi.org"
+    assert rss._format_author(release) == release.author_email
```
yeraydiazdiaz: Nit: this could also include the names in the form `"No Reply <noreply@pypi.org>"` as well to be consistent, unless I'm missing something?

di:
Given that author metadata for a release is optional and not guaranteed to be correct, would it make more sense to include the uploader's PyPI username or a link to their profile instead?

scop:
Content of `<author>` is specified to be the author's email address in RSS, using some other info would be against that. http://www.rssboard.org/rss-specification#ltauthorgtSubelementOfLtitemgt

tests/unit/rss/test_views.py
scop: Content of `<author>` is specified to be the author's email address in RSS, that's why I'm omitting it if we don't have one. http://www.rssboard.org/rss-specification#ltauthorgtSubelementOfLtitemgt

tests/unit/rss/test_views.py
scop: That would be nice, but would require opening the can of worms of heuristic parsing of author and email fields to look for multiple entries. Do we want to support semicolon as a separator in addition to comma? Do we support both mixed in one string (e.g. `foo, bar; quux` contains three or two elements -- if two, which two)? And author names can contain at least commas.

Of course we could try it out for some simple cases, e.g. split both author and author_email on commas, and if the lists are of equal length, zip-format addresses.

scop: Added a couple of commits along these lines.

scop:
While on the topic of RSS feeds, #7013 would be a really nice addition ;)

tests/unit/rss/test_views.py
yeraydiazdiaz: Got it, thanks for the link. If we're adhering this closely to the spec it looks like the proper form should be `lawyer@boyer.net (Lawyer Boyer)`, which is slightly different than the current `No Reply <noreply@pypi.org>`.

Not a blocker by any means, I'm just a bit unsure how much we can deviate from the spec or not.

tests/unit/rss/test_views.py
yeraydiazdiaz: Having a look at the test database rows for `author` I can already see examples that would break when splitting on `,` like `Cloudistics, Inc.` and similar examples in other languages. There's also rows including the email addresses and even str'd lists of authors i.e. `['John Doe', 'Jane Doe']`.

I don't think we can reliably parse these into `(name, email)` pairs.

`author_email` seems to be more structured, althought there's quite a bit of `UNKNOWN` entries. Maybe we should just rely on that and filter the `UNKNOWN` and nulls?

scop: Done, rebased and simplified.

tests/unit/rss/test_views.py
scop: I think the formatting in the spec is pretty much just an example. Anyway the current revision of the PR uses the bare email address, and ignores the realname.

warehouse/rss/views.py
```diff
@@ -18,6 +20,19 @@
 from warehouse.xml import XML_CSP
 
 
+def _format_author(release):
```
yeraydiazdiaz: Could you add a quick docstring on this function to explain the rationale behind it?

scop: Done.

PR #7076, opened 2019-12-10
Fix frontend locale and localized dates

### LINKED ISSUE #6848: HTML templates hardcoded with "en" language attributes
The base template for warehouse pages have the following hardcoded "language" bits:

https://github.com/pypa/warehouse/blob/bc6a10bf5c58332a03e389781546ce963bfee128/warehouse/templates/base.html#L65

https://github.com/pypa/warehouse/blob/bc6a10bf5c58332a03e389781546ce963bfee128/warehouse/templates/base.html#L71-L72

This was noticed in https://github.com/python/pypi-infra/pull/49 and should probably be addressed?

I'm not certain what purpose these have and am not sure how to best address.

### LINKED ISSUE #6862: Dates should be localized too
So while browsing PyPI with translation switched, I noticed that dates are still displayed in English.
This should be adjusted accordingly.

kilrogg:
Fixes #6862, fixes #6848 by:

- Changing "lang" attribute for html document to current request locale
- Fixing hard coded english localization of dates and times to that document locale
- Filling META-Tag "availableLanguages" with the language codes from KNOWN_LOCALES


di:
Thanks @kilrogg!

PR #7117, opened 2019-12-18
Add limit and max_age params to RSS feeds

alexbecker:
Implements https://github.com/pypa/warehouse/issues/7116, introducing `limit` and `max_age` query parameters to the `packages.xml` and `releases.xml` feeds. The `limit` parameter controls how many items are returned, defaulting to 40 if not present (same as current behavior) and respecting values up to 200 (an arbitrary maximum I chose, please suggest a more appropriate value). The `max_age` query parameter filters down to items at most `max_age` seconds old; I chose to use seconds of age instead of datetimes/timestamps in the request because they are easier to parse (unless I used unix timestamps, I suppose) and less sensitive to clock skew.

Creates a small `_get_int_query_param` utility to parse these parameters and 400 if they are not valid integers. Please let me know if there is an existing utility I should use or if I should move this to the `utils` module.

Currently `limit` requests over `MAX_RESULTS` (200) are silently reduced to returning `MAX_RESULTS` items. Should this be changed to a 400 error?

Thank you for maintaining PyPI and thanks in advance for your time reviewing this PR!


tests/unit/rss/test_views.py
```diff
@@ -44,6 +48,72 @@ def test_rss_updates(db_request):
...
+        )
+    )
+
+    db_request.session = pretend.stub()
+
+    project1 = ProjectFactory.create()
+    project2 = ProjectFactory.create()
+
+    release1 = ReleaseFactory.create(project=project1)
+    release1.created = datetime.date(2011, 1, 1)
```
yeraydiazdiaz: Nit: Any reason to not pass this in the `create` call above?

warehouse/rss/views.py
```diff
@@ -10,13 +10,31 @@
...
+MAX_RESULTS = 200
+
+
+def _get_int_query_param(request, param, default=None):
+    value = request.params.get(param)
+    if not value:
+        # Return default if 'param' is absent or has an empty value.
+        return default
+    try:
+        return int(value)
```
yeraydiazdiaz: Maybe it'd be worth asserting it must be a _positive_ integer as well?

warehouse/rss/views.py
```diff
@@ -10,13 +10,31 @@
...
 from warehouse.cache.origin import origin_cache
 from warehouse.packaging.models import Project, Release
+from warehouse.utils import now
 from warehouse.xml import XML_CSP
 
+DEFAULT_RESULTS = 40
+MAX_RESULTS = 200
+
+
+def _get_int_query_param(request, param, default=None):
```
yeraydiazdiaz: Nit: I feel defaulting to `None` might be dangerous here as the caller might expect an integer. The two options I see is raising an expection or returning a sentinel value (maybe `-1` or `0` would make sense?).

tests/unit/rss/test_views.py
alexbecker: No reason, I just cargo-culted this from the existing tests in this file.

di:
I think we probably need a bit more discussion before continuing work on this, see https://github.com/pypa/warehouse/issues/7116#issuecomment-570303891.

firefly9966 (approved):

PR #7124, opened 2019-12-19
GitHub token scanning api view

### LINKED ISSUE #6051: GitHub Token Scanning Partnership

GitHub has a "Token Scanning" feature, which enables them to detect API keys/tokens, and notify the service provider who issued the token. The service providers can then act on this information according to their own policies (revoke, contact owner etc).

This helps avoid situations where an active API key is available publicly on the project's repository. GitHub has a Partnership program for this, with various service providers (incl. [npm][1] and [more][2]).

[1]: https://github.blog/changelog/2019-06-18-github-adds-npm-as-a-token-scanning-partner/
[2]: https://help.github.com/en/articles/about-token-scanning

It would be worthwhile to explore partnering with GitHub for Token Scanning, when the support for API Keys has been completed.

**Additional context**

https://developer.github.com/partnerships/token-scanning/ contains details about how to contact GitHub for this partnership, and details on how the integration works.

ewjoachim:
Fixes #6051 

## High level description to ease review:
(Assuming you're familiar with the ticket).

### The view
In this PR, we create a view (`github_disclose_token`) which:
- Is a json POST API view meant to be called by GitHub
- Returns either 204 no content or 40x and a json error payload
- Expects a json body and 2 custom headers

This view uses 2 main components described below.

### GitHubTokenScanningPayloadVerifyService
This service validates that the call originates from GitHub by doing the following:
- A request to GitHub meta API & parsing the result to retrieve a public key, given a fingerprint
- Check that the provided signature of the json payload of the request is valid given this public key

### TokenLeakAnalyzer
This part is voluntarily independant from GitHub
The request JSON payload contains leaked tokens to analyse. Each token might either be a string of the form `pypi-xxx` or a b64 encoded form of `__token__:pypi-xxx`.
- We extract the token from whichever form it is
- We check if the token exists in the DB and it's signature is correct
- If so we invalidate it
- And send a mail to the user

### The rest
- We needed a new function `check_if_macaroon_exists` in the macaroon service (& associated tests)
- There's a new email function send_password_compromised_email_leak and the associated email
- There's a FAQ page

TODO:
- [x] Get some answers regarding:
  - [x] How to cache things?
  - [x] Do we have a github token, and if not, would we plan to get one to avoid IP based rate limiting? **Well, I've implemented it, but it's optionnal** 
  - [x] When things go bad, how should we log?
  - [x] How to check multiple headers in the view config predicates?
- [x] tests:
  - [x] GitHubTokenScanningPayloadVerifyService
  - [x] NullGitHubTokenScanningPayloadVerifyService
  - [x] check_if_macaroon_exists
  - [x] TokenLeakAnalyzer
  - [x] github_disclose_token view
  - [x] send_password_compromised_email_leak
- [x] FAQ in the doc
- [x] Get an answer from GitHub and implement the signature crypto check
- [x] Test that it works on the dev env


ewdurbin:
Hi @ewjoachim https://github.com/pypa/warehouse/blob/master/warehouse/email/ses/views.py has a lot of good reference that may help answer many of your open questions.

ewjoachim:
Ah, I'll look at this, thank you a lot.

ewjoachim:
I'm not sure I understand why the `MessageVerifier` is implemented as a class in `/utils` rather than as a service. I believe the architecture of what I'm trying to achieve is indeed very similar, so I'm enclined to do the same with a `GitHubTokenScanningRequestVerifier`, but I'm not sure I really understand why this should not be a service ?

ewjoachim:
Ok, it still misses tests & docs, but it should start being readable.

I'm interested in pointers for the remaining todos, feedback if my code style was right (escpecially if I need to redo large parts of the code, before I start writing all the tests) and maybe wording advice for the email.

dstufft:
Services are for things where you might realistically want to replace the implementation with something else at runtime. So the code that interacts with ``MessageVerifier`` is never going to want to use anything *but* a ``MessageVerifier``, it makes no sense to make it swapable because that code is so directly tied to SES that there's not really a second implementation that makes sense.

However something like... where we store file objects is something that we might absolutely want to swap at runtime (and we do, we use different storage engines at runtime versus at development time) so it makes perfect sense for that to be a service. Another example is the HIBP code, one option when we were developing that was taking the HIBP data and running our own copy locally, using a service let us make that a more feasible option with minimal changes to the code if we ever decide to go that route.

ewjoachim:
It feels like we could have a reason to swap the fetching and signature verification part at run or test time (for example for a version that would not contact github and work with local keys and key ids).

On the other hand, the token check, revocation and user warning code seems pretty much straightfoward.

Is that to say I should have implemented this a a 3 parts:
- Payload parsing & basic checks, not as a service
- Signature check (including contacting github to check the public key), as a service
- Examining tokens and revoking the valid ones, not as a service

?

Side note: I've tried to implement part 3 completely independently from GitHub, with the idea that we may want to have revocation protocols from other services (google ? manual with a rate limit ? Who knows ?), but whatever the origin of the disclosure is, it doesn't mean the implementation would change, so this would still say "not a service" if I understand correctly.

So, shall I change it all ?

ewjoachim:
> So, shall I change it all ?

I've done that.

New iteration, far less `# TODOs`. Next step, get an answer from GitHub, and write some tests.

warehouse/accounts/utils.py
```diff
@@ -0,0 +1,151 @@
...
+        # That being said, except logging, there's not much we can do.
+        try:
+            token = disclosure.matcher.extract(text=disclosure.token)
+        except ExtractionFailed:
+            metrics.increment(f"warehouse.token_leak.{origin}.error.extraction")
+            return
+
+        macaroon_service = self._request.find_service(IMacaroonService, context=None)
+
+        database_macaroon = macaroon_service.find_macaroon_from_raw(raw_macaroon=token)
```
ewjoachim: OMG ! I nearly missed that I need to somehow verify that the signature on that macaroon is correct, otherwise I'm just going to check that there is a macaroon with this ID in the DB.
It's probably not that bad, given I'd need to guess an existing UUID, but... Still...

ewjoachim:
Missing coverage:
```
warehouse/accounts/services.py                      335      1     88      0    99%   805
warehouse/accounts/utils.py                          67     39      8      0    40%   37, 43-47, 82-104, 111, 115-146
warehouse/accounts/views.py                         284     19     90      0    94%   731-765
warehouse/email/__init__.py                          72      1     10      0    99%   174
```

Still a few open questions for which I'll need someone's help though (see in the PR message at the top)

ewjoachim:
Ok, I'm ready for review. I've been told the PR is likely too big to be easily review, which I would totally understand, but then I need suggestion on how to break it up into smaller ones.

Also, there are still unanswered questions regarding cache, logging and Pyramid's header predicate (see at the top in the todo-list)

ewjoachim:
Rebased, fixed conflict.

@ewdurbin / @dstufft  Do you think you might have some time to make a first round of review ? Is there something I can do to make it easier to you ?

ewjoachim:
Hey 👋 ! Crazy month... Any chance you could start reviewing or give me clues on what I should do to help this PR be reviewed ?

Meanwhile, I'll rebase and fix conflicts :)

tests/unit/accounts/test_views.py
```diff
@@ -1889,6 +1890,7 @@ def test_edit_profile_button(self):
         request = pretend.stub()
 
         assert views.edit_profile_button(user, request) == {"user": user}
+<<<<<<< HEAD
```
merwok, line 1893: wrong merge here :slightly_smiling_face:

warehouse/accounts/views.py
```diff
@@ -778,3 +782,54 @@ def edit_profile_button(user, request):
...
+
+
+@view_config(
+    require_methods=["POST"],
+    require_csrf=False,
+    renderer="json",
+    route_name="accounts.github-disclose-token",
+    header="GITHUB-PUBLIC-KEY-IDENTIFIER",
+    # TODO: How to check multiple headers in the predicates ?
+    # header="GITHUB-PUBLIC-KEY-SIGNATURE"
```
merwok, line 794: It is not possible at the moment; I submitted https://github.com/Pylons/pyramid/issues/3575

A custom predicate is needed here to be able to write `headers=("header1", "header2")`
see here: https://raw.githubusercontent.com/Pylons/pyramid/master/docs/narr/urldispatch.rst
(cloudflare issue prevents seeing the docs and github rendering misses the code blocks, so it’s source markup)

tests/unit/accounts/test_utils.py
```diff
@@ -0,0 +1,221 @@
...
+        {"type": type, "token": token, "url": "http://example.com"}
+    )
+
+    assert request.token == "pypi-1234"
+    assert request.public_url == "http://example.com"
+
+
+def test_token_leak_analyzer_analyze_disclosure(monkeypatch):
+
+    metrics = collections.defaultdict(int)
```
webknjaz, line 96: Did you mean counter?
```suggestion
    metrics = collections.Counter()
```

tests/unit/macaroons/test_services.py
```diff
@@ -197,3 +197,61 @@ def test_get_macaroon_by_description(self, macaroon_service):
...
+        with pytest.raises(services.InvalidMacaroon):
+            macaroon_service.check_if_macaroon_exists(raw_macaroon)
+
+    def test_check_if_macaroon_exists_no_macaroon(self, macaroon_service):
+        raw_macaroon = pymacaroons.Macaroon(
+            location="fake location",
+            identifier=str(uuid4()),
+            key=b"fake key",
+            version=pymacaroons.MACAROON_V2,
+        ).serialize()
```
webknjaz, line 218: Maybe move the setup part into a fixture?
```suggestion
    @pytest.fixture
    def raw_macaroon()
        return pymacaroons.Macaroon(
            location="fake location",
            identifier=str(uuid4()),
            key=b"fake key",
            version=pymacaroons.MACAROON_V2,
        ).serialize()

    def test_check_if_macaroon_exists_unprefixed_macaroon(self, macaroon_service, raw_macaroon):
        with pytest.raises(services.InvalidMacaroon):
            macaroon_service.check_if_macaroon_exists(raw_macaroon)

    def test_check_if_macaroon_exists_no_macaroon(self, macaroon_service, raw_macaroon):
```

tests/unit/accounts/test_views.py
ewjoachim: Woops :)

warehouse/accounts/views.py
ewjoachim: So you think the best course of action would be for your PR to land on master and then in warehouse (no idea if we're talking days, weeks, years) or build a custom predicate now ?

tests/unit/accounts/test_utils.py
ewjoachim: Ah, yes, both work. In this case, I'll do ``metrics.update(key)``.

I'll do it, no worry, but is there a specific reason I should learn to use one over the other ?

ewjoachim: (done)

warehouse/accounts/views.py
merwok: It could be weeks before Warehouse can update to a Pyramid release that includes my PR.
Go for a custom predicate now!

tests/unit/accounts/test_utils.py
webknjaz: It's just something that fits better semantically, hence the code would read better.

warehouse/accounts/views.py
ewjoachim: Is there a predicate that combines predicates ? In that case maybe I could simply have something like `And(Header(...), Header(...))`

ewjoachim: Ok, seems I get an easy solution with:
```
class HeadersPredicate:
    def __init__(self, val: List[str], config):
        self.sub_predicates = [
            predicates.HeaderPredicate(subval, config) for subval in val
        ]

    def text(self):
        return ", ".join(sub.text() for sub in self.sub_predicates)

    phash = text

    def __call__(self, context, request):
        return all(sub(context, request) for sub in self.sub_predicates)
```

ewjoachim: ok done, with tests and all.

ewjoachim:
Should there be a way to quickly enable/disable/silent the feature in case it doesn't work as expected ?

ewjoachim:
Hey there :) I hope it's ok if I ping this PR every month or so :) Tell me otherwise !

Again, if there is anything that needs to be done before reviewing, please tell me :)

warehouse/accounts/views.py
```diff
@@ -814,3 +818,52 @@ def edit_profile_button(user, request):
...
 def profile_public_email(user, request):
     return {"user": user}
+
+
+@view_config(
+    require_methods=["POST"],
+    require_csrf=False,
+    renderer="json",
+    route_name="accounts.github-disclose-token",
+    headers=["GITHUB-PUBLIC-KEY-IDENTIFIER", "GITHUB-PUBLIC-KEY-SIGNATURE"],
```
di, line 828: This almost feels like overkill for something that would only be an extra two lines in the view body:

```
if key_id is None or signature is None:
    return Response(status=400)
```

If we really think we want to keep this for future use, I'd recommend naming it something like `require_headers` instead, and adding some additional comments/documentation to the predicate. For example, it's not clear to me what a request with incorrect headers would get in response.

warehouse/accounts/services.py
```diff
@@ -10,18 +10,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import base64
 import collections
+import enum
```
di, line 15: Unused?

warehouse/accounts/services.py
```diff
@@ -10,18 +10,29 @@
...
 import functools
 import hashlib
 import hmac
+import json
 import logging
 import posixpath
+import re
 import secrets
+import typing
```
di, line 24: Unused?

warehouse/accounts/services.py
```diff
@@ -10,18 +10,29 @@
...
 import urllib.parse
 import uuid
 
+import attr
```
di, line 28: Unused?

warehouse/accounts/utils.py
```diff
@@ -0,0 +1,181 @@
...
+}
+
+
+class InvalidTokenLeakRequest(Exception):
+    def __init__(self, message, reason):
+        self.reason = reason
+        super().__init__(message)
+
+
+@attr.s(auto_attribs=True, frozen=True, slots=True)
```
di, line 86: Currently `attrs` is not a top-level dependency of Warehouse, it's only available as a sub-dependency of `automat`. If we truly need to use `attrs` here, it should be added to https://github.com/pypa/warehouse/blob/master/requirements/main.in and those dependencies should be re-compiled.

However I think I'd prefer to just use a regular class here.

warehouse/macaroons/services.py
```diff
@@ -118,6 +118,32 @@ def verify(self, raw_macaroon, context, principals, permission):
...
+        except MacaroonDeserializationException:
+            raise InvalidMacaroon("malformed macaroon")
+
+        dm = self.find_macaroon(m.identifier.decode())
+
+        if dm is None:
+            raise InvalidMacaroon("deleted or nonexistent macaroon")
+
+        verifier = Verifier(m, context=None, principals=None, permission=None)
+        verifier.verify_signature(dm.key)
```
di, line 143: Why do we need to do any additional verification here? If `find_macaroon` says the macaroon exists, that's enough to prove its existence, no?

Specifically in the context of the GitHub integration, if they are able to produce a valid macaroon that exists in our DB, that should be enough to warrant it's deletion.

warehouse/templates/email/token-compromised-leak/body.html
```diff
@@ -0,0 +1,68 @@
...
+  someone else acquired and published your token.{% endtrans %}
+</p>
+<p>
+  {% trans site_name=site_name %}
+  If the token was not published by you, it means it was stolen, either from
+  you or from {{ site_name }}. We urge you to take all the necessary steps to secure all your
+  credentials.{% endtrans %}
+  {% trans site_name=site_name, href="https://python-security.readthedocs.io/packages.html#vulnerabilites-in-the-package-index" %}
+  In the case of a breach of the {{ site_name }} database, you will find informations
+  and support in the <a href="{{ href }}">python-security</a> documentation.{% endtrans %}
```
di, line 40: If PyPI got breached and one of these tokens showed up on GitHub as a result, this may be the first indication we get. I'm not sure we'd already have information at that URL. I think the following are mutually exclusive:

* PyPI got breached, we found out and reset all tokens before they could make it on GitHub, and added some documentation to that effect.
* PyPI got breached, we didn't find out before they could make it on GitHub, an so there's no documentation to that effect.

In general, I don't think tokens leaking to GitHub would likely be indicative of a breach, so I'm not sure we need to devote this much discussion to it in the email.

warehouse/templates/email/token-compromised-leak/body.html
```diff
@@ -0,0 +1,68 @@
...
+  {# TODO how much of an effort should we do to replace PyPI with {{ site_name }} ? #}
```
di, line 41: If we're talking about a specific site, e.g. "Your PyPI|TestPyPI token was leaked" or linking to something on the site (which should done via `request.help_url` or something similar), then we should make every effort.

I think the only exception would be the admin email, which should always be `@pypi.org`.

warehouse/templates/email/token-compromised-leak/body.html
```diff
@@ -0,0 +1,68 @@
...
+  you and release a malicious package.{% endtrans %}
+</p>
+<h3>{% trans %}How do you know this?{% endtrans %}</h3>
+<p>
+  {% trans site_name=site_name %}This is an automated message. Our partner {{ origin }} analyzes all the
+  contents it recieves for potential {{ site_name }} token leaks and warns us every time it finds one.
+  We check every disclosure we recieve and take action when the token appears valid.{% endtrans %}
+</p>
+<p>
+  {% trans trimmed faq_url=request.help_url(_anchor='TODO'), email_href='mailto:admin@pypi.org', email_address='admin@pypi.org' %}
```
di, line 64: TODO here

warehouse/accounts/utils.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self._metrics.increment(f"warehouse.token_leak.{origin}.recieved")
+
+        try:
+            disclosure = TokenLeakDisclosureRequest.from_api_record(
+                record=disclosure_record
+            )
+        except InvalidTokenLeakRequest as exc:
+            # TODO Logging something here would be useful in case we recieve
+            # unexpected pattern type.
```
di, line 145: We don't really rely on logs very much. Ideally we would be watching the metrics for the types of exceptions occurring here instead.

If this is truly unexpected and "shouldn't happen", we might want to let this exception bubble up into our exception handling service instead, where we'd be immediately notified, and have a lot more context than the metrics service or a log message would be able to provide.

warehouse/accounts/services.py
```diff
@@ -759,3 +772,163 @@ def check_password(self, password, *, tags=None):
...
+                f"Invalid response code {response.status_code}: {response.text[:100]}",
+                f"public_key_api.status.{response.status_code}",
+            ) from exc
+        except json.JSONDecodeError as exc:
+            raise GitHubPublicKeyMetaAPIError(
+                f"Non-JSON response received: {response.text[:100]}",
+                "public_key_api.invalid_json",
+            ) from exc
+        except requests.RequestException as exc:
+            # TODO Log
```
di, line 850: I think incrementing a metric is sufficient here.

warehouse/accounts/services.py
```diff
@@ -759,3 +772,163 @@ def check_password(self, password, *, tags=None):
...
+        token_scanning_pubkey_api_url = (
+            "https://api.github.com/meta/public_keys/token_scanning"
+        )
+        headers = self._get_headers()
+        try:
+            response = self._session.get(token_scanning_pubkey_api_url, headers=headers)
+            response.raise_for_status()
+            return response.json()
+        except requests.HTTPError as exc:
+            # TODO Log, including status code and body
```
di, line 839: This seems like an exception we would want to see and should just let occur.

warehouse/accounts/utils.py
```diff
@@ -0,0 +1,181 @@
...
+            )
+        for disclosure_record in disclosure_records:
+            try:
+                self.analyze_disclosure(
+                    disclosure_record=disclosure_record, origin=origin
+                )
+            except Exception:
+                # TODO log, but don't stop processing other leaks.
+                # It seems logger.exception() is not used in the codebase. What is
+                # expected ?
```
di, line 180: Somewhat unfortunate that we might have multiple disclosures to verify, but verifying any one of them may result in an exception (either intentional or unintentional) that may block the verification of one or more disclosures.

This makes me think it might make sense to make each verification it's own background task (via our Celery workers), so each can fail independently of another instead.

warehouse/accounts/views.py
```diff
@@ -814,3 +818,52 @@ def edit_profile_button(user, request):
...
+    has_translations=True,  # Not the view itself, but the email it sends
```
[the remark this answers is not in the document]
ewjoachim: Hm, I think I did it because I copied that from the Have I Been Pwned emails, but I'll happily remove that part !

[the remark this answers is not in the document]
di: The HIBP emails are generated from a user request that may have a locale set, so it's appropriate there and slightly different than what's happening here, where these are always generated by an automated system that will never have a locale set.

warehouse/accounts/views.py
```diff
@@ -814,3 +818,52 @@ def edit_profile_button(user, request):
...
+    # https://developer.github.com/partnerships/token-scanning/
+
+    body = request.body
+
+    # Thanks to the predicates, we know the headers we need are defined.
+    key_id = request.headers.get("GITHUB-PUBLIC-KEY-IDENTIFIER")
+    signature = request.headers.get("GITHUB-PUBLIC-KEY-SIGNATURE")
+
+    verifier = request.find_service(
+        IGitHubTokenScanningPayloadVerifyService, context=None
```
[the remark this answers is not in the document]
ewjoachim: Verifying the payload is costly and very complex:

- we call GitHub's meta API to extract key IDs and keys
- we check that the key ID is in our known keys
- we check the cryptographic signature

There are a lot of moving parts, and I imagined we may want to have simpler alternatives:
- Caching the meta API call in some sort of storage
- Actually storing those in the code directly, and avoid calling GitHub altogether if we want to avoid that for some reason
- Change the crypto impelementation
- ...

I though this part both was well contained, and had quite a few moving parts, so that's why I tried it this way. See https://github.com/pypa/warehouse/pull/7124#issuecomment-568281765 and below for what led me to that :)

warehouse/accounts/services.py
```diff
@@ -759,3 +772,163 @@ def check_password(self, password, *, tags=None):
...
+            )
+            return False
+
+        self._metrics.increment("warehouse.token_leak.github.auth.success")
+        return True
+
+    def _get_headers(self):
+        if self._api_token:
+            return {"Authorization": f"token {self._api_token}"}
+        return {}
```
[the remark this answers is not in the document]
ewjoachim: As far as I know, without PAT, GitHub does rate-limiting based on IP, which means depending o how we deploy and what our externally seen IP is, we may be rate-limited du to calls from other people on the same datacenter.

Using a PAT is better, but it's not required. For example, if someone tries to use this code from a local environment, it's perfectly ok if they decide not to use PAT. This is what led me to having those 2 branches. I can add a comment if you want :)

warehouse/accounts/services.py
```diff
@@ -759,3 +772,163 @@ def check_password(self, password, *, tags=None):
...
+
+    def _retrieve_public_key_payload(self):
+        # TODO: cache ?
```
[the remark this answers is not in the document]
ewjoachim: I would agree, but honnestly I have no idea how often this would happen. Every minute ? Hour ? Day ? Week ?

Also, what would happen if someone purposefully (...or not?) posted millions of "things that look like a pypi token" ? If we check once per match, GitHub might end up thinking we're trying to DDOS them :)

warehouse/accounts/services.py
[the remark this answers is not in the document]
di: I guess I'm not sure what we'd want to do in production: it's my understanding that PATs are user-specific and can't be generated for organizations, so whose PAT should we use in prod?

warehouse/accounts/services.py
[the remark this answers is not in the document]
di: That's a fair point, we don't have a way to determine the request is legitimate before we get the public keys. That makes me think we probably _should_ cache this, which also should reduce the need for using a PAT with the request.

warehouse/accounts/views.py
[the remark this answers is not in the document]
ewjoachim: Ah, locale is not stored in the DB, got it

warehouse/accounts/services.py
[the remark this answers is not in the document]
ewjoachim: Hm, I’ll admit I’m not really sure. In my experience, organisations always had dedicated users for that, but it’s more of a habit than a rule of course. Also, I was under the impression that it was possible to use oauth apps for that but I didn’t investigate.

If it turns out it has to be someone’s token and there’s no « bot » Pypi user, and you wish not to create one, I guess all that’s left is no token. I’ve been bitten by the rate limit earlier so I tend to be mindful of that now, but if you think this won’t be a problem, I’m perfectly oK to drop it, as long as we’re aware of this, and if it breaks someday, we will know why.

warehouse/accounts/services.py
ewjoachim: Yep ! I always like to `raise BlaError() from exc` just to make it super duper extra easy to figure out what happened, and have the code know this is semi-expected.

warehouse/accounts/utils.py
ewjoachim: Yep, sounds nice! I’ll look at that.

warehouse/accounts/services.py
```diff
@@ -759,3 +772,163 @@ def check_password(self, password, *, tags=None):
...
+
+
+@implementer(IGitHubTokenScanningPayloadVerifyService)
+class NullGitHubTokenScanningPayloadVerifyService:
+    """
+    Doesn't really check anything, avoids HTTP calls, always validate
+    """
+
+    def verify(self, *, payload, key_id, signature):
+        return True
```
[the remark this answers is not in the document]
ewjoachim: I did it for local development, because I guess you don’t always want your local machine to call github. I may be missing the proper setup ?

warehouse/accounts/views.py
```diff
@@ -814,3 +818,52 @@ def edit_profile_button(user, request):
...
+    )
+
+    if not verifier.verify(payload=body, key_id=key_id, signature=signature):
+        request.response.status_int = 403
+        return {"error": "invalid signature"}
```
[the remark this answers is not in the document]
ewjoachim: I think github does’t care, it fires and forgets.

I’ve tried to put very minimal explainations In the payload to help whoever human person trying to debug this when it stops working someday, but honnestly, empty response would be fine to me, especially if there’s a side method for knowing what’s going on (logs,...)

warehouse/accounts/views.py
ewjoachim: I had (what I understood as) the opposite suggestion earlier in this review...

I’ll do my best to make everyone happy but if you can make sure that choice will be final, I’d hate to redo the same code a 4th time :)
For what it’s worth, it’s kinda what I had at first.

warehouse/accounts/services.py
ewjoachim: Hm, stange, when I submitted it, flake8 was happy. Probably a merge issue. I’ll make flake8 happy again.

warehouse/accounts/views.py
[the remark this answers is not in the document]
di: Ref: https://github.com/pypa/warehouse/issues/7583

warehouse/accounts/utils.py
ewjoachim: Hm, I’m 100% sure that when this code was written, I copy pasted this from elsewhere in the codebase, so either it was removed since then or something happened that I don’t understand, but I’m usually not in the habbit of adding unneeded deps. I’ll reove it.

warehouse/accounts/services.py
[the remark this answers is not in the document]
di: Maybe this is something we should talk to GitHub about what they expect users to do here? (I'm not on the original thread).

warehouse/accounts/services.py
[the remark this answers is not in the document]
di: In that case I would expect it to be configurable via `dev/environment` similar to: https://github.com/pypa/warehouse/blob/557ca0ece02f8570be7da9de7c5c1cc713d96ff4/dev/environment#L30

However as mentioned above, I think we should give some thought to whether this needs to be a service at all, i.e. do we imagine we'd want to use this environment variable to change this to some other implementer of `IGitHubTokenScanningPayloadVerifyService` in prod? I think we probably won't.

warehouse/macaroons/services.py
ewjoachim: I’ll look at the code more closely but if memory serves, find_macaroon looks for macaroon id but doesn’t check its signature.

Now it would be extremely unlikely that someone guesses a macaroon_id given it’s a UUID, but it’s not a secret per se. We could imagine that there is a way to get macaroon IDs through a side channel, and having a macaroon ID should not be sufficient to interact with this macaroon, that’s what signatures are for.

If we don’t check signature, it means someone who has access to a macaroon ID could revoke it by pushing on GitHub a macaroon with this ID and a bogus signature. I’m not tempted to explore this path, it seems much saner to check that a macaroon that we’ll revoke can only be revoked if it actually leaked.

warehouse/accounts/views.py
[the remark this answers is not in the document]
di: Might be worth checking w/ GitHub to see if they care at all about a response. For example, I'd hope that if they get a 500 they would retry the request, since it's somewhat important that we receive these. Specifically, if this is the case, I'd want to make sure that they consider any 200-level response a success (and not just `200 OK`).

warehouse/templates/email/token-compromised-leak/body.html
ewjoachim: I’m ok with that, just thought it was a possibility.

I agree that the page wouldn’t exist yet when the mail is sent if we learn about the breach, but by the time people read their emails and click, a few hours/days may have passed, so the page will likely exist.

That being said, again, no problem dropping that

warehouse/templates/email/token-compromised-leak/body.html
ewjoachim: Ok, I’llnkeep that in mind thanks !

warehouse/accounts/services.py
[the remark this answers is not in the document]
ewjoachim: Ok ! But I don’t know how caching works on Warehouse. Any pointer ?

warehouse/accounts/views.py
di: Are you referring to https://github.com/pypa/warehouse/pull/7124#discussion_r414072228? I think that discussion was more about whether it was possible, not whether it was advisable 🙂 

Since you've already done the work though, I think it's fine to keep it unless the other maintainers (@dstufft / @ewdurbin) object for some reason. I do think we should be more clear about it's purpose though, just `headers` is not descriptive enough.

warehouse/accounts/services.py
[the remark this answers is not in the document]
di: I don't believe we do anything like this at the moment unfortunately. This is (I think) the first time we're reaching out to a third-party service like this (that isn't on the frontend).

Perhaps just a simple in-memory TTL cache with https://pypi.org/project/cachetools/ would suffice?

warehouse/accounts/services.py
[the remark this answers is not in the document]
ewjoachim: Regarding that part, I'd love to have your opinion on https://github.com/pypa/warehouse/pull/7124#discussion_r453870408

warehouse/accounts/views.py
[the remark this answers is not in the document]
di: I agree that it makes sense to encapsulate this in a class or helper method, but a service is really for more generic things that we might want to someday swap out the underlying implementation.

The potential future changes you're describing mostly seem like things we'd just modify the service to support as necessary, rather than creating a new service entirely.

Like @dstufft mentioned before, a good example is our file storage. We want a generic interface to store files, and the ability to implement them as services that write to S3 or GCS or whatever, and the ability to swap them out via configuration whenever we need to.

Since this endpoint is very GitHub-specific, and the verifier is very GitHub-specific, I don't realistically see us ever replacing this with something else.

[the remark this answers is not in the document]
ewjoachim: Ok, I understand better now.

warehouse/routes.py
```diff
@@ -155,6 +155,11 @@ def includeme(config):
     config.add_route(
         "accounts.verify-email", "/account/verify-email/", domain=warehouse
     )
+    config.add_route(
+        "accounts.github-disclose-token",
+        "/account/disclose-token/github",
```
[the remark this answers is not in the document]
ewjoachim: Would that be `warehouse/integrations/github.py` or `warehouse/integrations/github/*.py` ? (I'm starting with the second one)

warehouse/macaroons/services.py
ewjoachim: (btw, https://github.com/pypa/warehouse/issues/7183)

warehouse/accounts/services.py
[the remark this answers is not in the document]
ewjoachim: I've asked github by e-mail, but I've tested in parallel, and unauthenticated calls, including to this URL, are limited to 60 calls / hour / IP. I think we can say this is clearly not enough. We'll need to cache the response and/or to make authenticated calls, but I don't think 60 calls/hour from the IP will be enough.

warehouse/accounts/views.py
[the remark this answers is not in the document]
ewjoachim: (Same, I've asked GitHub, let's wait for their answer)

warehouse/accounts/views.py
ewjoachim: > For example, it's not clear to me what a request with incorrect headers would get in response.

As far as I can tell, a view without those headers will not be considered during URL matching, and thus I'd expect 404.

warehouse/accounts/utils.py
ewjoachim: I'll remove it, but you may want to fix this in another ticket/PR:

https://github.com/pypa/warehouse/blob/b344984ee83126eb6fc541406d1f34719fad8b6b/warehouse/email/services.py#L36

di: Ah, good catch!

warehouse/accounts/services.py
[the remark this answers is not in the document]
ewjoachim: Ok, it's been 5 days and no answer so far. Let's assume token is important, according to field research, and make it mandatory. It will always be time to change our minds later.

And for how we'll get a token for production: the more I think about it, the more I think a dedicated account is probably the easiest way to go. The token doesn't even need special access to anything: it only accesses an open resource.

warehouse/accounts/views.py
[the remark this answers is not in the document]
ewjoachim: For now I'll go for "empty 200 / empty 400". We'll change this if GitHub gives us more info.

warehouse/accounts/services.py
ewjoachim: Ok found.

This line (predating the PR) actually disables flake8 for the whole file
https://github.com/pypa/warehouse/blob/696162baf8a0efc140b487ac888c8458fa8a9b9b/warehouse/accounts/services.py#L109

di: Ah, that looks like an issue with our auto formatter moving the comment, it should be on the following line(s).

ewjoachim: No this comment always acts on the whole file: https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#ignoring-entire-files

I've made a PR: https://github.com/pypa/warehouse/pull/8317

ewjoachim: The regex is sused to mach any line in the file, and ignore the whole file.

https://github.com/PyCQA/flake8/blob/076dfeee372affb1dbe849d12a1cc20f8ea37d40/src/flake8/defaults.py#L43
https://github.com/PyCQA/flake8/blob/94304de68480ac3007377404837130df72534ee3/src/flake8/processor.py#L386-L388

ewjoachim:
Okay ! @di I've addressed all your comments except the one about switching to celery. In dedicated commits, I made sure not to merge them with previous commits so that you can continue the review from where you were. When we're done, I can squash if you want.

What's left:
- [ ] I've avoided re-doing the translations (messages.pot) so far because with my setup, it's a bit complicated, but it's a generated file, I'll get it right when the rest is ok.
- [ ] The tests. I haven't touched the test after the big changes. I hope it will mostly be about moving tests more than writing tests, but we'll see
- [ ] and the celery part (and this time, it will involve writing new tests)

Am I missing something ? I'll be continuing to work on that, when I find some time :)

warehouse/accounts/utils.py
ewjoachim: So just to make sure I got it right:
- Everything that is done only once would be run by the view
- Everything that is done for every individual disclosure would be done in the celery task

Yes ? (I could also imagine having the view launch a main task and then 200, and the main task would do the validation, and create other tasks for each disclosure)

warehouse/integrations/github/utils.py
```diff
@@ -273,57 +273,61 @@ def _check_signature(self, payload, public_key, signature):
             ) from exc
 
 
-class TokenLeakAnalyzer:
-    def __init__(self, request):
-        self._request = request
-        self._metrics = self._request.find_service(IMetricsService, context=None)
-        self._macaroon_service = self._request.find_service(
-            IMacaroonService, context=None
+def _analyze_disclosure(request, disclosure_record, origin):
```
ewjoachim, line 276: Honnestly, I'd prefer if this method recieved the services it needs instead of the request, but `send_token_compromised_email_leak` requires passing a request. That leaves us with 3 choices:

- Pass the request and use find_service in the function, what I did but not a huge fan
- Pass the services AND the request. Seems unnatural to pass `x` and `f(x)` as arguments to a function. 
- Only pass the services: but then how to send the email ?

Would be easier if email sending used a service too.

ewjoachim:
Okay !

Took some time but I think we're at the end of the road.

- All the review comments were take into account
- Tests are at 100% coverage again
- Translations rebuilt
- Linters pass
- All TODOs are gone. All checkboxes in this PR (which is approaching 100 comments 🎉) are checked.

Note: I haven't been able to test the whole thing for real, this would require GitHub calling us and I don't know how to that.

Once review is good, I can redo the commits if you want, fixing up and squashing things for a cleaner history, but at this point, I have to admit that I'll be glad if we could call it a day.

di:
Response from GitHub about some of the open questions here:

>> Are we expected to use tokens of some sort to call https://api.github.com/meta/public_keys/secret_scanning to avoid rate-limiting or is this URL exempt of rate-limiting ? If this URL is NOT exempt, can you confirm that, without a token, our ability to get a response will depend on how many calls there were recently, from the same IP address ? If we'd rather avoid using a personal access token here, what would be the alternatives ? Or should we use a personal access token from a dedicated account ?
>
> Authenticated requests (associated with a user) have a higher rate limit than unauthenticated ones. AFAIK, Secret scanning does not have any special rate limiting rules. The general rules defined here should apply: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting
>
> The public key used to sign a hook is provided, and these keys do not change too often. We recommend caching the key locally, so it only needs to be fetched if not present in the cache, or when the key changes.
>
>> When processing GitHub's HTTP call, upon detection of a token leak, is there any change of behaviour on GitHub side that depends on the status code or body of the secret ? Or is the response completely ignored ? What if the response is 5XX, will the call be tried again later ?
>
> We do retry requests that do not receive a 2xx HTTP success response. Apart from that, the response code does not influence behaviour. Specific types of secrets may have custom processing applied to them after they are detected.

I think my preference would be to not use a PAT here and instead do some light caching (via Redis).

> Once review is good, I can redo the commits if you want, fixing up and squashing things for a cleaner history, but at this point, I have to admit that I'll be glad if we could call it a day.

I don't think this is necessary! We'll just squash this all together into one commit when merging anyways.

Really appreciate you seeing this through @ewjoachim! I think we're nearly there 😅

ewjoachim:
> I think my preference would be to not use a PAT here and instead do some light caching (via Redis).

I understand how easier for deployment and maintenance this will be, and in the end you'll be the judge. Just sharing my experience on the matter: it all depends on the deployment specifics and whether we'll be the only ones using the outbound IP. Without a PAT, we're limited to 60 calls / hour, for all the users of the same IP, which will include warehouse workers, but can also include e.g. many other unrelated tenants in the same public cloud, or this kind of stuff. If you're sure that we won't have to share our rate limit bucket, then depending on how many web workers we have, even an in-memory cache might suffice. If we're sharing the IP though, we're pretty much guaranteed that 95% calls will fail, so caching won't help.

di:
Sure. I think that's a question @ewdurbin can probably answer better than I can.

I also just realized we have https://github.com/browntruck which we could use to generate a PAT (although I'm not sure who has control of that account right now)

ewjoachim:
If the cache is used and the key wasn't found, we should bust the cache and retry before dclaring it a failure. I missed that part.

ewjoachim:
> If the cache is used and the key wasn't found, we should bust the cache and retry before declaring it a failure. I missed that part.

Done. Now the cache will be used if it exists and it's not too old, but if the key is not found in the cache, the "normal" way will be used. Also, added metrics for cache hits and misses to ease potential debugging.

ewjoachim:
I was under the impression that we were about to be ready. Is there anything more I could be doing ? @di ?

warehouse/templates/pages/help.html
```diff
@@ -122,7 +127,12 @@ <h2><a href="#my-account">{% trans %}My Account{% endtrans %}</a></h2>
         <ul>
           <li><a href="#verified-email">{{ verified_email() }}</a></li>
           <li><a href="#compromised-password">{{ compromised_password() }}</a></li>
+<<<<<<< HEAD
           <li><a href="#suspicious-activity">{{ suspicious_activity() }}</a></li>
+||||||| merged common ancestors
+=======
+          <li><a href="#compromised-token">{{ compromised_token() }}</a></li>
+>>>>>>> FAQ
```
di, line 135: Looks like a bad merge here.

di (approved):
LGTM once the bad merge is fixed. I'd like @ewdurbin to be able to review but if he isn't able to get to it, I'll merge by the end of the week.

ewjoachim:
Rebases, conflicts solved, all good !

di:
OK, one last small change to request: can you update this so that it's not required that `GITHUB_TOKEN` is set? I.e. if the env var is not present, this should just not attempt to include the `Authorization` header at all.

di:
(Also, looks like linting is failing because we're using a newer version of black now, and the translations need regenerated)

ewjoachim:
> OK, one last small change to request: can you update this so that it's not required that GITHUB_TOKEN is set? I.e. if the env var is not present, this should just not attempt to include the Authorization header at all.

Yep no problem :+1: (we did a full 360 from https://github.com/pypa/warehouse/pull/7124#discussion_r453686794 :tada: )

> (Also, looks like linting is failing because we're using a newer version of black now, and the translations need regenerated)
I'll rerun black. Translations: I regenerated them completely, I'm surprised.

ewjoachim:
ok, `make translations` does not create diff, neither does `make reformat`. The api token is now optional.

di:
Thanks @ewjoachim for all your work here!

Issue #7128, opened 2019-12-19
Primary Key Constraint missing on some models


ewdurbin:
The following models/tables have a UUID `id` column but are missing a Unique Index and Primary Key Constraint on the column:

- `File`/`release_files`
- `Dependency`/`release_dependencies `
- `Role`/`roles`

I'm not sure how we ended up in this state, but it seems something was missed in #4958 when [migrating to surrogate keys](https://github.com/pypa/warehouse/pull/4958/files#diff-b90527dff68c0ac1452ea267d46d85aa).

ewdurbin:
At least for `File`/`release_files` it appears to just be a relic from when it was [initially mapped into the ORM](https://github.com/pypa/warehouse/pull/383)

ewdurbin:
Since SQLAlchemy _believes_ these primary keys exist

```
>>> from sqlalchemy.inspection import inspect
>>> from warehouse.packaging.models import File
>>> inspect(File).primary_key
(Column('id', UUID(as_uuid=True), table=<release_files>, primary_key=True, nullable=False, server_default=DefaultClause(<sqlalchemy.sql.elements.TextClause object at 0x7fe5445588d0>, for_update=False)),)
>>> from warehouse.packaging.models import Dependency
>>> inspect(Dependency).primary_key
(Column('id', UUID(as_uuid=True), table=<release_dependencies>, primary_key=True, nullable=False, server_default=DefaultClause(<sqlalchemy.sql.elements.TextClause object at 0x7fe5445588d0>, for_update=False)),)
>>> from warehouse.packaging.models import Role
>>> inspect(Role).primary_key
(Column('id', UUID(as_uuid=True), table=<roles>, primary_key=True, nullable=False, server_default=DefaultClause(<sqlalchemy.sql.elements.TextClause object at 0x7fe5445588d0>, for_update=False)),)
```

but the underlying schema does not have them errors such as 

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.InvalidForeignKey) there is no unique constraint matching given keys for referenced table "release_files"
```

Arise when referencing them in relations.

PR #7129, opened 2019-12-19
add missing primary key constraints

ewdurbin:
addresses #7128


ewdurbin:
generated SQL:

```
ALTER TABLE release_files ADD PRIMARY KEY (id);

ALTER TABLE release_dependencies ADD PRIMARY KEY (id);

ALTER TABLE roles ADD PRIMARY KEY (id);

UPDATE alembic_version SET version_num='b5bb5d08543d' WHERE alembic_version.version_num = '08aedc089eaf';

COMMIT;
```

ewdurbin:
I've confirmed that this migration is rather quick, and syncs SQLAlchemy's view of the DB with reality, so I'm gonna go ahead and ship it.

Issue #7133, opened 2019-12-20
Add admin interface for malware check administration


xmunoz:
This issue is part of the work defined by [2019-Q4-PyPI Milestone 2](https://github.com/python/request-for/blob/master/2019-Q4-PyPI/RFP.md#milestone-2---systems-for-automated-detection-of-malicious-uploads).

Add an interface for an admin to view all the configured checks, and enable/disable checks.

PR #7134, opened 2019-12-20
Add admin interface to view and enable checks

xmunoz:
- Implement list, detail and change_state views (#7133)
- Add unit tests for check admin view

### DIFF
--- a/tests/unit/admin/views/test_checks.py
+++ b/tests/unit/admin/views/test_checks.py
@@ -10,12 +10,113 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import uuid
+
+import pretend
+import pytest
+
+from pyramid.httpexceptions import HTTPNotFound
+
 from warehouse.admin.views import checks as views
+from warehouse.malware.models import MalwareCheckState
 
 from ....common.db.malware import MalwareCheckFactory
 
 
-class TestGetChecks:
+class TestListChecks:
+    def test_get_checks_none(self, db_request):
+        assert views.get_checks(db_request) == {"checks": []}
+
     def test_get_checks(self, db_request):
         checks = [MalwareCheckFactory.create() for _ in range(10)]
         assert views.get_checks(db_request) == {"checks": checks}
+
+    def test_get_checks_different_versions(self, db_request):
+        checks = [MalwareCheckFactory.create() for _ in range(5)]
+        checks_same = [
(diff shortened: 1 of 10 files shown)


warehouse/admin/views/checks.py
```diff
@@ -0,0 +1,69 @@
...
+    renderer="admin/malware/checks/detail.html",
+    permission="moderator",
+    request_method="GET",
+    uses_session=True,
+)
+def get_check(request):
+    query = (
+        request.db.query(MalwareCheck)
+        .filter(MalwareCheck.name == request.matchdict["check_name"])
+        .order_by(MalwareCheck.created.desc())
```
ewdurbin: might make sense to order by revision, though one would hope that it has the same result.

xmunoz: Yup, I agree.

warehouse/admin/views/checks.py
```diff
@@ -0,0 +1,89 @@
...
+            request.db.query(MalwareCheck)
+            .filter(MalwareCheck.id == request.POST["id"])
+            .one()
+        )
+    except NoResultFound:
+        raise HTTPNotFound
+
+    try:
+        check.state = getattr(MalwareCheckState, request.POST["check_state"])
+    except AttributeError:
```
xmunoz: What do you think about also catching a `KeyError` here, for the case where `"check_state not in "request.POST`?

woodruffw: +1 from me on catching `KeyError` 🙂

tests/unit/admin/views/test_checks.py
```diff
@@ -0,0 +1,122 @@
...
+from pyramid.httpexceptions import HTTPNotFound
+
+from warehouse.admin.views import checks as views
+from warehouse.malware.models import MalwareCheckState
+
+from ....common.db.malware import MalwareCheckFactory
+
+
+class TestListChecks:
+    def test_get_checks_none(self, db_request):
```
woodruffw: Nit: You could probably use `pytest.mark.parametrize` here to reduce these independent methods into a single `test_get_checks`, but the current structure also LGTM 🙂

xmunoz: I looked a bit into parameterization, and I don't think it would work cleanly here because I need to call a `Factory.create()` method to generate the checks. It seems better suited for situations where the arguments are static objects (strings, lists, etc). Thanks for the tip though! This is also a super common testing pattern in golang.

Issue #7093, opened 2019-12-13
Implement code and set up infrastructure for check execution


xmunoz:
This issue is part of the work defined by [2019-Q4-PyPI Milestone 2](https://github.com/python/request-for/blob/master/2019-Q4-PyPI/RFP.md#milestone-2---systems-for-automated-detection-of-malicious-uploads).

The malware checks should be dispatched to a task queue (such as Celery) and run as tasks in an isolated environment (e.g. as callbacks to an AWS Lambda instance). Set up the infrastructure to run the checks in this way.

xmunoz:
Of note: the max run time for a lambda instance is [15 minutes](https://aws.amazon.com/about-aws/whats-new/2018/10/aws-lambda-supports-functions-that-can-run-up-to-15-minutes/).

PR #7160, opened 2019-12-26
Add initial hook-based check execution mechanism

xmunoz:
Progress on #7093

### DIFF
--- a/dev/environment
+++ b/dev/environment
@@ -29,6 +29,7 @@ MAIL_BACKEND=warehouse.email.services.SMTPEmailSender host=smtp port=2525 ssl=fa
 
 BREACHED_PASSWORDS=warehouse.accounts.NullPasswordBreachedService
 
+#TODO: change this to PrinterMalwareCheckService before deploy
 MALWARE_CHECK_BACKEND=warehouse.malware.services.DatabaseMalwareCheckService
 
 METRICS_BACKEND=warehouse.metrics.DataDogMetrics host=notdatadog
--- a/tests/unit/malware/test_tasks.py
+++ b/tests/unit/malware/test_tasks.py
@@ -14,7 +14,10 @@ import celery
 import pretend
 import pytest
 
-# from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm.exc import NoResultFound
+
+import warehouse.malware.checks as checks
+
 from warehouse.malware.models import MalwareVerdict
 from warehouse.malware.tasks import run_check
 
@@ -34,9 +37,21 @@ def test_run_check(monkeypatch, db_request):
 
 
 def test_run_check_missing_check_id(monkeypatch, db_session):
-    task = pretend.stub(
-        retry=pretend.call_recorder(pretend.raiser(celery.exceptions.Retry))
-    )
+    exc = NoResultFound("No row was found for one()")
+
+    class FakeMalwareCheck:
+        def __init__(self, db):
+            raise exc
+
+    class Task:
+        @staticmethod
+        @pretend.call_recorder
+        def retry(exc):
+            raise celery.exceptions.Retry
+
+    task = Task()
+
+    checks.FakeMalwareCheck = FakeMalwareCheck
 
     request = pretend.stub(
         db=db_session,
@@ -44,19 +59,19 @@ def test_run_check_missing_check_id(monkeypatch, db_session):
     )
 
     with pytest.raises(celery.exceptions.Retry):
-        run_check(task, request, "ExampleCheck", "d03d75d1-2511-4a8b-9759-62294a6fe3a7")
+        run_check(
+            task, request, "FakeMalwareCheck", "d03d75d1-2511-4a8b-9759-62294a6fe3a7"
+        )
 
     assert request.log.error.calls == [
         pretend.call(
-            "Error executing check %s: %s", "ExampleCheck", "No row was found for one()"
+            "Error executing check %s: %s",
+            "FakeMalwareCheck",
+            "No row was found for one()",
         )
     ]
 
-    """
-    assert task.retry.calls == [
-        pretend.call(exc=NoResultFound("No row was found for one()"))
-    ]
-    """
+    assert task.retry.calls == [pretend.call(exc=exc)]
 
 
 def test_run_check_missing_check(db_request):
--- a/tests/unit/malware/test_tasks.py
+++ b/tests/unit/malware/test_tasks.py
@@ -1,0 +1,70 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import celery
+import pretend
+import pytest
+
+# from sqlalchemy.orm.exc import NoResultFound
+from warehouse.malware.models import MalwareVerdict
+from warehouse.malware.tasks import run_check
+
+from ...common.db.malware import MalwareCheckFactory
+from ...common.db.packaging import FileFactory, ProjectFactory, ReleaseFactory
+
+
+def test_run_check(monkeypatch, db_request):
+    project = ProjectFactory.create(name="foo")
+    release = ReleaseFactory.create(project=project)
+    file0 = FileFactory.create(release=release, filename="foo.bar")
+    MalwareCheckFactory.create(name="ExampleCheck", state="enabled")
+
+    task = pretend.stub()
+    run_check(task, db_request, "ExampleCheck", file0.id)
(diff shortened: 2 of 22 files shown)


ewdurbin:
One option rather than instantiating checks directly from the upload method would be to add a db listener:

```diff
diff --git a/warehouse/malware/__init__.py b/warehouse/malware/__init__.py
index 0d989ac0..b4b4973b 100644
--- a/warehouse/malware/__init__.py
+++ b/warehouse/malware/__init__.py
@@ -10,8 +10,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from warehouse import db
 from warehouse.malware.interfaces import MalwareCheckService
 from warehouse.malware.services import database_malware_check_factory
+from warehouse.packaging.models import File, Project, Release
+
+
+@db.listens_for(db.Session, "after_flush")
+def determine_malware_checks(config, session, flush_context):
+
+    malware_checks = session.info.setdefault("warehouse.malware.checks", set())
+
+    for obj in session.new:
+        if obj.__class__ in (File,):
+            malware_checks.update([f"example_file_check:{obj.id}"])
+        if obj.__class__ in (Release,):
+            malware_checks.update([f"example_release_check:{obj.id}"])
+        if obj.__class__ in (Project,):
+            malware_checks.update([f"example_project_check:{obj.id}"])
+
+
+@db.listens_for(db.Session, "after_commit")
+def queue_malware_checks(config, session):
+
+    malware_checks = session.info.setdefault("warehouse.malware.checks", set())
+
+    print(malware_checks)
 
 
 def includeme(config):
```

This is similar to how our [cache purging](https://github.com/pypa/warehouse/blob/3edcc330ad79660cc430e797a2bfbb84755b7161/warehouse/cache/origin/__init__.py#L26-L55) works. The example above is a bit rough, but implementing it in this way would allow for instantiating the checks from the event and even instantiating different checks based on what changed.

warehouse/malware/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+
+import warehouse.malware.checks as checks
+
+from warehouse.tasks import task
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
+def run_check(task, request, check_name, check_id, file_id):
+    try:
+        check = getattr(checks, check_name)(check_id)
```
woodruffw: Just to make sure I understand: `check_name` here would be something like `ExampleCheck` (i.e. the class name), right?

So part of the development process for new checks includes explicitly importing them into the `checks` namespace (which SGTM, just want to make sure I'm on the same page 😄)

warehouse/malware/services.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@implementer(MalwareCheckService)
+class DatabaseMalwareCheckService:
+    def __init__(self, session):
+        self.db = session
+
+    @functools.lru_cache()
+    def get_enabled_checks(self):
+        return self.db.query(MalwareCheck).filter(MalwareCheck.state == "enabled").all()
```
woodruffw: Nitpick: Maybe `MalwareCheckState.Enabled` instead of a string literal?

warehouse/malware/tasks.py
xmunoz: Yes, that's what I'm thinking now. Open to other suggestions though!

warehouse/malware/checks/example.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def run(self, request, file_id):
+        verdict = MalwareVerdict(
+            check_id=self.id,
+            file_id=file_id,
+            classification=VerdictClassification.benign,
+            confidence=VerdictConfidence.High,
+            message="Nothing to see here!",
+        )
+        request.db.add(verdict)
```
woodruffw: As a general design question: What do you think about performing the database transactions behind the scenes and exposing an imperative API? Something like this:

```python3
class MalwareCheckBase:
  def __init__(self):
    self._verdicts = []

  def verdict(self, **kwargs):
    self._verdicts.append(MalwareVerdict(**kwargs))

  def perform(self, ...):
    self.run(...)
    request.db.add_all(self._verdicts)

  ...
```

(With `perform` being the new top-level method submitted to the task queue)

With individual checks doing something like:

```python3
def run(self, request, file_id):
  self.verdict(...)
```

(or `deliver_verdict` or similar if just `verdict` isn't imperative enough)

The main advantages to this (that I can think of) is that it hides the involvement of the DB from check implementors, and gives us the ability to do additional bookkeeping (or future refactoring) within the `verdict` method.

warehouse/malware/tasks.py
woodruffw: That sounds good to me! I've seen and done hacky things with dynamic loading of classes into a fake namespace before and that's consistently been a mistake, so leveraging the actual Python module model sounds good.

warehouse/malware/checks/example.py
xmunoz: Yes! I like that a lot. This PR is just a PoC, but this is definitely the direction that I want to go in.

tests/unit/malware/test_tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
+    assert request.log.error.calls == [
+        pretend.call(
+            "Error executing check %s: %s", "ExampleCheck", "No row was found for one()"
+        )
+    ]
+
+    """
+    assert task.retry.calls == [
+        pretend.call(exc=NoResultFound("No row was found for one()"))
```
ewdurbin: This comparison will not work since it is comparing instances of `NoResultFound`, one instantiated by the code under test and one instantiated in test (in this line). Since those will be different objects, the comparison will fail.

For this test to work, whatever is raising the exception in the code would need to be mocked to raise _the same_ exception object. See [this test](https://github.com/pypa/warehouse/blob/64d223a255e0150232a33ebdb1b737d9a2fd5c5d/tests/unit/email/test_init.py#L214-L234) for an example.

xmunoz: Thanks! I just pushed an update.

warehouse/malware/__init__.py
```diff
@@ -9,3 +9,51 @@
...
+            obj.__class__.__name__
+            for obj in session.new
+            if obj.__class__.__name__ in utils.valid_check_types()
+        ]
+    ):
+        return
+
+    malware_checks = session.info.setdefault("warehouse.malware.checks", set())
+    enabled_checks = utils.get_enabled_checks(session)
+    for obj in session.new:
```
ewdurbin: This assumes all objects in `session.new` are _also_ an object that we run checks on. Should it filter out objects that we don't care about? It seems like we have the potential for a KeyError on the next line when `enabled_checks[obj.__class__.__name__]` is called.

warehouse/malware/tasks.py
```diff
@@ -0,0 +1,26 @@
...
+
+import warehouse.malware.checks as checks
+
+from warehouse.tasks import task
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
+def run_check(task, request, check_name, obj_id):
+    try:
+        check = getattr(checks, check_name)(request.db)
```
ewdurbin: It probably shouldn't happen, but we should probably watch for an `AttributeError` being raised here in the event that a `check_name` is not found for some reason.

warehouse/malware/utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_enabled_checks(session):
+    checks = (
+        session.query(MalwareCheck.name, MalwareCheck.hooked_object)
+        .filter(MalwareCheck.check_type == MalwareCheckType.event_hook)
+        .filter(MalwareCheck.state == MalwareCheckState.enabled)
+        .all()
+    )
+    results = defaultdict(list)
```
ewdurbin: Aha, I think this negates my concern about a KeyError in the event listener :)

ewdurbin:
I'm not sure that any of the above concerns are blockers to merge here.

warehouse/malware/utils.py
xmunoz: Yeah, the `defaultdict` will just return an empty list for a missing key, so we should raise a `KeyError` on `enabled_checks[obj.__class__.__name__]`.

warehouse/malware/__init__.py
xmunoz: Answered below.

warehouse/malware/tasks.py
xmunoz: I agree that there needs to be more robust error handling here. Maybe we can discuss that offline.

warehouse/malware/utils.py
ewdurbin: Using a defaultdict here certainly makes what's going on in the event listener a bit implicit, but it works. I'm torn on doing it this way versus filtering in `determine_malware_checks`.

dev/environment
```diff
@@ -29,6 +29,8 @@ MAIL_BACKEND=warehouse.email.services.SMTPEmailSender host=smtp port=2525 ssl=fa
 
 BREACHED_PASSWORDS=warehouse.accounts.NullPasswordBreachedService
 
+MALWARE_CHECK_BACKEND=warehouse.malware.services.DatabaseMalwareCheckService
```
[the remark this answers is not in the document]
xmunoz: Added.

warehouse/malware/utils.py
xmunoz: I'll update it to use `.get` instead.

Issue #7143, opened 2019-12-22
"Invalid credentials" when revoking a token


ewjoachim:
**Describe the bug**
I cannot revoke a token I have created, because it asks for my password and refuses to accept it. I'm using a password manager so low risk of typo, and I've tried displaying the password and they're the same. I've tried logging out and in with the same password and it works

**Expected behavior**
The crenetial check on token revocation should work with my password

**To Reproduce**
- Create a token
- From the creation screen, click on the remove button
- Type your password and validate

- Additionally, try removing it from your profile page.

**My Platform**

Firefox 71, Ubuntu

ewdurbin:
I have been unable to reproduce this locally at this point.

It points to something going wrong when [validating the submitted form](https://github.com/pypa/warehouse/blob/64d223a255e0150232a33ebdb1b737d9a2fd5c5d/warehouse/manage/views.py#L725) that isn't necessarily associated with the credential supplied.

ewdurbin:
@ewjoachim and @JokerQyou, is there any chance that a browser extension is modifying the contents of the form? In particular the `macaroon_id` input which is hidden. If the `csrf_token` input was being manipulated you'd get a different error.

JokerQyou:
I've tried with `test.pypi.org` and it worked fine. This issue is only reproducible with `pypi.org`.

As for the `csrf_token` value, I've checked with the developer tool, the final value in the POST request matches the initial value received from the server. See the screenshots:

![Initial CSRF token value received from server](https://user-images.githubusercontent.com/1465267/71678158-77322580-2dbf-11ea-98a2-319b748e87b7.png)

![CSRF token sent in the POST request](https://user-images.githubusercontent.com/1465267/71678216-98931180-2dbf-11ea-892b-8fac5db0f243.png)

JokerQyou:
Oh, sorry I misread your comment. I checked the `macaroon_id` field and it was indeed modified by the password manager! It was incorrectly filled with my username. If I manually copy and paste the password, the deletion works fine.

ewjoachim:
Hm, that's interesting. I reproduce if I let my password manager, bitwarden, fill the password field for me. If I copy/paste manually, it works. I'm going to try and explore how the payload is different in both cases.

ewdurbin:
This should be resolved. Can either of you confirm?

JokerQyou:
Just tested on `pypi.org` and it's fine now.

PR #7174, opened 2020-01-02
token revocation: disable autocomplete for hidden macaroon_id field

ewdurbin:
as discovered in #7143, some password managers are being overzealous and filling this input as the username.

### DIFF
--- a/warehouse/templates/manage/account.html
+++ b/warehouse/templates/manage/account.html
@@ -208,7 +208,7 @@
       {% set action=request.route_path('manage.account.token') %}
       {% set confirm_button_label=gettext("Remove API token") %}
       {% set extra_fields %}
-        <input type="text" name="macaroon_id" value="{{ macaroon.id }}" autocomplete="off" hidden>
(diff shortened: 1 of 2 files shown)


ewjoachim:
I think autocomplete won't change anything. The password field just below has the same properties and is filled by all password managers.

But I wonder why it's a `<input type="text" hidden>` rather than a `<input type="hidden">`.

BTW, there are 2 places to fix:
- https://github.com/pypa/warehouse/blob/ed618a928b945cb48631cc7cd7f9e54e9800e413/warehouse/templates/manage/token.html#L56
- https://github.com/pypa/warehouse/blob/4a51c7a2e1ddd7d0765c7cee9b7fdf23e81d1c1b/warehouse/templates/manage/account.html#L211

JokerQyou:
@ewjoachim I think that's the point. The `type="text"` part confuses Bitwarden, and it fills the field with a `User name`, because this field is the first `text` field in the form. If I manually edit the HTML via developer tool, changing `type` to `hidden`, this deletion works smoothly with Bitwarden autofill.

PR #7190, opened 2020-01-06
Add malware check syncing mechanism

### LINKED ISSUE #7097: Write db migration tooling for user-contributed malware checks
This issue is part of the work defined by [2019-Q4-PyPI Milestone 2](https://github.com/python/request-for/blob/master/2019-Q4-PyPI/RFP.md#milestone-2---systems-for-automated-detection-of-malicious-uploads).

User-contributed malware checks should be automatically aggregated and inserted into the database upon deployment, and re-inserted after any change to that code.

xmunoz:
Fixes #7097.

### DIFF
--- a/warehouse/malware/tasks.py
+++ b/warehouse/malware/tasks.py
@@ -36,30 +36,32 @@ def sync_checks(task, request):
 
     all_checks = request.db.query(MalwareCheck).all()
     active_checks = {}
-    wiped_out = {}
+    wiped_out_checks = {}
     for check in all_checks:
         if not check.is_stale:
             if check.state == MalwareCheckState.wiped_out:
-                wiped_out[check.name] = check
+                wiped_out_checks[check.name] = check
             else:
                 active_checks[check.name] = check
 
     if len(active_checks) > len(code_checks):
+        code_check_names = set([name for name, cls in code_checks])
+        missing = ", ".join(set(active_checks.keys()) - code_check_names)
         request.log.error(
             "Found %d active checks in the db, but only %d checks in \
-                code. Please manually move superfluous checks to the wiped_out state \
-                in the check admin."
-            % (len(active_checks), len(code_checks))
+code. Please manually move superfluous checks to the wiped_out state \
+in the check admin: %s"
+            % (len(active_checks), len(code_checks), missing)
         )
         raise Exception("Mismatch between number of db checks and code checks.")
 
     for check_name, check_class in code_checks:
         check = getattr(checks, check_name)
 
-        if wiped_out.get(check_name):
+        if wiped_out_checks.get(check_name):
             request.log.error(
-                "%s is wiped_out and cannot be synced. \
-                    Please remove check from codebase."
+                "%s is wiped_out and cannot be synced. Please remove check from \
+codebase."
                 % check_name
             )
             continue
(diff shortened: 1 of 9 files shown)


warehouse/malware/utils.py
```diff
@@ -27,6 +28,15 @@ def valid_check_types():
     return set([t.value for t in MalwareCheckObjectType])
 
 
+def get_check_fields(check):
```
ewdurbin: I'm not sure why, but I feel a little apprehensive passing all "non-dunder"/non-routine class attributes to the `MalwareCheck` instantiation above

```python
            request.log.info("Adding new %s to the database." % check_name)
             fields = get_check_fields(check)
             request.db.add(MalwareCheck(**fields))
```

It kinda stinks to hardcode the specific class attributes we're interested in, but could have the benefit of allowing us to implement a `validate_check_definition` or similar utility that could ensure that checks conform over time.

I'm imagining such a utility could ensure that all objects subclassing `MalwareCheckBase` contain the specific attributes and methods the interface defines.

warehouse/malware/tasks.py
```diff
@@ -24,3 +27,60 @@ def run_check(task, request, check_name, obj_id):
     except Exception as exc:
         request.log.error("Error executing check %s: %s", check_name, str(exc))
         raise task.retry(exc=exc)
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
+def sync_checks(task, request):
+    code_checks = inspect.getmembers(checks, inspect.isclass)
```
ewdurbin: I think this could be more specific by calling `MalwareCheckBase.__subclasses__()` rather than relying on `warehouse/malware/checks/__init__.py` to import.

Either that or verifying that the the classes found in the `checks` module are subclasses of `MalwareBaseCheck` directly.

warehouse/malware/tasks.py
```diff
@@ -24,3 +27,60 @@ def run_check(task, request, check_name, obj_id):
...
+        if not check.is_stale:
+            if check.state == MalwareCheckState.wiped_out:
+                wiped_out[check.name] = check
+            else:
+                active_checks[check.name] = check
+
+    if len(active_checks) > len(code_checks):
+        request.log.error(
+            "Found %d active checks in the db, but only %d checks in \
+                code. Please manually move superfluous checks to the wiped_out state \
```
ewdurbin: +1 to logging and crashing out, It would be great to note which checks are being considered superfluous in the error message. could a comparison of the sets of `name` and/or `version` provide this?

warehouse/malware/tasks.py
xmunoz: I like the `__subclassess__()` suggestion.

warehouse/malware/utils.py
xmunoz: This could definitely be used maliciously, but has a future-proof/convenience benefits. I'll do the hard-coding thing.

warehouse/malware/tasks.py
xmunoz: Sure!

warehouse/malware/tasks.py
xmunoz: There is some difficulty with the tests if we add this constraint to the task. Namely, once something is created and added as a subclass, it will persist for the entire execution of the program. This breaks the hermetic isolation of tests. I have preserved the original logic, but added something to `check_tests` to verify that every class exported from `checks` is, in fact, a subclass of `MalwareCheckBase`.

Related:
https://mail.python.org/pipermail/python-list/2003-August/210297.html

PR #7196, opened 2020-01-07
Refactor MalwareCheckBase. Fixes #7091.

### LINKED ISSUE #7091: Add base class for new malware checks
This issue is part of the work defined by [2019-Q4-PyPI Milestone 2](https://github.com/python/request-for/blob/master/2019-Q4-PyPI/RFP.md#milestone-2---systems-for-automated-detection-of-malicious-uploads).

Write a base class for the malware checks. This base class will be inherited by the event-based and schedule-based check classes, and may be used to aggregate all malware checks and update checks at release time.

xmunoz:
Add Foreign Keys in MalwareVerdicts for other types of objects
(Releases, Projects).

### DIFF
--- a/warehouse/malware/checks/base.py
+++ b/warehouse/malware/checks/base.py
@@ -10,7 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from warehouse.malware.models import MalwareCheck, MalwareCheckState
+from warehouse.malware.models import MalwareCheck, MalwareCheckState, MalwareVerdict
 
 
 class MalwareCheckBase:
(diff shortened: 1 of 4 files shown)


warehouse/malware/checks/base.py
```diff
@@ -10,18 +10,30 @@
...
 
-from warehouse.malware.models import MalwareCheck, MalwareCheckState
+from warehouse.malware.models import MalwareCheck, MalwareCheckState, MalwareVerdict
 
 
 class MalwareCheckBase:
     def __init__(self, db):
         self.db = db
         self._name = self.__class__.__name__
         self._load_check()
```
woodruffw: Nit: Maybe move this attr above the `_load_check()` call, just for consistency with the other attrs?

warehouse/malware/checks/base.py
```diff
@@ -10,18 +10,30 @@
...
+        self._verdicts = []
+
+    def add_verdict(self, verdict):
```
woodruffw: What do you think about making `verdict` into a kwargs instead of a direct dictionary? That would make it consistent with the `record_event` API.

xmunoz: Sure!

warehouse/malware/checks/base.py
xmunoz: I _could_ do this, but I use `self._name` in `_load_check` to look up the id.

woodruffw: Ah! I missed that, sorry. In that case this LGTM 🙂

PR #7202, opened 2020-01-07
Add wipe-out functionality

xmunoz:
Related: #7133

### DIFF
--- a/tests/common/db/malware.py
+++ b/tests/common/db/malware.py
@@ -37,9 +37,9 @@ class MalwareCheckFactory(WarehouseFactory):
     version = 1
     short_description = factory.fuzzy.FuzzyText(length=80)
     long_description = factory.fuzzy.FuzzyText(length=300)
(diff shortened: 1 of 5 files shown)


tests/common/db/malware.py
```diff
@@ -39,3 +43,14 @@ class Meta:
...
     )
+
+
+class MalwareVerdictFactory(WarehouseFactory):
+    class Meta:
+        model = MalwareVerdict
+
+    check = factory.SubFactory(MalwareCheckFactory)
+    release_file = factory.SubFactory(FileFactory)
+    classification = factory.fuzzy.FuzzyChoice([e for e in VerdictClassification])
```
woodruffw: Looked this up out of curiosity and thought I'd share: `list(VerdictClassification)` _should_ also work here, although I see no reason for a change 😄

xmunoz: Ooo yeah I like that. Let me see if I can change it everywhere.

PR #7203, opened 2020-01-07
Add additional email validation

di:
We later try to turn the address into an `Address`, so we should do it as soon as possible in the validation of the form to determine whether it will fail later or not.

(Also moves the domain checking up earlier in the validation so we can bail out before needing to make an additional query).


di:
This is for https://sentry.io/organizations/python-software-foundation/issues/1423903897/

Issue #7094, opened 2019-12-13
Implement malware check backfill functionality


xmunoz:
This issue is part of the work defined by [2019-Q4-PyPI Milestone 2](https://github.com/python/request-for/blob/master/2019-Q4-PyPI/RFP.md#milestone-2---systems-for-automated-detection-of-malicious-uploads).

Implement the functionality to backfill data for a malware check. This functionality should include the ability to:
- Run a check on the entire corpus of PyPI data
- Run a check on a sampled subset of files to evaluate the efficacy of a check

xmunoz:
What was settled upon was punting on the full backfill functionality for now, and instead running an evaluation on a fixed subset of each object type for hooked checks. Currently, the subset is limited to 10k elements: https://github.com/pypa/warehouse/blob/master/warehouse/admin/views/checks.py#L20

PR #7232, opened 2020-01-14
Add backfill functionality to check admin #7094

xmunoz:
- Add backfill task
- Change lookup of checks to check_name instead of id
- Load checks that are also in "evaluation" state

### DIFF
--- a/warehouse/admin/templates/admin/malware/checks/detail.html
+++ b/warehouse/admin/templates/admin/malware/checks/detail.html
@@ -68,14 +68,11 @@
   <div class="box box-primary">
     <div class="box-header with-border">
       <h3 class="box-title">Run Backfill</h3>
-      <p>Please select the percentage of the corpus that you'd like to sample.</p>
     </div>
     <form method="POST" action="{{ request.route_path('admin.checks.run_backfill', check_name=check.name) }}">
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
       <div class="box-body">
-        <div class="form-group col-sm-4">
-          <input class="form-control" type="number" value="1" name="sample-rate">
-        </div>
+        <p>Run this check against 10,000 {{ check.hooked_object.value }}s, selected at random. This is used to evaluate the efficacy of a check.</p>
         <div class="pull-right col-sm-4">
           <button type="submit" class="btn btn-primary pull-right">Run</button>
         </div>
--- a/warehouse/admin/templates/admin/malware/checks/detail.html
+++ b/warehouse/admin/templates/admin/malware/checks/detail.html
@@ -67,7 +67,7 @@
   </div>
   <div class="box box-primary">
     <div class="box-header with-border">
-      <h3 class="box-title">Run Backfill</h3>
+      <h3 class="box-title">Run Evaluation</h3>
     </div>
     <form method="POST" action="{{ request.route_path('admin.checks.run_backfill', check_name=check.name) }}">
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
--- a/warehouse/admin/templates/admin/malware/checks/detail.html
+++ b/warehouse/admin/templates/admin/malware/checks/detail.html
@@ -48,11 +48,10 @@
       <h3 class="box-title">Change State</h3>
     </div>
     <form method="POST" action="{{ request.route_path('admin.checks.change_state', check_name=check.name) }}">
-      <input type="hidden" name="id" value="{{ check.id }}">
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
       <div class="box-body">
         <div class="form-group col-sm-4">
-          <select name="check_state" id="check_state">
+          <select class="form-control" name="check_state">
             {% for state in states %}
             <option value="{{ state.value }}" {{'disabled selected' if check.state == state else ''}}>
               {{ state.value }}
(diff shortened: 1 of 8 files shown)


xmunoz:
Failing on coverage, but I wanted get input about my implementation of the [`backfill` function](https://github.com/pypa/warehouse/pull/7232/files#diff-f38ebe83acd6404974dcfd5d049a163eR35). There are two ways that this could run:
- Calling the `run_task` method directly, which causes the backfill task to be long-running, synchronously executing each `run_check` call.
- Having the backfill task enqueue a bunch of `run_check` tasks.

I have the first implementation in this PR for a few reasons. On my local machine, there appears to be a huge overhead incurred by enqueue thousands of tasks. To execute a backfill on 10,000 items took about 15 seconds with the first method, and about 5 minutes with the second. This overhead possibly goes away on production. My other reasons for preferring the first method:
 - It's easy to kill a single, rogue long-running task, and more difficult to play task whack-a-mole.
- Each `run_check` executes a few read and at least one write db query. This could potentially cause lock contention on the database, requiring task whack-a-mole...
- There may be a concern about the rate at which certain types of checks run, e.g. ones that query external APIs. Backfill checks executing synchronously means we get rate-limiting for free!

Let me know what you think!

warehouse/admin/views/checks.py
```diff
@@ -43,19 +43,61 @@ def get_checks(request):
...
+    if check.state not in (MalwareCheckState.enabled, MalwareCheckState.evaluation):
+        request.session.flash(
+            f"Check must be in 'enabled' or 'evaluation' state to run a backfill.",
+            queue="error",
+        )
+        return HTTPSeeOther(
+            request.route_path("admin.checks.detail", check_name=check.name)
+        )
+
+    if sample <= 0 or sample > 100:
```
woodruffw: It might make sense to have a dedicated form for validation here, instead of performing the validation in the request handler.  It looks like the admin views don't follow the same pattern of dedicated forms, though, so feel free to ignore.

warehouse/malware/tasks.py
```diff
@@ -29,6 +32,23 @@ def run_check(task, request, check_name, obj_id):
         raise task.retry(exc=exc)
 
 
+@task(bind=True, ignore_result=True, acks_late=True)
+def backfill(task, request, check_name, sample_rate=1):
```
woodruffw: A docstring for this method would be nice (particularly mentioning that `sample_rate` is a ratio in `[0, 1]` and not in `[0, 100]`, which is what I thought before I noticed the `sample * 0.01` in the request handler 😅

warehouse/admin/views/checks.py
xmunoz: I do think it makes sense to implement the form validation found in the user-facing site, but maybe not in this PR. This will be needed when I add advanced filtering capabilities in the verdicts view, so I'll add it here then as well.

ewdurbin: It's a mix in admin for sure, but see `admin/views/users.py` for examples of a form with validators (EmailForm and UserForm)

warehouse/admin/views/checks.py
```diff
@@ -67,18 +109,16 @@ def get_check(request):
     require_csrf=True,
 )
 def change_check_state(request):
+    check = get_check_by_name(request.db, request.matchdict["check_name"])
+
```
ewdurbin: For viewing... fetching by name seems reasonable to me, but for submitting a state change fetching by ID seems preferable. Could consider a final sanity check that the check being modified is the latest?

xmunoz: I did it this way because I view it as safer. A user of the web admin should never be able to craft a request that modifies the state of an arbitrary check via `id`. This way, the view logic determines which check is the correct entry to modify.

ewdurbin:
Sorry, I missed this comment, and the approach in the backfill task itself...

> Failing on coverage, but I wanted get input about my implementation of the [`backfill` function](https://github.com/pypa/warehouse/pull/7232/files#diff-f38ebe83acd6404974dcfd5d049a163eR35). There are two ways that this could run:
> 
> * Calling the `run_task` method directly, which causes the backfill task to be long-running, synchronously executing each `run_check` call.
> * Having the backfill task enqueue a bunch of `run_check` tasks.
> 
> I have the first implementation in this PR for a few reasons. On my local machine, there appears to be a huge overhead incurred by enqueue thousands of tasks. To execute a backfill on 10,000 items took about 15 seconds with the first method, and about 5 minutes with the second. This overhead possibly goes away on production. My other reasons for preferring the first method:

I think the primary reason why the first approach is significantly more performant is that we currently aren't _doing_ anything in any of the checks. I suspect that if a check were preforming more substantial work it would likely be problematic.

The second approach is likely more feasible in the long run, though perhaps not via the web UI due to timeouts. I doubt the overhead would disappear in prod, as enqueueing tasks still require network calls to the queue.
 
> * It's easy to kill a single, rogue long-running task, and more difficult to play task whack-a-mole.
> * Each `run_check` executes a few read and at least one write db query. This could potentially cause lock contention on the database, requiring task whack-a-mole...
> * There may be a concern about the rate at which certain types of checks run, e.g. ones that query external APIs. Backfill checks executing synchronously means we get rate-limiting for free!

These are all reasonable benefits of the first approach, but it looks to me like a better approach would be to strip out the ability to do a full backfill from the web and make it a CLI entrypoint under the malware module. Perhaps this view should only allow kicking off evaluations with very conservative corpus sizes.

xmunoz:
> The second approach is likely more feasible in the long run, though perhaps not via the web UI due to timeouts. 
I don't understand the concern about the web UI. The admin simply enqueues the backfill task and returns view to the user. The backfill execution happens asynchronously.

ewdurbin:
> I don't understand the concern about the web UI. The admin simply enqueues the backfill task and returns a view to the user. The backfill execution happens asynchronously. I don't think timeouts are a concern.

I got mixed up where the iteration was occurring. You're right.

xmunoz:
After more discussion around scalability, we are removing the ability to run full backfills from the project scope, and limited the admin triggered backfills to a fixed size for check evaluation purposes. Update pending.

warehouse/admin/templates/admin/malware/checks/detail.html
```diff
@@ -66,5 +65,19 @@ <h3 class="box-title">Change State</h3>
       </div>
     </form>
   </div>
+  <div class="box box-primary">
+    <div class="box-header with-border">
+      <h3 class="box-title">Run Backfill</h3>
```
ewdurbin: Probably worth renaming this to "Run Evaluation"

ewdurbin:
```suggestion
      <h3 class="box-title">Run Evaluation</h3>
```

PR #7249, opened 2020-01-16
Event-based Malware check

### LINKED ISSUE #7198: PoC Event-based malware check
Once #7196 has settled, we'll need an initial (non-stub) malware check for proofing the event-based check system.

The current proposal: our PoC check will be based on YARA, and will include patterns for detecting unusual/malicious patterns in `setup.py` files. Examples of unusual and malicious patterns:

* Attempting to spawn or invoke processes: `os.system`, `os.exec*`, `os.posix_spawn*`, etc.
* Attempting to perform network requests
* Attempting to call deserialization routines frequently used for ACE and/or obfuscation

See #7096.

woodruffw:
Introduces SetupPatternCheck, an implementation of an event-based
check that scans the `setup.py`s of release files for suspicious
patterns.

Closes #7198.


warehouse/malware/checks/example.py
```diff
@@ -26,7 +26,11 @@ class ExampleCheck(MalwareCheckBase):
     def __init__(self, db):
         super().__init__(db)
 
-    def scan(self, file_id):
+    def scan(self, **kwargs):
+        file_id = kwargs.get("obj_id")
```
woodruffw: N.B.: It's probably overkill to guard on `obj_id`, since it's explicitly supplied during `run_check`.

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, db):
+        super().__init__(db)
+        self._yara_rules = self._load_yara_rules()
+
+    def _load_yara_rules(self):
+        return yara.compile(filepath=self._yara_rule_file)
+
+    def scan(self, **kwargs):
+        file_id = kwargs.get("obj_id")
```
woodruffw: Same here. It probably makes sense to check on `file_url`, though.

warehouse/malware/checks/base.py
```diff
@@ -20,17 +21,28 @@ def __init__(self, db):
         self._load_check_id()
         self._verdicts = []
 
+    @classmethod
+    def prepare(cls, request, obj_id):
+        # TODO(ww): Document.
+        kwargs = {}
+
+        if cls.hooked_object == "File":
+            path = request.db.query(File.path).filter(File.id == obj_id).one()
```
xmunoz: I believe you can use `get(obj_id)` here instead of `filter(File.id == obj_id).one()`

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,116 @@
...
+            except tarfile.TarError:
+                return None
+
+    def scan(self, **kwargs):
+        file_id = kwargs.get("obj_id")
+        file_url = kwargs.get("file_url")
+        if file_id is None or file_url is None:
+            return
+
+        file = self.db.query(File).filter(File.id == file_id).one()
```
xmunoz: I think you can simply use `.get(File.id)` here instead of `.filter(File.id == file_id).one()`

https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.get

woodruffw: c570cf55

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+        # The check harness will catch everything eventually,
+        # so these might not be worth handling explicitly.
+        file_request = requests.get(file_url)
+        if not file_request.ok:
+            # TODO: Indeterminate verdict here?
+            return
+
+        setup_py_contents = self._extract_setup_contents(BytesIO(file_request.content))
+        if setup_py_contents is None:
+            self.add_verdict(
```
woodruffw: For review: it might not be worth reporting a verdict at all here, since the absence of a `setup.py` is neither a positive nor negative indicator.

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+            return
+
+        matches = self._yara_rules.match(data=setup_py_contents)
+        for match in matches:
+            self.add_verdict(
+                file_id=file_id,
+                classification=VerdictClassification(match.meta["classification"]),
+                confidence=VerdictConfidence(match.meta["confidence"]),
+                message=match.rule,
```
woodruffw: For review: I could make the message here lengthier (e.g. by adding more information to the `meta` section of each YARA rule). Thoughts?

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+    hooked_object = "File"
+
+    def __init__(self, db):
+        super().__init__(db)
+        self._yara_rules = self._load_yara_rules()
+
+    def _load_yara_rules(self):
+        return yara.compile(filepath=self._yara_rule_file)
+
+    def _extract_setup_contents(self, archive_stream):
```
woodruffw: For review: we could probably expose a `malware.checks.services` or similar module somewhere that contains routines like this one, since they'll probably get reused across multiple checks. Thoughts?

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+        if zipfile.is_zipfile(archive_stream):
+            with zipfile.ZipFile(archive_stream) as zipobj:
+                for name in zipobj.namelist():
+                    parts = os.path.split(name)
+                    if len(parts) == 2 and parts[1] == "setup.py":
+                        return zipobj.read(name)
+                return None
+        else:
```
woodruffw: https://github.com/python/cpython/pull/18090 eventually :wink:

ewdurbin: nice work!

warehouse/malware/checks/setup_patterns/check.py
ewdurbin: I'm definitely +1 for setting up some helper methods for common operations like fetching the file from CDN. I'm not sure if we'll have many checks that fetch a specific file aside from setup.py... but a `_extract_file_contents(self, archive_stream, file_path)` type method for doing so would also be pretty rad.

warehouse/malware/checks/setup_patterns/check.py
ewdurbin: looks like the message would currently be something like `process_spawn_in_setup` or `networking_in_setup `? If so that's _probably_ sufficient but something human readable would be nice.

warehouse/malware/checks/setup_patterns/check.py
ewdurbin: I suppose it depends. The only way we know the check ran is by a verdict being created, so I think creating the verdict is sane. "neither a positive nor negative" seems like a match to Indeterminate, but I'm not positive?

warehouse/malware/models.py
```diff
@@ -32,34 +32,39 @@
 from warehouse.utils.attrs import make_repr
 
 
+@enum.unique
```
ewdurbin: nice call.

warehouse/malware/checks/base.py
```diff
@@ -43,12 +57,12 @@ def backfill(self, sample=1):
         """
 
     def _load_check_id(self):
-        self.id = (
+        (self.id,) = (
```
ewdurbin: Is this necessary? I think the parenthesis here are for a multiline continuation of the `self.db.query(MalwareCheck.id)` statement, and not a tuple.

warehouse/malware/checks/setup_patterns/check.py
woodruffw: > I suppose it depends. The only way we know the check ran is by a verdict being created, so I think creating the verdict is sane. "neither a positive nor negative" seems like a match to Indeterminate, but I'm not positive?

Sounds good! And yes, my understanding of our model for verdicts is that "indeterminate" matches "neither positive nor negative" 🙂

warehouse/malware/checks/setup_patterns/check.py
woodruffw: > looks like the message would currently be something like `process_spawn_in_setup` or `networking_in_setup `?

Yep, exactly. I'll go ahead and add brief descriptions via `meta` as well. It also probably wouldn't be difficult to convert the match's internal `offset, pattern` list into a `dict` for `MalwareVerdict.details`, if desired.

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+            self.add_verdict(
+                file_id=file_id,
+                classification=VerdictClassification.Indeterminate,
+                confidence=VerdictConfidence.High,
+                message="sdist does not contain a suitable setup.py for analysis",
+            )
+            return
+
+        matches = self._yara_rules.match(data=setup_py_contents)
+        for match in matches:
```
ewdurbin: I'm not sure if we want multiple verdicts or a single verdict here. I think it's understandable to have multiple verdicts, but a single verdict would certainly reduce noise.

Perhaps a single verdict with the max(classification) and max(confidence), with each of the rules spelled out in the message and the full breakout of matches with their individual classification and confidence in `Verdict.details`?

warehouse/malware/checks/setup_patterns/check.py
woodruffw: > I'm definitely +1 for setting up some helper methods for common operations like fetching the file from CDN. I'm not sure if we'll have many checks that fetch a specific file aside from setup.py... but a `_extract_file_contents(self, archive_stream, file_path)` type method for doing so would also be pretty rad.

Sounds good! I can make a helper for the fetch operation. One challenge with `_extract_file_contents` as a generic helper is that `file_path`'s base directory is indeterminate -- for the `sdist`s that I've observed it tends to be `$packagename-$version/` but I don't believe that's strictly enforced anywhere. We could make "ignores the base directory" an explicit part of the behavior, though.

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def scan(self, **kwargs):
+        file_id = kwargs.get("obj_id")
+        file_url = kwargs.get("file_url")
+        if file_id is None or file_url is None:
+            # TODO: Maybe raise here, since the absence of these
+            # arguments is a use/user error.
+            return
+
+        file = self.db.query(File).get(file_id)
```
ewdurbin: maybe `prepare` should be doing this for us? tacking `obj` into kwargs...

or maybe a helper method on the base class that allows fetching the object.

```
file = self.hooked_object_instance()
```

Something to standardize the method that checks use to access objects.

warehouse/malware/checks/setup_patterns/check.py
ewdurbin: that's absolutely correct, AFAIK there's no enforcement on the name of the base directory so it should probably be presumed guilty of breaking that convention.

warehouse/malware/checks/setup_patterns/check.py
ewdurbin: I think leveraging `details` here is a great way to showcase how it is useful :)

warehouse/malware/checks/base.py
woodruffw: I didn't think it was, but `black` complained about it. (This was the CI lint failure that was confusing me, since my local `black` hadn't complained about it until I ran a full `make lint`).

warehouse/malware/checks/setup_patterns/check.py
woodruffw: Sounds good!

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+                            return tarobj.extractfile(member).read()
+                        member = tarobj.next()
+                    return None
+            except tarfile.TarError:
+                return None
+
+    def scan(self, **kwargs):
+        file_id = kwargs.get("obj_id")
+        file_url = kwargs.get("file_url")
+        if file_id is None or file_url is None:
```
xmunoz: I think it makes sense to implement a custom malware exception that is not retryable, and throw such an error in this case. This would require modifying `run_check` a bit. In https://github.com/pypa/warehouse/pull/7271 I'm making a lot of changes to `run_check` anyway so I can take care of that there.

warehouse/malware/checks/setup_patterns/check.py
```diff
@@ -0,0 +1,120 @@
...
+        file = self.db.query(File).get(file_id)
+        if file.packagetype != "sdist":
+            # Per PEP 491: bdists do not contain setup.py.
+            # This check only scans dists that contain setup.py, so
+            # we have nothing to perform.
+            return
+
+        # TODO(ww): Handle connection errors, timeouts?
+        # The check harness will catch everything eventually,
+        # so these might not be worth handling explicitly.
```
xmunoz: I agree with the approach outlined in this comment. Maybe remove the `file_request.ok` checking and let it raise an `AttributeError` a few lines down on `file_request.content` to that `run_check` retries a few times.

tests/unit/malware/test_setup_patterns.py
```diff
@@ -0,0 +1,243 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
xmunoz: Can you create a new directory under `malware` called `checks` and put this file in there?

xmunoz:
Some minor comments about exception handling, otherwise LGTM.

warehouse/malware/checks/setup_patterns/check.py
woodruffw: Thanks!

warehouse/malware/checks/setup_patterns/check.py
woodruffw: Sounds good! I'm refactoring this into a general utility but I'll make it as noisy as possible (in terms of exceptions) to ensure that we propagate them outwards for retrying.

warehouse/malware/checks/setup_patterns/check.py
woodruffw: Yeah, I think does this in `prepare` makes sense -- the object ID is only available in a few contexts, and that's one of them.

PR #7256, opened 2020-01-17
[WIP] Download boosts in search

yeraydiazdiaz:
Addressing https://github.com/pypa/warehouse/issues/3932 though not completely.

Part two split from https://github.com/pypa/warehouse/pull/5311/

- Adds "Downloads last 30 days" option from the search ordering mode applying a [function score modifier](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html) to the search to allow more popular packages to score higher.
- Adds `downloads_last_30_days` column in database and ElasticSearch models.
- Amends `compute_trending` BigQuery query to retrieve the downloads in the last 30 days.
- Adapt `compute_trending` and `reindexing` task to include the downloads data.

Here's a capture of the functionality:

downloads_query_order

You'll notice popular packages rank higher using this order, however, you may have noticed that `black` scores quite high as well when it has little to do with the terms of the search.

Taking the query for "django" as an example, this happens because in `black`s description there's a testimonial from a Django developer and the number of downloads is so large compared to the other more relevant but less popular results that it scores above them.

I've played around with changing the search field boosts and the function score, but of course this affects the "Relevance" mode, since the function score wraps it.

My first reaction would be to use a different combination of boosts for this "order", but I'd like to get some feedback first and probably implement it as a separate PR.


yeraydiazdiaz:
In order to test this feature properly you'll need the downloads data from BigQuery. As I did in the [original PR](https://github.com/pypa/warehouse/pull/5311#issuecomment-457851778) I've put together a [database dump with the extra data](https://ydz.s3.eu-west-2.amazonaws.com/example_downloads_data_2020.sql.xz). The database includes all the migrations and download data in the new column. To use it:

1. Download the file and place it in the `dev` subdirectory.
2. Edit the Makefile line 5 to read `DB := example_downloads_data`.
3. Run `make initdb`.
4. Run `make serve`.
5. Perform the searches as needed for testing.

yeraydiazdiaz:
Closing this PR as it's not really an improvement on our current search anyway.

We can always introduce it later if need be.

Issue #5800, opened 2019-05-07
Implement 2FA recovery codes


ronaldoussoren:
**What's the problem this feature will solve?**
The new 2FA feature adds a TOTP second factor to the authentication proces. This can be problematic when you lose access to the device you use for the 2FA authentication (e.g. the phone running the TOTP app). 

**Describe the solution you'd like**
Most TOTP using sites I've used have an option of recovery codes that can be used to log in with a recovery code when you've lost access to the TOTP device.

**Additional context**

[Github documentation on account recovery for 2FA](https://help.github.com/en/articles/recovering-your-account-if-you-lose-your-2fa-credentials)

di:
https://github.com/pypa/warehouse/issues/5758 ("Define manual account recovery process") is related.

brainwane:
Per [conversation in IRC today](http://kafka.dcpython.org/day/pypa-dev/2019-05-16) -- recovery codes, while a useful feature, are not in-scope for the deliverables we promised in [our grant-funded work](https://pyfound.blogspot.com/2018/12/upcoming-pypi-improvements-for-2019.html) from the Open Technology Fund ([more details](https://pyfound.blogspot.com/2019/03/commencing-security-accessibility-and.html)). So right now we'll continue with the current workflow, where it falls on PyPI site admins to perform the recovery process on behalf of users, and if we have time in this grant, we'll prioritize doing this, but otherwise we will work on it on a volunteer basis afterward.

woodruffw:
To follow up on @nurupo's question in #996:

> All 2FA implementations I have seen have a backup method you can use if you lose your TOTP/U2F device. Often times they provide backup codes. Sometimes it's TOTP with SMS as backup, or U2F with TOTP as backup. But I think I have yet to see a 2FA without a backup method.

If a user has both TOTP and WebAuthn enabled but has lost access to one or the other, they'll still be able to use the one they still have access to. That's not a backup scheme per se since we impose no ordering between TOTP and WebAuthn, but it's something we'll support.

Re: SMS and recovery codes: like @brainwane said these are not in scope for the OTF work, but it's my professional opinion that SMS should **not** be either a second-factor **or** recovery option in any greenfield 2FA implementation. Recovery codes are a great idea and IIRC are on the PyPI roadmap, but (again IMO) should not be mandatory, as they circumvent the two factor scheme entirely. Users who don't enable recovery codes will then be at the mercy of the recovery process determined by the site admins.

nurupo:
[...]

PR #7260, opened 2020-01-17
Implement account Recovery Codes

ewdurbin:
Resolves: #5800 

Upon initial generation or user requests regeneration, a set of 8 recovery codes are generated and stored hashed in the database. If being regenerated all existing codes are invalidated (deleted). The plaintext codes are presented to the user *once* so that they can store them offline.

If a user logs in using a valid recovery code, it is invalidated (deleted).


ewdurbin:
@woodruffw @dstufft @di @yeraydiazdiaz, this is a first pass at this I've been working on in the evenings over the last couple weeks, if I could get feedback to sanity check the implementation I'll work to push this over the line :)

warehouse/templates/accounts/recovery-code.html
```diff
@@ -0,0 +1,66 @@
...
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "base.html" %}
+
+{% block title %}
+  {% if form and form.errors %}{% trans %}Error processing form{% endtrans %} –{% endif %}
+  {% trans %}Password reset{% endtrans %}
```
yeraydiazdiaz:
```suggestion
  {% trans %}Recovery Codes{% endtrans %}
```

warehouse/templates/manage/account/recovery_codes-provision.html
```diff
@@ -0,0 +1,76 @@
...
+{% block mobile_search %}{% endblock %}
+
+{% block content %}
+<div class="horizontal-section">
+  <div class="site-container">
+    <div class="col-half">
+      <h1 class="page-title page-title--wsubtitle heading-wsubtitle__heading">{{ title }}</h1>
+      <p>If you lose access to your authentication application or security key(s), you’ll need to use one of these recovery codes to log into your PyPI account. Each code can only be used <strong>once</strong>.</p>
+      <p>These codes should <strong>only</strong> be used for account recovery, not for typical logins.</p>
+      <p><strong>Keep these somewhere safe</strong>. If you lose your authentication application or security key(s) and do not have access to these recovery codes, you may permanently lose access to your PyPI account!</p>
```
yeraydiazdiaz: Missing `{% trans %}` blocks?

warehouse/templates/manage/manage_base.html
```diff
@@ -93,7 +94,7 @@ <h3 class="sidebar-section__title">{% trans %}Your account{% endtrans %}</h3>
         <i class="fa fa-times" aria-hidden="true"></i>
         <span class="sr-only">{% trans %}Close{% endtrans %}</span>
       </a>
-      <form method="POST" class="modal__form" action="{{ action or request.current_route_path() }}">
+      <form method="{{ method }}" class="modal__form" action="{{ action or request.current_route_path() }}">
```
yeraydiazdiaz: Good catch 👍

warehouse/accounts/views.py
```diff
@@ -322,6 +323,50 @@ def webauthn_authentication_validate(request):
...
+
+            resp = HTTPSeeOther(redirect_to)
+            resp.set_cookie(
+                USER_ID_INSECURE_COOKIE,
+                hashlib.blake2b(str(userid).encode("ascii"), person=b"warehouse.userid")
+                .hexdigest()
+                .lower(),
+            )
+
+            return resp
```
yeraydiazdiaz: Idea: Should we maybe add a flash informing the user that the code they've used will no longer be valid?

warehouse/templates/manage/account/recovery_codes-provision.html
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+    <div class="col-half">
+      <div>
+        <h2>{% trans %}Save your Recovery Codes{% endtrans %}</h2>
+        <div class="recovery-codes">
```
yeraydiazdiaz: Are we missing an SCSS file? I can't find this class in master.

warehouse/accounts/views.py
```diff
@@ -322,6 +323,50 @@ def webauthn_authentication_validate(request):
...
+        if form.validate():
+            _login_user(request, userid, two_factor_method="recovery-code")
+
+            resp = HTTPSeeOther(redirect_to)
+            resp.set_cookie(
+                USER_ID_INSECURE_COOKIE,
+                hashlib.blake2b(str(userid).encode("ascii"), person=b"warehouse.userid")
+                .hexdigest()
+                .lower(),
+            )
```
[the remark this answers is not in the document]
ewdurbin: It would be super tidy if we could bundle all of this up in to the `_login_user` function. It might be a bit messy though. If you have a tidier suggestion for how to incorporate that refactor into this PR I'm happy to take a crack at it, but it's probably better handled in a separate changeset.

warehouse/templates/manage/account/recovery_codes-provision.html
ewdurbin: I didn't actually define any new CSS here, just tried to put classes on it so someone who knows how to do that can fix my ugly work 😅

warehouse/accounts/services.py
```diff
@@ -255,6 +256,14 @@ def has_webauthn(self, user_id):
 
         return len(user.webauthn) > 0
 
+    def has_recovery_codes(self, user_id):
+        """
+        Returns True if the user has generated recovery codes.
+        """
+        user = self.get_user(user_id)
+
+        return len(user.recovery_codes) > 0
```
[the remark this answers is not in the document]
ewdurbin: good eye.

warehouse/accounts/views.py
ewdurbin: great idea!

warehouse/accounts/views.py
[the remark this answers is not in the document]
yeraydiazdiaz: Sounds good 👍

warehouse/accounts/services.py
```diff
@@ -455,6 +469,81 @@ def record_event(self, user_id, *, tag, ip_address, additional=None):
         user = self.get_user(user_id)
         return user.record_event(tag=tag, ip_address=ip_address, additional=additional)
 
+    def generate_recovery_codes(self, user_id):
+        user = self.get_user(user_id)
+
+        if user.has_recovery_codes:
+            self.db.query(RecoveryCode).filter_by(user=user).delete()
+
+        recovery_codes = [secrets.token_hex(8) for _ in range(8)]
```
woodruffw: Ultra nit: Might be worth parametrizing the `8` in `range(8)` as `RECOVERY_CODE_COUNT` or similar.

warehouse/accounts/services.py
```diff
@@ -455,6 +469,81 @@ def record_event(self, user_id, *, tag, ip_address, additional=None):
...
+
+        if valid:
+            self._metrics.increment("warehouse.authentication.recovery_code.ok")
+        else:
+            self._metrics.increment(
+                "warehouse.authentication.recovery_code.failure",
+                tags=["failure_reason:invalid_recovery_code"],
+            )
+            # If we've gotten here, then we'll want to record a failed attempt in our
+            # rate limiting before returning False to indicate a failed totp
```
woodruffw: `failed totp` -> `failed recovery code`?

warehouse/accounts/views.py
```diff
@@ -322,6 +323,61 @@ def webauthn_authentication_validate(request):
...
+
+            user_service.record_event(
+                userid,
+                tag="account:recovery_codes:recovery_code_used",
```
woodruffw: Proposal: abbreviate to `account:recovery_code:used` or similar, for consistency with the other events.

warehouse/manage/views.py
```diff
@@ -616,6 +616,70 @@ def delete_webauthn(self):
...
+    def __init__(self, request):
+        self.request = request
+        self.user_service = request.find_service(IUserService, context=None)
+
+    @view_config(
+        request_method="GET",
+        route_name="manage.account.recovery-codes.generate",
+        renderer="manage/account/recovery_codes-provision.html",
+    )
+    def recovery_codes_generate(self):
```
woodruffw: Here and for regenerate: does it make sense to allow users to generate recovery codes if they don't have an active second factor method? The template hides this logic in the WebUI if the user doesn't have any 2FA, but it might also make sense to enforce it in the view itself.

warehouse/manage/views.py
```diff
@@ -616,6 +616,70 @@ def delete_webauthn(self):
...
+        if self.user_service.has_recovery_codes(self.request.user.id):
+            self.request.session.flash(
+                _("Recovery codes already generated"), queue="error"
+            )
+            return HTTPSeeOther(self.request.route_path("manage.account"))
```
di: IMO it would be more user-friendly to still render this page, and just render this warning instead of the codes, rather than flashing/redirecting. Could also guide the user to re-generating as that might be what they want to do.

warehouse/manage/views.py
```diff
@@ -616,6 +616,70 @@ def delete_webauthn(self):
...
+        self.user_service.record_event(
+            self.request.user.id,
+            tag="account:recovery_codes:generated",
+            ip_address=self.request.remote_addr,
+        )
+        return {
+            "recovery_codes": self.user_service.generate_recovery_codes(
+                self.request.user.id
+            )
+        }
```
di: This allows the user to generate recovery codes without having any 2FA provisioned -- should it? If not, the button to generate codes should probably be disabled as well.

warehouse/templates/accounts/two-factor.html
```diff
@@ -104,12 +104,20 @@ <h2>{% trans %}Authenticate with an app{% endtrans %}</h2>
 
           {% trans trimmed href='/help/#totp' %}
           <p>Generate a code using the authentication application connected to your PyPI account. Enter this code in the form to verify your identity.</p>
-          <p>Lost your application? Not working? <a href="{{ href }}">Get help</a>.</p>
+          <p>Lost your application?  Not working? <a href="{{ href }}">Get help</a>.</p>
```
di:
```suggestion
          <p>Lost your application? Not working? <a href="{{ href }}">Get help</a>.</p>
```

warehouse/templates/accounts/two-factor.html
```diff
@@ -104,12 +104,20 @@ <h2>{% trans %}Authenticate with an app{% endtrans %}</h2>
...
       {% endif %}
     </div>
+
+    <hr>
+    <div class="twofa-recovery">
+      <p>{% trans %}Lost your security key or application?{% endtrans %}</p>
+      <a href="{{ request.route_path('accounts.recovery-code') }}?{{request.query_string}}" class="button button--secondary">
+          {% trans %}Login using a Recovery Code{% endtrans %}
+      </a>
+    </div>
```
di: This is displayed whether or not the user has recovery codes. Perhaps intentionally? But this might be a good opportunity to direct a user that does have 2FA but hasn't generated recovery codes that they should generate codes once logging in.

warehouse/accounts/views.py
```diff
@@ -322,6 +323,61 @@ def webauthn_authentication_validate(request):
...
+                ip_address=request.remote_addr,
+            )
+
+            request.session.flash(
+                _("Recovery code accepted. The supplied code cannot be used again."),
+                queue="success",
+            )
+
+            return resp
```
di: Should we maybe redirect to `/manage/account/` so the user can disable/reset their 2FA here? Seems like the most likely thing they'd want to do after using a recovery code.

waynew:
Not a detailed review, but FWIW 2FA experiences have been that the recovery codes are shown to me after I've turned on 2fa. On GitHub, I think it is, they display a message that we will no longer see codes after they're shown that once.

I think for folks who want to live dangerously, simply not saving the codes anywhere at that point is equivalent to just deleting theirs.

Implementation question: does hashed mean bcrypt or equivalent? Seems like a rainbow table would be pretty trivial otherwise? (Apologies if the code could answer that - phones aren't super for source code)

warehouse/manage/views.py
ewdurbin: put a guard in the view in https://github.com/pypa/warehouse/pull/7260/commits/199adf0ca51dbb6fbaf488bb2a15b61ae3f57eaa, I'll hide the button in the template as well in next commit.

warehouse/accounts/views.py
ewdurbin: seems reasonable to me.

ewdurbin:
@di, good discussion points

> I think this should probably also include a way for the user to delete their recovery codes (w/ a password prompt) in case they've lost control of their backup (or just want to live dangerously).

This seems reasonable, but I think we'd have to track recovery code "opt-out" explicitly somehow? Particularly given the next point.

> Should we also force everyone with 2FA enabled but no recovery codes generated to generate them after login?

I'm iffy on this. And similar to above, I think we'd need to have some way to explicitly "opt-out" if people do not want to for some reason?

> I think we also need to handle the case where the user has used all of their recovery codes, perhaps forcing them to regenerate as well.

Good point, perhaps a warning when only two remain with a flash and a forced regenerate on the last one?

ewdurbin:
@waynew

> Not a detailed review, but FWIW 2FA experiences have been that the recovery codes are shown to me after I've turned on 2fa. On GitHub, I think it is, they display a message that we will no longer see codes after they're shown that once.

I did a test enrollment with GitHub two-factor and they display recovery codes _up front_ before any other method is provisioned. This makes a ton of sense from a customer support perspective and we should probably consider it.

GitHub and most other services I've checked _do_ allow for viewing previously generated codes again after some confirmation (password or a two factor device). I'm not a huge fan of this and think encouraging regeneration over viewing them again is preferable.

> I think for folks who want to live dangerously, simply not saving the codes anywhere at that point is equivalent to just deleting theirs.

I suppose it is infinitesimally less secure than explicitly having no recovery codes, but would tend to agree.
 
> Implementation question: does hashed mean bcrypt or equivalent? Seems like a rainbow table would be pretty trivial otherwise? (Apologies if the code could answer that - phones aren't super for source code)

We're using the same strategy as passwords, which is currently argon2. Attempts to authenticate with recovery code are also rate limited.

warehouse/templates/manage/account/recovery_codes-provision.html
```diff
@@ -0,0 +1,90 @@
...
+              <a href="#complete" class="button button--small button--primary">
+                {% trans %}Continue{% endtrans %}
+              </a>
+              {% set custom_warning_text %}
+                <p><strong>{% trans %}These codes will not be visible again.{% endtrans %}</strong></p>
+              {% endset %}
+              {% set extra_description %}
+                <p>{% trans %}Ensure that you have securely stored them before continuing.{% endtrans %}</p>
+              {% endset %}
+              {{ confirm_modal(title, gettext("Username"), "username", request.user.username, slug, confirm_button_label=confirm_button_label, extra_description=extra_description, confirm_string_in_title=False, custom_warning_text=custom_warning_text, action=action, method="GET") }}
```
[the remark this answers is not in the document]
ewdurbin: I thought about "confirm" or "yes" or some other string, but thought that might be weird from a translations perspective.

Do we we have a modal macro for just a confirmation button?

warehouse/templates/accounts/two-factor.html
```diff
@@ -110,6 +110,21 @@ <h2>{% trans %}Authenticate with an app{% endtrans %}</h2>
       </div>
       {% endif %}
     </div>
+
+    <hr>
+    <div class="twofa-recovery">
+      {% if request.user.has_recovery_codes %}
```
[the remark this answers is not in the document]
ewdurbin: whoops! good catch. yeah since this is before we've auth'd the view will need to return something to switch on here.

warehouse/templates/manage/account/recovery_codes-provision.html
```diff
@@ -0,0 +1,90 @@
...
+            <div class="recovery-codes__-codes">
+              <ul>
+                {% for recovery_code in recovery_codes %}
+                <li><code>{{ recovery_code }}</code></li>
+                {% endfor %}
+              </ul>
+              <button type="button" class="button button--small copy-tooltip copy-tooltip-w" data-tooltip-label="{% trans %}Copy to clipboard{% endtrans %}" data-clipboard-text="{{ 'PyPI Recovery Codes\n' + recovery_codes|join('\n') }}">
+                {% trans %}Copy{% endtrans %}
+              </button>
+              <a href="data:text/plain;,{{ recovery_codes|join('%0A') }}" download="PyPI-Recovery-Codes.txt" class="button button--small" data-tooltip-label="{% trans %}Download as file{% endtrans %}" target="_blank">
```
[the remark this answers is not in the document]
ewdurbin: I think I added that to make it actually download as a file, but will try without it.

warehouse/accounts/models.py
```diff
@@ -134,6 +137,10 @@ def email(self):
     def has_two_factor(self):
         return self.totp_secret is not None or len(self.webauthn) > 0
 
+    @property
+    def has_recovery_codes(self):
+        return len(self.recovery_codes) > 0
```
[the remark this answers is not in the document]
ewdurbin: If `self.recovery_codes.count()` worked like it does in the Django ORM I'd probably say yeah let's go for that.

I think in order to use sqlalchemy's `func.count` we'd have to get access to the db session, which I'm not sure how to do in the context of the model.

This is probably fine, as we've been using the same method for `has_two_factor` and webauthn keys.

warehouse/templates/manage/account/recovery_codes-provision.html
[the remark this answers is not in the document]
ewdurbin: seems that was a relic from my initial testing, I'll remove.

warehouse/templates/manage/account/recovery_codes-provision.html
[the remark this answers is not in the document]
yeraydiazdiaz: I don't think we do, but that's fine, confirming the username makes sense as well 👍

PR #7271, opened 2020-01-20
Implement scheduled checks #7093

xmunoz:
- Rename `run_backfill` to `run_evaluation` in admin malware view
- Modify `run` and `scan` method signatures to accept `**kwargs`
- Extend `run_check` to accomodate scheduled check functionality

### DIFF
--- a/tests/unit/admin/views/test_checks.py
+++ b/tests/unit/admin/views/test_checks.py
@@ -47,6 +47,7 @@ class TestGetCheck:
             "check": check,
             "checks": [check],
             "states": MalwareCheckState,
+            "evaluation_run_size": 10000,
         }
 
     def test_get_check_many_versions(self, db_request):
@@ -57,6 +58,7 @@ class TestGetCheck:
             "check": check2,
             "checks": [check2, check1],
             "states": MalwareCheckState,
+            "evaluation_run_size": 10000,
         }
 
     def test_get_check_not_found(self, db_request):
@@ -196,7 +198,7 @@ class TestRunEvaluation:
             ]
             assert db_request.task.calls == [pretend.call(backfill)]
             assert backfill_recorder.delay.calls == [pretend.call(check.name, 10000)]
-        else:
+        elif check_type == MalwareCheckType.scheduled:
             assert db_request.session.flash.calls == [
                 pretend.call("Running %s now!" % check.name, queue="success",)
             ]
@@ -204,3 +206,5 @@ class TestRunEvaluation:
             assert backfill_recorder.delay.calls == [
                 pretend.call(check.name, manually_triggered=True)
             ]
+        else:
+            raise Exception("Invalid check type: %s" % check_type)
--- a/tests/unit/admin/views/test_checks.py
+++ b/tests/unit/admin/views/test_checks.py
@@ -16,7 +16,8 @@ import pytest
 from pyramid.httpexceptions import HTTPNotFound
 
 from warehouse.admin.views import checks as views
-from warehouse.malware.models import MalwareCheckState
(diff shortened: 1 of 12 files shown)


warehouse/admin/views/checks.py
```diff
@@ -56,32 +56,36 @@ def get_check(request):
...
         return HTTPSeeOther(
             request.route_path("admin.checks.detail", check_name=check.name)
         )
 
-    request.session.flash(
-        f"Running {check.name} on {num_objects} {check.hooked_object.value}s!",
-        queue="success",
-    )
+    if check.check_type == MalwareCheckType.event_hook:
+        num_objects = 10000
```
woodruffw: Ultra nit: Maybe parametrize this somewhere (either as a module-level constant or as something loaded into the application-level config via `includeme`)?

tests/unit/admin/views/test_checks.py
```diff
@@ -177,13 +184,23 @@ def test_sucess(self, db_request):
...
+            assert db_request.session.flash.calls == [
+                pretend.call(
+                    "Running %s on 10000 %ss!"
+                    % (check.name, check.hooked_object.value),
+                    queue="success",
+                )
+            ]
+            assert db_request.task.calls == [pretend.call(backfill)]
+            assert backfill_recorder.delay.calls == [pretend.call(check.name, 10000)]
+        else:
```
ewdurbin: I'm not sure how I feel about tests for different behavior all in one, since we'll end up with a cascading `elif` here. But for now might `elif check_type == MalwareCheckType.scheduled` and then `else:` with an invalid assert. Would make it easier to tell where the asserts for a given type are and explicitly fail in the implicit case.

warehouse/malware/tasks.py
```diff
@@ -21,10 +23,30 @@
 
 
 @task(bind=True, ignore_result=True, acks_late=True, retry_backoff=True)
-def run_check(task, request, check_name, obj_id):
-    check = getattr(checks, check_name)(request.db)
+def run_check(task, request, check_name, obj_id=None, manually_triggered=False):
+    try:
+        check = getattr(checks, check_name)(request.db)
+    except NoResultFound:
+        request.log.info("Check %s isn't active. Aborting." % check_name)
```
ewdurbin: Is it the case that it isn't _active_ here or that it doesn't exist in the codebase? This might be a bit misleading.

warehouse/malware/tasks.py
```diff
@@ -21,10 +23,30 @@
...
+        request.log.info(
+            "%s is in the `evaluation` state and must be manually triggered to run."
+            % check_name
+        )
+        return
+
+    kwargs = {}
+
+    # Hooked checks require `obj_id`s.
+    if obj_id:
```
ewdurbin: Probably best to be explicit: `if obj_id is not None:`

ewdurbin:
Direction looks good and sensible. Just a couple of small code questions.

warehouse/malware/tasks.py
xmunoz: If it does not exist in the code base, an `AttributeError` will be raised, which isn't caught here.

xmunoz: > Is it the case that it isn't active here or that it doesn't exist in the codebase?
It is the case that it isn't active (e.g. in the `enabled` or `evaluation` state). Loading info from the database happens when the check class is initialized. That is why I'm only catching the `NoResultFound` exception.

ewdurbin: Ah! I misinterpreted NoResultFound with AttributeError in my mind :)

PR #7275, opened 2020-01-21
Tell dependabot what Python version to use

di:
This is causing issues with kombu's dependency on `importlib-metadata>=0.18; python_version<"3.8"`


scop:
FYI, this causes some annoyances locally -- I use pyenv-virtualenv and have set .python-version to the virtualenv I have for warehouse (named warehouse), and it conveniently auto-enables whenever I cd to the project dir. Now that this file comes from git, my virtualenv setting either no longer works like that, or I have to remember to manage to this file manually back and forth between how it is in git, and how I want it to be.

It's not that _I_ will be much affected by this as I have no plans here after the RSS project feeds are in, but other, more frequent contributors might be similarly affected.

Maybe there are some other ways to poke dependabot?

di:
Hi @scop, sorry this is causing problems for you. AFAIK this is the only way to set the version for Dependabot.

Maybe I'm missing something, but you shouldn't need to use a virtualenv for working on Warehouse. Most of the local development happens inside containers, and the few things that don't (like linting) use their own virtualenv: https://github.com/pypa/warehouse/blob/15c0a1c8d62744a47b0d874e251e2630d0950458/Makefile#L55-L58

Furthermore, this is the version of Python that we actually run Warehouse on, so you shouldn't need to change it to a different version -- or is there something else you need in this file?

yeraydiazdiaz:
FWIW I use `pyenv-virtualenv` too, `.python-version` is the file it reads from to activate not only your Python version but your virtual environment, for instance mine just reads `warehouse`. I created a venv to be able to run linters in VSCode and manually in the terminal without spinning up docker containers.

That being said, I just edited the `.python-version` file to have my virtualenv and avoid adding it in commits. I guess it depends on the way you interact with git as well.

di:
> I created a venv to be able to run linters in VSCode and manually in the terminal without spinning up docker containers.

I'm not following why this is necessary, the linting doesn't run in a container ([...yet](https://github.com/pypa/warehouse/pull/6205)).

That said, it might be about time to move to 3.8, which would obviate the need for this file at the moment, if anyone would like to explore that avenue.

yeraydiazdiaz:
I don't mean the linters as defined in the makefile, I mean linters that format and show errors as I type and save files in VSCode. I find it quite convenient. Again, it depends a lot on your workflow 🙂

scop:
Code completion in editors (Emacs in my case) requires dependencies installed in order for it to work for them, also for my setup python-language-server installed for the same feature, and also various linters (not necessarily ones defined in warehouse's files) etc, just like @yeraydiazdiaz's setup. Not using pyenv-virtualenv is a non-starter for me; I'll manage it manually if needed, but would be nice if that wasn't necessary.

PR #7286, opened 2020-01-22
Fix classifiers search requiring a querystring

yeraydiazdiaz:
Fixes https://github.com/pypa/warehouse/issues/7279


warehouse/search/queries.py
```diff
@@ -53,12 +53,12 @@ def get_es_query(es, querystring, order, classifiers):
     Returns an Elasticsearch query from data from the request.
     """
     if not querystring:
```
ewdurbin: Took me a moment to realize this isn't the full query string in terms of a URL 😅. This led me to wonder how we'd ever _not_ have a `querystring`... but still have a classifier.

I realize this was refactored out of the initial view where it could have been equally confusing, so it's not critical but renaming this might be more helpful. I don't have any good suggestions off the top of my head though.

PR #7308, opened 2020-01-28
Clarify that /stats wants `Accepts` instead

qaisjp:


ewdurbin:
Good catch @qaisjp, can you also update the sample request to remove the two references Content-Type?

qaisjp:
Yes

ewdurbin:
Thank you @qaisjp

qaisjp:
Thank you!

Issue #7199, opened 2020-01-07
PoC schedule-based malware check


woodruffw:
Like #7198, we'll need a schedule-based malware check for proofing.

The current proposal: a scheduled check for complete turnovers (i.e., no original or previous maintainers/role-holders remaining) in package ownership, graded against other suspicious indicators (IPs not previously seen for the package and/or its maintainers, new maintainers(s) are not in the community of any the the previous maintainers beyond a fixed window).

See #7096.

xmunoz:
This might be too complicated for v1, but I think the idea behind sourcerank is really interesting:
https://libraries.io/pypi/requests

A high score combined with complete turnovers might be an interesting signal.

woodruffw:
Merged; will close when this hits master.

PR #7321, opened 2020-01-30
malware/checks: PackageTurnover skeleton

woodruffw:
See #7199.

### DIFF
--- a/warehouse/malware/checks/package_turnover/check.py
+++ b/warehouse/malware/checks/package_turnover/check.py
@@ -1,0 +1,62 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime, timedelta
+from textwrap import dedent
+
+from sqlalchemy.sql import exists
+
+from warehouse.packaging.models import Release
+from warehouse.malware.models import MalwareVerdict
+from warehouse.malware.checks.base import MalwareCheckBase
+
+
+class PackageTurnoverCheck(MalwareCheckBase):
+    version = 1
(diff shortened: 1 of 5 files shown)


warehouse/malware/checks/package_turnover/check.py
```diff
@@ -0,0 +1,114 @@
...
+                .filter(UserEvent.user_id == user.id)
+                .filter(UserEvent.time >= self._scan_interval)
+                .filter(UserEvent.tag == "account:two_factor:method_removed")
+                .exists()
+            ).scalar()
+
+            if has_removed_2fa_method and not user.has_two_factor:
+                self.add_verdict(
+                    project_id=project.id,
+                    classification=VerdictClassification.Threat,
```
xmunoz: I'm not sure disabling 2fa merits the `Threat` classification.

woodruffw: I agree, although I'll point out that it's not just disabling 2FA -- it's disabling 2FA right before a new release and not adding a new method. But yes, I think it'd be a good idea to get a feel for the noise here and lower this if it's too much.

ewdurbin: I reviewed the last few days of activity for that tag, and was actually rather delighted to find that most instances of account:two_factor:method_removed were right before or after adding another method.

On average it's been fewer than ten instances per day.

warehouse/malware/checks/package_turnover/check.py
```diff
@@ -0,0 +1,114 @@
...
+        prior_verdicts = (
+            self.db.query(MalwareVerdict.release_id).filter(
+                MalwareVerdict.check_id == self.id
+            )
+        ).subquery()
+
+        # NOTE: We query for releases here, but the rest of this check is
+        # project-oriented. Maybe just query for projects directly instead?
+        releases = (
+            self.db.query(Release)
```
xmunoz: `self.db.query(Release.project)` perhaps?

woodruffw: Hmm, I'm having some trouble getting this to work, probably because `project` is a backref and not a realized column. I could do something like this:

```
self.db.query(Release.project).join(Release.project).filter(...)
```

but this only returns a list of `bool`s. Thoughts?

xmunoz: Meh, it seems fine as is. Maybe just remove/modify the comment?

woodruffw:
🎉

PR #7339, opened 2020-02-04
Add verdict administrator review. Fixes #6062.

### LINKED ISSUE #6062: admin interface for review of flagged packages
We're working on a system to detect malicious uploads (per #4998). It's going to be a pipeline where automated systems run checks, flag packages/projects for deletion/review/ok, etc.

So this issue is for a feature in the admin interface where administrators/moderators (#4011) can review those flagged projects and releases and decide what to accept/reject.

We'll probably also want to use this if/when we implement a mechanism for users to report packages (#3896), and if/when we start automatically checking uploaded packages for compliant metadata and installability (#194).

xmunoz:
- Add new `admin.verdicts.review` endpoint
- Change layout of verdict list and detail view and add forms
- Change sort order of the MalwareChecks, and update the tests

### DIFF
--- a/warehouse/admin/views/verdicts.py
+++ b/warehouse/admin/views/verdicts.py
@@ -71,7 +71,7 @@ def get_verdict(request):
 
 @view_config(
     route_name="admin.verdicts.review",
-    permission="admin",
+    permission="moderator",
     request_method="POST",
     uses_session=True,
     require_methods=False,
@@ -86,7 +86,7 @@ def review_verdict(request):
         raise HTTPBadRequest("Invalid verdict classification.") from None
 
     verdict.manually_reviewed = True
-    verdict.administrator_verdict = classification
+    verdict.reviewer_verdict = classification
 
     request.session.flash(
         "Verdict %s marked as reviewed." % verdict.id, queue="success"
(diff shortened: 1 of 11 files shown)


xmunoz:
Here is a way you can generate verdicts to test this. First open the warehouse shell ` docker-compose run web python -m warehouse shell`. Then run this:

```
from warehouse.malware.models import MalwareCheck, MalwareVerdict
from warehouse.packaging.models import File
check = db.query(MalwareCheck).filter(MalwareCheck.name == "SetupPatternCheck").one()
release_files = db.query(File).limit(10)
for f in release_files:
	db.add(MalwareVerdict(release_file=f, check=check, classification="threat", confidence="medium", message="Just a test"))

db.flush()
db.commit()
```

warehouse/admin/views/verdicts.py
```diff
@@ -61,11 +61,43 @@ def get_verdict(request):
...
+            "verdict": verdict,
+            "classifications": list(VerdictClassification.__members__.keys()),
+        }
 
     raise HTTPNotFound
 
 
+@view_config(
+    route_name="admin.verdicts.review",
+    permission="admin",
```
ewdurbin: permission="moderator" is probably good here so that we can easily add additional volunteers to review.

ewdurbin: with that in mind, it might make sense to change the name from `administrator_verdict` to `reviewer_verdict`

ewdurbin:
Really just the thoughts around permissions on the ability to review verdicts, and semantics on the name for the db column. Otherwise, this looks great.

PR #7440, opened 2020-02-25
Add pointer to pypi-support on faq page

vinayak-mehta:
Added a pointer to the pypi-support repo in case someone loses access to their authenticator app and don't have recovery codes.


ewdurbin:
Hi @vinayak-mehta thank you for your pull request, I've addressed this separately in #7441

vinayak-mehta:
:+1:

PR #7443, opened 2020-02-25
Check for translation updates

di:
This adds a CI job to check whether the translations need to be updated, as well as a more user-friendly command, `make translations`.

It also recompiles the POT file to get the job passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,3 +55,6 @@ matrix:
     - script: make licenses BINDIR="$(dirname $(which python))"
       env:
         - SUITE=Licenses
+    - script: make translations BINDIR="$(dirname $(which python))"
+      env:
+        - SUITE=Translations
--- a/Makefile
+++ b/Makefile
@@ -207,4 +207,10 @@ build-mos: compile-pot
 		L=$$LOCALE $(MAKE) compile-po ; \
 		done
 
+translations: compile-pot
+ifneq ($(TRAVIS), false)
+	git diff --quiet ./warehouse/locale/messages.pot || (echo "There are outstanding translations, run 'make translations' and commit the changes."; exit 1)
+else
+endif
+
 .PHONY: default build serve initdb shell tests docs deps travis-deps clean purge debug stop compile-pot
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -177,9 +177,7 @@ stop:
 compile-pot: .state/env/pyvenv.cfg
 	PYTHONPATH=$(PWD) $(BINDIR)/pybabel extract \
 		-F babel.cfg \
-		--copyright-holder="PyPA" \
-		--msgid-bugs-address="https://github.com/pypa/warehouse/issues/new" \
-		--project="Warehouse" \
+		--omit-header \
```
ewdurbin: good call here.

ewdurbin:
@di I think omitting the location notes might break the "Nearby strings" feature in weblate?
Screen Shot 2020-02-25 at 3 42 09 PM

di:
@ewdurbin I think it might. Tradeoffs... Those strings don't look super helpful for that particular translation (but I have not actually used Weblate so I can't really say).

Makefile
```diff
@@ -177,9 +177,8 @@ stop:
...
+		--no-location \
```
webknjaz: I think it'd be a good idea to maybe have separate targets just for identifying if regeneration of `messages.pot` is needed and keep the original one as is.

webknjaz: Location info is actually useful as it allows us to track down places in the templates if the context is unclear.

.travis.yml
```diff
@@ -55,3 +55,6 @@ matrix:
     - script: make licenses BINDIR="$(dirname $(which python))"
       env:
         - SUITE=Licenses
+    - script: make translations BINDIR="$(dirname $(which python))"
```
webknjaz: The alternative could look like this: https://github.com/pypa/warehouse/issues/7409#issuecomment-591443217. Having an up-to-date translations source is more useful in master after the PR merge, not in PRs themselves.

di:
I added the location comments back, we'll see how annoying it is.

webknjaz:
I confirm that new strings appeared in the UI and I got a notification from Weblate about it.

PR #7449, opened 2020-02-26
Order search with filters

yeraydiazdiaz:
Fixes https://github.com/pypa/warehouse/issues/7439

### DIFF
--- a/tests/unit/test_search.py
+++ b/tests/unit/test_search.py
@@ -34,7 +34,7 @@ EXPECTED_SEARCH_FIELDS = [
 
 
 class TestQueries:
-    def test_no_querystring(self):
+    def test_no_terms(self):
         es = Search()
 
         query = queries.get_es_query(es, "", "", [])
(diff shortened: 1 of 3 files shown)


tests/unit/test_search.py
```diff
@@ -171,22 +171,36 @@ def test_with_classifiers_with_querystring(self):
...
+        classifiers = [("c", "foo :: bar")]
+
+        query = queries.get_es_query(es, terms, "-created", classifiers)
+
+        query_dict = query.to_dict()
+        assert query_dict["query"] == {"prefix": {"classifiers": ("c", "foo :: bar")}}
+        assert query_dict["sort"] == [
+            {"created": {"order": "desc", "unmapped_type": "long"}}
+        ]
+
```
ewdurbin:
```suggestion
        ]
```

warehouse/search/queries.py
```diff
@@ -48,22 +48,22 @@
 )
 
 
-def get_es_query(es, querystring, order, classifiers):
+def get_es_query(es, terms, order, classifiers):
```
ewdurbin: Thanks for this rename :)

PR #7467, opened 2020-02-28
Add reload interval to hupper to allow time for gunicorn to shutdown

yeraydiazdiaz:
Fixes https://github.com/pypa/warehouse/issues/6578

Add [`reload_interval`](https://docs.pylonsproject.org/projects/hupper/en/latest/api.html#hupper.start_reloader) to the web command. Seems to solve the issue for me but it would be good to do some further testing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -45,7 +45,7 @@ services:
         # working on pypi-theme, this is a private repository due to the fact
         # that other people's IP is contained in it.
         #THEME_REPO:
-    command: hupper -w 'warehouse/locale/**/*.mo' -m gunicorn.app.wsgiapp -b 0.0.0.0:8000 -c gunicorn.conf.py warehouse.wsgi:application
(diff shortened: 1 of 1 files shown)


di:
So, I tried both this and just `shutdown_interval` (which I think is probably actually the important one since the issue is around shutdown) and neither actually worked for me.

Increasing this to a really large number just caused Hupper to wait that long before force-killing and going into the same state.

I also noticed that occasionally (but rarely) it would reload successfully, so maybe you're seeing that?

IIUC increasing `reload_interval` just increases how often hupper checks for changes to initiate a reload. But maybe I'm missing something?

yeraydiazdiaz:
Yeah, just tried this morning and didn't actually work. I also tried with `shutdown_interval` and it does seem to help a bit, but it might be just perception. I tried logging into the docker container and checking the processes as a reload occurs and it seems gunicorn survives the "forceful killing" somehow. 😕

PR #7488, opened 2020-03-03
[WIP] warehouse: TUF initialization

woodruffw:
Adds CLI and task support for initializing a TUF repository.

Adds a service interface (IKeyService) with a filesystem-based
implementation (LocalKeyService). Deployment will (potentially cloud-based)
KMS implementation.


warehouse/tuf/tasks.py
```diff
@@ -0,0 +1,39 @@
...
+
+from tuf import repository_tool
+
+from warehouse.tasks import task
+from warehouse.tuf.interfaces import IKeyService
+
+TOPLEVEL_ROLES = ["root", "snapshot", "targets", "timestamp"]
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
```
ewdurbin: Will we ever be invoking this as a Celery task, or is this just being implemented to match other CLI commands that call tasks?

woodruffw: The latter -- we shouldn't ever need to run this either manually or periodically after initial setup. I did it as a Celery task to get access to a `request` object; is there another way to do that just in the context of the CLI?

ewdurbin: Gotcha, then I think you can use the `warehouse.cli.warehouse.command` decorator to receive a `config` object on your CLI entry point function instead. See [the shell cli command](https://github.com/pypa/warehouse/blob/f12bcb4bcfa557fb5a3a0a50a35ab52163db5491/warehouse/cli/shell.py#L60) for example.

warehouse/tuf/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+        repo_service = request.find_service(IRepositoryService)
+        key_service = request.find_service(IKeyService)
+        repository = repo_service.load_repository()
+
+        dirty_roles = ["snapshot", "timestamp"]
+        for role in dirty_roles:
+            role_obj = getattr(repository, role)
+            [role_obj.load_signing_key(k) for k in key_service.privkeys_for_role(role)]
+
+        # NOTE(ww): I think this should be targets("bins") instead of just targets,
```
woodruffw, line 60: cc @trishankkarthik / @mnm678 / @lukpueh: I *think* this is a bug in `repository_tool.load_repository`, as each bin-n should be delegated from `bins` and not from the top-level `targets`. The repository creation code in `warehouse/cli/tuf.py` works as expected and delegates correctly from `bins`, so it only seems to happen on repo reload.

woodruffw:
This is currently blocked on some bugs in the TUF reference implementation, namely missing roledb state when reloading the repository. Will reference tracking information when it's created.

warehouse/tuf/tasks.py
woodruffw: Tracking issue: https://github.com/theupdateframework/tuf/issues/1045

warehouse/tuf/tasks.py
```diff
@@ -0,0 +1,79 @@
...
+from warehouse.tasks import task
+from warehouse.tuf import utils
+from warehouse.tuf.interfaces import IKeyService, IRepositoryService
+
+
+@task(bind=True, ignore_result=True, acks_late=True)
+def bump_role(task, request, role):
+    r = redis.StrictRedis.from_url(request.registry.settings["celery.scheduler_url"])
+
+    with utils.RepoLock(r):
```
webknjaz, line 24: I don't think that you need another lock CM wrapper. The lib already exposes a context manager: https://github.com/andymccurdy/redis-py/blob/ce88fca/redis/lock.py#L151-L159.
```suggestion
    with r.lock("tuf-repo"):
```

warehouse/tuf/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self, redis_client):
+        self.lock = redis_client.lock("tuf-repo")
+
+    def __enter__(self):
+        self.lock.acquire()
+        return self
+
+    def __exit__(self, *_exc):
+        self.lock.release()
```
webknjaz, line 115: This does not look necessary per https://github.com/pypa/warehouse/pull/7488/files#r490291493

warehouse/tuf/tasks.py
woodruffw: Looks good, thanks for finding that~

miketheman:
Now that RSTUF is part of the dev setup from #15241, is this PR still useful, or should it be closed?

lukpueh:
This can be closed.

lukpueh:
> This can be closed.

It is superseded by 
https://github.com/pypi/warehouse/pull/10870
https://github.com/pypi/warehouse/pull/13943
https://github.com/pypi/warehouse/pull/15241

PR #7496, opened 2020-03-04
Minimize the development database

ewdurbin:
Historically, we've loaded a copy for test.pypi.org that has been sanitized to give developers something to look at when running warehouse locally.

This has two major drawbacks:

1) It is easy to leak private information still, in the event that we miss something in our "clean" script
2) The file is _large_. The current dump is 67MB and a recent dump would be over 200.

This PR creates a new minimal dev db. Since we're already paying for the dump in history, it also introduces a new make target `exampledb` that allows for loading that.


di:
While we're at it, can update the existing TestPyPI dump here to have all the latest migrations?

ewdurbin:
> While we're at it, can update the existing TestPyPI dump here to have all the latest migrations?

I thought about doing that, but it'd drop another 67MB blob into our git history. :/

di:
I guess if we're using that dump less, the time to run all those migrations matters less.

We might want to consider updating our docs to suggest cloning with `git clone --depth=1` or something to cut down on the initial load for new contributors regardless.

PR #7529, opened 2020-03-12
Fully implement PEP 527

### LINKED ISSUE #6792: Fully implement PEP 527
We still retain the ability for some projects to upload the legacy filetypes listed in [PEP 527](https://www.python.org/dev/peps/pep-0527/): 

https://github.com/pypa/warehouse/blob/aa0d54019c322b52cb6428780808816d417abbd1/warehouse/packaging/models.py#L124

https://github.com/pypa/warehouse/blob/e1d0e4e41738fd07a0edeb77f95fedba0fdd41f8/warehouse/forklift/legacy.py#L463-L480

https://github.com/pypa/warehouse/blob/e1d0e4e41738fd07a0edeb77f95fedba0fdd41f8/warehouse/forklift/legacy.py#L1167-L1174

We should audit which projects currently have this ability, and whether they are still publishing deprecated filetypes. For example, [Pillow is no longer publishing `bdist_wininst` files](https://github.com/python-pillow/Pillow/pull/4029).

di:
Fixes #6792. To be merged on 2020-04-12.


warehouse/migrations/versions/b265ed9eeb8a_fully_deprecate_legacy_distribution_.py
```diff
@@ -0,0 +1,42 @@
...
+
+revision = "b265ed9eeb8a"
+down_revision = "5c029d9ef925"
+
+
+def upgrade():
+    op.drop_column("projects", "allow_legacy_files")
+
+
+def downgrade():
```
dstufft, line 32: This should probably just fail, going backwards doesn't make sense here because re-adding this isn't going to restore the values in the database that marked whether a project could accept legacy uploads or not.

ewdurbin:
Note: as is this will cause upload failures temporarily during deploy when the old upload fleet goes looking for this column after the migration runs, but before they’re replaced.

pradyunsg:
It's no longer 2020-04-12 anywhere on earth. Is there a new timeline for this PR now?

ewdurbin:
To avoid confusion, I'll set the admin flag `disallow-new-upload` momentarily before deploy so folks get [a useful error](https://github.com/pypa/warehouse/blob/7cbd78c24cbc7acb5ecc45ed787c95e4607f0db8/warehouse/forklift/legacy.py#L755-L762) instead of a 500.

PR #7530, opened 2020-03-12
Refactoring of email templates and add translations

MVrachev:
This pr is dependent on pr: https://github.com/pypa/warehouse/pull/7498

Signed-off-by: Martin Vrachev <mvrachev@vmware.com>

### DIFF
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -1094,7 +1094,7 @@ class TestAddedAsCollaboratorEmail:
         assert send_email.delay.calls == []
 
 
-class TestRemovedPackageEmail:
+class TestRemovedProjectEmail:
     def test_removed_project_email_to_maintainer(
         self, pyramid_request, pyramid_config, monkeypatch
     ):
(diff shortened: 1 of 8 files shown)


ewdurbin:
My only concern here is that the translation will be dependent on the user initiating the action, since the locale will be determined from the request. I'm not sure how we would best handle this situation, not all maintainers of a given project will necessarily be able to read the same language (English or otherwise).

MVrachev:
To be honest @ewdurbin I don't know how to make it independent on the user interaction.
I thought that was the idea? 
If a German user does the interaction then an email in German is sent to him/her, isn't that the idea?

About the maintainers they will look at the code written in english and that's enough for them isn't it? 

Maybe it's good to bring that question on the other open pr https://github.com/pypa/warehouse/pull/7498 where I was requested by @di  to make my strings translatable.

di:
Yeah, I think @ewdurbin has a good point that I originally missed here, and as a result we should probably ensure that all emails are not translated and sent in English.

The rest of the refactoring in this PR looks good to me though.

MVrachev:
Sorry, but I am not sure I understand what you mean.
The other emails are same as mine with the exception that they are not translating any strings like me:
image

Is the problem that I am making a LazyString object from `an owner/a maintainer` and then casting it to string?

di:
Sorry for the confusion. Ernest is saying that since the translations are done based on the locale of the user that initiated the action, all the emails that are sent to the collaborators will also be translated to that user's locale, not the locale of the email's recipients, which could be confusion.

So, we should not translate _any_ strings which will be used in an email.

MVrachev:
A good consideration.

Then the next logical question is should the emails for removed package, release or release file be translatable as a whole?

di:
Hi @MVrachev, can you resolve the merge conflicts here?

MVrachev:
I resolved them and updated my pr, but it seems Travis CI is stale...

di:
Opening/closing to try to resolve CI issues...

di:
Travis seems totally stuck here, but I verified that [the build](https://travis-ci.org/github/pypa/warehouse/builds/666543737) passed.

PR #7543, opened 2020-03-13
Invalidate Two Factor state after login

ewdurbin:
Currently our Two Factor Auth state is carried in the query string and expires after 300s, but this state remains reusable after a login is complete.

This checks against the last login for the user and refuses Two Factor attempts if the user has logged in since the token was created.

Thanks to @Julian for the report that led to this discovery!

### DIFF
--- a/warehouse/accounts/services.py
+++ b/warehouse/accounts/services.py
@@ -566,19 +566,14 @@ class TokenService:
             raise TokenMissing
 
         try:
-            if return_timestamp:
-                data, timestamp = self.serializer.loads(
-                    token, max_age=self.max_age, return_timestamp=True
-                )
-            else:
-                data = self.serializer.loads(token, max_age=self.max_age)
+            data = self.serializer.loads(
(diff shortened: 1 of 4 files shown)


warehouse/accounts/services.py
```diff
@@ -561,17 +561,24 @@ def __init__(self, secret, salt, max_age):
...
         return self.serializer.dumps({key: str(value) for key, value in data.items()})
 
-    def loads(self, token):
+    def loads(self, token, return_timestamp=False):
         if not token:
             raise TokenMissing
 
         try:
-            data = self.serializer.loads(token, max_age=self.max_age)
+            if return_timestamp:
```
dstufft: This could probably just be ``data = self.serializer.loads(token, max_age=self.max_age, return_timestamp=return_timestamp)``. It'll return a tuple in return timestamp cases, and a single value in not.

ewdurbin: ah, and just return `data` to the caller. yeah that's probably tidier.

PR #7582, opened 2020-03-23
External classifiers

di:
This PR moves the canonical source for classifiers out of PyPI's database and into the [`trove-classifiers`](https://github.com/pypa/trove-classifiers) project.

This is a necessary step to resolve https://github.com/pypa/packaging/issues/147 and move PyPI's metadata validation out of PyPI and into `packaging`.

PyPI's behavior here for the end user remains unchanged. For administrators, new & deprecated classifiers are no longer created through the UI, but as pull requests for `trove-classifiers` and dependency upgrades here.

PyPI still retains a `Classifier` model to help with efficiently searching for projects based on classifiers, but it is now essentially just a mapping from the classifier name to id. New rows in the classifiers table are created for new classifiers the first time a project is uploaded using that classifier.


tests/unit/admin/views/test_classifiers.py
```diff
@@ -1,134 +0,0 @@
...
-
-        db_request.params = {"parent_id": parent.id, "child": "Foobar"}
-        db_request.session.flash = pretend.call_recorder(lambda *a, **kw: None)
-        db_request.route_path = lambda *a: "/the/path"
-
-        with pytest.raises(sqlalchemy.exc.IntegrityError):
-            views.AddClassifier(db_request).add_child_classifier()
-
-    @pytest.mark.parametrize("parent_classifier", ["private", "Private", "PrIvAtE"])
-    def test_add_private_parent_classifier(self, db_request, parent_classifier):
```
xavfernandez: I liked that there were tests preventing the upload of `Private ::` classifiers.
Is it now tested elsewhere ?

ewdurbin: We still retain a CheckConstraint on the Classifier model to ensure that nothing that case-insensitive matches `'private ::%'` can be created in the DB. Which would act as a backstop in the event that a `trove-classifiers` release included such a classifier. It probably makes sense to do two things 1) include a test in warehouse to verify that all classifiers from a given release of `trove-classifiers` can be created which would catch such a condition in CI rather than at deploy 2) include a test in `trove-classifiers` itself.

di: Great points. Done!

tests/unit/forklift/test_legacy.py
```diff
@@ -2786,6 +2820,102 @@ def test_upload_succeeds_creates_release(self, pyramid_config, db_request, metri
...
+        [
+            "private",
+            "private",
+            "private",
+            "Private",
+            "Private",
+            "Private",
+            "PrIvAtE",
+            "PrIvAtE",
+            "PrIvAtE",
```
merwok: I see 3 differrent forms in these 9 lines, or am I mistaken?

di: Oops, good catch. 👍

pradyunsg:
Hey @di, are we waiting on reviews from @ewdurbin and possibly others here?

di:
Need to merge/release https://github.com/pypa/trove-classifiers/pull/6 and update this here first, and waiting on a review from @ewdurbin at least (although we've chatted about it and I don't think there's much else to do here).

PR #7689, opened 2020-03-27
Roles: db migration to create constraints

### LINKED ISSUE #2745: De-duplicate "Owner" and "Maintainer" roles
Currently on pypi-legacy, it's possible for a user to have both the "Owner" and "Maintainer" roles.

However, an "Owner" can do everything a "Maintainer" can do, so it doesn't make sense for a user to be able to have both roles.

In the short term, we decided to add a hack to hide any "duplicate" roles in #2705, and remove the ability to add multiple roles for the same user in Warehouse.

Once pypi-legacy is shut down, we can run a migration which removes any of these "duplicate" roles., and remove this hack.

ewdurbin:
Fixes #2745


Makefile
```diff
@@ -154,9 +154,11 @@ ifneq ($(PR), false)
 endif
 
 initdb:
+	docker-compose run --rm web psql -h db -d postgres -U postgres -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname ='warehouse';"
```
ewdurbin: disconnects any other sessions to make `make initdb` run regardless of if `make serve` containers are accessing the db.

Makefile
```diff
@@ -154,9 +154,11 @@ ifneq ($(PR), false)
...
 	docker-compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS warehouse"
 	docker-compose run --rm web psql -h db -d postgres -U postgres -c "CREATE DATABASE warehouse ENCODING 'UTF8'"
 	xz -d -f -k dev/$(DB).sql.xz --stdout | docker-compose run --rm web psql -h db -d warehouse -U postgres -v ON_ERROR_STOP=1 -1 -f -
+	docker-compose run -T --rm web psql -h db -d warehouse -U postgres < dev/example-cleanroles.sql
```
ewdurbin: cleans testpypi dump that has duplicates.

warehouse/packaging/models.py
```diff
@@ -57,9 +57,12 @@
...
 
     __tablename__ = "roles"
-    __table_args__ = (Index("roles_user_id_idx", "user_id"),)
+    __table_args__ = (
+        Index("roles_user_id_idx", "user_id"),
+        UniqueConstraint("user_id", "project_id", name="_roles_user_project_uc"),
+    )
 
-    __repr__ = make_repr("role_name", "user_name", "package_name")
+    __repr__ = make_repr("role_name")
```
ewdurbin: This is weird, it never worked before and I just noticed that it explodes when tests kept breaking when it tried to render the repr.

ewdurbin: looks like it is a relic of when we had strings for 'user' and 'package' on the model.

di:
@dstufft Yeah, seemed roughly the same (and the CI runs look roughly the same as well)

ewdurbin:
I think this is all set! Caught a template bug in manual testing see https://github.com/pypa/warehouse/pull/7689/commits/91992428b1f1a7726da08fa1098d76da65082ee8

Otherwise, just further tidied up the helper method for `get_user_role_in_project`. This may be worth refactoring to be on the model itself a la... `Project.get_user_role(user)` which should return one of `Owner`, `Maintainer`, or None? But I don't think it's worth tacking into this PR.

ewdurbin:
Also confirming that https://github.com/pypa/warehouse/pull/7689/commits/b38a674d4f446a868bb8826175927f03a9384aa5 doesn't seem to have a significant impact on our test suite performance.

PR #7699, opened 2020-03-30
Fix 7639

xmunoz:


tests/common/db/malware.py
```diff
@@ -59,5 +59,6 @@ class Meta:
     classification = factory.fuzzy.FuzzyChoice(list(VerdictClassification))
     confidence = factory.fuzzy.FuzzyChoice(list(VerdictConfidence))
     message = factory.fuzzy.FuzzyText(length=80)
+    run_date = factory.fuzzy.FuzzyNaiveDateTime(datetime.datetime.utcnow())
```
ewdurbin: I think this leaves a nonzero chance that two resulting fixtures will end up with the same timestamp, but given the change to testing against a `SqlalchemyOrmPage` object with query, that shouldn't cause a failure.

PR #7725, opened 2020-04-03
Remove comments from CSS bundle

yeraydiazdiaz:
I noticed our current CSS bundle includes all the comments in the SCSS files which results in quite a bit of bloat:

Screenshot 2020-04-03 at 14 36 50

This PR configures `cssnano` to remove the comments, reducing the final CSS file size from 168K to 74K.


nlhkabu:
Thanks @yeraydiazdiaz !

pradyunsg:
Should 1 instance of the license be included?

yeraydiazdiaz:
That's a good point, I thought about that too but assumed the practice is to remove as much as possible from the final artifact to optimize the bundle size.

That being said GitHub, for instance, does include the license in the minified CSS but a quick check on my currently opened tabs suggests it's not common at all.

I guess it's up to us to make the right decision, but frankly I don't know what that would be.

Do we include our Apache 2 license only?
Do we include the licenses of all dependencies if their license requires it as well?

PR #7726, opened 2020-04-03
Switch to psycopg2-binary

di:


yeraydiazdiaz:
I noticed this snippet the other day from their [README](https://github.com/psycopg/psycopg2#installation):

> The binary package is a practical choice for development and testing *but in production it is advised to use the package built from sources*.

Not sure if that applies to Docker images as well, but thought I should mention it.

di:
Hmm, good catch. Compiling this from source in CI is costing us a lot of time, but after chatting with @ewdurbin, we decided to hold off on merging this until we can run our tests in the same container image we use for production.

PR #7743, opened 2020-04-05
Pluralize stats on the front page

webknjaz:


di:
Thanks, fixed in https://github.com/pypa/warehouse/pull/7751.

PR #7749, opened 2020-04-06
Webpack frontend pipeline

yeraydiazdiaz:
Fixes https://github.com/pypa/warehouse/issues/6141

This PR moves our frontend pipeline to using Webpack only (currently not including admin, I'll get to that later).

Benefits:

- Overall faster compilation times, in the `static` Docker container (keep in mind I'm using an old MBP late 2013):
   - a "cold" production build (first build after container spin up) takes 41 seconds,
   - a "warm" production build (build triggered after update) takes 20 seconds,
   - a "cold" development build 35 seconds,
   - a "warm" development build 5.5 seconds. 🚀 
- Less tooling for our frontend pipeline.
- Allows production and development builds simply changing the `mode` argument.
- Leverage Webpack's built-in watch options.

Drawbacks:

- Webpack configuration is less straightforward than Gulp's and harder to debug.
- Our frontend setup differs from the way Webpack like to operate which makes the configuration a bit awkward in some places.
- There's a few extra files generated, mostly empty ones, as a consequence of the above.

Notes:

- I think the entrypoint for the `static` container should be a `development` mode. As [@dstufft's commented in the issue](https://github.com/pypa/warehouse/issues/6141#issuecomment-508089860) it might be safer to have the developer change the mode while developing and setting it back to production before commiting, though arguably the optimization steps to do the right thing anyway, but I'd like opinions on the matter.
- I've delayed adding the admin to the configuration mainly because it complicates things even further as there's a second dist directory to write to. Also because I feel most devs focus on warehouse or admin, having the two in the same config means having more assets to build. I'm leaning towards creating a separate configuration for admin and have a second container build those assets separately.
- From what I could gather production does not behave exactly like dev. In dev static assets are served via Pyramid and leverage the manifest.json which doesn't seem like it's happening in production? It'd be good to have some context as to how production works so we can minimize problems on deployment.

This is a fairly risky change but I think it's a worthwhile one. I'd like devs to try it out and see how it feels. Personally in my very old laptop things are much quicker than before.


dstufft:
Production should still be serving via Pyramid IIRC, there's just a CDN in front of it so it rarely actually hits the backend servers.

ewdurbin:
Indeed, the production static assets are served in exact same manner as development.

di:
> I think the entrypoint for the static container should be a development mode.

Given the extra tests for the JS pipeline in https://github.com/pypa/warehouse/pull/7658, I'm thinking I'm OK with this, especially given the speedup it will give.

> I'm leaning towards creating a separate configuration for admin and have a second container build those assets separately.

This makes sense to me.

yeraydiazdiaz:
Ok, this is ready for review and testing.

I had to call defeat on one issue: I could not get the manifest to produce a proper mapping from copied filenames to their equivalent filenames with hashes so they're simply pointing to the same name, i.e. `"/path/to/zxcvbn.js": "/path/on/dist/zxcvbn.js"`. This applies to copied webfonts as well. Not sure how much of a problem that is, but it's worth mentioning, otherwise it's all working as far as I've tested. 

To test development mode you will need to change the `docker-compose` line:

`command: bash -c "npx webpack --mode production --config webpack.warehouse.js --watch --info-verbosity verbose"`

To:

`command: bash -c "npx webpack --mode development --config webpack.warehouse.js --watch --info-verbosity verbose"`

Same thing for the `static_admin` service though I don't think there's a lot of editing done there. There's probably a more clever way of doing this but I am all webpack'd out atm, suggestions are welcome.

di:
> I could not get the manifest to produce a proper mapping from copied filenames to their equivalent filenames with hashes so they're simply pointing to the same name

Unfortunately I think this is a requirement for us and why we were using `gulp-rev-all` in the first place: our static assets are set to ~never expire.

Using the hash ensures that a given page in the cache is always getting the version of the JS/CSS bundle that was originally intended to work with it.

Without this hash in the filename, every time we publish a "breaking" change to any of the static assets, we'd have to purge all HTML pages from the cache.

It looks like Webpack might support this out of the box, have you tried using the configuration mentioned in https://webpack.js.org/guides/caching/#output-filenames ?

yeraydiazdiaz:
Makes sense, it's being handled properly now.

> It looks like Webpack might support this out of the box, have you tried using the configuration mentioned in https://webpack.js.org/guides/caching/#output-filenames ?

I did, yes, but it caused some issues when combining the CopyPlugin and the ManifestPlugin.

The problem was that I was trying adhere too much to the Gulp config and working against how Webpack likes to do things.

The solution I came up with is to not use the CopyPlugin at all and let Webpack extract the fonts as they're referenced in the CSS. We were copying _all_ the fonts before, even if they weren't being referenced from CSS, so you may see some difference in the number of assets.

Vendored JS (only `zxcvbn` atm) was slightly different because it's not explicitly imported from our JS code but included in the HTML template, so it has to be emitted explicitly. There's probably a more webpack-y way of doing this but it works fine for now.

I [added](https://github.com/pypa/warehouse/pull/7749/commits/ab7e082ba210e1ecff7a053fbeb949c46106c7f0) a `WEBPACK_MODE` env var to make it easier to select the build mode:`WEBPACK_MODE=development make serve` is all you need now.

I've done some light testing of this navigating through the main site and admin, everything seems to load and it looks fine but can I ask devs to please do test it thoroughly in case I've missed something. I usually open a Network dev panel while I navigate around  to make sure all assets are loaded, you should see no 404s, if you do please let me know.

Dockerfile
```diff
@@ -12,7 +12,7 @@ WORKDIR /opt/warehouse/src/
 RUN set -x \
     && apt-get update \
     && apt-get install --no-install-recommends -y \
-        libjpeg-dev nasm
```
yeraydiazdiaz, line 15: This [avoids compiling cwepb-bin from source](https://github.com/imagemin/cwebp-bin/issues/31#issuecomment-588383966) reducing the time it takes to complete `npm ci`.

yeraydiazdiaz:
Sorry for the succession of commits yesterday, I thought I fixed something then noticed something else broke, rinse and repeat... 😞 

Anyway, it's now ready for test and review.

di:
Hey @yeraydiazdiaz, now that #8127 is merged want to take a look at resolving the conflicts here?

di:
@yeraydiazdiaz Gentle ping here!

yeraydiazdiaz:
Hey @di, I solved the conflicts but it no longer works due to support for RTL. I tried with https://github.com/romainberger/webpack-rtl-plugin/ but it generated an RTL version per CSS file and I don't have the time to look into how to solve it right now.

yeraydiazdiaz:
I [~worked out~ hacked](https://github.com/pypa/warehouse/pull/7749/files#diff-f179f4047296dfa0661d6569ff14c3e9R41-R43) a solution that seems to work. The "warm" recompiling time has increased slightly to ~6-8s but it's still quite an improvement from what we had.

I have to say that it took me a fair bit of time to reacquaint myself with the setup and its complexities. While I think it the developer speed improvement justify this change, especially while developing for frontend tasks, it's likely configuration changes to it down the line will cause some frustration as the Gulp setup is much easier to grok.

PR #7751, opened 2020-04-06
Adding pluralization to the statistics block

Mingun:
The statistics block on main page always uses the plural form for the displayed messages. This change makes the message dependent on the actual number of statistics items.


di:
Thanks!

PR #7754, opened 2020-04-06
Fix pluralization of front page stats

ewdurbin:
Initial change caused issues as `pluralize` was trying to interpret a formatted number instead of an integer: https://sentry.io/share/issue/dd54ac64ea7d4550855117f85a460a85/

Jinja docs say that we can pass a different value to `pluralize`: https://jinja.palletsprojects.com/en/2.11.x/templates/#i18n


ewdurbin:
Thanks @di was trying to figure that out 😂

PR #7760, opened 2020-04-06
Normalize language names

webknjaz:
All locale names are capitalized, except for Spanish and French. I think it makes sense to have the same style for them too.

### DIFF
--- a/warehouse/i18n/__init__.py
+++ b/warehouse/i18n/__init__.py
@@ -23,8 +23,8 @@ from warehouse.cache.http import add_vary
 # https://github.com/django/django/blob/master/django/conf/locale/__init__.py
 KNOWN_LOCALES = {
     "en": "English",  # English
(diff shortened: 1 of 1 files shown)


di:
This is intentional: in these languages, language names are lowercased. We take these directly from https://github.com/django/django/blob/master/django/conf/locale/__init__.py so we shouldn't have to think about which ones follow this rule and which don't.

webknjaz:
@di that file has Ukrainian uppercase https://github.com/django/django/blob/master/django/conf/locale/__init__.py#L537. But in fact, there's no capitalization for language names in Ukrainian. So I'm not sure how trustworthy that source is.

My point is that there's a list of languages in the footer of PyPI. And they appear inconsistently. This is weird. Also, in languages that don't capitalize the language names, there's usually an additional word "language" appended in the canonical form.

PR #7765, opened 2020-04-07
don't add new projects/users to sitemap for 2 weeks

ewdurbin:
This might help to dissuade some spammers by inducing a 14 day lag from initial project creation until we publicize it in the sitemap and thus assist in getting it into search results.

It also provides a bit longer for detection so we can keep crap from landing there in the first place.

14 days feels reasonable to me for brand new projects and users?


warehouse/sitemap/views.py
```diff
@@ -65,13 +68,20 @@ def sitemap_index(request):
...
         .all()
     )
     users = (
         request.db.query(
             User.sitemap_bucket, func.max(User.date_joined).label("modified")
         )
+        .filter(
+            or_(
+                User.date_joined < datetime.utcnow() - AGE_BEFORE_INDEX,
+                User.date_joined == None,  # noqa: E711 SQLAlchemy requires ==
```
di, line 82:
```suggestion
                User.date_joined.is_(None),
```
?

warehouse/sitemap/views.py
```diff
@@ -108,10 +118,21 @@ def sitemap_bucket(request):
...
         .all()
     )
-    users = request.db.query(User.username).filter(User.sitemap_bucket == bucket).all()
+    users = (
+        request.db.query(User.username)
+        .filter(User.sitemap_bucket == bucket)
+        .filter(
+            or_(
+                User.date_joined < datetime.utcnow() - AGE_BEFORE_INDEX,
+                User.date_joined == None,  # noqa: E711 SQLAlchemy requires ==
```
di, line 131:
```suggestion
                User.date_joined.is_(None),
```

warehouse/sitemap/views.py
ewdurbin: hooray!

PR #7768, opened 2020-04-07
add rel="nofollow" to trending/latest on index page

ewdurbin:
this will help to ensure that new projects aren't indexed until they are added to the sitemap after 14 days, see #7765


ewdurbin:
the... the... the tests passed... travis are you there?

PR #7785, opened 2020-04-10
fix rendering of help page for simplified chinese

ewdurbin:
after 9d9324cc05a5e97be5f14dfefafa5f18b8896ac4 we started seeing [these sentry errors](https://sentry.io/share/issue/5285167912184934a4be88fe03b923c6/) which i tracked down to loading /help/ with the simplified chinese locale.

i'd be lying if i said i understood exactly what is going wrong at this point... but this does resolve it

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -274,40 +274,40 @@ msgstr ""
 #: warehouse/templates/pages/help.html:456
 #: warehouse/templates/pages/help.html:468
 #: warehouse/templates/pages/help.html:474
-#: warehouse/templates/pages/help.html:477
-#: warehouse/templates/pages/help.html:479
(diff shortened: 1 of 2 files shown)


ewdurbin:
It actually looks like the issue is a bad translation:

https://github.com/pypa/warehouse/blob/755d0c4f6310ef33fda89fbdb0279e6fdd8de54a/warehouse/locale/zh_Hans/LC_MESSAGES/messages.po#L4454-L4462

ewdurbin:
Now I'm even more confused as to why this PR stopped the failure.

I'm going to delete that translated string for now.

ewdurbin:
Ah, I see why this change "fixed" it, as it effectively marked the string as untranslated.

PR #7812, opened 2020-04-18
Tweak invalid-auth help

bhrutledge:
I discovered this addition from https://github.com/pypa/warehouse/pull/7424 while reviewing https://github.com/pypa/twine/issues/577. Much appreciated!

It seems like a common source of confusion with `twine upload` is the need for distinct accounts on PyPI and TestPyPI, so I attempted to address that here.

I still haven't set up a dev environment, so 🤞 that this works.


di:
Just an FYI @bhrutledge, when you change or add any translated text, you'll need to run `make translations` to rebuild the compiled list of translatable strings.

bhrutledge:
Thanks @di! I saw that in the Makefile, and was aiming to give it a try, but I'm glad you beat me to it.

Issue #7782, opened 2020-04-10
Eliminate extra space where email is hidden in logged-out profile view


di:
The optional email on a user's profile is leaving extra space when it's not being shown for logged-out users. For example, on https://pypi.org/user/admin:

Logged out | Logged in 
------------ | -------------
<img width="329" alt="Screen Shot 2020-04-09 at 7 01 55 PM" src="https://user-images.githubusercontent.com/294415/78950596-daa1b280-7a94-11ea-9966-7c020a1ef1a4.png"> | <img width="356" alt="Screen Shot 2020-04-09 at 7 01 43 PM" src="https://user-images.githubusercontent.com/294415/78950591-d83f5880-7a94-11ea-8abf-0c131480e016.png">

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/). If you are working on this issue and have questions, feel free to ask them here, [`#pypa-dev` on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or the [pypa-dev mailing list](https://groups.google.com/forum/#!forum/pypa-dev).

**Screenshot Required**: *If your pull request makes a visual change*, include a screenshot of your update. This helps our team give you feedback faster.

JohnHillegass:
Hey @di, I tried doing this with css like :blank but was having issues getting it to apply correctly. Another approach is to check request.authenticated_userid, I just need to set uses_session=True on the view_config. What do you think about taking this approach? Do you see any issues?

di:
Yes, we don't want this page to require a session because it would mean it's not cacheable, hence why we're including some parts of it which _do_ require a session (like this email address, which is only displayed for logged-in users) via a client-side include.

JohnHillegass:
Thanks for clarifying. I was eventually able to figure out a way to do it with html and css alone. Not 100% sure if I am making efficient use of the css. Please take a look when you can and let me know what you think.

PR #7820, opened 2020-04-20
7782 user profile eliminate extra email space

JohnHillegass:
For 
#7782

Use HTML and CSS to get rid of the extra space. Seems that this was an issue because p is a block element. The divs allow us to style them exactly as we'd like so offer more flexibility for this.


di:
Seems good to me, I'll let @nlhkabu review as well. Could you include a screenshot?

JohnHillegass:
Awesome! Thanks @di. Here are the screenshots:

Logged Out
Screen Shot 2020-04-20 at 7 56 45 PM

Logged In
Screen Shot 2020-04-20 at 7 56 15 PM

di:
LGTM!

nlhkabu:
Thank you @JohnHillegass! :tada:

PR #7828, opened 2020-04-22
upgrade to Python 3.8.2

ewdurbin:
`:=`


.travis.yml
```diff
@@ -31,6 +31,7 @@ install:
   - nvm install $TRAVIS_NODE_VERSION
   - pip install -U pip setuptools wheel
   - pip install -r requirements/main.txt -r requirements/deploy.txt -r requirements/docs.txt -r requirements/lint.txt -r requirements/tests.txt
+  - pip install -r requirements/dev.txt
```
ewdurbin, line 34: I'm honestly not _exactly_ sure what changed that required this, it seems `pip-tools` was being installed some other way before, might relate to bumping travis dist to `bionic`

requirements/main.txt
```diff
@@ -724,44 +720,50 @@ yara-python==3.11.0 \
...
-    --hash=sha256:a0c39e2535a7e9c195af956610dba5a1073071d2d85e9d2e5d789463f63e52ab \
-    --hash=sha256:a15e75d284178afe529a536b0e8b28b7e107ef39626a7809b4ee64ff3abc9127 \
-    --hash=sha256:a6a6ff82f5f9b9702478035d8f6fb6903885653bff7ec3a1e011edc9b1a7168d \
-    --hash=sha256:b639f72b95389620c1f881d94739c614d385406ab1d6926a9ffe1c8abbea23fe \
-    --hash=sha256:bad44274b151d46619a7567010f7cde23a908c6faa84b97598fd2f474a0c6891 \
-    --hash=sha256:bbcef00d09a30948756c5968863316c949d9cedbc7aabac5e8f0ffbdb632e5f1 \
-    --hash=sha256:d788a3999014ddf416f2dc454efa4a5dbeda657c6aba031cf363741273804c6b \
-    --hash=sha256:eed88ae03e1ef3a75a0e96a55a99d7937ed03e53d0cffc2451c208db445a2966 \
-    --hash=sha256:f99451f3a579e73b5dd58b1b08d1179791d49084371d9a47baad3b22417f0317 \
+zope.interface==4.7.2 \
```
ewdurbin, line 727: zope.interface==4.7.0 added Python 3.8 support, upgrade to latest zope.interface before 5.0.0

requirements/main.txt
```diff
@@ -294,10 +294,6 @@ idna==2.9 \
     --hash=sha256:7588d1c14ae4c77d74036e8c22ff447b26d0fde8f007354fd48a7814db15b7cb \
     --hash=sha256:a068a21ceac8a4d63dbfd964670474107f541babbd2250d61922f029858365fa \
     # via email-validator, requests
-importlib-metadata==1.6.0 \
```
ewdurbin, line 297: no need for this back port anymore!

requirements/main.txt
jamadden: Just curious: Are there known problems with zope.interface 5.0 in this code base? I contributed almost all of the changes to zope.interface 5, and though it is technically backwards incompatible, in practice I think I found two legacy test cases in all the dependencies of Zope and Plone that had very minor issues.

docs/api-reference/legacy.rst
```diff
@@ -120,7 +120,7 @@ The API endpoint served at `upload.pypi.org/legacy/
 <https://upload.pypi.org/legacy/>`_ is Warehouse's emulation of the
 legacy PyPI upload API. This is the endpoint that tools such as `twine
 <https://twine.readthedocs.io/>`_ and `distutils
-<https://docs.python.org/3.7/distutils/packageindex.html#the-upload-command>`_
+<https://docs.python.org/3.8/distutils/packageindex.html#the-upload-command>`_
```
di, line 123: This link hasn't existed since 3.6, we should probably just talk about `twine` here.

requirements/main.txt
ewdurbin: I didn't try 5.0, just took the shortest leap I could to get this working with 3.8 and will let dependabot upgrade us further. Though I'm really not sure why dependabot hasn't offered us an upgrade yet otherwise.

di: Perhaps one of our dependencies that depend on `zope.interface` (pyramid, pyramid-retry, pyramid-services, repoze.sendmail, transaction, wired, zope.sqlalchemy) has a `zope.interface<5` subdependency?

.travis.yml
ewdurbin: also while tracking this down I noticed that in CI we are installing everything into the virtualenv travis automatically creates for us, but then reinstalling it into the `.state/env` we construct... but never use.

di:
It seems like `pip-compile` is doing a kind of funny thing where it won't add additional hashes or upgrade a subdependency if the "outfile" already exists and the dependency is already listed there. 

I removed all the `.txt` files and recompiled all dependencies, which gave us some additional annotations, some new hashes for late-added built distributions (`cryptography`, `sqlalchemy`), and upgrades for some subdependencies (`zope.deprecation`, `zope.interface` and `zope.sqlalchemy`)

pradyunsg:
*looks at this* nice.
*looks at pip* oh right, Python 2. One day, we'll get here too.

PR #7841, opened 2020-04-23
Yank updates

di:
This PR:

* displays the yanked status of a release in the admin UI
* ensures that if only yanked releases are available

Fixes https://sentry.io/organizations/python-software-foundation/issues/1504635020/

### DIFF
--- a/warehouse/packaging/views.py
+++ b/warehouse/packaging/views.py
@@ -36,6 +36,7 @@ def project_detail(project, request):
         return HTTPMovedPermanently(request.current_route_path(name=project.name))
 
     try:
+        # Try to find a non-yanked release
         release = (
             request.db.query(Release)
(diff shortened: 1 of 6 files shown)


warehouse/packaging/views.py
```diff
@@ -36,17 +36,33 @@ def project_detail(project, request):
         return HTTPMovedPermanently(request.current_route_path(name=project.name))
 
     try:
+        # Try to find a non-yanked release
         release = (
             request.db.query(Release)
             .filter(Release.project == project, Release.yanked.is_(False))
```
dstufft, line 42: If you remove ``Release.yanked.is_(False)`` from here, you can make the first condition of the ``ORDER BY`` be ``Release.yanked``. That should implement this behavior without duplicating the query.

PR #7853, opened 2020-04-25
Allow manylinux_x_y wheel upload (PEP600)

mayeut:
manylinux_x_y platform tag is official per PEP600

see also pypa/manylinux#542


mayeut:
[PEP600](https://www.python.org/dev/peps/pep-0600/) does not restrict x & y other than those being numbers. However, it is said that a package index might decide to be conservative and reject wheels by restricting x & y a bit more.

It might make sense to restrict `x` to `2` for now. For `y`, it could be left unchecked or restricted to be between `5` (manylinux1) & `31` (latest glibc version is `2.31`) however this will require to be updated now and then.

mayeut:
@di,

Could you review and/or assign reviewers please ?

Support for PEP600 [under development](https://github.com/pypa/packaging/pull/293) in `packaging` as a prelude to `pip`.

mattip:
Packaging, pip, and wheel now support PEP 600 tags (the heavy lifting is done by packaging, pip vendors in a copy and wheel has a dependency on packaging).

di:
Thanks @mattip and @mayeut for your work here!

PR #7882, opened 2020-05-02
Add icon for change log etc project URL

scop:
Getting to package change logs, release notes or the like is currently a manual task, as links to them are not defined just about anywhere except perhapse in description prose. I think it would be good to promote adding them.

For example, when I see a new release from a project RSS feed being made, I go to PyPI and... start manually sifting through pages here and there to find what's changed in the release. Would be very nice to have a direct link in project links for that.

Although this is beyond PyPI, adding a recognizable specific icon could be a step towards that direction.

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -21,6 +21,8 @@
    {% set icon = "fas fa-cloud-download-alt" %}
 {% elif name|lower in ["home", "homepage", "home page"] %}
   {% set icon = "fas fa-home" %}
+{% elif name|lower in ["changelog", "change log", "changes", "release notes", "news", "what's new"] %}
+  {% set icon = "fas fa-gift" %}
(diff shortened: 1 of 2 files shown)


di:
Thanks @scop!

warehouse/templates/packaging/detail.html
```diff
@@ -21,6 +21,8 @@
    {% set icon = "fas fa-cloud-download-alt" %}
 {% elif name|lower in ["home", "homepage", "home page"] %}
   {% set icon = "fas fa-home" %}
+{% elif name|lower in ["changelog", "change log", "changes", "release notes", "news", "what's new"] %}
```
webknjaz, line 24: @scop I often use category prefixes for links. See https://pypi.org/project/aiohttp/. So it'd be better to match if any of these are in the name and not just match exactly. In this case, I'd rather use `Docs: changelog` to follow the same style as `Docs: RTD`.

scop: I'm sure the warehouse maintainers would consider such a PR :)

takluyver:
BTW, are these recognised names documented anywhere? Or even just one preferred form for each category? As a project maintainer, it would be good to be able to see at a glance what names I should use, because you're never going to cover every possibility.

di:
@takluyver Not yet: https://github.com/pypa/warehouse/issues/5947

PR #7896, opened 2020-05-06
Use change log icon for history URLs

scop:
Related to #7882, some projects call their change logs by the name "History".

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -21,7 +21,7 @@
    {% set icon = "fas fa-cloud-download-alt" %}
 {% elif name|lower in ["home", "homepage", "home page"] %}
   {% set icon = "fas fa-home" %}
-{% elif name|lower in ["changelog", "change log", "changes", "release notes", "news", "what's new"] %}
+{% elif name|lower in ["changelog", "change log", "changes", "release notes", "news", "what's new", "history"] %}
   {% set icon = "fas fa-gift" %}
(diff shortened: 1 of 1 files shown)


di:
"History" makes me think that a clock icon might make more sense here, like https://fontawesome.com/icons/clock?style=solid. Thoughts?

scop:
I think it serves the same purpose as changelog etc, and therefore should have the same icon.

di:
Sorry, I meant that all of these should change. I'm not sure the "gift" icon is very intuitive. @nlhkabu thoughts?

scop:
FWIW gift is what I've seen used in various services elsewhere. At least Slack, maybe Firefox, I forget exactly which all.

nlhkabu:
How about we use `fa-history` for:  "changelog", "change log", "changes", "history"
We are already using this icon in other places, so that would be consistent.

Then we can use `fa-newspaper` for: "release notes", "news", "what's new"

What do you think?

scop:
`fa-history` is currently used for in-warehouse release history etc links, and I would not reuse it here where it always points outside -- otherwise links that look the same would kind of be acting entirely differently from end user POV.

I think `fa-gift` is a better and likely more familiar choice (to others besides me as well ;)) than `fa-newspaper` for this. Subjectively also better looking than `fa-newspaper` or `fa-history`.

Also, as these project URLs serve the same purpose, they should share the same icon.

BTW the `release history` in addition to `history` title should likely also be added to this list too. Then if `fa-history` would be used for these, it would be even more odd as the link would possibly look exactly the same as the one in the Navigation section but lead elsewhere altogether.

di:
@nlhkabu Can you make a final call here?

PR #7916, opened 2020-05-09
Add ability for maintainers to provide reason for yanking a release

### LINKED ISSUE #7856: Give maintainers the ability to provide reason for yanking
PEP 592 permits the use of an arbitrary string which gives a reason for yanking:

> The value of the `data-yanked` attribute, if present, is an arbitrary string that represents the reason for why the file has been yanked. Tools that process the simple repository API MAY surface this string to end users.

PyPI does not currently support this. Completing this issue would include a UI to allow the user to set this when yanking, an update to the DB model to change this from a boolean to a text field, and an update to the simple page to surface this value if present.

---

### LINKED ISSUE #7886: /simple serves HTML that can't be parsed by Python's xml.etree if package has yanked releases
**Describe the bug**

Parsing HTML served by `/simple` endpoint results in `xml.etree.ElementTree.ParseError`.

**Expected behavior**

No parse error, as it was before when there were no yanked releases yet or with packages that don't have any yanked releases (yet).

**To Reproduce**

* Python script `test.py` that contains:
  ```python
  import requests
  from xml.etree import ElementTree
  simple_pip = requests.get('https://pypi.python.org/simple/pip')
  ElementTree.fromstring(simple_pip.text)
  ```

* run it with `python test.py`, for example (on macOS):
  ```
  $ python3 test.py
    File "test.py", line 4, in <module>
      ElementTree.fromstring(simple_pip.text)
    File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xml/etree/ElementTree.py", line 1315, in XML
    parser.feed(text)
  xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 143, column 306
  ```

The problem is the `data-yanked` part in lines like:

```
<a href="https://files.pythonhosted.org/packages/8c/5c/c18d58ab5c1a702bf670e0bd6a77cd4645e4aeca021c6118ef850895cc96/pip-20.0.tar.gz#sha256=5128e9a9401f1d16c1d15b2ed766a79d7813db1538428d0b0ce74838249e3a41" data-requires-python="&gt;=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*" data-yanked>pip-20.0.tar.gz</a><br/>
```

**My Platform**

* macOS 10.15.4 with Python 2.7.16 or 3.7.7 (but same issue occurs on other platforms too)

**Additional context**

* EasyBuild bug report: https://github.com/easybuilders/easybuild/issues/619
* we've worked around this in the upcoming EasyBuild version by stripping out the `data-yanked` part (see https://github.com/easybuilders/easybuild-framework/pull/3303), but this issue still occurs in EasyBuild releases that worked fine perfectly before package releases were getting yanked

daneah:
Resolves #7856 
Fixes #7886


warehouse/templates/manage/release.html
```diff
@@ -142,7 +142,17 @@ <h3>{% trans %}Yank release{% endtrans %}</h3>
...
+    {% set extra_description %}
+      <p>
+        {% trans project_name=project.name, version=release.version %}
+          You may provide a reason for yanking this release, which will be displayed by pip to users who install <code>{{ project_name }}=={{ version }}</code>.
+        {% endtrans %}
+      </p>
+      <label for="yanked_reason">{% trans %}Reason (optional){% endtrans %}</label>
+      <input name="yanked_reason" id="yanked_reason" type="text" autocomplete="off" autocapitalize="off" spellcheck="false">
+    {% endset %}
+    {{ confirm_button(gettext("Yank release"), gettext("Version"), "yank_version", release.version, modifier="--primary", warning=False, extra_description=extra_description) }}
```
daneah, line 155: Two things I don't think I feel great about here:

1) `extra_description` probably isn't meant for this, but there wasn't another existing way to tie into `confirmation_button`/`confirmation_modal` and I wasn't sure I should introduce a whole new block for this singular use case.
2) The [duplication of this in `releases.html`](https://github.com/pypa/warehouse/pull/7916/files#diff-aa165515152660253988628f2a9af2caR118-R127) doesn't feel good so I'd need a suggestion about an approach befitting this code base to, say, put it in a shared location both templates could use.

tests/unit/email/test_init.py
```diff
@@ -2057,6 +2066,7 @@ def test_send_removed_project_release_file_email_to_maintainer(
             version="0.0.0",
             project=pretend.stub(name="test_project"),
             created=datetime.datetime(2017, 2, 5, 0, 0, 0, 0),
+            yanked_reason="",
```
daneah, line 2069: I had to add `yanked_reason` to all the stubs in this file, with the alternative being to use `getattr(release, "yanked_reason")` in `warehouse/email/__init__.py`, which felt less clean 🤔

daneah:
@di Let me know if you have any guidance on the above notes! Is it okay if I force push a rebase of this branch, or do y'all prefer a fresh PR?

di:
@daneah This is on my list to review today. A force push is totally fine!

warehouse/templates/manage/release.html
di: The `confirm_button`/`confirm_modal` pattern is generally getting a bit unwieldy, and we probably should think about ways to improve it, but I think this is probably fine as is for now.

daneah:
@di thanks for the review! I can definitely address some of those points readily, and I'm sure I'll have a couple of follow up questions 😄 will report back.

di:
Thanks @daneah!

pradyunsg:
Wohoo! Thanks @daneah and @di! ^>^

PR #7927, opened 2020-05-11
Support RTL locales

di:
This gets us most of the way to supporting RTL locales, but will probably need some specific tweaks from @nlhkabu.

This PR also enables Hebrew for testing RTL locales.

TODO:
- [x] Update tests
- [x] Update translations
- [x] ???

| LTR  | RTL | 
| - | - |
| Screen Shot 2020-05-11 at 3 44 06 PM  | Screen Shot 2020-05-11 at 3 49 14 PM |

Fixes #6683.

PR #7968, opened 2020-05-17
Fix strings not marked for translation

nlhkabu:
Found while doing #7927

### DIFF
--- a/warehouse/templates/manage/account.html
+++ b/warehouse/templates/manage/account.html
@@ -437,7 +437,8 @@
         </ul>
       </div>
       <div>
-        <input value="Update password" class="button button--primary" type="submit" data-target="password-match.submit">
+        {% set updateButtonText=gettext("Update password") %}
+        <input value="{{ updateButtonText }}" class="button button--primary" type="submit" data-target="password-match.submit">
       </div>
     </form>
   </section>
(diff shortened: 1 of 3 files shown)


warehouse/templates/manage/account.html
```diff
@@ -437,7 +437,8 @@ <h2 class="sub-title">{% trans %}Change password{% endtrans %}</h2>
         </ul>
       </div>
       <div>
-        <input value="Update password" class="button button--primary" type="submit" data-target="password-match.submit">
+        {% set updateButtonText=gettext("Update password") %}
+        <input value="{{ updateButtonText }}" class="button button--primary" type="submit" data-target="password-match.submit">
```
di, line 441: Just curious, why use `gettext` over `trans` here?

```suggestion
        <input value="{% trans %}Update Password{% endtrans %}" class="button button--primary" type="submit" data-target="password-match.submit">
```

nlhkabu: I have no idea :tired_face: :laughing:

di: OK, fine with me either way I think. If you want to accept this suggestion you'll likely need to run `make translations` again though.

nlhkabu:
Thanks @di.
Translations updated. This should be ready to merge.

PR #8011, opened 2020-05-28
WIP: Docs: Update sample author name and email

brainwane:
Per https://groups.google.com/d/msg/pypa-dev/rUNsfIbruHM/LCEx-CB5AgAJ
the pypa-dev Google Group is now decommissioned.
Using this opportunity to, as in
pypa/sampleproject#118, improve the example author
name and email address.

Signed-off-by: Sumana Harihareswara <sh@changeset.nyc>

### DIFF
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -28,135 +28,316 @@ Project
         HTTP/1.1 200 OK
         Content-Type: application/json; charset="UTF-8"
 
-        {
-            "info": {
-                "author": "A. Random Developer",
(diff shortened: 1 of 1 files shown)


di:
Since this is an example of an actual API response, we should probably make a new release of `sampleproject` and then update this to be what PyPI would actually return instead.

brainwane:
That sounds fair.

di:
A new release of `sampleproject` has been made: https://pypi.org/project/sampleproject/2.0.0/

Issue #5222, opened 2018-12-21
Display "last released" date in sorted-by-relevance search results 


brainwane:
When the user [searches for, e.g., `short-url`](https://pypi.org/search/?q=short-url), in the default view (sorted by relevance), we could display the last-released date on each item, just as we do [when sorting by last updated](https://pypi.org/search/?q=short-url&o=-created). 

Requested by @koobs in IRC (`#pypa-dev`) today:

> Makes it easier to filter 'relevance results' by last updated for large result sets
> when you search on PyPI, it defaults to a 'Relevance' search. the results in this Relevance (order by) mode dont display 'Last updated' attribute of the package in the results
> When you switch to 'Last updated' order by, 'Last updated' values are shown
> What I'm proposing is exposing/displaying the 'Last updated' values in the 'Relevance' order by case as well
> Allowing users to use the 'relevance' matching which highlights (as expected) 'more' relevant results, but also allows them to see last updated, instead of going to the package page, to see last updated
> my example case: https://pypi.org/search/?q=short+url
> when you switch that search to last updated, most of the results aren't relevant for the search terms any more
<koobs> im looking for 'relevent' results 'updated recently'
<koobs> without 'Last updated' being displayed in relevence mode, i cant do that
~~~

sanjaysiddhanti:
I can work on this issue

brainwane:
Great! Please go ahead!

brainwane:
Thanks, @sanjaysiddhanti! What will you be working on next?

sanjaysiddhanti:
@brainwane I'm not sure -- do you have any suggestions?

I was working on the spec for https://github.com/pypa/warehouse/issues/8158, but we might be at a blocker there because we don't want to commit such a large file to the git repo and slow everything down.

sanjaysiddhanti:
Would be happy to work on any issues you have in mind, or I can look through the open issues

brainwane:
@sanjaysiddhanti I think you should go ahead and look through the ["help needed" issues](https://github.com/pypa/warehouse/labels/help%20needed)!

PR #8027, opened 2020-06-02
Move show password checkbox under password field

### LINKED ISSUE #6722: Move "show password" checkbox on login form

sanjaysiddhanti:
@di 

This PR implements https://github.com/pypa/warehouse/issues/6722. 

Screenshot before:

Screen Shot 2020-06-01 at 10 47 17 PM

Screenshot after:

Screen Shot 2020-06-01 at 10 50 33 PM

Let me know if you'd like any additional changes.

Fixes #6722.


sanjaysiddhanti:
I could also move "show password" down to be horizontally aligned with "Log in" and "forgot password." I wasn't sure where exactly you wanted it. This does make password one tab away from username as the ticket requests.

di:
> I could also move "show password" down to be horizontally aligned with "Log in" and "forgot password."

Yes, please do that! I think that can be done with the `split-layout` class. We should also remove the `split-layout` class where it's no longer being used, like https://github.com/pypa/warehouse/issues/6722#issuecomment-538182534 describes.

sanjaysiddhanti:
Ok, I removed the `split-layout` where it is no longer used. When I apply the rest of the diff from https://github.com/pypa/warehouse/issues/6722#issuecomment-538182534, it basically looks like my screenshot above, without horizontal alignment.

How does d845979 look? I did the following:

- Moved "show password" into the form group that has the log in and forgot password buttons
- This broke the "show password" functionality. Looks like clicking on "show password" triggers some JS code, but it needs to be in a block that connects it to the Password Controller. So, I moved the entire form group into the `<div data-controller="password" class="form-group">` block. Is there a better solution?

New screenshot below

image

warehouse/templates/accounts/login.html
```diff
@@ -86,16 +82,19 @@ <h1 class="page-title">{% trans title=title %}Log in to {{ title }}{% endtrans %
...
+              <div>
+                <input type="submit" value="{% trans %}Log in{% endtrans %}" class="button button--primary">
+              </div>
+              <label for="show-password">
+                <input data-action="change->password#togglePasswords" data-target="password.showPassword"
+                  id="show-password" type="checkbox">&nbsp;{% trans %}Show password{% endtrans %}
+              </label>
+              <span>
+                <a href="{{request.route_url('accounts.request-password-reset')}}">{% trans %}Forgot password?{% endtrans %}</a>
+              </span>
```
di, line 96: I _think_ wrapping these in a `div` will cause them to both be moved to the right side, but stacked on top of each other

```suggestion
              <div>
                <label for="show-password">
                  <input data-action="change->password#togglePasswords" data-target="password.showPassword"
                    id="show-password" type="checkbox">&nbsp;{% trans %}Show password{% endtrans %}
                </label>
                <span>
                  <a href="{{request.route_url('accounts.request-password-reset')}}">{% trans %}Forgot password?{% endtrans %}</a>
                </span>
              </div>
```

sanjaysiddhanti:
>  will cause them to both be moved to the right side, but stacked on top of each other

Ah, I didn't realize that was the goal. Your suggested diff doesn't quite do it, since `label` and `span` are both inline (see below).

Screen Shot 2020-06-02 at 5 56 23 PM

But when I wrap the `label` in a `div` in b98f393, I think it looks right

Screen Shot 2020-06-02 at 5 58 04 PM

sanjaysiddhanti:
@nlhkabu No rush, but just a friendly ping on this one in case it got lost.

warehouse/templates/accounts/login.html
```diff
@@ -86,12 +82,15 @@ <h1 class="page-title">{% trans title=title %}Log in to {{ title }}{% endtrans %
...
             {% endif %}
           </div>
-        </div>
-
-        <div class="form-group">
-          <div class="split-layout split-layout--table">
-            <div>
-              <input type="submit" value="{% trans %}Log in{% endtrans %}" class="button button--primary">
+          <div class="form-group">
+            <div class="split-layout margin-top--large margin-bottom--large">
```
sanjaysiddhanti, line 86: removed `split-layout--table`, as it causes elements to be vertically centered in the container, whereas Nicole's screenshot shows "Show password" aligned to the top of the div

sanjaysiddhanti:
Thanks for the feedback! Here is an updated version:

image

The only difference I see is that Nicole's version has double the margin between "Log in" and "Show password." I tried to achieve this with the following, but the margins were getting collapsed with the sibling element:

```
            <span class="display-block margin-top--large">
              <a href="{{request.route_url('accounts.request-password-reset')}}">{% trans %}Forgot password?{% endtrans %}</a>
            </span>
```
I also tried wrapping the `span` in a `div class="margin-top--large"` 

@di did you have an idea of how to get this extra margin and does it matter to you?

patelneel55:
Since there's now a new feature request regarding a "Keep me signed in" button at #8033, would it be beneficial to switch to a :eye: icon, as mentioned by @bittner at [#6722](https://github.com/pypa/warehouse/issues/6722#issuecomment-613121450), to prevent a clutter of text fields under the password field?

Any thoughts? @di @sanjaysiddhanti

bittner:
Nice! Adding an :eye: icon can maybe done easily now. Let's see. :smiley_cat: 

Thanks for this improvement so far! :1st_place_medal:

bittner:
Wow! This change is already rolled out. Nice!

Is the pipeline that does this public? It doesn't seem to be the [GitHub Actions](https://github.com/pypa/warehouse/actions) of this repository. :thinking:

brainwane:
Hi @sanjaysiddhanti! Congrats on landing your first pull request in PyPA projects. Could I ask you to take a look at #5222 and comment there if you can work on it?

PR #8093, opened 2020-06-11
Public BigQuery table for distribution metadata 

### LINKED ISSUE #7403: Public Dataset for distribution metadata
As requested in https://github.com/pypa/packaging-problems/issues/323, we should explore publishing the metadata for each released distribution in a public dataset via BigQuery.

I'm imagining that each row would contain all the core metadata fields included in each release, as well as filename, digests, file size, upload time, URL to the distribution, etc. Essentially everything in the ["Release" JSON API](https://warehouse.readthedocs.io/api-reference/json/#release), with the per-release `info` field included for every individual distribution.

Once we're publishing to the dataset on upload, we'd also need to backfill prior distributions as well.

Not entirely sure what we'd name it, does `the-psf:pypi.distributions` make sense?

patelneel55:
Per the feature request in #7403, this PR creates a new celery task every time a release is uploaded to PyPI using the `legacy.py` endpoint. The task extracts the metadata from the release and pushes it to a public BigQuery table asynchronously to avoid blocking and failing the main upload thread. The `schema` was decided, with input from @di, based on the [core metadata specifications](https://packaging.python.org/specifications/core-metadata/) and the [Release JSON API](https://warehouse.pypa.io/api-reference/json/#release).

<strike>Before this can be merged and deployed,</strike>* There is still some discussion to be had on the optimal approach to create the table and bulk load pre-existing releases to BigQuery. This could be in the form of a script which can be provided to one of the maintainers to run without the need to include it into the `warehouse` repo. 

~~Also, it would be great if we could notify maintainers when certain tasks fail so that they can take action if necessary. In the case of BigQuery, if BigQuery happens to be down, the task will be re-run 5 times before it fails silently. I could not find a service that could notify maintainers on any such errors. However, I believe this can be its own issue and PR~~. We now use the logger to notify `on_failure` for all celery tasks.

Resolves #7403 

Checklist to fulfill the issue:
- [x] [Create script to Bulk load data into BigQuery](https://github.com/pypa/warehouse/pull/8142)
- [x] ~~Update readthedocs documentation~~ Updated help text to indicate that there is a public table for metadata now
- [ ] (Maintainers) Add the table to the appropriate environment variable mapping in prod (`WAREHOUSE_RELEASE_FILES_TABLE`)

<details>
<summary>
<strike>Table Schema</strike>
</summary>

```
[
    {
        "name": "id",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "metadata_version",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "project_id",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "name",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "version",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "summary",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "description_id",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "description",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "description_content_type",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "author",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "author_email",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "maintainer",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "maintainer_email",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "license",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "keywords",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "classifiers",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "platform",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "home_page",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "download_url",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "requires_python",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "requires",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "provides",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "obsoletes",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "requires_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "provides_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "obsoletes_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "requires_external",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "project_urls",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "created",
        "type": "TIMESTAMP",
        "mode": "REQUIRED"
    },
    {
        "name": "yanked",
        "type": "BOOLEAN",
        "mode": "REQUIRED"
    },
    {
        "name": "yanked_reason",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "uploader_id",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "uploaded_via",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "files",
        "type": "RECORD",
        "mode": "REPEATED",
        "fields": [
            {
                "name": "filename",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "size",
                "type": "INTEGER",
                "mode": "REQUIRED"
            },
            {
                "name": "path",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "python_version",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "packagetype",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "comment_text",
                "type": "STRING",
                "mode": "NULLABLE"
            },
            {
                "name": "has_signature",
                "type": "BOOLEAN",
                "mode": "REQUIRED"
            },
            {
                "name": "md5_digest",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "sha256_digest",
                "type": "STRING",
                "mode": "REQUIRED"
            },
            {
                "name": "blake2_256_digest",
                "type": "STRING",
                "mode": "REQUIRED"
            }
        ]
    }
]
```
</details>

<details>
<summary>
Table Schema (Updated)
</summary>

```
[
    {
        "name": "metadata_version",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "name",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "version",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "summary",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "description",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "description_content_type",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "author",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "author_email",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "maintainer",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "maintainer_email",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "license",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "keywords",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "classifiers",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "platform",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "home_page",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "download_url",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "requires_python",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "requires",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "provides",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "obsoletes",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "requires_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "provides_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "obsoletes_dist",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "requires_external",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "project_urls",
        "type": "STRING",
        "mode": "REPEATED"
    },
    {
        "name": "uploaded_via",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "upload_time",
        "type": "TIMESTAMP",
        "mode": "REQUIRED"
    },
    {
        "name": "filename",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "size",
        "type": "INTEGER",
        "mode": "REQUIRED"
    },
    {
        "name": "path",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "python_version",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "packagetype",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "comment_text",
        "type": "STRING",
        "mode": "NULLABLE"
    },
    {
        "name": "has_signature",
        "type": "BOOLEAN",
        "mode": "REQUIRED"
    },
    {
        "name": "md5_digest",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "sha256_digest",
        "type": "STRING",
        "mode": "REQUIRED"
    },
    {
        "name": "blake2_256_digest",
        "type": "STRING",
        "mode": "REQUIRED"
    }
]
```
</details>

<ins>*Update (6/17):</ins> 
After discussion with @di, the `schema` was updated to reflect each row in BigQuery as a `release file` rather than a `release` with nested `files` objects. This change was made because there's a possibility that new files are added to a release after its been published. BigQuery isn't designed for high-frequency updates which would make a lot of transactions, including the data migration, convoluted and expensive. 

Once the script is ready, the migration of data can take place as follows to maintain data integrity:
- Create table with the schema above on BigQuery
- Merge this PR first so that the task can start uploading new data to BigQuery
- Note the first record with the earliest `upload_time` and provide it to the [migration script](https://github.com/pypa/warehouse/pull/8142)
- The script will migrate all the necessary data points until the provided `upload_time` ensuring no release file was missed


warehouse/packaging/tasks.py
```diff
@@ -111,3 +116,108 @@ def update_description_html(request):
...
+    bq = request.find_service(name="gcloud.bigquery")
+
+    # Using the schema to populate the data allows us to automatically
+    # set the values rather than assigning values individually
+    def populate_data_using_schema(schema, row_data):
+
+        for sch in schema:
+            if sch.field_type == "RECORD":
+                row_data[sch.name] = dict()
+                populate_data_using_schema(sch.fields, row_data[sch.name])
```
patelneel55, line 142: If a `RECORD` type field is set to the `REPEATED` mode, they need to be enclosed in an array

patelneel55: This is now resolved as the `schema` change invalidated this bug

tests/unit/forklift/test_legacy.py
```diff
@@ -1359,6 +1360,14 @@ def storage_service_store(path, file_path, *, meta):
                 IMetricsService: metrics,
             }.get(svc)
         )
+        db_request.registry.settings = {
+            "warehouse.distribution_table": "example.pypi.distributions"
+        }
+
+        update_database = pretend.stub(
```
ewdurbin, line 1367: nit: suggest changing this to "update_bigquery" just to clarify that it isn't our "core" database.

warehouse/tasks.py
```diff
@@ -119,6 +122,9 @@ def _after_commit_hook(self, success, *args, **kwargs):
         if success:
             super().apply_async(*args, **kwargs)
 
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
```
ewdurbin, line 125: This is great! Thank you.

I'm actually wondering if we could get this into the Error logger so Sentry would capture them.

warehouse/packaging/tasks.py
```diff
@@ -111,3 +113,54 @@ def update_description_html(request):
...
+@tasks.task(
+    bind=True,
+    ignore_result=True,
+    acks_late=True,
+    autoretry_for=(Exception,),
+    retry_backoff=15,
+    retry_jitter=False,
+    max_retries=5,
+)
+def upload_bigquery_distributions(task, request, file, form):
```
ewdurbin, line 127: `update_bigquery_release_files` might be a better name to match things up across the DB models and service names.

warehouse/config.py
```diff
@@ -159,6 +159,7 @@ def configure(settings=None):
     maybe_set(settings, "aws.region", "AWS_REGION")
     maybe_set(settings, "gcloud.credentials", "GCLOUD_CREDENTIALS")
     maybe_set(settings, "gcloud.project", "GCLOUD_PROJECT")
+    maybe_set(settings, "warehouse.distribution_table", "WAREHOUSE_DISTRIBUTION_TABLE")
```
ewdurbin, line 162: I'd prefer `warehouse.release_files_table`/`WAREHOUSE_RELEASE_FILES_TABLE`, again to keep terms consistent among DB models. I don't recall that we use the word "distribution" anywhere in the codebase yet...

patelneel55:
@ewdurbin Thanks for the review! I'll get working on the requested changes.

> Also, I'd be interested in a "synchronization" task. Tasks can fail and it'd be great if we pushed that information eventually.
>
>I'd suggest something as simple as querying for all `blake2_256_digest` that exist in the DB, then all `blake2_256_digest` that exist in BigQuery and then shipping the diff. (Including removing deleted releases.)

Yep a "synchronization task" would help us update BigQuery in case the normal BigQuery uploads fail, however, its a little tricky in the case of deleted releases. I forgot to add this to the PR description but, initially the schema contained the `yanked`, `yanked_reason` and `deleted` flags which would be updated when the respective action occurs. However, it seems that BigQuery is designed to be used as a log and is not optimized for updating and removing rows, which ends up processing the whole table to manipulate one row. After some design discussions with @di, the plan was that if there's increased interest in performing retrospective analysis on data we can create another table mapping it to the current table thus manipulating a smaller subset of data. Any thoughts?

di:
> (Including removing deleted releases.)

To echo what @patelneel55 said, when we chatted w/ Felipe about this dataset, he advised that removing/updating rows sort of goes "against the grain" of how BigQuery is designed to operate, and could result in some expensive operations.

I think if we also view this as a sort of historical record, we shouldn't remove rows when the corresponding file is deleted. Adding the additional context of what has been yanked/deleted can be done in a separate table at a later date, like @patelneel55 said.

di:
(That said a sync task sounds like a good idea otherwise!)

warehouse/tasks.py
di: To add more context here, @ewdurbin is talking about [this logging handler](https://github.com/pypa/warehouse/blob/9f3fb2f7cfb2563e714fc0a452e5ad79bcc11124/warehouse/logging.py#L68-L74) which is attached to the [root logger](https://github.com/pypa/warehouse/blob/9f3fb2f7cfb2563e714fc0a452e5ad79bcc11124/warehouse/logging.py#L78).

warehouse/packaging/tasks.py
```diff
@@ -111,3 +113,54 @@ def update_description_html(request):
...
+                field_data = None
+
+            if field_data is None and sch.mode == "REPEATED":
+                row_data[sch.name] = []
+            else:
+                row_data[sch.name] = field_data
+
+    distribution_row_data = dict()
+    table_name = request.registry.settings["warehouse.distribution_table"]
+    populate_data_using_schema(bq.get_table(table_name).schema, distribution_row_data)
```
di, line 163: Do we actually need to do `bq.get_table(table_name).schema` here every time? Or can we keep some representation of this in the codebase instead?

di: In addition, I'd recommend not passing in a variable to be mutated here, making it the return value instead, and passing in the file/form as arguments. Something like:

```
json_rows = [file_and_form_to_row(file, form)]
```

warehouse/tasks.py
di: I think maybe this is already happening as-is?

warehouse/packaging/tasks.py
patelneel55: > Do we actually need to do `bq.get_table(table_name).schema` here every time? Or can we keep some representation of this in the codebase instead?

The reason for querying the schema every time is so that if the schema changes in the future or if a new table is created, we would only need to change it in BigQuery rather than also in the codebase. If query cost is an issue, I believe it is very minimal compared to the `insert_rows_json` command

di: OK, that makes sense. Comment still stands about refactoring this a bit, though.

ewdurbin:
Should do those things _before_ merge?

patelneel55:
It could be better to do that before the merge, in case anything goes wrong when streaming data before its been set. Hopefully nothing does. 🤞

PR #8125, opened 2020-06-17
Use os.path instead of posixpath

di:
This seems to cause issues with isort on some systems.

### DIFF
--- a/warehouse/accounts/services.py
+++ b/warehouse/accounts/services.py
@@ -15,7 +15,7 @@ import functools
 import hashlib
 import hmac
 import logging
-import posixpath
+import os
 import secrets
 import urllib.parse
 import uuid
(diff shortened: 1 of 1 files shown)


pradyunsg:
> This seems to cause issues with isort on some systems.

I'm curious, what kind of issues?

di:
@pradyunsg: https://github.com/timothycrosley/isort/issues/1216

I had a contributor tell me that running isort locally was attempting to move `posixpath` from the stdlib section to the 3rd-party section, likely because it's not technically in the stdlib on all platforms.

andersk:
This change is wrong on Windows systems. URLs use the POSIX separator `/` regardless of whether the host filesystem uses `/` or `\`, so the use of `posixpath` was intentional. 

`posixpath` *is* in the standard library on all platforms. It’s just only aliased to `os.path` on POSIX.

It could be replaced with another `urljoin` in this case. (Note though that the semantics would be different if not for the trailing slash.)

di:
PyPI isn't deployed on Windows. If we ever do, I'm guessing we'll have bigger things to fix first, but this will be one of them.

dstufft:
There is basically a 0% chance we ever deploy to Windows.

PR #8142, opened 2020-06-19
Public BigQuery table for distribution metadata - Bulk Load Script

patelneel55:
This PR extends from #8093. It contains the script that can be used to populate the BigQuery tables  from the existing releases on PyPI. This PR doesn't need to be merged as the script just needs to be used once for migrating metadata.

Once #8093 is merged and data starts populating on the BigQuery table, note the oldest timestamp and other config parameters into the script to make sure all the metadata in the PostgreSQL database is migrated to BigQuery. 

This script was tested on the local dev database which is similar to the TestPyPI database and the results as follows:
```
# of releases: 131,392
Total time: 34m 15s
Batch size: 1000               (how many releases were queried and uploaded at once)
Avg time / branch: 15.57s
```
Assuming that the prod PyPI is approx. 6 times TestPyPI, an estimated runtime of the migration would be around `3h 25min`. This migration can occur concurrently to other PyPI processes and doesn't block any additions or updates to releases. These runtimes occurred with the PostgreSQL server running on `i7 3Ghz` with `4GB` of container ram, so the true runtime could be faster than the estimated amount.

### DIFF
--- a/bin/update_bigquery
+++ b/bin/update_bigquery
@@ -161,21 +161,6 @@ def get_schema():
             "mode": "REPEATED"
         },
         {
-            "name": "yanked",
-            "type": "BOOLEAN",
-            "mode": "REQUIRED"
-        },
-        {
-            "name": "yanked_reason",
-            "type": "STRING",
-            "mode": "NULLABLE"
-        },
-        {
-            "name": "deleted",
-            "type": "BOOLEAN",
-            "mode": "REQUIRED"
-        },
-        {
             "name": "uploaded_via",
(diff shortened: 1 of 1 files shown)


patelneel55:
The change made in aa387b6cd8ce4154641766e206cc3deab317b9e4 under #8093 now contains an implementation of a sync task which will not only act as a fail safe in case of a failure during file upload but also be used as a migration script to move existing release files from the current database to BigQuery. Thus, we longer have a use for an independent migration script such as the one proposed in this PR.

PR #8170, opened 2020-06-26
List recent sent emails in Security history

### LINKED ISSUE #7992: list recent sent emails in security/account settings
Sometimes administrators want to send our users email notifications to ask that they do something, but we don't want to inure them to clicking on links in emails, and we want a way to credibly demonstrate that the email is legitimate.

Via <a href="https://twitter.com/runasand/status/1248385736677576706">this tweet</a> I saw that Facebook archives recent automated emails sent to users so that each user can check in their settings to find those notifications:

> To confirm whether an email claiming to be from Facebook is authentic, review recent emails we've sent in your Security and Login Settings

Could we do this?

**Additional context**
In the next 12-18 months, PyPI will probably add more features or increase the strictness of authentication, etc. for project maintainers and owners. When we do those things, we'll want to notify users in a credible way.

patelneel55:
As per the discussion in #7992, this PR creates a new security event every time an email is sent from PyPI to users which will help indicate if an email is legitimate or not. Resolves #7992 

Security History

FAQ page:
image

### DIFF
--- a/warehouse/admin/views/emails.py
+++ b/warehouse/admin/views/emails.py
@@ -99,6 +99,8 @@ def email_mass(request):
                         "body_text": row["body_text"],
                         "body_html": row.get("body_html"),
                     },
+                    user.id,
+                    request.remote_addr,
                 )
         request.session.flash("Mass emails sent", queue="success")
     else:
--- a/warehouse/admin/views/emails.py
+++ b/warehouse/admin/views/emails.py
@@ -100,7 +100,8 @@ def email_mass(request):
                         "body_html": row.get("body_html"),
                     },
                     {
-                        "sending_user_id": user.id,
+                        "tag": "account:email:sent",
+                        "user_id": user.id,
                         "ip_address": request.remote_addr,
                         "additional": {
                             "from_": request.registry.settings.get("mail.sender"),
--- a/warehouse/admin/views/emails.py
+++ b/warehouse/admin/views/emails.py
@@ -99,8 +99,16 @@ def email_mass(request):
                         "body_text": row["body_text"],
                         "body_html": row.get("body_html"),
                     },
-                    user.id,
-                    request.remote_addr,
+                    {
+                        "sending_user_id": user.id,
+                        "ip_address": request.remote_addr,
+                        "additional": {
+                            "from_": request.registry.settings.get("mail.sender"),
+                            "to": email.email,
+                            "subject": row["subject"],
+                            "redact_ip": True,
+                        },
+                    },
                 )
         request.session.flash("Mass emails sent", queue="success")
     else:
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -76,9 +76,10 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
 
     # We should only store/display IP address of an 'email sent' event if the user
     # who triggered the email event is the one who receives the email. Else display
-    # 'Redacted' to prevent user privacy concerns
+    # 'Redacted' to prevent user privacy concerns. If we don't know the user who
+    # triggered the action, default to showing the IP of the source.
     user_email = request.db.query(Email).filter(Email.email == email.email).one()
-    redact_ip = user_email.user_id != request.user.id
+    redact_ip = user_email.user_id != request.user.id if request.user else False
 
     request.task(send_email).delay(
         _compute_recipient(user, email.email),
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -35,31 +35,23 @@ def _compute_recipient(user, email):
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, recipient, msg):
+def send_email(task, request, recipient, msg, user_id, ip_address):
     msg = EmailMessage(**msg)
     sender = request.find_service(IEmailSender)
+    from_address = parseaddr(
+        sender.sender if hasattr(sender, "sender") else sender._sender
+    )[1]
 
     try:
         sender.send(recipient, msg)
 
-        # We want to only the show the IP address for the email
-        # if the email was sent to the user who triggered the email
-        user_email = (
-            request.db.query(Email).filter(Email.email == parseaddr(recipient)[1]).one()
-        )
-
-        if user_email.user_id != request.user.id:
-            ip_address = "Redacted"
-        else:
-            ip_address = request.remote_addr
-
         user_service = request.find_service(IUserService, context=None)
         user_service.record_event(
-            request.user.id,
+            user_id,
             tag="account:email:sent",
             ip_address=ip_address,
             additional={
-                "from_": parseaddr(sender._sender)[1],
+                "from_": from_address,
                 "to": parseaddr(recipient)[1],
                 "subject": msg.subject,
             },
@@ -81,8 +73,16 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
     if email is None or not (email.verified or allow_unverified):
         return
 
+    # We should only store/display IP address of an 'email sent' event if the user
+    # who triggered the email event is the one who receives the email. Else display
+    # 'Redacted' to prevent user privacy concerns
+    user_email = request.db.query(Email).filter(Email.email == email.email).one()
+    ip_address = (
+        request.remote_addr if user_email.user_id == request.user.id else "Redacted"
+    )
+
     request.task(send_email).delay(
-        _compute_recipient(user, email.email), attr.asdict(msg)
+        _compute_recipient(user, email.email), attr.asdict(msg), user.id, ip_address
     )
 
 
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -35,7 +35,9 @@ def _compute_recipient(user, email):
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, recipient, msg, user_id, ip_address):
+def send_email(
+    task, request, recipient, msg, sending_user_id, ip_address, redact_ip=True
+):
     msg = EmailMessage(**msg)
     sender = request.find_service(IEmailSender)
     from_address = parseaddr(sender.from_address)[1]
@@ -45,13 +47,14 @@ def send_email(task, request, recipient, msg, user_id, ip_address):
 
         user_service = request.find_service(IUserService, context=None)
         user_service.record_event(
-            user_id,
+            sending_user_id,
             tag="account:email:sent",
             ip_address=ip_address,
             additional={
                 "from_": from_address,
                 "to": parseaddr(recipient)[1],
                 "subject": msg.subject,
+                "redact_ip": redact_ip,
             },
         )
     except Exception as exc:
@@ -75,12 +78,14 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
     # who triggered the email event is the one who receives the email. Else display
     # 'Redacted' to prevent user privacy concerns
     user_email = request.db.query(Email).filter(Email.email == email.email).one()
-    ip_address = (
-        request.remote_addr if user_email.user_id == request.user.id else "Redacted"
-    )
+    redact_ip = user_email.user_id != request.user.id
 
     request.task(send_email).delay(
-        _compute_recipient(user, email.email), attr.asdict(msg), user.id, ip_address
+        _compute_recipient(user, email.email),
+        attr.asdict(msg),
+        user.id,
+        request.remote_addr,
+        redact_ip,
     )
 
 
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -13,6 +13,7 @@
 import functools
 
 from email.headerregistry import Address
+from email.utils import parseaddr
 
 import attr
 
@@ -21,6 +22,7 @@ from first import first
 
 from warehouse import tasks
 from warehouse.accounts.interfaces import ITokenService, IUserService
+from warehouse.accounts.models import Email
 from warehouse.email.interfaces import IEmailSender
 from warehouse.email.services import EmailMessage
 from warehouse.email.ses.tasks import cleanup as ses_cleanup
@@ -40,14 +42,25 @@ def send_email(task, request, recipient, msg):
     try:
         sender.send(recipient, msg)
 
+        # We want to only the show the IP address for the email
+        # if the email was sent to the user who triggered the email
+        user_email = (
+            request.db.query(Email).filter(Email.email == parseaddr(recipient)[1]).one()
+        )
+
+        if user_email.user_id != request.user.id:
+            ip_address = "Redacted"
+        else:
+            ip_address = request.remote_addr
+
         user_service = request.find_service(IUserService, context=None)
         user_service.record_event(
             request.user.id,
             tag="account:email:sent",
-            ip_address=request.remote_addr,
+            ip_address=ip_address,
             additional={
-                "from_": sender._sender,
-                "to": recipient,
+                "from_": parseaddr(sender._sender)[1],
+                "to": parseaddr(recipient)[1],
                 "subject": msg.subject,
             },
         )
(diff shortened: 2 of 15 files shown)


warehouse/templates/pages/help.html
```diff
@@ -270,6 +272,18 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
...
         {% endtrans %}
 
+        <h3 id="suspicious-activity">{{ suspicious_activity() }}</h3>
+        {% trans trimmed reset_pwd_href=request.route_url('accounts.request-password-reset') %}
+          <p>
+            All PyPI user events are stored under <a href="/manage/account/#account-events">Security history</a> in Account settings.
+            If there are any events that seem suspicious, take the following steps:
+            <ul>
+              <li><a href="{{ reset_pwd_href }}">Reset your password</a></li>
+              <li>Contact the PyPI admins about the event at <a href="mailto:admin@python.org">admin at python dot org</a></li>
```
di, line 282:
```suggestion
              <li>Contact the PyPI admins about the event at <a href="mailto:admin@python.org">admin@python.org</a></li>
```

warehouse/templates/manage/account.html
```diff
@@ -662,7 +669,7 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
       {% endif %}
     {%- endmacro %}
 
-    <p>{% trans %}Events appear here as security-related actions occur on your account.{% endtrans %}</p>
+    <p>{% trans %}Events appear here as security-related actions occur on your account. If you notice anything suspicious, please <a href="/help#suspicious-activity">secure your account</a> as soon as possible.{% endtrans %}</p>
```
di, line 672: We should use `request.help_url` to generate the URL here. Also see https://github.com/pypa/warehouse/issues/8040#issuecomment-652076266 for an example of passing this into the translated string.

warehouse/email/__init__.py
```diff
@@ -56,8 +73,16 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
     if email is None or not (email.verified or allow_unverified):
         return
 
+    # We should only store/display IP address of an 'email sent' event if the user
+    # who triggered the email event is the one who receives the email. Else display
+    # 'Redacted' to prevent user privacy concerns
+    user_email = request.db.query(Email).filter(Email.email == email.email).one()
+    ip_address = (
+        request.remote_addr if user_email.user_id == request.user.id else "Redacted"
+    )
```
di, line 82: Why not:

```suggestion
    ip_address = (
        request.remote_addr if email == request.user.primary_email else "Redacted"
    )
```
?

warehouse/email/__init__.py
```diff
@@ -33,12 +35,27 @@ def _compute_recipient(user, email):
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, recipient, msg):
+def send_email(task, request, recipient, msg, user_id, ip_address):
     msg = EmailMessage(**msg)
     sender = request.find_service(IEmailSender)
+    from_address = parseaddr(
+        sender.sender if hasattr(sender, "sender") else sender._sender
+    )[1]
```
di, line 43: I'm not sure what reason for the difference in the class signature for `SMTPEmailSender` and `SESEmailSender` are, but I think we should probably unify this as something each service provides, so we can just do something like:

```python
from_address = sender.from_address
```
(or just put that inline below)

warehouse/templates/manage/account.html
```diff
@@ -600,6 +600,13 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
           {{ event.additional.new_primary }}
         </small>
         {% endif %}
+      {% elif event.tag == "account:email:sent" %}
+      <strong>{% trans %}Email sent{% endtrans %}</strong><br>
+      <small>
+        {% trans %}From:{% endtrans %} {{ event.additional.from_ }}<br>
+        {% trans %}To:{% endtrans %} {{ event.additional.to }}<br>
+        {% trans %}Subject:{% endtrans %} {{event.additional.subject}}
```
di, line 608: I think we should probably format these emails like we do in the Account Emails table:

Screen Shot 2020-06-30 at 5 54 53 PM

On the fence about whether we should do it for the subject too, but I'm thinking yes.

warehouse/email/__init__.py
patelneel55: The user_email was queried in the off-chance that an email is sent to a non-primary email, such as when sending a `verify-email` email.

warehouse/templates/manage/account.html
patelneel55: image

warehouse/templates/pages/help.html
```diff
@@ -270,6 +272,18 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
           </p>
         {% endtrans %}
 
+        <h3 id="suspicious-activity">{{ suspicious_activity() }}</h3>
+        {% trans trimmed reset_pwd_href=request.route_url('accounts.request-password-reset'), admin_email='mailto:admin@python.org'%}
```
di, line 276:
```suggestion
        {% trans trimmed reset_pwd_href=request.route_url('accounts.request-password-reset'), admin_email='admin@python.org'%}
```

warehouse/templates/pages/help.html
```diff
@@ -270,6 +272,18 @@ <h3 id="compromised-password">{{ compromised_password() }}</h3>
...
+          <p>
+            All PyPI user events are stored under security history in account settings.
+            If there are any events that seem suspicious, take the following steps:
+            <ul>
+              <li><a href="{{ reset_pwd_href }}">Reset your password</a></li>
+              <li>Contact the PyPI admins about the event at <a href="{{ admin_email }}">admin@python.org</a></li>
```
di, line 282:
```suggestion
              <li>Contact the PyPI admins about the event at <a href="mailto:{{ admin_email }}">{{ admin_email }}</a></li>
```

warehouse/email/__init__.py
```diff
@@ -33,12 +35,25 @@ def _compute_recipient(user, email):
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, recipient, msg):
+def send_email(task, request, recipient, msg, user_id, ip_address):
```
di, line 38: The `send_email` task is also used in the admin: https://github.com/pypa/warehouse/blob/de634ab4f46adf34cb5bb1af701f1f807df01343/warehouse/admin/views/emails.py#L95-L102

In that use case, the IP address should always be redacted, so it might make more sense to pass `redact_ip` in instead.

Also we should be more clear what `user_id` is here. Something like:

```suggestion
def send_email(task, request, recipient, msg, sending_user_id, redact_ip):
```

warehouse/email/__init__.py
```diff
@@ -33,12 +35,25 @@ def _compute_recipient(user, email):
...
+    from_address = parseaddr(sender.from_address)[1]
 
     try:
         sender.send(recipient, msg)
+
+        user_service = request.find_service(IUserService, context=None)
+        user_service.record_event(
+            user_id,
+            tag="account:email:sent",
+            ip_address=ip_address,
```
di, line 50:
```suggestion
            ip_address=request.remote_addr,
```

warehouse/email/__init__.py
```diff
@@ -56,8 +71,16 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
     if email is None or not (email.verified or allow_unverified):
         return
 
+    # We should only store/display IP address of an 'email sent' event if the user
+    # who triggered the email event is the one who receives the email. Else display
+    # 'Redacted' to prevent user privacy concerns
+    user_email = request.db.query(Email).filter(Email.email == email.email).one()
+    ip_address = (
+        request.remote_addr if user_email.user_id == request.user.id else "Redacted"
+    )
```
di, line 80: I think instead of the `ip_address` literally becoming `"Redacted"` here, we should set something like `additional.redact_ip` and just change how we display it in the UI with something like:

```
{{ "Redacted" if event.additional.redact_ip else event.ip_address }}
```
So this would become something like:

```suggestion
    user_email = request.db.query(Email).filter(Email.email == email.email).one()
    redact_ip = user_email.user_id != request.user.id
```

patelneel55: My thought process here was that if we store the IP in the database and redact it in the UI, then there is a possibility that a user can manipulate the client and still extract the IP. Hence, rather not have that data to begin with. However, I'm not sure how feasible extracting the data would be from the client and if we need to worry about that.

di: I think we still want the data available for administrators. I don't think it would be possible to extract the IP.

warehouse/email/__init__.py
patelneel55: It seems that the `request` object in the `send_email` task is missing many of the attributes like `remote_addr` and `db`. Any ideas on why that could be?

di: Are they getting lost on serialization? I.e. do they exist before it's passed to `delay`?

patelneel55: Yes, it seems that the `WarehouseTask` class is overwriting the `request` that is passed into the task

patelneel55: I have updated the function so that we now pass the `remote_addr` to the task along with a `redact_ip` flag for the UI mark as "Redacted"

warehouse/email/__init__.py
```diff
@@ -56,8 +74,18 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
     if email is None or not (email.verified or allow_unverified):
         return
 
+    # We should only store/display IP address of an 'email sent' event if the user
+    # who triggered the email event is the one who receives the email. Else display
+    # 'Redacted' to prevent user privacy concerns
+    user_email = request.db.query(Email).filter(Email.email == email.email).one()
+    redact_ip = user_email.user_id != request.user.id
```
di, line 81: We need to handle the case where `request.user` is `None` (for example, when requesting a password reset)

patelneel55: Hmm, I printed `request.user` in the console and it seems to have a valid value during password reset and it also displays the appropriate event, are there any other scenarios that `request.user` can be `None`? If so, should we marked it as redacted since we can't verify the source?

image

di: Here's the full stacktrace I get with this branch when requesting a password reset:

```
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid_debugtoolbar/panels/performance.py", line 58, in resource_timer_handler
    result = handler(request)
  File "/opt/warehouse/src/warehouse/raven.py", line 40, in raven_tween
    return handler(request)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid_tm/__init__.py", line 178, in tm_tween
    reraise(*exc_info)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid_tm/compat.py", line 36, in reraise
    raise value
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid_tm/__init__.py", line 143, in tm_tween
    response = handler(request)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/tweens.py", line 43, in excview_tween
    response = _error_handler(request, exc)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/tweens.py", line 17, in _error_handler
    reraise(*exc_info)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/compat.py", line 179, in reraise
    raise value
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/tweens.py", line 41, in excview_tween
    response = handler(request)
  File "/opt/warehouse/src/warehouse/cache/http.py", line 74, in conditional_http_tween
    response = handler(request)
  File "/opt/warehouse/src/warehouse/sanity.py", line 76, in sanity_tween_egress
    return unicode_redirects(handler(request))
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/router.py", line 147, in handle_request
    response = _call_view(
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/view.py", line 667, in _call_view
    response = view_callable(context, request)
  File "/opt/warehouse/src/warehouse/cache/http.py", line 33, in wrapped
    return view(context, request)
  File "/opt/warehouse/src/warehouse/cache/http.py", line 33, in wrapped
    return view(context, request)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/viewderivers.py", line 514, in csrf_view
    return view(context, request)
  File "/opt/warehouse/src/warehouse/metrics/views.py", line 34, in wrapper_view
    return view(context, request)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/viewderivers.py", line 436, in rendered_view
    result = view(context, request)
  File "/opt/warehouse/lib/python3.8/site-packages/pyramid/viewderivers.py", line 144, in _requestonly_view
    response = view(request)
  File "/opt/warehouse/src/warehouse/accounts/views.py", line 541, in request_password_reset
    send_password_reset_email(request, (user, email))
  File "/opt/warehouse/src/warehouse/email/__init__.py", line 141, in wrapper
    _send_email_to_user(
  File "/opt/warehouse/src/warehouse/email/__init__.py", line 81, in _send_email_to_user
    redact_ip = user_email.user_id != request.user.id
AttributeError: 'NoneType' object has no attribute 'id'
```

The `request.user` attribute is `None` whenever there isn't a logged-in user.

patelneel55: Ah I see, `request.user` is `None` when you reset password during login. I assumed when resetting under `Account settings`. Good catch! I will fix that.

warehouse/email/__init__.py
```diff
@@ -33,12 +35,28 @@ def _compute_recipient(user, email):
 
 
 @tasks.task(bind=True, ignore_result=True, acks_late=True)
-def send_email(task, request, recipient, msg):
+def send_email(
+    task, request, recipient, msg, sending_user_id, ip_address, redact_ip=True
+):
```
di, line 40: I think I'd prefer:

```suggestion
def send_email(task, request, recipient, msg, success_event):
```

and below we just do:

```python
user_service = request.find_service(IUserService, context=None)
user_service.record_event(success_event)
```

patelneel55: I was able to reduce the number of parameters we pass into `send_email` by passing a dict but I can't call the `record_event` function exactly as you're expecting because some of the parameters for `record_event` such as the `from_address` come from inside the `send_email` task. Does that work?

di: I think we can determine `from_address` in the view as well.

warehouse/email/__init__.py
```diff
@@ -33,12 +34,20 @@ def _compute_recipient(user, email):
...
     try:
         sender.send(recipient, msg)
+
+        user_service = request.find_service(IUserService, context=None)
+        user_service.record_event(
+            success_event["sending_user_id"],
+            tag="account:email:sent",
+            ip_address=success_event["ip_address"],
+            additional=success_event["additional"],
+        )
```
di, line 50:
```suggestion
        user_service.record_event(**success_event)
```

warehouse/email/__init__.py
```diff
@@ -56,8 +65,26 @@ def _send_email_to_user(request, user, msg, *, email=None, allow_unverified=Fals
...
+    # triggered the action, default to showing the IP of the source.
+    user_email = request.db.query(Email).filter(Email.email == email.email).one()
+    redact_ip = user_email.user_id != request.user.id if request.user else False
+
     request.task(send_email).delay(
-        _compute_recipient(user, email.email), attr.asdict(msg)
+        _compute_recipient(user, email.email),
+        attr.asdict(msg),
+        {
+            "sending_user_id": user.id,
```
di, line 79:
```suggestion
            "user_id": user.id,
            "tag": "account:email:sent",
```

warehouse/admin/views/emails.py
```diff
@@ -99,6 +99,16 @@ def email_mass(request):
                         "body_text": row["body_text"],
                         "body_html": row.get("body_html"),
                     },
+                    {
+                        "sending_user_id": user.id,
```
di, line 103:
```suggestion
            "user_id": user.id,
            "tag": "account:email:sent",
```

PR #8181, opened 2020-06-30
Re-Authentication for Sensitive Actions/Endpoints

callmecampos:
Implements https://github.com/pypa/warehouse/issues/8139.

Whenever a "sensitive" endpoint is hit, we check to see how long it has been since the user last authenticated. We do this by recording the timestamp of each authentication for a given session in Redis. If it has been less than specified period of time (1 hour by default), we perform the action normally, otherwise we first display a password confirmation page:

Screen Shot 2020-06-30 at 12 36 20 AM

Once the user re-authenticates, the timestamp of is recorded in the session (and subsequently in Redis), they are redirected to their original destination, and the user won't be asked to re-authenticate again for another hour for such sensitive endpoints.

Below is a draft checklist of endpoints that should require re-authentication:

- [x] /manage/account/
  - [x] adding emails
  - [x] deleting emails
  - [x] changing primary email
  - [ ] deleting account (should reauthenticate every time, modal view already exists doing this)
- [x] /manage/project/[project_name]/releases/
  - [x] adding collaborators
  - [x] removing collaborators
  - [x] changing collaborator roles
  - [x] yanking releases
  - [x] unyanking releases
  - [x] deleting releases
  - [x] deleting files
- [x] /manage/project/[project_name]/collaborators/
- [x] /manage/project/[project_name]/settings/
  - [x] generating API tokens
  - [x] deleting API tokens
  - [x] deleting projects

### DIFF
--- a/warehouse/sessions.py
+++ b/warehouse/sessions.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import functools
 import time
 
@@ -82,6 +83,9 @@ def _changed_method(method):
 
 @implementer(ISession)
 class Session(dict):
+    reauth_timestamp_cookie_name = "session_reauth_timestamp"
+    reauth_max_age = 30  # 60 * 60 # 1 hour
+    max_age = 12 * 60 * 60  # 12 hours
 
     _csrf_token_key = "_csrf_token"
     _flash_key = "_flash_messages"
@@ -139,6 +143,34 @@ class Session(dict):
 
     def should_save(self):
         return self._changed
+
+    def record_auth_timestamp(self, request, response):
+        auth_signer = crypto.Signer(
+            request.registry.settings["sessions.secret"], salt="session"
+        )
+
+        response.set_cookie(
+            self.reauth_timestamp_cookie_name,
+            auth_signer.sign(str(datetime.datetime.now().timestamp())),
+            max_age=12 * 60 * 60,
+            httponly=True,
+        )
+
+    def needs_reauthentication(self, request):
+        reauth_signer = crypto.Signer(
+            request.registry.settings["sessions.secret"], salt="session"
+        )
+
+        try:
+            auth_time = float(
+                reauth_signer.unsign(
+                    request.cookies.get(self.reauth_timestamp_cookie_name)
+                )
+            )
+            current_time = datetime.datetime.now().timestamp()
+            return current_time - auth_time >= self.reauth_max_age
+        except (crypto.BadSignature, ValueError):
+            return True
 
     # Flash Messages Methods
     def _get_flash_queue_key(self, queue):
--- a/warehouse/sessions.py
+++ b/warehouse/sessions.py
@@ -83,7 +83,7 @@ def _changed_method(method):
 
 @implementer(ISession)
 class Session(dict):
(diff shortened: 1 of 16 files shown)


warehouse/sessions.py
```diff
@@ -140,6 +145,35 @@ def invalidate(self):
     def should_save(self):
         return self._changed
 
+    def record_auth_timestamp(self, request, response):
+        auth_signer = crypto.TimestampSigner(
+            request.registry.settings["sessions.secret"], salt="session"
+        )
+
+        response.set_cookie(
```
dstufft, line 153: Why not just store a datetime inside the session itself? Currently this seems like it's defeatable because there's nothing tying this timestamp cookie to the session, so if I come across a computer that is signed in, but outside of this timestamp window, I just fire up a second browser, login with any account, get the cookie, and then inject that cookie into the logged in browser.

Is having it be a cookie buying us something? Otherwise it seems simpler and more secure to just throw it in the session.

callmecampos: By storing inside the session itself, I'm assuming you mean storing the datetime in Redis? So adding a field in the cookie dictionary like `request.session["last_auth_timestamp"] = str(datetime.datetime.now().timestamp())` which would then be serialized with the rest of the cookie fields (e.g. `_csrf_token`) and then unpacked with every new request?

With regards to the signature, I tried what you mentioned and you're right, with the same secret and salt used in the signature, you would be able to exploit this by copying over a cookie from a separate browser with a different account more recently logged in. However, changing the salt to the session ID instead of just "session" (as it is currently) would prevent this sort of attack since the signatures would then be unique to each session regardless of timestamp. I just tested adding this locally and I'm no longer able to successfully perform the exploit you mentioned.

Either way would be an easy change, let me know which you think is best.

callmecampos: To your question of whether the cookie gets us anything, I don't think it does. I'm going ahead and adding the timestamp as a key/value pair in the session.

warehouse/sessions.py
```diff
@@ -140,6 +145,22 @@ def invalidate(self):
...
+        reauth_timestamp = self.get(self._reauth_timestamp_key)
+        if reauth_timestamp is None:
+            return True
+
+        try:
+            auth_time = float(reauth_timestamp)
+            current_time = datetime.datetime.now().timestamp()
+            return current_time - auth_time >= self.time_to_reauth
+        except (ValueError, TypeError):
+            return True
```
di, line 162: When would this happen? Do we actually need to handle these exceptions or can we just guard against them?

warehouse/sessions.py
```diff
@@ -82,11 +83,13 @@ def wrapped(self, *args, **kwargs):
 
 @implementer(ISession)
 class Session(dict):
+    time_to_reauth = 30  # 60 * 60  # 1 hour
```
di, line 86:
```suggestion
    time_to_reauth = 60 * 60  # 1 hour
```

warehouse/sessions.py
callmecampos: `TypeError` doesn't need to be there since I already handled the case where the timestamp is `None`. A `ValueError` could be raised if the timestamp can't be cast to a float for some reason, though I can't see how that field would ever get corrupted. So yeah, it should be fine to just guard against the `None` case, which I already do.

PR #8182, opened 2020-06-30
Invite collaborators to project

### LINKED ISSUE #5790: Send invitations when adding owner/maintainer roles

patelneel55:
Resolves #5790 and closes #6054. Wraps up work done by @DavidBord, @di and @ewdurbin


warehouse/accounts/views.py
```diff
@@ -735,6 +744,104 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+    except TokenExpired:
+        return _error(request._("Expired token: request a new project role invite"))
+    except TokenInvalid:
+        return _error(request._("Invalid token: request a new project role invite"))
+    except TokenMissing:
+        return _error(request._("Invalid token: no token supplied"))
+
+    # Check whether this token is being used correctly
+    if data.get("action") != "email-project-role-verify":
+        return _error(request._("Invalid token: not an email verification token"))
```
di, line 773:
```suggestion
        return _error(request._("Invalid token: not a collaboration invitation token"))
```

warehouse/accounts/views.py
```diff
@@ -735,6 +744,104 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+
+    user = user_service.get_user(data.get("user_id"))
+    if user != request.user:
+        return _error(request._("Invite is not valid"))
+
+    project = (
+        request.db.query(Project).filter(Project.id == data.get("project_id")).one()
+    )
```
di, line 781: If we mounted this view at `/manage/project/{project_name}/...` instead, we could use URL traversal to get the current project, rather than having to do it manually in the view here.

We'd need to change how we generated the URL for the email slightly as well, but that shouldn't be an issue.

warehouse/accounts/views.py
```diff
@@ -735,6 +744,104 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+        request.db.query(RoleInvitation)
+        .filter(RoleInvitation.project == project)
+        .filter(RoleInvitation.user == user)
+        .one()
+    )
+
+    if role_invite.invite_status != RoleInvitationStatus.Pending.value:
+        return _error(
+            request._("Role invitation is already assigned or has been revoked.")
+        )
```
di, line 794: Maybe we can provide some more specific errors here based on what `invite_status` is instead of what it is not, as well as a generic fallback error message if it falls through all the conditionals?

warehouse/accounts/views.py
```diff
@@ -735,6 +744,104 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+    role_invite.invite_status = RoleInvitationStatus.Accepted.value
+    request.db.add(Role(user=user, project=project, role_name=desired_role))
+    request.db.add(
+        JournalEntry(
+            name=project.name,
+            action=f"accepted {desired_role} {user.username}",
+            submitted_by=request.user,
+            submitted_from=request.remote_addr,
+        )
+    )
```
di, line 805: Probably should start doing a `record_event` here as well, for the user and for the project.

warehouse/legacy/api/xmlrpc/views.py
```diff
@@ -186,7 +186,6 @@ def mapply(self, fn, args, kwargs):
             memo = typeguard._CallMemo(fn, args=args, kwargs=kwargs)
             typeguard.check_argument_types(memo)
         except TypeError as exc:
-            print(exc)
```
di, line 189: 👍

warehouse/manage/views.py
```diff
@@ -1439,65 +1444,154 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
+            if user_invite:
+                if (
+                    user_invite.invite_status == RoleInvitationStatus.Pending.value
+                    or user_invite.invite_status == RoleInvitationStatus.Accepted.value
+                ):
+                    request.session.flash(
+                        f"User '{username}' already has an active invite/role. "
+                        f"Please try again later.",
+                        queue="error",
+                    )
```
di, line 1474: Again, let's split this apart so we can provide a more accurate error message.

warehouse/manage/views.py
```diff
@@ -1439,65 +1444,154 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
+            else:
+                send_email = True
+                request.db.add(
+                    RoleInvitation(
+                        user=user,
+                        project=project,
+                        invite_status=RoleInvitationStatus.Pending.value,
+                        token=token,
+                    )
                 )
```
di, line 1488: Can we refactor this so the `send_email` boolean isn't necessary, and so that the conditionals are only one deep? We can always return early if we need to.

warehouse/templates/email/verify-project-role/body.html
```diff
@@ -0,0 +1,24 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+{% block content %}
+<p>{{ initiator_username }} has invited you (<code>{{ email_address }}</code>) to join their PyPI project "{{ project_name }}".</p>
```
di, line 17: I think we could make `initiator_username` a link to the user's public profile here.

warehouse/templates/email/verify-project-role/body.html
```diff
@@ -0,0 +1,24 @@
...
+
+<p>If you wish to accept this invitation, <a href="{{ request.route_url('accounts.verify-project-role', _query={'token': token}) }}">click this link</a></p>
```
di, line 19:
```suggestion
<p>If you wish to proceed, <a href="{{ request.route_url('accounts.verify-project-role', _query={'token': token}) }}">click this link to accept this invitation</a></p>
```

warehouse/accounts/views.py
di: Can't do this because it would require modifying the project ACL, so marking this as resolved.

di:
@patelneel55 Looks like there's still unresolved comments here, please re-request a review when this is ready.

warehouse/accounts/views.py
```diff
@@ -735,6 +744,122 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+        role=desired_role,
+    )
+
+    request.session.flash(
+        request._(
+            "Added collaborator '${username}'", mapping={"username": user.username},
+        ),
+        queue="success",
+    )
+    return HTTPSeeOther(request.route_path("manage.projects"))
```
di, line 860: I think we should redirect to the management page for the project they just got access to instead.

warehouse/accounts/views.py
```diff
@@ -735,6 +744,122 @@ def _get_two_factor_data(request, _redirect_to="/"):
...
+        request,
+        user,
+        submitter=submitter_user,
+        project_name=project.name,
+        role=desired_role,
+    )
+
+    request.session.flash(
+        request._(
+            "Added collaborator '${username}'", mapping={"username": user.username},
```
di, line 856: This is the message displayed to the user after accepting an invitation, might be better for this to say something like:

> You are now Owner/Maintainer of the 'project_name' project

warehouse/manage/views.py
```diff
@@ -1443,66 +1455,149 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
-            request.db.add(
-                Role(user=user, project=project, role_name=form.role_name.data)
+        elif (
+            user_invite
+            and user_invite.invite_status == RoleInvitationStatus.Pending.value
+        ):
+            request.session.flash(
+                f"User '{username}' already has an active invite. "
+                f"Please try again later.",
+                queue="error",
```
di, line 1465: There's an edge case and a 5 minute window here where the invite might be expired, but it's still "Pending" in the database. I think in the case that the user already has an invite, we need to check whether it's expired as well, and manually handle the status change.

warehouse/manage/views.py
```diff
@@ -1443,66 +1455,149 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
             )
+        elif (
+            user_invite
+            and user_invite.invite_status == RoleInvitationStatus.Accepted.value
+        ):
+            request.session.flash(
+                f"User '{username}' is already a collaborator on the project.",
+                queue="error",
+            )
```
di, line 1474: What happens here if we're just changing the user's role on the project (e.g. from Maintainer to Owner), so there's already one accepted invitation, but we need to generate another with a different role?

warehouse/migrations/versions/80018e46c5a4_create_role_invitation_table.py
```diff
@@ -0,0 +1,59 @@
...
+        sa.ForeignKeyConstraint(
+            ["project_id"], ["projects.id"], onupdate="CASCADE", ondelete="CASCADE"
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"], ["users.id"], onupdate="CASCADE", ondelete="CASCADE"
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "user_id", "project_id", name="_role_invitations_user_project_uc"
+        ),
```
di, line 50: This seems to imply that we can't generate multiple invitations for a given user, e.g. w/ a role change, or to invite a user, accept the invitation, revoke the role, re-invite, etc.

warehouse/manage/views.py
```diff
@@ -1443,66 +1455,149 @@ def manage_project_roles(project, request, _form_class=CreateRoleForm):
...
-            request.session.flash(
-                f"Added collaborator '{form.username.data}'", queue="success"
+    try:
+        user = user_service.get_user(request.POST["user_id"])
+        user_invite = (
+            request.db.query(RoleInvitation)
+            .filter(RoleInvitation.project == project)
+            .filter(RoleInvitation.user == user)
+            .one()
+        )
```
di, line 1565: I think we need to use the `RoleInvitation` ID here instead.

warehouse/manage/views.py
```diff
@@ -1578,6 +1673,13 @@ def delete_project_role(project, request):
             request.session.flash("Cannot remove yourself as Owner", queue="error")
         else:
             request.db.delete(role)
+            role_invite = (
+                request.db.query(RoleInvitation)
+                .filter(RoleInvitation.project == project)
+                .filter(RoleInvitation.user == role.user)
+                .one()
+            )
```
di, line 1681: Same thing here, use the ID instead.

warehouse/accounts/views.py
patelneel55: Since only project owners can access the project management page, we can redirect to the project page instead?

di: Ah, right. Maybe this should be dependent on what the role was then.

warehouse/migrations/versions/80018e46c5a4_create_role_invitation_table.py
patelneel55: Since we decided to leave a minimal footprint in the role invitation table, this constraint still applies thus marking this as resolved.

warehouse/manage/views.py
patelneel55: As discussed, we decided to leave a minimal footprint in the role invitation table, thus marking this as resolved.

warehouse/manage/views.py
patelneel55: As discussed, we decided to leave a minimal footprint in the role invitation table, thus marking this as resolved.

warehouse/packaging/models.py
```diff
@@ -77,6 +77,38 @@ class Role(db.Model):
...
+    __table_args__ = (
+        Index("role_invitations_user_id_idx", "user_id"),
+        UniqueConstraint(
+            "user_id", "project_id", name="_role_invitations_user_project_uc"
+        ),
+    )
+
+    __repr__ = make_repr("invite_status", "user", "project")
+
+    invite_status = Column(Text, nullable=False)
```
di, line 98: This should probably be an `Enum` instead, e.g: https://github.com/pypa/warehouse/blob/91a1767337b5e3a0ef01817208aba2b6a4c8f554/warehouse/packaging/models.py#L486-L497

warehouse/templates/email/verify-project-role/body.txt
```diff
@@ -0,0 +1,26 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.txt" %}
+
+{% block content %}
+{{ initiator_username }} has invited you ({{ email_address }}) to their PyPI project "{{ project_name }}".
```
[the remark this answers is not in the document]
patelneel55: Is it possible to add hyperlinks in `body.txt`? I'm not really sure how to do this and I couldn't find any other email txt files doing this either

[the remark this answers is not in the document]
di: They can just be added as plaintext:

```suggestion
{{ initiator_username }} has invited you ({{ email_address }}) to their PyPI project "{{ project_name }}" ({{ request.route_path('packaging.project', name=project_name) }})
```

patelneel55:
@di I made the requested changes and also, like you suggested, made the project page display any pending invitations. Accepting it will take you to the accepted project page/project manage page based on your target role and declining it will remove the invite and keep you at the project list page.

image

warehouse/templates/accounts/invite-confirmation.html
```diff
@@ -0,0 +1,37 @@
...
+    <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+    <span class="form-group__text">
+      {% trans project_name=project_name, role_name=desired_role, project_link=request.route_url('packaging.project', name=project_name) %}
+      Would you like to accept this invitation to join '<a href="{{ project_link }}">{{ project_name }}</a>' as a project {{ role_name }}?
+      {% endtrans %}
+    </span>
+  </div>
+  <div>
+    <input name="accept" type="submit" value="{% trans %}Accept{% endtrans %}" class="button button--primary">
+    <input name="decline" type="submit" value="{% trans %}Decline{% endtrans %}" class="button button--danger">
```
di, line 33:
```suggestion
    <input name="decline" type="submit" value="{% trans %}Decline{% endtrans %}" class="button">
```
Very small nit here but this button is not actually "dangerous". We want to be careful how much we use the "danger" color.

warehouse/templates/manage/projects.html
```diff
@@ -18,6 +18,43 @@
...
+            {% trans creation_date=humanize(project.created) %}Created on {{ creation_date }}{% endtrans %}
+          </p>
+          {% endif %}
+        </div>
+        <div class="package-snippet__buttons">
+          <form method="POST" action="{{ request.route_path('accounts.verify-project-role') }}">
+            <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+            <input name="token" type="hidden" value="{{ token }}">
+            <input name="accept" type="submit" value="{% trans %}Accept{% endtrans %}" class="button button--primary">
+            <input name="decline" type="submit" value="{% trans %}Decline{% endtrans %}" class="button button--danger">
```
di, line 50:
```suggestion
            <input name="decline" type="submit" value="{% trans %}Decline{% endtrans %}" class="button">
```

warehouse/templates/accounts/invite-confirmation.html
patelneel55: I was attracted to the red 😜

pradyunsg:
\o/

PR #8183, opened 2020-06-30
Document marking strings and views for translation

callmecampos:
Fixes https://github.com/pypa/warehouse/issues/8040.

### DIFF
--- a/docs/translations.rst
+++ b/docs/translations.rst
@@ -34,11 +34,25 @@ To add a new known locale:
 .. |warehouse/i18n/__init__.py| replace:: ``warehouse/i18n/__init__.py``
 .. _warehouse/i18n/__init__.py: https://github.com/pypa/warehouse/blob/master/warehouse/i18n/__init__.py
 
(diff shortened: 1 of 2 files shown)


docs/translations.rst
```diff
@@ -33,3 +33,45 @@ To add a new known locale:
...
+      This link will expire in {{ n_hours }} hours.
+      {% endtrans %}
+
+In Python, given a request context, call :code:`request._(message)` to mark
+:code:`message` for translation. Without a request context, you can do the following:
+
+.. code-block:: python
+
+   from warehouse.i18n import localize as _
+   message = _("Your message here.")
```
di, line 60: Can we split this into two sections? One "Marking strings for translation" (with HTML/Python examples) and the second "Marking pluralized strings for translation" (also with HTML/Python)?

di:
Could you add another section, "Using non-traslatable values in translated strings" based on https://github.com/pypa/warehouse/issues/8040#issuecomment-652076266? Sorry this was missing from the original issues.

PR #8189, opened 2020-07-01
detect assign and import aliasing, HOF use, and chr/ord

lcol3117:
This PR contains a few changes to the yara rules to detect many previously undetected bypass mechanisms.

see lcol3117/better-python-malware-rules for the development of this change

The changes include detection of the following, previously undetected, almost certainly malicious code:

- Attempting to alias a suspicious module by `import ? as ?`, and then using suspicious methods
- Doing the same with `? = ?` instead of `import ? as ?`
- Currently, for calls such as `eval`, the check can be bypassed by passing `eval` to a HOF (higher order function). This fixes that, but things such as the word `evaluate` are not detected.
- the functions `chr` and `ord` can be used for caesar-cipher-style obfuscation, and are almost never used for anything else. These will now detected as (indeterminate) potential malware. 

These, and only these, have been fixed in this PR. 

Hopefully this can make PyPI safer. If there is anything else I can do please tell me. Thanks for keeping python safe, working, and powerful.


xmunoz:
Take a look at the failing unit tests @lcol3117

lcol3117:
Sorry, I had forgotten to increment the version number per the docs.

xmunoz:
Great, I should be able to get to a review in the coming days. Thanks for your contribution!

lcol3117:
No problem, thanks for this!

warehouse/malware/checks/setup_patterns/setup_py_rules.yara
```diff
@@ -10,7 +10,9 @@ rule process_spawn_in_setup {
         classification = "threat"
 
     strings:
-        // NOTE(ww): We don't detect `import os as ...`
+        // EDITED to add alias import detection
```
xmunoz, line 13:
```suggestion
        // Alias import detection
```

lcol3117:
Will do, thanks so much. Completely agree, sorry I left the comments so messy. Should have this done in a couple of days at most. Thanks again.

di:
@lcol3117 FYI, you can run `make reformat` to auto-format any linting issues.

lcol3117:
Thanks, sorry I didn’t know that. Once that is done all the tests were successful so the PR should be good to go for review again.

warehouse/malware/checks/setup_patterns/setup_py_rules.yara
lcol3117: These changes have been done, thanks.

lcol3117:
Sorry about that. The comments are cleaned up, and I have introduced quite a few (113 to be precise) new tests for the new malware rules, and both the lint, static pipeline, and unit test issues have been resolved. Thanks for helping out @xmunoz and @di . The only other change since the review is that passing a suspicious function to a higher order function embedded in a dict will know be detected (it can be passed as either the key or the value). Thanks again for helping out.

di:
Thanks!

PR #8191, opened 2020-07-01
Re-Authentication for Sensitive Actions/Endpoints

callmecampos:
Implements https://github.com/pypa/warehouse/issues/8139. (Re-creating this PR from #8181 since Travis wasn't able to recognize the last one for some reason)

Whenever a "sensitive" endpoint is hit, we check to see how long it has been since the user last authenticated. We do this by recording the timestamp of each authentication for a given session in Redis. If it has been less than specified period of time (1 hour by default), we perform the action normally, otherwise we first display a password confirmation page:

Screen Shot 2020-06-30 at 12 36 20 AM

Once the user re-authenticates, the timestamp of is recorded in the session (and subsequently in Redis), they are redirected to their original destination, and the user won't be asked to re-authenticate again for another hour for such sensitive endpoints.

Below is a draft checklist of endpoints that should require re-authentication:

- [x] /manage/account/
  - [x] adding emails
  - [x] deleting emails
  - [x] changing primary email
  - [x] (Separate PR) deleting account (should reauthenticate every time, modal view already exists doing this)
- [x] /manage/project/[project_name]/releases/
  - [x] adding collaborators
  - [x] removing collaborators
  - [x] changing collaborator roles
  - [x] yanking releases
  - [x] unyanking releases
  - [x] deleting releases
  - [x] deleting files
- [x] /manage/project/[project_name]/collaborators/
- [x] /manage/project/[project_name]/settings/
  - [x] generating API tokens
  - [x] deleting API tokens
  - [x] deleting projects

### DIFF
--- a/warehouse/routes.py
+++ b/warehouse/routes.py
@@ -140,6 +140,9 @@ def includeme(config):
         domain=warehouse,
     )
     config.add_route(
+        "accounts.reauthenticate", "/accounts/reauthenticate/", domain=warehouse
+    )
+    config.add_route(
         "accounts.recovery-code", "/account/recovery-code/", domain=warehouse,
     )
     config.add_route("accounts.logout", "/account/logout/", domain=warehouse)
@@ -295,9 +298,6 @@ def includeme(config):
         domain=warehouse,
     )
     config.add_route("packaging.file", files_url)
-
-    # Re-Authentication Route
-    config.add_route("reauthenticate", "/reauthenticate/", domain=warehouse)
 
     # SES Webhooks
     config.add_route("ses.hook", "/_/ses-hook/", domain=warehouse)
--- a/warehouse/routes.py
+++ b/warehouse/routes.py
@@ -296,6 +296,9 @@ def includeme(config):
     )
     config.add_route("packaging.file", files_url)
 
+    # Re-Authentication Route
+    config.add_route("reauthenticate", "/reauthenticate/", domain=warehouse)
+
     # SES Webhooks
     config.add_route("ses.hook", "/_/ses-hook/", domain=warehouse)
 
--- a/warehouse/sessions.py
+++ b/warehouse/sessions.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import functools
 import time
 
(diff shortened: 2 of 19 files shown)


warehouse/sessions.py
```diff
@@ -98,7 +101,7 @@ class Session(dict):
     setdefault = _changed_method(dict.setdefault)
     update = _changed_method(dict.update)
 
-    def __init__(self, data=None, session_id=None, new=True):
+    def __init__(self, data=None, session_id=None, new=True, max_age=12 * 60 * 60):
```
di, line 104:
```suggestion
    def __init__(self, data=None, session_id=None, new=True):
```

warehouse/sessions.py
```diff
@@ -116,6 +119,8 @@ def __init__(self, data=None, session_id=None, new=True):
         # We'll track all of the IDs that have been invalidated here
         self.invalidated = set()
 
+        self.max_age = max_age
+
```
di, line 123:
```suggestion
```
Not used

warehouse/sessions.py
```diff
@@ -247,7 +265,7 @@ def _process_request(self, request):
 
         # If we were able to load existing session data, load it into a
         # Session class
-        session = Session(data, session_id, False)
+        session = Session(data, session_id, False, max_age=self.max_age)
```
di, line 268:
```suggestion
        session = Session(data, session_id, False)
```

warehouse/templates/pages/help.html
```diff
@@ -470,6 +472,15 @@ <h3 id="apitoken">{{ apitoken() }}</h3>
         </p>
         <p>{% trans %}Advanced users may wish to inspect their token by decoding it with base64, and checking the output against the unique identifier displayed on PyPI.{% endtrans %}</p>
 
+        <h3 id="sensitiveactions">{{ sensitiveactions() }}</h3>
+
+        {% trans %}
+          <p>PyPI asks you to confirm your password before you want to perform a <i>sensitive action</i>. Sensitive actions include things like adding or removing maintainers, deleting distributions, generating API tokens, and setting up 2FA.</p>
```
di, line 478:
```suggestion
          <p>PyPI asks you to confirm your password before you want to perform a <i>sensitive action</i>. Sensitive actions include things like adding or removing maintainers, deleting distributions, generating API tokens, and setting up two-factor authentication.</p>
```

warehouse/routes.py
```diff
@@ -296,6 +296,9 @@ def includeme(config):
     )
     config.add_route("packaging.file", files_url)
 
+    # Re-Authentication Route
+    config.add_route("reauthenticate", "/reauthenticate/", domain=warehouse)
+
```
di, line 301: I'm thinking this should probably be mounted under `/account/reauthenticate` instead, which will ensure that we won't have any cache issues because everything under `/account` is not cached.

This means that this should move up where the rest of the `/account` routes are defined above, and the view should move into `warehouse/accounts/views.py`, and tests as appropriate.

PR #8210, opened 2020-07-05
Fix #8188 - incorrect stats in testpypi

### LINKED ISSUE #8188: Ordering at /stats is incorrect for TestPyPI
The list at https://test.pypi.org/stats/ is incorrect and lists projects with 0 bytes first:

Screen Shot 2020-07-01 at 10 06 59 AM

The 0-byte projects should not be included in that list. 

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

If you are working on this issue and have questions, feel free to ask them here, in the [`#pypa-dev` chat channel on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or on the [distutils-sig.python.org mailing list](https://mail.python.org/mailman3/lists/distutils-sig.python.org/).

kobibarhanin:
Closes #8188.

I've made a small modification to the view that collects packages stats so that now it'll not display any packages that are of size 0.

I've tested it locally:
- Before fix, after adding a zero size package to display:
before
- After fix:
after

I've also tested and linted locally with make.

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -405,15 +405,6 @@ def stats(request):
     )
     # Move top packages into a dict to make JSON more self describing
 
-    print("TOP PACKAGES:")
-    print(top_100_packages)
-    print("TOP PACKAGES WITH ZERO_SIZE:")
-    zero_package = ('zero_size_package', 0)
-    none_package = ('none_package', None)
-    top_100_packages.append(zero_package)
-    top_100_packages.append(none_package)
-    print(top_100_packages)
-
     top_packages = {
         pkg_name: {"size": int(pkg_bytes)}
         for pkg_name, pkg_bytes in top_100_packages if pkg_bytes not in [0, None]
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -404,9 +404,19 @@ def stats(request):
         .all()
     )
     # Move top packages into a dict to make JSON more self describing
+
+    print("TOP PACKAGES:")
+    print(top_100_packages)
+    print("TOP PACKAGES WITH ZERO_SIZE:")
+    zero_package = ('zero_size_package', 0)
+    none_package = ('none_package', None)
+    top_100_packages.append(zero_package)
+    top_100_packages.append(none_package)
+    print(top_100_packages)
+
     top_packages = {
-        pkg_name: {"size": int(pkg_bytes) if pkg_bytes is not None else 0}
-        for pkg_name, pkg_bytes in top_100_packages
+        pkg_name: {"size": int(pkg_bytes)}
+        for pkg_name, pkg_bytes in top_100_packages if pkg_bytes not in [0, None]
     }
 
     return {"total_packages_size": total_size, "top_packages": top_packages}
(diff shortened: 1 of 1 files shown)


warehouse/views.py
```diff
@@ -405,8 +405,9 @@ def stats(request):
     )
     # Move top packages into a dict to make JSON more self describing
     top_packages = {
-        pkg_name: {"size": int(pkg_bytes) if pkg_bytes is not None else 0}
+        pkg_name: {"size": int(pkg_bytes)}
         for pkg_name, pkg_bytes in top_100_packages
+        if pkg_bytes not in [0, None]
```
di, line 410: It seems like this would result in less than 100 packages being displayed. Maybe the query should be adjusted instead?

kobibarhanin: Yep, thanks. I've modified the query to filter for above 0 packages.

warehouse/views.py
```diff
@@ -399,14 +399,16 @@ def stats(request):
...
         .order_by(Project.total_size.desc())
         .limit(100)
         .all()
     )
     # Move top packages into a dict to make JSON more self describing
     top_packages = {
-        pkg_name: {"size": int(pkg_bytes) if pkg_bytes is not None else 0}
+        pkg_name: {"size": int(pkg_bytes)}
         for pkg_name, pkg_bytes in top_100_packages
+        if pkg_bytes is not None
```
di, line 411: This conditional would still have the effect of removing some packages from the 100 found in the query, no?

I think the original issue is that `total_size` is null, not that it's zero. The reason it's appearing at the top of the results is because our ORM orders nulls before non-null values. The reason it's appearing as zero is because of the `if pkg_bytes is not None else 0` in the comprehension.

I think instead we can just tell it to put nulls last when ordering:

```python
.order_by(Project.total_size.desc().nullslast())
```

imsahil007:
@kobibarhanin  Are you going to write a jquery instead for filtering out those packages with 0 or None size so that we can print exact 100 rows instead of this? I would like to work on this. If that's okay with you?

warehouse/views.py
kobibarhanin: No problem, I'll add it - but shouldn't this conditional remain in case null sizes still make it to the top 100 even with reordering? or it's too unlikely....

di: I think it's pretty unlikely, but the original `if pkg_bytes is not None else 0` should handle that if it happens, so that should probably remain.

kobibarhanin: Ok thanks - and from behavioural POV, in case of None do we want to have it displayed and keep the cap at 100 or do we prefer to filter it out and have potentially less then 100 (in the unlikely event)?

di: We want the current behavior (displayed as zero) we just want it to come last.

kobibarhanin: Ok, so essentially I've simply added the _.nullslast()_ like you've suggested and that produces the behaviour needed - nulls will only get pulled from db in an unlikely event and in that case get assigned a zero value and appear last on the table.

di:
Thanks!

Issue #8187, opened 2020-07-01
Display file and total project size limits on project settings page


di:
We should display the current file size limit (`Project.upload_limit`) and total size limit (`Project.total_size_limit`)
on the project settings page for all projects, with external links to https://pypi.org/help/#file-size-limit and https://pypi.org/help/#project-size-limit, respectively.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

If you are working on this issue and have questions, feel free to ask them here, in the [`#pypa-dev` chat channel on Freenode](https://webchat.freenode.net/?channels=%23pypa-dev), or on the [distutils-sig.python.org mailing list](https://mail.python.org/mailman3/lists/distutils-sig.python.org/).

**Screenshot Required**: *If your pull request makes a visual change*, include a screenshot of your update. This helps our team give you feedback faster.

imsahil007:
Hey, @di I have updated the template for this. Which file do I have to update for updating the text value inside the anchor tag?

di:
Re-opening this, I missed that in #8221 it is showing `0 bytes` instead of the default values when a limit is not set.

@imsahil007, can you take a look?

imsahil007:
> Re-opening this, I missed that in #8221 it is showing `0 bytes` instead of the default values when a limit is not set.
> 
> @imsahil007, can you take a look?

I missed that! I checked the values only after updating the limits. I will look into that.

imsahil007:
![Screenshot from 2020-07-08 12-14-13](https://user-images.githubusercontent.com/35963992/86886484-a29e2000-c114-11ea-9b02-04c9f593d476.png)
I think this will close the issue for now.
The default values in the database are set to 0. Even in the admin panel if the value is not there we are printing the warehouse.forklift constants MAX_FILESIZE and so... I used the same approach.

PR #8214, opened 2020-07-06
Add project size limit in project settings 

imsahil007:
#8187 
Screenshot from 2020-07-06 22-40-20
Which file I have to change for updating the values inside these parameters?

### DIFF
--- a/warehouse/templates/manage/settings.html
+++ b/warehouse/templates/manage/settings.html
@@ -20,9 +20,9 @@
 {% block main %}
   <h2>{% trans %}Project settings{% endtrans %}</h2>
   {% trans project_upload_limit=project.upload_limit | int / 1048576 , help_url=request.help_url(_anchor='file-size-limit') %}Project upload limit is: 
-  <a href="{{ help_url }}">{{ project_upload_limit }} MiB{% endtrans %}</a><br><br>
+  <a href="{{ help_url }}">{{ project_upload_limit }} MiB</a>{% endtrans %}<br><br>
   {% trans project_total_limit=project.total_size_limit | int / 1073741824 , help_url=request.help_url(_anchor='project-size-limit') %}Project total size limit is: 
-  <a href="{{ help_url }}">{{ project_total_limit }} GiB{% endtrans %}</a><br><br>
+  <a href="{{ help_url }}">{{ project_total_limit }} GiB</a>{% endtrans %}<br><br>
   <h3>{% trans %}API tokens{% endtrans %}</h3>
   <p>{% trans %}API tokens provide an alternative way to authenticate when uploading packages to PyPI.{% endtrans %}</p>
   {% if user.has_primary_verified_email %}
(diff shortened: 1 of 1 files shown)


warehouse/templates/manage/settings.html
```diff
@@ -20,6 +20,12 @@
 {% block main %}
   <h2>{% trans %}Project settings{% endtrans %}</h2>
 
+  {% trans project_upload_limit=project.upload_limit %}Project upload limit is: 
+  <a href="https://pypi.org/help/#file-size-limit">{{ project_upload_limit }}{% endtrans %}</a><br><br>
+  
+  {% trans project_total_size_limit=project.total_size_limit %}Project total size limit is: 
+  <a href="https://pypi.org/help/#project-size-limit">{{ project_total_size_limit }}{% endtrans %}</a><br>
+
```
di, line 28: The actual values should come from the `project` variable which is available in this template (see how it's used elsewhere in this file)

The help links should be created with `request.help_url` instead.

imsahil007:
Screenshot from 2020-07-07 01-58-15
I think this is okay. Followed both your advice and made the necessary changes.

imsahil007:
I think this is okay. Followed both your advice and made the necessary changes.
Screenshot from 2020-07-07 02-24-33

warehouse/templates/manage/settings.html
```diff
@@ -19,7 +19,10 @@
 
 {% block main %}
   <h2>{% trans %}Project settings{% endtrans %}</h2>
-
+  {% trans project_upload_limit=project.upload_limit | int / 1048576 , help_url=request.help_url(_anchor='file-size-limit') %}Project upload limit is: 
+  <a href="{{ help_url }}">{{ project_upload_limit }} MiB</a>{% endtrans %}<br><br>
+  {% trans project_total_limit=project.total_size_limit | int / 1073741824 , help_url=request.help_url(_anchor='project-size-limit') %}Project total size limit is: 
+  <a href="{{ help_url }}">{{ project_total_limit }} GiB</a>{% endtrans %}<br><br>
```
di, line 25: You can use the [`filesizeformat`](https://jinja.palletsprojects.com/en/2.11.x/templates/#filesizeformat) filter here instead to convert these to human-readable values.

imsahil007: Okay. I will use that. One more thing
Do I have to run 
$make translation 
Before committing?
It's not passing the test case.

di: Yes, see https://warehouse.readthedocs.io/development/submitting-patches/#translations

imsahil007: Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/sahil/Downloads/warehouse/warehouse/i18n/__init__.py", line 15, in <module>
    from babel.core import Locale
ModuleNotFoundError: No module named 'babel'
/.state/env/bin/isort -rc *.py warehouse/ tests/
/bin/sh: 1: /.state/env/bin/isort: not found
make: *** [Makefile:103: reformat] Error 127

I have already installed babel. I don't have any idea why this is happening.

PR #8217, opened 2020-07-06
Ensure Project.total_size is updated on Release removal

### LINKED ISSUE #8213: Project total_size is not updated when a release/file is removed
In https://github.com/pypa/pypi-support/issues/478 users are saying they have removed files/releases but this is not changing the `Project.total_size` for their project.

ewdurbin:
Previously this trigger was not being fired when Release objects were deleted.

It seems as though row level Postgres triggers are not fired on CASCADE'd deletes.

Resolves #8213

### DIFF
--- a/warehouse/migrations/versions/87509f4ae027_update_project_size_on_release_removal.py
+++ b/warehouse/migrations/versions/87509f4ae027_update_project_size_on_release_removal.py
@@ -1,0 +1,158 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


ewdurbin:
CC: @DavidBord I think this makes sense based on your work but would appreciate your review as well.

di:
Do we need to add a trigger on `projects` too, for when project deletion cascades to releases/release files?

ewdurbin:
> Do we need to add a trigger on `projects` too, for when project deletion cascades to releases/release files?

Since we're storing Project total_size on the Project, we do not! We're deleting the row that contains the column in question.

di:
It seems like we should probably also trigger a one-time recalculation of `total_size` for all projects, so users don't have to manually trigger this to get an updated/correct `total_size`.

ewdurbin:
> It seems like we should probably also trigger a one-time recalculation of `total_size` for all projects, so users don't have to manually trigger this to get an updated/correct `total_size`.

It's in the migration :-D

di:
> It's in the migration :-D

D'oh, missed that!

ewdurbin:
> > It's in the migration :-D
> 
> D'oh, missed that!

2020, grace and patience abounds <3

DavidBord:
@ewdurbin seems legit

PR #8230, opened 2020-07-08
Modal will close by pressing Esc key

### LINKED ISSUE #8204: Close the file hashes dialog box if Esc is pressed on the keyboard

It's annoying to not be able to close the "Hashes" dialog box on package pages via the keyboard (at least, not without hitting Tab a million times). [Example](https://pypi.org/project/wheel/#copy-hash-modal-00fac0d9-361c-40bf-acf3-12bf93359ccf).

Definitely not a major issue, but a papercut that it would be nice to solve nonetheless :-)

Hitting Esc should close this dialog box.

**Additional context**

I'm on Firefox 78.0.1. Thanks for maintaining PyPi, your hard work is much appreciated <3

imsahil007:
Closes #8204

### DIFF
--- a/warehouse/static/js/warehouse/index.js
+++ b/warehouse/static/js/warehouse/index.js
@@ -119,8 +119,8 @@ docReady(() => {
 // Close modals when escape button is pressed
 docReady(() => {
   document.addEventListener("keydown", event => {
-    // Only handle the escape key press when a modal is open
-    if (document.querySelector(".modal:target") && event.keyCode === 27) {
+    if (event.keyCode === 27) {
+      window.location.href = "#modal-close";
       for (let element of document.querySelectorAll(".modal")) {
         application
           .getControllerForElementAndIdentifier(element, "confirm")


imsahil007:
ezgif-3-f73995a54550

warehouse/static/js/warehouse/index.js
```diff
@@ -119,8 +119,8 @@ docReady(() => {
 // Close modals when escape button is pressed
 docReady(() => {
   document.addEventListener("keydown", event => {
-    // Only handle the escape key press when a modal is open
-    if (document.querySelector(".modal:target") && event.keyCode === 27) {
+    if (event.keyCode === 27) {
+      window.location.href = "#modal-close";
```
di, line 123: Hmm, it seems like this listener is already working as expected for other modals. I think this change will make it so that the location is changed any time the user presses the <kbd>Esc</kbd> key, which is probably not what we want.

I think it's probably not working for the hash modal because there is no "confirm" element for it to call `.cancel()` on (see below). Maybe a better solution would be to make this work in the same way as the existing modals instead?

di:
Closing due to lack of response.

Issue #4440, opened 2018-08-02
Implement soft deletes for projects, releases and files


di:
**What's the problem this feature will solve?**
Currently there is some user frustration/friction about PyPI's policy of disallowing the reuse of filenames for distributions. I.e., once you delete a given distribution or release, it cannot be re-uploaded.

**Describe the solution you'd like**
One way we can improve this experience without reversing our policy is to allow for "soft" deletes. This would allow a maintainer to "delete" a file, release or project (same behavior as currently exist), but have the ability to see "deleted" files/releases/projects in the management UI, and be able to reverse the deletion.

Because we are currently not removing the actual files from our storage service, this will not result in significantly increased disk space. And while this will result in a slight increase in data in our database, it should be negligible because deletes happen relatively infrequently.

This feature is slightly complicated by the fact that currently we hook into the SQLAlchemy creation/deletion events to determine when to purge various pages from the cache. Removing true deletion of database objects will necessitate finding another way to initiate the purges.

**Additional context**
See also https://github.com/pypa/packaging-problems/issues/74, where this was originally described in https://github.com/pypa/packaging-problems/issues/74#issuecomment-137775034.

dstufft:
I actually don't think it's complicated at all by our SQLa hooks, because assuming soft delete is implemented as a column on the object, then that'll trigger the hook into the SQLAlchemy modified event (which we already have) to purge the cache.

It would only be complicated if we implemented soft delete as an additional table or something silly like that.

di:
Right, of course. Should be pretty simple then, actually.

dstufft:
I think the biggest questions I have is what should this look like in practice?

Soft deletes for files are obvious, we treat them as immutable anyways so there isn’t much question about them. 

Releases are a bit trickier. Right now you can delete a release and upload a new release with the same version but different content. However this only really works if you are able to upload files that didn’t already exist (e.g. you could upload a .zip if you had previously uploaded a .tar.gz). However it might be reasonable to just treat this as closing another loophole to where there is no longer a way to change a release metadata after it was first created. One edge case about this is what would we do if you tried to upload a file that had never existed to a soft deleted release? Would we undelete implicitly? What if the metadata changed?

[...]

PR #8237, opened 2020-07-09
Soft deletes for releases, projects, and files

VikramJayanthi17:
As discussed in #4440, this draft PR is being used to figure out how to best solve the soft-deletes issue, specifically using the 1st approach laid out [here](https://github.com/pypa/warehouse/issues/4440#issuecomment-656332919). This doesn't contain all of the logic for soft-deletes(multiple UI changes, Etc.) yet and is still WIP.


di:
Closing this since it's quite stale.

PR #8240, opened 2020-07-09
Update Warehouse documentation to inform about BigQuery datasets

### LINKED ISSUE #7403: Public Dataset for distribution metadata

patelneel55:
Closes #7403.


docs/api-reference/index.rst
```diff
@@ -68,3 +68,4 @@ Available APIs
     stats
     xml-rpc
     integration-guide
+    bigquery-dataset
```
pradyunsg, line 71: nit: bigquery or bigquery-datasets

PR #8241, opened 2020-07-10
Changing token template to reflect PyPI site

### LINKED ISSUE #8132: Section name in .pypirc for testpypi should not be pypi
**Describe the bug**
When I create a new API token in test.pypi.org, it gives tips as following screenshot.

However, after I copied the content to ~/.pypirc, then twine command still asks for password . I tried to resolve and found that it only works when the section name change from [pypi] to [testpypi].

It seemed that the tip is same with main PYPI site, but a bit misleading. Make it sync with current site could be more clear.

The problem should be not related to OS platform or browser types.

**Additional context**

图片

callmecampos:
Closes #8132


merwok:
Thanks for this! :sparkles:

PR #8243, opened 2020-07-10
Fix crontab parameter for celery task

patelneel55:


ewdurbin:
Travis status not propagating... :/

PR #8255, opened 2020-07-13
GitHub Actions CI

### LINKED ISSUE #7001: Look at using GitHub Actions for CI

callmecampos:
Testing Warehouse using GitHub Actions. We run all of our jobs in parallel using the native `ubuntu-latest` Docker containers provided by GitHub Actions.

- The testing job runs inside of a PostgreSQL [service container](https://docs.github.com/en/actions/configuring-and-managing-workflows/about-service-containers), specifically `postgres:10.1` instead of `postgres:latest`.
- ~Ideally we would cache Docker containers with GitHub Package Registry instead of Docker Hub (because of [faster build time when compared to other caching methods](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei)), however it doesn't yet seem possible to use for container actions because accessing even public packages on GitHub Package Registry requires an extra authentication step (see https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358 for ongoing discussion). In case GitHub fixes this and allows a simpler integration of Actions and Package Registry, I've left the (skipped) job in [.github/workflows/main.yml](https://github.com/callmecampos/warehouse/blob/github-actions-ci/.github/workflows/main.yml#L55-L101).~
- We use GitHub Actions' native caching action to cache pip dependencies and node modules for the following jobs: `test`, `lint`, `deps`, `static-tests`, `static-pipeline`.

Resolves #7001. You can see the latest runs with runtimes at the bottom of this PR, or in the Checks tab.

~This won't yet be able to run on pypa/warehouse since the secret environment variables for Docker Hub haven't been set up in this repo, only in my fork.~ This now runs fine on pypa/warehouse.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,60 @@
+name: 'Setup'
+description: 'Set up CI environment'
+runs:
+  using: "composite"
+  steps:
+    - name: Check out repository
+      uses: actions/checkout@v2
+
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.8.2
+
+    - name: Cache Python dependencies
+      uses: actions/cache@v2
+      env:
+        cache-name: warehouse-cache-pip
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('requirements.txt', 'requirements/*.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ github.job }}-${{ env.cache-name }}-
+          ${{ runner.os }}-${{ github.job }}-
+          ${{ runner.os }}-
+
+    - name: Install Python dependencies
+      run: |
+        pip install -U pip setuptools wheel
+        pip install -r requirements.txt
+        pip install -r requirements/dev.txt
+
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 14.4.0
+
+    - name: Cache Node dependencies
+      uses: actions/cache@v2
+      env:
+        cache-name: warehouse-cache-npm-modules-static
+      with:
+        path: |
+          ~/.npm
+          **/node_modules
+        key: ${{ runner.os }}-build-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-build-${{ github.job }}-${{ env.cache-name }}-
+          ${{ runner.os }}-build-${{ github.job }}-
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+
+    - name: Install Node dependencies
+      run: npm ci
+
+    - name: Check versions
+      run: |
+        node -v
+        npm -v
+        python --version
+
+    - run: echo "Setup success!"
+      shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,11 @@
+on: [push]
+
+jobs:
+  hello_world_job:
+    runs-on: ubuntu-latest
+    name: A job to say hello
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ./.github/actions/setup
+    - run: echo Done
+      shell: bash


pavanagrawal123:
hey @callmecampos, this looks really good! I had #7881 open for this, but I didn't have much time recently to fix it up. This is a way better approach then I had going. I'm going to close #7881 in favor of this!

tests/conftest.py
```diff
@@ -144,7 +145,13 @@ def database(request):
...
         janitor.drop()
 
-    return "postgresql://{}@{}:{}/{}".format(pg_user, pg_host, pg_port, pg_db)
+    pass_field = ""
+    if pg_pass:
+        pass_field = f":{pg_pass}"
+
+    return "postgresql://{}{}@{}:{}/{}".format(
+        pg_user, pass_field, pg_host, pg_port, pg_db
+    )
```
di, line 154: Are these changes necessary for GitHub Actions?

Makefile
```diff
@@ -115,8 +116,8 @@ ifneq ($(TRAVIS), false)
 	# We're on Travis, so we can lint static files locally
 	./node_modules/.bin/eslint 'warehouse/static/js/**' '**.js' 'tests/frontend/**' --ignore-pattern 'warehouse/static/js/vendor/**'
 	./node_modules/.bin/sass-lint --verbose
-else
-	# We're not on Travis, so we should lint static files inside the static container
+else ifneq ($(ACTIONS), true)
+	# We're not on Travis or GitHub Actions, so we should lint static files inside the static container
```
di, line 120: We should do the same thing as if we were on Travis here. As-is, this is not running any static linting on Actions.

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+        ports:
+          # Maps tcp port 5432 on service container to the host
+          - 5432:5432
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
di, line 29: Can we use `3.8.2` here?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+
+      - name: Cache pip dependencies
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-pip-deps-warehouse-p
```
di, line 151: What's with the `-p` on the end of this?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+  static-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules-static-dh
```
di, line 214: Same question here and below about the `-dh`

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
+name: CI
+
+on: [push, pull_request]
+
+env:
+  CACHE_SEED: 1
```
di, line 6: If this is always `1`, why are we including it in the keys for the cache?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+        run: sudo apt -y install libcurl4-openssl-dev libssl-dev pkg-config
+
+      - name: Install pip requirements
+        run: |
+          mkdir ./.cache && mkdir ./.cache/pip
+          chown -R $(whoami) ./.cache/pip
+          pip install -U pip setuptools wheel --cache-dir ./.cache/pip
+          pip install -r requirements/main.txt --cache-dir ./.cache/pip
+          pip install -r requirements/tests.txt --cache-dir ./.cache/pip
+          pip install -r requirements/deploy.txt --cache-dir ./.cache/pip
```
di, line 84: Is there a reason we can't use the default (`~/.cache/pip`) here?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+          
+      - name: Setup node
+        run: |
+          curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.0/install.sh | bash
+          export NVM_DIR="$HOME/.nvm" 
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"
+          nvm install $NODE_VERSION
```
di, line 92: Can we just use https://github.com/actions/setup-node to set up a specific node version here instead?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+          npm i -g npm
+          npm install -g gulp-cli
+          npm ci
+
+      - name: Run tests
+        run: |
+          python -m coverage run -m pytest --strict --postgresql-host localhost
+          python -m coverage html
+          python -m coverage report -m --fail-under 100
```
di, line 101: Why can't we just run `bin/tests` here instead?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+        env:
+          # The hostname used to communicate with the PostgreSQL service container
+          POSTGRES_HOST: localhost
+          # The default PostgreSQL port
+          POSTGRES_PORT: 5432
```
di, line 106: Are these used? The test command above is specifying `localhost` not `POSTGRES_HOST`. Also, are these necessary if they're the defaults?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+          key: ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-${{ env.cache-name }}-
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Dependency Setup
+        run: npm install -D babel
```
di, line 228: Why not `npm ci`? Why does `babel` have to be specifically installed here?

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+          cache-name: cache-nvm-pipeline-dh
+        with:
+          path: ~/.nvm/versions/
+          key: ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-${{ env.cache-name }}-
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-${{ github.job }}-
+            ${{ runner.os }}-build-${{ env.CACHE_SEED }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
```
di, line 251: I think this is unnecessary if we use https://github.com/actions/setup-node instead.

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+
+      - name: Dependency Setup
+        run: npm install -D babel
```
di, line 270: Same question here about babel.

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+
+      - name: Static Pipeline Check
+        run: bin/static_pipeline # this periodically fails for no reason, just rerun the job if needed
```
di, line 273:
```suggestion
        run: bin/static_pipeline
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,273 @@
...
+          restore-keys: |
+            ${{ runner.os }}-${{ env.CACHE_SEED }}-${{ github.job }}-${{ env.cache-name }}-
+            ${{ runner.os }}-${{ env.CACHE_SEED }}-${{ github.job }}-
+            ${{ runner.os }}-${{ env.CACHE_SEED }}-
+            ${{ runner.os }}-
+
+      - name: Set cache permissions
+        run: |
+          mkdir ~/.cache && mkdir ~/.cache/pip
+          chown -R $(whoami) ~/.cache/pip
```
di, line 164: Is this necessary? The later pip-install commands in `make deps` don't use this custom cache directory.

tests/conftest.py
callmecampos: Nope, good catch. That was for when we were using the `postgres:latest` container which requires a password to access the db.

.github/workflows/main.yml
callmecampos: Should be able to, I'll try it out.

.github/workflows/main.yml
callmecampos: TLDR: It shouldn't be there.

The caching action is strange since it doesn't update the cache after the first time it caches dependencies, you have to change the key if you want a new one. So, I was sort of mindlessly adding characters whenever I would change stuff lol. However, I just found that it's recommended to [hash the dependency files](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#using-contexts-to-create-cache-keys) such that the keys change when the dependencies change, so I'm gonna update the .yml file to reflect that.

.github/workflows/main.yml
callmecampos: Same as [the comment above](https://github.com/pypa/warehouse/pull/8255#discussion_r468752304). TLDR: shouldn't be there.

.github/workflows/main.yml
callmecampos: Same as [the comment above](https://github.com/pypa/warehouse/pull/8255#discussion_r468752304). TLDR: shouldn't be there. I believe I was using it just as a generic way to update the cache.

.github/workflows/main.yml
callmecampos: I'll test this out with the default `~/.cache/pip` now that we're using the regular `ubuntu-latest` container, but I was having issues in the multi-stage Docker build where the permissions on `~/.cache/pip` were set such that I wasn't able to read or write to it.

.github/workflows/main.yml
callmecampos: Actually, it seems I'm already hashing the dependency files where I need to be. I think I did this because I was changing which directories were being cached (e.g. `~/.cache/pip` vs `./.cache/pip`, `~/.npm` vs `~/.npm && **/node_modules`) and needed the updated cache to reflect that. Still, I'll go ahead and change the `cache-name` to be something readable.

.github/workflows/main.yml
callmecampos: Yep, not necessary, was using this in a previous iteration when I was trying different hosts/ports. Good catch.

webknjaz:
@callmecampos Most (if not all) of the jobs here can be collapsed into a nice matrix. Are there any reasons for not implementing this?

.github/workflows/main.yml
```diff
@@ -0,0 +1,272 @@
+name: CI
+
+on: [push, pull_request]
```
webknjaz, line 3: I think adding a cron schedule here would be nice too

.github/workflows/main.yml
```diff
@@ -0,0 +1,272 @@
...
+
+env:
+  PYTHON_VERSION: 3.8.2
+  NODE_VERSION: 14.4.0
```
webknjaz, line 7: This won't be necessary if you make a matrix...

di:
Hey @webknjaz, @callmecampos was working on this as part of his internship which is now finished. Would you like to take over here?

webknjaz:
@di ah, I didn't know. If he is not going to keep working on this PR, I'll try to find some time myself, but no promises.

PR #8293, opened 2020-07-20
Fix BigQuery OOM issue by yielding rows into memory in smaller batches

patelneel55:


warehouse/packaging/tasks.py
```diff
@@ -244,8 +227,39 @@ def populate_data_using_schema(file):
...
+
+        json_rows = [populate_data_using_schema(file) for file in release_files]
 
-    bq.load_table_from_json(
-        json_rows, table_name, job_config=LoadJobConfig(schema=table_schema)
-    ).result()
+        bq.load_table_from_json(
+            json_rows, table_name, job_config=LoadJobConfig(schema=table_schema)
+        ).result()
+        break
```
di, line 265:
```suggestion
```

Issue #8299, opened 2020-07-21
Using undeclared dependency attrs


ewjoachim:
**Describe the bug**

From @di 

Regarding:
https://github.com/pypa/warehouse/blob/b344984ee83126eb6fc541406d1f34719fad8b6b/warehouse/email/services.py#L36

> Currently attrs is not a top-level dependency of Warehouse, it's only available as a sub-dependency of automat. If we truly need to use attrs here, it should be added to https://github.com/pypa/warehouse/blob/master/requirements/main.in and those dependencies should be re-compiled.

(source: https://github.com/pypa/warehouse/pull/7124#discussion_r453729298)

**Expected behavior**

attrs should be added in main.in OR removed from the codebase.

di:
I've added the "good first issue" label to this issue.

I also said on that comment: 

> However I think I'd prefer to just use a regular class here.

ewjoachim:
Ah, I didn't know if it applied here too ! :)

PR #8313, opened 2020-07-24
Update main.in

Manthan03583:


ewdurbin:
Hi @Manthan03583, can you explain more about what you're attempting to do with this change? This file is not intended to be interpreted by Python.

Manthan03583:
> Hi @Manthan03583, can you explain more about what you're attempting to do with this change? This file is not intended to be interpreted by Python.

I am just trying do this with python

di:
Looks like this is trying to fix #8299, is that correct?

ewdurbin:
Documentation on adding new dependencies (which is effectively the same as declaring one explicitly) is available at https://warehouse.pypa.io/development/patterns/#adding-new-dependencies

eamanu:
Hi! yes this PR seems to be a fix of #8299. Do you need some help @Manthan03583

Manthan03583:
> Looks like this is trying to fix #8299, is that correct?

Issue #6099, opened 2019-06-26
Create keyboard shortcut to focus search bar


nlhkabu:
**What's the problem this feature will solve?**
We have decided *not* to autofocus the search bar on the PyPI homepage (see #6088)

However, we should consider creating a keyboard shortcut to move users to the search bar.

**Describe the solution you'd like**
The solution should be:

- Easy for PyPI users to discover
- Simple
- It should *not* interfere with keyboard shortcuts used by screenreader users (we need to research this)

**Additional context**
Suggestion from @amcgregor:

On YouTube typing `/` activates search, while typing `?` shows a shortcut reference.  See the [original comment](https://github.com/pypa/warehouse/issues/3440#issuecomment-505631589) for more info.

amcgregor:
Here's a short story summary for this, if this helps in any way.  :)

---

As a Python developer attempting to find back ("jump to") a specific project, or looking for a package that meets my requirements, I want to be able to rapidly search so that I can get on with the real work.

### Discoverability

**Given** a _new visitor_ to the package index,
**And** they haven not seen the hint yet (i.e. this is their first visit),
**Then** a _hint should be temporarily shown_ suggesting `/` (for quick search) and `?` to discover all available shortcuts.

**Given** a _visitor to the package index unaware of the availability of shortcuts_,
**When** the user _presses keys not assigned to shortcuts_,
**Then** a _hint should be temporarily shown_ suggesting `/` (for quick search) and `?` to discover all available shortcuts.

### Proposed Involved Shortcuts

**Given** a _visitor_ to the package index,
**When** that visitor _presses_ `/`,
**Then** the _search field_ should become focused.

**Given** a _visitor_ to the package index,
**When** that user _presses_ `?`,
**Then** a _popover reference of shortcuts_ should become visible.

---

Rationale: a growing de-facto standard across multiple sites, e.g. GitHub, YouTube, where `/` replicates the command-mode Vim shortcut for search.  Reddit has keyboard shortcuts for most navigation and actions (voting, replying, saving, hiding, collapsing/expanding), with `Q` as the shortcut for accessing search-like filters, preserving `?` as the discoverability option.  Additional discovery of sites involving shortcuts like these is somewhat hampered by the generic nature of the terms, "websites using keyboard shortcuts" finds nothing but keyboard shortcut references for Windows and things like that.  ;)

nlhkabu:
Note: WCAG 2.1 - Criteria 2.1.4:

[...]

PR #8319, opened 2020-07-25
Autofocus for the search field

brejoc:
I guess the primary use case of the warehouse is searching or looking libs or applications. Having the search field focused right from the start would speed this task up tremendously.

### DIFF
--- a/warehouse/templates/index.html
+++ b/warehouse/templates/index.html
@@ -51,7 +51,7 @@
 
     <form class="search-form search-form--large search-form--fullwidth" action="{{ request.route_path('search') }}" role="search">
       <label for="search" class="sr-only">{% trans %}Search PyPI{% endtrans %}</label>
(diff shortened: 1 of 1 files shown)


ewdurbin:
Hi @brejoc! Thanks for this PR. This feature is not desirable from an accessibility perspective.

This topic has been extensively discussed and researched, see #6088. 

The feature in #6099 is another way to consider contributing.

brejoc:
@ewdurbin Thanks for pointing this out. On question. There where some numbers presented in #6088. Did somebody make sure that those numbers are not tainted by bots and crawlers?

ewdurbin:
@brejoc that's a great question to present to #6088. I don't have information to add here.

brejoc:
@ewdurbin I'd love to, but #6088 is locked. I can't comment there.

PR #8320, opened 2020-07-25
Add a HTTP-308 redirect for /legacy

### LINKED ISSUE #8104: Incorrect http response from invalid upload endpoint
**Describe the bug**

As a user attempting to upload distributions to PyPI, I want to get a failure response from the PyPI http API when I attempt to upload files to an incorrect endpoint, so that I am able to correctly identify failures and successes and correct them if necessary or provide meaningful error messages.

The correct endpoint for uploading is `https://test.pypi.org/legacy/` (notice the trailing slash)

If an upload is attempted at `https://test.pypi.org/legacy` (without a trailing slash), the upload does not succeed, but a `200` response is returned.

**Expected behavior**

A 4xx or 5xx error code should be returned.

**To Reproduce**

Attempt to upload a distribution to `https://test.pypi.org/legacy`

**My Platform**

Mac, python 3.8, using `flit` as the upload front-end

**Additional context**

fbidu:
If an user makes a POST to /legacy instead of to /legacy/, the API should return a HTTP-308 Permanent Redirect status.

This PR:

* Adds a route in forklift to accomplish that
* Implements the redirect inside forklift.legacy
* Adds and fixes the necessary unit and functional tests

Closes #8104

The route was added with a plain `add_route` in `forklift.includeme`, in contrast with all the previous routes that used `add_legacy_action_route`. Because of this new method being called, I had to patch the current stub at `tests/unit/forklift/test_legacy.py:test_includeme`


warehouse/forklift/legacy.py
```diff
@@ -1504,3 +1510,18 @@ def doc_upload(request):
...
+)
+def permanent_redirect(request):
+    """
+    Deals with requests aimed at /legacy and points the user to the correct
+    /legacy/ route.
+    """
+    return _exc_with_message(
+        HTTPPermanentRedirect,
+        "It looks like you forgot a trailing /"
+        "Please make your POST request to /legacy/. This is /legacy",
```
Julian, line 1526: I think the POST bit may actually be a bit confusing to users seeing this, since I think the typical way someone will encounter this message will be by e.g. typoing the `--repository-url` in a call to `twine`, not by explicitly themselves making a POST request.

Maybe something like `An upload was attempted to /legacy, but the expected upload URL is /legacy/, with the trailing /.` avoids discussing the specifics of HTTP?

warehouse/forklift/legacy.py
```diff
@@ -1504,3 +1510,18 @@ def doc_upload(request):
         "Uploading documentation is no longer supported, we recommend using "
         "https://readthedocs.org/.",
     )
+
+
+@view_config(
+    route_name="forklift.legacy.redirect", require_csrf=False, require_methods=["POST"],
```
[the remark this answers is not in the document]
fbidu: Makes sense, I'll rename those

warehouse/forklift/legacy.py
fbidu: +1 on the wording change

warehouse/forklift/legacy.py
```diff
@@ -1504,3 +1510,18 @@ def doc_upload(request):
...
+)
+def permanent_redirect(request):
+    """
+    Deals with requests aimed at /legacy and points the user to the correct
```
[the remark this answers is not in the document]
fbidu: Oh, good catch. I also prefer the imperative tense

tests/functional/legacy_api/test_basic.py
```diff
@@ -0,0 +1,29 @@
...
+    See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+    """
+    resp = webtest.post("/legacy", status=308)
+    assert resp.status == (
+        "308 An upload was attempted to /legacy, but the expected upload URL is "
+        "/legacy/, with the trailing /"
+    )
+
+    assert "location" in resp.headers
+    assert resp.headers["location"].endswith("/legacy/")
```
fbidu, line 29: There probably is a better way to check that the provided location is indeed a correct URL for a given route but I'm not sure how

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/fbidu/miniconda3/envs/python38/bin/python"
+}
```
ewjoachim, line 3: I think you didn't want to commit that.
Suggestion:
```
git config --global core.excludesfile ~/.gitignore_global
echo ".vscode" > ~/.gitignore_global
```

fbidu: Oh, sorry! And thanks for the tip :sweat_smile:

fbidu:
> I'm seeing already existing views that do the same but are not implemented the same way. Did you consider harmonizing ?
> 
> https://github.com/pypa/warehouse/blob/3fe4e45e6dd014ca6409f101b0fae9463e18f1f9/warehouse/legacy/api/json.py#L79-L89

I had to use `_exc_with_message` due to some `forklift` legacy compatibility of what I'm not fully aware... In my latest commit I did change the location redirection to something similar to `request.route_path("legacy.api.json.project", name=project.name)`

Julian:
Woohoo! Think lgtm from me at this point! A maintainer will likely give you a review soon.

warehouse/forklift/legacy.py
```diff
@@ -1504,3 +1509,21 @@ def doc_upload(request):
...
+)
+def missing_trailing_slash_redirect(request):
+    """
+    Redirect requests to /legacy to the correct /legacy/ route with a
+    HTTP-308 Permanent Redirect
+    """
+    return _exc_with_message(
+        HTTPPermanentRedirect,
+        "An upload was attempted to /legacy, but the expected upload URL is "
+        "/legacy/, with the trailing /",
```
di, line 1527:
```suggestion
        "An upload was attempted to /legacy but the expected upload URL is "
        "/legacy/ (with a trailing slash)",
```

tests/unit/forklift/test_legacy.py
```diff
@@ -3430,3 +3430,17 @@ def test_doc_upload(pyramid_request):
...
+def test_missing_trailing_slash_redirect(pyramid_request):
+
+    pyramid_request.route_path = pretend.call_recorder(lambda *a, **kw: "/legacy/")
+
+    resp = legacy.missing_trailing_slash_redirect(pyramid_request)
+
+    assert resp.status_code == 308
+    assert resp.status == (
+        "308 An upload was attempted to /legacy, but the expected upload URL is "
+        "/legacy/, with the trailing /"
```
di, line 3444:
```suggestion
        "308 An upload was attempted to /legacy but the expected upload URL is "
        "/legacy/ (with a trailing slash)"
```

tests/functional/legacy_api/test_basic.py
```diff
@@ -0,0 +1,29 @@
...
+    Per issue #8104, we should issue an HTTP-308 for a POST
+    in /legacy and point the user to the correct endpoint,
+    /legacy/
+
+    See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+    """
+    resp = webtest.post("/legacy", status=308)
+    assert resp.status == (
+        "308 An upload was attempted to /legacy, but the expected upload URL is "
+        "/legacy/, with the trailing /"
```
di, line 25:
```suggestion
        "308 An upload was attempted to /legacy but the expected upload URL is "
        "/legacy/ (with a trailing slash)"
```

fbidu:
> LGTM, just want to tweak the message here a bit.

Thanks for the suggestion, @di! I have applied them. While working on this issue, I noticed a comment about the `forklift` module being expected to be broken into a different package. Is that refactor still on the roadmap? If so I could try to help with that if you think it would be useful

di:
> While working on this issue, I noticed a comment about the forklift module being expected to be broken into a different package. Is that refactor still on the roadmap? If so I could try to help with that if you think it would be useful

I suppose this is still on the roadmap but it isn't currently high priority. We might want to make a separate issue to discuss.

di:
Thanks!

PR #8323, opened 2020-07-26
Add minutes login block on message to users

eamanu:
In this PR add the information about how many minutes users
need wait before login again.

Fixes https://github.com/pypa/warehouse/issues/1981.


Julian:
@eamanu nice! Looks like a good start.

There are some tests failing in CI that look relevant / need updating: https://travis-ci.org/github/pypa/warehouse/jobs/711839967

Lemme know if you need a hand deciphering those.

Also will leave one comment about units on the code directly.

warehouse/accounts/forms.py
```diff
@@ -121,11 +121,12 @@ def validate_password(self, field):
                     raise wtforms.validators.ValidationError(
                         _("The password is invalid. Try again.")
                     )
-            except TooManyFailedLogins:
+            except TooManyFailedLogins as err:
                 raise wtforms.validators.ValidationError(
                     _(
                         "There have been too many unsuccessful login attempts. "
-                        "Try again later."
+                        f"You have been locked out for {err.resets_in.total_seconds()} "
```
Julian, line 128: So here the units are in seconds, rather than minutes -- but I think there was a suggestion to use the [humanize](https://pypi.org/project/humanize/) module made as well here, given that e.g. if the timeout period is 600 seconds, seeing "10 minutes" is a bit nicer than seeing "600 seconds".

Perhaps after the test suite passes it's worth giving a shot at adding that module as a dependency and seeing whether you can wire it up here.

Julian:
There's one other thing that I think I mentioned in the discord that we'll need to figure out (or a warehouse maintainer may tell us) -- which is how to handle translations. This message is translated in Warehouse's internationalization support -- e.g. [here it's defined](https://github.com/pypa/warehouse/blob/fd22cb615885f1cda09b05e7c86290285ac16597/warehouse/locale/messages.pot#L35) and [here it's translated into Spanish](https://github.com/pypa/warehouse/blob/fd22cb615885f1cda09b05e7c86290285ac16597/warehouse/locale/es/LC_MESSAGES/messages.po#L68-L69).

It looks like there are [some docs here](https://warehouse.readthedocs.io/translations/) on how Warehouse does translations (specifically that it uses a service called [weblate](https://weblate.org/en/) and that the translations should get applied automatically on this PR) -- I'm sure that will work here (or at least I hope so?) but we may need to do something to ensure the humanized value that is passed in can be templated in across languages, not sure yet.

warehouse/accounts/forms.py
```diff
@@ -121,11 +121,12 @@ def validate_password(self, field):
...
                         _("The password is invalid. Try again.")
                     )
-            except TooManyFailedLogins:
+            except TooManyFailedLogins as err:
                 raise wtforms.validators.ValidationError(
                     _(
                         "There have been too many unsuccessful login attempts. "
-                        "Try again later."
+                        "You have been locked out for {0} minutes. "
+                        "Please try again later.".format(err.resets_in.total_seconds())
```
Julian, line 129: Here needs conversion to minutes again. Maybe now that CI is passing though is a good time to try out humanize?

eamanu:
Hi @di sorry,  Idon't found the way to use a macro template inside a python response?

cc @Julian

eamanu:
Maybe we need add humanize dependency?

eamanu:
@Julian @di Hi, this is a proposal of fixing this issue using a new dependency: humanize.

eamanu:
Hi I let here an screenshot of the changes made

image

warehouse/accounts/forms.py
```diff
@@ -122,11 +123,14 @@ def validate_password(self, field):
...
-            except TooManyFailedLogins:
+            except TooManyFailedLogins as err:
                 raise wtforms.validators.ValidationError(
                     _(
                         "There have been too many unsuccessful login attempts. "
-                        "Try again later."
+                        "You have been locked out for {0}. "
+                        "Please try again later.".format(
+                            humanize.naturaldelta(err.resets_in.total_seconds())
+                        )
```
di, line 133: It seems like this is the first time we're trying to pass in a value to a translated string like this. As-is, this produces a string that might get mis-translated and break the formatting. 

Instead, the `translate` function takes a `mapping` argument that should be used for this instead: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/i18n.html#pyramid.i18n.Localizer.translate

We should be able to just do:

```suggestion
                        "You have been locked out for ${time}. "
                        "Please try again later.",
                        mapping={
                            'time': humanize.naturaldelta(err.resets_in.total_seconds()),
                        }
```

I'm not 100% sure that will work though, we might need to allow the `mapping` argument to be explicitly passed through to here: https://github.com/pypa/warehouse/blob/15cc65e57d20268ee2ce0a9ef4cdeb410d5069c7/warehouse/i18n/__init__.py#L98-L108

An explanation for how to do this is also missing from https://warehouse.readthedocs.io/translations/#passing-non-translatable-values-to-translated-strings. Once we figure this out, we should add it to the docs as well.

tests/unit/accounts/test_views.py
```diff
@@ -47,7 +48,11 @@ def test_too_many_failed_logins(self, pyramid_request):
 
         assert resp.status == "429 Too Many Failed Login Attempts"
         assert resp.detail == (
-            "There have been too many unsuccessful login attempts. Try again later."
+            "There have been too many unsuccessful login attempts. "
+            "You have been locked out for {0}. "
+            "Please try again later.".format(
+                humanize.naturaldelta(exc.resets_in.total_seconds())
```
miketheman, line 54: **suggestion:** ‏make the return value explicit here instead of using humanize to calculate it, replacing this line with what we expect humanize to emit, like “5 minutes”.
This reduces the chance that a change to humanize’s behavior pass unchecked into production, since the intent is expressed more clearly in the test case.

miketheman (commented):
Was going through old open issues and came across this one.

@eamanu are you still interested in resolving this? It’s okay if you’ve moved on after this time period, so happy to pick up if you don’t want to continue.

di:
Assigning this to @miketheman to take over!

eamanu:
OMG, I don't know why I lost this PR into my things! from 2020. Sorry sorry for this. I'm re-taking my interest in Python ecosystem. I'm going to pay more attention now.

Please, if you need some help/or something for me just ping me, @miketheman.  

sorry again.

miketheman:
@eamanu no worries! Let me know when you're ready for a review.

eamanu:
Ok! let me a couple of days. Thanks!

eamanu:
Hi, @miketheman I've push an update.

requirements/main.txt
```diff
@@ -790,6 +790,10 @@ html5lib==1.1 \
     --hash=sha256:0d78f8fde1c230e99fe37986a60526d7049ed4bf8a9fadbad5f00e22e58e041d \
     --hash=sha256:b2e5b40261e20f354d198eae92afc10d750afb487ed5e50f9c4eaf07c184146f
     # via -r requirements/main.in
+humanize==2.5.0 \
```
miketheman, line 793: This is a pretty old version of the library. 4.6.0 is latest, and includes type hints that should solve the lint failing.
Would you mind removing this line and running `make deps` again to get the current version?

tests/unit/accounts/test_forms.py
```diff
@@ -216,7 +217,9 @@ def test_validate_password_too_many_failed(self):
         user_service = pretend.stub(
             find_userid=pretend.call_recorder(lambda userid: 1),
             check_password=pretend.call_recorder(
-                pretend.raiser(TooManyFailedLogins(resets_in=None))
+                pretend.raiser(
+                    TooManyFailedLogins(resets_in=datetime.timedelta(seconds=600))
```
miketheman, line 221: Nice! Love the usage of timedelta.

miketheman (changes requested):
Thanks for the update! One change, and then this ought to be good to go.

requirements/main.txt
eamanu: Ready @miketheman thanks!

miketheman (approved):
Thanks for working through this!

PR #8324, opened 2020-07-26
Generic events

di:
This PR introduces a generic `HasEvents` subclass for models that have related events, and migrates the existing `UserEvents` and `ProjectEvents` models to it. It uses generic associations but maintains essentially the same underlying "table per related" structure.

This ensures that events are created and associated with their sources in a consistent and generic way, and allows us to add event support to any new or existing model simply by adding the `HasEvents` mixin to the models class definition.

### DIFF
--- a/warehouse/events/models.py
+++ b/warehouse/events/models.py
@@ -1,0 +1,57 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from sqlalchemy import Column, DateTime, ForeignKey, String, orm, sql
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.ext.declarative import declared_attr
+
+from warehouse import db
+
+
+class Event:
+    tag = Column(String, nullable=False)
+    time = Column(DateTime, nullable=False, server_default=sql.func.now())
+    ip_address = Column(String, nullable=False)
+    additional = Column(JSONB, nullable=True)
+
+
+class HasEvents:
+    @declared_attr
+    def events(cls):  # noqa: N805
+        cls.Event = type(
+            "%sEvent" % cls.__name__,
+            (Event, db.Model),
+            dict(
+                __tablename__="%s_events" % cls.__tablename__,
+                source_id=Column(
+                    UUID(as_uuid=True),
+                    ForeignKey(
+                        "%s.id" % cls.__tablename__,
+                        deferrable=True,
+                        initially="DEFERRED",
+                    ),
+                    nullable=False,
+                ),
+                source=orm.relationship(cls),
+            ),
+        )
+        return orm.relationship(cls.Event, cascade="all, delete-orphan", lazy=True)
+
+    def record_event(self, *, tag, ip_address, additional=None):
+        session = orm.object_session(self)
+        event = self.Event(
+            source=self, tag=tag, ip_address=ip_address, additional=additional
+        )
+        session.add(event)
+        session.flush()
+
+        return event
--- a/warehouse/events/models.py
+++ b/warehouse/events/models.py
@@ -23,31 +23,50 @@ class Event(AbstractConcreteBase):
     ip_address = Column(String, nullable=False)
     additional = Column(JSONB, nullable=True)
 
+    @declared_attr
+    def __tablename__(cls):
+        return "_".join([cls.__name__.removesuffix("Event").lower(), "events"])
+
+    @declared_attr
+    def __mapper_args__(cls):
+        return (
+            {"polymorphic_identity": cls.__name__, "concrete": True}
+            if cls.__name__ != "Event"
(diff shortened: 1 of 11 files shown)


warehouse/migrations/versions/5e02c4f9f95c_generic_events.py
```diff
@@ -0,0 +1,178 @@
...
+        "(id, source_id, tag, time, ip_address, additional) "
+        "SELECT id, project_id, tag, time, ip_address, additional "
+        "FROM project_events"
+    )
+    op.execute(
+        "INSERT INTO users_events "
+        "(id, source_id, tag, time, ip_address, additional) "
+        "SELECT id, user_id, tag, time, ip_address, additional "
+        "FROM user_events"
+    )
```
di, line 84: Not sure if this is the fastest way to do this, but it does work.

di: For a sense of size of these tables:

```
warehouse=> select count(*) from user_events;
 count
--------
 481808
(1 row)

warehouse=> select count(*) from project_events;
 count
--------
 811383
(1 row)
```

ewdurbin:
I'm going to validate the migration locally, but have no issues with this.

I will ask to coordinate merge so I can be on hand though.

warehouse/events/models.py
```diff
@@ -0,0 +1,61 @@
...
+    tag = Column(String, nullable=False)
+    time = Column(DateTime, nullable=False, server_default=sql.func.now())
+    ip_address = Column(String, nullable=False)
+    additional = Column(JSONB, nullable=True)
+
+
+class HasEvents:
+    @declared_attr
+    def events(cls):  # noqa: N805
+        cls.Event = type(
```
dstufft, line 30: My only concern with this, is that ``@declared_attr`` is designed to dynamically compute a declared attribute (which this is doing) but it's not really intended to *also* directly mutate the class. I think it will probably work fine?

However, It might be cleaner to do something like:

```python
class HasEvents:
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.Event = ...
        return cls

    @declared_attr
    def events(cls):
         return orm.relationship(cls.Event, cascade="all, delete-orphan", lazy=True)
```

You might also move some of the dynanism up a level:


```python
class Event(AbstractConcreteBase):
    tag = Column(String, nullable=False)
    time = Column(DateTime, nullable=False, server_default=sql.func.now())
    ip_address = Column(String, nullable=False)
    additional = Column(JSONB, nullable=True)

    @declared_attr
    def __tablename__(cls):
       "_".join([cls.__name__.removesuffix("Event").lower(), "events"])

    @declared_attr
   def __mapper_args__(cls):
        return {'polymorphic_identity': cls.__name__,
                'concrete': True} if cls.__name__ != "Event" else {}


 
class HasEvents:
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.Event = type(f"{cls.__name__}Event", (Event, db.Model), dict(source_id=..., source=...))
        return cls

    @declared_attr
    def events(cls):
         return orm.relationship(cls.Event, cascade="all, delete-orphan", lazy=True)

```

This is all completely untested, and you may be able to even do something like:

```python
class Event(AbstractConcreteBase):

    tag = Column(String, nullable=False)
    time = Column(DateTime, nullable=False, server_default=sql.func.now())
    ip_address = Column(String, nullable=False)
    additional = Column(JSONB, nullable=True)

    @declared_attr
    def __tablename__(cls):
       "_".join([cls.__name__.removesuffix("Event").lower(), "events"])

    @declared_attr
   def __mapper_args__(cls):
        return {'polymorphic_identity': cls.__name__,
                'concrete': True} if cls.__name__ != "Event" else {}

    @declared_attr
    def source_id(cls):
        return  source_id=Column(
                    UUID(as_uuid=True),
                    ForeignKey(
                        "%s.id" % cls._parent_class.__tablename__,
                        deferrable=True,
                        initially="DEFERRED",
                    ),
                    nullable=False,
                )
    @declared_attr
    def source(cls):
        orm.relationship(cls._parent_class)

    def __init_subclass__(cls, /, parent_class, **kwargs):
        super().__init_subclass__(cls, **kwargs)
        cls._parent_class = parent_class
        return cls


 
class HasEvents:
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.Event = type(f"{cls.__name__}Event", (Event, db.Model), dict(source_id=..., source=...))
        return cls

    @declared_attr
    def events(cls):
         return orm.relationship(cls.Event, cascade="all, delete-orphan", lazy=True)
```

Feel free to ignore this of course, just feels a little easier to read and follow what's going on.

dstufft (approved):
One comment, this looks fine as is but the comment above may make it cleaner. Either way it looks fine to me.

warehouse/events/models.py
di: I was able to get the 3rd option working: ef24030

(This also simplified the migration a bit)

warehouse/events/models.py
```diff
@@ -0,0 +1,80 @@
...
+            ),
+            nullable=False,
+        )
+
+    @declared_attr
+    def source(cls):
+        return orm.relationship(cls._parent_class)
+
+    def __init_subclass__(cls, /, parent_class, **kwargs):
+        # super().__init_subclass__(cls, **kwargs)
```
dstufft, line 55: This line probably wants to be uncommented or deleted.

dstufft:
👍

di:
@ewdurbin I'll let you merge whenever you're free to babysit this.

ewdurbin:
🖖🏼 20.778 month old PR 🚢

PR #8358, opened 2020-08-01
Autofocus project search bar

carlbordum:
Almost everytime I visit https://pypi.org/ my intention is to search. This change would save me a second or two :-)


di:
Hi @carlbordum! Thanks for this PR. This feature is not desirable from an accessibility perspective.

This topic has been extensively discussed and researched, see #6088. 

The feature in #6099 is another way to consider contributing.

PR #8393, opened 2020-08-07
Refactor BigQuery sync task to optimize memory consumption

patelneel55:
This PR is one last attempt at optimizing the `sync_bigquery_release_files` task. Instead of creating new variables for every computation, these changes minimize the memory footprint of the variables and removes variables from memory if not required. Reducing memory footprint did increase execution time but not by a considerable amount.

Expected usage for 3 million release files,
Peak memory usage for variables: ~50-60MB
Task execution time: ~1 min


di:
This task is currently disabled, so we don't need this.

PR #8404, opened 2020-08-10
Mark implied non-nullable fields as non-nullable

di:
This sets some columns of `release_files` to be `NOT NULL` where a) the production DB has no null values and b) we sort of imply these will not be null via how we use them in the codebase.

Towards #14403.


di:
```
warehouse=> SELECT count(*)
FROM release_files
WHERE python_version IS NULL
   OR packagetype IS NULL
   OR filename IS NULL
   OR size IS NULL
   OR upload_time IS NULL;
 count
-------
     0
(1 row)
```

```
warehouse_test=> SELECT count(*)
FROM release_files
WHERE python_version IS NULL
   OR packagetype IS NULL
   OR filename IS NULL
   OR size IS NULL
   OR upload_time IS NULL;
 count
-------
     0
(1 row)
```

ewdurbin (approved):

PR #8444, opened 2020-08-17
pip install <foo> quote <foo>

### LINKED ISSUE #8426: Copyable 'pip install <foo>' line should likely quote <foo> in some cases
**Describe the bug**
PyPI shows a copyable `pip install <foo>` command line, and includes the version when visiting a particular release. (Though there's some discussion in #1563 and #3006 about improving or dropping it).

For a release containing a [`PEP 440 Epoch`](https://www.python.org/dev/peps/pep-0440/#version-epochs), Warehouse shows:

Screen Shot 2020-08-13 at 19 11 54

The line there is likely not something someone wants to copy paste.

In various shells (zsh, bash) the `!` there will trigger history completion or the like (so `3!2` will complete in the history line number 2 in the shell history file).

**Expected behavior**
`<foo>` should be quoted, at least in the above case, perhaps others, though I can't think of any.

**To Reproduce**
https://pypi.org/project/composition/8!6.75309/

**My Platform**
Firefox 80.0b2 (64-bit), macOS

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypa/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

imsahil007:
Fixes #8426.

### DIFF
--- a/warehouse/templates/packaging/detail.html
+++ b/warehouse/templates/packaging/detail.html
@@ -82,7 +82,11 @@
     {% set project_version = "==" + release.version %}
   {%- endif -%}
 
-  pip install{{ index_url }} {{ release.project.name }}{{ project_version }}
+  {%- if release.version.epoch -%}
+    pip install{{ index_url }} '{{ release.project.name }}{{ project_version }}'
+  {%- else -%}
+    pip install{{ index_url }} {{ release.project.name }}{{ project_version }}
+  {%- endif -%}
 {%- endmacro -%}
 
 {% block title %}{{ release.project.name }}{% endblock %}
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -249,7 +249,7 @@ msgstr ""
 #: warehouse/templates/manage/release.html:171
 #: warehouse/templates/manage/releases.html:142
 #: warehouse/templates/manage/releases.html:175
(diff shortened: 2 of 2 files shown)


warehouse/templates/packaging/detail.html
```diff
@@ -77,12 +77,15 @@ <h3 class="package-snippet__title">
   {%- if test_environment -%}
     {% set index_url = " -i https://test.pypi.org/simple/" %}
   {%- endif -%}
-
   {%- if request_route == "packaging.release" -%}
     {% set project_version = "==" + release.version %}
   {%- endif -%}
 
-  pip install{{ index_url }} {{ release.project.name }}{{ project_version }}
+  {%- if '!' in release.version -%}
```
di, line 84: The `Version` class has an [`epoch` property https://github.com/pypa/packaging/blob/2a883ca86300bab76ad6296860797907b3432871/packaging/version.py#L348-L352] that can be used for this instead.

di:
Thanks!

imsahil007:
> Thanks!

You're welcome!

PR #8463, opened 2020-08-23
Add help for pasting password in Windows

bhrutledge:
Following up on https://github.com/pypa/twine/issues/671#issuecomment-674510081 and the subsequent discussion. In short, I think tokens are causing more folks to copy/paste instead of typing by hand, but that doesn't work great in the Windows Command Prompt. It seemed worth documenting, and this felt like an appropriate location.


bhrutledge:
@di Can you take a look at this, both for the content, and the pending Travis check?

PR #8533, opened 2020-09-09
Fixes #4703 Implement API for obtaining prohibited names

### LINKED ISSUE #4703: Publish a list of malicious packages that have been taken down
Users who may have possibly installed malicious packages don't have insight into what packages have been taken down by PyPI administrators.

PyPI should publish both a human-readable and machine-readable (API) list of malicious packages that have been taken down. Ideally the human-readable list would be sortable by package name, or by the date it was created/taken down.

**Additional context**
Feature request to automatically uninstall packages via this API in `pip`: https://github.com/pypa/pip/issues/5777

zooba:
This is very much a work in progress for implementing #4703, but I wanted to get feedback on the direction as I go.

In particular, the API route can't (trivially) go under `/api/prohibited_project_name` because that's a lookup for a package by that name. I replaced the slash with an underscore so I could test it, but open to other ideas on where to put it.

At least in my dev environment, putting it at `/admin/prohibited_project_name/json` seems to work okay. But totally open to other suggestions.

### DIFF
--- a/warehouse/legacy/api/json.py
+++ b/warehouse/legacy/api/json.py
@@ -19,7 +19,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from warehouse.cache.http import cache_control
 from warehouse.cache.origin import origin_cache
-from warehouse.packaging.models import File, Project, Release
+from warehouse.packaging.models import File, ProhibitedProjectName, Project, Release
 
 # Generate appropriate CORS headers for the JSON endpoint.
 # We want to allow Cross-Origin requests here so that users can interact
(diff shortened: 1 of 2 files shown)


zooba:
Ping @di

zooba:
Going to move it under the admin section where the existing route is, but leave it unauthenticated and just return the list as JSON.

ewdurbin:
we really shouldn't place this under `/admin` as that is explicitly _not_ cached.

ewdurbin:
Speaking of caching, we'll want to instrument some way of purging this URL when a new project is prohibited or removed from the prohibited list.

zooba:
If we move it under a route that Fastly will cache, will they respect expiry headers? This is an information API, so there's really no harm in it being up to a day behind.

zooba:
Also, any suggestions/preferences for where to route it? Under /api/{project_name} can't work (unless we go ahead and ban the package name matching the API... 👀) None of the other existing paths looked like good options.

firefly9966 (approved):

PR #8562, opened 2020-09-15
Add check_if_macaroon_exists

ewjoachim:
Addresses the revert of #7124

Root causes:
- There was a test for the error branch of Verifier.verify but no test for the success case
- It was not very clear whether the method was supposed to
  - return True or False
  - return True or raise (:point_left:  what it does)
  - pass or raise
- The unit test mocks a lot, and there's no functional check that checks the whole chain.

This PR:
- Extracts all the code that touches macaroon from #7124 
- Fixes the problem: the function `Verifier.verify` needs to return True if it's ok, not raising is not enough.
- Fixes the test.

This PR doesn't:
- Address that https://github.com/pypa/warehouse/blob/faa49c6dbdf45f64648d96ae367b0a9014fbc2ac/warehouse/macaroons/services.py#L118 is dead code: `verify()` either returns `True` or raise, there's no `False`.  This is confusing and should be address too. I can adress it here or in a separate PR.

### DIFF
--- a/warehouse/macaroons/caveats.py
+++ b/warehouse/macaroons/caveats.py
@@ -33,7 +33,7 @@ class Caveat:
 
 
 class V1Caveat(Caveat):
-    def verify_projects(self, projects):
+    def verify_projects(self, projects) -> None:
         # First, ensure that we're actually operating in
         # the context of a package.
         if not isinstance(self.verifier.context, Project):
@@ -43,13 +43,13 @@ class V1Caveat(Caveat):
 
         project = self.verifier.context
         if project.normalized_name in projects:
-            return True
+            return
 
         raise InvalidMacaroon(
             f"project-scoped token is not valid for project '{project.name}'"
         )
 
-    def verify(self, predicate):
+    def verify(self, predicate) -> None:
         try:
             data = json.loads(predicate)
         except ValueError:
@@ -64,13 +64,13 @@ class V1Caveat(Caveat):
 
         if permissions == "user":
             # User-scoped tokens behave exactly like a user's normal credentials.
-            return True
+            return
 
         projects = permissions.get("projects")
         if projects is None:
             raise InvalidMacaroon("invalid projects in predicate")
 
-        return self.verify_projects(projects)
+        self.verify_projects(projects)
 
 
 class Verifier:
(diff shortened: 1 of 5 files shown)


warehouse/macaroons/caveats.py
```diff
@@ -81,11 +81,14 @@ def __init__(self, macaroon, context, principals, permission):
...
-    def verify(self, key):
+    def verify(self, key: str) -> bool:
         self.verifier.satisfy_general(V1Caveat(self))
+        self.verify_signature(key=key)
+        return True
 
+    def verify_signature(self, key: str) -> None:
         try:
-            return self.verifier.verify(self.macaroon, key)
+            self.verifier.verify(self.macaroon, key)
```
di, line 91:
```suggestion
            return self.verifier.verify(self.macaroon, key)
```
I'll agree that the pymacaroons API doesn't make a ton of sense here, but in case a future version begins returning `False` here, I think we should continue returning the result.

warehouse/macaroons/caveats.py
```diff
@@ -81,11 +81,14 @@ def __init__(self, macaroon, context, principals, permission):
         self.permission = permission
         self.verifier = pymacaroons.Verifier()
 
-    def verify(self, key):
+    def verify(self, key: str) -> bool:
         self.verifier.satisfy_general(V1Caveat(self))
+        self.verify_signature(key=key)
+        return True
```
di, line 87:
```suggestion
        return self.verify_signature(key=key)
```

warehouse/macaroons/caveats.py
merwok: Isn’t it more pythonic to return None for success, raise exception for problem?

di: Sure, but in that case we'd still need to change this to raise an exception if `verify` ever returns false. Either works for me.

merwok: Right, and one would hope that the project maintainer would have suitable version bump and changelog entries in that case!

ewjoachim: Yep, I’d much rather raise if False. "True or raise" feels to me
Line a receipe for a misunderstanding (like what happened there)

ewjoachim: I will be adding code to raise if False, but in parallel, I've opened https://github.com/ecordell/pymacaroons/issues/55 to try and officialize "raise or pass".

(I admit I'm not a huge fan of writing code imagining potential dependency API breakages, and I'm not a fan that in order to test this code we **will** have to monkeypatch and make the stub do things that cannot happen in real life. For all intended purposes we introducing new dead code :/ )

(Except if you think it's not worth it, I'll happily celebrate a change of mind :) )

ewjoachim:
Fixed. I've removed dead code, enforced 'pass or raise', and added a special branch if case pymacaroon ever attempts to return False.

di:
@ewjoachim You've got failing tests here.

ewjoachim:
damn :(

ewjoachim:
fixed :)

di:
Thanks!

PR #8580, opened 2020-09-17
Limit password gauge latency

bodograumann:
Currently the complete password is given to zxcvbn to estimate it’s strength and display that in the gauge.
If the password is sufficiently long, this takes an exponential amount of time.
For that reason it is recommended to truncate the password to a sensible length when gauging the strength.
Cf. https://github.com/dropbox/zxcvbn#runtime-latency


di:
Thanks!

Issue #8487, opened 2020-08-28
Pre-generate and serve simple index metadata


woodruffw:
**What's the problem this feature will solve?**

As part of the TUF rollout (#7488), we will need to store hashes for the simple indices that `pip` and other resolvers use.

These indices are currently generated dynamically from a template when requested, making that difficult. Instead, they should be generated once per relevant event (file upload/release) and stored somewhere (probably GCS). Stale indices should _not_ be deleted from the store, as the TUF metadata may still refer to them.

cc @ewdurbin @dstufft

di:
For those that aren't as familiar with TUF, a few additional questions about this:

* what hashes should we store?
* do we need to include other metadata with the file as well? timestamp?
* do we need this for `/simple` or just `/simple/projectname`?
* how would TUF metadata refer to an old index?

woodruffw:
Yep! Thanks for the clarifying questions.

> * what hashes should we store?

TUF is using BLAKE2 for the other target metadata (i.e., actually distribution packages), so it probably makes sense to use it here as well.

> * do we need to include other metadata with the file as well? timestamp?

I don't believe so; I _think_ just the file itself should be sufficient. @jku may be able to correct me here, if I'm missing something.

> * do we need this for `/simple` or just `/simple/projectname`?

Just `/simple/projectname`, for TUF purposes. My understanding (again, please correct me if I'm wrong!) is that no resolvers currently use `/simple`, and TUF won't be using it whatsoever.

> * how would TUF metadata refer to an old index?

This is probably the knottiest part. My first thought for this is that `/simple/projectname` should be mapped to `/simple/{hash}-projectname`, where `{hash}` is the BLAKE2 content hash. The TUF metadata would only ever refer to the `{hash}-projectname` variant, ensuring that we always fetch a version of the index that's consistent with the other target metadata.

jku:
Yeah this all seems correct to me. The TUF metadata for a project index will look roughly like this:
```
   "sampleproject": {
    "custom": {},
    "hashes": {
     "blake2b": "7a8bc0d0a15f6289b184f86a56d398467fd465c6293535182ba0f6cc2f04e703",
    },
    "length": 3080
   }
```
[...]

PR #8586, opened 2020-09-18
Create utility to render and store transactional snapshot of simple detail page for a project

ewdurbin:
Needed for #8487.

Exposes `packaging.utils.render_simple_detail` for use in TUF integration.

```
def render_simple_detail(project, request, store=False):
```

Returns `(content_hash, simple_detail_path)`

The `store` flag can be used to control wether the resulting rendered/hashed simple detail page is pushed to our storage service.

### DIFF
--- a/warehouse/legacy/api/simple.py
+++ b/warehouse/legacy/api/simple.py
@@ -49,6 +49,20 @@ def simple_index(request):
     return {"projects": projects}
 
 
+def _simple_detail(project, request):
+    # Get all of the files for this project.
+    files = sorted(
+        request.db.query(File)
+        .options(joinedload(File.release))
+        .join(Release)
+        .filter(Release.project == project)
+        .all(),
+        key=lambda f: (parse(f.release.version), f.filename),
+    )
+
+    return {"project": project, "files": files}
+
+
 @view_config(
     route_name="legacy.api.simple.detail",
     context=Project,
@@ -74,14 +88,4 @@ def simple_detail(project, request):
     # Get the latest serial number for this project.
     request.response.headers["X-PyPI-Last-Serial"] = str(project.last_serial)
 
-    # Get all of the files for this project.
-    files = sorted(
-        request.db.query(File)
-        .options(joinedload(File.release))
-        .join(Release)
-        .filter(Release.project == project)
-        .all(),
-        key=lambda f: (parse(f.release.version), f.filename),
-    )
-
-    return {"project": project, "files": files}
+    return _simple_detail(project, request)
--- a/warehouse/legacy/api/simple.py
+++ b/warehouse/legacy/api/simple.py
@@ -18,7 +18,7 @@ from sqlalchemy import func
 from warehouse.cache.http import cache_control
 from warehouse.cache.origin import origin_cache
 from warehouse.packaging.models import JournalEntry, Project
-from warehouse.packaging.utils import _simple_detail, render_simple_detail
+from warehouse.packaging.utils import _simple_detail
 
 
 @view_config(
@@ -72,7 +72,5 @@ def simple_detail(project, request):
 
     # Get the latest serial number for this project.
     request.response.headers["X-PyPI-Last-Serial"] = str(project.last_serial)
-
-    render_simple_detail(project, request, store=True)
 
     return _simple_detail(project, request)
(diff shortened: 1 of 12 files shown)


ewdurbin:
@di can I get a review and sanity check here before I go about implementing tests?

ewdurbin:
cc @woodruffw

warehouse/legacy/api/simple.py
```diff
@@ -74,14 +73,6 @@ def simple_detail(project, request):
...
-    # Get all of the files for this project.
-    files = sorted(
-        request.db.query(File)
-        .options(joinedload(File.release))
-        .join(Release)
-        .filter(Release.project == project)
-        .all(),
-        key=lambda f: (parse(f.release.version), f.filename),
-    )
+    render_simple_detail(project, request, store=True)
```
ewdurbin, line 76: This is only inline here at the moment to actually trigger a render/store.

woodruffw: Makes sense! It looks like the current implementation unconditionally pushes to the object store when `store=True` on every request; is that intended (vs. checking whether the stored version is stale first)?

woodruffw (approved):
LGTM! One tiny question about when simple detail files are being stored 🙂

warehouse/legacy/api/simple.py
ewdurbin: My assumption is that `render_simple_detail` would be called primarily in response to uploads, thus would only be retouched if a task retries. So with that in mind and given the fact that these are immutable and for all intents and purposes impossible to collide, it seemed like it wouldn't be a concern.

Is there something lurking I need to be concerned with 😅. Given that we *do* generate these primarily based on user supplied inputs... I guess a very clever and lucky attacker could force an overwrite of an existing index page?

woodruffw: > Given that we _do_ generate these primarily based on user supplied inputs... I guess a very clever and lucky attacker could force an overwrite of an existing index page?

I _think_ this is safe, at least assuming that an attacker can't cause a conflict with `project.normalized_name` and that there isn't a feasible preimage attack on BLAKE2b. I think both of those assumptions hold, so it should be fine 🙂

ewdurbin:
Ready for review @di and @woodruffw.

di (approved):

warehouse/packaging/services.py
```diff
@@ -58,6 +58,38 @@ def store(self, path, file_path, *, meta=None):
...
+    @classmethod
+    def create_service(cls, context, request):
+        return cls(request.registry.settings["simple.path"])
+
+    def get(self, path):
+        return open(os.path.join(self.base, path), "rb")
+
+    def store(self, path, file_path, *, meta=None):
+        destination = os.path.join(self.base, path)
+        print(destination)
```
kairoaraujo, line 86: I think using it was used for debugging.

warehouse/packaging/utils.py
```diff
@@ -0,0 +1,77 @@
...
+    content_hash = content_hasher.hexdigest().lower()
+
+    simple_detail_path = (
+        f"{project.normalized_name}/{content_hash}.{project.normalized_name}.html"
+    )
+
+    if store:
+        storage = request.find_service(ISimpleStorage)
+        with tempfile.NamedTemporaryFile() as f:
+            f.write(content.encode("utf-8"))
```
kairoaraujo, line 57: I think missing to flush the temporary file here. Otherwise, the content will be empty in the following HTML files.
```suggestion
            f.write(content.encode("utf-8"))
            f.flush()
```

kairoaraujo (commented):
I am working on the following implementation PR on top of this PR (related to PEP 458, TUF).
I noticed that the simple HTML files generated by ``render_simple_details`` are empty (see my comment on ``warehouse/packaging/utils``).

warehouse/packaging/utils.py
```diff
@@ -0,0 +1,83 @@
...
+        .join(Release)
+        .filter(Release.project == project)
+        .all(),
+        key=lambda f: (parse(f.release.version), f.filename),
+    )
+
+    return {"project": project, "files": files}
+
+
+def render_simple_detail(project, request, store=False):
```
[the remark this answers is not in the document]
ewdurbin: Since I believe TUF needs the hash/path, TUF is the intended invoker of this method and it shouldn't be invoked anywhere yet.

PR #8591, opened 2020-09-21
Revert "Add check_if_macaroon_exists (#8562)"

### LINKED ISSUE #8565: Invalid API Token: invalid macaroon!r for url
**Describe the bug**
Possibly related to #6259
When using the same token generated from pypi a couple of months ago, now I got the error:
```
HTTPError: 403 Client Error: Invalid API Token: invalid macaroon!r for url: https://upload.pypi.org/legacy/
```
I figured it was based on an old key and re-generated a token, but I get the same error.

I verified that the version of `twine` did not change between the successful execution and the failure.

**Expected behavior**
No error on upload.

**To Reproduce**

TWINE_USERNAME=`__token__`
TWINE_PASSWORD=`<token generated at pypi.org>`

twine upload ...

**My Platform**
```
Build dist: xenial
Runtime kernel version: 4.15.0-1077-gcp
```
**Additional context**
https://travis-ci.com/github/pyproj4/pyproj-wheels/jobs/384840987

di:
This reverts commit 64b393c9f771686b7d5ce518eee12063b48ed31f.

Fixes #8565.

### DIFF
--- a/tests/unit/macaroons/test_caveats.py
+++ b/tests/unit/macaroons/test_caveats.py
@@ -55,7 +55,7 @@ class TestV1Caveat:
         caveat = V1Caveat(verifier)
         predicate = '{"permissions": "user", "version": 1}'
 
-        caveat(predicate)
+        assert caveat(predicate) is True
 
     def test_verify_project_invalid_context(self):
         verifier = pretend.stub(context=pretend.stub())
@@ -90,9 +90,9 @@ class TestV1Caveat:
         project = ProjectFactory.create(name="foobar")
         verifier = pretend.stub(context=project)
         caveat = V1Caveat(verifier)
-        predicate = {"version": 1, "permissions": {"projects": ["foobar"]}}
 
-        caveat(json.dumps(predicate))
+        predicate = {"version": 1, "permissions": {"projects": ["foobar"]}}
+        assert caveat(json.dumps(predicate)) is True
 
 
 class TestVerifier:
@@ -108,7 +108,7 @@ class TestVerifier:
         assert verifier.principals is principals
         assert verifier.permission is permission
 
-    def test_verify_raises(self, monkeypatch):
+    def test_verify(self, monkeypatch):
         verify = pretend.call_recorder(
             pretend.raiser(MacaroonInvalidSignatureException)
         )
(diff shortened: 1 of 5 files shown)


ewjoachim:
But I had reduced the scope, increased coverage and tested it manually 😫

ewjoachim:
Ok, found it. I think the best way to explain how the bug happened is simply describe it :)

`DatabaseMacaroonService` has a `verify()` method that instantiates a `Verifier` object called `verifier`  and call its `verify()` method.
That `Verifier` class has a `verifier` attribute which point to an instance of PyMacaroon's `Verifier` class. PyMacaroons' `Verifier` is given a `V1Caveat` object which has a method named `verify`. So when I saw in our `Verifier` a call to `self.verifier.verify()`, but it was actually PyMacaroonmy mind jumped to the `V1Caveat.verify()`, instead of following the path through PyMacaroon's `verify`.

But the crucial difference is that PyMacaroons expects its caveat to return "True or False" (well actually it calls `bool(sum())` on the results but I think they're designed to be booleans), whereas PyMacaroon's `verify()` is "pass or raise". But this was obscured by the fact our caveat could never return False: it either returned True or raised. I though the raising part was the important one, but actually no: raising in the caveat is a bypass of the API, we're supposed to return False, and PyMacaroon will be raising for us. When we raise, it's unexpected for PyMacaroon, so it bubbles up.

So when I replaced `return True` with `return`, this meant the result of the caveat, instead of being a boolean, became a None, and as I mentioned, PyMacaroon groups boolean by doing a `sum()`, which raised  a `ValueError`. Because PyMacaroon doesn't guarantee a base Exception class when calling `verify`, we have to wrap it in a `try: except Exception:` which backfired, catching the `ValueError` as if it was PyMacaroon's way of saying "the macaroon is invalid" when, for real, it was the codebase crashing because of an unexpected bug.

So root causes:
- I didn't understand the code properly
- Because of that, I removed the return True, instead of replacing the raise with `return False`
- `ERR_TOO_MANY_THINGS_CALLED_VERIFY`
- In the original code, V1Caveat raised instead of returning False
- PyMacaroon uses `bool(sum())` instead of `any()`
- PyMacaroon's caveats are "True and False", but PyMacaroon's `verify` is "True or raise"
- (PyMacaroon is not much undocumented)
- `except Exception` in our code because of https://github.com/ecordell/pymacaroons/issues/51

goanpeca:
Hi :)

Thanks for looking into it!

Has this been released to production?

I am trying to run this on a Github Action and I am getting:

```
Uploading spyder_remote_client-0.0.2-py3-none-any.whl

  0%|          | 0.00/17.3k [00:00<?, ?B/s]
 46%|████▌     | 8.00k/17.3k [00:00<00:00, 43.5kB/s]
HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/
Invalid API Token: InvalidMacaroon('invalid macaroon signature')
100%|██████████| 17.3k/17.3k [00:00<00:00, 54.3kB/s]
Content received from server:
<html>
 <head>
  <title>403 Invalid API Token: InvalidMacaroon('invalid macaroon signature')</title>
 </head>
 <body>
  <h1>403 Invalid API Token: InvalidMacaroon('invalid macaroon signature')</h1>
  Access was denied to this resource.<br/><br/>
Invalid API Token: InvalidMacaroon(&#x27;invalid macaroon signature&#x27;)


 </body>
</html>
```

di:
@goanpeca Please open a new issue, and let us know if you're using a API token that was previously working, or if this is the first time you're attempting this.

PR #8615, opened 2020-09-28
Implement 'latest' API and release-view endpoints 

### LINKED ISSUE #4663: API for fetching latest and stable versions

Tools  could use simple lookup logic for checking their latest and stable versions. The way it is currently done in `pip` is crazy:

https://github.com/pypa/pip/blob/35d51f1f405f6f31d68c800e73307686ec029254/src/pip/_internal/utils/outdated.py#L104-L118

It should be as simple as fetching https://pypi.org/pypi/pip/latest/json or https://pypi.org/pypi/pip/stable/json endpoints.

bskinn:
Implements and closes #4663. See [here](https://github.com/pypa/warehouse/issues/4663#issuecomment-695012184) for the semantics of the three new endpoints.

Per [here](https://github.com/pypa/warehouse/issues/4663#issuecomment-701972862) and [here](https://github.com/badges/shields/issues/4692), there was interest expressed in also having similar `latest` endpoints available for release views on the web UI. This PR has been expanded beyond #4663 to include all three of the analogous release-view redirects, at `/project/{name}/latest`, etc.

As the release calculations for the new `latest` JSON and web UI endpoints use the `.latest_version` property that was previously defined on `Project`, their implementations are coupled to the behavior used to serve the `/project/{name}/` endpoint. 

Conversely, the `latest-stable` and `latest-unstable` endpoints for both the JSON and web UI are based on new functionality, and are not coupled to any other behaviors.

- [x] Implementation
- [x] Tests
- [x] Docs
- ~Changelog/Announcement (?)~ N/A


di:
Nice! Just an FYI @bskinn, we'll hold off on giving this a full review until you take it out of draft status, but if you want input at any point feel free to @-mention me with specific questions.

bskinn:
Yep, @di, that was my intention with draft status... no need to take maintainer time/attention until I've got it fleshed out. :+1:

bskinn:
Ok, @di, I think this is ready for review on the implementation and tests. I'll augment the API docs as part of this PR, too--still TODO for now. (If you want to wait to review until I've finished the docs update, let me know.)

One big question: with the arc I took on the code, I ended up implementing the logic directly in `warehouse/legacy/api/json.py`, and not as methods on the `Project` model. Would you like me to refactor it over to such methods? I could see `Project.latest_stable_version` and `Project.latest_unstable_version` as potentially being useful elsewhere; or, they could just be YAGNI at this point.

bskinn:
Looks like there *is* existing interest in having access to behaviors that would use `latest-unstable` (at least) in addition to `latest`. I'm inclined to refactor the lookup into `Project`.

https://github.com/badges/shields/issues/4692

https://github.com/pypa/warehouse/issues/4663#issuecomment-701972862

bskinn:
Ok, @di, this is ready for review. Code/tests/docs are all completed for both the JSON and web UI `latest` endpoints. 

I couldn't find anywhere to put a news snippet (as per, e.g., towncrier), or to include a CHANGELOG entry. Please advise if that's something I need to do, and how.

bskinn:
Hi, @di -- I'm still interested in shepherding this PR through, if the features remain of interest. Please let me know if there's anything I can do on my end to facilitate the process.

di:
Thanks @bskinn, appreciate your patience here, this is still of interest.

ewjoachim:
I might be confused but while redirections work for a browser, it doesn't have a default behaviour for an API consumer, so the idea is that the API client will implement handling redirection ? Or is this done with a specific client in mind ?
Having 3XX redirects in an API is not a practice I often see.

While I think the approach is fine for the HTML pages, I think we need to approach it differently for the JSON APIs: the redirect should be server-side (as in, the `latest` view should call the normal view function with the proper arguments and return whatever this function returns). Of course, it might be cleaner to split the `json_release` view into 2 parts: one part that is a pyramid view and calls a simple Python function and one part that is a simple python function and does all the implementation. This will avoid going through the decorator code twice.

bskinn:
@ewjoachim:

> I might be confused but while redirections work for a browser, it doesn't have a default behaviour for an API consumer, so the idea is that the API client will implement handling redirection ? Or is this done with a specific client in mind ?
Having 3XX redirects in an API is not a practice I often see.

I've almost certainly done it incorrectly. :-) I have almost no experience writing web apps, so I greatly appreciate the critique. Manually testing the API endpoints using `links` with the `warehouse` dev server running seemed to work ok...but I guess that might be because it was still a browser handling the redirect.

> While I think the approach is fine for the HTML pages, I think we need to approach it differently for the JSON APIs: the redirect should be server-side (as in, the latest view should call the normal view function with the proper arguments and return whatever this function returns). Of course, it might be cleaner to split the json_release view into 2 parts: one part that is a pyramid view and calls a simple Python function and one part that is a simple python function and does all the implementation. This will avoid going through the decorator code twice.

Pretty sure I have a good idea what you're aiming at here. I'll give a go at revising.

ewjoachim:
I realize I've forgotten an important part: thank you a *lot* and it's a very nice PR you submitted here :)

> I've almost certainly done it incorrectly. :-) I have minimal experience writing web apps, so I greatly appreciate the critique. 

Then your work is all the more impressive, and thank you for stepping out of your confort zone for the project :)

bskinn:
Thanks! It really wasn't too hard -- the project is so well laid out & designed, that really all I had to do was copy what I needed from what was already there, and innovate (too much on the API endpoints, it seems :-D) on the relatively few novel bits.

It's been a great way to get my hands dirty inside a substantial web app...I have a much better idea now of all the sorts of things I have yet to learn....

bskinn:
Ah -- @ewjoachim, now that I look at it, I implemented those redirects in what I thought was analogy to the redirects already in place to handle URLs with/without trailing slashes; see, e.g., [here](https://github.com/pypa/warehouse/blob/fb550767791e40f0435dcc83b6885516b970ea5f/warehouse/legacy/api/json.py#L201-L215).

Chrome and Python `requests` both seem able to follow the `HTTPMovedPermanently` redirect ok:

```
>>> import requests as rq
>>> resp = rq.get("https://pypi.org/pypi/requests/json/")    # With slash
>>> resp
<Response [200]>
>>> resp.url
'https://pypi.org/pypi/requests/json'      # No slash
```

Is this sort of redirect ok for something as minimal as slash-handling, but improper for the more substantial redirects of this PR?

ewjoachim:
Hm, I might be wrong, but the way I understand RESTful API design is to use all the expressivity of the HTTP standard to return information that is not tied to how browsers typically use it. For example, a 401 in a browser cause the display of a login/password modal. In a RESTful API, a 401 only conveys the information that the request was Unauthorized but not the behaviour of displaying a login/password modal, that is left to the Client to decide and implement.

Similarily, I'm not sure a 302 response should be interpreted by the HTTP layer directly, but left to the client to decide. One thing that tells me something is fishy here is that 3XX answers have a body (that typically no-one reads because we follow redirects instead) and in a RESTful fashion, we should expect this body to be a JSON payload, and it actually is plain text, and kinda no-one cares, which would tell me that this approach isn't that much RESTful (arguably)

But what I say could well be very "theological" and theoretical, and I'm perfectly fine if other people step in and say that they don't agree. It's purely my interpretation at this point.

> Is this sort of redirect ok for something as minimal as slash-handling, but improper for the more substantial redirects of this PR?

I think the main difference is that the / redirect (which, to me, have the same problem) don't exist as a way to extract information from the API but as a very pragmatic courtesy to developpers who mistyped the URL. It's either this or a 404, and a 3XX redirect is a bit more useful (yet actually encourages mispractice). I wouldn't have done it and I'm not a fan that it's there, but I understand why it was put there (it doesn't really hurt) and it's not something that is required to handle in order to work with the API. We could even argue it's not a part of the API itself.

Does that make my opinion (and I insist that this is merely an opinion) clearer?

bskinn:
Nod, definitely clearer -- it's essentially what I had assumed (minus all of the details of what's RESTful; I have no expertise there), where the slash-fix redirect is "less improper/unRESTful" because it's effectively just a typo-fix, but what I've implemented contains "semantically-significant redirects" and is more fraught.

tests/unit/legacy/api/test_json.py
```diff
@@ -206,6 +191,142 @@ def test_normalizing_redirects(self, db_request):
         assert resp.headers["Location"] == "/project/the-redirect"
 
 
+@pytest.fixture
+def check_json_release(monkeypatch):
```
bskinn, line 195: Would this fixture be better sited in `/conftest.py`, next to the other fixtures I've added?

Is this the 'right way' to go about this?

tests/unit/packaging/test_views.py
```diff
@@ -333,3 +337,93 @@ def test_edit_project_button_returns_project(self):
         assert views.edit_project_button(project, pretend.stub()) == {
             "project": project
         }
+
+
+class TestProjectLatestRedirects:
+    def check_release(self, db_request, project, release, endpoint):
+        db_request.route_path = pretend.call_recorder(
```
bskinn, line 344: Should I refactor this monkeypatch to happen within the context of an actual use of `monkeypatch` fixture? If so, should it be relocated to `conftest.py` (and probably renamed)?

Is it possible to decorate a class method as a fixture?

warehouse/legacy/api/json.py
```diff
@@ -217,3 +217,120 @@ def json_release_slash(release, request):
...
+
+@view_config(
+    route_name="legacy.api.json.latest",
+    context=Project,
+    renderer="json",
+    decorator=_CACHE_DECORATOR,
+)
+def json_latest(project, request):
+    release = project.latest_version
+    version = release.version if release else None
```
bskinn, line 230: Stylistically/functionally, is this the best way to accommodate the possibility of `project.latest_version` being `None`?

warehouse/legacy/api/json.py
```diff
@@ -217,3 +217,120 @@ def json_release_slash(release, request):
...
+
+    if version is None:
+        return HTTPNotFound(headers=_CORS_HEADERS)
+
+    release = (
+        request.db.query(Release)
+        .filter(Release.project == project)
+        .filter(Release.version == version)
+        .first()
```
bskinn, line 239: Is `.first()` the best way to trim this to one result? Would it be faster to use, e.g., `.one()` or `.limit(1)`?

warehouse/legacy/api/json.py
```diff
@@ -217,3 +217,120 @@ def json_release_slash(release, request):
...
     )
+
+
+@view_config(
+    route_name="legacy.api.json.latest",
+    context=Project,
+    renderer="json",
+    decorator=_CACHE_DECORATOR,
+)
+def json_latest(project, request):
```
bskinn, line 228: As written, this function (and those analogous to it) make two database calls. The first is within the implementation of `project.latest_version`, and the second is a few lines below, when `release` is defined.

Would it be better from a performance or other perspective to figure out how to refactor this down to a single database access? (I'm not entirely sure if it's possible, though, given that the web view logic needs something different from the JSON view logic.)

bskinn (commented):
@ewjoachim, this is ready for your re-review. If you have the bandwidth, your feedback on my comments here would be hugely appreciated. Thanks!

warehouse/legacy/api/json.py
ewjoachim: `one` would raise if there are more than one result, which is not what you want.
`limit()` would return an iterable which is not what you want. There's no reason calling SQLAlchemy's function exactly for its main use-case would be suboptimal. `first` is absolutely fine, as far as I can tell.

Here's the implementation for first:
https://github.com/sqlalchemy/sqlalchemy/blob/201893e3f6f7f3bdc8b9c93cd25092cd6fa8cba3/lib/sqlalchemy/engine/result.py#L1476-L1487
Here's the implementation for one:
https://github.com/sqlalchemy/sqlalchemy/blob/201893e3f6f7f3bdc8b9c93cd25092cd6fa8cba3/lib/sqlalchemy/engine/result.py#L1501-L1511

ewjoachim: In this case, if there was 0 or multiple results, it would be a bug, so it could make sense to call `one()` (if we had mypy, it would complain that release might be None here). `limit(1)` wouldn't bring a lot to the table since you don't want an iterable. I think it makes the most sense to use the function whose Python API is most adapted to what you want and trust sqlalchemy to optimize it correctly.

tests/unit/packaging/test_views.py
bskinn: Looks like yes, it _is_ possible to decorate an instance method as a fixture. Definitely seems prudent to `monkeypatch` this mock. Will revise accordingly.

tests/unit/legacy/api/test_json.py
bskinn: Will be refactoring this into an instance-method fixture within the `TestJSONLatestReleases` class. Best to keep it tightly associated with these tests, I think.

warehouse/legacy/api/json.py
bskinn: Nod, I like that change; I'd looked at the docs but hadn't appreciated the raising behavior of `.one()`.

docs/api-reference/json.rst
```diff
@@ -264,3 +264,35 @@ Release
...
+
+    .. _api_json_latest:
+
+    There are three special ``<version>`` names that can be passed for any
+    ``<project_name>``, to obtain a `Release`_ JSON response for various flavors
+    of the latest available release for that project:
+
+    * ``/pypi/<project_name>/latest/json``
+
+        Redirects to the latest non-prerelease version of ``<project_name>``,
```
ewjoachim, line 277: Not a redirect anymore

tests/conftest.py
```diff
@@ -337,3 +338,49 @@ def monkeypatch_session():
...
     yield m
     m.undo()
+
+
+# Standardized dummy projects for testing version-search
+# behavior under different stable/prerelease circumstances.
+# In particular, created to support the 'latest' endpoints of
+# https://github.com/pypa/warehouse/pull/8615
+
+ProjectData = namedtuple("ProjectData", ["project", "latest_stable", "latest_pre"])
```
ewjoachim, line 348: Do you think we could `typing.NamedTuple` instead ? It's more readable (IMHO).

tests/conftest.py
```diff
@@ -337,3 +338,49 @@ def monkeypatch_session():
...
+
+
+@pytest.fixture(scope="function")
```
ewjoachim, line 351: IIRC `scope="function"` is the default value, and if there's no argument, you can omit parens.
So that could be 
```suggestion
@pytest.fixture
```

warehouse/legacy/api/json.py
ewjoachim: IMHO, 2 DB calls is still early optimization, if we're not on the critical path.

ewjoachim: IMHO, trying to remove 1 DB call if it has readability implications and we're not on the critical path is early optimization.

warehouse/legacy/api/json.py
```diff
@@ -217,3 +217,120 @@ def json_release_slash(release, request):
...
+    release = project.latest_version
+    version = release.version if release else None
+
+    if version is None:
+        return HTTPNotFound(headers=_CORS_HEADERS)
```
ewjoachim, line 233: Aswering https://github.com/pypa/warehouse/pull/8615/files#r618602220

We could do
```suggestion
    if release is None:
        return HTTPNotFound(headers=_CORS_HEADERS)
        
    version = release.version
```
I believe a release MUST have a non-None version.

warehouse/packaging/models.py
```diff
@@ -268,6 +270,37 @@ def latest_version(self):
             .first()
         )
 
+    @property
+    def latest_stable_version(self):
```
ewjoachim, line 274: What if those properties returned a `Release` object instead of just a few fields ? I believe it would simplify the View code.

warehouse/packaging/views.py
```diff
@@ -53,6 +57,53 @@ def project_detail(project, request):
...
+)
+def project_latest_stable(project, request):
+    release = project.latest_stable_version
+
+    if release:
+        return HTTPTemporaryRedirect(
+            request.route_path(
+                "packaging.release",
+                name=project.name,
+                version=project.latest_stable_version.version,
```
ewjoachim, line 86:
```suggestion
                version=release.version,
```

warehouse/packaging/views.py
```diff
@@ -53,6 +57,53 @@ def project_detail(project, request):
...
+@view_config(
+    route_name="packaging.project_latest_unstable",
+    context=Project,
+)
+def project_latest_unstable(project, request):
+    return HTTPTemporaryRedirect(
+        request.route_path(
+            "packaging.release",
+            name=project.name,
+            version=project.latest_unstable_version.version,
```
ewjoachim, line 102: what if `project.latest_unstable_version` is None ?

warehouse/packaging/views.py
```diff
@@ -53,6 +57,53 @@ def project_detail(project, request):
...
+@view_config(
+    route_name="packaging.project_latest",
+    context=Project,
+)
+def project_latest(project, request):
+    return HTTPTemporaryRedirect(
+        request.route_path(
+            "packaging.release",
+            name=project.name,
+            version=project.latest_version.version,
```
ewjoachim, line 69: What if `project.latest_version` is None ?

ewjoachim:
(I think I'm still a little unclear regarding the difference between accessing `https://pypi.org/project/somepackage/` vs `https://pypi.org/project/somepackage/latest`)

bskinn:
Thanks for the reviews, @ewjoachim! I'll work through them as soon as I can.

To answer your uncertainty: there currently is _no difference whatsoever_ between `project/{name}` and `project/{name}/latest`. Per the discussion at the original issue (starting at @di's comment [here](https://github.com/pypa/warehouse/issues/4663#issuecomment-695012184), and following), my implementation intentionally includes this redundancy. It could easily be eliminated; however, including it allows for the current `latest` behavior to be easily retained if the desired behavior for `project/{name}` changes in the future.

The redundancy might well be YAGNI, though, and I'm happy to strip the `latest` view and JSON endpoint if that's the decision.

ewjoachim:
Ok. I think there's a little difference though in that `myproject/latest` will redirect to `myproject/1.2.3` (if `1.2.3` is latest) which will display things a little differently (e.g. instructions at the top of the page will be `pip install myproject==1.2.3`. Would that be a good thing or not, I don't know.

warehouse/packaging/models.py
bskinn: Huh. Yeah, it looks like the test suite still passes fine if I switch the three `latest` properties to just query into `Release`:

```
    @property
    def latest_version(self):
        # Supply the latest stable version, if any stable releases exist.
        # If only pre-releases exist, supply the latest pre-release version.
        return (
            orm.object_session(self)
            .query(Release)                   <--------
            .filter(...
```

So, yeah, that should enable the tightening of the view code as you suggest. I'll make those changes.

bskinn:
Ah, I hadn't thought about that difference. But yeah, IMO having that explicit `pip install myproject==1.2.3` at the top of the page would be overall a good thing -- if someone goes to use that `myproject/latest/` view, it seems to me they're specifically asking the question, "What is the latest version of `myproject`, right now?", and this behavior gives it to them unambiguously.

ewjoachim:
Yeah but given it's usually not a good practice to pin a dependency (in one's source reqs file not in the lock file of course), I don't know if it's something we should encourage.

I'm not requesting any change here, and I'm not saying you should definitely have a strong opinion on that (you already explained that the main reason for implementing these browser views was because it was suggested in the issue) but I'm
Wondering what the usecase will be.

Will people link to the `/latest-stable` and `/latest-prerelease` pages from their github readme badges ? Or is it for PyPI power users to manually add those to the URLs ? Or...?

ewjoachim:
( @graingert if you have the time/the will to weight in and precise your usecase from what you exposed in the ticket ?)

graingert:
I'd want to link to latest pre-release unless there's a newer main release

ewjoachim:
And on the other GitHub badge you mentioned that pointed to the latest stable, dis you have in mind a page pointing on a specific release or the project page we have today? (Upon visiting the link, would you imagine the install instructions be `pip install foo` or `pip install foo==1.2.3`)

warehouse/packaging/models.py
```diff
@@ -268,6 +270,37 @@ def latest_version(self):
...
+    @property
+    def latest_unstable_version(self):
+        # Supply the latest available version, regardless of pre-release status.
+        return (
+            orm.object_session(self)
+            .query(Release.version, Release.created, Release.is_prerelease)
+            .filter(
+                Release.project == self,
+                Release.yanked.is_(False),
+                Release.is_prerelease is not None,
```
ewjoachim, line 298: I might misread it but this seems to contradict with the docstring: If I'm not mistaken, this selects only prereleases, where we'd want a stable if it's the most recent.

bskinn: I think this is ok -- the filter is `Release.is_prerelease is not None`, which is just a (possibly unnecessary) filter against a malformed table row; as opposed to `Release.is_prerelease is True` (or simply `Release.is_prerelease`), which I think would be a filter that would return only pre-releases.

docs/api-reference/json.rst
bskinn: Fixed in 0306080

warehouse/packaging/views.py
bskinn: Fixed in dc838308

warehouse/packaging/views.py
bskinn: Applied in dc838308

warehouse/packaging/views.py
bskinn: Fixed in dc838308

warehouse/legacy/api/json.py
bskinn: Implemented along these lines in 688d842e. Also removed redundant database lookups.

bskinn:
@ewjoachim, I've implemented your requested changes on a branch in my fork. See the PR over in my fork, for just this subset of the changes, here: https://github.com/bskinn/warehouse/pull/1

----

Separately, do you have any further thoughts on the web view showing, e.g., `pip install foo --pre` versus `pip install foo==1.2.3b17`?  While `latest` would be pretty easy (just redirect to `project/{name}`), I think that `latest-stable` and `latest-unstable` are problematic. 

Unless I miss my guess, actually implementing display of `pip install foo --pre` for `latest-unstable` on the web view would require addition of new frontend logic ... no idea how desirable that is, for something that's so specific to this new feature.

For `latest-stable`, I don't know if there's even a pip invocation that matches the behavior.

bskinn:
Ok, @ewjoachim, the auxiliary PR is merged. Ready for review starting with ee873dd.

(Not sure what's up with the CI error -- by my read of `requirements/main.txt`, the `google-api-core` requirement line looks to be properly formed, with a `==` version spec.)

di:
The error is due to https://github.com/pypa/pip/issues/9644

bskinn:
Marked back as draft to remind about reverting c6f664a before merge

di:
@bskinn c6f664a should be unnecessary now that #9543 is merged.

ofek:
Any update on this?

bskinn:
@ofek Given the extensive discussion ongoing about modernizing/standardizing the JSON API  (see, e.g., [here](https://github.com/pypa/packaging-problems/issues/367)), I'm guessing this PR is on hold. (AFAIK the current JSON API is technically in legacy status.)

*If* the feature set is still desired after all the turnover, I expect I'll need to revise the PR to fit into the new, standardized API, once it's finalized and implemented. (I haven't looked closely at the [new PEP draft](https://discuss.python.org/t/pep-rfc-python-package-index-warehouse-json-api-v1/9205); it's possible that the functionality of this PR is already part of that plan, but I doubt it since the stated goal was just to PEPify the current API.)

bskinn:
With the launch of the PEP 691 JSON API, are the semantics of this PR relevant any more? I haven't looked closely at what PEP691 provides in terms of version aliases.

If so, and if the features are still of interest, I'll be happy to reimplement under the PEP691 rubric.

Probably would close this PR and start from scratch.

bskinn:
&lt;looks closer&gt; Oh. PEP691 deals with a JSON simple API. That's ~unrelated to this, isn't it?

bskinn:
Closing as likely obsolete.

PR #8624, opened 2020-09-29
Remove undeclared dependency on attrs

### LINKED ISSUE #8299: Using undeclared dependency attrs

di:
Fixes #8299.

### DIFF
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -12,7 +12,6 @@
 
 import datetime
 
-import attr
 import celery.exceptions
 import pretend
 import pytest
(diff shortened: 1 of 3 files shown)


ewjoachim (commented):
:+1:

PR #8636, opened 2020-09-30
Manage releases: Move brackets tighter (like this) to avoid inner spaces ( like this )

hugovk:
On release management pages (e.g. https://test.pypi.org/manage/project/prettytable/releases/) there's extra space inside the brackets:

image

Add minuses to `{%- ... -%}` to strip the extra space.

---

# Before

Tested with this minimal repro:

```python
from jinja2 import Template

things = [
    ("Wheel", 1),
    ("Source", 1),
]

template = Template(
    """
        (<span dir="ltr">
        {% for packagetype, packagetype_count in things %}
          {%- if True -%}
          {{ packagetype_count }} {{ packagetype }}{{ ", " if not loop.last }}
          {%- endif -%}
        {% endfor %}
        </span>)
    """
)

print(template.render(things=things))
```
Outputs:
```html
        (<span dir="ltr">
        1 Wheel, 1 Source
        </span>)
```

# After

```python
from jinja2 import Template

things = [
    ("Wheel", 1),
    ("Source", 1),
]

template = Template(
    """
        (<span dir="ltr">
        {%- for packagetype, packagetype_count in things -%}
          {%- if True -%}
          {{ packagetype_count }} {{ packagetype }}{{ ", " if not loop.last }}
          {%- endif -%}
        {%- endfor -%}
        </span>)
    """
)

print(template.render(things=things))
```
Outputs:
```html
        (<span dir="ltr">1 Wheel, 1 Source</span>)
```

### DIFF
--- a/warehouse/templates/manage/releases.html
+++ b/warehouse/templates/manage/releases.html
@@ -47,11 +47,11 @@
           {{ count }} files
         {% endtrans %}
         (<span dir="ltr">
-        {% for packagetype, packagetype_count in version_to_file_counts[release.version].items() %}
+        {%- for packagetype, packagetype_count in version_to_file_counts[release.version].items() -%}
           {%- if packagetype != 'total' -%}
           {{ packagetype_count }} {{ packagetype|format_package_type }}{{ ", " if not loop.last }}
           {%- endif -%}
-        {% endfor %}
+        {%- endfor -%}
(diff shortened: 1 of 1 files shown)


di (approved):
Glad I'm not the only one bothered by these sort of things 😅

PR #8656, opened 2020-10-03
Update README.rst links

### LINKED ISSUE #8677: Broken readthedocs links in README
sorry if there's already an issue for this, but I wasn't able to find anything after a few searches

**Describe the bug**

The warehouse.readthedocs.io links in the readme are broken (for me)

[a broken (for me, at least) link in the readme](https://warehouse.readthedocs.io/development/getting-started/)

[A tweaked link that works](https://warehouse.readthedocs.io/development/getting-started.html)

imsahil007:
The links in the Readme file were not working.

Fixes #8677.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,17 +44,17 @@ Everyone interacting in the Warehouse project's codebases, issue trackers, chat
 rooms, and mailing lists is expected to follow the `PSF Code of Conduct`_.
 
 .. _`PyPI`: https://pypi.org/
-.. _`our development roadmap`: https://warehouse.readthedocs.io/roadmap/
-.. _`architectural overview`: https://warehouse.readthedocs.io/application/
+.. _`our development roadmap`: https://warehouse.readthedocs.io/roadmap.html
+.. _`architectural overview`: https://warehouse.readthedocs.io/application.html
 .. _`documentation`: https://warehouse.readthedocs.io
-.. _`Getting started`: https://warehouse.readthedocs.io/development/getting-started/
+.. _`Getting started`: https://warehouse.readthedocs.io/development/getting-started.html
 .. _`issue tracker`: https://github.com/pypa/warehouse/issues
 .. _`pypi.org`: https://pypi.org/
 .. _`on Freenode`: https://webchat.freenode.net/?channels=%23pypa-dev,pypa
 .. _`distutils-sig mailing list`: https://mail.python.org/mailman3/lists/distutils-sig.python.org/
-.. _`Running tests and linters section`: https://warehouse.readthedocs.io/development/getting-started/#running-tests-and-linters
+.. _`Running tests and linters section`: https://warehouse.readthedocs.io/development/getting-started.html#running-tests-and-linters
 .. _BrowserStack: https://browserstack.com/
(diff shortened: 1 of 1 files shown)


ewjoachim:
These links used to work, and they don't work anymore. Rather that fixing those, I think we should fix ReadTheDocs and make those links works, who knows how many links we have broken on other websites...

imsahil007:
> These links used to work, and they don't work anymore. Rather that fixing those, I think we should fix ReadTheDocs and make those links works, who knows how many links we have broken on other websites... (except if those were very old links and it's the last places that don't reference the new RTD url scheme ?)

Yeah, you are right. They used to work earlier.
But I checked out some other projects like this:
* https://github.com/scrapy/scrapy#install
* https://github.com/readthedocs/readthedocs.org#contributing
It seems like even readthedocs people are using that (.html) extension scheme

ewjoachim (approved):
I'd tend to agree, I think it would be interesting to know what changed, though. 

[Google](https://webcache.googleusercontent.com/search?q=cache:SL2L-lSm_NQJ:https://warehouse.pypa.io/development/getting-started/+&cd=1&hl=en) said it was still ok less than a month ago.

ewjoachim:
I'd bet the RTD configuration was switched on the website. Ideally, it would be much better if the configuration was versionned in a readthedocs.yml file in the repo to avoid this kind of surprises.

Capture d’écran 2020-10-03 à 19 47 32
Capture d’écran 2020-10-03 à 19 47 25

ewjoachim:
So, I don't think one or the other would be better, but I can't find such a change being discussed, so I wonder if it could have been an unvoluntary mistake. Best would be if it can be fixed on RTD.

imsahil007:
@ewjoachim 
People at RTD says that the URL without (.html) is wrong. We can redirect our websites using Sphinx Html in the above banner you mentioned.
This seems to be the only place where the URLs are not working.

di:
The current docs are configured to use https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.dirhtml.DirectoryHTMLBuilder. According to that:

> For example, the document `markup/rest.rst` will not result in an output file `markup/rest.html`, but `markup/rest/index.html`. When generating links between pages, the `index.html` is omitted, so that the URL would look like `markup/rest/`.

So I'm not sure why this stopped working.

imsahil007:
Hey @di
Read this https://docs.readthedocs.io/en/latest/user-defined-redirects.html#sphinx-redirects
If we want the absolute URL we need to change from HtmlDir to HTML only

di:
Changing that doesn't seem to fix this. It doesn't seem to change anything, actually. I wonder if this is a RTD bug?

imsahil007:
We will have to change this file https://github.com/pypa/warehouse/blob/ee96211cde2ad170f462f53a14c02a35e23bdf92/.readthedocs.yml#L6
Adding this configuration below will solve our problem
```
version: 2

sphinx:
  builder: html
  configuration: conf.py
  fail_on_warning: true
```

di:
Fixed by https://github.com/pypa/warehouse/issues/8677#issuecomment-705808480

PR #8719, opened 2020-10-17
New functional tests (behaviour)

### LINKED ISSUE #8616: Functional tests
We've had 2 regressions lately that lead to reverts (#8591 and #8555) and [more](https://github.com/pypa/warehouse/pulls?q=is%3Apr+sort%3Aupdated-desc+in%3Atitle+revert+is%3Aclosed+-author%3Aapp%2Fdependabot-preview+) in the last months. While not all of those could have been caught before impacting production, I believe some may.

Our current testing strategy has its own qualities & defaults. On the one hand, we have 100% coverage, which helps limiting cases of incorrectly transcribing our thought into code. On the other hand, its large use of stubs, and its positioning that is relatively low-level doesn't give us that much confidence into how good the code we wrote fits real life. For this we would need higher-level tests.

I've seen a few tests in the test codebase labeled functional, and it doesn't strike me as something fundamentally different from the other kinds of tests we have.

I'm thinking about starting a higher-level test suite that would be more likely to catch "real-life" issues, by avoiding mocks & stubs where possible, and running on the complete stack. I'm ok to try and work on this but I'm also perfectly ok if someone else volunteers.

I've got a few thoughts of where I'd tend to go in the long run, but I don't want to get too precise before making a first working iteration. To be clear though, I'm talking about running a server, and then trying to run a scenario while interacting only through the HTTPS interface, as much as possible. This does not necessarily imply a headless browser, but at some point, we could decide to go this way. What I'd need though even for the first iteration would be an idea of a realistic and useful scenario.

The place where we had troubles recently were the authentication, so I'd tend to say:

- create a user
- verify email
- login (failure, success)
- add token
- upload package with credentials (failure, success)
- upload package with token (failure, success)
- delete package
- delete user

This should help us with #8598 in the short term and also to envision #7266

ewjoachim:
Closes #8616

I'm introducing new kind of high-level test: behaviour tests. I already spend a little while playing with this and I'm not sure it's reasonable to invest more before having feedback.

Note:
- We can keep the browser-interaction part without the BDD part
- Of course this lack documentation as of now
- Using MechanicalSoup (requests + bs4) makes it as simple as can be for a high level test, but we don't have JavaScript, so things as simple as "check what's in the home page top-right menu" are made complicated by the fact this menu is actually populated by JS. Using a real browser would probably be much more complicated to run in the development enviornment & travis
- The 3 tests run when launched in isolation, but they fail as of now, because I haven't figured out how to reset the DB between test. **Any kind of help for doing that would be highly welcome**

Thanks to anyone who takes time to express feedback :)

(and ping @nlhkabu because you introduced me to BDD a while back)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,7 +48,9 @@ def pytest_collection_modifyitems(items):
         )
 
         module_root_dir = module_path.split(os.pathsep)[0]
-        if module_root_dir.startswith("functional"):
+        if module_root_dir.startswith("behaviour"):
+            item.add_marker(pytest.mark.behaviour)
+        elif module_root_dir.startswith("functional"):
             item.add_marker(pytest.mark.functional)
         elif module_root_dir.startswith("unit"):
             item.add_marker(pytest.mark.unit)
@@ -163,7 +165,15 @@ def mock_manifest_cache_buster():
 
 
 @pytest.fixture(scope="session")
-def app_config(database):
+def redis():
+    return {
+        "sessions.url": "redis://redis:0/",
+        "warehouse.xmlrpc.cache.url": "redis://redis:0/",
+    }
+
+
+@pytest.fixture(scope="session")
+def app_config(database, redis):
     settings = {
         "warehouse.prevent_esi": True,
         "warehouse.token": "insecure token",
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -163,7 +165,15 @@ def mock_manifest_cache_buster():
 
 
 @pytest.fixture(scope="session")
-def app_config(database):
+def redis():
+    return {
+        "sessions.url": "redis://redis:0/",
+        "warehouse.xmlrpc.cache.url": "redis://redis:0/",
+    }
```
ewjoachim, line 172: Ah yeah, almost forgot about that. I had to change the URL for redis, but not exactly sure how I should do about that:
- Either find a way so that it's not actually required, to ease Travis configuration
- Or configure redis in Travis (well, if it's not already there)

ewjoachim:
On the other hand, Browserstack is free for opensource. Also, I was given the oportunity to try out Playwright, and it was nice. Maybe it's worth the extra mile of doing real browser tests.

ewjoachim:
I'm going to close this PR for now to reduce the noise, but maybe someday we'll do something about that !

PR #8807, opened 2020-11-10
remove initial attempt at squat detection

ewdurbin:
This table is now well over 3.2M rows, is unused as far as I am aware, and causes unpleasant performance in the admin site.


pradyunsg:
Well, the CI doesn't like this change, but I certainly do! 🙃

di (approved):
Yeah, this didn't work.

ewjoachim (approved):

Issue #8321, opened 2020-07-25
Search by package name returns no matches


andy-z:
**Describe the bug**
Searching package on pypi.org by its full name returns zero results for package `pyemf`:

    0 projects for "pyemf" Did you mean 'pyamf'?

Searching for `pyemf` in Google returns PyPI link as first result: https://pypi.org/project/pyemf/

**Expected behavior**

PyPI should be able to find a package.

**To Reproduce**

Click https://pypi.org/search/?q=pyemf

**My Platform**

Web browser.

eamanu:
Hi! I would like to work on this if there any objection :)

di:
It seems like the package is not in the search index for some reason. One thing we could do to help here would be to provide a way for PyPI administrators to trigger the `reindex_project` task from the Admin UI: 
https://github.com/pypa/warehouse/blob/12c2b81b190c6bff5c1664d44879e6eca1135a68/warehouse/search/tasks.py#L229-L250

di:
Taking a second look at this, it's likely that the age of the package in question (last published Jan 1 2006) has something to do with this.

DavidDoran-Hub:
I am getting the same issue with my brand new package, “oligoss”. It does not appear on any searches. However, direct link to the project works.

To reproduce: https://pypi.org/search/?q=oligoss&o=

Link to project: https://pypi.org/project/oligoss/

enkidulan:
I've just observed this issue. Searching by `networkx` returns 0 results, however searching by `"networkx"` returns some results with the https://pypi.org/project/networkx/ package on top.

FOUREX:
My library is not searchable, what should I do?

Library: https://pypi.org/project/alerts-in-ua.py/

mscisdk:
I am encountering a similar issue with my new package, "msci.sdk". 
It does not show up in any searches, but the direct link to the project works fine.

https://pypi.org/search/?q=msci.sdk&o=
https://pypi.org/project/msci.sdk/

If you're interested, I wanted to let you know that the search functionality is working perfectly fine on testpypi for the same package.

miketheman:
@mscisdk search finds your project now. Sometimes our search index lags behind.
![Search results · PyPI](https://github.com/pypi/warehouse/assets/529516/8f3863db-ec80-4792-83eb-440dc1f88625)

mscisdk:
Thanks for checking @miketheman . the search functionality is working fine now.

miketheman:
[...]

PR #8822, opened 2020-11-12
Admin UI project detail reindex button

nicholasserra:
Refs #8321

Was mentioned in issue that an admin UI reindex button would be helpful. This adds a simple reindex button to the admin UI project detail page to fire off the reindex_project task. 

Unit test mocks the task function and just checks that the task was called and that the UI alert happens. I didn't see any frontend tests for admin views, correct me if i'm wrong, so I didn't add any.

YwYZ79gfuj

### DIFF
--- a/warehouse/admin/views/projects.py
+++ b/warehouse/admin/views/projects.py
@@ -22,6 +22,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from warehouse.accounts.models import User
 from warehouse.forklift.legacy import MAX_FILESIZE, MAX_PROJECT_SIZE
 from warehouse.packaging.models import JournalEntry, Project, Release, Role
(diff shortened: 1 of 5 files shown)


warehouse/admin/views/projects.py
```diff
@@ -465,3 +466,20 @@ def delete_project(project, request):
     remove_project(project, request)
 
     return HTTPSeeOther(request.route_path("admin.project.list"))
+
+
+@view_config(
+    route_name="admin.project.reindex",
+    permission="admin",
```
di, line 473: I think we can make this available to moderators, it's not destructive.

nicholasserra: Cool. Just pushed that change. View is now allowing moderators and I removed the superuser checks from the DOM.

di:
Thanks!

PR #8881, opened 2020-12-01
allow uploads of wheels targeting macosx_11_0, arm64

### LINKED ISSUE #8870: macOS 11 wheel upload is not allowed
**Describe the bug**

Upload of wheels targeting latest stable macOS (11.0) is rejected due to unsupported platform tag:

```
Binary wheel 'pyzmq-20.0.0-cp38-cp38-macosx_11_0_x86_64.whl' has an unsupported platform tag 'macosx_11_0_x86_64'.
```

**Expected behavior**

Upload of wheels built for current stable macOS (11.0) should be allowed.

**To Reproduce**

Build a wheel with tag `macosx_11_0_x86_64` and try to upload with `twine`. The simplest path is probably to `brew install python` on macOS 11.0 and build a wheel with that. I'm using pyzmq-20.0.0, but anything should do:

```python
brew install python@3.8
/usr/local/opt/python@3.8/bin/python3 -m venv ./env
source ./env/bin/activate
pip wheel pyzmq
twine upload ...
```

Folks who know more about wheels could probably hand-craft an os-tagged wheel on another platform without needing to install the latest macOS.

**My Platform**

macOS 11.0.1, Python 3.8, 3.9 from homebrew. Only homebrew Python targets the latest OS as the minimum version, but probably pyenv or similar would as well, or any Python built from source, I think. Pythons with an earlier target OS (conda, Python.org) do not have this issue.

**Additional context**

This is related to https://github.com/pypa/packaging/pull/319. I don't know where warehouse gets its supported tag list, and it may be that that PR to packaging is actually the only fix necessary, but as of today upload of valid tags is not allowed.

minrk:
- relaxed regex for mac version to allow specific major versions (currently 10, 11)
- adds support for `arm64`, `universal2` arch tags defined in https://github.com/pypa/packaging/pull/319

closes #8870

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -124,18 +124,25 @@ _allowed_platforms = {
     "linux_armv7l",
 }
 # macosx is a little more complicated:
-_macosx_platform_re = re.compile(r"macosx_10_(\d+)_(?P<arch>.*)")
+_macosx_platform_re = re.compile(r"macosx_(?P<major>\d+)_(\d+)_(?P<arch>.*)")
 _macosx_arches = {
     "ppc",
     "ppc64",
     "i386",
     "x86_64",
+    "arm64",
     "intel",
     "fat",
     "fat32",
     "fat64",
     "universal",
+    "universal2",
 }
+_macosx_major_versions = {
+    "10",
+    "11",
+}
+
 # manylinux pep600 is a little more complicated:
 _manylinux_platform_re = re.compile(r"manylinux_(\d+)_(\d+)_(?P<arch>.*)")
 _manylinux_arches = {
@@ -154,7 +161,11 @@ def _valid_platform_tag(platform_tag):
     if platform_tag in _allowed_platforms:
         return True
     m = _macosx_platform_re.match(platform_tag)
-    if m and m.group("arch") in _macosx_arches:
+    if (
+        m
+        and m.group("major") in _macosx_major_versions
+        and m.group("arch") in _macosx_arches
+    ):
         return True
     m = _manylinux_platform_re.match(platform_tag)
     if m and m.group("arch") in _manylinux_arches:
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2566,6 +2566,9 @@ class TestFileUpload:
             "manylinux_3_0_s390x",
             "macosx_10_6_intel",
             "macosx_10_13_x86_64",
+            "macosx_11_0_x86_64",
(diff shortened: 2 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -124,17 +124,19 @@ def namespace_stdlib_list(module_list):
...
 }
 # macosx is a little more complicated:
-_macosx_platform_re = re.compile(r"macosx_10_(\d+)_(?P<arch>.*)")
+_macosx_platform_re = re.compile(r"macosx_(?P<major>\d+)_(\d+)_(?P<arch>.*)")
 _macosx_arches = {
     "ppc",
     "ppc64",
     "i386",
     "x86_64",
+    "arm64",
```
minrk, line 133: arm64, universal2 discussed [here](https://discuss.python.org/t/apple-silicon-and-packaging/4516/21), implemented [here](https://github.com/pypa/packaging/pull/319).

warehouse/forklift/legacy.py
```diff
@@ -154,6 +156,8 @@ def _valid_platform_tag(platform_tag):
     if platform_tag in _allowed_platforms:
         return True
     m = _macosx_platform_re.match(platform_tag)
+    if m and int(m.group("major")) < 10:
```
minrk, line 159: If it's preferable to make new mac major version support opt-in, I'll switch this to a `_macosx_major_versions` allowed-set like `_macosx_arches`

di: Yes, let's do that.

minrk: Sure thing! Done.

ronaldoussoren (commented):
LGTM.

ronaldoussoren:
One thing to consider: Until now the minor version in the release changed, and was not checked for validity (I could upload a wheel for macosx_10_25_intel).   I expect that the next major release for macOS will be 12.0, which would require a change to PyPI as well because the PR checks that the major release is on of the currently known major releases.

That's a yearly change, so not too bad.

dstufft:
We can probably re-evaluate when/if that happens if it ends up being that macOS jus bumps the major every year and we end up with purely mechanical PRs, we can relax the requirement.

fxcoudert:
Hi, I'd like to understand this PR better. Before, the version numbering only included the major version numbers (`10_15`, `10_14`, etc.) and not the minor OS version. But since the numbering scheme has changed, now the minor version number is included (`11_0`). Is that on purpose?

di:
Previously it only allowed `10.x`. Now it allows `10.x` or `11.x`. Does that answer your question?

PR #8882, opened 2020-12-01
allow macosx_11 wheel upload

### LINKED ISSUE #8870: macOS 11 wheel upload is not allowed

minrk:
only allow macos 10 and 11, not new amd64 arch tags or not-yet-released macos major versions

stricter short-term alternative to #8881 in case that wants to wait for some further discussion

closes #8870

#8881 or similar is still needed for amd64 wheels, but this expands upload versions to allow for wheels built with current stable Python on current stable macOS.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2566,6 +2566,7 @@ class TestFileUpload:
             "manylinux_3_0_s390x",
             "macosx_10_6_intel",
             "macosx_10_13_x86_64",
+            "macosx_11_0_x86_64",
             # A real tag used by e.g. some numpy wheels
             (
                 "macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64."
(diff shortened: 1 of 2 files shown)


di:
Let's go with #8881 instead.

PR #8922, opened 2020-12-14
infra: add a vault container and scaffolding

woodruffw:
I broke this out of my local changes to #7488.

It sets up a new development container, `vault`, running an instance of Hashicorp Vault. A single (root) token is used everywhere for simplicity.

This PR doesn't contain any of the Pyramid-level service work needed to interact with the Vault service via `transit`. I'll break my local work into a follow-up PR for that.


docker-compose.yml
```diff
@@ -69,6 +89,13 @@ services:
       - packages:/var/opt/warehouse/packages
     ports:
       - "80:8000"
+    links:
```
ewdurbin, line 92: is this just making the links explicit, I'm not familiar with why this would be necessary for web but not worker in that case.

woodruffw: Yep, correct. My understanding (and experience) is that all containers are exposed by name on the same network by default; `links` just forces `docker-compose` to start `vault` before it starts `web`.

`worker` should also probably have a `links` entry for it, since we'll be accessing vault via Celery tasks. So I'll add it 🙂

di: It's advised to not use `links`: https://docs.docker.com/compose/compose-file/#links

woodruffw: Thanks for the ref! The other Docker page that documents `links` doesn't mention their deprecation: https://docs.docker.com/compose/networking/#links, so I missed it. I'll drop these.

ewdurbin (approved):
Since we will need to encode retries and resilience into the client anyway, this looks ready to go even without links/dependency managed by docker-compose

woodruffw:
> Since we will need to encode retries and resilience into the client anyway, this looks ready to go even without links/dependency managed by docker-compose

Yep -- part of the local deployment process for TUF also involves running `make inittuf` to fill Vault with signing keys and perform the initial backfill, so this isn't any more or less broken than Warehouse's state prior to `make initdb` already is 🙂

PR #8982, opened 2021-01-12
Switch to using the "system font stack"

### LINKED ISSUE #8634: Italic font gets a fallback in Cyrillic texts
**Describe the bug**

TL;DR normal and bold text gets rendered with "Source Sans Pro" but italics gets some locally-available Arial-based fallback. This makes the paragraph in the example below look inconsistent. I suppose "Source Sans Pro" does not contain any Cyrillic glyphs in italics.

pypi org--weird--italics

**Expected behavior**

The same font used for all text content consistently.

**To Reproduce**

1. Switch PyPI to Ukrainian
2. Go to https://pypi.org/help/#file-size-limit
3. Look closer at the word in italics ("меншого")
4. Check the font that got used for this word compared to the surrounding text. It is visible in DevTools if you highlight the corresponding `<em>` tag, then go to the "Computed" tab in the sidebar and scroll all the way down to the "Rendered Fonts" section.

**My Platform**

Google Chrome on Gentoo Linux (two of my laptops choose different fonts; because of the present fallbacks one system it is DejaVu Sans and the other it's Arial)
I've also tested it on a Chromebook (ChromeOS is technically Gentoo under the hood but it shouldn't matter) — it uses Arimo.
I'm pretty sure that this behavior will happen on any system that is able to download that main " Source Sans Pro" from the Internet so that it'll get used in some places but not the others.

**Additional context**

N/A

betatim:
This is a first attempt to close #8634 by switching to a system font stack.

Should I put a line break for the list of fonts? I ran `make reformat` hoping it would sort out my formatting issues but no luck :D

Screenshot 2021-01-12 at 13 44 28

I copied меншого to before the `<em>` tag to make it easier to compare (that is why it appears twice). I think it now uses the same font.

What should I do next? Are there things that can be cleaned up now that Source Sans Pro isn't in use any more?

Are there any conventions around adding `[WIP]` to the PR title or marking it as draft for stuff which needs a bit of hand-holding from an experienced person?

### DIFF
--- a/warehouse/static/sass/settings/_fonts.scss
+++ b/warehouse/static/sass/settings/_fonts.scss
@@ -14,7 +14,7 @@
 
 // SETTINGS - FONTS
 
-$base-font-family: "Source Sans Pro", "Helvetica", Arial, sans-serif;
+$base-font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
 $code-font: "Source Code Pro", monospace;
 
 


warehouse/static/sass/settings/_fonts.scss
```diff
@@ -14,7 +14,7 @@
 
 // SETTINGS - FONTS
 
-$base-font-family: "Source Sans Pro", "Helvetica", Arial, sans-serif;
+$base-font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
 $code-font: "Source Code Pro", monospace;
```
webknjaz, line 18: Unrelated but I think we also need to consider something like `Fira Code` here (separate discussion needed)

betatim: You mean a "system font stack" for the monospaced case?

Maybe https://github.com/pradyunsg/furo/blob/ecfee9c5a3872cfac72f5ee8c63c7d474ee8e60f/src/furo/assets/styles/variables/_index.scss#L41-L42 would work?

webknjaz:
> I think it now uses the same font.

Have you checked this with DevTools? The italics version seems visually smaller which suggests that the font may still be different.

> Should I put a line break for the list of fonts? I ran `make reformat` hoping it would sort out my formatting issues but no luck :D

I don't think it knows how to format CSS. I guess you can change it any way you want as long as it remains readable.

> Are there any conventions around adding `[WIP]` to the PR title or marking it as draft for stuff which needs a bit of hand-holding from an experienced person?

I think you can mark PRs as Draft when you yourself think that it's not ready for review.

P.S. I don't think I currently have the warehouse repo set up on my laptop but I've made this change via DevTools on the production instance. And in my env, it now selects Arial for both normal paragraph text and the emphasized one.

warehouse/static/sass/settings/_fonts.scss
webknjaz: Probably. But it's out of the scope of this PR.

betatim:
Devtools for the two bits of text below. I think this means they are using the same font (not quite sure why FA shows up for the normal text)

Screenshot 2021-01-12 at 15 07 52
Screenshot 2021-01-12 at 15 08 06


This is with firefox on OSX

betatim:
As far as I can tell this is ready for reviewing. System font stack is used for all normal text (normal == not monospaced).

di:
Thanks @betatim for taking a look at this!

> Are there things that can be cleaned up now that Source Sans Pro isn't in use any more?

Looks like it's still in use by the [admin interface CSS](https://github.com/pypa/warehouse/blob/master/warehouse/admin/static/css/AdminLTE.min.css) as well as some [SVG logos](https://github.com/pypa/warehouse/blob/01f87cfdb6275a8c8bc11b838618df246fc10fec/warehouse/static/images/logo-large.svg) and [other templates](https://github.com/pypa/warehouse/blob/01f87cfdb6275a8c8bc11b838618df246fc10fec/warehouse/static/html/mirror.html).

I'd be fine with dropping it in all those places and removing [our dependency on `fonts.google.com`](https://github.com/pypa/warehouse/blob/01f87cfdb6275a8c8bc11b838618df246fc10fec/warehouse/templates/base.html#L86) for all Warehouse pages (except our 404 page, this should remain).

However I'll leave this for @nlhkabu to approve since ultimately it has the effect of changing the font all of PyPI will render with.

nlhkabu:
On my side, I'd prefer to maintain Source Sans Pro where we can.
Are we able to target this per language (at least for now), as per https://www.w3.org/International/questions/qa-css-lang ?

FYI, support in this font is coming (slowly....) - see:
https://github.com/adobe-fonts/source-sans/issues/87 and
https://github.com/google/fonts/issues/1744

nlhkabu:
oh, and I should have said - thanks for your work on this @betatim! :tada:

betatim:
>  Are we able to target this per language (at least for now), as per [w3.org/International/questions/qa-css-lang](https://www.w3.org/International/questions/qa-css-lang) ?

Oh wow! I didn't know that this existed. I'll do some reading to find out what the answer is. If someone knows the answer to this question please post it (don't wait for me to find out first :D )

warehouse/static/sass/settings/_fonts.scss
fitojb: Please don’t use “BlinkMacSystemFont”; just using “system-ui” is enough in Blink-based browsers.

divbzero:
We could close this pull request and instead rely on the new version of Source Code Pro as @nlhkabu described on [23 Apr 2021](https://github.com/pypa/warehouse/pull/8982#issuecomment-826043873).

The two third-party changes that @nlhkabu are now resolved:
- adobe-fonts/source-sans#87 was resolved in 2019
- google/fonts#1744 was merged on 15 Sep 2021

I confirmed today that https://pypi.org/help/#file-size-limit now displays italic ”*меншого*” using:
- Chrome 100.0 on macOS 12.3
- Firefox 98.0 on Ubuntu 20.04 LTS
- Safari 15.4 on macOS 12.3

betatim:
> We could close this pull request and instead rely on the new version of Source Code Pro

That would work for me. I'll close this PR. Thanks for following up!

PR #9032, opened 2021-01-28
Remove [sqs] extra from celery dependency

di:
This is currently causing CI to fail due to https://github.com/pypa/pip/issues/8785 (pip cannot consider `celery[sqs]` and `celery` as the same dependency during resolution).

Since the [two extra dependencies this includes](https://github.com/celery/celery/blob/29eda054555fa95c83210e5e6bc3e839c80bcd3b/requirements/extras/sqs.txt#L1-L2) are already top-level dependencies for Warehouse, we don't actually need to use this extra.

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -4,7 +4,7 @@ argon2-cffi
 Babel
 bcrypt
 boto3
-celery[sqs]>=4,<5
+celery>=4,<5
 celery-redbeat
 certifi
 click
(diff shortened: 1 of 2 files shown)


ewdurbin:
This is kind of an unfortunate way to resolve (heh, pun), this... should we file an issue to re-add the extra at some point when pip can sort it out?

Issue #9033, opened 2021-01-28
"500: Internal Server Error" during a routine package update


MKuranowski:
**Describe the bug**

During a routine package upload the servers (both the "normal" and "test" repos) reply with a non-descriptive "500 Internal Server Error".

Apart from not being able to upload updated packages, I am also tempted to say that a simple "500 Internal Server Error" reply is an issue in and of itself, and that more information about what went wrong should be returned. I have no clue what could I do to resolve this issue on my own and I have to bother you and create an issue.

**Expected behavior**

I would expect the package to be normally uploaded, or, at least an error message that would describe what went wrong.

**To Reproduce**

```
git clone git@github.com:MKuranowski/osmiter.git
python setup.py sdist bdist_wheel
twine upload --verbose dist/osmiter-1.1.1.tar.gz dist/osmiter-1.1.1-py3-none-any.whl
```

<details>

<summary>Output of twine</summary>

```
Using configuration from /home/mikolaj/.pypirc
Uploading distributions to https://upload.pypi.org/legacy/
  dist/osmiter-1.1.1-py3-none-any.whl (19.0 KB)
  dist/osmiter-1.1.1.tar.gz (18.8 KB)
username set from config file
password set from config file
username: __token__
password: <hidden>
Uploading osmiter-1.1.1-py3-none-any.whl
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.7k/29.7k [00:00<00:00, 35.9kB/s]
Received "500: Internal Server Error" Package upload appears to have failed.  Retry 1 of 5
Uploading osmiter-1.1.1-py3-none-any.whl
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.7k/29.7k [00:00<00:00, 156kB/s]
Received "500: Internal Server Error" Package upload appears to have failed.  Retry 2 of 5
Uploading osmiter-1.1.1-py3-none-any.whl
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.7k/29.7k [00:00<00:00, 157kB/s]
Received "500: Internal Server Error" Package upload appears to have failed.  Retry 3 of 5
Uploading osmiter-1.1.1-py3-none-any.whl
[...]

PR #9034, opened 2021-01-28
Ensure that headers returned from upload endpoint are valid ISO-8895-1

ewdurbin:
Because the legacy upload endpoint returns user supplied text, we need to guard against characters in the Status header that the web server (gunicorn in our case) cannot handle.

This roundtrips the message that may contain user supplied text through ISO-8895-1 encoding/decoding before sending it off as a header.


ewdurbin:
Resolves #9033

di (approved):
For future reference: https://github.com/benoitc/gunicorn/issues/2319

PR #9064, opened 2021-02-09
Fix travis build

di:
Looks like the Travis base image no longer contains Python 2, which is a dependency of a sub-dependency of `gulp-sass`, which wraps `node-sass`, which is deprecated I think. It appears that [`gulp-dart-sass`](https://www.npmjs.com/package/gulp-dart-sass) is a drop-in replacement.


ewdurbin (approved):
indeed it is! thanks @di

PR #4961, opened 2018-10-27
Run "make lint" and "make reformat" inside docker container 

shevelevs:
This is so that it's not necessary to install python3.6 on the host machine. See #4948 for details.
This is not yet finished, but just to get feedback before I try to do more

Couple of implementation notes:
* static container doesn't have python3.6 installed, so it looks like it's easier to just use python3.6 image from docker registry
* technically  the python3.6 "build" cointainer doesn't need to be in docker-compose (it can be just used from Makefile via docker run directly)
* "docs" target is the thing that currently won't work, it needs some modifications because of recursive make calls

Issue #4948, opened 2018-10-27
`make reformat` fails (I'm using Python 3.7)


xni:
$ make reformat
# Create our Python 3.6 virtual environment
rm -rf .state/env
python3.6 -m venv .state/env
make: python3.6: No such file or directory
make: *** [.state/env/pyvenv.cfg] Error 1

di:
At the moment you'll need Python 3.6 installed on your machine to run this command. I'd recommend using [`pyenv`](https://github.com/pyenv/pyenv) to have multiple Python versions installed.

If you're interested, we might want to look into a way to move this into a Docker container instead, to avoid this issue in the future.

xni:
Yes, it would be great! Same for linting...

shevelevs:
@di, I can work on this - should be pretty straightforward

shevelevs:
I guess the question is whether you should be able to run stuff (especially build docs) without having to run docker. If that's the case, it makes sense to still preserve venv on the host machine. _Or_ it could be either separate target (IMO messy) or some kind of command line option or environment variable so you can select whether you want host or docker (or just default to docker if py3.6 is not found). Thoughts?

di:
Now that #9083 is merged, the remaining things that aren't running in containers are:

- [ ] [`make docs`](https://github.com/pypa/warehouse/blob/961575f8405566f8d567d51a0ea928c7a9eecb63/Makefile#L114-L116)
- [ ] [`make deps`](https://github.com/pypa/warehouse/blob/961575f8405566f8d567d51a0ea928c7a9eecb63/Makefile#L122-L131)
- [ ] [`make translations`](https://github.com/pypa/warehouse/blob/961575f8405566f8d567d51a0ea928c7a9eecb63/Makefile#L175-L209) (and it's subcommands)

PR #9083, opened 2021-02-12
GitHub Actions

### LINKED ISSUE #7001: Look at using GitHub Actions for CI

di:
Fixes #7001, closes #8255, closes #4961.


Makefile
```diff
@@ -139,6 +141,12 @@ deps: .state/env/pyvenv.cfg
 	rm -r $(TMPDIR)
 	$(BINDIR)/pip check
 
+github-actions-deps:
+ifneq ($(GITHUB_HEAD_REF), false)
+	git fetch origin $(GITHUB_HEAD_REF):refs/remotes/origin/$(GITHUB_HEAD_REF)
+	git diff --name-only $(GITHUB_HEAD_REF) | grep '^requirements/' || exit 0 && $(MAKE) deps
+endif
+
```
di, line 149: Something's broken with this: https://github.com/pypa/warehouse/pull/9083/checks?check_run_id=1885648471

```
Run make github-actions-deps BINDIR="$(dirname $(which python))"
/bin/sh: 1: .state/env/bin/python: not found
git fetch origin github-actions-ci:refs/remotes/origin/github-actions-ci
From https://github.com/pypa/warehouse
 * [new branch]        github-actions-ci -> origin/github-actions-ci
git diff --name-only github-actions-ci | grep '^requirements/' || exit 0 && make deps
fatal: ambiguous argument 'github-actions-ci': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

di: Fixed!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  test:
```
ichard26, line 4: Internal PRs (that come from branches of pypa/warehouse) match both the push and pull_request events. This leads to a duplication of ci workflows checks on such PRs. Not only is it a waste of resources, it makes the Status Checks UI more frustrating to go through.
```suggestion
  test:
    # We want to run on external PRs, but not on our own internal PRs (for pull_request events) as 
    # they'll be run by the push to the branch. Without this if check, checks are duplicated since 
    # internal PRs match both the push and pull_request events.
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ github.job }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ github.job }}-${{ env.cache-name }}-
+            ${{ runner.os }}-build-${{ github.job }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install Node dependencies
+        run: npm ci
```
webknjaz, line 71: Is this necessary in all the jobs? Should it be conditional?

di: You're correct, but this wouldn't shorten the parallel runtime of the CI suite. I opted to keep this workflows simple instead.

di: (because the longest-running job still requires the Node dependencies to be installed)

ewdurbin:
My only concern here is that the check with the busted git command... still went green?

Are we doing something wrong that an error like that doesn't surface? Could lead to some broken ships.

.github/workflows/ci.yml
webknjaz: I think it's worth saving the minutes because the number of parallel jobs affects other runs in the repo. This will be visible in other PRs when the jobs don't start for some time without any obvious reason. Besides, I'm pretty sure there's some monthly cap on the used minutes in the org...

di:
> My only concern here is that the check with the busted git command... still went green?
>
> Are we doing something wrong that an error like that doesn't surface? Could lead to some broken ships.

The current command just assumes that the `git diff` will always succeed. I added a guard in d5d0347 just to be safe.

.github/workflows/ci.yml
di: Thanks for pushing on this, this brought the CI run down to 3-ish minutes and also got me to make some progress towards #4948 as well.

.github/workflows/ci.yml
webknjaz: @di this thing will avoid unnecessary runs for dependabot PRs because it makes pushes to upstream.

di: I think we want to retain the existing behavior.

webknjaz: For any pushes? I often add exclude-branches matching dependabot because it always causes PR builds.

di: Yep, for all pushes.

Cu2y (approved):
reviewed changes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,99 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  test:
+    strategy:
+      matrix:
+        include:
+          - name: Tests
+            command: bin/tests --postgresql-host localhost
+            needs-python: ${{ true }}
```
webknjaz, line 10: Nice trick! I normally do something like `fromJSON()` in the workflow itself.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install Python dependencies
+        if: ${{ matrix.needs-python }}
+        run: |
+          pip install -U pip setuptools wheel
+          pip install -r requirements.txt
+          pip install -r requirements/dev.txt
+      - uses: actions/setup-node@v2
+        if: ${{ matrix.needs-node }}
+        with:
+          node-version: 14.4.0
```
webknjaz, line 81: It's not important but I normally put versions into the matrix. Even if it's just one matrix item, it's nicer to keep all the declarations in one place at the top. And updating it is easier — no need to look for this place in the middle of the workflow file.

But it's up to you, of course. I just wanted to share what I do to improve the maintainability.

webknjaz (approved):
LGTM unless you want to do some polishing

ewdurbin (approved):

ewdurbin:
Thank you @di and @webknjaz!

PR #9096, opened 2021-02-15
Rephrase search error not to say temporarily

nanonyme:
It is entirely unnecessary to communicate why xmlrpc is unavailable


ewdurbin (approved):

ewdurbin:
That's a good edit, thank you @nanonyme

PR #9166, opened 2021-03-04
Update sponsor grid CSS to reflect new design for PSF sponsors

nlhkabu:
Supports https://github.com/python/pypi-theme/pull/37


di (approved):
Will need to pull in the new version of `pypi-theme` with this PR as well once it's published.

PR #9264, opened 2021-03-19
Use pypitoken to generate, check, introspect tokens

### LINKED ISSUE #9018: Improve "invalid macaroon signature" error
Right now, an API token can fail for a number of reasons, most of which produce an "invalid macaroon signature" exception.

https://github.com/pypa/warehouse/blob/aafc5185e57e67d43487ce4faa95913dd4573e14/warehouse/macaroons/caveats.py#L93

Common issues that could use more detailed error messages:
* the API token is scoped for a different project than it's being used for
* the API token is malformed and cannot be deserialized
* the API token doesn't begin with `pypi-`
* (probably more)

### LINKED ISSUE #9184: Managing API tokens with `pypitoken`
This ticket is the counterpart for [that discussion](https://discuss.python.org/t/pypitoken-a-library-for-generating-and-manipulating-pypi-tokens/7572) for discussing the integration of [pypitoken](https://pypi.org/project/pypitoken/) into warehouse. I might start working on the PR at some point, but if there are huge structural problems to doing so, or things that should be figured out beforehand, I'll gladly welcome remarks before working on the integration :)

- https://github.com/ewjoachim/pypitoken
- https://pypitoken.readthedocs.io/en/latest
- https://pypi.org/project/pypitoken/

I plan to remove `caveats.py`, and have the Macaroons service use pypitoken directly. This should simplify code. Also, the Form for creating tokens will likely be simplified too, as it currently holds some information on the caveats JSON structure, which is not needed if `pypitoken` takes responsibility for crafting tokens. All in all, I hope, we'll remove much more code than we'll add, while making future evolutions easier.

@

ewjoachim:
Fixes #9184 
Fixes #9018 

Ah, a PR where we remove code without removing features, I love those :D

So, normally after this PR, the warehouse codebase won't know about the format of the caveats anymore, and we should be able to add new restriction types with much more ease :)


tests/unit/macaroons/test_services.py
```diff
@@ -69,17 +58,12 @@ def test_find_macaroon(self, user_service, macaroon_service):
         assert isinstance(dm, Macaroon)
         assert macaroon.id == dm.id
 
-    def test_find_userid_no_macaroon(self, macaroon_service):
-        assert macaroon_service.find_userid(None) is None
```
ewjoachim, line 73: The MacaroonService.find_userid method was never called with None. I'm not sure it makes sense to leave this codepath.

tests/unit/macaroons/test_services.py
```diff
@@ -131,82 +106,53 @@ def test_verify_no_macaroon(self, macaroon_service):
...
-    def test_deserialize_raw_macaroon_when_none(self, macaroon_service):
-        raw_macaroon = pretend.stub()
-        macaroon_service._extract_raw_macaroon = pretend.call_recorder(lambda a: None)
-
-        with pytest.raises(services.InvalidMacaroon):
-            macaroon_service._deserialize_raw_macaroon(raw_macaroon)
-
-        assert macaroon_service._extract_raw_macaroon.calls == [
-            pretend.call(raw_macaroon),
-        ]
```
ewjoachim, line 161: `_extract_raw_macaroon` doesn't exist anymore, and the `load` function will raise when no token is extracted, and `test_deserialize_raw_macaroon` tests that.

tests/unit/macaroons/test_services.py
```diff
@@ -131,82 +106,53 @@ def test_verify_no_macaroon(self, macaroon_service):
...
-        ],
-    )
-    def test_deserialize_raw_macaroon(self, monkeypatch, macaroon_service, exception):
-        raw_macaroon = pretend.stub()
-        macaroon_service._extract_raw_macaroon = pretend.call_recorder(
-            lambda a: raw_macaroon
-        )
-        monkeypatch.setattr(
-            pymacaroons.Macaroon, "deserialize", pretend.raiser(exception)
-        )
```
ewjoachim, line 183: `pypitoken` has a test for various types of errors https://github.com/ewjoachim/pypitoken/blob/master/tests/test_token.py#L325-L344 and a catchall because of https://github.com/ecordell/pymacaroons/issues/50 : https://github.com/ewjoachim/pypitoken/blob/master/pypitoken/token.py#L328-L332

tests/unit/manage/test_forms.py
```diff
@@ -431,6 +431,7 @@ def test_validate_token_scope_valid_user(self):
         )
 
         assert form.validate()
+        assert form.validated_restrictions == {}
```
ewjoachim, line 434: This wasn't tested, I added a test, it's free :)

warehouse/macaroons/services.py
```diff
@@ -104,40 +85,66 @@ def verify(self, raw_macaroon, context, principals, permission):
...
         """
         Returns a tuple of a new raw (serialized) macaroon and its DB model.
         The description provided is not embedded into the macaroon, only stored
         in the DB model.
         """
         user = self.db.query(User).filter(User.id == user_id).one()
 
-        dm = Macaroon(user=user, description=description, caveats=caveats)
+        identifier = uuid.uuid4()
+        key = _generate_key()
```
ewjoachim, line 113: The previous flow was

1. we generate the caveats object
2. we create the database macaroon, letting the DB choose the identifier and have the key computed by a python function set as model default
3. we read the id and key from the object after the DB flush
4. we generate the pymacaroons object from those

Because we want pypitoken to generate the caveats object and pypitoken needs the key and id, the new flow is:

1. we generate a random key and id
2. we generate the pypitoken object
3. we extract the caveats
4. we create the database macaroon with the already computer key, id and caveats

warehouse/macaroons/services.py
```diff
@@ -67,30 +56,22 @@ def find_macaroon(self, macaroon_id):
...
         try:
-            return pymacaroons.Macaroon.deserialize(raw_macaroon)
-        except (
-            MacaroonDeserializationException,
-            Exception,  # https://github.com/ecordell/pymacaroons/issues/50
-        ):
-            raise InvalidMacaroon("malformed macaroon")
+            return pypitoken.Token.load(raw_macaroon)
+        except pypitoken.LoaderError as exc:
+            raise InvalidMacaroon(str(exc))
```
ewjoachim, line 62: It's one of the promises of the pypitoken API that the exception message is displayable to the user (same when calling `check()`)

warehouse/macaroons/services.py
merwok: The comment above the model field should be updated (it discusses having the default factory in database vs python).

ewjoachim: (done)

warehouse/templates/pages/help.html
```diff
@@ -496,7 +496,7 @@ <h3 id="apitoken">{{ apitoken() }}</h3>
...
           {% endtrans %}
         </p>
-        <p>{% trans %}Advanced users may wish to inspect their token by decoding it with base64, and checking the output against the unique identifier displayed on PyPI.{% endtrans %}</p>
+        <p>{% trans trimmed pypitoken_href='https://pypitoken.readthedocs.io/', title=gettext('External link') %}Advanced users may wish to inspect their token by decoding it with <a href="{{ pypitoken_href }}" title="{{ title }}" target="_blank" rel="noopener">the <code>pypitoken</code> library</a>, and checking the identifier against the unique identifier displayed on PyPI. Please make sure to keep your token identifier private, though. Knowing just the identifier of a token is not enough to impersonate the user, but it's sufficient to have a token be disabled.{% endtrans %}</p>
```
ewjoachim, line 499: Added a little bit more info than strictly required but I thought it was interesting :)

PR #9279, opened 2021-03-22
Switch from custom smtp service to maildev

ewjoachim:
Because I love those stats: Capture d’écran 2021-03-22 à 21 59 35

Should allow us to actually test that HTML messages work before shipping them 🚀 💥 

Capture d’écran 2021-03-22 à 21 52 20
Capture d’écran 2021-03-22 à 21 52 48
Capture d’écran 2021-03-22 à 21 53 12


docker-compose.yml
```diff
@@ -125,14 +125,10 @@ services:
...
-    build:
-      context: .
-    command: python /opt/warehouse/dev/smtp.py 0.0.0.0:2525
+  maildev:
+    image: maildev/maildev
     ports:
-      - "2525:2525"
-    volumes:
-      - ./dev/smtp.py:/opt/warehouse/dev/smtp.py
+      - "1080:80"
```
ewjoachim, line 131: I haven't exposed the 25 port externally because the worker container doesn't need port forwarding to access it, but if you have the habit of exposing ports to the host, e.g. so that you can run the worker outside of docker, let me know :)

di:
Two things:

 * Can you add some documentation about how to use this? E.g. what URL to visit.
 * One handy feature of the current service is that it logs the entire message to the console, which makes it easy to quickly follow a confirmation link in development. Does maildev do the same? Could we configure it to do the same?

ewjoachim:
> Can you add some documentation about how to use this? E.g. what URL to visit.

Of course, I'll do that :)

> One handy feature of the current service is that it logs the entire message to the console, which makes it easy to quickly follow a confirmation link in development. Does maildev do the same? Could we configure it to do the same?

No idea. I'll try, and the --verbose mode, and I'll let you know.
We could also use a dedicated MailSender subclasses that prints the email ?

ewjoachim:
What the emails look like in the term:

Capture d’écran 2021-03-24 à 00 07 05

The doc:

Capture d’écran 2021-03-25 à 18 18 42
Capture d’écran 2021-03-25 à 18 18 51

ewdurbin (approved):

PR #9316, opened 2021-03-31
Update pypi-theme

di:


Babai143:
sResult(list):
        def __init__(self, key):
            self._key = key

        def __repr__(self):
            result = ["  WCS key '{}':".format(self._key or ' ')]
            if len(self):
                for entry

PR #9322, opened 2021-04-01
Clarify source of metadata in JSON API

### LINKED ISSUE #9274: Clarification of where the release metadata (notably requires_dist) in the JSON API comes from
(I chose the "feature request" template, as there isn't a "request for information" option, so treating this as a feature request for the documentation to be improved seemed best. But just getting an answer here would be sufficient for me).

The JSON API documented [here](https://warehouse.pypa.io/api-reference/json.html) includes `requires_dist` metadata. But it's not clear where that data comes from or when it's filled in, so it's essentially useless - without knowing how it's derived, applications can't reliably use it for anything.

A clarification on how Warehouse determines that data for projects. Specifically:

1. Which release file is it extracted from (in theory, a project release could contain different metadata in its Linux and Windows wheels, for example). My assumption is "one of the wheels, but which one is essentially arbitrary", which is probably good enough for my purposes (see below).
2. Was the data backfilled to older releases when this was added, or should I assume that for entries older than some date, missing data just means "we weren't collecting the data at this time"? (I'm pretty sure that it wouldn't have been backfilled).
3. For releases recent enough that the data should be there, how should I interpret a value of null? I can imagine multiple possibilities:
   3a. The project explicitly declares that it has no dependencies.
   3b. The project didn't upload wheels, and you don't extract metadata from sdists, so the project might have dependencies.
   3c.  The project initially uploaded a sdist but later added a wheel, and you don't update the data in this case.

**Additional context**
I am looking for this information for the purposes of research into projects and their dependencies on PyPI. As a consequence, I don't need 100% accurate data, but understanding the limitations of what is available would be extremely useful for me, as it would save me from having to download potentially thousands of wheels from PyPI and process them myself. I'm also mostly interested in the latest releases of projects, so historical data isn't critical to me, but being able to look at whether metadata changes over time might be of interest if history *is* available.

The main thing I'd like to have is some heuristics on how to interpret a value of null. One of the key questions I want to answer is "what proportion of projects have dependencies at all" and it's hard to know that without being able to distinguish between "not known" and "definitely not there".

I know there is work going on to standardise and formalise the JSON API, but I don't know how far along that is, and I would still find it useful to know the current situation.

If someone can give me a pointer to the relevant parts of the Warehouse code, and a rough summary, I'm happy to go and read the code and work out the details for myself, but at the moment I'm unfamiliar with the Warehouse codebase, so I don't know where to start.

pfmoore:
Fixes #9274.

### DIFF
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -13,7 +13,8 @@ Project
     the release name, URL, and MD5 and SHA256 hash digests, and are keyed by
     the release version string. Metadata returned comes from the values provided
     at upload time and does not necessarily match the content of the uploaded
-    files.
+    files. The first uploaded data for a release is stored, subsequent uploads
+    do not update it.
 
     **Example Request**:
 
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -11,7 +11,9 @@ Project
     Returns metadata (info) about an individual project at the latest version,
     a list of all releases for that project, and project URLs. Releases include
     the release name, URL, and MD5 and SHA256 hash digests, and are keyed by
-    the release version string.
+    the release version string. Metadata returned comes from the values provided
+    at upload time and does not necessarily match the content of the uploaded
+    files.
 
     **Example Request**:
 


ewjoachim (approved):

docs/api-reference/json.rst
```diff
@@ -11,7 +11,9 @@ Project
     Returns metadata (info) about an individual project at the latest version,
     a list of all releases for that project, and project URLs. Releases include
     the release name, URL, and MD5 and SHA256 hash digests, and are keyed by
-    the release version string.
+    the release version string. Metadata returned comes from the values provided
+    at upload time and does not necessarily match the content of the uploaded
```
ewjoachim, line 15: Do you want to add that it's the first upload that sets the value in case multiple distributions are uploaded ?

pfmoore: I didn't do that, because I wasn't sure if it was true 🙂 If you're confirming that this is the case, then I can add that.

(Personally, "doesn't necessarily match the content of the uploaded files" is the deal-breaker for my use case, so I don't have a personal need to be more precise than that, but I do want whatever goes here to be accurate, obviously).

di: It's (unfortunately) true, at least until https://github.com/pypa/warehouse/issues/8090 is resolved.

pfmoore: Done.

di (approved):

ewjoachim (approved):

di:
Thanks @pfmoore!

PR #9341, opened 2021-04-03
Admin: ensure exactly one primary email is set

### LINKED ISSUE #9321: Reset password doesn't work if user doesn't have verified/primary email address.
**Describe the bug**

Password reset is not working. I get the email, follow the link, enter the password twice, submit the form, and get back a page with no error messages. But the page title is "Error processing form -- Log in".

**Expected behavior**

I expected a positive confirmation that my password was reset, and to be able to login with the new password. Neither one happens.

**To Reproduce**

1. Start at `https://pypi.org/account/login/`
2. Click "Forgot password?" link
3. Enter the email address associated with my PyPI account (gerg.ward at gmail dot com)
4. Check my inbox and the email is already there.
5. Click the link: `https://pypi.org/account/reset-password/?token=.eJw1...`
6. Generate a new password outside my browser.
7. Copy/paste the new password into both password fields; text changes to "Passwords match" and the "Reset password" button is enabled.
8. Click "Reset password"
9. Receive the same "Reset password" form again, but with positive confirmation and no error messages. Only the page title gives a clue: "Error processing form –  Reset your password · PyPI". When I try to login with the new password (again, copy/pasted), it does not work.
 
I'll attach HTML and screenshot.

**My Platform**

Firefox 87.0 on Ubuntu 20.04.2.

Same behaviour with Chromium.

No HTTP proxy.

**Additional context**

ewjoachim:
Fixes #9321


ewjoachim:
I'm fixing the tests. Do you think there might be a user with no email at all or is this something that never happens ?

ewjoachim:
Hm, I'm afraid that modifying `primary_email` could have a lot of other consequences, it might be safe to create a new property.

di:
I'm not sure this is the right approach, we probably don't want to make assumptions about what the primary email is.

The bug in question seemed to happen because the email in question was reset/added but not marked as primary. We should ensure that it's not possible to add an email address via the admin UI without the primary field checked if there are no other primary email addresses on the account.

I'm not super concerned about old accounts that might not have emails or might not have a primary email.

ewjoachim:
Thanks for your insight!

I had assumed the problem was old emails because my tests so far didn't yield any way of not having a primary email:
- when you create an account, you have to provide an email, and it's set as primary
- There's no delete button if the email is primary

So maybe it's a UI bug somewhere (e.g. maybe if I open 2 webpages and swith primary in one, I can delete the primiry email in the other one, I'll check that) but if there's no such bug, I would say emails should be sent to any email if there's no primary (which should not happen in the first place)

ewjoachim:
Capture d’écran 2021-04-03 à 21 36 55

Nope. I've checked and the view successfully detects an attempt in deleting the primary address with an error.

di:
I was specifically talking about the PyPI admin UI, in which it is possible to make an account have no primary email addresses.

ewjoachim:
Do we have an idea how many accounts have no primary address ?

ewjoachim:
Aaaaaaah ok, I'll check.

ewjoachim:
@di new approach :) Better ?

https://user-images.githubusercontent.com/1457576/113491652-bbd0d300-94d2-11eb-89b5-25dbba3fd577.mov

di:
Yep, that's exactly what I was thinking. Will review this shortly!

warehouse/admin/views/users.py
```diff
@@ -92,6 +93,14 @@ class UserForm(forms.Form):
 
     emails = wtforms.fields.FieldList(wtforms.fields.FormField(EmailForm))
 
+    def validate_emails(self, field):
+        # If there's no email on the account, it's ok. Otherwise, ensure
+        # we have 1 primary email.
+        if field.data and len([1 for email in field.data if email["primary"]]) != 1:
```
di, line 99: Here's a fun trick
```suggestion
        primaries = iter([email["primary"] for email in field.data])
        if any(primaries) and not any(primaries):
```
I'm not sure that's more readable than what you had here, but it came to mind when reviewing.

di (approved):

warehouse/admin/views/users.py
ewjoachim: Haha, fun. But why do `iter([a for a in b])` when you can do `(a for a in b)` ? 😄 

It's fun, it's also less readable IMHO. (also, if there's no email at all we should be forced to add one, which I believe your example doesn't do) ((also I think you meant it the other way around, we want to `not` this whole clause so that's `if not any(primaries) or any(primaries)`)

Other solutions (that also don't handle the no email case for the sake of simplicity):
```python
import collections
if collections.Counter(email["primary"] for email in field.data).get(True) != 1:
```
Or good ol':
```
# bools are just integers with values 0 or 1
if sum(email["primary"] for email in field.data) != 1:
```
I actually quite like the last one :o

ewjoachim:
```suggestion
        if field.data and sum(email["primary"] for email in field.data) != 1:
```

PR #9378, opened 2021-04-13
Admin: Allow moderators to perform project ownership transfers

yeraydiazdiaz:
The current process for PEP 541 is suffering from a bottleneck in availability of admins because adding and removing users to projects in the admin panel is limited to users with the `admin` permission. This PR reduces the required permission to `moderator`.

This is the lowest effort measure to solve the problem, I can see how this may be seen as a potential security risk, but I thought I'd try and jumpstart the conversation.

### DIFF
--- a/warehouse/admin/templates/admin/projects/detail.html
+++ b/warehouse/admin/templates/admin/projects/detail.html
@@ -124,7 +124,7 @@
                   <td><a href="{{ request.route_path('admin.user.detail', user_id=role.user.id) }}">{{ role.user.username }}</a></td>
                   <td>{{ role.role_name }}</td>
                   <td>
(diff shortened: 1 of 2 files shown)


ewjoachim (approved):
I'm surprise that there is no change in the tests.

I'm :+1: for the change, but probably biased as someone who would gain a new ability.
An alternative could be to make you admin, I guess :)

yeraydiazdiaz:
> I'm surprise that there is no change in the tests.

I was too, I didn't see any tests for permissions, possibly because they're implemented using the decorator which one can argue it's part of Pyramid.

ewdurbin (approved):

PR #9398, opened 2021-04-17
Display 2FA status on package co-owners/maintainers

### LINKED ISSUE #5791: Display whether fellow maintainers/owners have enabled 2FA
Currently package owners can't determine if their fellow owners/maintainers have enabled 2FA on their accounts, which makes it difficult to maintain a certain level of security for a given package.

On the list of roles for a project, display a label next to each user which indicates whether 2FA is enabled or disabled.

**Additional context**
Blocked on #5790, otherwise this would allow any attacker to determine the 2FA status for a given user.

ewjoachim:
Closes #5791

I believe that given this is 100% a template fix, we should still have 100% coverage, and I believe we don't write automated tests for this, so, yolo :)

I choose not to display a badge if no 2FA rather than displaying a red badge, but this can be discussed.

EDIT: I originally added this to invites too, but re-reading the ticket highlighted that this might not be desirable. I'm not sure if it's really problematic that an attacker can know if you have 2FA or not but I can imagine it's not something to handle lightly.

Capture d’écran 2021-04-17 à 14 37 28

The screenshot below shows than even if user ewdurbin has 2FA, the interface doesn't show it because the invite is pending.
Capture d’écran 2021-04-17 à 14 56 34

### DIFF
--- a/warehouse/templates/manage/roles.html
+++ b/warehouse/templates/manage/roles.html
@@ -19,14 +19,14 @@
 
 {% macro two_factor_badge(user) -%}
   {% if user.has_two_factor %}
-    <span class="badge badge--success">
-      <i class="fa fa-check" aria-hidden="true"></i>
-      {% trans %}2FA enabled{% endtrans %}
-    </span>
+  <span class="badge badge--success" title="{% trans %}2FA enabled{% endtrans %}">
+    {% trans %}2FA{% endtrans %}
+    <i class="fa fa-check" aria-hidden="true"></i>
+  </span>
   {% else %}
-  <span class="badge badge--danger">
+  <span class="badge badge--danger" title="{% trans %}2FA disabled{% endtrans %}">
+    {% trans %}2FA{% endtrans %}
     <i class="fa fa-times" aria-hidden="true"></i>
-    {% trans %}No 2FA{% endtrans %}
   </span>
   {% endif %}
 {% endmacro %}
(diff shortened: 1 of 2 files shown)


ewjoachim:
Ah I just saw in the original ticket "Blocked on #5790, otherwise this would allow any attacker to determine the 2FA status for a given user."

So I guess the goal is NOT to be able to see this for invites.

warehouse/templates/manage/roles.html
```diff
@@ -17,6 +17,15 @@
 
 {% block title %}{% trans project_name=project.name %}Manage '{{ project_name }}' collaborators{% endtrans %}{% endblock %}
 
+{% macro two_factor_badge(user) -%}
+  {% if user.has_two_factor %}
+    <span class="badge badge--success">
+      <i class="fa fa-check" aria-hidden="true"></i>
+      {% trans %}2FA enabled{% endtrans %}
```
di, line 24:
```suggestion
      {% trans %}2FA{% endtrans %}
```
What do you think about adding a red badge for users without 2FA enabled?

ewjoachim: To quote myself from above,
> I choose not to display a badge if no 2FA rather than displaying a red badge, but this can be discussed.

:D

I'm perfectly fine doing it, I just couldn't decide by myself if it was akin to "blaming users who haven't activated 2FA" which there might be good reasons for (e.g. you don't have the necessary hardware)

ewjoachim: Capture d’écran 2021-04-23 à 21 54 46

di (approved):

di:
Thanks!

PR #9411, opened 2021-04-22
Fix JSON endpoint 404 when only yanked releases are available

### LINKED ISSUE #8966: JSON endpoint for project is 404 if only yanked releases are available
From https://github.com/pypa/warehouse/issues/3709#issuecomment-754973958, it appears that the JSON API for a project is 404 if the only releases available are yanked releases.

E.g. https://pypi.org/pypi/django-twilio2/json should probably look like https://pypi.org/pypi/django-twilio2/0.9.0/json instead of being 404.

AustinTSchaffer:
Fixes #8966 

This PR removes the check on "yanked" from the filter clause and adds it to the order by clause in the JSON endpoint's releases query. This ensures that yanked releases are returned from the DB query, but are prioritized lower than all other release variants.

This PR also adds unit tests for this behavior.

### DIFF
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -118,6 +118,23 @@ class TestJSONProject:
         assert resp is response
         assert json_release.calls == [pretend.call(release, db_request)]
 
+    def test_latest_release_yanked(self, monkeypatch, db_request):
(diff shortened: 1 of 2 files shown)


di (approved):

di:
Thanks!

PR #9449, opened 2021-04-30
Always show version in pip command

bfirsh:
The most common reason I find myself going to PyPi is to find the latest version to copy & paste into my `requirements.txt`. It'd be really nice if the main package page had the string that goes into requirements.txt. What I usually do is copy and paste the title, and edit the space to be `==`.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -308,7 +308,7 @@ msgstr ""
 #: warehouse/templates/manage/release.html:175
 #: warehouse/templates/manage/releases.html:140
 #: warehouse/templates/manage/releases.html:173
-#: warehouse/templates/packaging/detail.html:314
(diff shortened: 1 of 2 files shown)


ewjoachim:
Pinning all the dependencies isn't always advisable. Have you, by any chance, considered using [pip-compile](https://github.com/jazzband/pip-tools) or a similar tool (poetry, pipenv, ...) to maintain a `requirements.in` file without pinning, and have a tool take the hastle of adding the pins for you ? Additionally, you'd get [hashes](https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode) for free.

miketheman:
I'm going to close this, as there's not a need to display the current version on the main page of a given project all the time.

We already have the ability to display the desired version when looking at the Release History of a given release. The version will display if the selected version is the latest.

PR #9494, opened 2021-05-07
Add user timezone to localized dates' alt text

### LINKED ISSUE #9476: Add the timezone in the release date alt text

It's hard to guess which timezone the timestamps use and while assuming the user assumes it is UTC is reasonable, it's not explicit.

image

Just add the timezone to the timestamp.

**Additional context**

None.

berinhard:
Fixes #9476

This PR updates the JS controller used to display localized dates and date time strings to also display user's timezone.


warehouse/static/js/warehouse/controllers/localized_time_controller.js
```diff
@@ -49,6 +49,8 @@ export default class extends Controller {
       }
     }
 
+    const userTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+    isoDate = isoDate + ` (${userTimezone})`;
```
ewjoachim, line 53: This will work, but wouldn't just adding a "X" to the format string above be enough ? (https://date-fns.org/v2.21.3/docs/format)

ewdurbin: If I'm reading correctly those formatting options display UTC offsets in various ways, which while accurate... are not quite as human readable. Result from the proposed change reads like `America/New_York` for me.

Personally I'd prefer that (since it's identifiable to me as my timezone) over `-4 Z`

berinhard: Thanks for the review! I'll update the PR accordingly :+1:

berinhard: I'll also fix the breaking build. I didn't realize we do have tests for JS code (yaaaaaay)

berinhard: @ewdurbin and @ewjoachim I've updated the PR considering your suggestions. The only thing to notice is that we're kind of limited on format options with `date-fns` because Pypi is still using version `v1.19` and, thus, the "X" token is not available. 

But, from [1.19 docs](https://date-fns.org/v1.29.0/docs/format) I was able to use "Z" token instead.

tests/frontend/localized_time_controller_test.js
```diff
@@ -39,7 +39,7 @@ describe("Localized time controller", () => {
       expect(el).toHaveTextContent("Sep 20, 2019");
       // The expected ISO string in the title is localized
       const date = new Date(el.getAttribute("datetime"));
-      const expectedDate = format(date, "YYYY-MM-DD HH:mm:ss");
+      const expectedDate = format(date, "YYYY-MM-DD HH:mm:ss (Z)");
```
ewjoachim, line 42: Hm, is `Z` excpected to be the timezone here? Could we setup the test browser to be based somewhere in particular ?

thedrow:
Is there anything I can do to progress this PR further?

di:
Looks like the tests are failing due to the change in 8bbebdc.

thedrow:
I honestly don't get the assertion error. The values look the same to me.
https://github.com/pypa/warehouse/runs/5021666038?check_suite_focus=true#step:11:29-32

di:
They're off by one hour:

```diff
- title="2019-09-20 18:06:58 (+00:00)"
+ title="2019-09-20 19:06:58 (+00:00)"
```

tests/frontend/localized_time_controller_test.js
thedrow: It's called [`libfaketime`](https://github.com/wolfcw/libfaketime). We need to preload it before starting the browser and to set the correct environment variables. @di Would this be an appropriate solution to the test failure?

divbzero:
Looks like the change of `datetime` from `+0000` to `+0100` in commit 8bbebdc broke the static tests: 
https://github.com/pypa/warehouse/blob/8bbebdcc8936476438e7665c7c145845b9dca26d/tests/frontend/localized_time_controller_test.js#L28

Our localized-time controller assumes UTC because that’s how timestamps are stored in the database:
https://github.com/pypa/warehouse/blob/8bbebdcc8936476438e7665c7c145845b9dca26d/warehouse/static/js/warehouse/controllers/localized_time_controller.js#L22

divbzero:
We should also change the expected timezone in static tests back to `(+00:00)`:
https://github.com/pypa/warehouse/blob/8bbebdcc8936476438e7665c7c145845b9dca26d/tests/frontend/localized_time_controller_test.js#L45

The static tests run in our Docker image which is apparently set to UTC by default.

I have prepared a fix for the static tests that can be merged from [topic/timezone-to-alt-text](https://github.com/divbzero/warehouse/tree/topic/timezone-to-alt-text) or cherrypicked from df84524.

divbzero:
Thanks @di looks like the static tests are running smoothly now.

di:
Thanks @berinhard and @divbzero!

Issue #9503, opened 2021-05-10
Migrate all Sponsor data to the database.


ewdurbin:
**What's the problem this feature will solve?**

Currently due to copyright and trademarks that are incompatible with the license of this repository, PyPI uses a private repository [`pypi-theme`](https://github.com/python/pypi-theme) to store Sponsor logos.

These logos are rendered on the [sponsors page](https://pypi.org/sponsors) [using this template](https://github.com/pypa/warehouse/blob/main/warehouse/templates/pages/sponsors.html) and this [include](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sponsors.html).

We also render them in the [footer](https://github.com/pypa/warehouse/blob/main/warehouse/templates/base.html#L352) using this [include](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sponsors-footer.html) which references the dataset in [the aforementioned include](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sponsors.html).

And finally we render a subset of sponsor logos in [the sidebar](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sidebar-sponsor-logo.html) again using the same [aforementioned include](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sponsors.html).

At this point, the only reason we have the [`pypi-theme`](https://github.com/python/pypi-theme) repo is to house the Sponsor logos.

This makes @di and myself a bottleneck in updating sponsors and their logos on pypi.org.

**Describe the solution you'd like**

We should create a `Sponsor` model in the warehouse database that stores all of the various information encoded into the [sponsors include](https://github.com/pypa/warehouse/blob/main/warehouse/templates/includes/sponsors.html), as well as the ability to store an image in our object store that can later be referenced when rendering the page.

Then, we should update the rendering of sponsor logos throughout the site to use this model.

Additionally a new `is_sponsor_admin` column should be added to the [`User`](https://github.com/pypa/warehouse/blob/main/warehouse/accounts/models.py#L58) model which we will use to grant a new permission `sponsor_admin` in the [admin views](https://github.com/pypa/warehouse/tree/main/warehouse/admin) to a newly created admin panel to manage Sponsors. Users with `sponsor_admin` permissions should *only* be able to make changes to the Sponsor objects. You can see a similar behavior for the `moderator`/`is_moderator` permission/column in the admin app.

ewdurbin:
[...]

PR #9512, opened 2021-05-11
Move sponsors to DB

berinhard:
@ewdurbin and @dstufft I'm opening this draft PR so we can async talk about the progress of #9503.

Right now this PR introduces:

1. The new `Sponsor` model;
2. A command to load the sponsors into the database;

I'll add inline comments with specific questions I have.


warehouse/cli/sponsors.py
```diff
@@ -0,0 +1,592 @@
...
+@warehouse.group()  # pragma: no branch
+def sponsors():
+    """
+    Manage operations on top of sponsors.
+    """
+
+
+@sponsors.command()
+@click.pass_obj
+def populate_db(config):
```
berinhard, line 563: I wasn't sure if a click command was the best option here. From my django background, I'd rather go with a data migration, but I couldn't figure out how to do so with sqlalchemy yet. Please, let me know what you think about this.

warehouse/cli/sponsors.py
```diff
@@ -0,0 +1,592 @@
...
+        name = data['name']
+        sponsor = session.query(Sponsor).filter_by(name=name).one_or_none()
+        if sponsor:
+            print(f'Skipping {name} sponsor because it already exists.')
+            continue
+
+        params = data.copy()
+        img = params.pop('image')
+        params["color_logo_url"] = img
+        params["white_logo_url"] = f"white/{img}"
```
berinhard, line 582: @ewdurbin as [you've comment](https://github.com/pypa/warehouse/issues/9503#issuecomment-836912380), I'm adding the URLs directly, but I don't think we can totally rely on current URLs in Pypi. I mean, take a look at current Google's logo URL:

https://pypi.org/static/images/sponsors/color/google.62f52fb9.png

It has this hash `62f52fb9` suffix that I'm assuming it changes every deploy. So, how can we define the URLs here?

Also, what do you think about the directory structure I proposed: having the colored logo in a directory and all the white logos should be placed in the `white` children directory with the same file name.

warehouse/sponsors/models.py
```diff
@@ -0,0 +1,38 @@
...
+    __repr__ = make_repr("name")
+
+    name = Column(String, nullable=False)
+    service = Column(String)
+    # activity should be a list of strings
+    activity = Column(PickleType, nullable=False)
+
+    url = Column(String, nullable=False)  # TODO use URL field
+    color_logo_url = Column(String, nullable=False)  # TODO use URL field
+    white_logo_url = Column(String)  # TODO use URL field
```
berinhard, line 31: `sqlalchemy-utils` lib implements a [URLType](https://sqlalchemy-utils.readthedocs.io/en/latest/data_types.html#module-sqlalchemy_utils.types.url) that we can use here. But the [requirements structure](https://github.com/pypa/warehouse/tree/main/requirements) is not familiar to me and I don't know how to add the lib as a dependency. Can I have some help with that?

ewdurbin: https://warehouse.readthedocs.io/development/patterns.html#adding-new-dependencies

berinhard: Thanks! I've probably skiped this part from the docs =S

berinhard:
Hey everyone, I did some progress with this! 

As you can see, I changed the `/sponsors/` page to read the sponsors from the database by replacing the existing template route to a custom view. Although this is the simplest solution I could think and it works for that page, I'm afraid this is not true for the sidebar or footer logos.

From what I've saw in the code, I feel that the [`csi` template tag](https://github.com/pypa/warehouse/blob/main/warehouse/utils/html.py#L17) is a good candidate to go with. But, this will also introduce a conceptual change: now both sidebar and footers will require JS code to display the sponsors. Is this a problem?

I don't think it'll be, but I'm more comfortable asking before since I'm new in the project.

(please, do not consider the failing tests because I'm still under seeking for solutions mode. I'll work on them once I know my suggestions are valid alternatives)

warehouse/sponsors/models.py
ewdurbin: At first I was iffy about storing this information as a URL, since we will be accepting uploads.

However, now I see that this is indeed the best way to go! Once we have an object store configured to accept logo uploads, we can have the Form/View in admin accept an image upload, immediately place it in the object store and populate the URL fields. This reduces having to implement a "FileField"/"ImageField" like Django has... and is only minorly problematic if we ever change object stores.

One note: We should be displaying sponsor logos [via camo](https://github.com/pypa/warehouse/blob/7fc3ce5bd7ecc93ef54c1652787fb5e7757fe6f2/warehouse/filters.py#L66-L85), if we are not already, rather than adding an additional [image-src CSP entry](https://github.com/pypa/warehouse/blob/dffce426d48d779f5495c01547234993e3a3e523/warehouse/csp.py#L97-L102).

ewdurbin:
> As you can see, I changed the `/sponsors/` page to read the sponsors from the database by replacing the existing template route to a custom view. Although this is the simplest solution I could think and it works for that page, I'm afraid this is not true for the sidebar or footer logos.
> 
> From what I've saw in the code, I feel that the [`csi` template tag](https://github.com/pypa/warehouse/blob/main/warehouse/utils/html.py#L17) is a good candidate to go with. But, this will also introduce a conceptual change: now both sidebar and footer will require JS code to display the sponsors. Is this a problem?

I think that the CSI approach has merit, and shows significant consideration on your part to note the additional query necessary on every request to render sponsor information throughout the site. But I think the query for fetching sponsor name, description, inclusion areas (footer/sidebar), and logo url should be fairly lightweight and doable in a single sub millisecond query for each request. Given that, I the caveat of requiring JS to show our sponsor information might be worth the trade off of adding a new query to every request.

It's also more simple :)

warehouse/sponsors/models.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Sponsor(db.Model):
+    __tablename__ = "sponsors"
+    __repr__ = make_repr("name")
+
+    name = Column(String, nullable=False)
+    service = Column(String)
+    # activity should be a list of strings
+    activity = Column(PickleType, nullable=False)
```
ewdurbin, line 28: I see the note above, but we definitely do not want to be using Pickles <3

warehouse/sponsors/models.py
```diff
@@ -0,0 +1,39 @@
...
+
+    url = Column(URLType, nullable=False)
```
ewdurbin, line 30: perhaps rename to be more explicit like `link_url` or `landing_page_url`

warehouse/sponsors/models.py
dstufft: > One note: We should be displaying sponsor logos via camo, if we are not already, rather than adding an additional image-src CSP entry.

Are we planning to host the images ourselves, or are sponsors going to be responsible for hosting them?

warehouse/templates/pages/sponsors.html
```diff
@@ -117,7 +115,7 @@ <h2 class="sponsor-grid__sponsor-name">Your company here</h2>
         {% if sponsor.psf_sponsor %}
         <div class="sponsor-grid__sponsor">
           <div class="sponsor-grid__sponsor-img">
-            <img src="{{ request.static_path('warehouse:static/dist/images/sponsors/color/' + sponsor.image) }}" alt="{{ sponsor.name }}">
+            <img src="{{ sponsor.color_logo_url }}" alt="{{ sponsor.name }}">
```
ewdurbin, line 118: [camoify](https://github.com/pypa/warehouse/blob/7fc3ce5bd7ecc93ef54c1652787fb5e7757fe6f2/warehouse/filters.py#L66-L85) this url

warehouse/templates/pages/sponsors.html
```diff
@@ -142,7 +140,7 @@ <h2 class="centered-heading__title">Infrastructure sponsors</h2>
         {% if sponsor.infra_sponsor %}
         <div class="sponsor-grid__sponsor">
           <div class="sponsor-grid__sponsor-img">
-            <img src="{{ request.static_path('warehouse:static/dist/images/sponsors/color/' + sponsor.image) }}" alt="{{ sponsor.name }}">
+            <img src="{{ sponsor.color_logo_url }}" alt="{{ sponsor.name }}">
```
ewdurbin, line 143: [camoify](https://github.com/pypa/warehouse/blob/7fc3ce5bd7ecc93ef54c1652787fb5e7757fe6f2/warehouse/filters.py#L66-L85) this url

warehouse/templates/pages/sponsors.html
```diff
@@ -180,7 +178,7 @@ <h2 class="sponsor-grid__sponsor-name">Your company here</h2>
         {% if sponsor.one_time %}
         <div class="sponsor-grid__sponsor">
           <div class="sponsor-grid__sponsor-img">
-            <img src="{{ request.static_path('warehouse:static/dist/images/sponsors/color/' + sponsor.image) }}" alt="{{ sponsor.name }}">
+            <img src="{{ sponsor.color_logo_url }}" alt="{{ sponsor.name }}">
```
ewdurbin, line 181: [camoify](https://github.com/pypa/warehouse/blob/7fc3ce5bd7ecc93ef54c1652787fb5e7757fe6f2/warehouse/filters.py#L66-L85) this url

warehouse/sponsors/models.py
ewdurbin: we'll be hosting them ourselves, but throwing them in a public object storage bucket.

that url will be gross, and honestly an unnecessary addition to CSP since we can camoify them (which has the added benefit of making all this work locally)

dstufft: 👍

warehouse/sponsors/models.py
dstufft: postgres has an array type or you can use jsonb, or obviously a related table that has it's defaults set to join on load.

berinhard: Oh sure! I didn't realize we were using postgres! I'll change this by an array field then :+1

ewdurbin:
Note: We should ensure that we sanitize (for JavaScript an any esoteric tags/properties) any strings submitted via Admin before displaying on PyPI.org. This will ensure that even a hijacked Admin/Moderator/Sponsor Coordinator account won't result in an XSS for PyPI.

warehouse/sponsors/models.py
```diff
@@ -0,0 +1,59 @@
...
+                  src="{ self.white_logo_url }" alt="{ self.name }">'
+
+    @property
+    def activity(self):
+        """
+        Render raw activity markdown as HTML
+        """
+        if not self.activity_markdown:
+            return ""
+        return readme.render(self.activity_markdown, "text/markdown")
```
berinhard, line 59: @ewdurbin an update about sanitizing markdown content. Actually, this is already implemented in `readme` library. I'll link the pipeline of method calls so you can follow my steps to get to this conclusion:

1. This [`readme.render` method](https://github.com/pypa/warehouse/blob/main/warehouse/utils/readme.py#L31) is and utilitarian indirection to the real `readme.render` that's implemented at the `readme` library;
2. The `readme` library's `render` method returns a [clean HTML](https://github.com/pypa/readme_renderer/blob/main/readme_renderer/markdown.py#L70) instead of directly returning the HTML from the raw input;
3. The clean method defines a set of [allowed tags and attributes](https://github.com/pypa/readme_renderer/blob/main/readme_renderer/clean.py#L24) and it uses [bleach](https://pypi.org/project/bleach/) to sanitize the final HTML considering this set;

So, it seems to me that the markdown is already being sanitized internally by `readme` library. To feel more confident about this, I realized a few tests inputting malicious Markdown and none of the malicious behavior had success. The references I've used to run such tests were:

- [Exploiting XSS via Markdown](https://medium.com/taptuit/exploiting-xss-via-markdown-72a61e774bf8);
- [Markdown-XSS-Payloads](https://github.com/cujanovic/Markdown-XSS-Payloads/blob/master/Markdown-XSS-Payloads.txt)
- [Testing for XSS in Markdown Fields](https://makandracards.com/makandra/481451-testing-for-xss-in-markdown-fields)

Anyway, I've also added a simple unit test to document this at ed003dbe8b8dffe482cfe8e679af17e46445a605. Please, let me know if there's something else to do on this front.

ewdurbin: that looks excellent! 👍

warehouse/migrations/versions/590c513f1c74_new_psf_staff_boolean_flag.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "users",
+        sa.Column(
+            "is_psf_staff",
+            sa.Boolean(),
+            server_default=sa.text("false"),
```
ewdurbin, line 44: sa.sql.false()

ewdurbin (approved):
We reviewed this together on a call to finish our the last few details and get it shippable.

There is one remaining task to enable image uploads that will be handled in a follow-on Pull Request.

Thank you for the thorough work here @berinhard!

Makefile
```diff
@@ -147,6 +147,7 @@ initdb:
 	xz -d -f -k dev/$(DB).sql.xz --stdout | docker-compose run --rm web psql -h db -d warehouse -U postgres -v ON_ERROR_STOP=1 -1 -f -
 	docker-compose run --rm web python -m warehouse db upgrade head
 	$(MAKE) reindex
+	docker-compose run web python -m warehouse sponsors populate-db
```
di, line 150: Do we have a plan to run this in production?

As an aside, this is a bunch of work for something we'll only need to do once! Wow!

di, line 150: Do we have a plan to run this in production?

As an aside, this is a lot of work for something we only need once. Wow!

warehouse/accounts/__init__.py
```diff
@@ -106,6 +106,12 @@ def _authenticate(userid, request):
         principals.append("group:admins")
     if user.is_moderator or user.is_superuser:
         principals.append("group:moderators")
+    if user.is_psf_staff or user.is_superuser:
+        principals.append("group:psf_staff")
+
+    # user must have base admin access if any admin permission
+    if principals:
+        principals.append("group:with_admin_access")
```
di, line 114: This name confused me a bit because of our existing "admins" group, maybe something like:
```suggestion
        principals.append("group:with_admin_dashboard_access")
```
?

di (approved):
This looks solid! Nice work!

Makefile
ewdurbin: Yes, this will be run then axed.

berinhard:
Not sure why, but my last build failed but I'm not sure how to fix it. I'm still confused with the dependency management tools we're using. I tried to update the `main.txt` file using the `make` command explained [in the docs](https://warehouse.readthedocs.io/development/patterns.html#dependency-management), but the output says it's up to date:

```
$ make requirements/main.txt
make: 'requirements/main.txt' is up to date.
```

The error from CI is:

```
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
Collecting google-api-core<2.0.0dev,>=1.21.0
    google-api-core<2.0.0dev,>=1.21.0 from https://files.pythonhosted.org/packages/f6/72/6dc3c3c4576fedc409fd825bf71e22d0f448bd68143c79cdbd6216744282/google_api_core-1.30.0-py2.py3-none-any.whl#sha256=92cd9e9f366e84bfcf2524e34d2dc244906c645e731962617ba620da1620a1e0 (from google-cloud-core==1.6.0->-r requirements/main.txt (line 280))
```

But `google-api-core` do have 2 hash entries in `requirements/main.txt` and this branch is up to date the `main` one after @ewdurbin commit 96d9d4f592421ec25cc0fbe225c3137cbb29bfef.

I'm not sure what went wrong =S

ewdurbin:
dang, looks like we're hitting the google api dependency breakage again 😵

berinhard:
Not sure if this is related, but they've released a new package version 1 hour ago: https://pypi.org/project/google-api-core/#history

di:
Technically not `google-api-core`'s fault, just how https://github.com/pypa/pip/issues/9644 manifests here.

The solution is to upgrade to the latest version.

berinhard:
Thanks @di! I hope I managed to fix this on #9636

ewdurbin:
Right, not blaming the app itself, to rephrase, once again struck by https://github.com/pypa/pip/issues/9644, which is triggered by google-api dependency. (our only with extras)

ewdurbin:
aight! let's rebase and ship this!

Issue #9280, opened 2021-03-22
Gitlab secrets detection


ewjoachim:
**What's the problem this feature will solve?**
We've just closed #6051 🎉 allowing GitHub to warn us when it sees PyPI tokens publicly posted.
Let's do the same with GitLab, what do you think ?

**Describe the solution you'd like**
GitLab offers an equivalent feature, currently quite early stage. We can apply to start discussing by filling this form https://docs.google.com/forms/d/e/1FAIpQLSckOkRFMylEOb0KgBTzvLIOovHl6CPayqVhWT16h2z678Japg/viewform

Should we ?

The form kinda-implies that we could have a public token-revocation enpoint, which is true. Do you have an opinion on that ?

**Additional context**

ewjoachim:
I've gone ahead and started the work to have gitlab detect our secrets, which seems to be the first step towards automated reporting.

https://gitlab.com/gitlab-org/security-products/analyzers/secrets/-/merge_requests/104

ewjoachim:
Ok, they've requested that the form is filled so I'll do it. Here goes nothing 😅

ewjoachim:
So, I've failed to update this ticket, but we're in contact with GitLab people. they sent the API spec they're expecting (or rather the draft of the spec). I'm going to start doing things.

Here's the link: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/76812/diffs

ewjoachim:
I think that while implementing GitLab, it's interesting to also add a way to disclose tokens without being a trusted partner. First because this will give us a generic API for future integrations if need be, also because GitLab can be self hosted, and it would be nice if each instance can disclose tokens, and lastly because if anyone comes across a token, it's nice if they can disclose it easily. There might be a different rate limit for those disclosures.

The only thing that makes me hesitate is that if we do so, disclosers will be able to send a public url alongside with the disclosure, and this could be a phishing risk. Maybe we shouldn't display the url or maybe we should be extremely clear that people should exercise cauthing when clicking...

Any opinion ?

di:
This seems to be the issue on GitLab's side, which doesn't seem to have any activity in over 2 years unfortunately: https://gitlab.com/groups/gitlab-org/-/epics/8237

Our existing secret reporting API is now documented [here](https://docs.pypi.org/api/secrets/), hopefully this will help move things forward.

PR #9269, opened 2021-03-20
Notgithub: a service to simulate GitHub token scanning

ewjoachim:
Because manually testing the GitHub token scanning code in my dev environment was nontrivial to say the least, I've made a service to automate the complicated parts, and here's the integration of this service to the Warehouse dev environment.

Also, adds token scanning doc!

Capture d’écran 2021-03-20 à 18 40 50

Ping @greysteil : I believe this may be of interest to you and whoever gets to integrate GitHub token scanning, see https://github.com/ewjoachim/notgithub-token-scanning

PR #9552, opened 2021-05-20
Add package vulnerability reporting API

dekkagaijin:
PyPI today does not store information about known vulnerabilities. We’d like to build the necessary changes such that known vulnerabilities can be associated with package releases. The ultimate goal is that users of “pip” will be automatically warned if their dependencies are vulnerable.

See: https://github.com/pypa/warehouse/issues/9407, https://discuss.python.org/t/proposing-a-community-maintained-database-of-pypi-package-vulnerabilities/8374

Signed-off-by: Jake Sanders <jsand@google.com>


dekkagaijin:
Tests are passing, but <100% coverage. Please direct hatred at the changelist

ewjoachim:
Woo, I'm seeing a lot of interactions with the Token Disclosure code I contributed :)
If I may suggest a few general things:
- In the token disclosure code, there are 2 places where we take an arbitrary JSON payload and extract data from it (the meta API, and the disclosure payload), while trying very hard to check that it has the expected format. There isn't (yet) a lot of warehouse APIs where we accept JSON input payloads, so I had to do a little bit of work there, but if we're going toward refactoring and/or imitating this code, I think it would be much much easier to integrate a validation lib such as https://pypi.org/project/jsonschema/ (we could also go with Pydantic, Marshmallow or others like that but they're also quite oriented toward serialization and we're only interested in deserialization here). This may make your work much easier, and if you think it's worth it, I can try and make a first PR or draft to do that on the token disclosure side.
- It really wasn't easy to test the full token disclosure scenario (PyPI has many unit tests but fewer integration tests and it's not always easy to write those). I made the "notgithub" service (see #9269) exactly to ease that, and I'll be glad if you want to extend it to be capable of sending vulnerabilities given the format seems to be quite close. As of today "notgithub"'s only known use-case is Warehouse, so I don't have a problem if we add some coupling.
- Finally, around this part of the code, there are also discussions to integrate GitLab secrets detection too (#9280), but the format is not yet defined. If we move this code, let's make sure we make implementing future token scanning inputs easier rather than harder (I know it's not very actionnable advice, but in case of a doubt during implementation regarding re-usability, I hope this can provide some guidance)

dekkagaijin:
This is in a reviewable state, at least for a first pass. I'll be adding more using tests for 100% coverage of what exists and modeling `retrieve_public_key_payload` and `extract_public_keys` after what exists or the token leak API

dekkagaijin:
@di This is in a reviewable state, at least for a first pass. I'll be adding more using tests for 100% coverage of what exists and modeling retrieve_public_key_payload and extract_public_keys after what exists or the token leak API

warehouse/packaging/models.py
```diff
@@ -424,6 +450,13 @@ def __table_args__(cls):  # noqa
         passive_deletes=True,
     )
 
+    vulnerabilities = orm.relationship(
+        Vulnerability,
+        backref="releases",
+        secondary=lambda: release_vulnerabilities,
+        passive_deletes=True,
+    )
+
```
di, line 459: Should probably define this on the `Vulnerability` model instead.

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,200 @@
...
+
+class VulnerabilityReport:
+    def __init__(
+        self,
+        project: str,
+        versions: List[str],
+        vulnerability_id: str,
+        advisory_link: str,
+        aliases: List[str],
+    ):
```
di, line 38: I think we can probably condense the `Vulnerability` database model and this class into a single class. The class can also live at `warehouse/integrations/vulnerabilities/models.py` instead, and be imported into `warehouse.packaging.models`.

di (commented):
Preliminary review

warehouse/packaging/models.py
dekkagaijin: Would we not want both?

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,200 @@
...
+                .one()
+            )
+        except NoResultFound:
+            # TODO: metric
+            continue
+
+        if release not in vulnerability.releases:
+            vulnerability.releases.append(version)
+
+    # Delete any releases that no longer apply.
```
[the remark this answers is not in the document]
dekkagaijin: done

warehouse/packaging/models.py
```diff
@@ -316,6 +316,32 @@ class Dependency(db.Model):
...
+    ),
+    Column(
+        "vulnerability_id",
+        ForeignKey("vulnerabilities.id", onupdate="CASCADE", ondelete="CASCADE"),
+        nullable=False,
+    ),
+)
+
+
+class Vulnerability(db.Model):
```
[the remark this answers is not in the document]
dekkagaijin: done

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,200 @@
...
+    def retrieve_public_key_payload(self):
+        # TODO
+        pass
+
+    def extract_public_keys(self, pubkey_api_data):
+        # TODO
+        pass
+
+
+def _analyze_vulnerability(request, vulnerability_record, origin):
```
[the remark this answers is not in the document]
dekkagaijin: to leave the top-level `analyze_vulnerability` to catch and emit metrics for unexpected errors

warehouse/integrations/vulnerabilities/utils.py
dekkagaijin: As in, do the model representation, report validating, and DB interaction in a single class?

dekkagaijin:
@di I'm still finishing up the last few test cases, but please TAL

dekkagaijin:
Addddd green

Zyckzyck (approved):
It good

dekkagaijin:
I genericized the vuln report handler a bit more and split the OSV-specific code into a sub-repo

di (approved):
LGTM. I'll give this another ~day for other maintainers to review before merging.

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,147 @@
...
+    request.db.add(vuln_record)
+
+
+def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
+    try:
+        report = vulnerabilities.VulnerabilityReportRequest.from_api_request(
+            request=vulnerability_report
+        )
+    except vulnerabilities.InvalidVulnerabilityReportRequest as exc:
+        metrics.increment(f"warehouse.vulnerabilities.{origin}.error.{exc.reason}")
```
ewdurbin, line 66: general feedback on metrics calls: use `tags=[f"origin:{origin}"]` rather than embedding them in the metric name.

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class VulnerabilityRecord(db.Model):
+
+    __tablename__ = "vulnerabilities"
+
+    source = Column(String, primary_key=True)
+    id = Column(String, primary_key=True)
+    link = Column(String)
+    aliases = Column(ARRAY(String))
```
ewdurbin, line 53: it's unclear to me what an alias is in this context.

ewdurbin (commented):
Overall, this looks solid. I appreciate that the implementation is generic enough to handle additional implementations.

Given that, I do however have a few questions/concerns around ensuring that this modeling is sufficiently generic to handle additional implementations.

1) `origin` is currently defined as a string which is certainly useful for metrics/storage/analysis, but doesn't provide any user friendly value for display, and also doesn't provide us any graphical element to show next to anything we do end up displaying. this will mean that we'll have to create a lookup sometime down the line to map `{'osv': {'display_name': "Open Source Vulnerabilities", 'logo_url': "https://osv.dev..."}` etc. I don't think is is a blocker... but it might be good to include up front. I'm open to thoughts on how to approach that though... is it a separate model?

2) The model for `VulnerabilityRecord` is unsurprisingly simple. I appreciate that, but think we may want to include a JSONB column for any pertinent metadata that may want to be supplied/stored aside from the `link`. Again, not sure this is a blocker per se, but would create a more flexible mechanism for querying/filtering/analyzing these en masse without having to scrape links.

3) I have no idea what the `aliases` mean in this context and just don't want to commit to them being top level of the model if they are not a generic term for vulnerability reporting. My intuition says that these might just be things like CVEs? if not... perhaps they belong in the suggested metadata column from part 2?

4) I'm wondering if having some "short description" attached to the `VulnerabilityRecord` model would also be a good idea, again in the interest of being able to make any sense of them internally or at a later time for display without having to scrape links.

oliverchang:
> 3. I have no idea what the `aliases` mean in this context and just don't want to commit to them being top level of the model if they are not a generic term for vulnerability reporting. My intuition says that these might just be things like CVEs? if not... perhaps they belong in the suggested metadata column from part 2?

aliases is the alternate IDs a vulnerability is known as. e.g. The advisories from https://github.com/pypa/advisory-db/tree/main/vulns have `PYSEC-` IDs as the main ID, but they also have CVE IDs, or maybe GHSA IDs. This can be useful for deduplication if more vulnerability sources are added.

ewjoachim:
(I'll do my best to try and review this on Friday)

oliverchang:
> The model for VulnerabilityRecord is unsurprisingly simple. I appreciate that, but think we may want to include a JSONB column for any pertinent metadata that may want to be supplied/stored aside from the link. Again, not sure this is a blocker per se, but would create a more flexible mechanism for querying/filtering/analyzing these en masse without having to scrape links.

If we do this, perhaps this could be a JSON encoding of the format used at https://github.com/pypa/advisory-db/tree/main/vulns, which is where this integration will source from? This is a standard format which many vulnerability databases have started to adopt. 

On the other hand scraping this information as needed is also not hard, because there is a clear link to the source of these entries, and we can always backfill this data when needed?

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+    except NoResultFound:
+        if not report.versions:
+            return
+
+        vulnerability_record = VulnerabilityRecord(
+            id=report.vulnerability_id,
+            source=origin,
+            link=report.advisory_link,
+            aliases=report.aliases,
```
oliverchang, line 88: aliases, link etc might change. Could we always update these when we get a request even when the vulnerability_record already exists?

warehouse/integrations/vulnerabilities/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+                reason="format",
+            )
+
+        return cls(
+            request["project"],
+            request["versions"],
+            request["id"],
+            request["link"],
+            request["aliases"],
+        )
```
ewjoachim, line 62: I'd suggest either
```suggestion
        return cls(
            project=request["project"],
            versions=request["versions"],
            id=request["id"],
            link=request["link"],
            aliases=request["aliases"],
        )
```
Or
```suggestion
        return cls(**request)
```
Overall, relying on the order of parameters when there are more than one parameter usually leads to strange and hard-to-debug bugs :)

warehouse/integrations/vulnerabilities/osv/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+
+from warehouse import integrations
+from warehouse.integrations import vulnerabilities
+
+
+class OSVPublicKeyAPIError(vulnerabilities.InvalidVulnerabilityReportRequest):
+    pass
+
+
+OSV_PUBLIC_KEYS_URL = "https://osv.dev/public_keys/pypa"
```
ewjoachim, line 26: As of today, this URL isn't set up yet. You may want to make sure it's all good before merging (or at least before sending payloads our way)

warehouse/integrations/vulnerabilities/utils.py
ewjoachim: (That's my bad, I did this in the PR this is inspired from. We can fix either mine or both in a subsequent PR, I guess. Do you have more general rules how metrics should be used ?)

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        _add_vuln_record(request, vulnerability_record)
+
+    try:
+        project = _get_project(request, report)
+    except NoResultFound:
+        metrics.increment(f"warehouse.vulnerabilities.{origin}.error.project_not_found")
+        raise
+
+    for version in report.versions:
```
ewjoachim, line 98: I believe this may do an awful lot of SQL queries. Can we fetch all the releases at once ?

ewjoachim: Also, I believe if we have a project with 1000 releases (some projects issue a release at every commit) and a report with 1000 versions, the double loop we'll have will be quadratic (1000^2 iterations). This would be both quite readable and more efficient using sets:

```
report_versions = set(report.versions)
record_versions = {r.version for r in vulnerability_record.releases}

# Add missing releases
releases_to_add = request.db.query(Release).do_some_magic(
    versions in report_versions - record_versions
)
for release in releases_to_add:
    vulnerability_record.releases.append(release)

# Remove outdated releases
releases_to_remove = request.db.query(Release).do_some_magic(
    versions in record_versions - report_versions
)
for release in releases_to_remove
    vulnerability_record.releases.remove(release)
```

I'm not familiar enough with SQLAlchemy's API to tell if there is a way to do it in one go, but just to check, there's isn't a way to just reassign `vulnerability_record.releases` to a list of releases, right ?

warehouse/integrations/vulnerabilities/osv/__init__.py
ewjoachim: I think it may be a good idea to set this as a envvar defined app setting. I had to do that with GitHub too, this made testing much much easier

tests/unit/integration/vulnerabilities/osv/test_package.py
```diff
@@ -0,0 +1,466 @@
...
+        )
+        with pytest.raises(osv.OSVPublicKeyAPIError) as exc:
+            vuln_report_verifier.retrieve_public_key_payload()
+
+        assert str(exc.value) == "Invalid response code 418: I'm a teapot"
+        assert exc.value.reason == "public_key_api.status.418"
+
+    def test_retrieve_public_key_payload_json_error(self):
+        response = pretend.stub(
+            text="Still a non-json teapot",
```
ewjoachim, line 249: (Is it bad that I laughed at this given this is a copy of a joke I made :D ) (👍)(🍵)

ewjoachim:
(Finished this round of review. I forgot to make a general comment. I appreciate that there was a lot of code inspired from my PR, but if we need to do this a 3rd time, I'm afraid that will start to make quite a lot of duplication. This is by no means a way to block this PR (https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) but we can think already about how we can avoid duplicating that much code next time)

warehouse/integrations/vulnerabilities/__init__.py
dekkagaijin: done

warehouse/integrations/vulnerabilities/osv/__init__.py
ewdurbin: This is still an empty/useless page which blocks merge.

warehouse/integrations/vulnerabilities/utils.py
ewdurbin: I would still like to see tagging implemented before merge. As this is the initial implementation and we are likely to see subsequent ones, they should be able to reference this as a best practice.

We should patch up the GitHub metrics in a separate PR @ewjoachim

warehouse/integrations/vulnerabilities/utils.py
ewdurbin: I want to second this concern. To my eye this rill create an "N+1" query situation. I believe the thing desired is something in https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html

ewdurbin (changes requested):

dekkagaijin:
> I would still like to see tagging implemented before merge.

Tagging?

ewdurbin:
> > I would still like to see tagging implemented before merge.
> 
> Tagging?

see the comment that is in reply to https://github.com/pypa/warehouse/pull/9552#discussion_r662557423

warehouse/integrations/vulnerabilities/utils.py
dekkagaijin: added a joined load for releases when fetching the top-level vulnerability record

dekkagaijin:
> I would still like to see tagging implemented before merge.

done

warehouse/integrations/vulnerabilities/utils.py
dekkagaijin: done

warehouse/integrations/vulnerabilities/osv/__init__.py
dekkagaijin: endpoint is up

warehouse/integrations/vulnerabilities/models.py
dekkagaijin: added descriptive comments for `link` and `aliases`

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
```diff
@@ -0,0 +1,70 @@
...
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "1dbb95161e5a"
+down_revision = "08ccc59d9857"
+
+# Note: It is VERY important to ensure that a migration does not lock for a
```
ewdurbin, line 28: comment(s) from template can be removed (not critical)

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
```diff
@@ -0,0 +1,70 @@
...
+            ["vulnerabilities.source", "vulnerabilities.id"],
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ewdurbin, line 67: comment(s) from template can be removed (not critical)

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
```diff
@@ -0,0 +1,70 @@
...
+#       not break compatibility with the *previous* version of the code base.
+#       This is because the migrations will be ran automatically as part of the
+#       deployment process, but while the previous version of the code is still
+#       up and running. Thus backwards incompatible changes must be broken up
+#       over multiple migrations inside of multiple pull requests in order to
+#       phase them in over multiple deploys.
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
ewdurbin, line 39: comment(s) from template can be removed (not critical)

warehouse/integrations/vulnerabilities/utils.py
ewdurbin: Thanks, it's still not super clear to me how efficient the whole situation will be, but this should help!

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    # Unassociate any releases that no longer apply.
+    for release in list(vulnerability_record.releases):
+        if release.canonical_version not in report.versions:
+            vulnerability_record.releases.remove(release)
+
+
+def analyze_vulnerability(request, vulnerability_report, origin, metrics):
+    metrics.increment("warehouse.vulnerabilities.received", tags=[f"origin:{origin}"])
+    try:
+        _analyze_vulnerability(
```
ewdurbin, line 126: let's wrap this call in `with metrics.timed('warehouse.vulnerabilities.analysis', tags=[f"origin:{origin}"]):`

this will emit timing metrics so we can keep tabs on this programatically.

ewdurbin (approved):
one request (the timing metric tag) and some suggestions. none of it strictly critical for merge.

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
dekkagaijin: done

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
dekkagaijin: done

warehouse/migrations/versions/1dbb95161e5a_add_vulnerability_and_release_.py
dekkagaijin: done

warehouse/integrations/vulnerabilities/utils.py
dekkagaijin: done, good idea

PR #9555, opened 2021-05-20
Add admin shortcuts to project detail page

di:
Adds some shortcuts to the project detail page that are only viewable by moderators/admins.

Looks like this:

image

### DIFF
--- a/tests/unit/admin/views/test_includes.py
+++ b/tests/unit/admin/views/test_includes.py
@@ -1,0 +1,22 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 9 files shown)


dstufft:
I think this will visually work, but will end up executing a 403? or something like that for everyone who isn't an admin user when they load one of these pages. I also think that 403 pages aren't cached. We might want to move the permission check into the template, and just render a blank page if the user isn't an admin.

di:
@dstufft We should probably do the same for this existing view as well, b/c it's currently 403-ing on most project pages: https://github.com/pypa/warehouse/blob/7fc3ce5bd7ecc93ef54c1652787fb5e7757fe6f2/warehouse/packaging/views.py#L136-L145

dstufft:
Yea that seems like a good idea.

PR #9675, opened 2021-06-16
Sort API tokens by name in /manage/account/

### LINKED ISSUE #9674: API tokens on https://pypi.org/manage/account/ appear to be unsorted
**Describe the bug**

We use a zope.wheelbuilder account to upload binary wheels from various CI systems (Appveyor, GHA, formerly Travis).  Thus it has a bunch of API tokens (one for each project for each CI system).  These tokens appear to be listed in random order, which makes it difficult to see whether a particular project already has a token or not.

Ekrano nuotrauka iš 2021-06-16 12-58-18

**Expected behavior**

I would expect to see the tokens sorted by name. or at least by something (creation time/last usage time/scope all make sense to me).

**To Reproduce**

Create a bunch of API tokens and visit https://pypi.org/manage/account/

mgedmin:
Closes #9674.

### DIFF
--- a/warehouse/templates/manage/account.html
+++ b/warehouse/templates/manage/account.html
@@ -533,7 +533,7 @@
         </tr>
       </thead>
       <tbody>
-        {% for macaroon in user.macaroons %}
+        {% for macaroon in user.macaroons | sort(attribute="description") %}
           {{ api_row(macaroon) }}
         {% endfor %}
       </tbody>


ewjoachim (approved):
That looks about right. Doing the `.order_by` with the ORM would not be really jinja-friendly. We could also set a default order_by on the `user.macaroons` relationship and get rid of this problem more generally, but I believe this should work. (EDIT: ah, that what you said too in https://github.com/pypa/warehouse/issues/9674#issuecomment-862305078 )

Given there's no unit test, would you mind taking a screenshot of the result ?

mgedmin:
Re: screenshot -- I still haven't found the time to bring up a dev env on my laptop.

Re: https://github.com/pypa/warehouse/issues/9674#issuecomment-872583099, it looks like this PR is not what's wanted by the maintainers anyway.

Issue #9700, opened 2021-06-21
Missing info when moving From xmlrpc to restful API


4383:
**Describe the bug**
On openstack, we noticed missing data during our transition from the xmlrpc to the restful API.

Indeed, few months ago, we started transitioning github.com/openstack/releases on the restful API.

Previously we relied on the [`package_role` method](https://warehouse.pypa.io/api-reference/xml-rpc.html#package-querying) to retrieve roles. Role is either Maintainer or Owner. The original intent here was to check whether an existing package registration allowed the account used by our automation to upload releases. 

Example:

```
>>> import xmlrpc.client
>>> client = xmlrpc.client.ServerProxy("https://pypi.org/pypi")
>>> client.package_roles('ansible-role-atos-hsm')
[['Owner', 'openstackci']]
```

Unfortunately the restful API doesn't seems to return similar information. A `maintainer` field is returned but this one is empty

```
>>> import requests
>>> data = requests.get('https://pypi.org/pypi/ansible-role-atos-hsm/json')
>>> print(data.json()['info']['maintainer'])

>>> # Empty result
```

So, we don't know how to finalize our transition without these data.

**Expected behavior**

Does the fact that the `data.json()['info']['maintainer']` is empty is normal?

Do we should expect an empty result here?

Is it possible to bind something else (another JSON key) to retrieve this information?  

**To Reproduce**

By using xmlrpc:

```
>>> import xmlrpc.client
>>> client = xmlrpc.client.ServerProxy("https://pypi.org/pypi")
>>> client.package_roles('ansible-role-atos-hsm')
[['Owner', 'openstackci']]
```

By using restful

```
>>> import requests
>>> data = requests.get('https://pypi.org/pypi/ansible-role-atos-hsm/json')
>>> print(data.json()['info']['maintainer'])

>>> # Empty result
```

**My Platform**

**Additional context**

To see the full context of our transition and the full context of the experienced problem, then, please have a look to https://review.opendev.org/c/openstack/releases/+/785008

Thanks for your help.

di:
The `maintainer` field in the JSON API corresponds to the [`Maintainer`](https://packaging.python.org/specifications/core-metadata/#maintainer) metadata field that is provided in the package metadata, not the maintainers on PyPI.

There is no equivalent API for `package_role`.

4383:
So, if no equivalent exist, I think that we can close this issue.

Thank you for your reply.

di:
[...]

PR #9770, opened 2021-07-08
[feature] transition the package_roles method onto the restful API

4383:
These changes adds the `package_roles` as results of the return of the
`admin.views.project` restful API.

The `package_roles` method was available into the xmlrpc, unfortunately
these informations are missing into the restful API [1].

These changes adds the creation of an equivalent into the restful API.

[1] https://github.com/pypa/warehouse/issues/9700

### DIFF
--- a/tests/unit/admin/views/test_projects.py
+++ b/tests/unit/admin/views/test_projects.py
@@ -99,6 +99,7 @@ class TestProjectDetail:
             "project": project,
             "releases": [],
             "maintainers": roles,
+            "package_roles": [],
             "journal": journals[:30],
             "ONE_MB": views.ONE_MB,
             "MAX_FILESIZE": views.MAX_FILESIZE,
--- a/warehouse/admin/views/projects.py
+++ b/warehouse/admin/views/projects.py
@@ -111,6 +111,16 @@ def project_detail(project, request):
         )
     ]
     maintainers = sorted(maintainers, key=lambda x: (x.role_name, x.user.username))
+    package_roles = [
+        role
+        for role in (
+            request.db.query(Role)
+            .join(User, Project)
+            .filter(Project.normalized_name == project)
+            .order_by(Role.role_name.desc(), User.username)
+            .all()
+        )
+    ]
     journal = [
         entry
         for entry in (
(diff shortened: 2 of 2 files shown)


tests/unit/admin/views/test_projects.py
```diff
@@ -99,6 +99,7 @@ def test_gets_project(self, db_request):
             "project": project,
             "releases": [],
             "maintainers": roles,
+            "package_roles": [],
```
4383, line 102: I'm not sure if we want to create some representations by using the `RoleFactory`. In this case what should be the difference between `maintainers` and `package_roles`?

4383:
These changes aim to address #9700

4383:
Hello team,

I hate harassing people for getting reviews, but, at least, can we have a minimal analyse from your side just to know if this is the right thing to do to fix #9700 and get some visibility on our side.

Thanks for your understanding.

abitrolly:
JSON API at https://warehouse.readthedocs.io/api-reference/json.html#get--pypi--project_name--json promises this.

> Metadata returned comes from the values provided at upload time and does not necessarily match the content of the uploaded 

And `package_roles` is not a part of this metadata.

I don't think it is a good idea to skew the data that goes into `[packaging] -> [uploading] -> [storing] -> [accessing through API]` pipelines without documenting all the transformations that may come unexpected for people. I mean https://github.com/pypa/warehouse/pull/3520 needs to be merged first to solve `[packaging] -> [uploading]` disability, and then a similar mapping will need to be done to decouple fields in REST API from what people have in packages and uploads.

Or, make a separate namespace that will expose PyPI admin interface for a package, allowing to assign maintainership roles etc. Right now PyPI evolution stuck between going into central command and control server and providing a more decentralized alternative. When all information including maintainers was contained in the package, the management was decentralized. When people started to need the account and configure access on the server side, the management is centralized.

Given the above, I would not just make a separate namespace in the JSON, like "pypi_rbac" or "pypi_config", but a separate endpoint like `api/v2/config/<project_name>/access`, which will only manage PyPI settings. Because you would likely need the API to limit role scopes, assign token, check their validity, audit logs.

4383:
Thanks for your interesting feedback

firefly9966 (approved):

miketheman:

PR #9785, opened 2021-07-13
prohibit automated password reset for specified users

ewdurbin:


di (approved):

dstufft (approved):

warehouse/accounts/models.py
```diff
@@ -163,6 +166,17 @@ def recent_events(self):
             .all()
         )
 
+    @property
+    def password_reset_prohibited(self):
```
ewjoachim, line 170: It's minor, but from my experience, things tend to get clearer when booleans are expressed through something positive, not negative (e.g. `can_reset_password`) otherwise it tends to involve double-negatives.

warehouse/migrations/versions/08ccc59d9857_add_manual_password_reset_prohibition.py
```diff
@@ -0,0 +1,41 @@
...
+def upgrade():
+    op.add_column(
+        "users",
+        sa.Column(
+            "prohibit_password_reset",
+            sa.Boolean(),
+            server_default=sa.text("false"),
+            nullable=False,
+        ),
+    )
```
ewjoachim, line 37: I'm not an alambic expert, but I believe adding a non-nullable column will likely force PG to rewrite every row (522,487 according to the front page). I trust your judgement & experience on this, but on my part, I'd tend to say it might be worth it to try and run this migration during a time where a few seconds freeze of the DB will likely not be seen.

(the way of doing this properly is most likely overkill if we don't have several million rows)
(and again, maybe alembic actually takes care of this, I don't know)

ewdurbin: we've run similar migrations using the `server_default` incantation in the past with no impact, see `is_psf_staff` addition a few weeks back. it's more on the PG side than the alembic/sqlachemy side, but using a server_default Postgres knows what to do.

warehouse/accounts/models.py
ewdurbin: 👍🏼 similar feedback from someone else so I'll refactor.

warehouse/accounts/models.py
```diff
@@ -163,6 +166,17 @@ def recent_events(self):
...
         )
 
+    @property
+    def can_reset_password(self):
+        return not any(
+            [
+                self.is_superuser,
+                self.is_moderator,
+                self.is_psf_staff,
+                self.prohibit_password_reset,
```
ewjoachim, line 176: I not that you didn't change the name of the column. I guess this is voluntary?

ewdurbin: This is configured via admin only. So from the administrators perspective, flipping the bit prohibits password reset for a user. In this case it would be "sentinel" users like `admin` and `pypi` which we _do not_ want to grant any special privileges.

ewjoachim: Ok, it makes sense

PR #9852, opened 2021-07-30
make requirements/*.txt output to stdout

ewjoachim:
Also, fixes a doc inconsistency on the subject

### DIFF
--- a/docs/development/patterns.rst
+++ b/docs/development/patterns.rst
@@ -81,7 +81,7 @@ process for adding new dependencies:
 1. Remove the project name from the appropriate ``.in`` file
 2. From the repositories root directory, recompile the dependencies for each modified ``.in`` file::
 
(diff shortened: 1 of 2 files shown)


docs/development/patterns.rst
```diff
@@ -81,7 +81,7 @@ process for adding new dependencies:
 1. Remove the project name from the appropriate ``.in`` file
 2. From the repositories root directory, recompile the dependencies for each modified ``.in`` file::
```
di, line 82: I think this line needs updated too

ewjoachim: Hm ? I'm not sure I understand how it should be replaced?

The best I understand from this is that you imply we should talk about the `.txt` file ? Like:
```suggestion
2. From the repositories root directory, recompile the ```.txt` for each modified ``.in`` file::
```

di: Sorry, read this too fast, you're right

ewjoachim:
(Thank you ❤️ )

PR #9860, opened 2021-08-03
Add a blog

di:
This PR adds a basic blog to PyPI:

 warehouse localdev_blog_
warehouse localdev_blog_03-08-2021-hello-world

In draft for now to get early thoughts before writing the tests.

### DIFF
--- a/blog/03-08-2021-hello-world.md
+++ b/blog/03-08-2021-hello-world.md
@@ -1,0 +1,18 @@
+# Hello World!
+_Published August 3rd, 2021_
+
+Welcome to the PyPI blog!
+
+We've got:
+* lists
+* of
+* things
+
+as well as:
+```
+code blocks
+for your
+codes
+```
+
+and *all* the `cheese` you can **eat**!
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,6 +87,7 @@ services:
       - .coveragerc:/opt/warehouse/src/.coveragerc:z
       - packages:/var/opt/warehouse/packages
       - sponsorlogos:/var/opt/warehouse/sponsorlogos
+      - ./blog:/opt/warehouse/src/blog:z
       - ./bin:/opt/warehouse/src/bin:z
     ports:
       - "80:8000"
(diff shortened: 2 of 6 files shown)


dstufft:
Is there a reason to do this versus something like ``blog.pypi.org`` using something out of the box? Not opposed, just feels like feature creep that could be better solved by something out of the box, or even some other service.

di:
I can think of a few advantages:

* we wouldn't have yet another system/service to maintain/learn/depend on
* something like django or wagtail is probably overkill for what we actually need here
* blog post content in the codebase means when we merge a new feature we can simultaneously publish a post for it
* anyone can author a blog post by making a PR, no special access needs to be granted
* similarly, blog posts are easy to review (and anyone can review them)

The last two are my main gripes with the PSF blog. Also, we've been talking about creating a blog for a while and it hasn't happened, and I think this adds the least amount of overhead while also producing something we can publish to in the future.

ewdurbin:
I'm +0 on this. I agree with @di that _something_ is better than nothing at this point and all of the major advantages of this lightweight approach do resonate with me and my experience with blogger. (not running another service ourselves is also preferable).

Given that this implementation more or less just renders templated files, I don't think this is overkill... yet. However, we'll probably want support for author/bylines, tags... then pagination/nav... then...?

The question becomes how far we push this before something like blogger/django/wagtail is _not_ overkill.

ewdurbin:
Another thought is serving from pypi itself has the advantage of keeping the theme/style consistent without a bunch of hoops to jump through.

di:
> However, we'll probably want support for author/bylines, tags... then pagination/nav... then search... then...?

I think we could support author/bylines the same way this is supporting the date/summary right now. (I'm also not convinced it's totally necessary.)

I agree we'll probably want to add pagination someday, but this would just be a matter of slicing the sorted list of filenames, or slicing them by year, etc. I still think the long-term overhead is going to be lower with this approach though.

ewjoachim:
What about adding a [Pelican](https://blog.getpelican.com/) (or Jekyll or Hugo or...) blog inside the folders hierarchy but not coding it ourselves ? We'd have a static website generator, authors/posts/... without the hustle of having to code it ourselves.

(Pelican is Python so we'd be supporting a Python project :) )

Among the features that would help make the blog more useful from day 1 is a RSS feed IMHO, especially given those can be plug into Slack channels.

ewjoachim:
I just realized we already have a static site generator in the project codebase: the sphinx doc :D

Do you believe we could leverage it?

PR #9908, opened 2021-08-11
Add shorthands for `/user` and `/simple`

tusharsadhwani:
Inspired by the `/p` shorthand for the `/project` endpoint.

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -418,7 +418,9 @@ def test_routes(warehouse):
 
     assert config.add_redirect.calls == [
         pretend.call("/sponsor/", "/sponsors/", domain=warehouse),
+        pretend.call("/u/{username}/", "/user/{username}/", domain=warehouse),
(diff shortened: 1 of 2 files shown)


di:
Thanks! I've been meaning to add `/u` for ages. Could you update the tests as well?

tusharsadhwani:
@di I would, if I hadn't painfully made this PR from my phone. I'll do this when I wake up tomorrow :)

ewjoachim (approved):

ewdurbin:
Ran across this when researching another issue. I'm not a huge fan of this for "API" endpoints like the `/s/` for `/simple/`, and even less so as a blind redirect that won't be cached at the edge.

@di thoughts on removing this? In a cursory overview it doesn't appear to be in regular use, and I'd like to remove it before it happens to.

tusharsadhwani:
If /u is staying, I'm completely fine with it too

di:
Agreed.

Issue #8254, opened 2020-07-13
Expose the METADATA file of wheels in the simple API


dstufft:
Currently a number of projects are trying to work around the fact that in order to resolve dependencies in Python you have to download the entire wheel in order to read the metadata. I am aware of two current strategies for working around this, one is the attempt to use the PyPI JSON API (which isn't a good solution because it's non standard, the data model is wrong, and it's not going to be secured by TUF) and the other is attempting to use range requests to fetch only the ``METADATA`` file from the wheel before downloading the entire wheel (which isn't a good solution because TUF can currently only verify entire files, and it depends on the server supporting range requests, which not every mirror is going to support).

It seems to me like we could side step this issue by simply having PyPI extract the ``METADATA`` file of a wheel as part of the upload process, and storing that alongside the wheel itself. Within TUF we can ensure that these files have not been tampered with, by simply storing it as another TUF secured target. Resolvers could then download just the metadata file for a wheel they're considering as a candidate, instead of having to download the entire wheel.

This is a pretty small delta over what already exists, so it's more likely we're going to get it done than any of the broader proposals of trying to design an entire, brand new repository API or by ALSO retrofitting the JSON API inside of TUF.

The main problems with it is that the ``METADATA`` file might also be larger than needed since it contains the entire long description of the wheel and that it still leaves sdists unsolved (but they're not currently really solvable). I don't think either problem is too drastic though.

What do folks thinks? This would probably require a PEP and I probably don't have the spare cycles to do that right now, but I wanted to get the idea written down incase someone else felt like picking it up.

@pypa/pip-committers @pypa/pipenv-committers @sdispater (not sure who else work on poetry, feel free to CC more folks in).

pfmoore:
Sounds like a good idea. It would probably need to be an optional feature of the API, as we need to keep the spec backward-compatible, and really basic "serve a directory over HTTP" indexes might not be able to support the API.

But that is a minor point. Basically +1 from me.

One pip-specific note on timing, though. It looks like pip will get range request based metadata extraction before this API gets formalised/implemented. That's fine, but I think that when this *does* become available, pip should drop the range approach and switch to just using this. That would be a performance regression for indexes that support range requests but not the new API, but IMO that's more acceptable than carrying the support cost for having both approaches.

uranusjr:
I agree, it seems like a reasonable solution. If we design how the metadata is listed carefully, it’d likely also be reasonable for the files-in-a-directory use case to optionally implement.

[...]

PR #9972, opened 2021-08-30
Extract and host wheel METADATA on upload

abitrolly:
Extract METADATA on wheel upload as *.whl.metadata (#8254)

This allows to download just the .metadata file for dependency
resolution instead of full wheels as it happens today
https://github.com/pypa/warehouse/issues/8254#issuecomment-908327875

The filename convention and download location is covered by PEP-0658
https://www.python.org/dev/peps/pep-0658/#specification


warehouse/forklift/legacy.py
```diff
@@ -1330,11 +1331,19 @@ def file_upload(request):
                         "Binary wheel '{filename}' has an unsupported "
                         "platform tag '{plat}'.".format(filename=filename, plat=plat),
                     )
+            # Extract .metadata file
+            # https://www.python.org/dev/peps/pep-0658/#specification
+            with zipfile.ZipFile(temporary_filename) as zfp:
```
pradyunsg, line 1336: What happens if the user uploads a non-zip file as a wheel? Is it possible to have zip files that effectively DoS the server with a maliciously crafted file?

(that's totally possible, especially if someone's trying to break PyPI or something)

abitrolly: The check if the uploaded wheel is valid is made here - https://github.com/pypa/warehouse/blob/b9c2e2a45529b64a44c4e8b225c20efab0442bfd/warehouse/forklift/legacy.py#L711 - It could be extended to include all kind of checks, but..

> Is it possible to have zip files that effectively DoS the server with a maliciously crafted file?

Yes, because , but [`zipfile` that doesn't protect against zip bombs](https://docs.python.org/3/library/zipfile.html#resources-limitations). We can try to throw a zip bomb at PyPI, but such security hazards are better be managed in another issue - this PR doesn't increase the attack surface.

abitrolly:
Can't say I didn't expect the tests to fail, but I had a hope that the new functionality will just not be covered.

Test are failing, because they are monkeypatching a lot of calls including `.whl` file itself, and because new `zipfile.ZipFile` is not monkeypatched, it fails with `E           zipfile.BadZipFile: File is not a zip file`. Looks like I need to figure out monkeypatching section and copy/paste it to 29 failing tests. A price to pay for linear readable piece of code.

abitrolly:
I would really prefer a fixture with real wheel, to ensure that my code works. Otherwise I feel like I am testing my expectations how my code should work, and not what would really happen after deployment.

abitrolly:
It appeared that the fix needs to be applied to just one test - 29 failures are just parametrisation for different platforms.

It is now ready for review.

abitrolly:
Fixed `black` warning and rebased.

abitrolly:
@pradyunsg can you unblock the tests?

warehouse/forklift/legacy.py
```diff
@@ -1330,11 +1331,19 @@ def file_upload(request):
...
+                metafile = wheel_info.group("namever") + ".dist-info/METADATA"
```
ewjoachim, line 1337: Just curious, is this the way PIP reads the metadata ? Is there a way we have multiple metadata files in a wheel such that the metadata says we depend on package A, but in reality when installing via pip, we depend on package B ?

ewjoachim: Here's how pip finds the .dist-info folder as far as I can tell https://github.com/pypa/pip/blob/bf91a079791f2daf4339115fb39ce7d7e33a9312/src/pip/_internal/utils/wheel.py#L84-L114

abitrolly: Maybe upload API should do these checks too. Right now it checks for the `WHEEL` presence and allowed platform tags .

https://github.com/pypa/warehouse/blob/e5cb0e251898e2b60e9d801054d97e12ae3b8e7b/warehouse/forklift/legacy.py#L711-L727

abitrolly: > Is there a way we have multiple metadata files in a wheel such that the metadata says we depend on package A, but in reality when installing via pip, we depend on package B ?

According to this code `pip` should fail if there are multiple `.dist-info` dirs
https://github.com/pypa/pip/blob/bf91a079791f2daf4339115fb39ce7d7e33a9312/src/pip/_internal/utils/wheel.py#L100

ewdurbin (commented):
I think the major concern missing so far in implementation is storing a boolean on the [File](https://github.com/pypa/warehouse/blob/442148581c61163a2c30bca25f8ef3f06457e932/warehouse/packaging/models.py#L509) model and rendering the `data-dist-info-metadata` value in the [simple detail](https://github.com/pypa/warehouse/blob/442148581c61163a2c30bca25f8ef3f06457e932/warehouse/templates/legacy/api/simple/detail.html#L23) template/view per [the PEP](https://www.python.org/dev/peps/pep-0658/#specification).

The rendering isn't strictly required for this PR, but we should at least be able to keep track of what files we _have_ stored metadata for.

Ideally this implementation would be as a general purpose utility that we could point at a given wheel file on disk, that would allow for extracting/storing the metadata during uploads, but *also* allow us to backfill over time.

ewdurbin:
Oh, hmmm, and it also seems like we should be calculating/storing a hash for the metadata file.

abitrolly:
> The rendering isn't strictly required for this PR, but we should at least be able to keep track of what files we have stored metadata for.

Yes, I would like to avoid big changes and https://warehouse.readthedocs.io/development/submitting-patches.html recommends keeping them small. After this PR is merged, all new wheels will have the metadata stored, so the upload datetime will be enough for now.

> Ideally this implementation would be as a general purpose utility that we could point at a given wheel file on disk, that would allow for extracting/storing the metadata during uploads, but also allow us to backfill over time.

Yes, this should be available as a library, but I don't know which. Exactly for backfilling job.

> Oh, hmmm, and it also seems like we should be calculating/storing a hash for the metadata file.

That's not a problem to calculate them when DB part is ready.

warehouse/forklift/legacy.py
uranusjr: *Technically* a wheel can contain multiple `.dist-info` directories as long as only one matches the name-version pair in the wheel’s file name. But pip adds this additional restriction to simplify implementation, and afaik no-one has complained yet.

abitrolly: Those extra `.dist-info` directories will not be used, because [a wheel](https://www.python.org/dev/peps/pep-0427/) ships only one `package-version` build, and the `.dist-info` that will is used should match the `name-version` pair in filename.

abitrolly:
@ewdurbin looks like it is easy to add hashes to `.metadata` upload. Are all three are needed?

abitrolly:
Oh no. There is no `File` entry created for `.metadata` file. A migration will be needed in any case. It it possible not to touch the DB in the PR, splitting it into 3 instead?

1. Extract and host metadata (this one)
2. DB changes to store metadata hash
3. View changes to show metadata hash in simple API

dstufft:
I would probably change the order of those slightly to add the DB changes first. Since storing the data doesnt help us without data in the DB to tell us that there is a metadata file associated with this release.

abitrolly:
@dstufft all new wheel files after this change will have the metadata file, so it is not strictly necessary to store info that metadata exists in DB. The only thing that needs to be stored is the hash, but I would prefer to have a the discussion about that in a separate issue.

If you can point me to example how to traverse storage filesystem, I can try to make migration script that will check and add metadata for the wheels that are already uploaded.

ewdurbin:
I agree with @dstufft that without DB storage... extracting/storing the metadata file isn't as useful unless the `data-dist-info-metadata` part of the PEP is implemented, clients would need to issue many requests that would end with a 404.

Also for backfill, rather than needing to traverse our file storage, we'd only need to iterate over `File` objects where `packagetype` is `bdist_wheel` and `has_metadata` is NULL  or `metadata_hash` is NULL to backfill.

abitrolly:
@ewdurbin I can try to add NULLable `metadata_hash`.

Is it enough to just add new field to `File` class https://github.com/pypa/warehouse/blob/3802138b4eed60a1b8218168919e6de3792e9105/warehouse/packaging/models.py#L558-L560 and then generate migration as described in https://warehouse.readthedocs.io/development/database-migrations.html ?

abitrolly:
While I am working on DB migrations, it will be nice to get the https://github.com/pypa/warehouse/pull/9993 merged. It removes the `xz` requirement, and allows me to test changes with `podman-compose`.

abitrolly:
@dstufft @ewdurbin added DB migration and hash storage in DB.

ofek:
Any update on this?

uranusjr:
The extraction logic looks good to me, but I don’t have enough expertise to decide how information should be stored and pass around internally in Warehouse.

abitrolly:
@uranusjr it is no problem for me to split the PR into separate parts to get at least something merged as it already 2 months since I've added that code for @dstufft and @ewdurbin with no feedback from them.

pradyunsg:
FYI the tests are failing.

abitrolly:
@pradyunsg logs look like it's CI issue, as if caches are surviving rebases, and Alembic complains, but from the other hand tests should not use such cashes. I need to resurrect the local dev setup, but I am pretty sure tests were running ok locally.

di:
> `alembic.script.revision.MultipleHeads: Multiple heads are present for given argument 'head'; d582fb87b94c, 9b9778779fe2`

The tests are failing because there has been an additional migration added since your migration was created.

abitrolly:
@di thanks. I am going to rebase this now. Local docker build fails for me for some reason.

abitrolly:
Rebased thanks to the almighty power of StackOverflow Driven Development https://stackoverflow.com/questions/22342643/alembic-revision-multiple-heads-due-branching-error/66067051#66067051

warehouse/forklift/legacy.py
abitrolly: Well, there is https://nvd.nist.gov/vuln/detail/CVE-2019-9674 which Python core developers decided to fix with a warning. How do you suppose me to fix this in my free time?

Ark-kun:
Looks like all tests are passing now.

abitrolly:
Now there is a formal way to skip merging this PR until I can find a way to leave Belarus.

Ark-kun:
@pradyunsg What are the remaining action items here?

> Now there is a formal way to skip merging this PR until I can find a way to leave Belarus.

Sent you a referral, lol.

pradyunsg:
What's missing is one of the PyPI admins/contributors with familiarity with the codebase  (~all of whom are volunteers), finding the time and energy to reviewing this change.

That person is not going to be me. :)

abitrolly:
@Ark-kun that's very handy. Thanks a lot.

@pradyunsg I collected the monies to solve that "we are all volunteers" problem https://gitcoin.co/grants/3537/python-package-index but PSF blocks it. They also refused to openly discuss opening Open Collective https://github.com/opencollective/opencollective/issues/4790 so that it is possible to [raise specifically](https://docs.opencollective.com/help/collectives/projects) for this issue. I don't know what else can I do. Money is a tool that seemed to work well for this project.

The only technical issue that I see with this PR is testing if uploaded wheels can be exploited with .zip bomb - Python CVE was closed with [documentation comment](https://bugs.python.org/issue36260#msg341256). I filled a request to sponsor this research with https://sos.dev/ but haven't received any answer.

pradyunsg:
And in both those instances, you've been explicitly told that those are not appropriate avenues for raising funds (Cryptocurrency and an alternative platform for fundraising). You've even gone so far as to raise funds via these unauthorised mechanisms, which might have legal implications (I dunno, IANAL).

Anyway, I'm not going to engage with you anymore on the funding discussion -- you've not been receptive of the constraints that the projects and the PSF operate within and, overall, this hasn't been a particularly a productive exercise for me to engage in.

Ark-kun:
Here is how this looks like from the complete outsider perspective:

Contributor: Here is an issue affecting everyone (lack of package metadata indexing causes pip downloading tens of gigabytes of packages just to resolve dependencies)
PyPA: We do not have resources (time/money) to think about the issue.
Contributor: Here is a design proposal then.
PyPA: We do not have resources to fix the issue.
Contributor: Here is a PR then.
PyPA: We do not have resources to review and merge the PR.
Contributor: Here is the money I've collected for you.
PyPA: We do not want money.

Are there any open-source contribution avenues remaining in this case?

abitrolly:
@Ark-kun it is not PyPA, but PSF who blocks the money flow. PyPA folks never told me that they do not want support. If PyPA was independent, they could had no problems fundraising like [https://gitcoin.co/grants/65/urllib3-python-http-library](`urllib3`). But from outside it indeed looks like PyPA is guilty while PSF is officially silent.

Hopefully [Deb Nicholson](https://pyfound.blogspot.com/2022/04/deb-nicholson-new-executive-director.html) will be able to fix these non-code issues.

Wouterkoorn:
Shouldn't the implementation of this also require the increment of the minor simple API version ("1.0" -> "1.1" at this point in time). https://peps.python.org/pep-0629/

domdfcoding:
PEP 658 doesn't day anything about bumping the API version, although with hindsight it probably should have. It may be too late to change that.

uranusjr:
We can change it if needed, but I don’t think we need to? A client does not (and probably shouldn’t) need the version to detect the feature; this attribute is provided optionally at a per-link basis, so even if a Simple API server uses a version that provides this, the client still needs to do feature detection for every link to ensure the metadata is being statically served. The only scenario a version bump is useful, from what I can tell, is for a client to say “oh this server is using an old version so I’ll skip checking `data-dist-info-metadata` attributes”, which does not seem to be particularly useful in practice.

dstufft:
It’s not really worth bumping the version for this. 

The idea behind the version number was pip could bake the version number that it knew at time of release into itself, then if a repo had a newer version it could warn that the repo is using a newer api version and results might not be as expected. 

But bumping the version doesn’t do much without that already being implemented.

cosmicexplorer:
*regarding version numbering:*
I have added the capability to parse this additional metadata from both the html and json responses within pip in pypa/pip#11111, which seems close to being mergeable. In there we do precisely as @uranusjr states:
> the client still needs to do feature detection for every link to ensure the metadata is being statically served

I believe that [the parallel discussion in PEP 691](https://peps.python.org/pep-0691/#versioning) would also support the decision to avoid bumping the version number here.

*regarding this PR's progress:*
I'm not yet familiar with the warehouse codebase, so I can't push this PR forward, but I have been gradually advancing some alternatives within the pip client to accomplish the same goal (I am hoping to extend `--use-feature=fast-deps` to support sdists; see e.g. https://github.com/pypa/pip/issues/8929). If that is accepted, it will probably have to use some heuristics and therefore won't be as bulletproof as implementing PEP 658 here, but I'm hoping that will help to relieve pressure on overburdened maintainers before we get PEP 658 online.

cadojo:
Is this PR still blocked indefinitely, or does this PR fall in the "it's on the to-do list, we (PyPA) just need some time" category?

merwok:
PyPA is a loose confederation, not a single organized group.

PyPI pull requests need review from PyPI maintainers, who are very few and quite busy (and already active in comments here).

From a quick look this PR needs to merge upstream and resolve some open discussions.

abitrolly:
> PyPA is a loose confederation, not a single organized group.

@merwok PyPA are subordinate to PSF and they can not make independent decisions, like accepting crypto to compensate for time needed to review and merge this PR.

> From a quick look this PR needs to merge upstream and resolve some open discussions.

Or the people should stop praising PSF, and take the [OSS Sustainability problem](https://github.com/zloirock/core-js/issues/1179) to their corporate level. Each one of you. I know I should not communicate in passive-aggressive, but staying in Belarus when you have no money to leave is depressive, and I am losing it. I was a leper before, and the whole stuff just reinforced it. Really, I don't know why people just don't merge this stuff just to stop reading my messages that bring them down.

merwok:
abitrolly: I was replying to `we (PyPA) just need some time` to correct PyPA → PyPI maintainers.
Was not commenting on PSF-PyPA relationship.

warehouse/forklift/legacy.py
```diff
@@ -767,6 +768,21 @@ def _is_duplicate_file(db_session, filename, hashes):
...
 
 
+def extract_wheel_metadata(path):
+    """
+    Extract METADATA file and return it as a content. The name of the
+    .whl file is used to find the corresponding .dist-info dir.
+
+    See https://www.python.org/dev/peps/pep-0658/#specification
+    """
+    global _wheel_file_re
```
gst, line 778: does not need this

gst (changes requested):

warehouse/forklift/legacy.py
abitrolly: REs are precompiled on load, and therefore global.
This one is also used more than once.
And explicit is better than implicit.

frostming: > And explicit is better than implicit.

`global` isn't for such purpose, it is needed only when you need to reassign the value.

gst: effectively. 

which is not done in this function, hence my comment @abitrolly

you do not do : 

```python

import some_module

def some_function(...):
    global some_module
    some_module.foobar()
```

for instance ;)

the less globals/`global` the better.

uranusjr:
I also commented on #10504; after CPython’s response to the ZIP-bomb CVE, that issue is no longer a blocker to this, since this PR is now where that issue should be addressed.

I don’t personally have enough expertise to recommend an appropriate mitigation to the potential ZIP-bombing. But I know the CVE authors personally and can connect them to anyone that wants guidance.

warehouse/forklift/legacy.py
```diff
@@ -767,6 +768,21 @@ def _is_duplicate_file(db_session, filename, hashes):
...
+    filename = os.path.basename(path)
+    namever = _wheel_file_re.match(filename).group("namever")
```
merwok, line 780: Can the group ever be None here?

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
abitrolly: @merwok Python allows all kinds of bad programming tricks, such as monkey-patching, but that's no reason to use them. Global variables in programming [are considered to be bad practice](https://www.baeldung.com/cs/global-variables), so I am just marking their usage here.

warehouse/migrations/script.py.mako
```diff
@@ -17,8 +17,9 @@ Revises: ${down_revision}
 Create Date: ${create_date}
 """
 
-from alembic import op
 import sqlalchemy as sa
+
+from alembic import op
```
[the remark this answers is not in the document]
abitrolly: I agree that it is not related to the PR. It is probably artifact left by some VIM autoformating plugin while solving conflicts during rebase. I am gonna remove this.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
gst: I'm not sure I get, because, then, should not you mark like that *every* object/name (imported module, defined class, etc.., really anything)  that you use in any function, following that logic ? but that would obviously not be practical..

and btw: using `global` in function is effectively considered "bad" practice. I mean : if you can do without/don't need it then it's really better to do without it.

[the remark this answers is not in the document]
astrojuanlu: I agree, putting `global` there is not just a cosmetic change, it alters the usage.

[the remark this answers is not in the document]
abitrolly: > It is a global access (just like referencing a module or class!) but not a mutation, so global is useless and unwanted.

Ok. I got your point. Without the `global`, the accidental assignment / modification of regex variable only affects the function code (global becomes shadowed), and with `global` it is possible to make modification that will ruin some other part of code where the global var is used.

abitrolly:
#12702 created conflicts. Trying to resolve them now.

warehouse/migrations/script.py.mako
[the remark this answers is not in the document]
abitrolly: Removed.

abitrolly:
Rebased and squashed some CI linter commits.

ewdurbin:
superseded by #13649

abitrolly:
@ewdurbin you could at least give me a credit when copying the code into new PR, so that I could reference that trying to find a job.

Now that this code is merged, it is at least possible to test `.whl` bomb protection live #10504.

Leaving aside an acute seizure of justice, I am glad to see it finally merged. Thanks everyone for support.

astrojuanlu:
I also found it very weird to see this PR coming "out of the blue" (which of course it was not, I'm sure lots of discussion went internally about it - it was just not visible for bystanders) with parts that were almost identical to this one, without at least retaining attribution in 1 commit, or acknowledgement of some sort.

But anyway, glad to see #8254 addressed. Thanks everyone.

ewdurbin:
the long story short here is that abitrolly has consistently behaved across multiple contribution attempts in ways that were dismissive of maintainer feedback, overscoped, financially dubious, and often made me feel uncomfortable or unnecessarily pressured.

i have no interest in collaborating with this contributor.

PEP 658 is important as Metadata 2.2 (Dynamic) near implementation completion and getting back in the quagmire of interacting with abitrolly wasn't something i have capacity for.

abitrolly:
@ewdurbin given that I've never offended or held anything personal towards you, that's very unprofessional, man. Not everybody is a nice person when it comes to doing extra work in their free time, but you guys are at least being paid for maintaining things here, and "uncomfortable" and "unnecessarily pressured" is a normal job situation for many of us paid professionals. I take the blame and full responsibility for pushing these thing while you guys are finding other things to do.

With that being said, I apologize to you for my bad and uncomfortable behavior. Should I know that it is the reason to hold back my PR, I would of course take the measures to apologize earlier, just to deal with it and forget. Even if I may sound passive aggresive, I appreciate you maintainership work that resulted in these changes finally merged in.

cosmicexplorer:
Thank you so much @ewdurbin for courageously speaking directly and publicly about the reasons this work was so difficult to accept, and for putting in the effort yourself on #13649 to get this team effort across the frontend and backend over the finish line. I am replying here because a lot of what I hope are really significant perf improvements for pip and pex on very common tasks were blocking on this data being published by pypi and I'm super excited to try them out now. I'm very sorry you were subjected to that treatment and really appreciate your impactful contributions here that have enabled so much other exciting work.

abitrolly:
image

cosmicexplorer:
I'm glad that the python community can now move on from interactions like these. Sorry, everyone.

PR #9979, opened 2021-08-31
Use logo placement server to place sidebar's logos

berinhard:
This PR updates pypi to use logo placement server to place logo instead of the sponsors db.

To enable it, we had to update our CSP policies to include not only the server but also the hashes from JS and style files used. To better fit the logo in the sidebar, there's a minor CSS customization to support it. Here's a print screen of the final result:

Screenshot from 2021-10-19 10-49-58


warehouse/templates/base.html
```diff
@@ -139,6 +139,11 @@
     <script async src="https://www.googletagmanager.com/gtag/js?id={{ request.registry.settings['ga.tracking_id'] }}"></script>
     {% endif -%}
     <script defer src="https://www.fastly-insights.com/insights.js?k=6a52360a-f306-421e-8ed5-7417d0d4a4e9&dnt=true"></script>
+    <script async
+      src="https://media.ethicalads.io/media/client/ethicalads.min.js"
+      integrity="sha384-09b96ad0b934a6bfd31d07eb5dd16cc879fa40b25817b8c0bf7fc4584e4361be1f44a35fa631144d997277ad29766a56"
```
ewdurbin, line 144: Given this, we need to check if ethicalads allows some mechanism to pin to a specific version, as the `ethicalads.min.js` file changes... so too will the sha.

ewjoachim: From https://ethical-ad-client.readthedocs.io/en/latest/releasing.html?highlight=ethicalads.min.js#releasing I see:

> In the future we plan to release the client to NPM directly for users, but there is still a lot of churn and we don’t want users pinning to old versions quite yet.

So... I would guess they deliberately won't let us pin. Which means we should not put an integrity script ?

Another solution could be to vendor the script and auto-update it daily with GitHub actions. If releases aren't too often, this would make it easier, but it's a lot of moving parts.

I think at this point it's worth taking a minute to wonder what the threat model is, and which kind of attack we're trying to mitigate with the integrity hash. MITM is already handled by https. Given we don't pin, it means we don't audit the JS script itself, so a legitimate update is technically exactly the same thing as a hostile takeover. So what is this integrity hash here for?

berinhard: @ewjoachim thanks for your comments. Pinning the script versions is something we are already addressing with them. They already have pinned versions, as [v1.2.0](https://media.ethicalads.io/media/client/ethicalads-v1.2.0.min.js), but since it wasn't in their initial docs, we didn't know about that.

The integrity hash is required because, otherwise, Pyramid would apply CSP rules to prevent the script from being imported. Possibly, the same thing happens with Sentry because [its script tag](https://github.com/pypa/warehouse/blob/main/warehouse/templates/base.html#L119) also has the integrity hash.

ewdurbin (approved):

PR #9983, opened 2021-08-31
Skip invalid releases on the vulnerability report

### LINKED ISSUE #9913: Vulnerability reporting task is not resilient to deleted releases
Currently attempting to publish a vulnerability report for a notification that includes release versions that have been deleted (or never existed) causes an error and prevents the entire report from being stored.

Example: https://sentry.io/share/issue/69362937025d4524be01918a80f830d1/

If the release is not found, we should just skip that version and continue storing the vulnerability report. If *none* of the versions are found, we should error.

DiddiLeija:
Closes #9913. Store the releases found on the vulnerability report, and skip those versions that are not found. If no releases were saved, then we can raise an exception.

CC @di

### DIFF
--- a/warehouse/integrations/vulnerabilities/utils.py
+++ b/warehouse/integrations/vulnerabilities/utils.py
@@ -11,7 +11,6 @@
 # limitations under the License.
 
 from pyramid.httpexceptions import HTTPBadRequest
-
 from sqlalchemy import func, orm
 from sqlalchemy.orm.exc import NoResultFound
 
--- a/warehouse/integrations/vulnerabilities/utils.py
+++ b/warehouse/integrations/vulnerabilities/utils.py
@@ -101,18 +101,24 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
         )
         raise
 
+    report_found_releases = []  # store the found releases
+
     for version in report.versions:
         try:
             release = _get_release(request, project, version)
+            report_found_releases.append(release)
         except NoResultFound:
             metrics.increment(
                 "warehouse.vulnerabilities.error.release_not_found",
                 tags=[f"origin:{origin}"],
             )
-            raise
 
         if release not in vulnerability_record.releases:
             vulnerability_record.releases.append(release)
+
+    if len(report_found_releases) < 1:
+        # no releases found, then raise an exception
+        raise NoResultFound("None of the releases were found")
 
     # Unassociate any releases that no longer apply.
     for release in list(vulnerability_record.releases):
--- a/warehouse/integrations/vulnerabilities/utils.py
+++ b/warehouse/integrations/vulnerabilities/utils.py
@@ -146,7 +146,7 @@ def analyze_vulnerability(request, vulnerability_report, origin, metrics):
     except (
         vulnerabilities.InvalidVulnerabilityReportRequest,
         NoResultFound,
-        HTTPBadRequest
+        HTTPBadRequest,
     ):
         raise
     except Exception:
--- a/warehouse/integrations/vulnerabilities/utils.py
+++ b/warehouse/integrations/vulnerabilities/utils.py
@@ -101,12 +101,13 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
         )
         raise
 
-    report_found_releases = []  # store the found releases
+    found_releases = False  # by now, we don't have any release found
 
     for version in report.versions:
         try:
             release = _get_release(request, project, version)
             report_found_releases.append(release)
+            found_releases = True  # at least one release found
         except NoResultFound:
             metrics.increment(
                 "warehouse.vulnerabilities.error.release_not_found",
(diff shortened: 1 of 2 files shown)


warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -101,19 +101,25 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
         )
         raise
 
+    report_found_releases = []  # store the found releases
```
di, line 104: Why not a boolean? If we found at least one release, this should succeed.

DiddiLeija: Ok. I'll modify that.

DiddiLeija: Done.

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -101,19 +101,26 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
...
+
     for version in report.versions:
         try:
             release = _get_release(request, project, version)
+            found_releases = True  # at least one release found
         except NoResultFound:
             metrics.increment(
                 "warehouse.vulnerabilities.error.release_not_found",
                 tags=[f"origin:{origin}"],
             )
```
DiddiLeija, line 114: Is this part necessary? Because I think that is making the tests to fail.

di: Yes, it's still necessary. The tests are failing because the behavior has changed here but the tests haven't been updated.

DiddiLeija: Got it. As I understand, we should change `tests/unit/integration/vulnerabilities/test_utils.py`, right?

DiddiLeija:
@di I finally succeded! Can you review this pull request?

DiddiLeija:
@di I don't have access to this repository, so I need your review to apply this :)

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -101,19 +101,26 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
...
-            raise
+            continue  # skip that release
 
         if release not in vulnerability_record.releases:
             vulnerability_record.releases.append(release)
 
+    if not found_releases:
+        # no releases found, then raise an exception
+        raise NoResultFound("None of the releases were found")
```
di, line 122: This is a SQLAlchemy exception, not one we want to manually raise. We should just raise a generic exception here for the time being.

I think we should also add a test for this, the existing test doesn't assert on this behavior in any way.

DiddiLeija: > This is a SQLAlchemy exception, not one we want to manually raise. We should just raise a generic exception here for the time being.

I was thinking about `KeyError`, or `ValueError`. Is that right?

di: No, I don't think so. Since this will eventually become a HTTP response, I think `pyramid.httpexceptions. HTTPBadRequest` would be best.

DiddiLeija:
@di as I undertsand, I sorted the `pyramid.httpexceptions.HTTPBadRequest` import well. Anyway, `isort` failed:

```
+ python -m isort --check gunicorn-prod.conf.py gunicorn-uploads.conf.py warehouse/ tests/
ERROR: /home/runner/work/warehouse/warehouse/warehouse/integrations/vulnerabilities/utils.py Imports are incorrectly sorted and/or formatted.
```

Any suggestions?

di:
You can automatically reformat linting failures with `make reformat`: https://warehouse.readthedocs.io/development/submitting-patches.html#code

DiddiLeija:
Ok, I'll try to do this now. If I don't, I'll update this PR the next week.

DiddiLeija:
The tests are failing, I just have to update them.

DiddiLeija:
@di any suggestions to fix the test?

di:
The error looks pretty straightforward:

```
>       assert metrics_counter == {
            ("warehouse.vulnerabilities.received", ("origin:test_report_source",)): 1,
            ("warehouse.vulnerabilities.valid", ("origin:test_report_source",)): 1,
            (
                "warehouse.vulnerabilities.error.release_not_found",
                ("origin:test_report_source",),
            ): 2,
        }
E       AssertionError: assert Counter({('wa...ource',)): 1}) == {('warehouse....source',)): 1}
E         Omitting 3 identical items, use -vv to show
E         Left contains 1 more item:
E         {('warehouse.vulnerabilities.error.unknown', ('origin:test_report_source',)): 1}
E         Use -v to get the full diff
```

This is emitting a new metric now, you need to add it to the test.

DiddiLeija:
@di everything passed. Do I have to change something more?

DiddiLeija:
@di Any update here?

di:
I think we actually need to add `HTTPBadRequest` here instead, the "unknown" metric is just for when this raises an unknown exception: https://github.com/pypa/warehouse/blob/3802138b4eed60a1b8218168919e6de3792e9105/warehouse/integrations/vulnerabilities/utils.py#L139

DiddiLeija:
Got it. I'll work on this.

DiddiLeija:
@di I made your suggestion, and it worked fine.

di (approved):

DiddiLeija:
@di Merge this pull request whenever you want (I can't do it by myself).

DiddiLeija:
Thanks @di!

PR #5980, opened 2019-06-08
Error if no xz installed

trishankatdatadog:
Try to fix #5872

PR #9993, opened 2021-09-02
Use docker initialization scripts for DB init

abitrolly:
The behavior is explained here https://hub.docker.com/_/postgres

>  After the entrypoint calls initdb to create the default postgres user and database, it will run any *.sql files, run any executable *.sh scripts, and source any non-executable *.sh scripts found in that directory to do further initialization before starting the service.

Also fixes #5980.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,9 @@ requirements/%.txt: requirements/%.in
 	docker-compose run --rm web bin/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
 resetdb: .state/docker-build-web
-	docker-compose up --recreate db
+	docker-compose stop db
+	docker volume rm warehouse_pgdata
+	docker-compose up -d --recreate db
 
 migrate: .state/docker-build-web
 	docker-compose run --rm web python -m warehouse db upgrade head
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,10 @@
+<<<<<<< HEAD
+=======
+BINDIR = $(PWD)/.state/env/bin
+COMPOSE_PROJECT_NAME ?= $(notdir $(abspath .))
+GITHUB_ACTIONS := $(shell echo "$${GITHUB_ACTIONS:-false}")
+GITHUB_BASE_REF := $(shell echo "$${GITHUB_BASE_REF:-false}")
+>>>>>>> e1349bda (Calculate compose project name instead of hardcoding it)
 DB := example
 IPYTHON := no
 
@@ -84,7 +91,7 @@ requirements/%.txt: requirements/%.in
 
 resetdb: .state/docker-build-web
 	docker-compose stop db
-	docker volume rm warehouse_pgdata
+	docker volume rm $(COMPOSE_PROJECT_NAME)_pgdata
 	docker-compose up -d --recreate db
 
 migrate: .state/docker-build-web
--- a/Makefile
+++ b/Makefile
@@ -82,11 +82,10 @@ translations: .state/docker-build-web
 requirements/%.txt: requirements/%.in
 	docker-compose run --rm web bin/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
-initdb: .state/docker-build-web
-	docker-compose run --rm web psql -h db -d postgres -U postgres -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname ='warehouse';"
-	docker-compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS warehouse"
-	docker-compose run --rm web psql -h db -d postgres -U postgres -c "CREATE DATABASE warehouse ENCODING 'UTF8'"
-	docker-compose run --rm web bash -c "xz -d -f -k dev/$(DB).sql.xz --stdout | psql -h db -d warehouse -U postgres -v ON_ERROR_STOP=1 -1 -f -"
+resetdb: .state/docker-build-web
+	docker-compose up --recreate
+
+migrate: .state/docker-build-web
 	docker-compose run --rm web python -m warehouse db upgrade head
 	docker-compose run --rm web python -m warehouse sponsors populate-db
 	$(MAKE) reindex
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ requirements/%.txt: requirements/%.in
 	docker-compose run --rm web bin/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
 resetdb: .state/docker-build-web
-	docker-compose up --recreate
+	docker-compose up --recreate db
 
 migrate: .state/docker-build-web
 	docker-compose run --rm web python -m warehouse db upgrade head
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,4 @@
-<<<<<<< HEAD
-=======
-BINDIR = $(PWD)/.state/env/bin
 COMPOSE_PROJECT_NAME ?= $(notdir $(abspath .))
-GITHUB_ACTIONS := $(shell echo "$${GITHUB_ACTIONS:-false}")
-GITHUB_BASE_REF := $(shell echo "$${GITHUB_BASE_REF:-false}")
->>>>>>> e1349bda (Calculate compose project name instead of hardcoding it)
 DB := example
 IPYTHON := no
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,12 +25,17 @@ services:
       - ./dev/vault:/etc/vault
 
   db:
-    image: postgres:10.1
+    image: postgres:10.18
     ports:
       # 5432 may already in use by another PostgreSQL on host
       - "5433:5432"
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "postgres", "-d", "postgres"]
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -24,10 +24,15 @@ services:
...
-    image: postgres:10.1
+    image: postgres:10.18
     ports:
       # 5432 may already in use by another PostgreSQL on host
       - "5433:5432"
+    environment:
+      POSTGRES_DB: warehouse
+      POSTGRES_PASSWORD: password
+    volumes:
+      - ./dev/example.sql.xz:/docker-entrypoint-initdb.d/example.sql.xz
```
ewjoachim, line 35:
```suggestion
      - ./dev/example.sql.xz:/docker-entrypoint-initdb.d/example.sql.xz:ro
```
While we're at it's, it's free.

ewjoachim: Do we want to add a named volume with the DB data ? This should help have the DB ready more often.

docker-compose.yml
```diff
@@ -24,10 +24,15 @@ services:
       - ./dev/vault:/etc/vault
 
   db:
-    image: postgres:10.1
+    image: postgres:10.18
```
ewjoachim, line 27:
```suggestion
    image: postgres:10
```
While we're at it :)

Makefile
```diff
@@ -140,11 +140,11 @@ ifneq ($(GITHUB_BASE_REF), false)
 	git diff --name-only FETCH_HEAD | grep '^requirements/' || exit 0 && $(MAKE) deps
 endif
 
+resetdb:
+	docker-compose rm --stop --force db
+	docker-compose up -d db
```
ewjoachim, line 145: If we add a named volume, we'll need to delete it here.

docker-compose.yml
abitrolly: > Do we also want to add a named volume with the DB data ? This should help have the DB ready more often.

I am not sure I understand why DB would be not ready in development process once initialized, and how named volume helps with that. For `make resetdb` it doesn't matter if it is a named volume or not.

abitrolly: Oh, I see. The named volume that will persist already imported data.

ewjoachim: yes

abitrolly: I will address this in the thread below.

ewjoachim: To elaborate: as of today, with an unnamed volume, if you delete (`docker-compose down`) and recreate your db container, the data is gone and you have to run `make initdb` (5 to 10 min) everytime. With a named volume, `docker-compose down` will leave the volume as-is, so the next time you run the stack, the DB is ready. If you need to reinit the db, you need to either delete the whole stack (`docker-compose down --volumes`) or delete the volume directly with `docker volume`.

Makefile
abitrolly: So the named volume should hold the DB after it was imported and initialized. To me that extra redirection is not necessary, because `docker-compose down` doesn't kill the container contents. The only way DB content can be killed if somebody explicitly removes the containers explicitly with `rm` command. Why would we want to preserve the DB content in this case?

ewjoachim: > The only way DB content can be killed if somebody explicitly removes the container with rm command

Or if the user uses `docker-compose down`.

As you want, really. I offered this as a suggestion. I know I do it in my projects usually and it works for me, but it doesn't have to be useful for everyone.

ewjoachim: To me `docker compose down` means 'free all useless disk space taken by containers of this project' (and `docker compose down -v` means free all disk space taken by containers of this project'). I like being able to remove quickly everything except the thing that would be long to recreate.

abitrolly: Can you confirm that `docker-compose down` really removes the containers? Because in my setup it doesn't happen. I will check out once more right now, but I am pretty sure the behavior is that compose just issues `stop` and not `rm`.

Or do you mean that `docker-compose down` cleans up volumes, and this is what you want? But then you will need to reset the DB and run migrations anyway.

abitrolly: I was wrong. The compose setup survives reboot, but all containers are removed when `down` is invoked. So we need a named volume.

ewjoachim: yep. `docker-compose down = docker-compose stop + docker-compose rm`, roughly

docker-compose.yml
abitrolly: Added the named volume.

Makefile
abitrolly: Added named volume removal with explicit `docker` command, because `docker-compose` doesn't support this https://stackoverflow.com/questions/45511956/remove-a-named-volume-with-docker-compose

abitrolly:
Added the named volume and rebased. PTAL.

Makefile
```diff
@@ -155,6 +155,7 @@ endif
 
 resetdb:
 	docker-compose rm --stop --force db
+	docker volume rm warehouse_pgdata
```
ewjoachim, line 158: Thinking about it, I realize `warehouse_pgdata` is actually `${COMPOSE_PROJECT_NAME:-$(basename $(pwd))}_pgdata`. Chance are people didn't change COMPOSE_PROJECT_NAME and their folder is named warehouse but this is just an assumption.

(note that `${COMPOSE_PROJECT_NAME:-$(basename $(pwd))}_pgdata` is the bash syntax, I don't know enough Makefile syntaxt to say what the equivalent would be)

Alternatives include specifying a specific `name` in the volume definition to avoid the compose project name being appended.

I realize this sucks a little bit :( I hadn't foreseen that deriving the volume name would be non-trivial until I saw this diff.

abitrolly: Yea, well, need to test this with the `make` first.

ewjoachim: Here's the make syntax to do it (ok I cheated)
```
$(shell python -c 'import os;print(os.getenv("COMPOSE_PROJECT_NAME", os.getcwd().rsplit("/", 1)[-1]) + "_pgdata")')
```

abitrolly: I used standard `make` functions from http://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html

abitrolly: See the last commit.

ewjoachim (approved):

abitrolly:
Interesting. The changes are approved, but workflow still doesn't run.

ewjoachim:
Yes, I'm a pypi moderator, but I'm not an admin of the repo, I don't have the right to unblock the workflow.

FFY00:
You need commit access.

ewjoachim:
I know :) But that's not something I have as of today :)

ewdurbin (commented):
Two questions:

1) does this handle decompressing the example.sql.xz file?
2) is there a mechanism to re-intialize, or does the Postgres container require being destroyed/recreated?

abitrolly:
> does this handle decompressing the example.sql.xz file?

Yes, here.

https://github.com/docker-library/postgres/blob/740bb6d8d44a955df42d88fd5660c7043340db90/10/stretch/docker-entrypoint.sh#L168

> is there a mechanism to re-intialize, or does the Postgres container require being destroyed/recreated?

`DROP database` then load the dump with .xz as before, but that's not how containers are supposed to work.

abitrolly:
I probably need to clarify the effects of this PR.

1. The DB initialization with creating `warehouse` DB and uploading initial data dump happens automattically when the `db` container starts for the first time, not waiting for the user to issue `initdb`. So when `db` becomes online, the only thing that needs to be run are migrations.
2. Because initial dump is already imported, `initdb` now doesn't kill the database, allowing to run migrations multiple times safely. Actually, it can now be renamed to `updatedb`.
3. `resetdb` command is added to do explicit resets to the state before migrations. This allows to better control for debugging migrations. The `resetdb` reuses the DB initialization mechanism from 1. to avoid maintenance of parallel SQL code that does the same thing. The DB initialization mechanism is triggered when the container starts for the first time and there is no existing database on disk. We need to stop or kill the container to clear the volume without unexpected errors. It is possible to just stop the `db` container, but that signals that `db` container contents outside volumes is actively modified, which is not the case here.

@ewdurbin does that makes things look better? I can insert the SQL code back to `resetdb` if you need it, and then the separate PR will be needed to address https://github.com/pypa/warehouse/pull/5980

abitrolly:
So where we go from here?

di:
While this seems like a nice feature of the postgres image, I'm not sure I like that database initialization is now split.

Ideally _all_ database initialization would be done automatically here, and we wouldn't need `initdb` at all, which means that we keep the dev database up-to-date with every new migration. This would require something like https://github.com/pypa/warehouse/pull/7496 but also we'd need to make the dump much smaller, so as to not introduce a large file into the git history every time there is a migration.

abitrolly:
Database initialization is done automatically here. I could rename that to `migratedb` target as it is actually what the `initdb` target does at the moment. The only blocker is the line `docker-compose run web python -m warehouse sponsors populate-db`. I am not sure yet if it can be replaced with static import/export.

abitrolly:
Okay. Sponsors line is a technical debt from https://github.com/pypa/warehouse/pull/9512/files#diff-b912d97d6c38546cdc81c0c2cc35833412c3b8ccd46a62ac5767f030a35bef73 Because there are not pathways to Alembic migrations for Django users.

abitrolly:
Finally. 3 years later, but it is done. I would consider learning from Godot best practices for supporting the community that sends patches.

PR #9994, opened 2021-09-02
Bump google libs

ewjoachim:
This is a problem we've seen multiple times, due to https://github.com/pypa/pip/issues/9644

In order to fix it slightly more easily next time, I've added a make target with a description. Of course, there should be better fixes and all. Maybe even a word in the doc. But at least now we'll know what to do next time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ requirements/%.txt: requirements/%.in
 	$(BINDIR)/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
 fix-google-deps:
-	# Fixes https://github.com/pypa/warehouse/pull/9422 and related issues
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -130,6 +130,19 @@ deps: .state/env/pyvenv.cfg
 requirements/%.txt: requirements/%.in
 	$(BINDIR)/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
+fix-google-deps:
+	# Fixes https://github.com/pypa/warehouse/pull/9422 and related issues
```
di, line 134:
```suggestion
	# Fixes issues caused by https://github.com/pypa/pip/issues/9644
```

PR #10002, opened 2021-09-03
Translations are now trimmed by default.

### LINKED ISSUE #9420: All trans tags should be trimmed by default
Having translation strings not be trimmed always results in ugly things for translators.

We can approach this in 2 ways:
- The simple but inelegant way: add `trimmed` to all existing trans tags and a CI check to ensure it's added to future tags too
- The elegant way: tell jinja2/gettext to trim by default and subsequently remove all `trimmed` parameters.

The problem for solution number 2 is that it requires a way to access the [policies on the Jinja2 environment](https://jinja.palletsprojects.com/en/2.11.x/api/#ext-i18n-trimmed), and I haven't been able to find how to do so using `pyramid_jinja2`.

If anyone finds how to do so...

ewjoachim:
Fixes #9420

This will make translators lives easier in the long run, though it does create a few strings to retranslate. Normally, Weblate will provide the previous translation so it's just 1 click times the number of trimmed translations.

Now we COULD try to make it better. It would involve fixing the existing files and reuploading them in Weblate (ideally in a short amount of time as to not loose translations made on Weblate in the meantime). The problem is, while it's "easy" using standard `msg` tools to apply `strip()` to every existing translation `msgstr`:
```
msgfilter -i warehouse/locale/fr/LC_MESSAGES/messages.po python -c 'import sys; print(sys.stdin.read().strip(), end="")'
```
... I haven't found a way to do the same for the `msgid`s, which makes the whole thing much harder.

So I think it's much easier as-is. And I'm positive this will improve the quality of life for translators.

### How to review

The first commits are pretty short. The last 2 are automated, so I really suggest reading commit-by-commit. The PR is quite short without the last 2 commits.


ewjoachim:
(damn, my local rebase alias command was set incorectly, I rebased on an old version. Let's do it again)

ewjoachim:
Here you are :)

ewdurbin (approved):

ewjoachim:
🎉 ❤️

PR #10003, opened 2021-09-04
Improve notation readability for aria parameters in jinja2 templates

ewjoachim:
Replace `**{"aria-describedby": "foo"}` with `aria_described="foo"`. Also added much needed newlines.


warehouse/templates/manage/account.html
```diff
@@ -312,7 +319,10 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
           {% endif %}
         </label>
         <p class="form-group__text">
-          {{ save_account_form.public_email(class_="form-group__field", **{"aria-secribedby":"public-email-errors"}) }}
+          {{ save_account_form.public_email(
+              class_="form-group__field",
+              aria_describedby="public-email-errors",
+          ) }}
```
ewjoachim, line 325: Note: this one was broken, so I fixed it too.

miketheman:
#Triage 
This looks like a pretty good idea from a while back, but has fallen into conflicts.
@ewjoachim would you want to try this again? It's totally fine if not and you want to close this PR.

ewjoachim:
@miketheman done

Hm, I would imagine I'd need to rerun the translations as the line numbers changed, but it seems they actually crash. What is happening...?

Issue #9988, opened 2021-09-02
Upload dev containers


abitrolly:
**What's the problem this feature will solve?**

Setting up development environment takes a lot of time, with 10+ minutes going into `make build`.

**Describe the solution you'd like**

Make GitHub Actions to upload latest built images to https://hub.docker.com/r/pypa/warehouse and make sure `docker-compose` can find them.

ewjoachim:
If we push cache, it probably is useful to consider inlining build cache too !
https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources

Also, if we have images, this would be the first step toward having a complete dev env on codespaces ([github.dev](https://github.com/features/codespaces))

abitrolly:
I didn't know Docker had the ability to distributed inline caches, and I like it.

For github.dev I've heard VS uses `.devcontainer` - not sure if it is a standardized approach, but it looks like either that or Kubernetes things like Google's [`skaffold`](https://skaffold.dev/) provide modern approach to multi-container development.

PR #10004, opened 2021-09-04
Refactoring multiple parts of the dockerfile

ewjoachim:
- Switched to a 5-staged build:
  - Static, like before (but we now use it instead of Dockerfile.static)
  - Base which has apt deps and main env vars
  - Build with the build-time apt deps, which does the main pip install
  - Dev, which is the developper image (there's no DEVEL arg anymore)
  - App, which is to be the production image
- Remove a lot of heavy & unused things from the docker context through .dockerignore
- Add apt & pip cache, to speed up rebuilds and limit bandwith requirements (with no impact on layer size)
- use `set -eux; \` style instead of `set -u \ &&` style
- Address the todo that was in Dockerfile: add a lockfile for pip, setuptools & wheel
- docker-compose uses python image directly for `notdatadog` & `files`, should be smaller on disk/memory

This is a good first step toward #9988 

@ewdurbin I believe there might be implications on the deployment pipeline, though it might be nothing more than adding `--target=app` to the docker build. And even without that, I believe it would be the default.


abitrolly:
Should it be added into GitHub Actions pipeline? I am interested to see how the changes improve the build time.

ewjoachim:
> I am interested to see how the changes improve the build time.

I'm not sure this will speed up CI a lot, because I doubt GitHub CI keeps `--mount=type=cache` between calls, this would be too dangerous because if users polluted the cache, it would be hard to recover.

For the rest, the layers here should roughly take as long as before, it's refactored for readability and ease of use.

In the end the operations we have are:

- apt install run time deps
- pip install run time deps
- apt install build time deps
- pip install dev
- build statics
and that's it. We had that before, we have this now, and it takes as long to run.

And even with a layer cache, it reduces bandwidth usage but from an human point of view, downloading 100ko with fiber or reading it from disk takes as long (0s).

ewdurbin (approved):
This is really incredible work. Thank you @ewjoachim!

Our deployment pipeline doesn't currently have support for setting a target when building, but since the app ends up being default, that's not a problem!

Dockerfile
```diff
@@ -39,109 +49,124 @@ COPY Gulpfile.babel.js /opt/warehouse/src/
...
-ENV PATH="/opt/warehouse/bin:${PATH}"
-
-# Next, we want to update pip, setuptools, and wheel inside of this virtual
-# environment to ensure that we have the latest versions of them.
-# TODO: We use --require-hashes in our requirements files, but not here, making
-#       the ones in the requirements files kind of a moot point. We should
-#       probably pin these too, and update them as we do anything else.
-RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip setuptools wheel
+# Pip configuration (https://github.com/pypa/warehouse/pull/4584)
+ENV PIP_NO_BINARY=hiredis PIP_DISABLE_PIP_VERSION_CHECK=1
```
ewdurbin, line 109: It completely escapes me why we do not use the binary releases for hiredis... but based on the date it seems to be related to https://github.com/redis/hiredis-py/issues/69 (nice) which was seemingly only related to that 0.2.0 release.

Perhaps we should go back to using the wheels provided?

ewdurbin: Will address in follow up PR

ewjoachim: :+1:

PR #10021, opened 2021-09-08
Restore "Refactoring multiple parts of the dockerfile (#10004)"

ewdurbin:
Reverts pypa/warehouse#10020


ewdurbin:
HOLD FOR: upgrade to BuildKit compatible builders in deployment pipeline.

miketheman:
The underlying code has changed a fair amount since this was introduced and no longer merged cleanly.

I think there's still good ideas on leveraging cache mounts, but they rely on BuildKit being available during deployment. @ewdurbin is this something we still want to entertain?

ewdurbin:
absolutely want to entertain this. getting us up to a more recent container runtime that supports buildkit is a priority in the coming months.

ewdurbin:
I thought it was just the docker service that needed BuildKit, but turns out its the client too.

Cabotage interacts with the docker build hosts via docker-py which notably does not support BuildKit. ref https://github.com/docker/docker-py/issues/2230

So this remains blocked even after getting our builder upgraded 😅

ewdurbin:
We have a buildkit enabled deployment pipeline! I am closing this in favor of a new implementation to avoid all the deltas that have been introduced.

PR #10037, opened 2021-09-13
Add make test-shell

ewjoachim:
Maybe the makefile could be improved and simplified a bit (after #10021 is merged) but meanwhile, this might help.

This PR adds a makefile target that lets you dramatically speed up the feedback loop of tests when you write some code and run tests and iterate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,8 +84,16 @@ debug: .state/docker-build
 	docker-compose run --rm --service-ports web
 
 tests: .state/docker-build
+	docker-compose up -d db
 	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
 								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
+								  bin/tests --postgresql-host db $(T) $(TESTARGS)
+
+test-shell: .state/docker-build
+	docker-compose up -d db
+	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
+								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
+								  OPEN_SHELL=1 \
 								  bin/tests --postgresql-host db $(T) $(TESTARGS)
 
 static_tests: .state/docker-build
--- a/bin/tests
+++ b/bin/tests
@@ -29,6 +29,12 @@ if [ $ATTEMPTS -eq 5 ]; then
   exit 1
 fi
 
+if [ -n "${OPEN_SHELL:-}" ]; then
+    export PYTHONPATH=/opt/warehouse/src/
+    echo "Welcome to the test shell. Use 'pytest' to run the tests."
+    exec bash
+fi
+
 # Print all the followng commands
 set -x
 
--- a/docs/development/getting-started.rst
+++ b/docs/development/getting-started.rst
@@ -552,6 +552,21 @@ If you want to run a specific test, you can use the ``T`` variable:
 
     T=tests/unit/i18n/test_filters.py make tests
 
+If you find yourself launching some tests multiple times in a row, you may
+find out that the startup time of docker can get in the way of a quick
+feedback loop. In this case, you may want to run:
+
+.. code-block:: console
+
+    make test-shell
+
+This will open a bash console in a configured docker-container, where you can
+run ``pytest`` directly. `Some pytest options`__ such as ``--last-failed`` or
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -84,10 +84,18 @@ debug: .state/docker-build
 	docker-compose run --rm --service-ports web
 
 tests: .state/docker-build
+	docker-compose up -d db
```
ewjoachim, line 87: This one is a freebie. I can make it in its own PR but I think it's ok to have it here.

bin/tests
```diff
@@ -29,6 +29,12 @@ if [ $ATTEMPTS -eq 5 ]; then
   exit 1
 fi
 
+if [ -n "${OPEN_SHELL:-}" ]; then
+    export PYTHONPATH=/opt/warehouse/src/
+    echo "Welcome to the test shell. Use 'pytest' to run the tests."
```
ewjoachim, line 34: Note: spaces are important. `"Welcome to the test shell" != "Welcome to the tests hell"`.

❤️

abitrolly:
```
$ docker-compose run web bash
```
```
# pytest
```
Works for me.

---

For #9972 I've created a helper `testlift.sh` like this.
```
#!/bin/bash

export T=tests/unit/forklift/test_legacy.py 

podman-compose run --rm web bin/tests --postgresql-host db "$T""$*"
```
Because editing long params while working on specific tests is not convenient.

And if I really need a fast feedback loop then I launch it with [`watchexec`](https://github.com/watchexec/watchexec).

ewjoachim:
>     $ docker-compose run web bash
>     # pytest
> Works for me.

It works in the sense that pytest starts or is pytest actually able to to run all of the tests correctly on main ? Because for me it definitely isn't.

But I'd be fine working the other way around, configuring pytest so that it works by default without the need for bin/test etc.

Anyone can have their own solution like your testlift, I find it easier if there's an official one.

abitrolly:
```
FAILED tests/unit/test_db.py::test_configure_alembic - TypeError: config_cls() takes 0 positional arguments but 1 was given
======================================================= 3 failed, 2440 passed, 99 warnings in 122.31s (0:02:02) ========================================================
root@644d3790e8c9:/opt/warehouse/src# 
```

3 tests failed, because I am testing Alembic. What failures do you get?

abitrolly:
> it works by default without the need for bin/test etc.

Actually I am getting error about Python not found every time.
```
/bin/sh: 1: .state/env/bin/python: not found
```
https://github.com/abitrolly/warehouse/runs/3640554285#step:8:4

I suspect that you have this Python, and it is mounted into the container.

docs/development/getting-started.rst
```diff
@@ -552,6 +552,21 @@ If you want to run a specific test, you can use the ``T`` variable:
...
+If you find yourself launching some tests multiple times in a row, you may
+find out that the startup time of docker can get in the way of a quick
+feedback loop. In this case, you may want to run:
+
+.. code-block:: console
+
+    make test-shell
+
+This will open a bash console in a configured docker-container, where you can
+run ``pytest`` directly. `Some pytest options`__ such as ``--last-failed`` or
```
miketheman, line 564: Suggestion: use a named link target instead of a double underscore, consistent with other links in the same file, or inline the URL.

miketheman (commented):
Not a committee, but saw this older PR and thought I’d share some thoughts.

Makefile
```diff
@@ -84,10 +84,18 @@ debug: .state/docker-build
...
 	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
 								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
 								  bin/tests --postgresql-host db $(T) $(TESTARGS)
 
+test-shell: .state/docker-build
+	docker-compose up -d db
+	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
+								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
+								  OPEN_SHELL=1 \
+								  bin/tests --postgresql-host db $(T) $(TESTARGS)
```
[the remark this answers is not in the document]
ewjoachim: That could have been a possibility, but I believe the need for this has been kinda removed by the adoption of [`96ef622` (#10803)](https://github.com/pypa/warehouse/pull/10803/commits/96ef62291f532cff07e32ffb70bcf5e452997429)

I believe this whole PR might be moot now. I'll close it and if it looks like something needs to be done still, I'll reopen it, or a new one, and ping you :)
(thanks for the review still)

PR #10061, opened 2021-09-20
De-vendor & update normalize.css

### LINKED ISSUE #10056: Display details tag in document more clearly

When viewing the markdown containing the details tag as a document on the PyPI project page, it is difficult to see that the details tag can be clicked and expanded because it looks like any other text.

Before expanding:
image

After expanding:
image

For example, I think it is better to display a right-facing triangle next to the summary text in the details tag like GitHub:

image

**Additional context**

Screenshots was taken at: <https://pypi.org/project/simpleverse>

ewjoachim:
Fixes #10056.

See the ticket in detail to understand why this PR fixes the issue, because the link is subtle :D
Tested, with success.
Here's a screenshot (taken from the Black project, which is present in the local db and has details/summary tags)

Capture d’écran 2021-09-22 à 22 20 25
Capture d’écran 2021-09-22 à 22 20 29

Also, a visit on different pages of the website didn't immediately show anything fishy


warehouse/static/sass/warehouse.scss
```diff
@@ -43,7 +43,7 @@
 
 // RESETS LAYER: applies CSS resets and sensible defaults
 @import "resets/boxsizing";
-@import "resets/normalize";
+@import "~normalize.css/normalize";
```
abitrolly, line 46: What does ~ prefix mean?

ewjoachim: Something I saw meaning `node_modules/normalize.css` but I have absolutely no idea whether it will work. I've been struggling with docker issues recently, so I have yet to test this.

abitrolly: Have you seen https://github.com/pypa/warehouse/pull/10052? I may need to squash it, but it ensures that `docker-compose` setup works okay on a clean system.

ewjoachim: Yeah, my problem was not with the image but with docker itself. See https://github.com/docker/compose/issues?q=is%3Aissue+sort%3Aupdated-desc+author%3Aewjoachim+ for details :)

ewjoachim: Aaaaaand the tilde is gone.

Gulpfile.babel.js
```diff
@@ -134,11 +134,12 @@ gulp.task("dist:js", () => {
 
 gulp.task("dist:noscript", () => {
   let sassPath = path.join(staticPrefix, "sass");
+  let nodeModules = "node_modules";
 
   return gulp.src(path.join(sassPath, "noscript.scss"))
     .pipe(sourcemaps.init())
     .pipe(
-      sass({ includePaths: [sassPath] })
+      sass({ includePaths: [sassPath, nodeModules] })
```
di, line 142: Does this have the effect of including all node modules?

ewjoachim: Afaict, it has the effect of making all node modules searchable by an `@import` clause. if I `@import foo`, and there's no foo in `warehouse/static/sass/`, it will look for a foo in `node_modules/`.

If you'd rather we add specifically the right folder, we could do it. I believe we could also `@import ../../../node_module/normalize.css` but I don't like it at all.

di: I think that's fine if it's only including explicitly imported modules in the bundle.

di (approved):

warehouse/static/sass/warehouse.scss
abitrolly: https://github.com/pypa/warehouse/pull/9993 fixes your problem with running this with `docker-compose` by using standard Postrgres import mechanism that supports .xz files without `xz` dependency and pipes.
```
xz -d -f -k dev/example.sql.xz --stdout | docker-compose run --rm web psql -h db -d warehouse -U postgres -v ON_ERROR_STOP=1 -1 -f -
```
If you can review this, that would help with my `podman` workflow too.

ewjoachim: I solved it by `pipx install docker-compose`, making sure I have the (working) docker-compose v1 on my path.

PR #10197, opened 2021-10-15
Collect additional vulnerability data, surface in JSON API

di:
This PR updates the vulnerability endpoint to collect additional fields from the vulnerability report, and also surfaces the vulnerabilities for a given release in the JSON API.

### DIFF
--- a/warehouse/integrations/vulnerabilities/__init__.py
+++ b/warehouse/integrations/vulnerabilities/__init__.py
@@ -11,7 +11,7 @@
 # limitations under the License.
 
 
-from typing import List
+from typing import Dict, List
 
 from warehouse import integrations
 
@@ -30,12 +30,16 @@ class VulnerabilityReportRequest:
         vulnerability_id: str,
         advisory_link: str,
         aliases: List[str],
+        details: str,
+        events: List[Dict[str, str]],
     ):
         self.project = project
         self.versions = versions
         self.vulnerability_id = vulnerability_id
         self.advisory_link = advisory_link
         self.aliases = aliases
+        self.details = details
+        self.events = events
 
     @classmethod
     def from_api_request(cls, request):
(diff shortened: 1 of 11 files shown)


tetsuo-cpp (approved):
LGTM!

warehouse/integrations/vulnerabilities/__init__.py
```diff
@@ -30,12 +30,16 @@ def __init__(
         vulnerability_id: str,
         advisory_link: str,
         aliases: List[str],
+        details: str,
+        events: List[Dict[str, str]],
```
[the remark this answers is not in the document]
di: This is a fair point: this is probably too OSV specific to belong in this generic class. What we really care about here si the version(s) that this vunlnerability was fixed in (if any), so we should extract that from the OSV-specific integration into a more generic way here.

warehouse/integrations/vulnerabilities/utils.py
```diff
@@ -89,6 +89,8 @@ def _analyze_vulnerability(request, vulnerability_report, origin, metrics):
             source=origin,
             link=report.advisory_link,
             aliases=report.aliases,
+            details=report.details,
```
ewdurbin, line 92: will these be backfilled by some mechanism or are they new from the integrator?

ewdurbin (approved):
Adding some information to the help page on PyPI.org for folks to understand where this information comes from is probably a good idea.

Some users may understand this as coming from maintainers?

Some maintainers may be surprised/confused to see this published via our API without their involvement?

warehouse/integrations/vulnerabilities/utils.py
di: Since these aren't currently being used for anything, we will 'backfill' by wiping the existing rows and re-ingesting all the reports.

Issue #10199, opened 2021-10-17
Sticky notification bar overlaps and hides search bar on website


avats-dev:
**Describe the bug**

The new alert bar on PYPI overlaps and hides search bar on [pypi](https://pypi.org/) when on a particular module page.

**Expected behavior**

Obviously one should have access to the search bar and I think there should be an option (like a ❌ to remove the bar if one need so). It should be placed in a way that pushes the whole screen below by its height and is above to avoid any usage issues.
> Also, if a user opts to visit and provide feedback from there, the bar should not be there then but currently this is not the case. 

**To Reproduce**

Visit [here](https://pypi.org/project/windows-curses/) or any other similar page.

**My Platform**

OS : Windows 10/11
Chrome : Version 94.0.4606.71 (Official Build) (64-bit)

**Additional context**

ewjoachim:
Hm, I could be wrong but it might be a race condition:

This line:
https://github.com/pypa/warehouse/blob/c6392cb8035d8f3f570dee4f4bfe664d1d2e17ef/warehouse/static/js/warehouse/index.js#L109

Says that within 200ms of the document being ready, the height of the HTML element is computed and added to an inline padding on body.

The flash messages are computed by loading a different page and embedding its partial-html content into our page. So depending on the total timing, the page already has the content or not, which results on the bar being sometimes correct, sometimes incorrect. Not 100% sure it's that, but this could be the reason.

ewjoachim:
Maybe it's linked to the new banner feature, and maybe the fastly caching parameters for the banner are slight different, or something ?

pradyunsg:
IIUC, this is the first time we're using https://github.com/pypa/warehouse/pull/9656.

miketheman:
I've confirmed today that the notification banners no longer overlap with the search bar on the index and any package page, in desktop and mobile modes.

We've also since introduced the ability to dismiss certain banners (if the admin sets it so).

Closing since it doesn't appear any longer, but if any reporter still has this problem, please reopen with details like browser version, pages, and a screenshot with the problem.

PR #10205, opened 2021-10-18
bind "PositionWarning" function to CSILoaded event

ewdurbin:
proposed fix for #10199


pradyunsg:
Would it make sense to remove the `setTimeout(PositionWarning)` earlier in the file or, perhaps, to replace that with this instead?

PR #10278, opened 2021-10-28
Requirements makefile target should include venv

di:


miketheman:
I believe this change has introduced a circular dependency, as perceived by these messages:

```shellsession
$ make -v | head -1
GNU Make 3.81
$ make build
make[1]: Circular requirements/docs.txt <- .state/env/pyvenv.cfg dependency dropped.
...
```

For funnies, I tried the same with [remake](http://bashdb.sourceforge.net/remake/), which uses a newer version of `make`:

```shellsession
$ remake --version | head -1
GNU Make 4.3+dbg-1.5
$ remake build
remake[1]: Entering directory '/Users/miketheman/workspace/miketheman/warehouse'
remake[1]: Circular requirements/docs.txt <- .state/env/pyvenv.cfg dependency dropped.
remake[1]: Circular requirements/lint.txt <- .state/env/pyvenv.cfg dependency dropped.
remake[1]: '.state/docker-build' is up to date.
...
```

I don't know if this is harmful yet, but I'm having other issues getting started with development, and wondered if this is somehow related or not.

Happy to move this to a new Issue, if needed!

di:
@miketheman Indeed it does. I'm not sure there's a way to resolve it other than moving towards #4948.

ewjoachim:
I think we could mostly solve the problem by making the targets more granular, but at core we need to solve something: in order to install pip-compile, we need to know which version to install it too, which means we need to run pip-compile and so on.
If we install pip-compile without the hashes, it kinda defeats the purpose of using hashes in the first place.

So one solution could be to not declare to make that the pip-compile target needs the pip-compile requirement file, considering that it doesn't often change, and the venv gets recreated once in a while anyway.

So we would have:
- A target for venv creation but not installation of dependencies
- a target for pip-tools that would depend on that one, which would actually use the requirements/pip-tools.txt file but wouldn't say it
- requirements/*.txt would depend on that one
- the full environement target, the one everyone uses, would use the relevant .txt files and declare it.

I believe it would work.

PR #10307, opened 2021-11-04
upgrade elasticsearch to 7.x

ewdurbin:


docker-compose.yml
```diff
@@ -43,9 +43,11 @@ services:
       - "4576:4576"
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:6.8.7
+    image: elasticsearch:7.10.1
```
ewdurbin, line 46: note: not going to greater version since this is max version of upstream ES supported by our hosting (Amazon OpenSearch Service)

PR #10319, opened 2021-11-05
support Apple Silicon/M1 development environments

ewdurbin:
Major blocker was incompatibility of Elasticsearch 6.x containers, but that is resolved in #10307.

This handles a few loose ends:

  - set appropriate flags to get static/node container build working on arm
  - poor performance of older version of localstack under emulation, newer release is acceptable! (no arm containers available [yet](https://github.com/localstack/localstack/pull/4754))
  - explicitly pull in sqlalchemy greenlet dependency to keep pip-compile results from churning between envs


ewjoachim:
Do you need me to update the notgithub image ?

ewdurbin:
@ewjoachim if you're able to produce an arm build for it, that'd be great! but not required!

PR #10338, opened 2021-11-10
add win_arm64 to allowed platforms

nsait-linaro:
This patch enables uploading win_arm64 binary wheels to the PyPI repository.

CPython and key packages have been already ported to win/arm64 platforms. Python upstream CI is running for win/arm64 and Python official support is expected from Python 3.11. There are many package maintainers keen to start pushing win_arm64 binary wheels and would be helpful if win_arm64 is added to the permitted platform list.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -107,6 +107,7 @@ STDLIB_PROHIBITED = {
 _allowed_platforms = {
     "any",
     "win32",
+    "win_arm64",
     "win_amd64",
     "win_ia64",
     "manylinux1_x86_64",


di (approved):
Looks like this is indeed the platform tag returned by CPython: https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/distutils/util.py#L44-L45

nsait-linaro:
Thanks, @di, and @ewdurbin for the quick review.

Do you know when the changes would be deployed on the production and test PyPI server?

di:
@nsait-linaro It was live as of ~2 minutes ago!

nsait-linaro:
> @nsait-linaro It was live as of ~2 minutes ago!

Thanks, it works. I can upload wheels now 🎉

PR #10368, opened 2021-11-16
dependabot: check monthly

hugovk:
Checking monthly rather than daily will reduce the number of PRs opened (and also reduce the number of newly outdated PRs closed, because a new update for a given package closes an older PR for that package).

Options are daily, weekly, monthly. It looks like monthly is fine.
 
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval

_Security updates_ can still be made outside the interval:

> Note: `schedule` defines when Dependabot attempts a new update. However, it's not the only time you may receive pull requests. Updates can be triggered based on changes to your `dependabot.yml` file, changes to your manifest file(s) after a failed update, or Dependabot security updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
 - package-ecosystem: pip
   directory: "/"
   schedule:
-    interval: daily
+    interval: monthly
   open-pull-requests-limit: 99
   allow:
   - dependency-type: direct


di:
Thanks for taking the time to do this! I'm not sure I'm in favor, though: we have dependencies like https://pypi.org/project/trove-classifiers/ for which we generally want to land the updates as quickly as possible, and don't want to wait a month for updates.

I'm holding out for https://github.com/dependabot/dependabot-core/issues/1190 to cut down on the # of notifications here, but not sure when we'll see that land. Switching to something like Renovate is another option, but I do like the fact that Dependabot is more integrated into GitHub.

hugovk:
Ah, good point about trove-classifiers.

I think it's possible to have two sections, one daily for trove-classifiers, and one monthly for everything else:

```yml
- package-ecosystem: pip
  schedule:
    interval: daily
  allow:
    - dependency-name: "trove-classifiers"
  ...

- package-ecosystem: pip
  schedule:
    interval: monthly
  ignore:
    dependency-name: "trove-classifiers"
  ...
```

Does that sound useful?

If so, I can test it out in a fork first and update this or open a new PR later.

alex:
I don't believe that works, you can only have one choice per package-ecosystem.

hugovk:
Indeed it's not! I just tested with:

```yml
version: 2

updates:
- package-ecosystem: pip
  directory: "/"
  schedule:
    interval: daily
  open-pull-requests-limit: 99
  allow:
  - dependency-name: "trove-classifiers"
  - dependency-type: direct
  - dependency-type: indirect
  rebase-strategy: "disabled"

- package-ecosystem: pip
  directory: "/"
  schedule:
    interval: weekly
  open-pull-requests-limit: 99
  allow:
  - dependency-type: direct
  - dependency-type: indirect
  ignore:
  - dependency-name: "trove-classifiers"
  rebase-strategy: "disabled"
```

And it failed with:

> Your .github/dependabot.yml contained invalid details
> Dependabot encountered the following error when parsing your .github/dependabot.yml:
> 
> The property '#/updates/1' is a duplicate. Update configs must have a unique combination of > 'package-ecosystem', 'directory', and 'target-branch'
> Please update the config file to conform with Dependabot's specification.

https://github.com/hugovk/warehouse/runs/4416292209

Here's the dependabot feature request: https://github.com/dependabot/dependabot-core/issues/2390

Closing until such a time this might be possible!

---

As an aside, I had earlier deleted `.github/dependabot.yml` from my fork to avoid spam. After restoring it for this test, here's what I got:

image

I'll delete `.github/dependabot.yml` again from my fork.

This is a known problem with dependabot: https://github.com/dependabot/dependabot-core/issues/2804

PR #10393, opened 2021-11-21
docs: prepare the user that they need python 3.8

miketheman:
Until all `make` targets are built in containers, a new contributor will
stumble on this error and there's nothing in the docs that tell you what
you need.

Refs: https://github.com/pypa/warehouse/issues/4948


di:
Thanks!

PR #10394, opened 2021-11-21
Clarify browser support

miketheman:
Edit: Scope changed.

Simplify the static table and inform user on how to see which browsers are supported.

---

As noted in the frontend development docs, the development target
supports current version and one back, plus some other versions.

As frontend tooling will generate code based on capabilities, create a
configuration item that confirms to the documentation.
This key will inform packages that use `browserslist`, which in turn use
`caniuse-lite` to generate output code, so having the setting in place
helps confirm to the documentation.

Refs: https://github.com/browserslist/browserslist#packagejson
Refs: https://warehouse.pypa.io/development/frontend.html#browser-support

Note: The defaults that are currently in place without this
configuration are:

    '> 0.5%',
    'last 2 versions',
    'Firefox ESR',
    'not dead'

The difference from defaults and this one is:

    $ diff defaults.txt warehouse.txt
    9,11d8
    < chrome 94
    < chrome 93
    < chrome 92
    16,18d12
    < firefox 92
    < firefox 91
    < firefox 78
    22,23d15
    < ios_saf 14.0-14.4
    < ios_saf 12.2-12.5
    29d20
    < opera 79
    32a24
    > safari 14
    33a26,34
    > safari 13
    > safari 12.1
    > safari 12
    > safari 11.1
    > safari 11
    > safari 10.1
    > safari 10
    > safari 9.1
    > safari 9

Once merged, this should be followed with a change based on running:

    npx browserslist@latest --update-db

which will update `package-lock.json` with the updated browser versions
based on the configuration - something that should be run periodically.

### DIFF
--- a/docs/development/frontend.rst
+++ b/docs/development/frontend.rst
@@ -89,6 +89,10 @@ Browser Support
  IE       11+
 ========= ====================
 
+See the ``browserslist`` key in ``package.json`` for the configuration of
+supported browsers.
+You can see the full list of supported browsers by running ``npx browerslist``
+in the root of the project.
 
 HTML Code Style
 ---------------
--- a/docs/development/frontend.rst
+++ b/docs/development/frontend.rst
@@ -78,20 +78,10 @@ of some pages, but the "new" version of others.
 Browser Support
 ---------------
 
-========= ====================
- Browser  Supported Versions
-========= ====================
- Chrome   Current, Current - 1
- Firefox  Current, Current - 1
- Edge     Current, Current - 1
- Opera    Current, Current - 1
- Safari   9.0+
- IE       11+
-========= ====================
+We aim to support all major browsers, including IE11. We also support one-back,
+and follow the ``defaults`` recommendation from ``browserslist``.
 
-See the ``browserslist`` key in ``package.json`` for the configuration of
-supported browsers.
-You can see the full list of supported browsers by running ``npx browerslist``
+You can see the full list of supported browsers by running ``npx browserslist``
 in the root of the project.
 
 HTML Code Style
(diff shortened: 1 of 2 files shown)


miketheman:
I'm of mixed mind here - part of me thinks the defaults from `browserslist` are acceptable, unless there's meaningful semantics as to why Safari 9+ is still there - so maybe the better change is to keep the docs update on how to show the list, and remove the static table?

miketheman:
Original commit for the frontend browser support table, hasn't changed since. https://github.com/pypa/warehouse/pull/1053

docs/development/frontend.rst
```diff
@@ -89,6 +89,10 @@ Browser Support
  IE       11+
 ========= ====================
 
+See the ``browserslist`` key in ``package.json`` for the configuration of
+supported browsers.
+You can see the full list of supported browsers by running ``npx browerslist``
```
di, line 94:
```suggestion
You can see the full list of supported browsers by running ``npx browserslist``
```

di (commented):
We're definitely due to revisit this list. I think I'd be OK with moving towards the browserslist's defaults as our supported browsers list.

The table is in our docs as a policy so we can point people there who have issues with our browser support, but we don't do anything too exotic and I haven't seen any complaints about our compatibility, so it's probably fine for this to just point to `npx browserslist`.

miketheman:
> ... so it's probably fine for this to just point to `npx browserslist`.

That sounds like a good path forward. Should I update the config / table as part of this PR, or should that be a different piece of work?

di:
Let's stick to the defaults (remove the config) and remove the table (as it will quickly get out of date) in this PR.

di (approved):

di:
Thanks!

PR #10395, opened 2021-11-22
Add controller for search focus for slash keypress

### LINKED ISSUE #6099: Create keyboard shortcut to focus search bar

miketheman:
While having the search key autofocus is forbidden, having a keypress
focus the search field is a desirable feature.

Add a new Stimulus Controller that listens to the `/` keypress and
focuses the input on the search field.

Note: Currently applies only to the home page.

Fixes #6099


di:
Thanks! I think if we add a check that the cursor is not currently in an input field, we could add this to the base template, so this is usable everywhere the search box is currently present.

miketheman:
Happy to look at that change as well - but then I'm not sure how the base template property would be inherited to the homepage's search box?

di:
It wouldn't, we'd need to add it to both the form on the homepage (which you've already done here) and to this form: https://github.com/pypa/warehouse/blob/597938cff04261fd55f59dd538982b4bfcad4621/warehouse/templates/base.html#L225-L233

miketheman:
@di I think I'm ready for another review! I'm happier with this implementation - it's more tightly scoped, albeit increasing the html line length.

di (approved):

di:
Thanks, this is great!

PR #9712, opened 2021-06-23
Replace factory.fuzzy by factory.Faker

berinhard:
Fixes #9707 

This PR replaces deprecated `factory.fuzzy` calls by `factory.Faker` ones. A few comments about this PR:

-  At several places I used `Faker("pystr")` instead of more meaningful providers such as `user_name`. This is because a few tests broke due to unique constraints violations;
- For some reason a few legacy tests are failing with no apparent reason to me. I spent some time investigating them, but couldn't figure out why. I decided to open this PR to get some help with them.

PR #10410, opened 2021-11-26
Handle some Deprecations

miketheman:
When running the test suite, I saw a bunch of deprecation warnings that looked to be relatively minimal to handle, so took a shot and cleaning up some of the output.

Each commit contains some text as to the change, and any links I could dig up with references to the source.

Reduced 175 warnings emitted in CI to 156.

### DIFF
--- a/bin/tests
+++ b/bin/tests
@@ -33,6 +33,6 @@ fi
 set -x
 
 # Actually run our tests.
-python -m coverage run -m pytest --strict $COMMAND_ARGS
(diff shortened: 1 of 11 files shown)


ewjoachim (approved):

di (approved):

di:
Thanks @miketheman! If you wanted to continue farther down this path, https://github.com/pypa/warehouse/issues/9707 might be a good next issue (there is a start at #9712 but it seems to have stalled).

PR #10431, opened 2021-12-01
api-reference/json: document `vulnerabilities` in responses

### LINKED ISSUE #10218: Documentation for the JSON API does not show the new "vulnerabilities" key.
#10197 added a new "vulnerabilities" key to the JSON API, but the [documentation](https://warehouse.pypa.io/api-reference/json.html) doesn't show it in its example response. Would it be possible to update the example to show the expected structure of the value?

woodruffw:
Confirmed that this builds locally and renders correctly with `make docs`. My ReST is a little rusty, though, so please give it a pass for idiomatic usage 🙂

Closes #10218.

### DIFF
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -164,9 +164,10 @@ Project
 
     :statuscode 200: no error
 
-    The ``vulnerabilities`` array is always empty for this endpoint. To get a list
-    of known vulnerabilities for a version of a project, refer to the endpoint
-    documented below.
+    On this endpoint, the ``vulnerabilities`` array provides a listing for
+    any known vulnerabilities in the most recent release (none, for the example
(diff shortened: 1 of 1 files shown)


docs/api-reference/json.rst
```diff
@@ -158,11 +158,16 @@ Project
...
-            ]
+            ],
+            "vulnerabilities": []
         }
 
     :statuscode 200: no error
 
+    The ``vulnerabilities`` array is always empty for this endpoint. To get a list
+    of known vulnerabilities for a version of a project, refer to the endpoint
+    documented below.
```
di, line 169: This endpoint defaults to the latest release -- if the latest release has a vulnerability, this would not be empty.

woodruffw: Fixed.

di (approved):

PR #10438, opened 2021-12-03
Sort release filenames naturally on details page

### LINKED ISSUE #10262: Use natural sort order for file listings

Currently on https://pypi.org/project/lxml/4.6.3/#files, the files are listed as:

- lxml-4.6.3-cp27-cp27mu-manylinux1_x86_64.whl
- lxml-4.6.3-cp310-cp310-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_24_i686.whl
- lxml-4.6.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.manylinux_2_24_x86_64.whl
- lxml-4.6.3-cp35-cp35m-manylinux1_i686.whl

This is because the strings are sorted as 27 < 310 < 35, for strings.

Use natural sorting order for filenames, similar to what we did for https://github.com/pypa/trove-classifiers/issues/56.

This _may_ also make sense for the simple pages, where it would be a nice-to-have when a human looks at the page.

miketheman:
Resolves #10262

I spent some time trying to understand if there was a "simple" way to pass `natsorted` (or something) to [SQLAlchemy Custom Comparators](https://docs.sqlalchemy.org/en/14/core/operators.html#operator-customization), but that bore little fruit. My understanding is that whatever custom comparisons added to the `order_by` of the relationship would be performed SQL-side, so we'd not be able use the Python library, and might have to implement a migration to add a stored function to accomplish this.

I also considered adding [`__gt__`-style operators](https://docs.python.org/3.8/reference/datamodel.html#object.__gt__) to `File` class, and then using Python's default `.sort()` or `sorted()`, but that also seemed wrong, since it would apply filename-based sorting to the entire class. I also pursued this in hopes of using [Jinja2 `sort()`](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.sort) directly in the template - but it has the same problem of not using natural sorting.

Finally, I landed on performing the sort in the view, which felt like the right balance.

Note: in trove-classifiers, the sort order is 2.7 => 3.9 => 3.10 - and I reverse the order here - happy to change that, but I felt like the most recent files are going to be more readily useful to a user.

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -24,6 +24,7 @@ limits
 lxml
 mistune
 msgpack
+natsort
 packaging>=15.2
 paginate>=0.5.2
 paginate_sqlalchemy
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -711,6 +711,10 @@ msgpack==1.0.2 \
     --hash=sha256:fae04496f5bc150eefad4e9571d1a76c55d021325dcd484ce45065ebbdd00984 \
     --hash=sha256:fe07bc6735d08e492a327f496b7850e98cb4d112c56df69b0c844dbebcbb47f6
     # via -r requirements/main.in
+natsort==8.0.0 \
+    --hash=sha256:5f5f4ea471d655b1b1611eef1cf0c6d3397095d2d3a1aab7098d6a50e4c3901a \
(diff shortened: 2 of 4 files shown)


di (approved):

di:
Beautiful, nice job! Want to try https://github.com/pypa/warehouse/issues/5792 or https://github.com/pypa/warehouse/issues/8456 next?

PR #10453, opened 2021-12-08
Update `gulp-image` to speed up image build

abitrolly:
Primarily removes dependency on `guetzli` which is slow by itself,
but also compiled from source when installing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2101,6 +2101,11 @@
         }
       }
     },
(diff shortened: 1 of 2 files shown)


miketheman:
Hit this today as well through an unrelated change, glad to see it wasn't just me!

ewdurbin:
closed by #12728

PR #10488, opened 2021-12-14
Update EA client to the latest non-beta version

ericholscher:
We have shipped the features that were originally in our beta client for the PSF into our main client. This PR updates the client version to use the released version.


ewdurbin (approved):

ewdurbin:
Something seems to be rendering incorrectly with this change:

Screen Shot 2021-12-14 at 2 19 45 PM

Looks like the view pixel is now "showing". I'll have to take a closer look.

ewdurbin:
Closing in favor of #10489

ericholscher:
@ewdurbin Thanks for fixing that. Will see what's going on..

PR #10489, opened 2021-12-14
Update EA client to the latest non-beta version

ewdurbin:
Supersedes #10488


ericholscher:
@ewdurbin Thanks for updating this. I'm not able to reproduce the image issue you were seeing. I'm able to reproduce the image failing to load, which seems to happen because of the Acceptable Ads block list doesn't include that URL. But even when the image fails to load, it should be `display: none`, and looks like it is for me:

Screen Shot 2021-12-14 at 11 52 05 AM

ewdurbin:
@ericholscher it boils down to our CSP blocking the ea client from adding the inline style for display: none. adding the hashes to the CSP allows that and correctly styles the pixel.

ericholscher:
@ewdurbin Gotcha, thanks for pushing this out so quickly. 🎉

PR #10497, opened 2021-12-17
Skip GCS upload if blob exists

di:
Our upload is not fully transactional, meaning that this upload may succeed, and the corresponding write to DB may fail. If/when that happens, the distribution will not be on PyPI, but the file will be in the object store, and future repeated upload attempts will fail due missing DB entries for this file, and due to our object store disallowing overwrites.

Because the file_path always includes the file's hash, we can be assured that any attempt to upload a blob that already exists is a result of this edge case, and we can safely skip the upload.

Fixes https://sentry.io/organizations/python-software-foundation/issues/2866042087/

### DIFF
--- a/warehouse/packaging/services.py
+++ b/warehouse/packaging/services.py
@@ -202,4 +202,16 @@ class GCSFileStorage(GenericFileStorage):
         blob = self.bucket.blob(path)
         if meta is not None:
             blob.metadata = meta
-        blob.upload_from_filename(file_path)
+
+        # Our upload is not fully transactional, meaning that this upload may
+        # succeed, and the corresponding write to DB may fail. If/when that
+        # happens, the distribution will not be on PyPI, but the file will be
+        # in the object store, and future repeated upload attempts will fail
+        # due missing DB entries for this file, and due to our object store
+        # disallowing overwrites.
+        #
+        # Because the file_path always includes the file's hash, we can be
+        # assured that any attempt to upload a blob that already exists is a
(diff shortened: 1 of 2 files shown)


warehouse/packaging/services.py
```diff
@@ -202,4 +202,16 @@ def store(self, path, file_path, *, meta=None):
...
-        blob.upload_from_filename(file_path)
+
+        # Our upload is not fully transactional, meaning that this upload may
+        # succeed, and the corresponding write to DB may fail. If/when that
+        # happens, the distribution will not be on PyPI, but the file will be
+        # in the object store, and future repeated upload attempts will fail
+        # due missing DB entries for this file, and due to our object store
+        # disallowing overwrites.
+        #
+        # Because the file_path always includes the file's hash, we can be
```
ewdurbin, line 213: update this comment to make clear that we create the path based on a hash we calculated and I think this is a no-brainer.

ewdurbin (approved):

warehouse/packaging/services.py
```diff
@@ -202,4 +202,16 @@ def store(self, path, file_path, *, meta=None):
...
+        # assured that any attempt to upload a blob that already exists is a
+        # result of this edge case, and we can safely skip the upload.
+        if not blob.exists():
+            blob.upload_from_filename(file_path)
```
ewdurbin, line 217: an

```python
        else:
                report_to_sentry
```

here would definitely help me sleep better, I think these are worth investigating if they happen outside of an outage

PR #10503, opened 2021-12-18
remove redirect to simple added in #9908

ewdurbin:
Simple is an API endpoint, and redirects can be costly to our backends. It _also_ could lead to people relying on this undocumented URL.


di (approved):

asottile:
I really liked this shorthand, it's a shame to see it is gone, I was using this :(

PR #10508, opened 2021-12-20
limit length of response status line for _exc_with_message in forklift/legacy

### LINKED ISSUE #10501: Test pypi upload fails with 502 Bad Gateway error
**Describe the bug**
During a GitHub Action to release the latest version of aurora, the package fails to upload to the pypi test server with a 502 Bad Gateway error

**Expected behavior**
The Action has worked fine in the past

**To Reproduce**
https://github.com/fsciortino/Aurora/runs/4558019561?check_suite_focus=true

**My Platform**
GitHub Actions

Your help is greatly appreciated.  Mention @fsciortino .

ewdurbin:
addresses cases like #10501 where the Response Status line will be much larger than is reasonable.

in theory this is crashing the request at some part of our HTTP stack, I believe between our gunicorn servers and the TLS terminator right above it.

Fixes #10501.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -870,7 +870,7 @@ def file_upload(request):
                 error_message = (
                     "{value!r} is an invalid value for {field}. ".format(
                         value=(
-                            field.data[:60] + "..."
+                            field.data[:30] + "..." + field.data[-30:]
                             if len(field.data) > 60
                             else field.data
                         ),
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -869,7 +869,12 @@ def file_upload(request):
             if field.description and isinstance(field, wtforms.StringField):
                 error_message = (
                     "{value!r} is an invalid value for {field}. ".format(
-                        value=field.data, field=field.description
+                        value=(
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -869,7 +869,12 @@ def file_upload(request):
             if field.description and isinstance(field, wtforms.StringField):
                 error_message = (
                     "{value!r} is an invalid value for {field}. ".format(
-                        value=field.data, field=field.description
+                        value=(
+                            field.data[:60] + "..."
```
di, line 873:
```suggestion
                            field.data[:30] + "..." + field.data[-30:]
```
Just to give a sense of what the entire string is.

ewdurbin: it'd be a little awkward for strings just over the limit... since we would return 63 chars for a 61 char field.. but makes sense :-D

di (approved):

mjon60:
> addresses cases like #10501 where the Response Status line will be much larger than is reasonable.
> 
> in theory this is crashing the request at some part of our HTTP stack, I believe between our gunicorn servers and the TLS terminator right above it.
> 
> Fixes #10501.

gh pr checkout 10508

PR #10547, opened 2022-01-04
chore: handle more warnings in test suite

miketheman:
Continued poking at the environment, and saw that the warnings crept up past 2,000.

Each commit message describes what the specifics of the change are.

Takes the test suite down from ~2,200 warnings to ~108.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -104,6 +104,7 @@ class TestValidation:
         form, field = pretend.stub(), pretend.stub(data=version)
         legacy._validate_pep440_version(form, field)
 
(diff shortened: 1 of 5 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -992,6 +993,7 @@ def test_fails_invalid_version(self, pyramid_config, pyramid_request, version):
             ),
         ],
     )
+    @pytest.mark.filterwarnings("ignore:Creating a LegacyVersion.*:DeprecationWarning")
```
miketheman, line 996: TODO: This warning isn't being captured, and I'm not sure why yet.

di:
Thanks!

Issue #5792, opened 2019-05-06
Add ability to require 2FA for a given project


di:
**What's the problem this feature will solve?**
Currently there is no way to require a certain level of security for a given project.

**Describe the solution you'd like**
Allow a project to require that 2FA is enabled before a new owners/maintainers can be added. This should also prevent users that are not in compliance from doing the activities permitted by their role.

**Additional context**
~Blocked on #5790, otherwise this would allow any attacker to determine the 2FA status for a given user.~

brainwane:
Some UX questions to be resolved include:

* If a user wants to send an invitation and the recipient doesn't have 2FA on, what happens?
* if you get an invite and don't have 2FA on, what happens?
* In projects with multiple owners: can any owner turn that on? Maybe it's not allowed until all the existing owners/maintainers have turned on 2FA? Once turned on, can any owner turn it off by themselves?
* Will we expose anywhere, in public, the info that this is a requirement for this project? (Leaning towards no, since that allows malfeasors to discover which projects DON'T have this requirement.)
* What happens to an existing owner/maintainer of a project if they try to disable 2FA on their account? Maybe they retain maintainer/owner status but are blocked from actually performing any of their privileged actions? Would they be unable to create project-scoped API tokens, for instance?

So we'd want to research this with user tests and check prior art on other similar platforms. That, plus developing flows and templates, makes this a feature that might take something like 2 weeks of UX work.

Until we get into it we won't know how much backend work is involved in accommodating flows and edge cases here, and how much PM work it'll take to document, publicize, triage related issues, etc., but it feels to me like it might take ~6 days of backend and ~4 days of PM work.

It would also be great to have an admin panel to make it easy for a PyPI site administrator to check whether a specific project has this turned on (to help with troubleshooting and potentially with quiet outreach to improve the security of specific high-profile packages). (Doing that, including underpinnings and methodical unit & integration tests, might take about 5 days of work.)

(Roughly scoping this out as we look for additional funding.)

miketheman:
Has there been any further movement on the UX decisions here? Or would this need further work scoping to make those decisions?

PR #10574, opened 2022-01-07
Add mechanisms for requiring 2FA

di:
Towards #5792: adds the bits to make it possible for a project to require 2FA, and adds an authorization policy that prevents authorization for a project if a user doesn't have 2FA enabled.

image

Doesn't include: mechanisms for flipping these bits.


warehouse/templates/manage/projects.html
```diff
@@ -82,6 +83,15 @@ <h3 class="package-snippet__title">{{ project.name }}
           {% endif %}
         </div>
         <div class="package-snippet__buttons">
+
+          {% if not user_has_two_factor and project.name in projects_requiring_2fa %}
+          <a
+            href="{{ request.route_path('manage.account.two-factor') }}"
+            class="button button--danger"
+            title="{% trans %}This project requires 2FA to be enabled to manage{% endtrans %}"
```
ewdurbin, line 91: Since there are multiple reasons 2FA might be required (owners_require_2fa/pypi_mandates_2fa), it's probably best to surface which source is requiring this here.

warehouse/accounts/auth_policy.py
```diff
@@ -45,3 +50,46 @@ def unauthenticated_userid(self, request):
...
+
+        # If the request is permitted by the subpolicy, check if the context is
+        # 2FA requireable, if 2FA is indeed required, and if the user has 2FA
+        # enabled
+        if subpolicy_permits and isinstance(context, TwoFactorRequireable):
+            if context.two_factor_required and not request.user.has_two_factor:
+                return WarehouseDenied(
+                    "This project requires two factor authentication to be enabled "
+                    "for all contributors.",
+                    reason="two_factor_required",
```
ewdurbin, line 86: Similarly, perhaps we should have additional context for what is requiring 2FA?

PR #10575, opened 2022-01-09
Add uploader next to author/maintainer in info json

MartinThoma:
I've noticed that on https://pypi.org/project/identifier/ I can see the user "Collie" as maintainer, but on https://pypi.org/pypi/identifier/json the user cannot be seen at all.

It would be nice to be able to get the same information in a structured way as via the GUI.

### DIFF
--- a/warehouse/legacy/api/json.py
+++ b/warehouse/legacy/api/json.py
@@ -190,6 +190,7 @@ def json_release(release, request):
             "author_email": release.author_email,
             "maintainer": release.maintainer,
             "maintainer_email": release.maintainer_email,
+            "uploader": release.uploader,
             "requires_python": release.requires_python,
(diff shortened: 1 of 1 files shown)


miketheman (commented):
Uploaded value is not guaranteed to be available, per https://github.com/pypa/warehouse/blob/bc54b9db85e331623abe8b51c5954bf1de688c59/warehouse/packaging/models.py#L477

Have you considered how to guard against that possibility?

Adding a test to cover this behavior in https://github.com/pypa/warehouse/blob/main/tests/unit/legacy/api/test_json.py

And specifically using the ReleaseFactory to create a release without an uploaded value would be the way to go.

firefly9966 (approved):

miketheman:

PR #10603, opened 2022-01-16
Implement a way to select preferred package manager

patrick91:
Hi folks! I've made this PR to add a way to chose what package manager you want to use to install a dependency, here's a video of how it works:

I didn't put much effort in the code so feel free to bash it :D I was mostly wondering if this is a feature PyPI needs/wants or not, I've asked on twitter and people seem to like the idea.

In case we want to go forward with this PR there's a few things left to do:

- [ ] Add a cookie to preselect the preferred package manager 
- [ ] Improve the code (based on the PR review)
- [ ] decide if I put the settings form in the right place :)


warehouse/static/js/warehouse/index.js
```diff
@@ -257,3 +257,43 @@ application.load(definitionsFromContext(context));
 docReady(() => {
   RepositoryInfo();
 });
+
+// utils for choosing a package manager
```
miketheman, line 261: **suggestion:** ‏Following patterns that other functionality here has, move the logic out of index.js and into a feature-specific location.
Existing behaviors follow a pattern where they are placed in `warehouse/utils/` and imported here as a single call - like `RepositoryInfo()` on L258 above.

Even better: More of the JS functionality has been moved to Stimulus Controllers in https://github.com/pypa/warehouse/tree/main/warehouse/static/js/warehouse/controllers - and then contextually imported here: https://github.com/pypa/warehouse/blob/479e6f3ca2e11348e70e197dc5b4bf7887411341/warehouse/static/js/warehouse/index.js#L253-L255

The Stimulus approach may even work better for this function.

miketheman (commented):
Very cool feature! I was thinking about this recently as well.

Thinking aloud, is a cookie the right place for storing this preference? Maybe! Or should it be a User property, and persisted on the Model? I don't know the answer!

warehouse/static/js/warehouse/index.js
patrick91: oh, didn't notice stimulus being used 😊

patrick91:
> Thinking aloud, is a cookie the right place for storing this preference? Maybe! Or should it be a User property, and persisted on the Model? I don't know the answer!

Very good questions! I don't know how many people use pypi while being signed it (maybe we have some stats on that?), but I don't think it would be the majority (since you don't need an account to download packages).

warehouse/static/js/warehouse/index.js
miketheman: And for testing - see https://github.com/pypa/warehouse/tree/main/tests/frontend for other stuffs.

miketheman:
> I don't think it would be the majority...

I'm inclined to agree - but am unable to make a data-driven assertion.

miketheman:
@patrick91 Casual ping.

I've been thinking about this recently, and have come around to thinking that using `[localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)` is likely the way to go. I [found a blog post](https://stevepolito.design/blog/rails-auto-save-form-data/) talking about Stimulus and localStorage that might prove helpful.

Are you still interested in pursuing this effort?

patrick91:
@miketheman sure, happy to make a PR in the weekend :D

feel free to ping me again if I don't do it!

patrick91:
@miketheman I've done an initial implementation using stimulus, let me know how it looks, I'm not familiar with it 😊 

I also needed to add a wrapper div to make the target work 🙂

I'll add the local storage config after the first review :)

warehouse/static/js/warehouse/controllers/package_manager_instructions_controller.js
```diff
@@ -0,0 +1,57 @@
...
+    packageValue: String,
+  };
+
+  static targets = ["settings", "command"];
+
+  connect() {
+    console.log(this.packageNameValue);
+    console.log(this.projectVersionValue);
+    console.log(this.indexUrlValue);
+  }
```
miketheman, line 30: **nitpick:** ‏gonna guess these `console.log` and others go away when ready?
Will the `connect()` go away completely as well?

warehouse/templates/packaging/detail.html
```diff
@@ -147,14 +149,39 @@ <h1 class="package-header__name">
...
+        data-package-manager-instructions-package-name-value="{{ release.project.name }}"
+        data-package-manager-instructions-project-version-value="{{ project_version }}"
+        data-package-manager-instructions-index-url-value="{{ index_url }}"
+      >
+        <p class="package-header__pip-instructions">
+          <span id="pip-command" data-package-manager-instructions-target="command">{{ pip_command(release, request.matched_route.name, testPyPI) }}</span>
+          <button type="button" class="copy-tooltip copy-tooltip-s" data-clipboard-target="#pip-command" data-tooltip-label="{% trans %}Copy to clipboard{% endtrans %}">
+            <i class="fa fa-copy" aria-hidden="true"></i>
+              <!-- todo: rename this to installation instructions? -->
+              <span class="sr-only">{% trans %}Copy PIP instructions{% endtrans %}</span>
```
miketheman, line 163: Yes please - since `sr-*` prefixes are for screen readers. If you can shim in the package manager name here, that'd be cool too!

warehouse/templates/packaging/detail.html
```diff
@@ -147,14 +149,39 @@ <h1 class="package-header__name">
...
-      </p>
-      {% endif %}
+      <div
+        data-controller="package-manager-instructions"
+        data-package-manager-instructions-package-name-value="{{ release.project.name }}"
+        data-package-manager-instructions-project-version-value="{{ project_version }}"
+        data-package-manager-instructions-index-url-value="{{ index_url }}"
+      >
+        <p class="package-header__pip-instructions">
+          <span id="pip-command" data-package-manager-instructions-target="command">{{ pip_command(release, request.matched_route.name, testPyPI) }}</span>
```
miketheman, line 159: The `pip_command` macro is located above and contains this snippet:
https://github.com/pypi/warehouse/blob/3b4a3113dcd0ad438c2ceebda3c274351561042f/warehouse/templates/packaging/detail.html#L85-L87

This has the affect of transforming the copyable command from

Screenshot 2023-01-08 at 5 28 29 PM

to:

Screenshot 2023-01-08 at 5 28 48 PM

See https://pypi.org/project/hellowebapp-deploy/ vs https://pypi.org/project/hellowebapp-deploy/1.0.5/

By changing `innerHTML`, the existing behavior is broken after a change away from `pip` and back.

It'd be great map out the correct syntax for each package manager and version specifics, but at the least we shouldn't break current behavior.

miketheman (changes requested):
@patrick91 This is an awesome start. It looks slick, and works as advertised, especially for your first time with Stimulus!
The wrapping `div` is totally the way to go, unless there's an existing `div` to hook on to - I think you made the right call.

I left a few inline notes - let me know if they need further clarity.

warehouse/static/js/warehouse/controllers/package_manager_instructions_controller.js
patrick91: yup, I left them so I don't forget to handle the project version and the index url (if this is needed)

ewdurbin:
I'm not in support of merging this. PyPI tries to remain as neutral as is reasonable when it comes to tooling.

I'm concerned that if we add a list of clients it will be assumed to be exhaustive and/or to act as an endorsement of those tools.

`pip` is bundled with all supported Python versions so I think it is a reasonable assumption and sane default as is.

It may be worth opening an issue to discuss the feature before continuing on implementation. I think it *could* be such a switcher, something more comprehensive, or no action.

PR #10628, opened 2022-01-20
Interfaces and services for JWK management

### LINKED ISSUE #10620: Limetime management for OIDC JWKS
The first step in verifying the OIDC JWTs is verifying their signatures, which means checking them against the OIDC provider's signing key set.

There are a few approaches we could take to acquiring and maintaining the updatedness of those keys:

1. Bake them into the Warehouse codebase. We could pull the current JWKS blob from the URI referenced in the provider's `openid-configuration`.
  * Pros: Simple.
  * Cons: More tedious to update in the event of key updates/rotations; requires a PR each time any provider does so. Gaps between the rotation and updates means that Warehouse would probably reject authentic JWTs.

2. Fetch the JWKS for each provider on Warehouse startup/initialization. Restarts always re-fetch the JWKS.
  * Pros: Simple.
  * Cons: Most of the same cons as (1).

3. All of (2), but we also include a periodic job that checks each OIDC provider's JWKS on a schedule.
  * Pros: No rotation downtime.
  * Cons: Requires the most code.

woodruffw:
This adds the initial set of interfaces and services (and a small `warehouse oidc` CLI) for JSON Web Key (JWK) management.

Some design notes:

* JWKs come in sets, and are keyed by their OIDC provider. For now, GitHub is the only provider Warehouse knows about.
* The current implementation stashes the JWKs in Redis. I'm not 100% sure if this is the best place for them, so please let me know if somewhere else is better. 
* The current implementation adds a periodic (daily) job for refreshing the JWKs from each service. A user standing up a fresh instance of Warehouse can also manually fetch the JWKs by running `warehouse oidc update-oidc-jwks`.
* I'm not sure whether or not it also makes sense to cache the OIDC configuration (and periodically re-fetch it). The only thing I can currently think of that we might want from it is the `issuer` key, but we should probably keep that hardcoded anyways. 

Needs:

- [x] More documentation
- [x] Unit tests

Closes #10620.


warehouse/tasks.py
```diff
@@ -207,7 +207,10 @@ def includeme(config):
             Queue("default", routing_key="task.#"),
             Queue("malware", routing_key="malware.#"),
         ),
-        task_routes={"warehouse.malware.tasks.*": {"queue": "malware"}},
+        task_routes={
+            "warehouse.malware.tasks.*": {"queue": "malware"},
+            "warehouse.oidc.tasks.*": {"queue": "oidc"},
```
ewdurbin, line 212: Do we really anticipate that these tasks will be long running/disruptive enough to require their own queue?

ewdurbin: I only ask because using SQS as a broker, requires setting up a completely separate SQS queue for each route.

woodruffw: Nope, they should in general be very fast. I can remove the queue!

warehouse/oidc/services.py
```diff
@@ -0,0 +1,73 @@
...
+            if not resp.ok:
+                logger.error(f"error querying JWKS JSON for {provider}: {jwks_url}")
+                continue
+
+            jwks_conf = resp.json()
+            keys = jwks_conf["keys"]
+
+            yield (provider, keys)
+
+    def keyset_for_provider(self, provider):
```
woodruffw, line 69: N.B.: It might make sense to return a more structured type here, perhaps just a `PyJWKSet` (even though that's just a thin wrapper over the list of dicts).

woodruffw: Removed this interface.

warehouse/cli/oidc.py
```diff
@@ -0,0 +1,57 @@
...
+def oidc():
+    """
+    Manage the Warehouse OIDC components.
+    """
+
+
+@oidc.command()
+@click.pass_obj
+@click.argument("provider")
+def prime_provider(config, provider):
```
woodruffw, line 30: N.B.: This is only useful for local testing now that PyPI fetches each keyset on first use.

woodruffw:
Updated the `warehouse oidc` CLI now that we do FOFU:

```
Usage: python -m warehouse oidc [OPTIONS] COMMAND [ARGS]...

  Manage the Warehouse OIDC components.

Options:
  -h, --help  Show this message and exit.

Commands:
  get-key         Return the JWK for the given provider's...
  prime-provider  Update Warehouse's JWK sets for the given...
```

woodruffw:
I haven't started on tests yet, but this should be good for another review pass.

warehouse/oidc/services.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Another sanity test: an OIDC provider should never return an empty
+        # keyset, but there's nothing stopping them from doing so. We don't
+        # want to cache an empty keyset just in case it's a short-lived error,
+        # so we check here, error, and return None instead.
+        if len(keys) == 0:
+            logger.error(f"{provider} returned JWKS conf but no keys")
+            return None
+
+        return keys
```
di, line 111:
```suggestion
        return {key["kid"]: key for key in keys}
```

warehouse/oidc/services.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_key(self, provider, key_id):
```
di, line 113: I'd like to see this simplified a bit, something like:

```python
    def get_key(self, provider, key_id):
        keyset = self._get_keyset()
        if key_id not in keyset:
            keyset = self._refresh_keyset()
        if key_id not in keyset:
            # emit metrics
            return None
        return PyJWT(keyset[key_id])
```

This function doesn't need to know anything about timeouts or how to store a keyset, which can both be handled by `_refresh_keyset()`. Similarly `_refresh_keyset()` doesn't need to know about what `key_id` we're looking for here, just that the keyset might be stale.

warehouse/utils/oidc.py
```diff
@@ -0,0 +1,27 @@
...
+
+# This is a map of OIDC providers supported by Warehouse.
+# The keys are human-readable names for each provider, and the values
+# are the "issuer" (i.e., `iss`) FQDN that JWTs signed by their respective
+# JWKs are checked against.
+OIDC_PROVIDERS = {
+    "github": "https://token.actions.githubusercontent.com",
+}
+
+WELL_KNOWN_OIDC_CONF = ".well-known/openid-configuration"
```
di, line 27: Is this required by the spec or could it vary from one provider to the next? If it's the latter, we probably want to do something like:

```suggestion
OIDC_PROVIDERS = {
    "github": { 
        "host": "https://token.actions.githubusercontent.com",
        "conf-path": ".well-known/openid-configuration"
    }
}
```

di (commented):
Overall looks good. We need to make sure we're emitting meaningful metrics at the right places. We probably want to use `sentry_sdk.capture_message` for the errors instead of logging them (would be nice to make this just output to the logger in development).

requirements/main.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with python 3.8
 # To update, run:
 #
-#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements/main.txt requirements/main.in
+#    pip-compile --allow-unsafe --generate-hashes --output-file=main.txt main.in
```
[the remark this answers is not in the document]
woodruffw: Yeah, I had some local development strangeness with this: running `make requirements/main.txt` wouldn't update it file, even though `main.in` had changed and I could see `pip-compile` running. I had to `cd requirements` and run `pip-compile` manually. But I'll fix up for now and try to debug that later...

warehouse/utils/oidc.py
woodruffw: Yep, it's required by the spec -- every OIDC provider *must* provide that well-known URI.

(That's in contrast to the JWKS URI, which is usually under `.well-known` but isn't actually required to be.)

warehouse/oidc/services.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        with redis.StrictRedis.from_url(self._cache_url) as r:
+            keys = r.get(self._provider_jwk_key(provider))
+            timeout = bool(r.exists(self._provider_timeout_key(provider)))
+            if keys is not None:
+                return (json.loads(keys), timeout)
+            else:
+                return (None, timeout)
+
+    def _fetch_keyset(self, provider):
```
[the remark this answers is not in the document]
woodruffw: Just making sure I understand: in that case, this function should be connecting to Redis to check the timeout, correct? As well as using/updating the cached keys as necessary?

warehouse/oidc/services.py
woodruffw: Sounds good! That's a much better isolation of concerns; I'll refactor towards this.

warehouse/oidc/services.py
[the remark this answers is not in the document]
di: We could factor the Redis stuff into helper classes (`_get_timeout`/`_set_timeout`) but yeah.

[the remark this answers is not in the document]
woodruffw: Yeah, I currently have it reusing the `_get_keyset` and `_store_keyset` helpers. Will push up in a moment for you to check 🙂

warehouse/oidc/services.py
```diff
@@ -0,0 +1,159 @@
...
+        return self.service_class(self.provider, cache_url)
+
+    def __eq__(self, other):
+        if not isinstance(other, JWKServiceFactory):
+            return NotImplemented
+
+        return (self.provider, self.service_class) == (
+            other.provider,
+            other.service_class,
+        )
```
woodruffw, line 159: cc @di: Let me know if this looks reasonable for the factory.

(I cargo-culted the `__eq__` implementation from `TokenServiceFactory` -- I assume it's necessary to help `find_service` disambiguate the factory-produced services?)

warehouse/oidc/services.py
```diff
@@ -0,0 +1,159 @@
...
+        Return a JWK for the given key ID, or None if the key can't be found
+        in this provider's keyset.
+        """
+
+        keyset, _ = self._get_keyset()
+        if key_id not in keyset:
+            keyset = self._refresh_keyset()
+        if key_id not in keyset:
+            # TODO: Metrics
+            return None
```
woodruffw, line 138: N.B.: I'll work on adding metrics here.

warehouse/oidc/services.py
```diff
@@ -0,0 +1,168 @@
...
+        in this provider's keyset.
+        """
+
+        keyset, _ = self._get_keyset()
+        if key_id not in keyset:
+            keyset = self._refresh_keyset()
+        if key_id not in keyset:
+            self.metrics.increment(
+                "warehouse.oidc.get_key.error", tags=[f"key_id:{key_id}"]
+            )
```
woodruffw, line 145: cc @di Does this metric look reasonable to you? I used the same pattern as I've seen elsewhere in Warehouse (fixed metric ID, dynamic tag formatted as `key:value`).

warehouse/oidc/services.py
```diff
@@ -0,0 +1,168 @@
...
+        Returns the refreshed keyset, or the cached keyset if a timeout is
+        in effect.
+
+        Returns the cached keyset on any provider access or format errors.
+        """
+
+        # Fast path: we're in a cooldown from a previous refresh.
+        keys, timeout = self._get_keyset()
+        if timeout:
+            self.metrics.increment("warehouse.oidc.refresh_keyset.timeout")
```
woodruffw, line 75: This is the only other metric I added -- I figured the sentry cases are error cases that don't need metrics, since they ping the maintainers and indicate errors that are outside of Warehouse's ability to prevent anyways.

warehouse/oidc/interfaces.py
```diff
@@ -0,0 +1,34 @@
...
+
+class IJWKService(Interface):
+    def create_service(context, request):
+        """
+        Create the service, given the context and request for which it is
+        being created.
+        """
+        pass
+
+    def get_key(provider, key_id):
```
di, line 25:
```suggestion
    def get_key(key_id):
```

warehouse/utils/oidc.py
```diff
@@ -0,0 +1,27 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# For now, only RS256 is supported.
+# We probably won't need to support providers with only symmetric keys
+# (e.g. HS256) in the foreseeable future.
+VALID_ALGS = {"RS256"}
```
di, line 17: This doesn't seem to be used anywhere.

warehouse/utils/oidc.py
```diff
@@ -0,0 +1,27 @@
...
+
+# This is a map of OIDC providers supported by Warehouse.
+# The keys are human-readable names for each provider, and the values
+# are the "issuer" (i.e., `iss`) FQDN that JWTs signed by their respective
+# JWKs are checked against.
+OIDC_PROVIDERS = {
+    "github": "https://token.actions.githubusercontent.com",
+}
```
di, line 25: I think we can just include this when we initialize the service.

warehouse/oidc/services.py
```diff
@@ -0,0 +1,168 @@
...
+            return keys
+
+        oidc_url = f"{oidc.OIDC_PROVIDERS[self.provider]}/{oidc.WELL_KNOWN_OIDC_CONF}"
```
di, line 78:
```suggestion
        oidc_url = f"{self.token_request_url}/.well-known/openid-configuration"
```

This is the only place this is used, don't think we need it to be a constant.

warehouse/utils/oidc.py
di, line 27: This can get moved to where it's used.

warehouse/oidc/services.py
```diff
@@ -0,0 +1,168 @@
...
+
+from warehouse.metrics.interfaces import IMetricsService
+from warehouse.oidc.interfaces import IJWKService
+from warehouse.utils import oidc
+
+
+@implementer(IJWKService)
+class JWKService:
+    def __init__(self, provider, cache_url, metrics):
+        self.provider
```
di, line 30:
```suggestion
    def __init__(self, provider, token_request_url, cache_url, metrics):
        self.provider = provider
        self.token_request_url = token_request_url
```

di (commented):
Some slight refactoring suggestions and one overall thought: Should we make this an `OIDCProviderService` instead? I don't see a need for having separate JWK service and OIDC service (with an interface like `.verify(token)`) for the same provider.

woodruffw:
> Should we make this an `OIDCProviderService` instead?

Yep, that makes sense to me! I'll make that change.

warehouse/utils/oidc.py
woodruffw: Yep, it isn't used anywhere yet -- my plan was to use it to verify the incoming JWTs, since we'll only accept ones that have `alg=RS256`.

di: I think this can probably live where that verification happens instead.

warehouse/oidc/services.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@implementer(IOIDCProviderService)
+class OIDCProviderService:
+    def __init__(self, provider, cache_url, metrics):
+        self.provider = provider
+        self.cache_url = cache_url
+        self.metrics = metrics
+
+        self._issuer_url = oidc.OIDC_PROVIDERS[self.provider]
```
woodruffw, line 34: @di Is this what you had in mind, re: https://github.com/pypa/warehouse/pull/10628/files#r794061431?

warehouse/utils/oidc.py
woodruffw: Deleted it entirely, since it's always the same path.

warehouse/utils/oidc.py
woodruffw: Deleted.

warehouse/utils/oidc.py
woodruffw: https://github.com/pypa/warehouse/pull/10628#discussion_r794085394

warehouse/oidc/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from warehouse.oidc.interfaces import IOIDCProviderService
+from warehouse.oidc.services import OIDCProviderServiceFactory
+
+
+def includeme(config):
+    config.register_service_factory(
+        OIDCProviderServiceFactory(provider="github"),
```
di, line 19: Sorry, this comment either never got saved or was lost in the refactor:

```suggestion
        OIDCProviderServiceFactory(
            provider="github",
            token_request_url="https://token.actions.githubusercontent.com",
        ),
```

warehouse/oidc/services.py
di: No, see https://github.com/pypa/warehouse/pull/10628#discussion_r794063016 and https://github.com/pypa/warehouse/pull/10628#discussion_r794061762 and https://github.com/pypa/warehouse/pull/10628#discussion_r794088544

warehouse/oidc/__init__.py
woodruffw: Ah! Okay, that makes sense now. Will refactor towards that.

woodruffw: In this case we don't actually need the `provider` kwarg at all, since it's only being used for logging at this point.

woodruffw:
```python
        OIDCProviderServiceFactory(
            issuer_url="https://token.actions.githubusercontent.com"
        ),
        IOIDCProviderService,
        name="github",
    )
```

woodruffw:
Dependencies failure in the CI looks unrelated to these changes:

```
# install ipython if enabled
/opt/hostedtoolcache/Python/3.8.9/x64/bin/pip-compile --upgrade --allow-unsafe -o /tmp/tmp.skKABnHln0/deploy.txt requirements/deploy.in > /dev/null
  File "/opt/hostedtoolcache/Python/3.8.9/x64/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/piptools/scripts/compile.py", line 342, in cli
    repository = PyPIRepository(pip_args, cache_dir=cache_dir)
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/piptools/repositories/pypi.py", line 106, in __init__
    self._setup_logging()
  File "/opt/hostedtoolcache/Python/3.8.9/x64/lib/python3.8/site-packages/piptools/repositories/pypi.py", line 455, in _setup_logging
    assert isinstance(handler, logging.StreamHandler)
AssertionError
make[1]: *** [Makefile:121: deps] Error 1
make[1]: Leaving directory '/home/runner/work/warehouse/warehouse'
make: *** [Makefile:139: github-actions-deps] Error 2
Error: Process completed with exit code 2.
```

di (approved):

Issue #9066, opened 2021-02-09
Add badges/info about abi3 packages forward compatibility


webknjaz:
**What's the problem this feature will solve?**

When looking at the list of wheels in a given release of a Python package distribution, there may be a lot of files and the compat column only takes a Python version part into consideration ignoring the ABI.

**Describe the solution you'd like**

Annotate the dists with something (probably a badge, like for pre- and yanked releases) that would clearly indicate support for Python 3.x+, not just 3.x, in those wheels.

In doing so, PyPI will improve the users' understanding of what to expect from `pip install`.

**Additional context**

I realized that this info is missing from the website after seeing the confusion in these issues:
* https://github.com/pypa/gh-action-pypi-publish/issues/56
* https://github.com/pyca/cryptography/issues/5787
* https://github.com/pyca/cryptography/issues/5798

![cryptography-py36-abi3](https://user-images.githubusercontent.com/578543/107395932-92608000-6afd-11eb-967d-ad237f90b8db.png)

di:
Since this issue was created, the release files page has changed a bit. The `cp36` column is now a flag on the release file:

<img width="601" alt="image" src="https://user-images.githubusercontent.com/294415/184384676-9e78fe9c-e761-41fe-a369-f96e61638640.png">

I think it'd be good to turn this into one or more tags that are more explicit and human-friendly about what Python versions the artifact is compatible with.

webknjaz:
xref #10646

webknjaz:
xref #15606

PR #10646, opened 2022-01-26
Revamp downloads table

di:
This separates source distributions and built distributions, and brings the sdist(s) to the top. It also handles long filenames better per https://github.com/pypa/warehouse/issues/10448#issuecomment-1021443408, while retaining all the same information we already provide.

Example for a project w/ just 1 sdist and 1 bdist:
image

Project with multiple bdists:
image

### DIFF
--- a/tests/unit/packaging/test_views.py
+++ b/tests/unit/packaging/test_views.py
@@ -196,6 +196,7 @@ class TestReleaseDetail:
                 release=r,
                 filename="{}-{}.tar.gz".format(project.name, r.version),
                 python_version="source",
+                packagetype="sdist",
             )
             for r in releases
         ]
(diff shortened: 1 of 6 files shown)


alex:
Do you think this will make it easier to resolve https://github.com/pypa/warehouse/issues/10308 ?

di:
@alex Yes

alex:
<3 you love to hear it.

ewdurbin:
Only thought here is that the python version identifiers don't have some kind of label. The table headers at least provide some pointer to what those somewhat arbitrary looking strings mean.

Otherwise 👍🏼

di:
> Only thought here is that the python version identifiers don't have some kind of label. The table headers at least provide some pointer to what those somewhat arbitrary looking strings mean. Perhaps working towards #10308 will resolve this.

Agreed, I think working towards #10308 and replacing these with better 'tags' will resolve this.

webknjaz:
xref #9066

Issue #8897, opened 2020-12-07
Encourage generation of 2FA recovery codes


di:
In order to cut down on 2FA reset requests, PyPI should encourage/require users to generate recovery codes.

Some ideas:

* Require recovery codes to be generated before upload can succeed if 2FA is enabled
* Require recovery codes to be generated before the user can log in
* Require recovery codes to be generated as part of the 2FA setup
  * Additionally, require the user to burn one of the codes immediately after generation to ensure the codes have actually been downloaded/stored.

ewjoachim:
> Additionally, require the user to burn one of the codes immediately after generation to ensure the codes have actually been downloaded/stored.

I think this is the only way to make sure users actually did write them somewhere.

If we could also check sometimes at login time that they have it (e.g. add a checkbox once every ~2 months asking them to check that they can still access their recovery code, without requesting that they enter one, just to make sure they're aware that it's a thing they might need, and highlighting that if they loose their 2FA app without recovery codes, they will be permanently locked out without any recovery procedure)

PR #10707, opened 2022-02-09
Improvements to recovery codes

di:
Some improvements to recovery codes, towards #8897. The big improvement is that instead of deleting recovery codes when they are used, they're marked as 'burned'.

### DIFF
--- a/warehouse/accounts/services.py
+++ b/warehouse/accounts/services.py
@@ -30,9 +30,11 @@ import warehouse.utils.otp as otp
 import warehouse.utils.webauthn as webauthn
 
 from warehouse.accounts.interfaces import (
+    InvalidRecoveryCode,
     IPasswordBreachedService,
     ITokenService,
     IUserService,
+    NoRecoveryCodes,
     TokenExpired,
     TokenInvalid,
     TokenMissing,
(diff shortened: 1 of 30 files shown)


warehouse/accounts/services.py
```diff
@@ -586,6 +587,7 @@ def database_login_factory(context, request):
     return DatabaseUserService(
         request.db,
         metrics=request.find_service(IMetricsService, context=None),
+        remote_addr=request.remote_addr,
```
di, line 612: Here's the key change in this refactor: the `remote_addr` is passed in when creating the `DatabaseUserService`, instead of being passed to every function on the `DatabaseUserService`.

warehouse/templates/manage/account/recovery_codes-provision.html
```diff
@@ -61,10 +61,8 @@ <h2>{% trans %}Save your recovery codes{% endtrans %}</h2>
               <button type="button" class="button button--small copy-tooltip copy-tooltip-w" data-tooltip-label="{% trans %}Copy to clipboard{% endtrans %}" data-clipboard-text="{{ 'PyPI recovery codes\n' + recovery_codes|join('\n') }}">
                 {% trans %}Copy{% endtrans %}
               </button>
-              {% set _recovery_codes_filename = "PyPI-Recovery-Codes.txt" %}
-              {% if testPyPI %}
-                {% set _recovery_codes_filename = "TestPyPI-Recovery-Codes.txt" %}
-              {% endif %}
+              {% set _sitename = "TestPyPI" if testPyPI else "PyPI" %}
+              {% set _recovery_codes_filename = _sitename + "-Recovery-Codes-" + request.user.recovery_codes[0].generated.isoformat() + ".txt" %}
```
ewdurbin, line 65: Thinking... should we add the user name also?

ewdurbin (commented):
Fine as is, but a thought and one concern.

di:
@ewdurbin Unfortunately our reauthentication functionality is somewhat naive and can't handle a view that must always be reauthenticated, as it results in an infinite loop of reauthentication. In addition, it doesn't work for requests like this that require a `POST` request, because it doesn't have the ability to 'forward' such a request after reauthentication.

As a compromise, I made the reauthentication period configurable and set it to 10s for this view, and also added notification emails when codes are generated and a code is used.

PR #10725, opened 2022-02-11
Send a recovery code reminder email on login

### LINKED ISSUE #8897: Encourage generation of 2FA recovery codes

di:
For users that have 2FA, but don't have recovery codes generated, email them on every login.

I think this and #10707 is enough to resolve #8897.

Fixes #8897.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -58,6 +58,7 @@ from warehouse.email import (
     send_email_verification_email,
     send_password_change_email,
     send_password_reset_email,
+    send_recovery_code_reminder_email,
 )
 from warehouse.packaging.models import (
     JournalEntry,
@@ -282,6 +283,10 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
                 .hexdigest()
                 .lower(),
             )
+
+            if not request.user.has_recovery_codes:
+                send_recovery_code_reminder_email(request, request.user)
+
             return resp
         else:
             form.totp_value.data = ""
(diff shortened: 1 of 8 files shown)


warehouse/accounts/views.py
```diff
@@ -372,6 +376,9 @@ def webauthn_authentication_validate(request):
             .hexdigest()
             .lower(),
         )
+
+        send_recovery_code_reminder_email(request, request.user)
```
ewdurbin, line 380: I don't see how this is conditional on _not_ having recovery codes... seems like it'd fire regardless?

warehouse/accounts/views.py
```diff
@@ -282,6 +283,9 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
                 .hexdigest()
                 .lower(),
             )
+
+            send_recovery_code_reminder_email(request, request.user)
```
ewdurbin, line 287: I don't see how this is conditional on _not_ having recovery codes... seems like it'd fire regardless?

warehouse/accounts/views.py
```diff
@@ -282,6 +283,10 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
                 .hexdigest()
                 .lower(),
             )
+
+            if not request.user.has_recovery_codes:
```
ewdurbin, line 287: We have the `two_factor_state` dict constructed above... I'm not sure if it costs us anything to call `request.user.has_recovery_codes`, but we might just replace this with `if not two_factor_state.get('has_recovery_codes', False):`

ewdurbin (approved):

PR #10752, opened 2022-02-14
Update pytest-postgresql to version supporting pytest 7

fizyk:
@di Here's a pytest-postgresql version supporting pytest 7, but still working psycopg2.


di:
Awesome, thank you!

PR #10753, opened 2022-02-14
Models, routes and views for creating OIDC publishers

### LINKED ISSUE #10617: Update Warehouse's models to accomodate OIDC verification
Original thought: Warehouse's `User` model, at minimum, will probably need one or more new columns (possibly relations) for each supported provider.

Upon conversation with @di: we don't actually need the AuthN step, so we can probably get away without any changes to the `User` model. Instead, each project model will need a GitHub repository URL and workflow name to check incoming JWTs against.

woodruffw:
~~This is nowhere near done; just wanted to get something up.~~

Remaining items:

- [x] Unit tests
- [x] Feature-gate these changes (admins only?)
- [x] Rate limiting

Closes #10617.


warehouse/oidc/models.py
```diff
@@ -0,0 +1,48 @@
...
+    repository_name = Column(String)
+    owner = Column(String)
+    owner_id = Column(String)
+    workflow_name = Column(String)
+
+    def repository(self):
+        return f"{self.owner}/{self.repository_name}"
+
+    def job_workflow_ref(self):
+        return f"{self.repository}/.github/workflows/{self.workflow_name}.yml"
```
woodruffw, line 45: cc @di: How clever do you want us to get with the "claims" here?

I could add a `@claim` decorator that transparently updates the instance's `self.claims` list as you suggested in your example, but I wasn't sure whether that amount of magic was worth it for what'll end up being just a handful of methods.

warehouse/oidc/services.py
```diff
@@ -148,10 +148,51 @@ def get_key(self, key_id):
...
+        """
+        Return a JWK suitable for verifying the given JWT.
+
+        The JWT is not verified at this point, and this step happens
+        prior to any verification.
+        """
+        unverified_header = jwt.get_unverified_header(token)
+        return self.get_key(unverified_header["kid"])
 
     def verify(self, token):
```
woodruffw, line 163: N.B.: It might make sense to rename this to `verify_signature` (or even `verify_signature_only`), to emphasize that this service is only providing a signature verification check, not authN or authZ for a specific action.

warehouse/oidc/services.py
```diff
@@ -148,10 +148,51 @@ def get_key(self, key_id):
...
+                # that enforces their actual validity.
+                require=["iss", "iat", "nbf", "exp", "aud"],
+                verify_iss=True,
+                verify_iat=True,
+                verify_nbf=True,
+                verify_exp=True,
+                verify_aud=True,
+                issuer=self.issuer_url,
+                audience="pypi",
+                leeway=30,
```
woodruffw, line 185: N.B.: `leeway` defines the permitted clock skew, i.e. +/- 30 seconds. I chose that number because it seems to be the recommended best practice for JWTs (higher numbers indicate a server problem).

warehouse/oidc/models.py
di: I mostly just to make it easy to verify all claims and make it hard to forget to verify some claim.

Regarding the decorator, it'd probably be just as likely that we forget to add it as it would be that we forget to verify the claim.

Maybe something like explicit like `__verified_claims__ == {'repository', 'job_workflow_ref', ...}` and `__unverified_claims__ == {'<something else>'}` and we ensure that every claim in the token is either explicitly verified or unverified? And in the case where there's a claim that's unaccounted for (either it's new or we aren't verifying it) an appropriate warning is generated? (probably not an outright failure b/c we probably don't want to fail on every new claim?)

warehouse/oidc/services.py
di: Agreed

warehouse/oidc/models.py
woodruffw: Sounds good to me!

woodruffw: `__unverified_claims__` will have to include the set of "standard" claims that are verified during signature verification, but we can document that explicitly.

woodruffw: Okay, this is what I've come up with:

```python
    __verified_claims__ = {
        "repository",
        "job_workflow_ref",
        "actor",
        "workflow",
    }

    __preverified_claims__ = {
        "iss",
        "iat",
        "nbf",
        "exp",
        "aud",
    }
    
    __unverified_claims__ = {
        "jti",
        "sub",
        "ref",
        "sha",
        "run_id",
        "run_number",
        "run_attempt",
        "head_ref",
        "base_ref",
        "event_name",
        "ref_type",
    }
```

We need to think some more about how we actually do the verification step for each claim: most claims are just a string equality check, but some are prefix checks instead (like `job_workflow_ref`). Do we want to something dynamic like fetching a `verify_{claim}` function for each claim and calling that? That's pretty magical, so I'm not sure if that's the best approach.

di: Maybe each claim could just be `@property` on the class? So verification would be something like:

```python
def verify_claims(self, signed_token):
    for claim_key in self.__verified_claims__:
        if not getattr(self, claim_key) == signed_token[claim_key]:
            return False
    return True
```

woodruffw: That's pretty close to what we want, although we need more complex behavior than just equality checking. Perhaps something like this?

```python
def verify_claims(self, signed_token):
    for claim_key in self.__verified_claims__:
        if not getattr(self, claim_key)(signed_token[claim_key]):
            return False
    return True
```

i.e., each would be a function instead of a property, allowing us to do e.g. substring matching.

warehouse/oidc/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+    oidc_provider_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey("oidc_providers.id"),
+        nullable=False,
+        primary_key=True,
+    )
+    project_id = Column(
+        UUID(as_uuid=True), ForeignKey("projects.id"), nullable=False, primary_key=True
+    )
```
woodruffw, line 33: N.B.: Let me know if there's a better way to do this many-many mapping. This is the pattern I'm familiar with, but there are probably others.

woodruffw:
Just pasting some screenshots for visibility.

Separate "publishing" page on each project for OIDC management:

Screen Shot 2022-02-18 at 4 51 38 PM

Flash on successful OIDC provider configuration; removal table:

Screen Shot 2022-02-18 at 4 52 35 PM


Security events for OIDC management:

Screen Shot 2022-02-18 at 4 51 01 PM

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,97 @@
...
+
+        owner, repository = repository_slug.split("/", 1)
+
+        # To actually validate the owner, we ask GitHub's API about them.
+        # We can't do this for the repository, since it might be private.
+        response = requests.get(
+            f"https://api.github.com/users/{owner}",
+            headers={"Accept": "application/vnd.github.v3+json"},
+            allow_redirects=True,
+        )
```
di, line 58: This should use `GITHUB_TOKEN` to authenticate, see an example [here](https://github.com/pypa/warehouse/blob/844499dc6c169a55adba4f7075fcdf64b89a6f9b/warehouse/integrations/github/utils.py#L144-L147).

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if response.status_code == 404:
+            raise wtforms.validators.ValidationError(
+                _("Unknown GitHub user or organization.")
+            )
+        elif not response.ok:
+            raise wtforms.validators.ValidationError(
+                _("Unexpected error from GitHub. Try again.")
+            )
```
di, line 67: Another thing that can fail here is that our API quota has been consumed. We should provide a meaningful error if that's the case (and also generate a sentry message so we are aware).

We probably also need to ratelimit this action per-user and per-IP so a malicious user can't exhaust our quota.

warehouse/templates/manage/publishing.html
```diff
@@ -0,0 +1,129 @@
...
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "manage_project_base.html" %}
+
+{% set active_tab = 'publishing' %}
+
+{% block title %}
+  {% trans %}OIDC provider management{% endtrans %}
```
di, line 20: We probably need to think up a more user-friendly way to refer to this feature, most users aren't going to know what OIDC is.

warehouse/templates/manage/publishing.html
```diff
@@ -0,0 +1,129 @@
...
+      {% trans trimmed %}
+        PyPI projects can use trusted OIDC providers to automate their release
+        processes, without having to explicitly provision or manage API tokens.
+      {% endtrans %}
+    </p>
+
+    <h2 class="no-bottom-padding">{% trans %}Add a new provider{% endtrans %}</h2>
+
+    <h3>GitHub</h3>
+
```
di, line 71: We should link to https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect somewhere, probably here.

warehouse/templates/manage/publishing.html
```diff
@@ -0,0 +1,129 @@
...
+    <form method="POST" action="{{ request.route_path('manage.project.settings.oidc', project_name=project.name) }}">
+      <div class="form-group">
+        <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+        <label for="repository" class="form-group__label">
+          {% trans %}Repository{% endtrans %}
```
di, line 76: This is both the owner and the repository (the slug?) so this label is not quite right. Furthermore I'm not sure it's necessary for us to take these as combined input if we're just going to split them eventually anyways.

warehouse/manage/views.py
```diff
@@ -1047,6 +1049,119 @@ def toggle_2fa_requirement(self):
...
+                self.request.session.flash(
+                    f"{provider} is already registered with {self.project.name}",
+                    queue="error",
+                )
+                return {**self.default_response, "github_provider_form": form}
+
+            self.project.oidc_providers.append(provider)
+
+            self.project.record_event(
+                tag="project:oidc:provider-added",
```
di, line 1119: Let's make sure to add a special case for these in the template where it gets rendered, so it doesn't show up as the tag name and we can include the `additional` metadata as well.

Additionally, some of the `additional` metadata seems fairly GitHub specific. We should make sure we either differentiate based on providers or publish provider-specific events.

warehouse/manage/views.py
```diff
@@ -1047,6 +1049,119 @@ def toggle_2fa_requirement(self):
...
+                ip_address=self.request.remote_addr,
+                additional={
+                    "provider": provider.provider_name,
+                    "id": str(provider.id),
+                    "repository": form.repository_slug.data,
```
di, line 1124: Again, let's avoid the conflation of 'repository' and 'repository slug' here.

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,97 @@
...
+from warehouse import forms
+from warehouse.i18n import localize as _
+
+# This roughly matches the "owner/repo" convention used by GitHub.
+_VALID_GITHUB_OWNER_REPO_SLUG = re.compile(
+    r"^[a-zA-Z0-9][a-zA-Z0-9-]*/[a-zA-Z0-9-_.]+$"
+)
+
+
+class GitHubProviderForm(forms.Form):
```
di, line 27: Doesn't seem like we're checking anywhere if this particular provider configuration already exists. We should make sure to enforce uniqueness here.

warehouse/oidc/models.py
```diff
@@ -0,0 +1,168 @@
...
+    def repository(self):
+        return f"{self.owner}/{self.repository_name}"
+
+    @property
+    def job_workflow_ref(self):
+        return f"{self.repository}/.github/workflows/{self.workflow_name}.yml"
+
+    @property
+    def actor(self):
+        return self.owner
```
di, line 161: Is this right? Will the actor always be the repo owner?

warehouse/manage/views.py
```diff
@@ -1047,6 +1049,119 @@ def toggle_2fa_requirement(self):
...
+                    "workflow": form.workflow_name.data,
+                },
+            )
+
+            self.request.session.flash(
+                f"Added {provider} to {self.project.name}",
+                queue="success",
+            )
+
```
di, line 1133: Probably should also send an email here to all maintainers.

warehouse/manage/views.py
```diff
@@ -1047,6 +1049,119 @@ def toggle_2fa_requirement(self):
...
+            self.project.record_event(
+                tag="project:oidc:provider-removed",
+                ip_address=self.request.remote_addr,
+                additional={
+                    "provider": provider.provider_name,
+                    "id": str(provider.id),
+                },
+            )
+
+            self.request.session.flash(f"Removed {provider} from {self.project.name}")
```
di, line 1160: Same comments as above. Let's additionally record the workflow/repo here as well.

di (commented):
Looks good overall. One broad thought: we probably want an `AdminFlag` to easily disable this entire feature at once.

warehouse/templates/manage/publishing.html
woodruffw: Yeah, agreed.  We use the language of "publishing"/"publisher" elsewhere, WDYT about making that consistent? In that context, an OIDC provider becomes something like a "trusted publisher".

warehouse/templates/manage/publishing.html
woodruffw: Yeah, good point. I made it a single form input because `user/repo` slugs are a common pseudo-format, but splitting them up simplifies validation and is just as good of a UX.

warehouse/templates/manage/publishing.html
di: I think "OpenID Connect Publisher" would probably be OK.

warehouse/oidc/models.py
woodruffw: Hmm, nope -- the `actor` can be anybody empowered to trigger the workflow. This should probably become one of our unvalidated claims, unless we want to enforce an allowlist of users here.

warehouse/oidc/forms.py
woodruffw: We have a uniqueness constraint on the DB, and I put the handling for that in the view logic itself (since the view is responsible for creating a model if it doesn't exist, and other views do similar things). But I can move it into the form validation layer, if you think that's more appropriate!

warehouse/manage/views.py
woodruffw: > Additionally, some of the additional metadata seems fairly GitHub specific. We should make sure we either differentiate based on providers or publish provider-specific events.

Yeah, this is a little fiddly with our polymorphic model. So far the only common interface between all children of `OIDCProvider` is `__str__`, but we can add more baseline functionality to make the `additional` metadata more useful (and generic).

di: Probably acceptable to just store `str(provider)` here then, and let the class determine how it should be displayed?

woodruffw: Yep. sounds good. I'll do the same for the deletion route.

warehouse/manage/views.py
woodruffw: Done (via `str(provider)`).

warehouse/oidc/forms.py
woodruffw: d2159312

warehouse/templates/manage/publishing.html
woodruffw: 62a795ff

Screen Shot 2022-02-25 at 6 04 47 PM

warehouse/templates/manage/publishing.html
woodruffw: 5ed70bc4

warehouse/oidc/models.py
woodruffw: 2c9722d4

ewjoachim (commented):
(So far so good for me)

warehouse/manage/views.py
woodruffw: 6c004872

woodruffw:
> we probably want an AdminFlag to easily disable this entire feature at once.

Just to clarify: we want the `AdminFlag` to disable OIDC publisher registration, right? Not the (currently unimplemented) OIDC authentication endpoint?

di:
The entire feature: creation/deletion of publishers, as well as the integration itself eventually.

woodruffw:
> The entire feature: creation/deletion of publishers, as well as the integration itself eventually.

Sounds good! Done with a6aa4a07.

warehouse/oidc/models.py
```diff
@@ -0,0 +1,164 @@
...
+        if not self.__verifiable_claims__:
+            return False
+
+        # All claims should be accounted for.
+        # The presence of an unaccounted claim is not an error, only a warning
+        # that the JWT payload has changed.
+        known_claims = self.__verifiable_claims__.keys().union(
+            self.__preverified_claims__, self.__unchecked_claims__
+        )
+        unaccounted_claims = known_claims.difference(signed_claims.keys())
```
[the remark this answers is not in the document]
woodruffw: e259dc81

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,121 @@
...
+        if not self._api_token:
+            return {}
+        return {"Authorization": f"token {self._api_token}"}
+
+    def validate_owner(self, field):
+        owner = field.data
+
+        # To actually validate the owner, we ask GitHub's API about them.
+        # We can't do this for the repository, since it might be private.
+        response = requests.get(
```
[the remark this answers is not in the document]
woodruffw: d0b37d25 refactors this to use exception-driven handling; let me know if that's what you had in mind!

It doesn't 500 the page since I figure it's form-internal logic and might be transient, and so (IMO) we should just spit back a vaguely descriptive error and tell the user to try again momentarily. But let me know if I've underthought this 🙂

woodruffw:
Here's what the flash looks like when the `AdminFlag` is tripped:

Screen Shot 2022-02-28 at 11 58 42 AM

warehouse/oidc/forms.py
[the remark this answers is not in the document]
ewjoachim: I do agree that it *should not* 500 the whole page, and it would have with the previous implementation, butt the way it's done now is that it will only 500 in case of a network error that is not a timeout, which should probably be enough.

Also, if we have a 403 for any other reason, we might get a confusing message, but I'm not sure I see a scenario where we have a 403, though I don't know GitHub's permission model that well.

I think that should be 👍 for me for a start, and we'll always be able to make it better on the next iteration 😀

warehouse/oidc/forms.py
woodruffw: https://github.com/pypa/warehouse/pull/10753/commits/1fc826a89e296274fcd57d2691b2d5aaba56ac73

warehouse/config.py
```diff
@@ -270,6 +270,18 @@ def configure(settings=None):
...
+        "warehouse.manage.oidc.user_provider_registration_ratelimit_string",
+        "USER_OIDC_PROVIDER_REGISTRATION_RATELIMIT_STRING",
+        default="10 per day",
+    )
+    maybe_set(
+        settings,
+        "warehouse.manage.oidc.ip_provider_registration_ratelimit_string",
+        "IP_OIDC_PROVIDER_REGISTRATION_RATELIMIT_STRING",
+        default="10 per day",
+    )
```
woodruffw, line 284: This might be way too conservative, given CGNAT and other many-to-one-IP networking scenarios. Thoughts?

di: This seems reasonable.

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+                return {**self.default_response, "github_provider_form": form}
+
+            # We only email project owners, since only owners can manage OIDC providers.
+            owner_roles = (
+                self.request.db.query(Role)
+                .filter(Role.project == self.project)
+                .filter(Role.role_name == "Owner")
+                .all()
+            )
+            owner_users = {owner.user for owner in owner_roles}
```
di, line 1207: This is an N+1 query: for every `Role.project` we do a query for that relationship, and for every `owner.user` we do a query to satisfy that relationship.

We should do this with a single query instead:

```suggestion
            owner_roles = (
                request.db.query(User.id)
                .join(Role.user)
                .filter(Role.role_name == "Owner", Role.project == self.project)
                .subquery()
            )
            owner_users =  (
                request.db.query(User)
                .join(owner_roles, User.id == owner_roles.c.id)
                .all()
            )
```

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+                tag="project:oidc:provider-removed",
+                ip_address=self.request.remote_addr,
+                additional={
+                    "provider": provider.provider_name,
+                    "id": str(provider.id),
+                    "specifier": str(provider),
+                },
+            )
+
+            self.request.session.flash(f"Removed {provider} from {self.project.name}")
```
di, line 1267: We should similarly email owners here.

warehouse/templates/manage/settings.html
```diff
@@ -98,6 +98,21 @@ <h3>{% trans %}<abbr title="two factor authentication">2FA</abbr> requirement{%
...
+  </p>
+  <!-- if OIDC is enabled for project -->
+  <p>
+    {% trans href=request.route_path('manage.project.settings.oidc', project_name=project.name), project_name=project.name %}
+    <a href="{{ href }}" class="button button--primary">
+      Manage OpenID Connect publishers for {{ project_name }}
+    </a>
+    {% endtrans %}
+  </p>
+
```
di, line 115: I think this can probably just exist as a new menu option in the sidebar and the "Publishers" page and doesn't need to be here.

di (commented):
Looking good. Let's address these changes and get tests passing.

One additional thing we need to do is put anything user-facing behind a feature flag: anything new in the HTML templates should be gated on the presence of an environment variable, and similarly any new routes should 404 if the envvar is not present. See #10856 for an example.

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class GitHubProviderForm(forms.Form):
+    __params__ = ["owner", "repository", "workflow_name"]
+
+    owner = wtforms.StringField(
+        validators=[
+            wtforms.validators.DataRequired(
+                message=_("Specify GitHub owner (username or organization)")
+            ),
```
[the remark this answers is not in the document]
woodruffw: Yeah, we'll need a separate regex. I'll add one!

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+            return HTTPTooManyRequests(
+                self.request._(
+                    "There have been too many attempted OpenID Connect registrations. "
+                    "Try again later."
+                ),
+                retry_after=exc.resets_in.total_seconds(),
+            )
+
+        form = GitHubProviderForm(
+            self.request.POST,
```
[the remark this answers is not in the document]
woodruffw: Is there a best practice for this documented anywhere (for either Pyramid or Warehouse in particular)? Some of the other views use the bare `self.request.POST`, e.g.:

```python
    @view_config(
        request_method="POST",
        request_param=AddEmailForm.__params__,
        require_reauth=True,
    )
    def add_email(self):
        form = AddEmailForm(
            self.request.POST,
            user_service=self.user_service,
            user_id=self.request.user.id,
        )
        
        # ...
```

I vaguely remember someone telling me *not* to use the splat form because it could conceivably allow the other (trusted) kwargs to be injected, like `user_id` in the example above. But I could be misremembering that.

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+                    repository_name=form.repository.data,
+                    owner=form.normalized_owner,
+                    owner_id=form.owner_id,
+                    workflow_name=form.workflow_name.data,
+                )
+
+                self.request.db.add(provider)
+
+            # Each project has a unique registration of OIDC providers,
+            # even though the relationship is many-many.
```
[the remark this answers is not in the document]
woodruffw: Sorry, the wording there is murky: each OIDC provider can only be registered to a project once, even though each OIDC provider can have many projects and many projects can share the same provider.

In other words, `list(set(Project.oidc_providers)) == Project.oidc_providers`.

I'll reword to clarify that.

warehouse/manage/views.py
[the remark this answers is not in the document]
woodruffw: Yeah, the WTForms documentation supports (and gives priority to) passing the `dict` in directly, rather than splatting it: https://wtforms.readthedocs.io/en/3.0.x/forms/#wtforms.form.Form.__init__

Specifically:

> kwargs – Merged with data to allow passing existing data as parameters. Overwrites any duplicate keys in data. Only used if formdata is not passed.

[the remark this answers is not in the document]
di: Yeah, great point and TIL. Thanks for clarifying.

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+            )
+
+    @property
+    def default_response(self):
+        return {
+            "project": self.project,
+            "github_provider_form": GitHubProviderForm(
+                api_token=self.request.registry.settings.get("github.token"),
+            ),
+        }
```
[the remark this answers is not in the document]
di: Updated this to not spread `self.request.POST`.

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+                    "There have been too many attempted OpenID Connect registrations. "
+                    "Try again later."
+                ),
+                retry_after=exc.resets_in.total_seconds(),
+            )
+
+        form = GitHubProviderForm(
+            self.request.POST,
+            api_token=self.request.registry.settings.get("github.token"),
+        )
```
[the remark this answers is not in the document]
woodruffw: I'm having a little bit of trouble getting this pattern to play nicely with the way we display form errors. I'm not 100% sure yet, but I think the problem is the implicit state after the `validate()` call -- somehow we're clearing it.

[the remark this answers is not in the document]
woodruffw: Figured it out. We have to be careful about how many times we `@property`-construct the form, since each will be "new" (even for the same `POST`) and `validate()` doesn't carry between them.

warehouse/oidc/forms.py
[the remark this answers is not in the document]
woodruffw: Done. Any thoughts on how to make WTForm's validators "fail fast", i.e. skip the GitHub API call if we know that the username is already invalid?

Screen Shot 2022-03-09 at 11 12 57 AM
e

warehouse/manage/views.py
```diff
@@ -1059,6 +1069,206 @@ def toggle_2fa_requirement(self):
...
+            )
+            return self.default_response
+
+        form = DeleteProviderForm(self.request.POST)
+
+        if form.validate():
+            provider = self.request.db.query(OIDCProvider).get(form.provider_id.data)
+            if provider not in self.project.oidc_providers:
+                self.request.session.flash(
+                    f"{provider} is not registered to {self.project.name}",
```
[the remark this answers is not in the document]
woodruffw: I'll add another default handling case for `None`, and fix up the flashes.

warehouse/oidc/forms.py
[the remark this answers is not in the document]
di: We need to find a way to raise a `wtforms.validators.StopValidation` before that validator is evaluated.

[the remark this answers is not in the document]
woodruffw: Looks like there isn't a straightforward way to do a `StopValidation` within the "inline" validator list, so I can promote this regex check to the `validate_owner` callable instead.

warehouse/manage/views.py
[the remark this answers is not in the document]
woodruffw: Reworded; let me know what you think!

warehouse/manage/views.py
```diff
@@ -1059,6 +1070,232 @@ def toggle_2fa_requirement(self):
...
+                    owner=form.normalized_owner,
+                    owner_id=form.owner_id,
+                    workflow_name=form.workflow_name.data,
+                )
+
+                self.request.db.add(provider)
+
+            # Each project has a unique set of OIDC providers; the same
+            # provider can't be registered to the project more than once.
+            if provider in self.project.oidc_providers:
```
di, line 1197: Note that this membership test is only working because these end up being the same instance, I believe. Otherwise, this wouldn't work:

```
>>> from warehouse.oidc.models import GitHubProvider
>>> import uuid
>>> id = uuid.uuid4()
>>> a = GitHubProvider(repository_name='foo', owner='bar', owner_id=id, workflow_name='workflow.yml')
>>> b = GitHubProvider(repository_name='foo', owner='bar', owner_id=id, workflow_name='workflow.yml')
>>> a == b
False
```

warehouse/manage/views.py
```diff
@@ -1059,6 +1070,232 @@ def toggle_2fa_requirement(self):
...
+            )
+            if provider is None:
+                provider = GitHubProvider(
+                    repository_name=form.repository.data,
+                    owner=form.normalized_owner,
+                    owner_id=form.owner_id,
+                    workflow_name=form.workflow_name.data,
+                )
+
+                self.request.db.add(provider)
```
di, line 1193: This has the potential effect of adding a duplicate provider. We need to check if the provider exists before adding it.

woodruffw: I might be misunderstanding what you mean, but isn't that what the `one_or_none()` query above does? We only create the provider if it doesn't already exist.

woodruffw:
Still working on the unit tests, here's what's remaining (for my own recordkeeping):

```
Name                         Stmts   Miss Branch BrPart  Cover   Missing
------------------------------------------------------------------------
warehouse/manage/views.py      687     57    198      0    91%   1087-1088, 1092, 1102-1103, 1106-1114, 1120, 1127, 1134-1143, 1147-1238, 1242-1296
warehouse/oidc/models.py        56     16     14      0    66%   79-102, 107, 151, 155, 159, 163, 166
warehouse/oidc/services.py      89      2     26      0    98%   160-161
------------------------------------------------------------------------
TOTAL                         9399     75   2637      0    99%
```

woodruffw:
:tada:

```
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
---------------------------------------------------
TOTAL    9422      0   2647      0   100%
```

warehouse/manage/views.py
woodruffw: Whoops, didn't mean to resolve.

woodruffw:
[cc0c21a](https://github.com/pypa/warehouse/pull/10753/commits/cc0c21a2cae85dac6afd51985d87731803f076c6) adds a site-wide environment flag for controlling the availability of the OIDC routes.

di (commented):
Seems like I had comments in a pending review that didn't get submitted, apologies if these are out of date.

warehouse/manage/views.py
```diff
@@ -1157,6 +1162,9 @@ def manage_project_oidc_providers(self):
 
     @view_config(request_method="POST", request_param=GitHubProviderForm.__params__)
     def add_github_oidc_provider(self):
+        if not self.oidc_enabled:
+            return self.default_response
```
di, line 1166: I think this should 404 instead.

warehouse/manage/views.py
```diff
@@ -1242,6 +1250,9 @@ def add_github_oidc_provider(self):
 
     @view_config(request_method="POST", request_param=DeleteProviderForm.__params__)
     def delete_oidc_provider(self):
+        if not self.oidc_enabled:
+            return self.default_response
```
di, line 1254: Same here

warehouse/manage/views.py
```diff
@@ -1138,12 +1139,16 @@ def github_provider_form(self):
...
+            "oidc_enabled": self.oidc_enabled,
             "project": self.project,
             "github_provider_form": self.github_provider_form,
         }
 
     @view_config(request_method="GET")
     def manage_project_oidc_providers(self):
+        # NOTE: We don't need a special "OIDC enabled" check here, since
+        # this view handler has no special logic.
+
```
di, line 1151: This should 404 if the feature flag is present.

warehouse/templates/includes/manage/manage-project-menu.html
```diff
@@ -45,6 +45,12 @@
       </a>
     </li>
     {% endif %}
+    <li>
+      <a href="{{ request.route_path('manage.project.settings.publishing', project_name=project.normalized_name) }}" class="vertical-tabs__tab vertical-tabs__tab--with-icon {% if active_tab == 'publishing' %}vertical-tabs__tab--is-active{% endif %} {% if mode == 'mobile' %}vertical-tabs__tab--mobile{% endif %}" {% if active_tab == 'publishing' %}aria-selected="true"{% endif %}>
+        <i class="fa fa-upload" aria-hidden="true"></i>
+        {% trans %}Publishing{% endtrans %}
+      </a>
+    </li>
```
di, line 53: This should be behind the feature flag.

warehouse/manage/views.py
```diff
@@ -1059,6 +1070,232 @@ def toggle_2fa_requirement(self):
...
+                self.request.db.query(User)
+                .join(owner_roles, User.id == owner_roles.c.id)
+                .all()
+            )
+            for user in owner_users:
+                send_oidc_provider_removed_email(
+                    self.request, user, project_name=self.project.name
+                )
+
+            self.project.oidc_providers.remove(provider)
```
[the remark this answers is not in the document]
woodruffw: Yeah, I wasn't sure whether it made sense to delete it (or what the deletion semantics should be, since it's a many-many mapping).

I'll add a note.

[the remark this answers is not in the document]
woodruffw: [c2ca980](https://github.com/pypa/warehouse/pull/10753/commits/c2ca980fc4ae3551631d5434212b4b7a0d2f1085)

warehouse/manage/views.py
woodruffw: [c2ca980](https://github.com/pypa/warehouse/pull/10753/commits/c2ca980fc4ae3551631d5434212b4b7a0d2f1085)

warehouse/manage/views.py
woodruffw: [c2ca980](https://github.com/pypa/warehouse/pull/10753/commits/c2ca980fc4ae3551631d5434212b4b7a0d2f1085)

warehouse/templates/includes/manage/manage-project-menu.html
woodruffw: [c2ca980](https://github.com/pypa/warehouse/pull/10753/commits/c2ca980fc4ae3551631d5434212b4b7a0d2f1085)

di (commented):
edit

di (approved):
This LGTM, and I think this is ready for @ewdurbin to make an additional/final review (also @ewjoachim if you're interested as well!)

ewdurbin (changes requested):
I'll have time for a more thorough review later, but right off the bat I see that we should instrument many of these views with metrics.

woodruffw:
> I'll have time for a more thorough review later, but right off the bat I see that we should instrument many of these views with metrics.

Sounds good. Are you thinking that the metrics should be in conjunction with the current ratelimit logic, e.g. counters whenever a ratelimit is exceeded? Or more in-depth than that?

warehouse/oidc/services.py
```diff
@@ -148,10 +148,51 @@ def get_key(self, key_id):
...
+            )
+            return True, valid_token
+        except jwt.PyJWTError:
+            return False, None
+        except Exception as e:
+            # We expect pyjwt to only raise subclasses of PyJWTError, but
+            # we can't enforce this. Other exceptions indicate an abstraction
+            # leak, so we log them for upstream reporting.
+            sentry_sdk.capture_message(f"JWT verify raised generic error: {e}")
+            return False, None
```
ewjoachim, line 195: The first return value seems to be `bool(second)`, I'm not sure why we need 2 return values. Is there a reason why we're not returning a token or raising an exception if invalid ?

warehouse/templates/manage/publishing.html
```diff
@@ -0,0 +1,156 @@
...
+      <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+      {{ form_errors(github_provider_form) }}
+      <div class="form-group">
+        <label for="owner" class="form-group__label">
+          {% trans %}Owner{% endtrans %}
+          {% if github_provider_form.owner.flags.required %}
+          <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+          {% endif %}
+        </label>
+        {{ github_provider_form.owner(placeholder=gettext("owner"), autocomplete="off", autocapitalize="off", spellcheck="false", class_="form-group__field", **{"aria-describedby":"owner-errors"}) }}
```
ewjoachim, line 89: We could leave it this way because we do the same elsewhere, but it wouldn't hurt to use the simple form
```suggestion
        {{ github_provider_form.owner(placeholder=gettext("owner"), autocomplete="off", autocapitalize="off", spellcheck="false", class_="form-group__field", aria_describedby="owner-errors") }}
```
(see #10003 / https://wtforms.readthedocs.io/en/2.3.x/widgets/#wtforms.widgets.html_params)

ewjoachim (commented):
A few comments here and there, really good job ! (I haven't paid much attention to the tests)

warehouse/manage/views.py
```diff
@@ -1059,6 +1081,230 @@ def toggle_2fa_requirement(self):
...
+                    "Try again later."
+                ),
+                retry_after=exc.resets_in.total_seconds(),
+            )
+
+        response = self.default_response
+        form = response["github_provider_form"]
+
+        if form.validate():
+            self._hit_ratelimits()
```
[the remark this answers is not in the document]
woodruffw: Yep, this looks like a typo on my part. Thanks for catching it!

warehouse/manage/views.py
```diff
@@ -1059,6 +1081,230 @@ def toggle_2fa_requirement(self):
...
+
+        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+            self.request.session.flash(
+                (
+                    "OpenID Connect is temporarily disabled. "
+                    "See https://pypi.org/help#admin-intervention for details."
+                ),
+                queue="error",
+            )
+            return self.default_response
```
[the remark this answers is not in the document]
woodruffw: That seems reasonable to me. I copied this style from similar views, none of which currently use a decorator to handle admin flags. Perhaps we could mark this for more general future refactoring and do it all in one fell blow 🙂?

warehouse/oidc/models.py
```diff
@@ -0,0 +1,185 @@
...
+
+    id = Column(UUID(as_uuid=True), ForeignKey(OIDCProvider.id), primary_key=True)
+    repository_name = Column(String)
+    owner = Column(String)
+    owner_id = Column(String)
+    workflow_name = Column(String)
+
+    __verifiable_claims__ = {
+        "repository": str.__eq__,
+        "job_workflow_ref": str.startswith,
```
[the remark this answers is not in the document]
woodruffw: This is the full structure of each `job_workflow_ref`:

```
woodruffw/gha-oidc-jwt-test-repo/.github/workflows/release.yml@refs/tags/v0.0.3
```

The `username/repo` prefix and `.github/workflows` component are what makes the `startswith` check safe -- a malicious user can't contrive a workflow name such that those would pass, since workflows have to live inside a flat directory (`.github/workflows`).

I can definitely add some additional documentation here to explain the structure we're matching here!

warehouse/oidc/services.py
woodruffw: Yeah, this can just be one return parameter. Will refactor!

warehouse/oidc/models.py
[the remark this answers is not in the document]
ewjoachim: Hm, but then if matching the workflow name itself don't contribute to security, why do we require the name of the workflow at all ? I'm not asking that we change it all at this stage, but rather making sure I understand the security model.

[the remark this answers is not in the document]
woodruffw: Sorry if I'm explaining poorly -- the workflow name *does* contribute to security. Here's the rough model:

*Identity*: The identity of the trusted repository is defined by `owner/repo` + `owner_id`, where `owner` is the GitHub owner identity (username/org name), `repo` is the repository name, and `owner_id` is the unique and never reused GitHub ID for `owner`.

*Authenticity*: The authenticity of a token is determined by its validity against GitHub's published JWK set. 

*Authority*: Only certain workflows are trusted for the purpose of minting access tokens from JWTs. We don't trust every workflow, because some workflows might be configured as runnable by untrusted users (e.g., on every commit). The workflow we trust for each provider setup is defined by `workflow.yml`, which is the unique name of the workflow specification in the repository.

Then, in terms of claim verification:

* Identity: `repository` ensures that we're validating the correct token.
* Authenticity: This is provided by the JWT payload verification iteself.
* Authority: `workflow` and `job_workflow_ref` ensure that we're only accepting tokens from our trusted workflow.

To summarize: `workflow` and `job_workflow_ref` are the two claims that ensure that we don't accept JWTs that are valid but aren't trusted for our purposes. They're technically slightly redundant against each other, but having both makes us a little more resilient against potential internal changes to GitHub (e.g., if GitHub one day allows workflows to override their name internally with a `workflow:` key, `job_workflow_ref` will still be safe because it contains the fully qualified repository path).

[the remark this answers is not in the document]
di: I think what @ewjoachim is trying to say is that if the project has been configured to trust `user/repo/.github/workflows/foo.yml`, using `startswith` will also implicitly trust `user/repo/.github/workflows/foo.yml@malicious.yml`:

```
>>> expected_job_workflow_ref = "user/repo/.github/workflows/foo.yml@"
>>> claimed_job_workflow_ref = "user/repo/.github/workflows/foo.yml@malicious.yml"
>>> claimed_job_workflow_ref.startswith(expected_job_workflow_ref)
True
```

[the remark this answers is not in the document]
ewjoachim: (That 👆)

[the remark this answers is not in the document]
ewjoachim: Hm reading from https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows it looks like the repository and the workflow ref might not match, so if we do it this way, it's possible that it won't be possible to use reusable workflows, no ?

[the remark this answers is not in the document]
di: I believe this is correct, and that we should be able to support reusable workflows as well, so we should probably revisit this to allow a user to specify a workflow at a ref outside the repo.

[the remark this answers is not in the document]
woodruffw: Sorry, was out for most of last week. I'll catch up here today!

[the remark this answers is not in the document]
woodruffw: Re: `workflow` and `job_workflow` ref: I think we should always require *at least* one, since the absence of both makes it much easier for someone to trick a repository owner into merging changes to *any* workflow and getting direct access to the PyPI project.

That being said, I think you're both right about the two not necessarily matching. So yes, we should probably allow users to configure one or the other, but always at least one (and for `job_workflow_ref`, support globs). Does that seem reasonable?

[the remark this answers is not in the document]
woodruffw: More re: `job_workflow_ref`: we should probably be careful about which positions we support globs in, since we don't want someone to accidentally configure `*/*` or even `foo/bar/.github/workflows/*.yml` (since that'll be the same insecure configuration as allowing any `workflow`).

ewdurbin (commented):
Likely my bad for never responding to https://github.com/pypa/warehouse/pull/10753#issuecomment-1071331748, but I'm still bummed to not see metrics.

Things I would expect to emit a metric:
  - Rate limiters (see account login flow rate limiters for examples)
  - Provisioning/Deprovisioning of providers
  - Attempted/Failed verifications
  - Attempted/Failed claims
  - Token issuance/verification

woodruffw:
> Rate limiters (see account login flow rate limiters for examples)

These are currently in place -- we're ratelimiting both users and IPs w/r/t publisher registrations, which are (so far) the only "abusable" requests that users can make (since they hit GitHub's API). We're also ratelimiting JWKs refetches, although that code isn't currently reachable with this changeset (since this PR doesn't include the authentication flows, only the core models and setup work).

I can add metrics for the rest now! Well, almost all of them -- we can't do token issuance/verification yet, since that's not part of this PR. But I'll mark that in the corresponding feature issue (#10970).

warehouse/manage/views.py
```diff
@@ -1165,6 +1167,8 @@ def add_github_oidc_provider(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
+        self.metrics.increment("warehouse.oidc.add_github_provider.attempt")
```
ewdurbin, line 1170:
```suggestion
        self.metrics.increment("warehouse.oidc.add_provider.attempt", tags=["provider:github"])
```

woodruffw:
@ewdurbin let me know what you think of the metrics in [5dafd98](https://github.com/pypa/warehouse/pull/10753/commits/5dafd988d994d3477246ab289db3e088ffc5532f) and [0ff3f01](https://github.com/pypa/warehouse/pull/10753/commits/0ff3f012ae4a681fd3f0e58add66187b10f13d31) -- those add counters for the relevant routes and JWT verification routines.

warehouse/manage/views.py
ewdurbin: and similar for others.

warehouse/manage/views.py
```diff
@@ -1247,13 +1252,17 @@ def add_github_oidc_provider(self):
...
+            self.metrics.increment("warehouse.oidc.add_github_provider.ok")
+
         return response
 
     @view_config(request_method="POST", request_param=DeleteProviderForm.__params__)
     def delete_oidc_provider(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
+        self.metrics.increment("warehouse.oidc.delete_provider.attempt")
```
ewdurbin, line 1264:
```suggestion
        self.metrics.increment("warehouse.oidc.delete_provider.attempt", tags=["provider:github"])
```

and similar for others.

warehouse/manage/views.py
woodruffw: Good call -- there's a few of these, so I'll roll them up.

warehouse/oidc/services.py
```diff
@@ -161,6 +161,7 @@ def _get_key_for_token(self, token):
         return self.get_key(unverified_header["kid"])
 
     def verify_signature_only(self, token):
+        self.metrics.increment("warehouse.oidc.verify_signature_only.attempt")
```
ewdurbin, line 164: hmmm, any chance we know the provider at this point? would be good to tag so we know what providers are being used.

warehouse/manage/views.py
woodruffw: `delete_provider` is generic over multiple provider types, so I'll figure out the best way to avoid hardcoding `github` here.

ewdurbin (approved):
once metric tagging is addressed, all good!

warehouse/oidc/services.py
woodruffw: Hmm, probably not directly -- we only test a JWT against a project's provider(s) once we're passed the signature verification, so plumbing the providers into this method would be inverting the dependency relationship.

The top level verification action is against a project, so perhaps a tag for the project name in `verify_for_project` is sufficient? WDYT?

warehouse/manage/views.py
woodruffw: c945ee7d

warehouse/manage/views.py
woodruffw: c945ee7d

warehouse/oidc/services.py
woodruffw: Thinking about this more: `verify_signature_only` should only ever be called from `verify_for_project`, so I can just remove these metrics and making the calling context more detailed.

woodruffw: 32cfd38bb43493cba21b23c133fe2617035baa40 and 63bce661e202f53ef704b21f353ebdb0d7e47689 address this.

warehouse/oidc/models.py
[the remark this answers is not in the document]
di: Let's not worry about supporting reusable workflows for now and assume the workflow lives in the repo (and file an issue capturing that deficiency).

I still think we need to resolve the use of `startswtih` here though.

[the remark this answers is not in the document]
woodruffw: Sounds good. I'll mark `job_workflow_ref` as an explicitly unverified claim for now and document the plans there.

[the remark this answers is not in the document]
woodruffw: Tracking here: https://github.com/pypa/warehouse/issues/11096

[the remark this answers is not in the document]
woodruffw: 04c72611

woodruffw:
Disabled `job_workflow_ref` for now. Should be good to go!

ewjoachim (approved):

ewjoachim:
Great job :)

warehouse/templates/email/oidc-provider-removed/subject.txt
```diff
@@ -0,0 +1,21 @@
...
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/subject.txt" %}
+
+{% block subject %}
+{% trans project_name=project_name %}
+OpenID Connect removed from {{ project_name }}
```
di, line 19:
```suggestion
OpenID Connect publisher removed from {{ project_name }}
```

warehouse/templates/manage/publishing.html
```diff
@@ -0,0 +1,156 @@
...
+    {% else %}
+    <p class="no-bottom-padding">{% trans %}No publishers are currently configured.{% endtrans %}</p>
+    {% endif %}
+    {% else %}
+    <p>
+      {% trans %}
+      OpenID Connect publishing isn't enabled yet! Come back soon.
+      {% endtrans %}
+    </p>
+    {% endif %}
```
di, line 151: This shouldn't be necessary, this route should just 404 if the feature hasn't been enabled yet.

warehouse/oidc/forms.py
```diff
@@ -0,0 +1,144 @@
...
+        owner_info = self._lookup_owner(owner)
+
+        # NOTE: Use the normalized owner name as provided by GitHub.
+        self.normalized_owner = owner_info["login"]
+        self.owner_id = owner_info["id"]
+
+    def validate_workflow_name(self, field):
+        workflow_name = field.data
+
+        if not (workflow_name.endswith(".yml") or workflow_name.endswith(".yaml")):
```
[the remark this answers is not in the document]
woodruffw: Possibly: [GitHub's docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows) say that the suffix has to be exactly `.yml` or `.yaml`, but it's possible that that's an underspecification as well. I can try some other variants with our dummy repo in a bit.

warehouse/templates/manage/publishing.html
woodruffw: Thanks; will delete.

warehouse/manage/views.py
```diff
@@ -1059,6 +1082,250 @@ def toggle_2fa_requirement(self):
...
+                    queue="error",
+                )
+                return self.default_response
+
+            # We only email project owners, since only owners can manage OIDC providers.
+            owner_users = project_owners(self.request, self.project)
+            for user in owner_users:
+                send_oidc_provider_removed_email(
+                    self.request, user, project_name=self.project.name
+                )
```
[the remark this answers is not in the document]
woodruffw: 73eef397

warehouse/email/__init__.py
```diff
@@ -466,6 +466,18 @@ def send_recovery_code_reminder_email(request, user):
...
+@_email("oidc-provider-added")
+def send_oidc_provider_added_email(request, user, project_name):
+    # We use the request's user, since they're the one triggering the action.
+    return {"username": request.user.username, "project_name": project_name}
+
+
+@_email("oidc-provider-removed")
+def send_oidc_provider_removed_email(request, user, project_name):
+    # We use the request's user, since they're the one triggering the action.
+    return {"username": request.user.username, "project_name": project_name}
```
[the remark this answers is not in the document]
woodruffw: This might be difficult to do in a polymorphic manner, since only `GitHubProvider` has those fields. Something like this should work, though:

* Publisher name: GitHub (via `OIDCProvider.provider_name`
* Publisher specification: foo.yml @ foo/bar (via `str(provider)`)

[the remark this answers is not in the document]
woodruffw: 4651ce87

warehouse/templates/manage/publishing.html
woodruffw: Done.

warehouse/oidc/forms.py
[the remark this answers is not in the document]
woodruffw: Looks like uppercase workflows aren't run: https://github.com/woodruffw/gha-oidc-jwt-test-repo/pull/5, so I think it's safe to treat this as case sensitive for now.

di (approved):

warehouse/config.py
```diff
@@ -270,6 +270,18 @@ def configure(settings=None):
...
+        "warehouse.manage.oidc.user_registration_ratelimit_string",
+        "USER_OIDC_REGISTRATION_RATELIMIT_STRING",
+        default="10 per day",
+    )
+    maybe_set(
+        settings,
+        "warehouse.manage.oidc.ip_registration_ratelimit_string",
+        "IP_OIDC_REGISTRATION_RATELIMIT_STRING",
+        default="10 per day",
+    )
```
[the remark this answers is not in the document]
mohmedsalem2026-maker: - [x] - - - - - - @**____[]()____**

PR #10766, opened 2022-02-16
Sync PSF sponsors with logo placement API

berinhard:
This PR adds a new scheduled task to keep Pypi's database in sync with Python.org logo placements API. It's scheduled to run every sunday at 03:00 AM. The goal of this work is to centralize sponsors management at pythondotorg and, thus, reduce any type of duplicated work on Pypi side. It also opens the possibility to replace our logo placement by integration with EthicalAds, the same way [we did on pythondotorg](https://github.com/python/pythondotorg/pull/1981).

This PR does not handle sidebar, footer, one time or infra structure sponsors. It only operates on top of PSF ones.

I ask reviewers to leave instructions on the new unit tests. I feel that the tests are complex due to `pretend` and also I feel that probably there's a clever way to mock the task request object.

cc/ @ewdurbin


warehouse/sponsors/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+        sponsor.link_url = sponsor_info["sponsor_url"]
+        sponsor.color_logo_url = sponsor_info["logo"]
+        sponsor.level_name = sponsor_info["level_name"]
+        sponsor.level_order = sponsor_info["level_order"]
+        sponsor.is_active = True
+        sponsor.psf_sponsor = True
+        sponsor.origin = "remote"
+
+        if not sponsor.id:
+            request.db.add(sponsor)
```
di, line 68: I think this could jsut get moved up under the `sponsor = Sponsor()` above.

warehouse/sponsors/__init__.py
```diff
@@ -22,3 +24,8 @@ def _sponsors(request):
 def includeme(config):
     # Add a request method which will allow to list sponsors
     config.add_request_method(_sponsors, name="sponsors", reify=True)
+
+    # Add a periodic task to update sponsors table
+    config.add_periodic_task(
+        crontab(minute=0, hour=3, day_of_week=0), update_pypi_sponsors
```
di, line 30: FYI, this is 3AM UTC. Not sure if that was the original intention (also not sure if it really matters!)

di (approved):
LGTM! Just some small comments

> I ask reviewers to leave instructions on the new unit tests. I feel that the tests are complex due to pretend and also I feel that probably there's a clever way to mock the task request object.

Tests look good to me, I think your request fixture is fine.

warehouse/sponsors/__init__.py
```diff
@@ -22,3 +24,8 @@ def _sponsors(request):
 def includeme(config):
     # Add a request method which will allow to list sponsors
     config.add_request_method(_sponsors, name="sponsors", reify=True)
+
+    # Add a periodic task to update sponsors table
+    config.add_periodic_task(
```
ewdurbin, line 29: We might want to wrap this in a conditional to check if `pythondotorg.api_token` is set, then "unconfigure" it in dev. Mostly since I'm going to suggest we run this hourly to reduce the delay for new sponsors and confusion from PSF staff.

warehouse/sponsors/__init__.py
ewdurbin, line 30:
```suggestion
        crontab(minute=10), update_pypi_sponsors
```

warehouse/sponsors/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+@tasks.task(ignore_result=True, acks_late=True)
+def update_pypi_sponsors(request):
+    """
+    Read data from pythondotorg's logo placement API and update Sponsors
+    table to mirror it.
+    """
+    host = request.registry.settings["pythondotorg.host"]
+    token = request.registry.settings["pythondotorg.api_token"]
+    headers = {"Authorization": f"Token {token}"}
+
```
ewdurbin, line 32: As an alternate to https://github.com/pypa/warehouse/pull/10766/files#r808505811 we could short circuit here if host and token are not both set.

di: Or maybe we could just prevent the task from being registered at all?

ewdurbin: yeah, that comment is what's linked :-D I'll resolve this one and let that suggestion stand alone

ewdurbin (approved):

PR #10776, opened 2022-02-18
Suggest twine-check

sgillies:
To diagnose description render failures.

This is my first contribution to the help page. I hope I understand the structure correctly.


ewdurbin:
We should be linking to the `#description_content_type` anchor when [errors occur rendering](https://github.com/pypa/warehouse/blob/a1335776acf11911232771ad1dbcd6599590449c/warehouse/forklift/legacy.py#L1055-L1072), so perhaps we should add this sentence to that help text.

sgillies:
@ewdurbin thanks for the suggestion, I'll add more detail. Does the help page get rendered to RTD or a staging site where I can see the changes? Or do I need to run a local warehouse to observe the results?

di:
I think this is due to `psyscopg2` requiring platform-level dependencies to be installed in advance before it can be installed (see https://www.psycopg.org/docs/install.html). What platform are you on?

sgillies:
@di I'm on Ubuntu 20.04 and was lacking libpq-dev. After installing that, a new `make build` completed. I'll add a note to the getting started doc about this. Or maybe we could find a way to install libpq-dev only in the image?

di:
I think that's a good short-term fix until we can fix https://github.com/pypa/warehouse/issues/4948.

sgillies:
I've got a local warehouse running and have some screenshots of the changes in my patch-1 branch.

Here is the "Troubleshooting" tables of contents. Does the warehouse project have a convention for ordering of FAQ items? I've recently advocated for putting newest items on top and periodically checking the ones at the bottom for relevance and correctness, and have done that here. I'm happy to move it, though.

troubleshooting-toc

Here is the head of the troubleshooting section:

troubleshooting-section

And here is the "different format" item:

different-format

As noted in f1b66c5, I moved some content from this section and left a pointer. Does that look okay?

warehouse/templates/pages/help.html
```diff
@@ -625,6 +626,13 @@ <h3 id="vulnerability-data">{{ vulnerability_data() }}</h3>
       <section class="faq-group" id="troubleshooting">
         <h2>{% trans %}Troubleshooting{% endtrans %}</h2>
 
+        <h3 id="description-render-failure">{{ description_render_failure() }}</h3>
+        <p>
+          {% trans trimmed readme_renderer_href='https://pypi.org/project/readme_renderer/', twine_check_href='https://twine.readthedocs.io/en/latest/#twine-check', title=gettext('External link') %}
+            PyPI will reject uploads if the package description fails to render. You may <a href="{{ twine_check_href }}" title="{{ title }}">use twine's check command</a> or <a href="{{ readme_renderer_href }}">readme_renderer</a>, the same description renderer used by PyPI, to locally check a description for validity.
```
di, line 632:
```suggestion
          {% trans trimmed twine_check_href='https://twine.readthedocs.io/en/latest/#twine-check', title=gettext('External link') %}
            PyPI will reject uploads if the package description fails to render. You may <a href="{{ twine_check_href }}" title="{{ title }}">use twine's check command</a> to locally check a description for validity.
```

I think `readme_renderer` is too low-level here, and `twine check` gives the same functionality, so let's just recommend that.

di:
> I've recently advocated for putting newest items on top and periodically checking the ones at the bottom for relevance and correctness, and have done that here.

There's no ordering, I like this though!

sgillies:
Okay, I believe this is ready for review. Thanks for the `twine check` tip on Twitter, last week @di!

The note I added can be seen at the end of the https://warehouse--10776.org.readthedocs.build/development/getting-started.html#building-the-warehouse-container section.

docs/development/getting-started.rst
```diff
@@ -148,6 +148,11 @@ run all of the needed services. The Warehouse repository will be mounted inside
 the Docker container at :file:`/opt/warehouse/src/`. After the initial build,
 you should not have to run this command again.
 
+.. note::
+
+   On some systems ``make build`` may require the prior installation of libpq
+   and its headers so that psycopg2 can be compiled and installed.
+
```
di, line 155: This seems unrelated?

sgillies: No, `make build` failed quickly, while installing requirements, and couldn't succeed until I installed libpq-dev on my laptop.

di: Got it, sorry, didn't realize that you were planning to include this in this PR as well.

PR #10801, opened 2022-02-21
tests: omit covered files

miketheman:
When the tests complete, we get a lovely list of files that are 100%
covered by tests - woohoo!

As the file count grows, seeing anything that isn't 100% covered is
harder, so we can omit any file that coverage.py reports is fully
covered.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -24,3 +24,6 @@ omit =
 exclude_lines =
     pragma: no cover
     class \w+\(Interface\):
+
+# Don't show us anything that's already 100% covered.
+skip_covered = True


miketheman:
See example output [from this run](https://github.com/pypa/warehouse/runs/5280670677?check_suite_focus=true#step:11:288).

```console
+ python -m coverage html
Wrote HTML report to htmlcov/index.html
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
---------------------------------------------------
TOTAL    9104      0   2551      0   100%

185 files skipped due to complete coverage.
```

di:
I'm generally in favor of this for times when we run the whole test suite, but sometimes I run a single test file that should exercise a specific file, and repeatedly run it as I improve the tests until it reaches 100%. Would this have the effect of hiding that 100% line as well?

miketheman:
I believe with this change, the results will hide the now-covered line.

With that workflow, don't you already get a coverage failure from this line?
https://github.com/pypa/warehouse/blob/479e6f3ca2e11348e70e197dc5b4bf7887411341/bin/tests#L38

I tried that recently, and got the full output and failing exit code.

miketheman:
This is all assuming the end user is using the `make tests` command, which is the presented ergonomics. If that's something we'd want to change, then let's talk about that. 😄

di:
Yeah, I'm using `make tests T=<filename>`, just not sure which take precedence, `skip_covered` or `--fail-under`.

miketheman:
I think there's no precedence here - `--fail-under` will still fail when the code-under-test, as specified by:
https://github.com/pypa/warehouse/blob/479e6f3ca2e11348e70e197dc5b4bf7887411341/.coveragerc#L4-L5

doesn't reach 100%

`skip_covered` will only remove the covered lines from the report.

When running `make tests T=<filename>`, do you still get a 0 exit code when only running a single test file?
For fun, without this change, I ran:

```console
$ make tests T=tests/unit/utils/test_paginate.py
...
TOTAL                                                     9104   7630   2551     12    15%
Coverage failure: total of 15 is less than fail-under=100
make: *** [Makefile:87: tests] Error 2
```

The reporting config setting doesn't change the `--fail-under` behavior.

di:
I think I can probably get used to looking for the file I'm trying to test to disappear from the list, rather than seeing it reported as 100%.

Thanks!

PR #10469, opened 2021-12-11
Fix broken development setup

abitrolly:
This removes circular dependency in Makefile added in #10278

Unblocks #10052
Fixes #10446
Fixes #10447

PR #10803, opened 2022-02-22
Run all dev actions in containers

### LINKED ISSUE #4948: `make reformat` fails (I'm using Python 3.7)

### LINKED ISSUE #10447: Circular dependency in Makefile
From https://github.com/pypa/warehouse/pull/10278#issuecomment-974902013: 

> I believe this change has introduced a circular dependency, as perceived by these messages:
> 
> ```shellsession
> $ make -v | head -1
> GNU Make 3.81
> $ make build
> make[1]: Circular requirements/docs.txt <- .state/env/pyvenv.cfg dependency dropped.
> ...
> ```
> 
> For funsies, I tried the same with [remake](http://bashdb.sourceforge.net/remake/), which uses a newer version of `make`:
> 
> ```shellsession
> $ remake --version | head -1
> GNU Make 4.3+dbg-1.5
> $ remake build
> remake[1]: Entering directory '/Users/miketheman/workspace/miketheman/warehouse'
> remake[1]: Circular requirements/docs.txt <- .state/env/pyvenv.cfg dependency dropped.
> remake[1]: Circular requirements/lint.txt <- .state/env/pyvenv.cfg dependency dropped.
> remake[1]: '.state/docker-build' is up to date.
> ...
> ```
> 
> I don't know if this is harmful yet, but I'm having other issues getting started with development, and wondered if this is somehow related or not.

_Originally posted by @miketheman in https://github.com/pypa/warehouse/issues/10278#issuecomment-974902013_

di:
Fixes #4948, fixes #10447. Closes #10469.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,15 +84,6 @@ requirements/%.txt: requirements/%.in
 	  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
 	  bin/pip-compile --allow-unsafe --generate-hashes --output-file=$@ $<
 
-github-actions-deps:
-ifneq ($(GITHUB_BASE_REF), false)
-	git fetch origin $(GITHUB_BASE_REF):refs/remotes/origin/$(GITHUB_BASE_REF)
-	# Check that the following diff will exit with 0 or 1
-	git diff --name-only FETCH_HEAD || test $? -le 1 || exit 1
-	# Make the dependencies if any changed files are requirements files, otherwise exit
-	git diff --name-only FETCH_HEAD | grep '^requirements/' || exit 0 && bin/deps
-endif
-
 initdb:
 	docker-compose run --rm web psql -h db -d postgres -U postgres -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname ='warehouse';"
 	docker-compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS warehouse"
--- a/Makefile
+++ b/Makefile
@@ -165,40 +165,11 @@ purge: stop clean
 stop:
 	docker-compose down -v
 
-compile-pot: .state/env/pyvenv.cfg
-	PYTHONPATH=$(PWD) $(BINDIR)/pybabel extract \
-		-F babel.cfg \
-		--omit-header \
-		--output="warehouse/locale/messages.pot" \
-		warehouse
+translations: .state/docker-build
+	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
+								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
+								  bin/translations
 
-init-po: .state/env/pyvenv.cfg
-	$(BINDIR)/pybabel init \
-		--input-file="warehouse/locale/messages.pot" \
-		--output-dir="warehouse/locale/" \
-		--locale="$(L)"
 
-update-po: .state/env/pyvenv.cfg
-	$(BINDIR)/pybabel update \
-		--input-file="warehouse/locale/messages.pot" \
-		--output-file="warehouse/locale/$(L)/LC_MESSAGES/messages.po" \
-		--locale="$(L)"
-
-compile-po: .state/env/pyvenv.cfg
-	$(BINDIR)/pybabel compile \
-		--input-file="warehouse/locale/$(L)/LC_MESSAGES/messages.po" \
-		--directory="warehouse/locale/" \
-		--locale="$(L)"
-
-build-mos: compile-pot
-	for LOCALE in $(LOCALES) ; do \
-		L=$$LOCALE $(MAKE) compile-po ; \
-		done
-
-translations: compile-pot
-ifneq ($(GITHUB_ACTIONS), false)
-	git diff --quiet ./warehouse/locale/messages.pot || (echo "There are outstanding translations, run 'make translations' and commit the changes."; exit 1)
-else
-endif
 
 .PHONY: default build serve initdb shell tests docs deps clean purge debug stop compile-pot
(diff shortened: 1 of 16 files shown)


bin/docs
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+# Click requires us to ensure we have a well configured environment to run
+# our click commands. So we'll set our environment to ensure our locale is
+# correct.
+export LC_ALL="${ENCODING:-en_US.UTF-8}"
+export LANG="${ENCODING:-en_US.UTF-8}"
+
+COMMAND_ARGS=$@
```
ewjoachim, line 10: we're not using this, or did I miss something ? (also, even if this isn't actually relevant, usually, a good rule of thumb is that`"$@"` is most often what we mean when we use `$@`. Missing the double quotes is a bear trap)

warehouse/locale/Makefile
```diff
@@ -0,0 +1,46 @@
...
+	cd ../../; \
+	pybabel compile \
+		--input-file="warehouse/locale/$(L)/LC_MESSAGES/messages.po" \
+		--directory="warehouse/locale/" \
+		--locale="$(L)"
+
+build-mos: compile-pot
+	cd ../../; \
+	$(eval LOCALES := $(shell python -c "from warehouse.i18n import KNOWN_LOCALES; print(' '.join(set(KNOWN_LOCALES)-{'en'}))"))
+	for LOCALE in $(LOCALES) ; do \
```
ewjoachim, line 35: I'm somehow surprised that Make sees those 3 lines as a single block ? or is it that complil-po doensn'y actually need to be launched from 2 levels above ?

Makefile
```diff
@@ -116,17 +91,10 @@ docs: .state/docker-build
...
-	$(BINDIR)/pip-compile --upgrade --allow-unsafe -o $(TMPDIR)/main.txt requirements/main.in > /dev/null
-	$(BINDIR)/pip-compile --upgrade --allow-unsafe -o $(TMPDIR)/lint.txt requirements/lint.in > /dev/null
-	echo "$$DEPCHECKER" | $(BINDIR)/python - $(TMPDIR)/deploy.txt requirements/deploy.txt
-	echo "$$DEPCHECKER" | $(BINDIR)/python - $(TMPDIR)/main.txt requirements/main.txt
-	echo "$$DEPCHECKER" | $(BINDIR)/python - $(TMPDIR)/lint.txt requirements/lint.txt
-	rm -r $(TMPDIR)
-	$(BINDIR)/pip check
+deps: .state/docker-build
+	docker-compose run --rm web env -i ENCODING="C.UTF-8" \
+								  PATH="/opt/warehouse/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" \
```
ewjoachim: Why can't we define those envvars in the Dockerfile ?

Makefile
```diff
@@ -137,7 +105,7 @@ ifneq ($(GITHUB_BASE_REF), false)
 	# Check that the following diff will exit with 0 or 1
 	git diff --name-only FETCH_HEAD || test $? -le 1 || exit 1
 	# Make the dependencies if any changed files are requirements files, otherwise exit
-	git diff --name-only FETCH_HEAD | grep '^requirements/' || exit 0 && $(MAKE) deps
+	git diff --name-only FETCH_HEAD | grep '^requirements/' || exit 0 && bin/deps
```
ewjoachim, line 93: So, that part would not be done inside Docker, right ?

bin/pip-compile
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+
+# Click requires us to ensure we have a well configured environment to run
+# our click commands. So we'll set our environment to ensure our locale is
+# correct.
+export LC_ALL="${ENCODING:-en_US.UTF-8}"
+export LANG="${ENCODING:-en_US.UTF-8}"
+
+COMMAND_ARGS=$@
```
ewjoachim, line 10: Here's, it might actually make a difference to use `"$@"` with double quotes.

From `man bash`:
```
       @      Expands to the positional parameters, starting from one.   When  the  expansion  occurs  within  double
              quotes,  each  parameter  expands to a separate word.  That is, "$@" is equivalent to "$1" "$2" ...  If
              the double-quoted expansion occurs within a word, the expansion of the first parameter is  joined  with
              the  beginning  part  of  the original word, and the expansion of the last parameter is joined with the
              last part of the original word.
```

So the difference is if one of the parameters is `"a b"`, with `"@"`, it stays that way, and with `$@` it becomes 2 distinct parameters.

ewjoachim (approved):
A few comments, but overall, 👍  from me

bin/docs
di: Just some leftover copypasta, I removed it.

Makefile
di: We can!

Makefile
di: Correct, and great point. I moved this out of the Makefile entirely, and also caught that I missed running this in CI.

bin/pip-compile
di: Updated to use quotes everywhere we do this.

ewdurbin:
Everything seems to work except translations. I did a full purge/clean etc and 

```
ee-imacpro:warehouse ee$ make translations
docker-compose run --rm web env -i  bin/translations
Creating warehouse_web_run ... done
bin/translations: line 7: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory
+ make -C warehouse/locale/ translations
make: Entering directory '/opt/warehouse/src/warehouse/locale'
cd ../../; \
PYTHONPATH=/opt/warehouse/src pybabel extract \
	-F babel.cfg \
	--omit-header \
	--output="warehouse/locale/messages.pot" \
	warehouse
/bin/sh: 2: pybabel: not found
make: *** [Makefile:4: compile-pot] Error 127
make: Leaving directory '/opt/warehouse/src/warehouse/locale'
ERROR: 2
make: *** [translations] Error 2
```

abitrolly:
Tests on dev still fail https://github.com/pypa/warehouse/pull/10052

PR #10819, opened 2022-02-23
Use EthicalAds to display PSF sponsor logos at sponsor page

berinhard:
This PR is a #10766 follow-up. Its core is the update on sponsors page to display PSF sponsors logo from EthicalAds instead of using from pypi's static images. This PR also improves sponsor detail admin by displaying the information from logo placement's API as read-only data. Here's a screen shot of how it looks like:

Screenshot from 2022-02-23 18-11-34

### DIFF
--- a/warehouse/admin/templates/admin/sponsors/edit.html
+++ b/warehouse/admin/templates/admin/sponsors/edit.html
@@ -136,6 +136,25 @@
             {{ render_field("One Time", form.one_time, "sponsor-one-time") }}
             {{ render_field("Sidebar", form.sidebar, "sponsor-sidebar") }}
 
+            {% if sponsor %}
+            <div class="box-header with-border">
+              <h3 class="box-title">Info from Python.org Logo Placement API</h3>
+              <hr>
+
+              <div class="">
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Origin</label>
+                <p>{{ sponsor.origin }}</p>
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Level Name</label>
+                <p>{{ sponsor.level_name|default("---", true) }}</p>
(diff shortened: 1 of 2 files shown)


warehouse/admin/templates/admin/sponsors/edit.html
```diff
@@ -136,6 +136,25 @@ <h3 class="box-title">Create sponsor</h3>
...
+              <hr>
+
+              <div class="">
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Origin</label>
+                <p>{{ sponsor.origin }}</p>
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Level Name</label>
+                <p>{{ sponsor.level_name|default("---", true) }}</p>
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Level Order</label>
+                <p>{{ sponsor.level_order|default("---", true) }}</p>
+                <label for="{{ input_id }}" class="col-sm-2 align-right">Slug</label>
```
merwok, line 151: What do labels mean without a form?  Also they all have the same `for` attribute which doesn’t make sense to me

miketheman:
#13953 fixed the task, so now it's running correctly.

Is this PR still desirable?

berinhard:
@miketheman I've been away from the project for a while, but I'm 100% sure @ewdurbin can answer you better.

ewdurbin:
Yeah, I think we'd want to proceed with this!

firefly9966 (approved):

warehouse/admin/templates/admin/sponsors/edit.html
merwok: But the real issue here is probably the use of labels outside of a form!

PR #10844, opened 2022-02-28
Deduplicate "instead" in `basic-auth-with-2fa`

webknjaz:
Ref #10836


webknjaz:
_(noticed during translation)_

di (approved):

PR #10848, opened 2022-03-01
Add support for limited repeated emails

### LINKED ISSUE #10847: "Migrate to API tokens" sent many many times
**Describe the bug**
I recently uploaded with `twine` and basic auth and was sent *12* emails .  I uploaded a total of 4 files so at most I expect 4 emails and ideally only 1

**Expected behavior**
not to receive 12 duplicate emails in quick succession

**To Reproduce**
upload a package using `twine` with basic auth

**My Platform**

```console
$ twine --version
twine version 3.7.1 (importlib_metadata: 4.10.0, pkginfo: 1.8.2, requests:
2.27.1, requests-toolbelt: 0.9.1, tqdm: 4.62.3)
$ ~/opt/venv/bin/python --version --version
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0]
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

**Additional context**
n/a

<summary> a raw email

<details>

```
Subject: [PyPI] Migrate to API tokens for uploading to PyPI
From: PyPI <noreply@pypi.org>
To: asottile <...>
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="===============3208068886958749528=="
Message-ID: <0101017f4330a776-e90c22a8-0638-478b-9aa3-cda048ea52e1-000000@us-west-2.amazonses.com>
Date: Tue, 1 Mar 2022 01:55:36 +0000
Feedback-ID: 1.us-west-2.ZRFc7KXPaRh5E01FOWpg1s1VfcS7E0CO2O03ZZnJB3U=:AmazonSES
X-SES-Outgoing: 2022.03.01-54.240.27.21
X-Area1Security-Origin: EXTERNAL
X-Area1Security-Processed: 086babc1a6160fb7de2a5f52e72e77ec;2;NONE;2022-03-01T01:55:47

--===============3208068886958749528==
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: quoted-printable


# What?

During your recent upload or upload attempt to PyPI, we noticed you used ba=
sic authentication (username & password). However, your account has two-fac=
tor authentication (2FA) enabled.

In the near future, PyPI will begin prohibiting uploads using basic authent=
ication for accounts with two-factor authentication enabled. Instead, we wi=
ll require API tokens to be used.

# What should I do?

First, generate an API token for your account or project at https://pypi.or=
g/manage/account/token/. Then, use this token when publishing instead of yo=
ur username and password. See https://pypi.org/help/#apitoken for help usin=
g API tokens to publish.





--===============3208068886958749528==
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: quoted-printable
MIME-Version: 1.0

<!DOCTYPE html>
<html>
  <head>
    <meta charset=3D"utf-8">
    <meta name=3D"viewport" content=3D"width=3Ddevice-width">

    </head>
  <body>
   =20
<h3>What?</h3>
<p>
  During your recent upload or upload attempt to PyPI, we noticed you used =
basic authentication (username &amp; password). However, your account has t=
wo-factor authentication (2FA) enabled.
</p>
<p>
  In the near future, PyPI will begin prohibiting uploads using basic authe=
ntication for accounts with two-factor authentication enabled. Instead, we =
will require API tokens to be used.
</p>

<h3>What should I do?</h3>
<p>
  First, generate an API token for your account or project at https://pypi.=
org/manage/account/token/. Then, use this token when publishing instead of =
your username and password. See https://pypi.org/help/#apitoken for help us=
ing API tokens to publish.
</p>


   =20
  </body>
</html>

--===============3208068886958749528==--
```

</details>
</summary>

di:
Fixes #10847.

(cc @asottile)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -30,6 +30,7 @@ from psycopg2.errors import InvalidCatalogName
 from pyramid.i18n import TranslationString
 from pyramid.static import ManifestCacheBuster
 from pyramid_jinja2 import IJinja2Environment
+from pyramid_mailer.mailer import DummyMailer
 from pytest_postgresql.config import get_config
 from pytest_postgresql.janitor import DatabaseJanitor
 from sqlalchemy import event
@@ -38,6 +39,9 @@ import warehouse
 
 from warehouse import admin, config, static
 from warehouse.accounts import services as account_services
+from warehouse.accounts.interfaces import ITokenService
+from warehouse.email import services as email_services
+from warehouse.email.interfaces import IEmailSender
 from warehouse.macaroons import services as macaroon_services
 from warehouse.metrics import IMetricsService
 
@@ -113,11 +117,14 @@ class _Services:
 
 
 @pytest.fixture
-def pyramid_services(metrics):
+def pyramid_services(metrics, email_service, token_service):
     services = _Services()
 
     # Register our global services.
     services.register_service(metrics, IMetricsService, None, name="")
+    services.register_service(email_service, IEmailSender, None, name="")
+    services.register_service(token_service, ITokenService, None, name="password")
+    services.register_service(token_service, ITokenService, None, name="email")
 
     return services
 
@@ -279,6 +286,13 @@ def macaroon_service(db_session):
 @pytest.fixture
 def token_service(app_config):
     return account_services.TokenService(secret="secret", salt="salt", max_age=21600)
+
+
+@pytest.fixture
+def email_service():
+    return email_services.SMTPEmailSender(
+        mailer=DummyMailer(), sender="noreply@pypi.dev"
+    )
 
 
 class QueryRecorder:
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -19,7 +19,7 @@ import pytest
 from sqlalchemy.orm.exc import NoResultFound
 
 from warehouse import email
-from warehouse.accounts.interfaces import ITokenService, IUserService
+from warehouse.accounts.interfaces import IUserService
 from warehouse.email.interfaces import IEmailSender
 from warehouse.email.services import EmailMessage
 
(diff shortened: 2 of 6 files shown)


ewdurbin (approved):
I'm still not sure why we send 3 per upload. My only concern is that the three sends could happen more or less simultaneously in separate workers, since the session/transaction won't be shared we could still send more than one on the first notification within a given time window.

ewdurbin:
Second thought is now that we're limiting by subject, maybe we should revert to sending one email per released version?

di:
> Second thought is now that we're limiting by subject, maybe we should revert to sending one email per released version?

The `repeat_window` param is set per-email. It's set to 1 day  for the 2FA reminder email, but all other emails will have a default delta of 0: https://github.com/pypa/warehouse/pull/10848/files#diff-43d0c472ea153e3fb6daafdedf32184357a81c6bce3ed3e8fc3069fc14ddc0c5R239-R243

di:
Realized that a default window of 0 seconds may still contain microseconds, so I set the default to `None` instead.

ewdurbin:
Parameter is set per email, but last_sent calculates based on to/subject.

So I'm proposing setting the subject back how it was initially in this implementation to include the project name and release version number. That way we notify per version upload per day. Maybe it's more mail than we like to send, but that information is probably helpful for releases coming from automation or large teams to understand where they need to migrate to API tokens.

ewdurbin:
I see where I goofed in the first comment now 😅. I meant revert to sending an email with the version number included

di:
> Parameter is set per email, but last_sent calculates based on to/subject.

But this change only prevents delivery if the email is sent with `repeat_window` set, and this is only set for the `basic-auth-with-2fa` email. The change in #10829 is for `removed-project-release` and `yanked-project-release` -- or is that not what you're talking about?

ewdurbin:
Ope yeah, confused that PR with this one. It's probably fine, but I still think there will be folks confused what action triggered the email if we don't include at least the project name somewhere.

di:
I agree, but we don't really have the ability to determine what the project/version triggered the email: The upload view has no sense of authentication method used, and the auth mechanisms have no sense of the context of the request.

We could try to cram something onto the request which distinguishes how it was authenticated that we could later check in the upload view if this ends up being too confusing. For now I just want to try and cut down on the volume a bit.

di:
Gonna merge this and revisit if we can add project name / version to the email tomorrow!

PR #10853, opened 2022-03-01
Issue #9203 Disable mismatch warning when confirm password field is empty

### LINKED ISSUE #9203: Confirm password prompts warning before we start giving input to it.
image

From the above image we can see that confirm password starts prompting "Passwords do not match" but I think that it should start the prompt when we have given at least one character input.

usernameandme:
Fixes #9203.

### DIFF
--- a/tests/frontend/password_match_controller_test.js
+++ b/tests/frontend/password_match_controller_test.js
@@ -45,14 +45,10 @@ describe("Password match controller", () => {
 
   describe("incorrect inputs", function() {
     describe("adding text on one of the fields", function() {
-      it("shows text warning of mismatch and disables submit", function() {
+      it("disables submit", function() {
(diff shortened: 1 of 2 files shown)


di:
Could you add a test for this here? https://github.com/pypa/warehouse/blob/main/tests/frontend/password_match_controller_test.js

usernameandme:
I removed the checks in that test case because that is not the expected functionality. Tests pass locally now.

di:
> I removed the checks in that test case because that is not the expected functionality. Tests pass locally now.

Could we update the test to explicitly check: when both fields are empty, when the first is empty, when the second is empty? Right now it's only testing when the second value is empty.

usernameandme:
Makes sense. Updated.

usernameandme:
@di Can you please review and confirm?

PR #10856, opened 2022-03-02
Models, views, tasks, etc for critical projects promo

di:
~Putting this up as a draft for early review before I write the tests. Everything should be functional.~ Ready for review.


alex:
Didn't review the substance of the PR, but this has me excited!

warehouse/packaging/tasks.py
```diff
@@ -17,11 +17,69 @@
...
+                GROUP BY
+                  file.project
+                ORDER BY
+                  num_downloads DESC
+                LIMIT
+                  {cohort_size}
+            """.format(
+                table=request.registry.settings["warehouse.downloads_table"],
+                cohort_size=request.registry.settings[
+                    "warehouse.two_facor_mandate.cohort_size"
```
reaperhulk, line 53:
```suggestion
                    "warehouse.two_factor_mandate.cohort_size"
```
```suggestion
                    "warehouse.two_facor_mandate.cohort_size"
```

warehouse/cli/2fa.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    request = config.task(_compute_2fa_mandate).get_request()
+    config.task(_compute_2fa_mandate).run(request, project_names)
+
+
+@warehouse.command()
+@click.argument("promofile", type=click.File("r"))
+@click.pass_obj
+def load_promo_codes(config, promofile):
```
ewdurbin, line 37: would probably be easier to have an admin view to upload this file than figure out how to get a file into the deployment environment.

warehouse/templates/pages/security-key-giveaway.html
```diff
@@ -0,0 +1,205 @@
...
+      <h2 id="faq">FAQ</h2>
+      <p>Answers to frequently asked questions regarding this effort:</p>
+
+      <h3 id="critical-eligiblity">What determines if project is a critical project?</h3>
+      <p>
+        PyPI determines project eligibility based on download counts derived
+        from PyPI's <a
+        href="https://warehouse.pypa.io/api-reference/bigquery-datasets.html">public
+      dataset of download statistics</a>. Any project in the top 1% of
+    downloads over the prior 6 months is designated as critical.
```
ewdurbin, line 104: are we not planning to include warehouse dependencies?

warehouse/templates/pages/security-key-giveaway.html
```diff
@@ -0,0 +1,205 @@
...
+      <h3 id="user-requirement-eligiblity">What users are included in the 2FA requirement?</h3>
+      <p>
+        Any maintainer of a critical project (both 'Maintainers' and 'Owners')
+        are included in the 2FA requirement.
+      </p>
+
+      <h3 id="user-key-eligibility">What users are eligible to receive security keys?</h3>
+      <p>
+        PyPI users who are eligible to receive security keys must be
+        maintainers of critical projects who have not previously enabled 2FA on
```
ewdurbin, line 129: I'd really like to encourage security keys over TOTP. could we exclude TOTP as a 2FA method from consideration and extend these to any maintainer who is not currently using WebAuthn?

ewdurbin (commented):
Overall looks like an acceptable setup to me!

warehouse/templates/pages/security-key-giveaway.html
di: I think if we do that we would need to revisit whether 1% is the right % for critical project designation, or whether we're OK with running out of promo codes.

Right now for the top 1%:
* 3853 total maintainers:
  * 2759 do not not have 2FA enabled
  * 1094 have some 2FA enabled:
    * 894 have only TOTP enabled
    * 113 have both TOTP and security keys
    * 87 have only security keys
    
So that change would take the # of eligible users from 2.7K to 3.6K, which means roughly half of them would not be able to get a promo code.

warehouse/cli/2fa.py
di: This was mostly for local testing, I figured we can either load directly into DB, seems like overkill to write an admin view for something we're only going to use ~once.

warehouse/templates/pages/security-key-giveaway.html
ewdurbin: I'm _very_ happy to see that coverage as it is! OK in that case, let's stick with initial plan for now to get greatest coverage.

warehouse/templates/pages/security-key-giveaway.html
di: I considered it but thought that this might open up a broader question of whose dependencies are important, e.g. are warehouse dependencies more critical than twine dependencies? wheel? etc.

warehouse/templates/pages/security-key-giveaway.html
di: I think we can probably ldo some auxiliary efforts (informational callouts, etc) to generally motivate the TOTP-only users to adopt hardware keys.

warehouse/templates/pages/security-key-giveaway.html
ewdurbin: Based on the centralized nature of PyPI, I would argue that yes it is more important to protect in depth. A compromise of PyPI that is undetected for any length of time could have catastrophic impacts downstream. Installers are probably next highest priority, and then upload tools by the metric of "how bad how fast".

di: Fine with me, should be trivial to add this to the task, as long as we're OK taking a runtime dependency on https://github.com/di/pip-api 😉

warehouse/packaging/tasks.py
```diff
@@ -14,14 +14,73 @@
...
+              file.project
+            ORDER BY
+              num_downloads DESC
+            LIMIT
+              {cohort_size}
+        """.format(
+            table=request.registry.settings["warehouse.downloads_table"],
+            cohort_size=request.registry.settings[
+                "warehouse.two_factor_mandate.cohort_size"
+            ],
```
woodruffw, line 61: I don't think there's any possibility for an injection vulnerability here, but just as a matter of best practice: maybe parametrized queries here instead of raw interpolation?

Looks like biquery supports parametrization with `@identifier`: https://cloud.google.com/bigquery/docs/parameterized-queries

warehouse/templates/email/two-factor-mandate/body.html
```diff
@@ -0,0 +1,37 @@
...
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+
+{% block content %}
+<p>
+Congratulations! A project you ('{{ username }}') maintain has been designated as a critical project.
```
woodruffw, line 19: Needs `trans` markers here and below I think, for eventual i18n?

warehouse/templates/email/two-factor-mandate/body.txt
```diff
@@ -0,0 +1,29 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.txt" %}
+
+{% block content %}
+Congratulations! A project you ('{{ username }}') maintain has been designated as a critical project.
```
woodruffw, line 17: Same note about `trans`.

warehouse/templates/pages/security-key-giveaway.html
```diff
@@ -0,0 +1,205 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "base.html" %}
+{% block title %}PyPI 2FA Security Key Giveaway{% endblock %}
+
+{% block content %}
```
woodruffw, line 17: Same note about `trans`.

warehouse/packaging/tasks.py
```diff
@@ -14,14 +14,73 @@
...
+
+    # Get their maintainers
+    users = request.db.query(User).join(Project.users).join(new_projects).all()
+
+    # Email them
+    for user in users:
+        send_two_factor_mandate_email(request, user)
+
+    # Add them to the mandate
+    new_projects.update({Project.pypi_mandates_2fa: True})
```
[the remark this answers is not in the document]
di: No return, the `update` mutates the ORM model and thus the row in storage.

warehouse/templates/email/two-factor-mandate/body.html
di: These email templates join some of our other email templates that don't have a logged in user, thus we don't have a locale for them to translate this with, so it doesn't really make sense to mark these for translation now.

warehouse/packaging/tasks.py
di: This is a good tip, but I'm not concerned about it here, as these values only set by PyPI admins.

warehouse/packaging/__init__.py
```diff
@@ -101,6 +102,10 @@ def includeme(config):
     config.add_periodic_task(crontab(minute="*/5"), update_description_html)
     config.add_periodic_task(crontab(minute="*/5"), update_role_invitation_status)
 
+    # Add a periodic task to recompute the critical projects list once a day
+    if config.get_settings().get("warehouse.two_factor_mandate.available"):
```
ewdurbin, line 106:
```suggestion
    if config.get_settings().get("warehouse.two_factor_mandate.available", False):
```

ewdurbin (approved):
a single nit. otherwise, 👍🏼

di:
Rollout plan for this is here: https://github.com/pypa/warehouse/issues/11625

PR #10859, opened 2022-03-02
Invalidate existing sessions for a given User when User.password is changed

### LINKED ISSUE #10849: Invalidate active sessions on password change
In the event of a password change, all other active sessions should be invalidated.

ewdurbin:
Closes #10849.


warehouse/accounts/services.py
```diff
@@ -80,14 +82,19 @@ def __init__(self, session, *, ratelimiters=None, remote_addr, metrics):
         )
         self.remote_addr = remote_addr
         self._metrics = metrics
+        self.cached_get_user = functools.lru_cache()(self._get_user)
```
ewdurbin, line 85: Side effect of scrambling with figuring out why I couldn't get the latest `password_date` value... but interestingly this may resolve a memory leak? https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html

warehouse/accounts/models.py
```diff
@@ -108,6 +108,15 @@ class User(SitemapMixin, db.Model):
...
     )
 
+    @property
+    def password(self):
+        return self._password
+
+    @password.setter
+    def password(self, value):
+        self._password = value
+        self.password_date = sql.func.now()
```
di, line 118: I think we can commit the transaction early to get the new date and thus retain the existing DB triggers. I think it's probably desirable for those to continue to remain in the DB so we don't easily forget to update `password_date`.

ewdurbin: I can't figure out a way to get the resulting trigger value to be accessible inside the request. Calling `tm.commit()` mid request crashes the request at the end since txn complete... savepoint() doesn't work...

warehouse/sessions.py
```diff
@@ -145,6 +146,15 @@ def record_auth_timestamp(self):
...
         self.changed()
 
+    def record_password_timestamp(self, timestamp):
+        self[self._password_timestamp_key] = timestamp
+        self.changed()
+
+    def password_outdated(self, current_password_timestamp):
+        stored_password_timestamp = self.get(self._password_timestamp_key, 0)
+
+        return current_password_timestamp != stored_password_timestamp
```
di, line 156: I think this would have the effect of logging out all currently active sessions because their sessions lack the `_password_timestamp` key, and thus this comparison will fail for them.

ewdurbin: Great catch. I was initially going to implement the password date check based on a >=, then realized equality was a better check.

I'll update the default to None and add a conditional.

ewdurbin: fixed in https://github.com/pypa/warehouse/pull/10859/commits/61ce40b0b30cadf1a907af1d552299ad474015fd

warehouse/accounts/models.py
ewdurbin: trying sqlalchemy expire call... might work!

ewjoachim: ... Ok, today is the day where my comments land 30s late.

ewdurbin: yah, absolutely not ideal, but this is no longer necessary with https://github.com/pypa/warehouse/pull/10859/commits/8309b0057b0677000a577de4d81facd39a23653a.

for clarity, sql.func.now() gets translated to `now()` on insert, it's not calling the function.

ewjoachim: (Yeah 👍  but to explicit my comment, if we had forced a commit at that point it would have been doing a SQL query in the setter)

ewdurbin: ah, I wasn't proposing a commit in the setter, I was saying I had attempted all of these things in the view.

ewdurbin: specifically, where refresh came to the rescue: https://github.com/pypa/warehouse/pull/10859/commits/8309b0057b0677000a577de4d81facd39a23653a#diff-55d42e8160b62924651b14b5c3649c6511c24a79876c8447cdad028da727d14fR364

ewdurbin:
Sanity check done locally:

- Switched back to `main`
- Logged in with a brand new session
- Switched to this branch
- Stayed logged in!
- Opened new window and changed password
- New session that changed password stayed logged in, Old session terminated.

warehouse/sessions.py
```diff
@@ -145,6 +146,20 @@ def record_auth_timestamp(self):
         self[self._reauth_timestamp_key] = datetime.datetime.now().timestamp()
         self.changed()
 
+    def record_password_timestamp(self, timestamp):
+        self[self._password_timestamp_key] = timestamp
+        self.changed()
+
+    def password_outdated(self, current_password_timestamp):
+        stored_password_timestamp = self.get(self._password_timestamp_key, None)
```
di, line 154:
```suggestion
        stored_password_timestamp = self.get(self._password_timestamp_key)
```
nit

warehouse/sessions.py
```diff
@@ -145,6 +146,20 @@ def record_auth_timestamp(self):
...
+
+        if stored_password_timestamp is None:
+            # This session predates invalidation by password reset... since
+            # we cannot say for sure, let it live its life.
+            return False
+
+        return current_password_timestamp != stored_password_timestamp
```
di, line 161: I like that the `!=` comparison also covers our bases against time-travelers.

di (approved):

warehouse/sessions.py
ewdurbin: but explicit 😭

Issue #10865, opened 2022-03-03
Explore adding static typing


di:
We should determine a way that we can gradually add static typing (and ensure that we're able to assert that it is correct in CI) without introducing a single massive PR.

fizyk:
mypy has configuration options that would allow doing that gradually

https://mypy.readthedocs.io/en/stable/config_file.html

We could either include paths that got converted or exclude those that do not (this would require all new ones to already be typed)

miketheman:
I'm interested in this kind of work.

A few questions/decisions arise at this juncture:

- Is mypy considered a lint tool, or a testing tool? (for requirements inclusion) I think lint - from the mypy docs:
  > As mypy is a static analyzer, or a lint-like tool, ...
- Is there a particular entry point or an area of focus? We can also follow the [recommended path](https://mypy.readthedocs.io/en/stable/existing_code.html)
- Where do we want to configure mypy? I'm partial to `pyproject.toml`, but that's not currently mounted to the developer container - should we mount that as well?

di:
> Is mypy considered a lint tool, or a testing tool? (for requirements inclusion)

I think this belongs with our linters.

> Is there a particular entry point or an area of focus? We can also follow the recommended path

There are some existing types that @fizyk did that could be a good starting point. Other good starting points would be our more critical codepaths.

> Where do we want to configure mypy? I'm partial to pyproject.toml, but that's not currently mounted to the developer container - should we mount that as well?

Using `pyproject.toml` is preferable, and this should already be getting included in the dev container, as there's configuration for other linters and pytest in there as well

miketheman:
I think this can be closed now, as we have a mechanism for adding more types as gradually as we want to now.

PR #10874, opened 2022-03-07
chore: add mypy

miketheman:
Starts #10865 

See each commit for reasoning and descriptions.

Check out this branch, run `pip install -r requirements/lint.txt && bin/lint` for a nice success.


miketheman:
I’ll correct the failing lint tomorrow. Feels weird to have to move lines to allow for an ignore, so I may also ignore the flake8 E501 on these two lines as well.

miketheman:
@di I think this is good to go - I didn't add mypy to `bin/lint` just yet, but could since it's passing right now.

bin/mypy
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+python -m mypy -p warehouse
```
di, line 4: Let's go ahead and add this to `bin/lint` instead.

di (approved):
This LGTM, I'll let @ewdurbin review as well. Nice to see this already catching bugs.

ewdurbin (commented):
I don't feel like I have enough understanding here to explicitly approve this 😅.

What configuration is allowing us to implement this gradually? Will mypy detect when files have changed, or does it just skip over unannotated code?

The only other concern I have here is the long list of explicitly untyped modules, is there some way for us to know when typing is added or stubs are available? Feels like it will fall on those motivated to keep track and send PRs which I suppose is fine.

I'm not at all against merging this and starting down the path, I'm interested to see what the impact on workflow will be for myself and other contributors!

bin/mypy
miketheman: Reverted inclusion of this file and added to `bin/lint` in recent commit.

miketheman:
I'll try to clarify as best as I understand!

> What configuration is allowing us to implement this gradually?

The config as I've started will warn if we added mypy configs that aren't used (useful to prevent config bloat), and examine every file in scope for potential inconsistencies, using type inference.

The fix made in [`69a07b9` (#10874)](https://github.com/pypa/warehouse/pull/10874/commits/69a07b9680daacb1af2e505ab3c9a251629609ff) is an example of this - the function signature declared that `since: int` and then redefined it inside the function scope to a `datetime`, so that triggered a mypy alert of:

    warehouse/legacy/api/xmlrpc/views.py:550: error: Incompatible types in assignment (expression has type "datetime", variable has type "int")

If we wanted to allow this behavior, we would configure `--allow-redefinition` ([ref](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition)) - but since there was a single instance in the entire codebase, it felt like a good opportunity to fix.

While this is not a bug per se, since the code works, and Python allows redeclaration of a variable, mypy points out that this is a confusing behavior to the reader of the code, so a rename to a more explicit variable helps comprehension and solves the error.

> Will mypy detect when files have changed, or does it just skip over unannotated code?

Yes! By running each time with CI, we lint everything in the `warehouse` module. (We _could_ also lint `tests/` but that's likely overkill.)

We can opt in to stricter configurations one at a time, preventing a huge PR, and then finally flip the switch to be `--strict` and prevent things from entering the codebase at that time.

Here's a list of flags under `--strict`:

> Strict mode; enables the following flags: --warn-unused-configs, --disallow-any-generics, --disallow-subclassing-any, --disallow-untyped-calls, --disallow-untyped-defs, --disallow-incomplete-defs, --check-untyped-defs, --disallow-untyped-decorators, --no-implicit-optional, --warn-redundant-casts, --warn-unused-ignores, --warn-return-any, --no-implicit-reexport, --strict-equality

So I'd advise taking on each one, adding to the config, running the lint, fixing/confirming, and shipping changes in iterative fashion.

> The only other concern I have here is the long list of explicitly untyped modules, is there some way for us to know when typing is added or stubs are available? Feels like it will fall on those motivated to keep track and send PRs which I suppose is fine.

Not really, which is why I've added PR links that I could find for projects that may add typing in the future.

The [typeshed project](https://github.com/python/typeshed) collects and maintains a lot of `types-*` packages, but it's really up to the package's to ship a `py.typed` file to signal they have type information, per [PEP-561](https://www.python.org/dev/peps/pep-0561/).

One excellent way to reduce the amount of untyped packages in scope is to... reduce the total number of packages in use. 😁 But barring that, it's reasonable to check back in on packages as part of their upgrade cycles.

Let me know if this makes sense!

ewdurbin:
Yes. Thanks for the clarification @miketheman! @di I'll let you push the big green button.

PR #10925, opened 2022-03-11
Add repository name to github section

### LINKED ISSUE #10917: Show repository name, in the GitHub statistics section
Currently, it's not possible to discover the name of the GitHub repository, associated with a project's metadata from the PyPI page directly. It is also occasionally not immediately clear which of the URLs associated with a project is a GitHub URL.

This makes it possible for projects to misrepresent what their GitHub statistics are, by using a more popular Python project's URL on their metadata.

Change "GitHub statistics:" to instead be "GitHub: {repo-name}". So, for pip, that would be "GitHub: pypa/pip".

This would:

- provide an "obvious thing" to make into a link to the GitHub repository.
- surface this information to help identify projects that are misrepresenting themselves (intentionally or otherwise).

**Additional context**
See https://discuss.python.org/t/14181/, where this was originally proposed.

merwok:
Fixes #10917


pradyunsg (commented):
LGTM, but I'll let one of the admins merge this. :)

merwok:
Do people think it’s a problem if a repo link ends up in the project URLs section and the github stats section?

pradyunsg:
I certainly don't! They both serve distinct functions IMO -- one presenting the preferred language from the package author and another presenting a consistent + at-a-glance view of what the package reports as its URL.

di:
Aside from the other discussion in #10917, this doesn't seem to actually do what's originally requested in that issue:

> Change "GitHub statistics:" to instead be "GitHub: {repo-name}". So, for pip, that would be "GitHub: pypa/pip".

merwok:
Yes, I disagreed with that specific part of the proposal!  Prefer to keep the existing full title, and clarify what we are reading (let’s not assume that everyone understands «some letters slash more letters», «some word hyphen sig», «r slash word», etc 😉)

pradyunsg:
And I think this is perfectly reasonable as well. :)

merwok:
@di @dstufft could you review this?  I can change the code if you both want the format `github: username/repo`, but I do think my version is clearer!

PR #10992, opened 2022-03-21
Add User.is_frozen

di:
Adds the ability for an admin to freeze/unfreeze a user account.


ewdurbin (approved):
+1 but a little bit of help text is probably a good idea.

Alternatively we set it up so when freezing we can state a reason and it is emailed to the user.

PR #11024, opened 2022-03-26
Replace `fakeredis` with a class

### LINKED ISSUE #11015: The fakeredis project is unmaintained
Per https://github.com/jamesls/fakeredis/commit/798d534bd4d6d396ba5082184e478f160f791940, the `fakeredis` project is unmaintained. This is one of our test dependencies:

https://github.com/pypa/warehouse/blob/6968519ad7de18b9adcad0bf8db02ab85075f7c5/requirements/tests.in#L4

And it's lack of support for `redis==4.2.0` (https://github.com/jamesls/fakeredis/pull/328) is blocking #11014.

We should evaluate our usage of `fakeredis` and determine if we need to find a way to keep the project updated or stop using it.

miketheman:
Replace usage of the now-unmaintained `fakeredis` library with a "good enough" pytest fixture that behaves close enough to the `redis-py` => redis server interaction for our testing purposes.

I tried to figure out the right way to construct these with `pretend.stub()`, I found it simpler with a class/fixture that we can reuse.

Many of the methods on `_MockRedis` are in support of the behaviors exercised in `tests/unit/legacy/api/xmlrpc/test_cache.py` - so if we ever removed the legacy API code and tests, we could remove methods like `pipeline()`, `scan_iter()`, `execute()` at that time.

Closes #11015
Closes #11014 (cherry picked)


miketheman:
@di I think this is now ready for review! I pulled in your change to the redis deps as well - once this is merged, I don't know if dependabot will retry automatically or not.

di:
Thanks @miketheman!

PR #11113, opened 2022-04-05
Anti-spam suggestion for pypi repositories

rakovskij-stanislav:
Hello!

I found out that there is a spammer who creates many repositories containing only a PKG-INFO file advertising his (probably fake) roblox currency generator, amazon gift codes, etc.

Even though this repository does not have a malicious effect on developers, it is definitely a malicious/unwanted activity.

One of such "developers" for example:

image

Google indexes this kind of links, it may harm the reputation of pypi:

image

So I wrote a heuristics for this kind of situations.

### DIFF
--- a/warehouse/malware/checks/pkg_info_patterns/check.py
+++ b/warehouse/malware/checks/pkg_info_patterns/check.py
@@ -1,0 +1,113 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


rakovskij-stanislav:
For example, one of this repos: https://pypi.org/project/roblox-free-robux-no-survey-2022-roblox-for-robux-v-2529/

image

warehouse/malware/checks/pkg_info_patterns/check.py
```diff
@@ -0,0 +1,115 @@
...
+        release_file = kwargs.get("obj")
+        file_url = kwargs.get("file_url")
+        if release_file is None or file_url is None:
+            raise FatalCheckError(
+                "Release file or file url is None, indicating user error."
+            )
+
+        archive_stream = fetch_url_content(file_url)
+
+        file_list = extract_file_list(archive_stream)
```
di, line 59: I'm a bit wary of extracting user-submitted files here, as we don't have any sort of sandboxing in place.

rakovskij-stanislav: https://github.com/pypa/warehouse/blob/main/warehouse/malware/checks/setup_patterns/check.py#L63

But one of existing modules already does it)

PR #11121, opened 2022-04-05
Noindex new projects

di:
This PR adds a Jinja filter `is_recent` that can be used to determine if a given timestamp is within the last 30 days. This is then used on the project page to add a `noindex` meta tag for projects created recently, which should give us sufficient time to remove spam before googlebot indexes the project.


ewdurbin:
I thought we already had SEO blocking measures in place, but perhaps that was only on links? Regardless, great idea.

PR #11122, opened 2022-04-05
Add `ExpiryCaveat`

### LINKED ISSUE #6255: Add caveats to macaroons for expiration (time) and version

### LINKED ISSUE #10792: Update Macaroon claims/constraints for OIDC token minting
We're going to want to mint short-lived Warehouse API tokens (which are Macaroons) based on incoming OIDC JWTs.

As part of the "short-lived" part, we'll need need to make sure the tokens support some kind of expiration claim.

woodruffw:
Closes #10792.

Closes #6255.


woodruffw:
This needs more testing, but the basic idea:

* All current macaroons continue to function as expected, and nothing about the construction of normal macaroons (i.e., ones minted by users manually through the PyPI website) is changed.

* In the future, macaroons minted via the OIDC JWT consumption route will include an additional expiry caveat, with two fields: `nbf` to indicate that the macaroon must not be used before the given timestamp, and `exp` to indicate when the macaroon ceases to be valid.

warehouse/macaroons/services.py
```diff
@@ -135,7 +135,11 @@ def create_macaroon(self, location, user_id, description, caveats):
         """
         user = self.db.query(User).filter(User.id == user_id).one()
 
-        dm = Macaroon(user=user, description=description, caveats=caveats)
+        # NOTE: This is a bit of a hack: we keep a separate copy of the
+        # permissions caveat in the DB, so that we can display scope information
+        # in the UI.
+        permissions = next(c["permissions"] for c in caveats if "permissions" in c)
+        dm = Macaroon(user=user, description=description, permissions=permissions)
```
woodruffw, line 142: Highlighting this as a bit of a hack: I'm not really happy with this trick, but the alternative would be a `create_macaroon` signature like this:

```python
def create_macaroon(self, location, user_id, description, caveats, permissions):
  ...
```

...where `caveats[N] == permissions`. That feels roughly as bad, so I figured reducing the parameter deduplication was a reasonable tradeoff. But I'm very open to better ideas.

warehouse/manage/views.py
```diff
@@ -879,7 +879,10 @@ def create_macaroon(self):
 
         response = {**self.default_response}
         if form.validate():
-            macaroon_caveats = {"permissions": form.validated_scope, "version": 1}
+            # NOTE: The caveat order here is important: the UI assumes
+            # that the permissions caveat comes first in any recorded
+            # events.
+            macaroon_caveats = [{"permissions": form.validated_scope, "version": 1}]
```
woodruffw, line 885: Highlighting this as another part I'm not satisfied with: we subsequently assume that `macaroon_caveats[0]` is the permissions caveat, e.g. when pretty-printing events. That's unlikely to change, but it's a piece of nasty implicit state here.

woodruffw: (The other option is to embed `form.validated_scope` into the event instead of the entire `macaroon_caveats` list. That would avoid the implicit list position.)

warehouse/macaroons/services.py
```diff
@@ -135,7 +135,13 @@ def create_macaroon(self, location, user_id, description, caveats):
         """
         user = self.db.query(User).filter(User.id == user_id).one()
 
-        dm = Macaroon(user=user, description=description, caveats=caveats)
+        # NOTE: This is a bit of a hack: we keep a separate copy of the
+        # permissions caveat in the DB, so that we can display scope information
+        # in the UI.
+        permissions = next(
+            c["permissions"] for c in caveats if "permissions" in c
+        )  # pragma: no cover
```
woodruffw, line 143: N.B.: I don't think there's any actual coverage missing here; I think `coverage` is getting tripped up by the `StopIteration`.

woodruffw:
Marking as ready for review since this is in a stable state, but see comments above about the current approach.

warehouse/manage/views.py
di: Or maybe we should update the UI to include all caveats?

di (commented):
This looks good, but I didn't give it a super deep review. I had one high level question: how will this be backwards-compatible for macaroons created before this change? The migration is just dropping one column and adding a new one without doing a data migration.

woodruffw:
> I had one high level question: how will this be backwards-compatible for macaroons created before this change?

Yeah, this was an oversight on my part. This will break the UI for pre-existing macaroons, at least in the form they're recorded in the DB.

The migration should probably just rename the column rather than dropping it and inserting a new one, since interior data is the same. I'll fix that now.

warehouse/manage/views.py
woodruffw: Alright, I've added a `caveat_detail` macro that handles both caveat types, removing the need for the implicit array positioning here.

The changes introduced with that macro don't currently manifest in the UI since we don't currently mint any API tokens with the new expiry caveat, but they'll show up once I begin the route work.

woodruffw:
Blocked on: #11157.

di:
#11157 has been merged!

woodruffw:
Excellent! I'll de-conflict here.

woodruffw:
N.B.: [1f65cff](https://github.com/pypa/warehouse/pull/11122/commits/1f65cff664bef00de542e7a1ea4a3eba2172ad52) is a naive pick of the original changeset, so it isn't ready for review yet. I'll do another pass over it later today.

woodruffw:
I also confirmed locally that this does not break verification of current tokens (i.e., ones missing an expiry caveat).

warehouse/macaroons/caveats.py
```diff
@@ -83,6 +103,7 @@ def __init__(self, macaroon, context, principals, permission):
 
     def verify(self, key):
         self.verifier.satisfy_general(V1Caveat(self))
+        self.verifier.satisfy_general(ExpiryCaveat(self))
```
di, line 106: I was expecting to see a test change as a result of this, but it doesn't look like we're testing the `Verifier` class that well. Could you add a test that captures the change in behavior that this line introduces?

woodruffw: This was a great call -- adding these tests ([3e60c4b](https://github.com/pypa/warehouse/pull/11122/commits/3e60c4b51e856a966a4a79259d57cf4ad27eed53)) caught a bug that would have caused us to unconditionally fail all expiry caveat verifications. I've refactored the code to avoid exception-driven failures in caveats going forwards, since that's the root of the problem.

warehouse/macaroons/caveats.py
```diff
@@ -24,9 +25,16 @@ class InvalidMacaroonError(Exception):
 class Caveat:
     def __init__(self, verifier):
         self.verifier = verifier
+        self.failure_reason = None
+
+    def _fail(self, reason):
+        # TODO: Surface this failure reason to the user.
+        # See: https://github.com/pypa/warehouse/issues/9018
```
woodruffw, line 32: This change will make #9018 much easier to implement -- we just need to plumb the `failure_reason` through the verifier and service layers.

ewdurbin (approved):

warehouse/macaroons/caveats.py
```diff
@@ -24,9 +25,16 @@ class InvalidMacaroonError(Exception):
...
+        self.failure_reason = reason
+        return False
 
     def verify(self, predicate):
-        raise InvalidMacaroonError
+        return self._fail("programming error")
```
di, line 37: I'm not sure I love this pattern, for two reasons:

- It's not hard to imagine someone calling `self._fail(...)` (and expecting that call to change the flow) instead of `return self._fail(...)`, accidentally returning the implicit `None` instead. 
- The fact that `self._fail` essentially wraps setting state and just always returns `False` seems unnecessary.

I think I'd prefer sticking with an exception-driven control flow but I'm not sure I understand why you opted to move aware from it here.

woodruffw: I'm also not in love with this pattern, but the original (exception-driven) control flow was actually masking a bug in caveat verification (one that we never experienced, since we've only ever had one caveat per macaroon before).

Using exceptions here was actually causing us to prematurely fail our caveat verification steps, since `pymacaroons` runs each verifier once per caveat body. This meant that the `V1Caveat` would raise as soon as it was passed an `ExpiryCaveat` body, failing the entire verification phase rather than returning control to the next verifier.

IMO we need _some_ kind of state handler to make the error messages returned for macaroons more useful (per #9018), although it definitely doesn't need to be this particular `_fail` thing I just made up :slightly_smiling_face:

di: Got it, that makes sense.

As a compromise, can we just do something like this instead::

```suggestion
        self.failure_reason = "programming error"
        return False
```

and add some static typing to ensure that `verify` always returns a boolean?

woodruffw: Yep, SGTM. I'll make these changes now.

woodruffw: [e0d1709](https://github.com/pypa/warehouse/pull/11122/commits/e0d1709a1898b9c1934f7c9860d6ed2c5daa468b)

di (approved):

fschulze:
What is the deployment policy of pypi.org? Is everything on main automatically deployed? If not, where can I see what version is deployed? In other words, how can I see that changes like this are available on pypi.org?

di:
@fschulze Everything is automatically deployed within 5-10 minutes after merge.

woodruffw:
But also, note that these changes aren't user-facing yet! We don't currently allow users to create API tokens with expiry restrictions.

Issue #11070, opened 2022-04-01
Create an Organization account


s-mm:
Feature request for organization account project in PyPI.

Description | Users can create a new Organization account. The name of the account will reflect the Organization’s name.
-- | --
User value | The Organization account will allow entities to own Projects on PyPI and assign permission levels across team members and projects
Acceptance criteria | Email notification to PyPI Admin and account creator. Pending Organization visible in the PyPI Admin panel including necessary information for review. Approval and rejection actions in the PyPI Admin to either create the organization or reject it.

divbzero:
### Tasks

- [ ] Database: Create Organization table
- [ ] Database: Create OrganizationRole table
- [ ] Database: Create OrganizationRoleInvitation table
- [ ] Database: Create OrganizationNameCatalog table
- [ ] Database: Create OrganizationProject table
- [ ] Interface: List organizations
- [ ] Interface: Find organization by name
- [ ] Interface: Create new organization request
- [ ] UI: List or create organization
- [ ] UI: Confirm new organization request
- [ ] Email: Create organization requested confirmation for user
- [ ] Email: Create organization requested notification for PSF admin

PR #11166, opened 2022-04-12
Request organization email templates

divbzero:
- Email templates for #11070

### DIFF
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -186,6 +186,22 @@ def _email(
     return inner
 
 
+# Email templates for administrators.
+
+
+@_email("admin-new-organization-requested")
+def send_admin_new_organization_requested_email(
+    request, user, *, organization_name, initiator_username
+):
+    return {
+        "initiator_username": initiator_username,
+        "organization_name": organization_name,
+    }
+
+
+# Email templates for users.
+
+
 @_email("password-reset", allow_unverified=True)
 def send_password_reset_email(request, user_and_email):
     user, _ = user_and_email
--- a/warehouse/email/__init__.py
+++ b/warehouse/email/__init__.py
@@ -283,6 +283,11 @@ def send_primary_email_change_email(request, user_and_email):
     }
 
 
+@_email("new-organization-requested")
+def send_new_organization_requested_email(request, user, *, organization_name):
+    return {"organization_name": organization_name}
+
+
 @_email("collaborator-added")
 def send_collaborator_added_email(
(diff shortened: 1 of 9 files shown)


divbzero:
Rebased against `main`.

warehouse/email/__init__.py
```diff
@@ -186,6 +186,22 @@ def wrapper(request, user_or_users, **kwargs):
     return inner
 
 
+# Email templates for administrators.
+
+
+@_email("admin-new-organization-requested")
+def send_admin_new_organization_requested_email(
```
ewdurbin, line 193: We will likely want the recipient of these emails to be configurable, and a null value for it means "no email we'll go look at the admin queue ourself".

We may want to construct an email message directly https://github.com/pypa/warehouse/blob/439836ffd7570be4dec17be2146b543544fb11f0/warehouse/email/__init__.py#L165 and call https://github.com/pypa/warehouse/blob/439836ffd7570be4dec17be2146b543544fb11f0/warehouse/email/__init__.py#L58 directly.

ewdurbin: Currently we _only_ ever send emails to user accounts, but in this case it will likely be either a distribution list or a comma separated list of email addresses. Not all PyPI Admins should be notified unless they want to opt in.

@di, thoughts?

di: Eh, I think just sending these to https://pypi.org/user/admin/ would be fine for now. Eventually when we support orgs/teams this could become more granular perhaps?

ewdurbin: that sparks a decent idea. a wrapper that sends to a pyramid principal rather than a user?

divbzero: > that sparks a decent idea. a wrapper that sends to a pyramid principal rather than a user?

@ewdurbin Are you thinking we should send the notification emails to every `"psf_staff"`, `"admins"`, or another Pyramid principal that we haven’t defined yet? I suppose this would be the same group with permission to approve/decline organizations in the admin interface?

divbzero: @ewdurbin If we want to send emails to all `"admins"` one approach would be to just grab all admin users from the database:

https://github.com/pypa/warehouse/blob/d38989f02ff7758e2178627ec5af788e53538b96/warehouse/accounts/services.py#L104-L106

https://github.com/pypa/warehouse/blob/d4611f07b47b6b684cabcb46fa5a2a86d3a43b57/warehouse/manage/views.py#L1021

divbzero:
The changes in this PR are now part of #11184.

Issue #11087, opened 2022-04-01
Approve account


s-mm:
Feature request for organization account project in PyPI.

Description | PyPI Admin approves organization account
-- | --
User value | PyPI Admin ensures that there is no spurious use of an Organization’s name. PyPI Admin can control the number of accounts being created.
Acceptance criteria | Email notification to owners and PyPI Admin

divbzero:
### Subtasks

- [ ] Interface: Approve new organization request
- [ ] Interface: Decline  new organization request
- [ ] UI: Approve-or-decline form for PSF admin
- [ ] Email: Create organization approved notification for user
- [ ] Email: Create organization approved notification for PSF admin
- [ ] Email: Create organization declined notification for user
- [ ] Email: Create organization declined notification for user

Issue #7583, opened 2020-03-24
Emails language to be based on receiver nationality 


MVrachev:
**What's the problem this feature will solve?**
While I was working to add email notifications for project/release removed we had a discussion that translation of those emails can't be used because the nationality of the initiator of that action could be different from that of the receiver.
For example if the initiator of the project removal is from Spain that doesn't mean that the rest of the contributors of that package will speak spanish.

Read more about that discussion here: https://github.com/pypa/warehouse/pull/7530#issuecomment-598897978

**Describe the solution you'd like**

I propose to use the receiver nationality when deciding what language to use in the emails.
That seems logical solution to me as we are doing just that in our daily lives - in our social interactions with other people we choose our language of communication based  on the language spoken from the other person.  
That way we would be able to translate the emails which are aimed for groups of people. 

The question is do we have information about the nationality of ever user of PyPI?

di:
> The question is do we have information about the nationality of ever user of PyPI?

We don't currently have this information. Presumably when a logged-in user selects a locale, we could store it along with their other account details:

https://github.com/pypa/warehouse/blob/519afb3b82a682258b5baa8263cdd7813bf6e961/warehouse/views.py#L234-L258

This would probably also require some changes to how we determine the locale based on the request, to take this new attribute into account.

di:
> This would probably also require some changes to how we determine the locale based on the request, to take this new attribute into account.

By this I actually mean, how we determine for the locale for the localizer when translating emails, since right now it is based on the request, but it will need to use the attribute on the user who the email is being sent to instead. We'd likely need to construct new localizers for each user with this attribute. We do something similar here:

https://github.com/pypa/warehouse/blob/519afb3b82a682258b5baa8263cdd7813bf6e961/warehouse/views.py#L253-L254

almet:
Thanks @miketheman for pointing me here.

Unless I'm missing something, the preferred language of the users [is not defined in the models](https://github.com/pypi/warehouse/blob/e194e8a11cfb7aff0ec123589faf528649528e99/warehouse/accounts/models.py#L72-L163).

Is there any other way to do this than adding a field to the model? (e.g. is there somewhere this information is already being stored?)

adamchainz:
A few days ago, I received an email notification in Russian, the language of my collaborator, rather than my preferred language, English:

[...]

PR #11168, opened 2022-04-12
Fix missing translations for *.txt files

divbzero:
`subject.txt` and `body.txt` files in email templates have translations that were not being picked up by pybabel. Fixing pybabel config allows us to run `make translations` against all existing `*.txt` files.

### DIFF
--- a/babel.cfg
+++ b/babel.cfg
@@ -1,5 +1,11 @@
 [python: **.py]
+
 [jinja2: **.html]
+encoding = utf-8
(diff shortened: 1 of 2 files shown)


divbzero:
This change would add a few dozen translation strings from `*.txt` files. I think this is the right move but let me know if we’ve left them out on purpose.

divbzero:
I’ll revisit this PR after #11087 is merged.

di:
Marking this as blocked on https://github.com/pypi/warehouse/issues/7583.

firefly9966 (approved):

PR #11192, opened 2022-04-15
Autofocus on the TOTP input

pradyunsg:
By autofocusing on this element, we remove the need for the user to
navigate to it, to perform the primary action on this page.


di:
This needs to handle the case where a user doesn't have TOTP, see https://github.com/pypa/warehouse/issues/9805

pradyunsg:
Closing since https://github.com/pypa/warehouse/pull/9851 has that behaviour.

PR #11209, opened 2022-04-19
correct my name

ewdurbin:
dev db dump is years old, screenshots using my user account are painful


di (approved):
Approved, but also: we've had a lot of migrations since the last dump, which makes tests slow. Maybe we're due to bring the dev db up to date and fix this while we're at it.

ewdurbin:
Also tests do not rely on this DB dump, they build a clean DB?

ewdurbin:
That will still generate another ~60MB file in the git-tree :/

miketheman:
Brings up an interesting question on whether we should try to restore from a DB Dump, or use [Factory Boy](https://factoryboy.readthedocs.io/) to generate known seed data for **development** purposes, since the test suite does generate its own data.

If this is desirable for better developer experience, that's something we could pursue, but it does continue the drift between the "real" database schema in prod vs what we create in dev/test.

miketheman:
Brings up an interesting question on whether we should try to restore from a DB Dump, or use [Factory Boy](https://factoryboy.readthedocs.io/) to generate known seed data for **development** purposes, since the test suite does generate its own data.

If this is desirable for better developer experience, that's something we could pursue, but it does continue the drift between the "real" database schema in prod vs what we create in dev/test.

PR #11218, opened 2022-04-20
Refactor: Migrate to 2.0-style security policies

### LINKED ISSUE #11211: Refactor Warehouse's Auth{N,Z} to use Pyramid 2.0-style security policies
Pyramid 2.x introduces a [refactored AuthN/AuthZ system](https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html#upgrading-auth-20), built around "security policies." These changes combine authentication and authorization into a single policy, and produce an "identity" that's more flexible than the concept of an authenticated user.

Why does Warehouse need this? The most immediate reason is that it simplifies the planned approach for #10970, which involves minting API tokens that aren't tied to any particular user. Because they aren't associated with a user identity, these tokens don't fit into the Pyramid 1.0 authentication scheme.

Here are some open things that need to be considered as part of this refactor:

* How are we going to handle multiple authentication policies? The current codebase uses [`pyramid_multiauth`](https://github.com/mozilla-services/pyramid_multiauth) to dispatch to multiple underlying providers, but that library doesn't support Pyramid 2.0-style policies yet. Tracking: https://github.com/mozilla-services/pyramid_multiauth/issues/28
* How much/how invasive will the `request.user` refactor be? That attribute currently pervades the codebase, and is currently assumed to be a `User` model (AFAICT). Should we remove that entirely, or keep it around as an `Optional[User]`?

Pinging @ewdurbin and @di for thoughts.

### LINKED ISSUE #11252: Don't flash warning on session expiration for token-based sessions
As raised in https://github.com/pypa/warehouse/pull/11218#discussion_r856926749, currently token-based sessions will get a flash message when the session expires, which is unnecessary.

woodruffw:
~~Very WIP; pushing up for visibility.~~

Closes #11211.

Closes #11252.


woodruffw:
This currently partially works (I can log in and navigate around), but permissions are currently broken for reasons I haven't fully debugged (trying to manage projects returns a `403`).

woodruffw:
This needs a lot of testing, but the basic structure is here:

* Each of session, basic, and macaroon-based authentication uses a "security policy" rather than 1.0-style AuthN APIs.
* We provide a new `MultiSecurityPolicy` that wraps multiple security policies in a manner similar to the now-removed `pyramid-multiauth`. One significant deviation: `MultiSecurityPolicy` also handles AuthZ, but it doesn't use each policy's separate AuthZ -- instead it keeps a singular AuthZ policy internal to itself, since Warehouse only has one.
* Our "identity" structure is now a dictionary with the following format:
    
    ```py
    { "entity": entity, "principals": principals }
    ```

...where `entity` can be `User` model or any other DB model suitable for a particular AuthN/AuthZ flow, and `principals` is the list of relevant security principles (previously a dedicated part of Pyramid's AuthZ APIs).

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+from zope.interface import implementer
+
+from warehouse.accounts.interfaces import IUserService
+
+
+class SecurityPolicy:
+    def __init__(self, callback=None):
+        self._callback = callback
+
+    def identity(self, request):
```
woodruffw, line 24: N.B. This is currently a generic `identity` for when policies always boil down to user identities. I'll probably need to make it less generic or remote it entirely when we allow for project identities.

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def remember(self, request, userid, **kw):
+        headers = []
+        for policy in self._policies:
+            headers.extend(policy.remember(request, userid, **kw))
+        return headers
+
+    def permits(self, request, context, permission):
+        identity = request.identity
+        principals = [Everyone]
```
woodruffw, line 90: Flagging this for review: I cargo-culted this from the Pyramid cookbook. I don't _think_ it causes access issues since `Everyone` means "everyone, all the time" and as such it's sort of a no-op check against most routes, but I'm not certain.

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+    * `forget`: Combined from all policies
+    * `remember`: Combined from all policies
+    * `permits`: Uses the AuthZ policy passed during initialization
+
+    These semantics mostly mirror those of `pyramid-multiauth`.
+    """
+
+    def __init__(self, policies, authz):
+        self._policies = policies
+        self._authz = authz
```
woodruffw, line 62: Also flagging this for design review: Warehouse only has one real AuthZ policy (well, multiple, but they're composed as a nested set), so I wasn't sure if it made sense for each security policy to keep its own copy of the AuthZ policy. Currently they don't and only the top-level `MultiSecurityPolicy` performs the `permits` check, but that's a bit of an abstraction violation.

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+        principals = []
+        if self._callback is not None:
+            principals = self._callback(user.id, request)
+            if principals is None:
+                return None
+
+        return {"entity": user, "principals": principals}
+
+    def permits(self, request, context, permission):
+        return NotImplemented
```
woodruffw, line 39: See https://github.com/pypa/warehouse/pull/11218/files#r854535588 for design review on this.

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+        if user is None:
+            return None
+
+        principals = []
+        if self._callback is not None:
+            principals = self._callback(user.id, request)
+            if principals is None:
+                return None
+
+        return {"entity": user, "principals": principals}
```
di, line 36: I think I'd prefer this to just return the ORM object that corresponds to the identity (here, the user). The computation of the principals can happen when we need it in `permits` instead, we don't need to pass it around, and we don't need principals available everywhere `request.identity` would be available.

I also think the `identity` function for `SecurityPolicy` should probably be `NotImplemented` and the subclass should be responsible for generating the corresponding identity instead. This means that `SessionSecurityPolicy` and `BasicAuthSecurityPolicy` could always return `Users`, but we could eventually update `MacaroonSecurityPolicy` to return other identities.

Since for now, they would all share the same logic for computing a user from a request, perhaps we turn this into a standalone function or mixin or something instead.

warehouse/utils/security_policy.py
di: I think this is OK, at least until we decide we want to overhaul our use of `principals` entirely.

(I think eventually `permits` would just determine outcomes based on the relationship (e.g. `Roles`) between `identity` and `context`, but that would be a big change that we don't need for this).

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,95 @@
...
+from warehouse.accounts.interfaces import IUserService
+
+
+class SecurityPolicy:
+    def __init__(self, callback=None):
+        self._callback = callback
+
+    def identity(self, request):
+        login_service = request.find_service(IUserService, context=None)
+        user = login_service.get_user(self.unauthenticated_userid(request))
```
di, line 26: This is also going to have to change once `MacaroonSecurityPolicy` is not always returning a `User`, so probably better to delegate this to the individual policies instead.

warehouse/accounts/__init__.py
```diff
@@ -206,19 +206,18 @@ def includeme(config):
...
     )
-    config.set_authorization_policy(
-        TwoFactorAuthorizationPolicy(
-            policy=MacaroonAuthorizationPolicy(policy=ACLAuthorizationPolicy())
+    config.set_security_policy(
+        MultiSecurityPolicy(
+            [
+                SessionSecurityPolicy(callback=_session_authenticate),
+                BasicAuthSecurityPolicy(check=_basic_auth_check),
+                MacaroonSecurityPolicy(callback=_macaroon_authenticate),
```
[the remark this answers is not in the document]
woodruffw: The only reason was "minimize code churn" 🙂 -- I could move them into their respective `security_policy.py` files and eliminate the callbacks if you don't mind a slightly bigger diff.

warehouse/utils/security_policy.py
woodruffw: Yeah, this is another case where I cargo-culted it from Pyramid's porting guide: they recommended having the identity contain the principles, but I agree that there's no reason for it to. I'll refactor so that it's just the ORM model.

warehouse/utils/security_policy.py
woodruffw: None that I can think of, they should be functionally equivalent. Will drop.

warehouse/accounts/__init__.py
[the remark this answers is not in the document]
di: I think that'd be preferable, would make it slightly easier to follow what's calling what.

warehouse/utils/security_policy.py
woodruffw: Oh, I remember why I did this now -- our AuthZ policies are still using the old APIs, which require the principals to be passed in explicitly. I couldn't figure out how else to get them into those APIs without stashing them on the `identity`, since the only context available is `request`, `context` (already set), and `permissions` (already set?)

warehouse/utils/security_policy.py
di: Not necessarily asking to drop it, just don't understand what it's purpose is. Maybe we should be doing it?

warehouse/utils/security_policy.py
di: I think they can be computed in `permits` right before we pass them to the AuthZ policy, no?

warehouse/utils/security_policy.py
woodruffw: If I understand how Pyramid does ACLs correctly, it's essentially the "no-op" principal -- every identity belongs to `Everyone`, including "no-one", so there's no difference between it being present and not.

That's per the docs:

> [](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/authorization.html#pyramid.authorization.Everyone)Everyone
The special principal id named Everyone. This principal id is granted to all requests. Its actual value is the string 'system.Everyone'.

warehouse/utils/security_policy.py
woodruffw: Yeah, I think you're right (I was overthinking it). `permits` can use `request.identity` as long as it's present, so we can do anything that the policy's `identity` would do anyways. I'll refactor!

woodruffw: [0b788d9](https://github.com/pypa/warehouse/pull/11218/commits/0b788d91af438990ebd4961eb070c0ce31af468e)

warehouse/accounts/__init__.py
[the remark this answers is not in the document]
woodruffw: Done. A lot of this boiled down to principal collection as well, so I was able to simply remove most of it and put the remainder in a `principals_for_authenticated_user` helper.

warehouse/accounts/security_policy.py
```diff
@@ -0,0 +1,238 @@
...
+    def identity(self, request):
+        # If we're calling into this API on a request, then we want to register
+        # a callback which will ensure that the response varies based on the
+        # Cookie header.
+        request.add_response_callback(add_vary_callback("Cookie"))
+        request.authentication_method = AuthenticationMethod.SESSION
+
+        # NOTE: Why is this authenticated_userid? I guess it's already
+        # been validated by session validation?
+        userid = self._session_helper.authenticated_userid(request)
```
woodruffw, line 120: Flagging for review: this is a Pyramid API so I'm assuming it's correct, but I'm a little bit confused by *how* a session gets authenticated. For basic auth or macaroons the logic is embedded in the `identity` body, but here we just seem to grab a `userid` from thin air. Is this helper using the (pre-validated?) session machinery internally?

warehouse/macaroons/security_policy.py
```diff
@@ -67,16 +68,18 @@ def _extract_http_macaroon(request):
...
+    def identity(self, request):
         # If we're calling into this API on a request, then we want to register
         # a callback which will ensure that the response varies based on the
         # Authorization header.
         request.add_response_callback(add_vary_callback("Authorization"))
+        request.authentication_method = AuthenticationMethod.MACAROON
+
+        # Macaroon authentication can only be used for uploading
+        if request.matched_route.name not in ["forklift.legacy.file_upload"]:
+            return False
```
woodruffw, line 82: Flagging this for review as well: we didn't actually have this before, which makes me think that our authn/z for macaroons might have been a little too permissive (we were probably allowing anything with `context=Project` that matched the token's actual permissions).

warehouse/utils/security_policy.py
woodruffw: Removed it and nothing broke.

warehouse/utils/security_policy.py
woodruffw: Removed this and fully delegated.

warehouse/macaroons/security_policy.py
di: Yes, this is known: https://github.com/pypa/warehouse/issues/7266

di: (I'm also not convinced that we need this, at least in this PR).

warehouse/accounts/security_policy.py
dstufft: There's two APIs here:

1. ``unauthenticated_userid`` which is intended to pull the userid out of the request.
2. ``authenticated_userid`` which is intended to consume ``unauthenticated_userid`` and ensure that it matches what is in the data store.

For sessions, the data store is the session key, and when ``remember`` is called, the user id is stored in the user's Session, then ``authenticated_userid`` pulls that userid out of the session. Since session storage doesn't have passwords or anything like that and relies on the session id being randomly generated and secret, it just accepts whatever is in the session storage as the authenticated user id.

warehouse/utils/security_policy.py
woodruffw: Removed.

warehouse/utils/security_policy.py
woodruffw: Removed.

warehouse/macaroons/security_policy.py
woodruffw: Totally missed that check! I can remove this route check then, since it's a little out of place in the AuthN section.

warehouse/accounts/security_policy.py
woodruffw: Makes sense, thanks for the explanation. For some reason I forgot that session tokens are bearers and that that's the trust relationship.

warehouse/macaroons/security_policy.py
woodruffw: [0bc2083](https://github.com/pypa/warehouse/pull/11218/commits/0bc208360ef2f1f94146e35396e4ef981a4b77c1)

warehouse/macaroons/security_policy.py
```diff
@@ -90,16 +89,18 @@ def unauthenticated_userid(self, request):
         # fetch the user that is associated with it.
         macaroon_service = request.find_service(IMacaroonService, context=None)
         userid = macaroon_service.find_userid(macaroon)
-        if userid is not None:
-            return str(userid)
+
+        # TODO: This will return either a user or a project soon.
+        login_service = request.find_service(IUserService, context=None)
+        return login_service.get_user(userid)
```
ewjoachim, line 95: `macaroon_service.find_userid` extracts the macaroon ID from the macaroon and then fetches the macaroon with the ORM joining with the user table, then extracts the user id from the user, and discards both the macaroon instance and the user instance. And then you make a second request to get the user from its id.

You'd gain a DB request by doing 
```python
        user = macaroon_service.find_from_raw(macaroon).user
```

(also, I wonder why there's no exception handling here but there wasn't before either)

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,120 @@
...
+    login_service = request.find_service(IUserService, context=None)
+    user = login_service.get_user(userid)
+
+    if user is None:
+        return
+
+    if request.session.password_outdated(login_service.get_password_timestamp(userid)):
+        request.session.invalidate()
+        request.session.flash("Session invalidated by password change", queue="error")
+        return
```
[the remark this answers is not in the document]
di: I believe you're right, but since that's not a bug introduced by this PR I've made a new issue to capture it: https://github.com/pypa/warehouse/issues/11252

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,120 @@
...
+            principals.extend([Authenticated, str(identity.id)])
+
+            if isinstance(identity, User):
+                principals.extend(
+                    principals_for_authenticated_user(identity.id, request)
+                )
+            elif isinstance(identity, Project):
+                return Denied("unimplemented")
+
+        return self._authz.permits(context, principals, permission)
```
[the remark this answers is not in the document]
woodruffw: Yeah, it's extremely confusing -- it's because we're really wrapping the old-style AuthZ APIs under the hood here. I'll add a comment that explains the behavior a bit.

warehouse/accounts/security_policy.py
```diff
@@ -0,0 +1,238 @@
...
+def _basic_auth_check(username, password, request):
+    # Basic authentication can only be used for uploading
+    if request.matched_route.name not in ["forklift.legacy.file_upload"]:
+        return False
+
+    login_service = request.find_service(IUserService, context=None)
+    breach_service = request.find_service(IPasswordBreachedService, context=None)
+
+    userid = login_service.find_userid(username)
+    if userid is not None:
```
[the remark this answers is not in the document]
di: Good point. I'd be in favor of making this change after this PR is merged.

warehouse/accounts/security_policy.py
```diff
@@ -0,0 +1,238 @@
...
+            # mechanisms (since we wouldn't have fell through to them anyways).
+            if disabled_for == DisableReason.CompromisedPassword:
+                raise _format_exc_status(
+                    BasicAuthBreachedPassword(), breach_service.failure_message_plain
+                )
+            elif disabled_for == DisableReason.AccountFrozen:
+                raise _format_exc_status(BasicAuthAccountFrozen(), "Account is frozen.")
+            else:
+                raise _format_exc_status(HTTPUnauthorized(), "Account is disabled.")
+        elif login_service.check_password(
```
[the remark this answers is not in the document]
di: Ditto here.

warehouse/accounts/__init__.py
```diff
@@ -206,19 +76,18 @@ def includeme(config):
...
-        MultiAuthenticationPolicy(
-            [
-                SessionAuthenticationPolicy(callback=_session_authenticate),
-                BasicAuthAuthenticationPolicy(check=_basic_auth_check),
-                MacaroonAuthenticationPolicy(callback=_macaroon_authenticate),
-            ]
-        )
+    # Register our security policies (AuthN + AuthZ)
+    authz_policy = TwoFactorAuthorizationPolicy(
+        policy=MacaroonAuthorizationPolicy(policy=ACLAuthorizationPolicy())
```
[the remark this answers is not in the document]
woodruffw: If the goal is to completely switch over to 2.0-style policies, we should probably eventually fold the big callback tower into a single AuthZ policy "helper", since that's the pattern that Pyramid itself uses for e.g. basic and session-based auth 🙂

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,120 @@
...
+        for policy in self._policies:
+            headers.extend(policy.remember(request, userid, **kw))
+        return headers
+
+    def permits(self, request, context, permission):
+        identity = request.identity
+        principals = []
+        if identity is not None:
+            # Our identity is always an ORM model, so add some generic principals first.
+            principals.extend([Authenticated, str(identity.id)])
```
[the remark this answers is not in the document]
woodruffw: > Writing the id in the principals seems slightly dangerous to me because there could be a confusion if a user and a project had the same id. I know it's very very unlikely due to both ids being UUIDs, but I don't think it's a good idea to receive an ID and to have to guess what model it belongs to.

I might be misunderstanding what you mean, but we should never actually have to guess the model -- `request.identity` will always be the correct ORM model, so the `id` here is solely for permission matching.

(That being said, your suggestion re: `f"user:{user.id}"` sounds reasonable to me, if we do decide that raw IDs are ambiguous.)

warehouse/macaroons/security_policy.py
di: This isn't going to be around very long, as we'll need a way to get either the User or the Project from the macaroon, but I'd be fine with making this change for now to make this slightly more efficient.

warehouse/accounts/security_policy.py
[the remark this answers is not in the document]
woodruffw: Yep, agreed. I'll create a rollup issue tracking these kinds of post-merge refactors.

warehouse/utils/security_policy.py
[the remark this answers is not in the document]
di: I think what @ewjoachim is that it's possible (but unlikely) for an `id` to be shared by both a `User` and a `Project`, which I think is correct.

I'm not seeing why we're doing this though, were we doing it before and I'm just missing it? Or is this something taken from the new 2.0-style examples? (I'm not sure it's necessary).

warehouse/accounts/security_policy.py
```diff
@@ -0,0 +1,238 @@
...
+        login_service = request.find_service(IUserService, context=None)
+        return login_service.get_user_by_username(username)
+
+    def forget(self, request, **kw):
+        # No-op.
+        return []
+
+    def remember(self, request, userid, **kw):
+        # NOTE: We could make realm configurable here.
+        return [("WWW-Authenticate", 'Basic realm="Realm"')]
```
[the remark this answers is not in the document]
woodruffw: I cribbed this directly from the previous (Pyramid 1.0-style) authn policy: we were already implicitly returning this header, so I figured we should continue to do so on the (exceedingly unlikely) chance that someone is actually expecting it.

> I'm especially doubtful given that our MultiAuthSecurityPolicy calls remember on every subpolicies when we call its remember method, so as far as I can tell, this would result in this header be sent back to the browser when we log in ? And I guess the reason the browser doesn't display an auth modal is because it's not accompanied by a 401 Unauthenticated ?

Yep, correct. The previous (`pyramid_multiauth`-based) implementation also did this, mind you: https://github.com/mozilla-services/pyramid_multiauth/blob/e3d2f5d75201b05daefd0c589f44fb55fdf8e877/pyramid_multiauth/__init__.py#L137-L146

(I think it's pretty ugly that we always return all `remember` and `forget` headers, but I'm not sure there's a straightforward way to do that while composing AuthN policies in the way we are.)

warehouse/accounts/security_policy.py
```diff
@@ -0,0 +1,238 @@
...
+from warehouse.packaging.models import TwoFactorRequireable
+from warehouse.utils.security_policy import AuthenticationMethod
+
+
+def _format_exc_status(exc, message):
+    exc.status = f"{exc.status_code} {message}"
+    return exc
+
+
+def _basic_auth_check(username, password, request):
```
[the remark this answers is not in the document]
woodruffw: I think you're right about this. I'll include it in the refactor rollup.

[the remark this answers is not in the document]
woodruffw: https://github.com/pypa/warehouse/issues/11253

warehouse/accounts/security_policy.py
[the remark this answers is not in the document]
woodruffw: https://github.com/pypa/warehouse/issues/11253

warehouse/accounts/security_policy.py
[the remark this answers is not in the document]
woodruffw: https://github.com/pypa/warehouse/issues/11253

warehouse/macaroons/security_policy.py
woodruffw: 5cdb53a6

warehouse/utils/security_policy.py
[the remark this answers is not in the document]
woodruffw: Yeah, it's another thing I cargo-culted from the upgrade guide. I can probably remove it entirely with nothing breaking, will experiment now.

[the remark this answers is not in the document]
woodruffw: Hmm. I tried to remove this and things *appear* to work, but package upload with HTTP basic auth always 403s without it. Any thoughts?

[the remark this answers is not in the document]
dstufft: https://github.com/pypa/warehouse/blob/main/warehouse/packaging/models.py#L217-L236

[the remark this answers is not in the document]
woodruffw: Thanks @dstufft! That's an easy fix.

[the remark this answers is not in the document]
di: Yep, that's it. So I think @ewjoachim's idea to prefix this with `user:` makes sense.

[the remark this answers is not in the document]
woodruffw: @dstufft do you happen to know how the original code was placing the user ID in the principals list? I'm having trouble finding it.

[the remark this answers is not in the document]
dstufft: https://github.com/pypa/warehouse/blob/main/warehouse/accounts/__init__.py#L66-L92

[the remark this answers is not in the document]
dstufft: I don't remember why I used strings there instead of ORM objects, there might have been a reason, but it might also work just fine to use objects. I think probably the examples all used strings, and I didn't feel like digging into it to figure out if not-strings would introduce some horrible security bug.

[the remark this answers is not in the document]
woodruffw: I basically preserved that function as `_principals_for_authenticated_user`, but I don't actually see the user's ID getting added to the principal set there. Unless I'm missing something?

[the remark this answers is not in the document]
dstufft: Oh right. Sorry, the ``request.effective_principals`` in Pyramid pre 2.0 (no idea what happens after that) gets populated by the authentication policy's ``effective_principals`` method. Both the Session and the Basic auth authN policy that ships (shipped?) with Pyramid used the ``CallbackAuthenticationPolicy``, which looks like:

https://github.com/Pylons/pyramid/blob/1.9-branch/pyramid/authentication.py#L107-L184

Which puts the special "Everyone" ID, the string of the user id, and then anything that comes from the "groups" callback, which is our ``_authenticate`` method.

It looks like the Macaroon policy also uses the ``CallbackAuthenticationPolicy``, so it will also do that.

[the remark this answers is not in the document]
woodruffw: Mystery solved! 😄 

Yeah, we've gotten rid of the callbacks entirely (since they're not part of the 2.0 APIs), so now everything is concentrated into that one `permits` method that dispatches into the underlying stack of AuthZ policies.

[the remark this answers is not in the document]
dstufft: Assuming something like this still exists in Pyramid 2.0, we could just ignore the fact there is a string-ified userid in the principals, and make the ``_authenticate`` method also return a ``f"{model}:{id}"`` prinicpal, and actually use that in our ACLs OR we can override some methods to make our own ``CallbackAuthenticationPolicy`` that has the behavior we want.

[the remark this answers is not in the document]
dstufft: Ah cool, that's easier then.

woodruffw:
I need to do a final round of local testing, but this should be nearly good to go.

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _principals_for_authenticated_user(user, request):
+    """Apply the necessary principals to the authenticated user"""
+    login_service = request.find_service(IUserService, context=None)
+
+    if request.session.password_outdated(login_service.get_password_timestamp(user.id)):
+        request.session.invalidate()
+        request.session.flash("Session invalidated by password change", queue="error")
+        return None
```
woodruffw, line 37: This was part of the original code so I haven't moved it, but it's oddly placed here -- it's really part of AuthN, not the AuthZ work we're doing here. Any objections to me moving it into the `identity`/AuthN phase instead?

woodruffw: Done with [29b40f9](https://github.com/pypa/warehouse/pull/11218/commits/29b40f969b7f5191b89189f82b061854295226cf)

ewjoachim (approved):

di (approved):
This looks great. Leaving this open for a little while to give @ewdurbin and @dstufft a chance to review if they are able.

woodruffw:
This is the main blocker for #11272, which will "finish" the core functionality here (allowing a user to go from a JWT to an ephemeral API token).

warehouse/utils/security_policy.py
```diff
@@ -0,0 +1,109 @@
...
+        for policy in self._policies:
+            if ident := policy.identity(request):
+                return ident
+
+        return None
+
+    def authenticated_userid(self, request):
+        if request.identity:
+            return str(request.identity.id)
+        return None
```
di, line 78: This is fine for now but is going to be problematic once we have non-user identities, since these won't have a prefix like the principals would. We might want to address that here in advance, e.g. by asserting that the identity is a user  and otherwise returning `None` or something similar.

di (approved):
This is ready to merge once this last comment is addressed.

warehouse/utils/security_policy.py
woodruffw: Yeah, I'll make it so that only `User` identities return an ID here. That'll be consistent with our expected behavior, i.e. that `request.user` will only ever be populated with `User` models.

woodruffw: [ec2c563](https://github.com/pypa/warehouse/pull/11218/commits/ec2c563d4003556435664911a23364c7bac20197)

di (approved):

PR #11239, opened 2022-04-22
OIDC: More claims for GitHub's provider

woodruffw:
GitHub has added claims that allow us to attest to the repository owner's stable ID, which in turn allows us to prevent "account resurrection"-style attacks.

Needs test updates, but this should be functionally complete. The migration is primarily to make the comparisons during claim verification more straightforward.

### DIFF
--- a/warehouse/migrations/versions/bb986a64761a_rename_githubprovider_fields.py
+++ b/warehouse/migrations/versions/bb986a64761a_rename_githubprovider_fields.py
@@ -1,0 +1,67 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+rename GitHubProvider fields
+
+Revision ID: bb986a64761a
+Revises: 9f0f99509d92
+Create Date: 2022-04-22 22:00:53.832695
+"""
+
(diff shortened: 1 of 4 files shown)


woodruffw:
Ref: https://github.com/pypa/warehouse/issues/10645#issuecomment-1106910136

warehouse/migrations/versions/bb986a64761a_rename_githubprovider_fields.py
```diff
@@ -0,0 +1,67 @@
...
+    op.drop_constraint(
+        "_github_oidc_provider_uc", "github_oidc_providers", type_="unique"
+    )
+    op.create_unique_constraint(
+        "_github_oidc_provider_uc",
+        "github_oidc_providers",
+        ["repository_name", "repository_owner", "workflow_filename"],
+    )
+    op.drop_column("github_oidc_providers", "owner_id")
+    op.drop_column("github_oidc_providers", "owner")
```
di, line 46: Doesn't actually matter for the production DB right now because there are no rows in this table, but probably better to make this a rename instead. Generally we don't want to drop/add columns.

di (approved):
LGTM aside from a bit of cleanup that should happen in the migration.

warehouse/migrations/versions/bb986a64761a_rename_githubprovider_fields.py
woodruffw: Good to know, will tweak.

woodruffw: (I wish alembic would catch these renames, but thinking about it more that wouldn't be possible in the general case...)

woodruffw: [dddb25d](https://github.com/pypa/warehouse/pull/11239/commits/dddb25df9df8ba89318f165148f6e727b75da1f9)

woodruffw:
Refactored the migration to use renames 🙂

di (approved):

Issue #11096, opened 2022-04-01
Support for JWTs from reusable workflows


woodruffw:
Our MVP implementation (#10753) assumes that the workflow is in the same repository, which is [not necessarily true](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows).

We should support reusable workflows, specifically via the `job_workflow_ref` claim.

woodruffw:
Just as a note: this will have to interoperate with #11263.

GergelyKalmar:
Any chance to expedite this issue? It seems there's many who would be eager to move over to trusted publishing now that it is the recommended approach, but not being able to do the flow from a reusable workflow is a pretty major blocker. This issue also does not seem to be documented anywhere so we only find out about it once we changed everything to use the new approach.

woodruffw:
> Any chance to expedite this issue? 

I can't personally promise a timeline here, but it is on my backlog of things to do.

> This issue also does not seem to be documented anywhere so we only find out about it once we changed everything to use the new approach.

Thank you for pointing this out; I'll make some changes to the docs to emphasize that reusable workflows are not currently supported.

woodruffw:
> Thank you for pointing this out; I'll make some changes to the docs to emphasize that reusable workflows are not currently supported.

I've opened https://github.com/pypi/warehouse/pull/13592 for this. Thanks again for bringing it to our attention!

webknjaz:
@woodruffw I just stumbled upon an example of a reusable workflow actually _working_ with OIDC here https://github.com/jorisroovers/gitlint/pull/486 — could you confirm this is because of `secrets: inherit` at https://github.com/jorisroovers/gitlint/blob/5247839/.github/workflows/ci.yml#L225 ?

woodruffw:
Yeah, I think that's because it's in the same repo, not because of `secrets: inherit`.

In other words: reusable workflows *do* work with the current implementation, just not reusable workflows that are external to the repository that the trusted publisher was configured with.

webknjaz:
> That workflow is in the same repository though, I think the issue is with reusable workflows that are in a _different_ repository (e.g. in a separate public repo).

Yes, you can't use `inherit` across orgs. The issue was that the OIDC context has a reusable workflow reference and that's currently not checked IIUC. So my concern is that it might be possible to fool the verification logic in PyPI with this...

webknjaz:
I also wonder if it's possible to do something like
```yaml
[...]

Issue #10644, opened 2022-01-25
OIDC: Threat model for Warehouse


woodruffw:
This is a summary of my own notes from conversations with @di concerning Warehouse's threat model around OIDC (and GitHub, in particular, as an OIDC provider).

### JWT reuse

Warehouse needs well-defined and well-documented behavior around handling of a JWT seen more than once. For example, a GitHub action might do the following (pseudocode):

```
mint a JWT

for each dist in dist/* {
    use the JWT to mint an access token
    upload dist with the access token
}
```

In this case, the JWT gets used `N` times: once for each access token minting. This is probably **not** what we want; instead, we want this:

```
mint a JWT
mint an access token using the JWT

for each dist in dist/* { ... }
```

In this use pattern, we invalidate the JWT in Warehouse's backend after its first use. The process for that is probably as simple as performing uniqing on the `jti` claim after JWT verification, and then adding `(jti, exp)` to some table, where `exp` is the JWT's expiration claim (which gives us the ability to automatically clear out old JWTs once they expire).

### Access token ephemerality

We should determine exactly how long-lived our ephemeral access tokens/API keys will be: how long is a reasonable period to allow uploads for? Should we cap the number of individual requests at *some* high number (e.g., is there any legitimate packaging workflow that creates more than 100 separate distribution files and uploads them)?

### GitHub: account resurrection/reuse

GitHub allows a deleted user's username to be reused. When this happens, JWTs minted by the new user are indistinguishable from JWTs minted by the old one. This is a potential problem if PyPI is configured to accept JWTs from `user/repo @ workflow`, where `user` changes hands on GitHub -- the new (malicious) `user` would still be able to authenticate with PyPI as if they were the old `user`.

We probably need a mitigation for this before we fully enable support for GitHub as an OIDC provider. Two potential solutions:

1. Verify the `repository_owner` claim in the JWT with some sidecar information: on trusted setup, retrieve the GitHub User ID (which is hopefully unique) for the user and cross-check it against the current ID. **This is not the preferred solution.**
2. Get GitHub to include a `repository_owner_id` claim in the JWT, or something similar (like a `repository_owner_epoch`, if they don't want to guarantee the ID's uniqueness). 

Neither of these is a great solution, because both still require some amount of sidecar state: we still have to either initially store the GitHub user's ID during trusted setup or during TOFU, so that we can check for changes during subsequent authentications. That complicates our DB schema, which we'd like to keep generic. 

[...]

PR #11263, opened 2022-04-27
GitHub OIDC: validate `job_workflow_ref`

woodruffw:
Also adds a bunch of counterexample tests to be certain.


woodruffw:
This is more or less the same as the original implementation, except that we now use a strict regex (`re.match`) instead of `str.startswith` as our predicate. That should eliminate the possibility for potential prefix confusion by a malicious user, although said confusion would still be outside of our threat model (since it assumes CI compromise).

tests/unit/oidc/test_models.py
```diff
@@ -132,3 +133,26 @@ def test_github_provider_verifies(self, monkeypatch):
         }
         assert provider.verify_claims(signed_claims=signed_claims)
         assert len(noop_check.calls) == len(verifiable_claims)
+
+    @pytest.mark.parametrize(
+        ("claim", "valid"),
+        [
+            ("foo/bar/.github/workflows/baz.yml@", True),
```
di, line 140: Isn't this going to be something like `foo/bar/.github/workflows/baz.yml@refs/pull/6/merge`?

woodruffw: Whoops, you're absolutely right. I need to tweak it to ensure that there's data following the `@`, including at least one `/` (to prevent the phony filename problem).

woodruffw: [798a29f](https://github.com/pypa/warehouse/pull/11263/commits/798a29ffbbb3e716ffadfee430eadd88235d4c0b)

ewjoachim: As far as I can tell, I'm not sure this addresses the issue of reusable workflows, since it checks that what's in that string is from the same repo as the main repo we're validating.

I think we can't do it properly without a new db field because the information is independant.

Also, what happens if there's an `@` in the branch/tag name ? I'm starting to think that it may not be possible to parse this without injecting knowledge. Is the `ref` part passed elsewhere ?

ewjoachim:
Hm, I've left a comment and github is not showing it to me. I wonder if github lost my comment. Let me know if this is the first comment you see of me in this PR.

Anyway, the discussion we had was like this: https://github.com/pypa/warehouse/pull/10753#discussion_r833505827

And I'm not seeing the same things as we discussed, but maybe I'm misunderstanding the goal of this PR

tests/unit/oidc/test_models.py
di: We've decided (short term) not to support reusable workflows, but long term we will want to support them.

di (approved):
This LGTM but want to make sure @ewjoachim's concerns have been addressed before merging.

tests/unit/oidc/test_models.py
woodruffw: Yeah, that long term goal is captured in #11096.

ewjoachim:
Ok so there's one question left from my other comment:

> What happens if there's an @ in the branch/tag name ?

I believe this will confuse rsplit.

woodruffw:
> Ok so there's one question left from my other comment:
> 
> > What happens if there's an @ in the branch/tag name ?
> 
> I believe this will confuse rsplit.

Yeah, I think you're right. I don't have a clean solution for this, then -- we could do an unambiguous parse by referencing the `ref` claim for the suffix, but that causes an abstraction leak (claim verification depending on other claims as inputs) that makes our verification step much more manual (and therefore user-error prone).

All of this is technically outside of our threat model (#10644), since the attack here assumes that the attacker manages to fool the repository owner into introducing a malicious workflow. Thoughts on just reverting back to the `startswith` or regex check?

woodruffw:
Okay, this should do the trick: we're now using the `ref` claim to complement the `job_workflow_ref` check. This prevents any sort of prefix or suffix sneakiness, since the verification is now an exact string comparison for the pattern:

```
{workflow_path}@{ref}
```

...where `{workflow_path}` is computed solely from trusted setup data and `{ref}` is a verified claim containing a git-style ref.

di (approved):
Will let @ewjoachim re-review as well.

ewjoachim (approved):
Haha, I'm sorry for taking nice code away and forcing you to write less nice code :D (blame GitHub for mixing parts of strings together in a bigger string without any kind of escaping and expecting us to parse the bits ourselves)

Seems good now :)

```console
$ python -m this | grep amb
In the face of ambiguity, refuse the temptation to guess.
```

woodruffw:
> Haha, I'm sorry for taking nice code away and forcing you to write less nice code :D (blame GitHub for mixing parts of strings together in a bigger string without any kind of escaping and expecting us to parse the bits ourselves)

It's greatly appreciated! Precision is a virtue, especially in anything remotely involving security 🙂

Issue #9524, opened 2021-05-13
Delete Project Scoped API Tokens When the Project is Deleted


Julian:
**What's the problem this feature will solve?**
On the rare case that a project is deleted entirely from PyPI, a user's scoped API token(s) to that project still remain present.

**Describe the solution you'd like**
Once the project is deleted, remove the API tokens automatically.

**Additional context**
I just noticed this when performing said "rare case" -- I'm not sure it's a bug since I can't think of anything that goes particularly bad myself, so this may be quite minor and just unsightly (which is why I marked it a feature request), but passing it along nonetheless.

ewjoachim:
Just to clarify, in case someone would be wondering: if someone else would be to create a new project under that name, the existing token wouldn't automatically give the old token bearer the capacity to upload packages on this new project. Project scoped tokens still require their owner to have upload right on the uploading project.

While we're at it, I guess we may also want to remove one's project scoped tokens if that person is removed from the project by other owners.

I'm noting though that this would require either of the following:
- Filtering token JSONB caveats field through SQL, which would probably not be ideal in conjunction with #9264 where we try to decouple the caveat format information from the warehouse code.
- Using Python to analyze all the tokens someone owns when their right on a project disappear (through deletion or role removal) to find the project-scoped tokens (let's hope no one has multiple thousands of tokens, but if they did, I'm guessing this specific check would be minor compared to simply displaying their project page). Still, it would probably be the best solution. It might be interesting to do it in a celery task.

woodruffw:
> * Using Python to analyze all the tokens someone owns when their right on a project disappear (through deletion or role removal) to find the project-scoped tokens (let's hope no one has multiple thousands of tokens, but if they did, I'm guessing this specific check would be minor compared to simply displaying their project page). Still, it would probably be the best solution. It might be interesting to do it in a celery task.


This is the cleanest solution, but #11272 will throw a wrench in it (since it'll make users optional on macaroons). In other words, we won't be able to rely on role changes on users as a reliable trigger for when to purge a macaroon.

Ultimately, we may have to go with a SQL query that pokes into the JSON caveats.

PR #11272, opened 2022-04-27
OIDC macaroon minting for GitHub

### LINKED ISSUE #10970: Routes and endpoints for JWT consumption
Once we allow projects to configure OIDC providers (termed "publishers" in user-facing views), we need to allow OIDC-minted JWTs to be exchanged for API tokens.

As discussed with @di, we should ensure that these can be configured to run on a separate subdomain during deployment, to simplify caching logic.

woodruffw:
This adds an API route for converting a JWT into a temporary API token.

~~Very WIP.~~

TODO:

- [x] Actually create the macaroon
- [x] Local testing
- [x] Fill in test coverage
- [x] Feature gate with the other OIDC functionality

~Needs #11218.~

Closes #10970.


warehouse/macaroons/models.py
```diff
@@ -37,15 +38,29 @@ class Macaroon(db.Model):
         UniqueConstraint(
             "description", "user_id", name="_user_macaroons_description_uc"
         ),
+        UniqueConstraint(
+            "description", "project_id", name="_project_macaroons_description_uc"
+        ),
+        CheckConstraint(
+            "(user_id::text IS NULL) <> (project_id::text IS NULL)",
+            name="_user_xor_project_macaroon",
+        ),
```
woodruffw, line 47: cc @di: This enforces the XOR of "every Macaroon is either user-associated or project-associated." Let me know what you think about this approach.

woodruffw:
This is ready for an initial review.

warehouse/forklift/legacy.py
```diff
@@ -807,24 +807,27 @@ def file_upload(request):
...
+                HTTPBadRequest,
+                (
+                    "User {!r} does not have a verified primary email address. "
+                    "Please add a verified primary email before attempting to "
+                    "upload to PyPI. See {project_help} for more information."
+                ).format(
+                    request.user.username,
+                    project_help=request.help_url(_anchor="verified-email"),
+                ),
+            ) from None
```
woodruffw, line 830: Flagging for review: the entire upload endpoint assumes that the request's identity is a user. I've handled most of these by checking whether `request.user is None` (with the precondition that `request.identity is not None`).

warehouse/forklift/legacy.py
```diff
@@ -1169,28 +1200,33 @@ def file_upload(request):
...
+            # TODO: This should be handled by some sort of database trigger or
+            #       a SQLAlchemy hook or the like instead of doing it inline in
+            #       this view.
+            request.db.add(
+                JournalEntry(
+                    name=release.project.name,
+                    version=release.version,
+                    action="new release",
+                    submitted_by=request.user,
+                    submitted_from=request.remote_addr,
```
woodruffw, line 1218: NTS: I'll relax `JournalEntry.submitted_by` to make it nullable, so that we can re-enable this journal entry for all request identities.

warehouse/forklift/legacy.py
```diff
@@ -1169,28 +1200,33 @@ def file_upload(request):
...
+                )
             )
-        )
 
         project.record_event(
             tag="project:release:add",
             ip_address=request.remote_addr,
             additional={
-                "submitted_by": request.user.username,
+                "ephemeral": request.user is None,
```
woodruffw, line 1226: Flagging: It might make sense for this to be `nonuser: bool` instead, since `request.user is None` might not always imply an OIDC ephemeral token in the future.

warehouse/forklift/legacy.py
```diff
@@ -1429,20 +1465,21 @@ def file_upload(request):
...
+            #       view.
+            request.db.add(
+                JournalEntry(
+                    name=release.project.name,
+                    version=release.version,
+                    action="add {python_version} file {filename}".format(
+                        python_version=file_.python_version, filename=file_.filename
+                    ),
+                    submitted_by=request.user,
+                    submitted_from=request.remote_addr,
```
woodruffw, line 1480: NTS: Same as https://github.com/pypa/warehouse/pull/11272/files#r870683888

warehouse/macaroons/models.py
```diff
@@ -37,13 +38,27 @@ class Macaroon(db.Model):
...
+    #   additional project scope restrictions as part of its caveats.
+    # * In the project case, a Macaroon has an associated project, and
+    #   is scoped to just that project.
     user_id = Column(
-        UUID(as_uuid=True), ForeignKey("users.id"), nullable=False, index=True
+        UUID(as_uuid=True), ForeignKey("users.id"), nullable=True, index=True
+    )
+
+    project_id = Column(
+        UUID(as_uuid=True), ForeignKey("projects.id"), nullable=True, index=True
```
woodruffw, line 61: Flagging: I'm not sure if `index=True` is necessary here. I just added it because we have it for the `user_id` FK.

warehouse/macaroons/security_policy.py
```diff
@@ -88,13 +87,17 @@ def identity(self, request):
...
+        except InvalidMacaroonError:
             return None
 
-        user_service = request.find_service(IUserService, context=None)
-        return user_service.get_user(userid)
+        # Every Macaroon has either a user or a project.
+        if dm.user is not None:
+            return dm.user
+        else:
+            return dm.project
```
woodruffw, line 100: NB: I could refactor this with a `@property` on `Macaroon`, e.g.:

```py
@property
def identity(self):
    if self.user is not None:
        return self.user
    return self.project
```

warehouse/oidc/services.py
```diff
@@ -23,6 +24,24 @@
...
+    def __init__(self, provider, issuer_url, cache_url, metrics):
+        warnings.warn(
+            "NullOIDCProviderService is intended only for use in development, "
+            "you should not use it in production due to the lack of actual "
+            "JWT verification.",
+            InsecureOIDCProviderWarning,
+        )
+
+    def verify_for_project(self, token, project):
+        return True
```
woodruffw, line 42: NB: I'm going to change this so that it actually checks `project.oidc_providers`. It still won't do the actual JWT signature verification, however.

warehouse/packaging/models.py
```diff
@@ -221,6 +225,10 @@ def __acl__(self):
             (Allow, "group:moderators", "moderator"),
         ]
 
+        # The project itself has an identity (for OIDC-minted tokens),
+        # and that identity has the ability to upload releases.
+        acls.append((Allow, f"project:{self.id}", ["upload"]))
```
woodruffw, line 230: Flagging: This is how we give each `Project` identity permission to upload to itself. I'm doing this the same way we do `User` identities, but just flagging it in case I'm missing anything.

woodruffw:
Here's what journal events from non-users currently look like:

Screen Shot 2022-05-16 at 1 48 28 PM

warehouse/macaroons/services.py
```diff
@@ -54,17 +54,12 @@ def find_macaroon(self, macaroon_id):
...
-            )
-        except NoResultFound:
-            return None
-
-        return dm
+        return (
+            self.db.query(Macaroon)
+            .options(joinedload("user"))
+            .options(joinedload("project"))
+            .get(macaroon_id)
```
di, line 61: This is deprecated, see https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-orm-usage

warehouse/macaroons/services.py
```diff
@@ -95,11 +90,16 @@ def find_userid(self, raw_macaroon):
         if dm is None:
             return None
 
+        # This can be None if the macaroon has no associated user
+        # (e.g., an OIDC-minted macaroon).
+        if dm.user is None:
+            return None
+
```
di, line 97: I wonder if `find_userid` would be better served as something like `find_identity`, which could then let the user do `identity.id` with the result.

warehouse/macaroons/services.py
```diff
@@ -127,20 +127,33 @@ def verify(self, raw_macaroon, context, principals, permission):
...
+        Either a user ID or a project ID (but not both) is required.
         """
-        user = self.db.query(User).filter(User.id == user_id).one()
+        user = None
+        if user_id is not None:
+            user = self.db.query(User).get(user_id)
+
+        project = None
+        if project_id is not None:
+            project = self.db.query(Project).get(project_id)
```
di, line 146: Is it actually necessary to query for the actual instance here? Or could we just initialize a `Macaroon` with the ID?

warehouse/oidc/__init__.py
```diff
@@ -15,10 +15,22 @@
...
         ),
         IOIDCProviderService,
         name="github",
     )
+
+    # During deployments, we separate auth routes into their own subdomain
+    # to simplify caching exclusion.
+    auth = config.get_settings().get("auth.domain")
+
+    config.add_route("oidc.mint_token", "/api/oidc/mint-token", domain=auth)
```
di, line 36: I'm not 100% sure this is where I want this route to live -- I don't think we have anything else at `/api`.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,86 @@
...
+    route_name="oidc.mint_token",
+    require_methods=["POST"],
+    renderer="json",
+    require_csrf=False,
+    has_translations=False,
+)
+def mint_token_from_oidc(request):
+    def _invalid(msg):
+        request.response.status = 422
+        return {"error": msg}
```
di, line 33: It might be nice for these to both have something short and unique to key an error off of, as well as a longer human-readable explanation of why this failed.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,86 @@
...
+    except ValueError:
+        return _invalid("missing body")
+
+    unverified_jwt = body.get("token")
+    if not unverified_jwt:
+        return _invalid("missing token")
+
+    project_name = body.get("project")
+    if not project_name:
+        return _invalid("missing project")
```
di, line 46: I'm a little confused why the body can't just be the OIDC token, and why we need a separate field for the project. Can't that just be inferred from the token based on the provider?

warehouse/oidc/views.py
```diff
@@ -0,0 +1,86 @@
...
+    now = time.time()
+    expires = int(now) + 900
+    caveats = [
+        {"permissions": {"projects": [project.normalized_name]}, "version": 1},
+        {"nbf": int(now), "exp": expires},
+    ]
+    serialized, dm = macaroon_service.create_macaroon(
+        location=request.domain,
+        project_id=project.id,
+        description=f"OpenID created ephemeral token ({now})",
```
di, line 75: Something other than 'ephemeral' here.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,86 @@
...
+        caveats=caveats,
+    )
```
di, line 77: Let's find a way to serialize the provider ID or some other identifier for the provider into the macaroon here

warehouse/oidc/views.py
woodruffw: This is because of our original data model decision: we decided that multiple projects can share an OIDC publisher, since multiple Python packages might live under the same logical source code repository. We probably don't want to mint a token that's scoped for _all_ of them at once, so we need a way to determine which one we should mint for.

woodruffw: (I don't see a significant security concern with minting for all of them, since the OIDC JWT bearer is trusted by definition, but that was my reasoning.)

warehouse/oidc/views.py
woodruffw: Sounds good. I can add another caveat to that effect.

warehouse/oidc/__init__.py
woodruffw: Yeah, I agree -- I think we discussed putting it under the current integration route namespace; I can do that once this is in a slightly more stable state 🙂

warehouse/oidc/views.py
di: I think we probably _do_ want to mint for all of them? I'm assuming that multiple projects that share the same publisher will want to use the same token to publish, instead of making multiple requests varying the project name to get multiple tokens.

warehouse/oidc/__init__.py
di: I'll think on this a bit, but also we should survey other providers to see if there's a common practice here.

warehouse/oidc/views.py
woodruffw: Makes sense. I'll drop the project from the body and mint for all projects.

di: Does it make sense to retain a body, or can it just be the OIDC token? Presumably everything we need is w/in the token, but leaving the door open to including additional data might help future-proof this.

woodruffw: IMO it makes sense to retain the body, even if it adds one or two additional error states (missing, invalid JSON) for the reason you mentioned.

warehouse/macaroons/interfaces.py
```diff
@@ -14,15 +14,6 @@
...
 class IMacaroonService(Interface):
-    def _extract_raw_macaroon(raw_macaroon):
-        """
-        Returns the base64-encoded macaroon component of a PyPI macaroon,
-        dropping the prefix.
-
-        Returns None if the macaroon is None, has no prefix, or has the
-        wrong prefix.
-        """
-
```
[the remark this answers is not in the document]
woodruffw: Yeah, we're only using it in the service. I think having it in the interface was a slipup on my part, >2 years ago at this point 🙂

warehouse/forklift/legacy.py
```diff
@@ -1439,7 +1474,7 @@ def file_upload(request):
                 action="add {python_version} file {filename}".format(
                     python_version=file_.python_version, filename=file_.filename
                 ),
-                submitted_by=request.user,
+                submitted_by=request.user if request.user else None,
```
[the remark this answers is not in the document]
woodruffw: Yeah, I think that's a good idea. Two items from this:

* We need some way to get the `OIDCProvider` in this context. `request.identity` isn't good enough (since it's just the `Project`) and our data model currently allows a project to have multiple providers. Maybe we eliminate the many-many relationship there so that we can reliably get `project.oidc_provider` when `isinstance(identity, Project)`?
* We'll need to retain `submitted_by` as non-optional, but drop the `User` relationship. It's currently a `CIText`, so maybe we have to change the column type as well?

warehouse/forklift/legacy.py
```diff
@@ -1015,27 +1031,42 @@ def file_upload(request):
...
+                    request.help_url(_anchor="project-name"),
+                )
+                if reason is None
+                else allowed.msg
+            )
+        else:
+            msg = (
+                (
+                    "The given token isn't allowed to upload to project '{0}'. "
+                    "See {1} for more information."
```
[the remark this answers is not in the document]
woodruffw: Yeah, I agree. We'll need some way to go from `request.identity` back to the specific publisher; I wrote an idea for that in https://github.com/pypa/warehouse/pull/11272#discussion_r878467712

warehouse/oidc/views.py
woodruffw: Make sense. Something like this?

```
{
  success: bool,
  description: Optional[str],
}
```

warehouse/forklift/legacy.py
```diff
@@ -1190,7 +1222,10 @@ def file_upload(request):
             tag="project:release:add",
             ip_address=request.remote_addr,
             additional={
-                "submitted_by": request.user.username,
+                "ephemeral": request.user is None,
+                "submitted_by": request.user.username
+                if request.user
+                else "OpenID created ephemeral token",
```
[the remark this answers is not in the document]
woodruffw: I agree about the overloading; I'll rewrite to drop "ephemeral" entirely.

warehouse/oidc/views.py
woodruffw: Oh, I realize why I did this originally -- I designed the original verification process to look like this:

1. Verify the signature on an untrusted JWT
2. Compare a list of projects against the JWT's claims, seeing if any projects have OIDC providers that match the claims

When really, our verification process should look like this:

1. Verify the signature on an untrusted JWT
2. Check whether we have an `OIDCProvider` corresponding to the now verified (but still untrusted) JWT
3. Return all projects registered against that `OIDCProvider`

That also simplifies our OIDC service interface, which can become something like `find_provider(unverified_jwt: str) -> OIDCProvider`, which in turn can be used to access the project list. I'll refactor towards that.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
di: I think we need to retain the many:many, especially for the future when we have multiple providers, and a project might need to use multiple providers to publish.

We probably need a way to retain the publisher used to authorize the request. We could possibly use `request.authentication_method` from https://github.com/pypa/warehouse/pull/10854 or something similar for this, getting these details from the macaroon (re my previous comment about adding this information to the macaroon).

warehouse/oidc/views.py
di: Maybe something like:

```
{
   "message": "Token request failed",
   "errors": [
     {
       "code": "some_unique_string",
       "description: "A nice human-readable description"
     }
   ]
 }
```

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: Gotcha. I think the macaroon is probably the right place for this -- I can add something like an `OIDCPublisherCaveat` that stores the DB ID for the publisher, which we can then sidecar somewhere in the `request`.

woodruffw:
Recapping the current approach, for discussion: 

1. The consumption endpoint requires two fields: an OIDC JWT and a project name
2. The project name is used to look up all OIDC providers registered to the project (if any), and check the token against each until a match (signature + claim set) is found
3. If a match is found, a temporary access token (a normal macaroon with some additional claims) is created and returned to the user via the HTTP response

The main advantage to this approach is simplicity: we know which OIDC providers we need to check against, since we know the project name we're checking the JWT against.

The main disadvantages are conceptual and practical:

* An OIDC JWT arguably corresponds to an access token granting access to *all* projects that it's registered against, not just the one the user happens to request. In other words, it should be sufficient for the request to just contain the JWT, and the response access token to be scoped to all projects.

As implied, the alternative solution:

1. The consumption endpoint requires one field: the OIDC JWT
2. The JWT is *partially* decoded and verified, to determine which provider it's valid against
3. The provider is used to collect the list of projects (via the projects registered against it)
4. A temporary access token (same as above) is created and returned to the user via the HTTP response (but this time, with potentially multiple projects scoped)

The main advantage to this is end-agent simplicity: the end agent only needs to know the JWT, not the name(s) of the projects that it's attempting to create tokens for/eventually upload. Moreover, the token will be scoped for all projects it's valid for, making it easier to reuse in the context of e.g. a release workflow that uploads multiple projects under an umbrella org.

The main disadvantages are internal complexity and state: accepting just the JWT makes the verification process much more involved, since we need to split signature verification apart from claim verification. That in turn makes logic errors easier, since the two will be pretty far apart in the codebase.

More perniciously, it requires us to rethink the way we're doing claim verification: right now we have a "declarative" approach where claims are explicitly grouped into different sets ("preverified", "verifiable", "unchecked") and misuse-resistant APIs for checking that the claim set is exactly as expected. With the "JWT only" approach, we'll need to push some claim verification into the SQL/ORM level, since we'll need to convert a partially verified JWT into an `OIDCProvider` model that we then re-verify. That in turn makes supporting multiple OIDC providers trickier, since they'll have different claim sets and therefore different SQL queries/ORM models necessary for retrieving the concrete model subclass.

---

TL;DR: The data model/flow for our current approach looks like this:

```
(jwt, project_name) -> Project -> [OIDCProvider]
```

...where each `OIDCProvider` is verified against `jwt`, and the matching one is used.

The data model for the alternative approach looks like this:

```
jwt -> OIDCProvider -> [Project]
```

...where `jwt` is used to retrieve `OIDCProvider`.

woodruffw:
Overall, I am *slightly* biased towards the current approach, since it allows us to retain our "declarative" APIs for claim verification (which I think are nice and avoid a lot of the misuse pitfalls with JWT verification). That being said, I think it's unfortunate that the current approach requires clients to send the project name along with the JWT, since conceptually the JWT should be sufficient on its own.

warehouse/macaroons/security_policy.py
```diff
@@ -88,13 +88,22 @@ def identity(self, request):
...
+        # Every Macaroon has either a user, or it's being used in the
+        # context of a project. For the latter, our authorization policy
+        # has already checked whether the project is valid for the Macaroon
+        # by checking the Macaroon's caveats.
+        if dm.user is not None:
+            return dm.user
+        else:
+            if not isinstance(request.context, Project):
+                return None
+            return request.context
```
woodruffw, line 106: Flagging for review: in order to get this PR into a working state, we need to get the intended `Project` model as the request identity here. 


The current approach doesn't actually work, since the upload endpoint doesn't provide a `context`.

Some ideas:

* We could parse the request's body as a form here, and extract the project name from the metadata. This would essentially duplicate the work already being done in the actual upload endpoint, so it's not ideal.
* We could figure out how to set `request.context` for the upload route. This will be a little thorny, since project creation involves sending a project name that doesn't exist yet. That should never happen with a macaroon (since a user can't create a macaroon that references a nonexistent project), but it's more state to deal with.
* Something else?

woodruffw: (And note: the reason we don't have `dm.project` to rely on is because a Macaroon can authenticate for multiple projects.)

dstufft: > That should never happen with a macaroon (since a user can't create a macaroon that references a nonexistent project), but it's more state to deal with.

I think I saw an issue recently where someone pointed out that you can have a Macaroon that refers to a now deleted project, unless we've fixed that and we're purging macaroons on deletion now.

woodruffw: I take that back, it's a annoying to do:

* A macaroon might be scoped for multiple projects, meaning that we'd invalidate other potential users of the token
* The only server-side reference to the Macaroon's projects is kept in a JSON blob in the DB, so we'd need a query that peeks into that blob and checks whether the deleted project is a member of the projects list

woodruffw: That being said, we definitely should purge or invalidate Macaroons on project (and account) deletion -- this PR makes users optional on Macaroons, giving us a potential security problem where an ephemeral orphaned macaroon could potentially authenticate for a project that's been deleted and then re-created by a different user. The window for that would be very small, but it is possible.

We could preempt that by ensuring that the project has an OIDC provider enabled, but that's a hack and not a real solution.

dstufft: https://github.com/pypi/warehouse/issues/9524 found the issue I was thinking of

dstufft: I haven't actually reviewed this PR yet, but are we allowing Macaroons that aren't attached to a user, and are attached to multiple projects?

woodruffw: > I haven't actually reviewed this PR yet, but are we allowing Macaroons that aren't attached to a user, and are attached to multiple projects?

Yep, correct -- any project that authorizes the OIDC provider that verifies the JWT will be scoped within the Macaroon. That's based on the redesign described in https://github.com/pypi/warehouse/pull/11272#issuecomment-1168835516 and forwards.

woodruffw:
Blocked on https://github.com/pypi/warehouse/pull/11857, to constrain minted macaroons down to their scoped projects' IDs.

woodruffw:
Just to summarize a long conversation I had with @dstufft:

* Projects don't make sense as identities; they make sense as contexts. I need to refactor this PR to use the `OIDCProvider` itself as the identity, and reuse the existing permissions code to set the right project as `context`. The right way to do this is _probably_ to embed a caveat specifying the backing `OIDCProvider` into the generated Macaroon?
* I also need to use the caveat we've added in https://github.com/pypi/warehouse/pull/11857, to ensure that any ephemeral API tokens we create here don't become (briefly) valid for re-created projects that aren't under the original owners' control.

warehouse/macaroons/services.py
```diff
@@ -120,27 +120,42 @@ def verify(self, raw_macaroon, context, principals, permission):
         if dm is None:
             raise InvalidMacaroonError("deleted or nonexistent macaroon")
 
-        verifier = Verifier(m, context, principals, permission)
+        verifier = Verifier(m, context, principals, permission, identity)
```
woodruffw, line 123: Flagging for review: I plumbed `request.identity` through here, since the macaroon AuthZ layer needs to know the current identity to pass/fail the token. There might be a cleaner way to do this, although I suspect that we can clean this up more generally by removing the legacy `IAuthorizationPolicy` implementations and doing everything directly though `ISecurityPolicy`.

warehouse/oidc/__init__.py
woodruffw: Just pinging back on this to say that I haven't forgotten; I plan to do this once we've done some review on the actual functionality 🙂 

(I think a route under `/integrations/` might be best?)

warehouse/utils/security_policy.py
```diff
@@ -99,8 +100,10 @@ def permits(self, request, context, permission):
             if isinstance(identity, User):
                 principals.append(f"user:{identity.id}")
                 principals.extend(_principals_for_authenticated_user(identity))
+            elif isinstance(identity, OIDCProvider):
+                principals.append(f"oidc:{identity.id}")
```
woodruffw, line 104: Same note as https://github.com/pypi/warehouse/pull/11272/files#r923830262 -- this is the right principal to apply, but it's probably effectively a no-op since we don't match against ACLs for Macaroons at the moment.

woodruffw:
This still needs lots of work (including unit tests), but I'm marking it as ready for an initial review (since I think the overall approach is now correct).

warehouse/oidc/views.py
```diff
@@ -0,0 +1,88 @@
...
+    now = time.time()
+    expires = int(now) + 900
+    caveats = [
+        {"permissions": "oidc", "version": 1},
+        # TODO: Does this caveat make sense anymore?
+        # Conceptually, an OIDC provider authorizes either for every project registered
+        # against it at the time of token creation, *or* for every project that happens
+        # to be registered against it at the time of use. This caveat constrains the
+        # token to the former, but maybe the latter is fine/expected?
+        {"project_ids": [str(p.id) for p in provider.projects]},
```
woodruffw, line 70: Flagging for review (cc @dstufft in particular).

warehouse/packaging/models.py
```diff
@@ -250,6 +250,11 @@ def __acl__(self):
             (Allow, "group:moderators", "moderator"),
         ]
 
+        # The project has zero or more OIDC "providers" registered to it,
+        # each of which serves as an identity with the ability to upload releases.
+        for provider in self.oidc_providers:
+            acls.append((Allow, f"oidc:{provider.id}", ["upload"]))
```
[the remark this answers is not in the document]
dstufft: We only have a singular AuthZ policy, the Macaroon policy is currently implemented as a wrapper around the ACL policy, where Macaroons apply further restrictions to the permissions that the user (or now OIDC Provider) has naturally via ACL.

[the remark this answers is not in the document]
dstufft: Or rather, a singular AuthZ policy object is all that is active at one time, and ours looks like

``2FAAuthZ(MacaroonsAuthZ(ACLAuthZ))``, and the underlying permission grants come from the ACL system, and past the remaining two policies that wrap the ACL AuthZ policy just add additional restrictions.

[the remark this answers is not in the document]
woodruffw: Whoops, you're absolutely right. I misread my own code as "only passes to the underlying ACL policy if we're not validating a macaroon," but we always pass through unconditionally.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: I took another look at this, and here's how `submitted_by` is defined and related:

```py
    _submitted_by = Column(
        "submitted_by",
        CIText,
        ForeignKey("users.username", onupdate="CASCADE"),
        nullable=True,
    )
    submitted_by = orm.relationship(User, lazy="raise_on_sql")
```

So we have a couple of options:

* Remove the FK constraint + relationship and turn `submitted_by` into an ordinary `CIText` column
* Add a separate column + relationship for `JournalEvent`s submitted by `OIDCProvider`s, additionally constraining them such that `JournalEvent -> User XOR OIDCProvider` (similar to the `Macaroon` table)
* More extremely: journal events have been deprecated for a while (since 2019), so maybe we can just delete this entirely? 

Thoughts?

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: This should be addressed by the error message refactor in #11885; we'll only return this error message if the authz layer doesn't provide a message, which it always will for macaroons now.

woodruffw:
Blocking on #11903 to keep the churn down.

woodruffw:
Rebased on the "v2" caveat format, and tested locally. I'll start to get the test coverage up as well.

warehouse/macaroons/caveats/__init__.py
```diff
@@ -100,6 +101,38 @@ def verify(self, request: Request, context: Any, permission: str) -> Result:
...
+            return Failure("OIDC scoped token used outside of a project context")
+
+        # Specifically, they are only valid against projects that are registered
+        # to the current identifying OpenID provider.
+        if context not in request.identity.projects:
+            return Failure(
+                f"OIDC scoped token is not valid for project '{project.name}'"
+            )
+
+        return Success()
```
woodruffw, line 133: cc @dstufft I'm pretty sure I got this right, but since you designed the new API here I figured I'd flag it specifically for your eyes 🙂

woodruffw:
Resuming work on this today. The big items are test coverage and feature gating.

woodruffw:
Tests are passing locally; here's the remaining coverage map (for my tracking purposes)

```
Name                                      Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------
warehouse/forklift/legacy.py                459      2    261      3    99%   796->817, 900, 960
warehouse/macaroons/caveats/__init__.py      95      9     40      0    87%   112-133
warehouse/macaroons/security_policy.py       80      1     24      1    98%   102
warehouse/macaroons/services.py              91      2     29      2    97%   96, 141
warehouse/manage/views.py                  1336      0    372      1    99%   3055->3058
warehouse/oidc/services.py                  122      2     36      1    98%   262-266
warehouse/oidc/utils.py                      14      9      4      0    28%   29-59
warehouse/oidc/views.py                      28     21      8      0    19%   30-88
warehouse/packaging/models.py               301      1     49      1    99%   264
warehouse/utils/security_policy.py           59      1     28      1    98%   104
-------------------------------------------------------------------------------------
TOTAL                                     11996     48   2994     10    99%
```

woodruffw:
Whoops, didn't mean to close...

woodruffw:
Okay, we should be back at 100% coverage now. I need to do some local manual testing to confirm that I haven't broken anything + feature-gate this with the rest of the OIDC functionality, but otherwise it should be good to go?

warehouse/oidc/__init__.py
woodruffw: @di any follow-up thoughts on the route path? It's still under `/api/` for now, but changing it is easy enough.

woodruffw:
Just had a thought: when OIDC is disabled via an admin flag, do we want to temporarily disable uploads via API tokens that were minted via OIDC? Right now we disable the minting itself, but any tokens that _were_ issued before the flag's enablement remain valid until their expiry.

di:
> Just had a thought: when OIDC is disabled via an admin flag, do we want to temporarily disable uploads via API tokens that were minted using OIDC? Right now we disable the minting itself, but any tokens that were issued before the flag's enablement remain valid until their expiry.

I think this is fine as is: we wouldn't want to break workflows unnecessarily. But we might want a separate flag to disable tokens as well if we need to?

woodruffw:
> But we might want a separate flag to disable tokens as well if we need to?

Meaning all API tokens? That makes sense to me; I can do that in a separate PR if you'd like (either in parallel or after this one, since it's functionally independent but will touch a lot of the same files).

warehouse/oidc/views.py
```diff
@@ -0,0 +1,103 @@
...
+                {"code": "invalid-token", "description": "malformed or invalid token"}
+            ]
+        )
+
+    # At this point, we've verified that the given JWT is valid for the given
+    # project. All we need to do is mint a new token.
+    macaroon_service = request.find_service(IMacaroonService, context=None)
+    not_before = int(time.time())
+    expires_at = not_before + 900
+    serialized, dm = macaroon_service.create_macaroon(
```
woodruffw, line 78: Flagging for review, since I can't remember if we discussed this: does it make sense to also register a timed task/DB trigger here, to delete this macaroon once it has expired?

I don't think it's a security issue for it to remain (since it'll be invalid), but I can see a couple of advantages to removal:

* We'll be able to emit a `project:api_token:removed` event for these tokens, the way we do for "ordinary" ones
* Without timed removal, these macaroons will be difficult to prune from the DB (since we don't copy the entire caveat set into the DB to filter with after the fact)

woodruffw:
Updated the title here to clarify that this primarily concerns GitHub's OIDC provider, although the changes themselves are mostly generic.

warehouse/oidc/__init__.py
di: I guess the question is whether this should be generic or GitHub-specific.

If it's GitHub-specific, which it looks like it is right now, this should probably be something like `/_/oidc/github/mint-token`

warehouse/oidc/views.py
di: We probably need to discern between short-lived and long-lived tokens in the events -- if we can do that, I don't think we need to emit a 'removed' event for short-lived tokens.

I think it's probably going to be better to keep these around in order to provide better error messages (and events) in the event that someone tries to use an expired token.

warehouse/forklift/legacy.py
```diff
@@ -1072,10 +1103,11 @@ def file_upload(request):
                     "requires_python",
                 }
             },
-            uploader=request.user,
+            uploader=request.user if request.user else None,
```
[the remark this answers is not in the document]
woodruffw: That would be ideal, although refactoring might be a little messy -- `uploader` is currently an `Optional[User]` via `users.id`, so replacing it with something more generic (like an `Optional[Identity]`?) would probably require a lot of unrelated changes to the codebase.

More generally though I agree, and I _think_ what we probably want is a polymorphic `Identity` model, which relates to either `User` or `OIDCProvider` (and potentially future things). Once we have that, every dependent model (like `Journal` and `Release`) could be re-related to make it generic over all identity types. How would you feel about me doing that in a separate PR?

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: Sounds good, I'll add that.

warehouse/oidc/__init__.py
woodruffw: Yeah, it's currently GitHub specific (since we need to select the right `OIDCProviderService` for the request). Ideally it'd be provider agnostic, but that would require partially inspecting the JWT before verifying it.

warehouse/oidc/views.py
woodruffw: Sounds good. I can update the event messages here to emphasize that these are short-lived tokens.

woodruffw: (Actually, even better: I can put that in the `additional` metadata, and the event table renderer can look for it and display it as a little "chip" or something like that.)

woodruffw: ...or just compute that it's "short-lived" from the `expires` key.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
di: I agree. OK to do that in a follow-up PR.

warehouse/oidc/views.py
woodruffw: [b8d6b0c](https://github.com/pypi/warehouse/pull/11272/commits/b8d6b0ca7d0c3875d4a81f1d6a6acfb65bdbeb10) should address this.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: Cool; I'll make a tracking issue.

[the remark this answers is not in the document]
woodruffw: https://github.com/pypi/warehouse/issues/12477

warehouse/macaroons/services.py
woodruffw: Yeah, I think so -- I can do this as part of #12477.

warehouse/oidc/views.py
woodruffw: I'm going to remove this TODO -- it's stale with the caveat refactor, and I'm convinced that it makes sense for an API token to be limited to only the projects that existed at the time it was created.

warehouse/macaroons/services.py
woodruffw: Stale.

warehouse/macaroons/services.py
woodruffw: af976cdd

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: I'll do this as part of #12477 as well.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    try:
+        body = request.json_body
+    except ValueError:
+        return _invalid(
+            errors=[{"code": "invalid-json", "description": "missing JSON body"}]
+        )
+
+    unverified_jwt = body.get("token")
```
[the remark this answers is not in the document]
woodruffw: We actually _do_ catch JSON decoding errors, since the `json_body` property raises if the request body isn't valid JSON.

```python
    try:
        body = request.json_body
    except ValueError:
        return _invalid(
            errors=[{"code": "invalid-json", "description": "missing JSON body"}]
        )
```

(Ref: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/webob.html#request-json-body)

In this case it's a type error, since payload _is_ valid JSON but the top-level JSON type is a string instead of a map. So I'll add another check here to make sure we're getting the type we actually expect out of `json_body` 🙂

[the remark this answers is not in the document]
woodruffw: 0f03cab7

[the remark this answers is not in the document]
di: Ah, good catch. I can confirm that this fixes the issue.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,110 @@
...
+                    "code": "invalid-payload",
+                    "description": "payload is not a JSON dictionary",
+                }
+            ]
+        )
+
+    unverified_jwt = body.get("token")
+    if not unverified_jwt or not isinstance(unverified_jwt, str):
+        return _invalid(
+            errors=[{"code": "invalid-token", "description": "missing or empty token"}]
```
di, line 71: Can we make this error more clear? Two things could be going wrong here: the 'token' field is missing entirely, or the value in the 'token' field is invalid.

woodruffw: Yep! Doing now.

woodruffw: e2df77ee

warehouse/macaroons/caveats/__init__.py
```diff
@@ -100,6 +101,38 @@ def verify(self, request: Request, context: Any, permission: str) -> Result:
...
+class OIDCProvider(Caveat):
+    oidc_provider_id: StrictStr
+
+    def verify(self, request: Request, context: Any, permission: str) -> Result:
+        # This caveat can only be used when the surrounding request is identified
+        # with OpenID, rather than a user or some other identity.
+        if not isinstance(request.identity, oidc_models.OIDCProvider):
+            return Failure(
+                "OIDC scoped token used outside of an OIDC identified request"
+            )
```
di, line 115: Does this make sense? The identity comes from the token provided, not some characteristic of the surrounding request.

warehouse/templates/manage/project/history.html
```diff
@@ -167,9 +172,16 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
     {% elif event.tag == EventTag.Project.APITokenAdded %}
     <strong>{% trans %}API token added{% endtrans %}</strong>
```
di, line 173: I think we probably want a different heading here entirely for these, something like "Short-lived API token created"

warehouse/oidc/views.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+
+    # At this point, we've verified that the given JWT is valid for the given
+    # project. All we need to do is mint a new token.
+    macaroon_service = request.find_service(IMacaroonService, context=None)
+    not_before = int(time.time())
+    expires_at = not_before + 900
+    serialized, dm = macaroon_service.create_macaroon(
+        request.domain,
+        f"OpenID token: {provider} ({not_before})",
```
di, line 98: OpenID token

di, line 98: Should this include the provider name as well? Currently it's just stringifying the provider, which doesn't include the provider name (maybe that should instead?). E.g., "GitHub".

warehouse/macaroons/caveats/__init__.py
woodruffw: I _think_ we need this because of Macaroons' ability to provide custom attenuation: a user could conceivably add a third-party `OIDCProvider` caveat to their otherwise valid user-identified macaroon, which would then fail because of the following ID comparison. But that comparison doesn't make sense to do in the first place, so having a check on the kind of requesting identity allows us to provide a slightly nicer error message.

(I believe that's the rationale for why we do a similar check in the `RequestUser` caveat.)

warehouse/oidc/views.py
di: Looking at this:

image

The 'token name' line is pretty confusing here, because it's not really a name like other tokens. The timestamp is not human-readable either. Maybe this could be something like:

```
Creator: https://github.com/di/nachos-test/.github/workflows/publish.yml
```

which could even be a link!

woodruffw: Yeah, I think so -- more generally this is making me realize that the current provider stringification isn't great (it's a syntax I made up).

warehouse/templates/manage/project/history.html
woodruffw: Sounds good. In that case, were you thinking they'd have an entirely separate `EventTag.Project` value?

warehouse/templates/manage/project/history.html
```diff
@@ -167,9 +172,16 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
...
+    {% if event.additional.user %}
     <small>
       {% trans %}Controlled by:{% endtrans %} <a href="{{ request.route_path('accounts.profile', username=event.additional.user) }}">{{ event.additional.user }}</a>
     </small>
+    {% endif %}
+    {% if event.additional.expires %}
+    <small>
+      {% trans %}Expiration: {% endtrans %} {{ humanize(event.additional.expires | ctime) }}
+    </small>
+    {% endif %}
```
di, line 184: Might be nice to add an additional field here before this, something like:

```
Expiration Status: Active
```
then:
```
Expiration Status: Expired
```

If that's a ton of pain to compare times in the template though, we can omit this.

warehouse/macaroons/caveats/__init__.py
di: Ah, I mostly meant the comment. The identity isn't coming from the 'surrounding request', it's coming from the identity we have correlated with this macaroon ID.

warehouse/oidc/views.py
di: I think it'd be super nice if these were links -- think about how we might expose these publicly alongside a release.

warehouse/macaroons/caveats/__init__.py
woodruffw: Oh, I see what you mean! I can clarify the comment's intent.

warehouse/templates/manage/project/history.html
di: If that's easier, yeah let's do it.

di: Would make it easier for us to filter different types of token events in the future as well.

woodruffw: Yep, I think that'll be easiest. Doing now.

warehouse/oidc/views.py
woodruffw: I agree! I think I'm going to refactor the `OIDCProvider` interface to require that every provider include its link, and I'll include that in the render.

warehouse/templates/manage/project/history.html
woodruffw: I think that's doable with another Jinja2 filter; will do.

warehouse/oidc/views.py
woodruffw: One hiccup with linking: we can probably link to the repository, but linking directly to the workflow itself might be a little tricky (since we don't actually know the repo's default branch name, or if the exact workflow ran even came from the default branch).

woodruffw: ...although we do have the `sha` claim, so maybe we could reuse that?

woodruffw: Never mind, I can't use the `sha` claim in this context unless we include it in the set of DB-stored claims.

di: We can bypass the issue with the default branch name by using something like:

https://github.com/pypi/warehouse/blob/HEAD/.github/workflows/ci.yml

I think we can punt on figuring out the exact SHA/workflow that resulted in a build until later.

warehouse/templates/manage/project/history.html
woodruffw: [0fb4679](https://github.com/pypi/warehouse/pull/11272/commits/0fb4679ca2394e08389e8a284910bdcb33b7f95e)

warehouse/oidc/views.py
woodruffw: [0fb4679](https://github.com/pypi/warehouse/pull/11272/commits/0fb4679ca2394e08389e8a284910bdcb33b7f95e)

warehouse/templates/manage/project/history.html
woodruffw: [0fb4679](https://github.com/pypi/warehouse/pull/11272/commits/0fb4679ca2394e08389e8a284910bdcb33b7f95e)

warehouse/macaroons/caveats/__init__.py
woodruffw: [b31c3b6](https://github.com/pypi/warehouse/pull/11272/commits/b31c3b625447dbf879d7b586de4c0ebdd89f989f)

warehouse/oidc/views.py
```diff
@@ -0,0 +1,119 @@
...
+        return _invalid(
+            errors=[
+                {"code": "invalid-token", "description": "malformed or invalid token"}
+            ]
+        )
+
+    # At this point, we've verified that the given JWT is valid for the given
+    # project. All we need to do is mint a new token.
+    # NOTE: For OIDC-minted API tokens, the Macaroon's description string
+    # is purely an implementation detail and is not displayed to the user.
```
woodruffw, line 94: Flagging for visibility: the other option here is to relax the `Macaroon.description` column and allow it to be nullable.

warehouse/templates/manage/project/history.html
```diff
@@ -169,8 +169,31 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
...
+      {% else %}
+        {% trans %}Expired{% endtrans %}
+      {% endif %}
+    </small>
+    {% endif %}
+    <small>
+      {% trans %}Creator: {% endtrans %}
+      {{ event.additional.provider_name }}:
+      <a href="{{ event.additional.provider_url }}">
+        {{ event.additional.provider_spec }}
```
di, line 192: I think this should just be `provider_url` (and there's at least one other place we're showing the spec as well)

woodruffw: Yeah, I believe we show it in the registered providers table. I'll update there as well.

woodruffw:
Current event render:

Screenshot 2022-11-02 at 6 15 22 PM

warehouse/templates/manage/project/history.html
woodruffw: (I'll do that after I'm merged on #12479, since it'll need deconflicting.)

alex:
Is including a link to/ID for the specific _run_ practical? That'd give a greater degree of auditability.

woodruffw:
> Is including a link to/ID for the specific _run_ practical? That'd give a greater degree of auditability.

Yeah, it should be (not with state on the OIDC provider itself, but with state from the JWT.) Let me see if I can figure out a way to pull that out.

di:
> Is including a link to/ID for the specific run practical? That'd give a greater degree of auditability.

Yeah, would be good to include this with the publishing event/journal, and we can figure out how to externalize it in a followup PR.

warehouse/templates/manage/project/history.html
di: Merged!

woodruffw:
> Yeah, would be good to include this with the publishing event/journal, and we can figure out how to externalize it in a followup PR.

Did this with 8e3d5b1d -- nothing is exposed yet, but we now have access to the claims set that the verified JWT contained.

warehouse/templates/manage/project/history.html
woodruffw: Updated and tweaked the table to contain the provider URL instead.

warehouse/oidc/views.py
```diff
@@ -0,0 +1,118 @@
...
+    if not isinstance(unverified_jwt, str):
+        return _invalid(
+            errors=[{"code": "invalid-token", "description": "token is not a string"}]
+        )
+
+    # For the time being, GitHub is our only OIDC provider.
+    # In the future, this should locate the correct service based on an
+    # identifier in the request body.
+    oidc_service = request.find_service(IOIDCProviderService, name="github")
+    provider, _claims = oidc_service.find_provider(unverified_jwt)
```
di, line 83: I don't _love_ this interface, partly because of the multiple return values, and partly because this is still called `find_provider`.

Would it make sense to have two functions here, one being `verify_jwt` which takes an unverified JWT and returns the signed payload, if any, and a second `find_provider` which takes a verified JWT and returns the provider?

warehouse/oidc/views.py
di: I'm thinking it might be more helpful for this to be the `provider_url`.

warehouse/oidc/views.py
woodruffw: Yeah, I agree. 

Part of why I joined them is because I want this API to be as misuse-resistant as possible: it should be impossible to look up an `OIDCProvider` without having a valid JWT corresponding to that provider, and it should be impossible to "verify" a JWT in a way that doesn't imply its intended use.

I think `verify_jwt` and `find_provider` would work, assuming signatures like this:

```python
def verify_jwt(unverified_token: str) -> VerifiedClaims: ...
def find_provider(verified_claims: VerifiedClaims) -> OIDCProvider: ...
```

...where `VerifiedClaims` is a `NewType` over the claim dictionary. That won't *prevent* API misuse, but it should help MyPy detect it. Thoughts?

di: That seems acceptable to me.

woodruffw: Great! I'll make those changes now.

woodruffw: Oh, one tweak: `verify_jwt` should probably be `verify_jwt_signature`, since the "verification" we do in the context of returning the signed claims is *only* that the JWT has a valid and timely signature.

woodruffw: [f48da18](https://github.com/pypi/warehouse/pull/11272/commits/f48da1898e31fc88240447e56f328a28a908828e) has the changes. I'll update the tests with a follow-up and confirm that everything still works.

woodruffw:
This should be up-to-date and ready to go again.

warehouse/oidc/views.py
woodruffw: Whoops, forgot to get this with the last changeset. Fixing now!

woodruffw: 667c9fc5

di (approved):
LGTM. Can you bring this up to date so we can merge? I don't have the ability to do so.

woodruffw:
Okay, should be good to go.

PR #11309, opened 2022-05-02
Add pre-commit

xsc27:
- chore: make scripts executable
- chore: remove execute bit from non-scripts
- ci(pre-commit): add config


miketheman:
This is very cool, thanks for doing it!
I’d suggest adding a blurb in the development docs to encourage the developer to install pre-commit as well.

Considering there’s still some listing issues in admin templates, are you planning on resolving those before this is ready for merge, or would it make sense to remove the admin scope for now, and follow this up with a fix when that’s done?

I’d also suggest setting the python version for black, per their advice: https://black.readthedocs.io/en/stable/integrations/source_version_control.html

di:
I'd like to continue ignoring the admin templates -- that was intentional, as they're not user-facing.

It looks like there's some ancillary things getting included (like an empty migration file) that shouldn't be here, unless I'm missing something.

xsc27:
additionally to adding `pre-commit` to the dev doc, `bin/lint` should be replaced with `pre-commit` to have a single check.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+    rev: v4.2.0
+    hooks:
+      - id: check-executables-have-shebangs
+      - id: check-merge-conflict
+      - id: check-shebang-scripts-are-executable
+      - id: check-toml
+      - id: check-yaml
+      - id: fix-byte-order-marker
+      - id: no-commit-to-branch
+        args: ["-b", "main"]
```
miketheman, line 12: question for maintainers: While I think this is a good idea, do y'all never commit and push direction to `main`? It appears that way, and hopefully there's a [GitHub branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#restrict-who-can-push-to-matching-branches) to prevent.

@xsc27 - I think this is a good idea for forked repos, so contributors are guided away from making commits to their own forked `main` and then get into trouble with upstream merges.

miketheman (commented):
@xsc27 Nice update. With `black` configured, we are seeing a few files that would be reformatted as a result. Any idea why they are being modified now as a result?

dstufft:
I'm going to go ahead and close this. It's bit rotted since this PR was made, but more foundational, I don't believe that we want to utilize pre-commit on this code base.

PR #11314, opened 2022-05-02
Pyramid 2.0 security policies: take two

woodruffw:
Take two on #11218, which was reverted by #11313.

No functional changes; we misdiagnosed user error as a regression.


woodruffw:
I should have done this with a revert-revert. Closing.

PR #11333, opened 2022-05-04
tests, warehouse: avoid potential recursion in authenticated_userid

woodruffw:
This should prevent a recursive cycle when trying to grab a user ID from a request.

Tests pass, but I also need to do some local testing to confirm.


woodruffw:
One of the few remaining uses of `unauthenticated_userid` is in `_redact_ip` in `warehouse/email/__init__.py`:

```py
def _redact_ip(request, email):
    # We should only store/display IP address of an 'email sent' event if the user
    # who triggered the email event is the one who receives the email. Else display
    # 'Redacted' to prevent user privacy concerns. If we don't know the user who
    # triggered the action, default to showing the IP of the source.

    try:
        user_email = request.db.query(Email).filter(Email.email == email).one()
    except NoResultFound:
        # The email might have been deleted if this is an account deletion event
        return False

    if request.unauthenticated_userid:
        return user_email.user_id != request.unauthenticated_userid
    if request.user:
        return user_email.user_id != request.user.id
    return False
```

If I'm reading this correctly, I think this should be safe to replace with:

```python
# always redact the IP if the request doesn't have an associated user
if not request.user:
    return True
else:
    return user_email.user_id != request.user.id
```

Thoughts? cc @dstufft in particular, since you know the auth components much better than I do and I'm not 100% why this was checking `unauthenticated_userid` in the first place.

di (approved):

PR #11341, opened 2022-05-05
warehouse, tests: check that session's user still exists

woodruffw:
This prevents a potential race condition during account deletion:
a user might delete their account and then refresh or navigate
while deletion is still processing, causing their session to still
be valid but their underlying `User` model to no longer exist.

This previously caused a server-side error; we now treat it the
same as a nonexistent session.


warehouse/accounts/security_policy.py
```diff
@@ -117,16 +117,25 @@ def identity(self, request):
...
         login_service = request.find_service(IUserService, context=None)
 
+        # A user might delete their account and immediately issue a request
+        # while the deletion is processing, causing the session check
+        # (via authenticated_userid above) to pass despite the user no longer
+        # existing. We catch that here to avoid raising during the password
+        # staleness check immediately below.
+        user = login_service.get_user(userid)
+        if user is None:
+            return None
```
woodruffw, line 137: I deleted this check from the original refactor because I thought session lookup guaranteed us `user` availability; hopefully this comment (and test coverage) will prevent any future removals 🙂

di (approved):

PR #11351, opened 2022-05-06
warehouse, tests: require a matched route for session auth

woodruffw:
Similar to #11350.

### DIFF
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -121,6 +121,10 @@ class SessionSecurityPolicy:
         request.add_response_callback(add_vary_callback("Cookie"))
         request.authentication_method = AuthenticationMethod.SESSION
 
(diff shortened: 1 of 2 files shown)


warehouse/accounts/security_policy.py
```diff
@@ -121,6 +121,10 @@ def identity(self, request):
         request.add_response_callback(add_vary_callback("Cookie"))
         request.authentication_method = AuthenticationMethod.SESSION
 
+        # A route must be matched
+        if not request.matched_route:
+            return None
+
         # Session authentication cannot be used for uploading
         if request.matched_route.name in ["forklift.legacy.file_upload"]:
```
di, line 129:
```suggestion
        if request.matched_route.name == "forklift.legacy.file_upload":
```

di (approved):

PR #11378, opened 2022-05-13
Fix SQLAlchemy warnings

divbzero:
Minor adjustments resolve SQLAlchemy warnings that would appear when running `make tests`. The following changes reduced the total number of warnings from 569 to 523.

Resolve warnings about `.events` relationships:

    SAWarning: relationship 'User.events' will copy column users.id to column user_events.source_id, which conflicts with relationship(s): 'UserEvent.source' (copies users.id to user_events.source_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="source"' to the 'User.events' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)

Resolve warnings about many-to-many relationships:

    SAWarning: relationship 'Project.users' will copy column users.id to column roles.user_id, which conflicts with relationship(s): 'Role.user' (copies users.id to roles.user_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="user"' to the 'Project.users' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)

Resolve warnings about `or_()` with dynamic args:

    SADeprecationWarning: Invoking or_() without arguments is deprecated, and will be disallowed in a future release. For an empty or_() construct, use or_(False, *args).

Resolve warnings about using subquery for `in_()`:

    SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly

Resolve warnings about implicit coercion of `SELECT`:

    SADeprecationWarning: Implicit coercion of SELECT and textual SELECT constructs into FROM clauses is deprecated; please call .subquery() on any Core select or ORM Query object in order to produce a subquery object.


di:
> consider if these relationships should be linked with back_populates

I think we should probably do what the warning is suggesting here instead?

di (approved):

di:
I resolved the merge conflicts but the tests are now failing. I'm happy to merge this if we can resolve those failures.

divbzero:
@miketheman and @di: I’ve fixed the tests for this PR, should be ready for merge.

divbzero:
@dstufft Thank you!

PR #11426, opened 2022-05-20
Fix tests with time-dependent token serializer

### LINKED ISSUE #11424: More flakey organization tests
See https://github.com/pypa/warehouse/runs/6518601158

(cc @divbzero @sterbo)

divbzero:
Fixes #11424.

### DIFF
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -2703,6 +2703,8 @@ class TestManageOrganizationRoles:
             send_organization_role_verification_email,
         )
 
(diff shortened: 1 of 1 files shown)


di:
The issue seems to be that the ordering of the calls is non-deterministic. I'd recommend either a) ordering the models that `send_organization_member_invited_email` is getting called on, or b) asserting that a set of calls has been made, not a list of calls, if the ordering doesn't matter.

di:
Fixed by #11521.

PR #11473, opened 2022-05-28
Manage organization "Projects"

### LINKED ISSUE #11085: Transfer a project
Feature request for organization account project in PyPI.

Description | The Owner can transfer a project to the Organization account
-- | --
User value | The project becomes associated with the Organization
Acceptance criteria | Email notification to Owner and project maintainers

### LINKED ISSUE #11086: Create a project
Feature request for organization account project in PyPI.

Description | The Owner can create a project
-- | --
User value | The project becomes associated with the Organization
Acceptance criteria | Email notification to Owner

divbzero:
- Closes #11085.
- Closes #11086.


warehouse/utils/project.py
```diff
@@ -28,15 +40,156 @@ def remove_documentation(task, request, project_name):
...
+STDLIB_PROHIBITED = {
+    canonicalize_name(s.rstrip("-_.").lstrip("-_."))
+    for s in chain.from_iterable(
+        _namespace_stdlib_list(stdlib_list.stdlib_list(version))
+        for version in stdlib_list.short_versions
+    )
+}
+
+
+def add_project(name, request):
```
ewdurbin, line 59: One concern with this refactor is that it has side effects that I believe are incorrect for all use cases as discussed in another comment. I would suggest refactoring further to separate the concerns of "can we create this project given its name?" and "actually create the project".

warehouse/manage/views.py
```diff
@@ -1367,6 +1378,132 @@ def delete_organization(self):
...
+        form = default_response["add_organization_project_form"]
+        if not form.validate():
+            return default_response
+
+        # Get existing project or add new project.
+        if form.add_existing_project.data:
+            project = self.project_factory[form.existing_project_name.data]
+        else:
+            try:
+                project = add_project(form.new_project_name.data, self.request)
```
ewdurbin, line 1462: After this line has successfully executed, we will have a new `Project` object and a new `Role` object associating `self.request.user` with the `Project` as `"Owner"`. I believe this is incorrect behavior for the purposes of this feature. The resulting `Role` is independent of membership in `self.organization`, so by leaving or being removed from an organization that user would retain access (not good). `Project` objects associated via `OrganizationProject` objects to an `Organization` should allow organization owners all necessary permissions to manage the resulting `Project`. Since `add_organization_project` should only be available to organization owners the end result _without_ the `Role` is then the same.

warehouse/manage/views.py
```diff
@@ -1367,6 +1378,132 @@ def delete_organization(self):
...
+        }
+
+    @view_config(request_method="GET")
+    def manage_organization_projects(self):
+        if self.request.flags.enabled(AdminFlagValue.DISABLE_ORGANIZATIONS):
+            raise HTTPNotFound
+
+        return self.default_response
+
+    @view_config(request_method="POST")
```
ewdurbin, line 1446: Seems the `view_config` here should only be accessible to those with `manage:organization` permissions on the org. Currently the class level permission is `view:organization`, so any member could add a project?

ewdurbin (commented):
Points that need clarity/discussion along with a specific request around refining the refactor of the `add_project` function.

warehouse/manage/views.py
divbzero: If a project is owned by an organization but has no individual owners, what happens if the project is removed from the organization? Ways of handling this include:

1. The organization owner who removes the project is added as an individual owner.
2. All organization owners are added as individual owners.
3. We don’t allow this operation. There must be at least one individual owner before the project can be removed from the organization.

ewdurbin: The "removed" from an organization operation should only be able to happen via deletion or transfer.

Deletion, easy! We remove the OrganizationProject relation, all Releases, and the Project itself.

For Transfer "out" which I now realize we don't have scoped... Yes the third option makes the most sense to me. The Project should have to be transferred to _another_ organization or have at least one Role object with role "Owner" associated with an individual User.

warehouse/manage/views.py
divbzero: Fixed with commit [`6f4928a`](https://github.com/pypa/warehouse/pull/11473/commits/6f4928a129bbb81acea7b0ea9710343364a3b4eb).

ewdurbin (commented):
In review, the nuts and bolts of this look great. A couple UX things and points of clarification for transfers:

UX:

* When viewing projects owned by an organization either by transfer or creation, as an Owner of the Organization I *am* able to directly hit the manage page, but the buttons and badges on the page are not accurate. Specifically the "Manage" button is greyed out. ACL is good though since I can navigate to the project manage page directly. I think the buttons/badges are determined based on https://github.com/pypa/warehouse/blob/028dffea1db70ca6f884cf9b81cedb3a12fc49a2/warehouse/manage/views.py#L128-L181 so we need to see some updates there, I'm sure this will be confusing during user testing.
* "Sole Owner" starts to lose meaning for Organization accounts but could be similarly determined by calculating the total number of Users with either a direct Owner Role or ownership via an Organization.
* I would suspect that people will be confused not seeing Organization Projects they have access to in their main project view. Perhaps this should come later? But again I'd suspect this would be flagged during user testing.

Clarification on transfers:
* After transferring a project, my User still holds an Owner Role on the project directly, and I cannot remove that role myself. I'm not *positive* what people would expect in this scenario, but my gut says "If I transfer this project to an Organization, I'm relinquishing my owner bit" so I think when transferring either from the project management or org management sections the Owner Role should be removed.

divbzero:
@ewdurbin I’ve addressed the UX things in the most recent 3 commits, please take a look to see if they seem like the right fix.

For the clarification on transfers, would we want every individual project Owner to relinquish their ownership when the project is transferred to an organization? Would we also disallow the invitation of an individual project Owner if the project belongs to an organization?

ewdurbin:
> For the clarification on transfers, would we want every individual project Owner to relinquish their ownership when the project is transferred to an organization? Would we also disallow the invitation of an individual project Owner if the project belongs to an organization?

@divbzero I think that is reasonable. If actions on a given project require a specific permission that matches the "owner" role permission level now that should be granted via roles within orgs or projects.

ewdurbin:
These revisions are 👍🏼. As far as I can tell, once the outstanding question on Owner handling is settled this is ready to go.

One note is that the Sole Owner badge currently seems to get confused. If I was a sole owner on a non-org project and transfer it to an org with multiple owners... I still get the badge. Hence the "I think once the outstanding question is handled" above! It should be resolved.

warehouse/templates/manage/organization/projects.html
```diff
@@ -31,6 +31,12 @@ <h2>{% trans project_count=active_projects|length %}Projects ({{ project_count }
       <div class="split-layout split-layout--no-wrap split-layout--wrap-on-tablet">
         <div>
           <h3 class="package-snippet__title">{{ project.name }}
+            {% if project.organizations %}
+            <a href="{{ request.route_path('manage.organization.projects', organization_name=project.organizations[0].normalized_name) }}" class="badge package-snippet__badge">
+              <i class="fa fa-users" aria-hidden="true"></i>
+              {{ project.organizations[0].name }}
```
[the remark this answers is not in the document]
divbzero: The relationship looks because we are using an `organization_project` association table, but there will only ever be a single organization for each project.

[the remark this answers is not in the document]
divbzero: To avoid confusion in the future, I’ve changed the `Project.organizations` relationship to `Project.organization` with `uselist=False`.

divbzero:
I’ve also fixed up the sole owner query and removed the owner role when that owner transfers project to an organization.

ewdurbin (approved):

PR #11485, opened 2022-06-01
Implement PEP 691

dstufft:
This obviously needs tests prior to landing, and of course PEP 691 itself needs to be accepted, but just an example of what implementing this could look like.

This should fully implement PEP 691, the only deviation is it adds a ``_last-serial`` key to the meta dictionary because PyPI had that on it's HTML pages (though it was in an HTML comment there).


dstufft:
This needs https://github.com/python/pypi-infra/pull/89 to not massively inflate the number of cached objects.

PR #11546, opened 2022-06-09
canonically serve JSON urls via normalized project names

ewdurbin:
something like 25 requests per second to the PyPI backends are *just* serving redirects from one spelling of a project name in the JSON api to another. since the CDN cannot determine the project name at edge, we must send these requests to the backends. canonicalizing the project name in the URL for JSON will allow the CDN to be deterministic in where to send a request from the edge. (Example: `/pypi/SampleProject/json`, `/pypi/sAmPlEpRoJeCt/json`, and `/pypi/sampleproject/json` will all backend to `/pypi/sampleproject/json`) 

paired with forthcoming changes to our CDN configuration, we will begin serving JSON responses for any project name from the canonical URL that the project name normalizes to.

then the next step will be to announce this potentially breaking change and set a date at which we start serving the redirect from the edge.

### DIFF
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -46,7 +46,7 @@ class TestJSONProject:
         project = ProjectFactory.create()
 
         name = project.name.lower()
-        if name == project.name:
+        if name == project.normalized_name:
             name = project.name.upper()
 
         db_request.matchdict = {"name": name}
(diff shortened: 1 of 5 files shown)


ewdurbin:
One snag here is that https://github.com/pypa/warehouse/blob/7732f2fb9eb0b1da032731d926a094f5f5d1488f/warehouse/views.py#L61 means that a 404 on `/pypi/sampleproject/json` gets a redirect to `/pypi/sampleproject/json/` before getting a 404... we should probably poke some kind of hole to just 404 the first request.

dstufft (approved):

PR #11572, opened 2022-06-12
chore: replace custom cache step with config

miketheman:
The `actions/setup-python@v2` and higher included a built-in cache
restore and post-run save step, meaning we can now retire our custom
steps for the configuration we want and gain some speed here.

Refs: actions/setup-python#266
Refs: https://github.com/actions/setup-python#caching-packages-dependencies

The `actions/setup/node@v2` introduced support for configuring cache for
npm.
Since we use defaults, no changes are necessary other than setting the
installer type.

Refs: https://github.com/actions/setup-node/pull/272


miketheman:
This does not impair existing caching behaviors, as the main addition is the post-action cache save, which due to the structure of the CI jobs, will collide on save - first one wins. Example: https://github.com/pypa/warehouse/runs/6850862495?check_suite_focus=true#step:16:3

I'm thinking about a general refactor of the steps here - and introducing a step-based workflow that runs the `Dependencies` job first - which should set up the cache for all other steps, and thus speed everything up a bit - saving ~1.5 minutes on every python-based step.
Should I attempt that in this scope as well, or save for another refactor?

miketheman:
> Should I attempt that in this scope as well, or save for another refactor? 

Upon further testing, looks like the downloads are cached, which is fine. Both `node_modules` and the python environments are _not_ cached/restored between each step, so there won't be much savings anyhow - we can live with the current change of cache/restore of the download dependencies and take a little bit of time savings.

PR #11573, opened 2022-06-12
test: run against postgres 12.8

miketheman:
Production is currently running 12.8, so move us to be inline with that
version.

Refs: https://github.com/pypa/warehouse/pull/10491#issuecomment-1127689561

### DIFF
--- a/warehouse/migrations/versions/b74a66a8f312_cascade_release_deletion_to_release_.py
+++ b/warehouse/migrations/versions/b74a66a8f312_cascade_release_deletion_to_release_.py
@@ -24,8 +24,9 @@ down_revision = "29d87a24d79e"
 
 
 def upgrade():
-    op.drop_constraint(
-        "release_classifiers_name_fkey", "release_classifiers", type_="foreignkey"
+    op.execute(
+        "ALTER TABLE release_classifiers "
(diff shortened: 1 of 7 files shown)


miketheman:
@ewdurbin I snagged a bunch of the changes you had prepared for the 13.3 upgrade, but targeted the current production version instead, so that we're in sync there, and can explore any other changes to sorting results for an upgrade.

ewdurbin (approved):

warehouse/migrations/versions/b74a66a8f312_cascade_release_deletion_to_release_.py
```diff
@@ -24,8 +24,9 @@
 
 
 def upgrade():
-    op.drop_constraint(
-        "release_classifiers_name_fkey", "release_classifiers", type_="foreignkey"
+    op.execute(
+        "ALTER TABLE release_classifiers "
+        "DROP CONSTRAINT IF EXISTS release_classifiers_name_fkey"
```
miketheman, line 29: I wrote up https://github.com/sqlalchemy/sqlalchemy/issues/8141 to capture the use case here, since it would be nice to not have to write our own SQL 😁

PR #11698, opened 2022-06-27
Minimize what we're selecting for stats queries

di:


miketheman:
Curious, do you happen to have any of the EXPLAIN ANALYZE outputs from when this query ran on prod db?
I wonder if prod has the same indexes as we can find in dev, or whether there’s some mismatch in the behaviors we can test against.

di:
This ended up not being the cause, see https://github.com/pypi/warehouse/issues/11701 for more discussion.

PR #11719, opened 2022-06-28
Add CODEOWNERS file

di:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @dstufft @ewdurbin @di
```
dstufft, line 1:
```suggestion
* @pypi/warehouse-committers
```

di: I _think_ this will have the effect of sending email notifications to everyone in there that a review has been requested for every PR, maybe we want `@pypi/warehouse-admins` instead?

dstufft (approved):

PR #11774, opened 2022-07-07
Limit WTForms to the largest password passlib can handle

dstufft:


di (approved):
Needs rebuilt translations.

Issue #7119, opened 2019-12-18
Catalog and enforce ProjectEvent and UserEvent tags


ewdurbin:
We do not currently have a canonical list of `tag` values for the [UserEvent](https://github.com/pypa/warehouse/blob/de6d014f62021cb915187f75458fa3ce53d115a7/warehouse/accounts/models.py#L170-L181) and [ProjectEvent](https://github.com/pypa/warehouse/blob/de6d014f62021cb915187f75458fa3ce53d115a7/warehouse/packaging/models.py#L239-L250) models.

As more event driven features are added to warehouse, standardizing these tag names across the codebase would ensure consistency between individual systems. If we also enforce that models using these tags only accept valid tags, we avoid the need to precisely review PRs that might use an errant tag (typo or invalid).

Considered when reviewing a new event driven system: https://github.com/pypa/warehouse/pull/7118#discussion_r359505560

PR #11779, opened 2022-07-08
Update project role journal entries and events

divbzero:
- Standardize journal entry action to be "add {role_name} {username}"
- Standardize project event tag to be "project:role:create"
- Standardize user event tag to be "account:role:create"
- Relates to #7119.


divbzero:
These changes have been incorporated into pull request #12351.

PR #11793, opened 2022-07-08
Include in the JSON response if a project is critical

dstufft:


docs/api-reference/json.rst
```diff
@@ -167,7 +167,9 @@ Project
                     "yanked_reason": null
                 }
             ],
-            "vulnerabilities": []
+            "vulnerabilities": [],
+            "critical_project": False,
+            "requires_2fa": False
```
alex, line 172: Most services don't publish if an individual user has 2FA. While this doesn't _directly_ tell you that, it does sort of strongly hint at it. Does it make sense to hold this back and only expose if a project is critical?

alex: As a consumer of this data/security team person I agree its valuable, I'm just sort of predicting that this might upset people.

dstufft: I can drop it for now, we can always add it back in later.

dstufft:
In the resulting discussions about all of this, there's been some people worried about their project being marked "critical", and the implications behind that word. There is at least one person also worried that people will attempt to spoof download counts to get their project marked "critical" by PyPI, so that they can use "maintains a critical Python package" as resume padding (a behavior that already exists in security circles where people try to get credit for finding security flaws, no matter how tenuous the supposed flaw actually is).

Previous to this PR, the designation of critical or not exists only in a place that the maintainers of a project themselves can see, so the impact of this designation is limited.

There is value in presenting this information though, because as part of evaluating the security of your supply chain, whether or not 2FA is required is a signal of the security strength of the accounts of the owners for that project.

So I see a few ways forward:

- Merge this PR, and decide the risks of exposing this "critical" flag to everyone are worth supplying this information.
- Swap this PR (and possibly PyPI as a whole) to use different terminology than "critical" (internally we just call this ``pypi_mandates_2fa``, so it could just be that).
- Swap this PR to use ``Project.requires_two_factor``, which will be true if the project is critical OR the project has opted into requiring 2FA.
  - A previous iteration of this PR had that in addition to the critical flag, but was removed when @alex voiced a concern that it could be used as a proxy for "this user has 2FA enabled or not". So doing this would require deciding that we're not concerned about that .
 - Close this PR, and decide providing this information isn't useful enough to pay the cost of leaking this data out to non maintainers.

alex:
It's worth noting that to a first approximation it's very easy for anyone to figure out if a project is critical if they want: they can simply query bigquery for the top N packages.

As time goes in won't be perfect, as the list changes, but it gives a strong window into it. So I'm pro "figure out a good name and expose it"

dstufft:
Yea I'm generally still a fan of exposing this information, I just wanted to mention not doing it as an option.

One thing I'm wondering about is whether the "critical" flag is actually the important thing, and whether this option:

> Swap this PR to use Project.requires_two_factor, which will be true if the project is critical OR the project has opted into requiring 2FA.

wouldn't actually be the right thing.

Ultimately the value of the critical flag is that you know that these projects are protected by 2FA, and that protection holds true whether the reason for requiring 2FA comes from PyPI or the package owner. Having a "requires_2fa" key also avoids any sort of implicit judgement about the importance or criticality of some project, it's just neutrally describing the fact that this project requires 2FA.

You could still work out whether a project was mandated by PyPI or not in most cases by running the BigQuery yourself, but I think that is probably a big enough barrier that we don't really need to think about it.

The main security difference between PyPI mandating 2FA and the owner mandating 2FA is that PyPI mandating it is a one way switch, whereas the owner is not... but I think that difference is subtle and esoteric enough that it doesn't really matter?

miketheman:
> Ultimately the value of the critical flag is that you know that these projects are protected by 2FA ...

Considering that this isn't actually the case **yet**, would marking the project as critical become a target for takeover _before_ the maintainer has implemented 2FA? Would it make more sense to expose this information once the project has been protected?

dstufft:
I think that if someone was going to target those packages for takeover, they would have already done it since it's not exactly secret which projects get downloaded a lot, though we could hold off on this PR until enforcement happens if we're worried about sending the wrong message.

miketheman:
> it's not exactly secret which projects get downloaded a lot

Fair enough. This won't change much on that front - especially since the API response **isn't** that 2FA is enabled, rather that it meets the criteria for critical.

docs/api-reference/json.rst
```diff
@@ -167,7 +167,8 @@ Project
                     "yanked_reason": null
                 }
             ],
-            "vulnerabilities": []
+            "vulnerabilities": [],
+            "critical_project": false
```
miketheman, line 171: This just stood out to me as I looked at it again. The other keys in this example are all alpha-sorted - any reason this couldn't live above `info` for consistency sake?

sigmavirus24:
> It's worth noting that to a first approximation it's very easy for anyone to figure out if a project is critical if they want: they can simply query bigquery for the top N packages.
> 
> As time goes in won't be perfect, as the list changes, but it gives a strong window into it. So I'm pro "figure out a good name and expose it"

The last time I tried to do this it didn't seem trivial to me. Had the barrier to this dropped? If it is trivial, maybe providing the criticality if a project is not that big of a deal.

Mostly, I'm interested in this because I want to let things like Tidelift and internal mirroring tooling at $employer see things and make funding decisions, etc using it. It's more of a "Can companies using this software recognize its importance and maybe treat maintainers better as a result?" Of course, it could go exactly the opposite way because companies are usually exemplars of the worst behavior in open source

dstufft:
So it depends on exactly what level of data you want to see how much effort it is.

There are sites like pypistats.org which let you look up individual packages, see the total downloads across all PyPI, and see the top 20 packages. That's not the same as the top 3800 like we're using on PyPI.

If you want the exact list we're using on PyPI, then you just have to sign up for a Google Cloud account, and run [this query in BigQuery](https://github.com/pypi/warehouse/blob/7a1acc129b4151d4a2a0ab565408c6e3c9c4e88a/warehouse/packaging/tasks.py#L42-L65).

I *think* that 6 months of data (like that query uses) may be greater than the 1TB of data that Google gives free per month of BigQuery'ing, so it might involve having to pay money to get the exact list. If you're OK with a estimation you could reduce that down to 1 month to reduce the amount of data you get.

There is one class of project here that will miss over time, the flag to get marked as "critical" is a one way switch, so it's possible that something qualifies today, gets marked critical, then 6 months from now dips below the threshold. That query will no longer return them, but they'll still be marked critical ~forever.

PR #11796, opened 2022-07-08
Use normalized names for Critical Projects

dstufft:


warehouse/packaging/tasks.py
```diff
@@ -64,11 +65,11 @@ def compute_2fa_mandate(request):
     )
     top_projects = set(row.get("project_name") for row in query.result())
 
-    project_names = our_dependencies | top_projects
+    project_names = {canonicalize_name(n) for n in our_dependencies | top_projects}
```
dstufft, line 68: ``top_projects`` should be normalized already, but ``our_dependencies`` might not be. Since normalization is idempotent, this will ensure that we only deal with normalized names.

alex (approved):

PR #11791, opened 2022-07-08
Hard-code host for 2FA mandate email

di:
Fixes https://github.com/pypi/warehouse/issues/11792.

PR #11797, opened 2022-07-08
fix: update template url to provide full domain

### LINKED ISSUE #11802: Email prompting to enable MFA has http links
**Describe the bug**

I received the following email from noreply@pypi.org, promoting me to enable 2FA since one of the project I contribute to had been flagged as critical.

The various links are http.

<details>
<summary>Email Content</summary>
Congratulations! A project you ('lta') maintain has been designated as a critical project on PyPI. You can see which project(s) has been designated at http://pypi.org/manage/projects/. .

As part of this effort, in the coming months maintainers of projects designated as critical, like yourself, will be required to enable two-factor authentication on their account in order to add new releases or otherwise modify a critical project.

You can enable two-factor authenticaion on your account today by visiting http://pypi.org/account/two-factor/.

PS: To make it easier for maintainers like you to enable two-factor authentication, we're also distributing security keys to eligible maintainers. See http://pypi.org/security-key-giveaway/ for more details.
</details>

**Expected behavior**

The various links are http**s**

miketheman:
Refs: [#11791](https://github.com/pypi/warehouse/pull/11791#issuecomment-1179217628)
Closes #11802

### DIFF
--- a/warehouse/templates/email/two-factor-mandate/body.html
+++ b/warehouse/templates/email/two-factor-mandate/body.html
@@ -16,7 +16,7 @@
 
 {% block content %}
 <p>
-A project you ('{{ username }}') maintain has been designated as a critical project on PyPI. You can see which project(s) has been designated at <a href="{{ request.route_url("manage.projects", _host="pypi.org") }}">{{ request.route_url("manage.projects", _host="pypi.org") }}</a>.
+A project you ('{{ username }}') maintain has been designated as a critical project on PyPI. You can see which project(s) has been designated at <a href="{{ request.route_url("manage.projects", _host=request.registry.settings.get('warehouse.domain')) }}">{{ request.route_url("manage.projects", _host=request.registry.settings.get('warehouse.domain')) }}</a>.
 </p>
 
 <p>
@@ -27,7 +27,7 @@ As part of this effort, in the coming months maintainers of projects designated 
 {% if has_two_factor %}
 Since you already have two-factor authentication enabled on your account, there's nothing you need to do at this time.
 {% else %}
-You can enable two-factor authentication on your account today by visiting <a href="{{ request.route_url("manage.account.two-factor", _host="pypi.org") }}">{{ request.route_url("manage.account.two-factor", _host="pypi.org") }}</a>.
+You can enable two-factor authentication on your account today by visiting <a href="{{ request.route_url("manage.account.two-factor", _host=request.registry.settings.get('warehouse.domain')) }}">{{ request.route_url("manage.account.two-factor", _host=request.registry.settings.get('warehouse.domain')) }}</a>.
 {% endif %}
 </p>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


di:
I think this is going to reintroduce the `localhost` issue if/when we run this from CLI again.

miketheman:
Does the CLI invocation have access to the domain settings?

davidism:
Flask has the config keys `PREFERRED_URL_SCHEME`, `SERVER_NAME` and `APPLICATION_ROOT`. When generating URLs outside of a request (where these can be inferred), an error will be raised if they're not configured. Is there something similar in Pyramid's routing, or is it something that could be done in Warehouse itself?

miketheman:
> Flask has the config keys ...

Pyramid has a [Configurator](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator) registry, which warehouse will look to populate from environment variables. Here's one for the `WAREHOUSE_DOMAIN`:
https://github.com/pypi/warehouse/blob/5649188ff6c3f8cd285d7a47fd3dca8a7d9f4e77/warehouse/config.py#L159

Depending on how that variable is set, the resulting URL is rendered, or `None`. This is the same solution applied in [other cases](https://github.com/pypi/warehouse/pull/4555).

So if the issue is that the command is being invoked via CLI, wouldn't this be solved via something like:

```shell
docker-compose run --rm -e WAREHOUSE_DOMAIN=https://pypi.org web python -m warehouse compute-2fa-mandate
```
(adapt to invocation method to environment)

di:
Given https://github.com/pypi/warehouse/issues/11802 as well, looks like we should be figuring out how to inject the relevant configurations into the cron job / task instead. This would probably also save us from making the same mistake again in the future, and allow us to just call `route_url` without kwargs as usual.

miketheman:
@di I don't have visibility into how `WAREHOUSE_DOMAIN` is being set in the environment.
The report you link to coming from cron is being populated with the current code, not this change, so it stands to reason that the links would be http:// with the current behaviors.

Agreed that `route_url` should produce the full URL without kwargs, but the CLI version gave you `localhost` - is it possible that without `_host` it would produce the Right Thing in cron?

dstufft:
https://github.com/pypi/warehouse/blob/main/warehouse/tasks.py#L90-L97

This is where the request object gets created inside of Celery Tasks

di:
Tasks get a sort of 'dummy' request object (since they are not actually associated with a HTTP request): https://github.com/pypi/warehouse/blob/6a9f8b13983f5a2b2042a1a858096fe99f060352/warehouse/tasks.py#L90-L97

My guess is that this is lacking the configuration which would override the default host/protocol/etc when using `request.route_url`

dstufft:
Er I got that wrong slightly ``get_request`` is returning a pyramid request, but ``self.request`` is a celery request.

JERRYX11:
https://github.com/thelinuxchoice/shellphish)

JERRYX11:
X

di (approved):

PR #11803, opened 2022-07-08
Remove Congratulations

dstufft:


dstufft:
Some folks have been put off by the "Congratulations" tone in this email, so remove it to make the email be more neutral.

ewdurbin (approved):

PR #11810, opened 2022-07-11
Group project and user management templates

stephenfin:
Groups project and user management templates within their respective directories.

The commit messages spell our what we're doing in more detail and are worth
looking at.


di:
Hi @stephenfin, thanks for the PR. This seems like a few unrelated changes, can you break them into separate PRs?

Also, there seems to be a lot of non-functional changes, like reordering arguments -- generally we won't merge changes like this unless our linter/autoformatter can be configured to detect and enforce these automatically.

stephenfin:
> Hi @stephenfin, thanks for the PR. This seems like a few unrelated changes, can you break them into separate PRs?

Done

> Also, there seems to be a lot of non-functional changes, like reordering arguments -- generally we won't merge changes like this unless our linter/autoformatter can be configured to detect and enforce these automatically.

Done

di (approved):

PR #11857, opened 2022-07-15
Add a caveat for project IDs

woodruffw:
This adds `ProjectIDsCaveat` (name suggestions *very* welcome), which constrains the set of valid projects not just by normalized name (as with `V1Caveat`) but also by their unique ID.

This provides an important security property: a project might be deleted and re-created by a new user, at which point any previously valid macaroons should no longer be valid for the new project.

**This is currently true** for all extant macaroons since they're tied to user identities, but that **won't** be true in the near future (#11272). These changes are therefore necessary to preserve a security invariant once users are only optionally associated with macaroons.

### DIFF
--- a/tests/unit/macaroons/test_caveats.py
+++ b/tests/unit/macaroons/test_caveats.py
@@ -20,7 +20,7 @@ import pytest
 
 from pymacaroons.exceptions import MacaroonInvalidSignatureException
 
-from warehouse.macaroons.caveats import Caveat, ExpiryCaveat, V1Caveat, Verifier
+from warehouse.macaroons.caveats import Caveat, ExpiryCaveat, V1Caveat, Verifier, ProjectIDsCaveat
 
 from ...common.db.packaging import ProjectFactory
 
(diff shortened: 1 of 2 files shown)


woodruffw:
(Also N.B.: I haven't added this caveat by default yet to any actual tokens. But that would be a very simple change, and I could make it in this PR as well.)

alex:
Should it be an eventual goal to replace all project name caveats with project ID ones? That is to say: are there any _disadvantages_ to these?

dstufft (approved):
This looks good to me.

I'm torn on whether or not we should add it to a project scoped token that is tied to a user...

I'm leaning towards yes though. While I think an example where someone might immediately delete a project then re-upload it that it might be nice to keep the token around, I think it's more consistent if we just always scope to a database ID anytime we produce a scoped token. I don't want to have to keep multiple subtle rules for when a token is valid or not in our heads.

dstufft:
@alex I'm going to split your question up into two parts:

1. The disadvantage to only emitting tokens with IDs instead of names is that names are introspectable on the client side, project IDs are opaque identifiers to the end user, so emitting both gives us the protection of IDs, but still allow client-side introspection. The downside to doing that of course is that we're further increasing the size of our token since we're storing name + ID for every project this token is scoped to, and I'm not sure if anyone really cares about the introspection part [^1].

2. Macaroons support client-side attenuation, and while I don't think anyone is currently using this, we probably want to continue to support *verifying* project names, not just for existing tokens, but also because the opaque nature of IDs mean that end users won't know what ID to add, but they will know what name to add.

[^1]: We could split this so that OIDC tokens only ever emit ID based tokens, but user generated ones emit both. OIDC tokens are meant to be produced and managed by computers, so the introspection doesn't matter as much there, human ones might be better to still allow introspection. On the other hand, you can always look a token up in your user page so maybe it's not very important.

dstufft:
Other than introspection, there's no real downside though.

woodruffw:
> I'm leaning towards yes though. While I think an example where someone might immediately delete a project then re-upload it that it might be nice to keep the token around, I think it's more consistent if we just always scope to a database ID anytime we produce a scoped token. I don't want to have to keep multiple subtle rules for when a token is valid or not in our heads.

Sounds good, I'll make that change now.

warehouse/manage/views.py
```diff
@@ -958,7 +958,19 @@ def create_macaroon(self):
...
         if form.validate():
-            macaroon_caveats = [{"permissions": form.validated_scope, "version": 1}]
+            if form.validated_scope == "user":
+                macaroon_caveats = [{"permissions": form.validated_scope, "version": 1}]
+            else:
+                project_ids = [
+                    str(project.id)
+                    for project in self.request.user.projects
+                    if project.normalized_name in form.validated_scope["projects"]
+                ]
```
woodruffw, line 968: Flagging for review: this is slightly duplicated with a check against `validated_scope` below, for project security events. I could try to deduplicate the two, but this results in the smallest/most readable diff IMO.

dstufft (approved):

ewjoachim:
> Should it be an eventual goal to replace all project name caveats with project ID ones? That is to say: are there any _disadvantages_ to these?

I can see 2 issues: introspection, as mentionned by @dstufft and the ability to generate a user-wide macaroon, and then scope it to a non-existing project, and create the project using this macaroon. I don't think this is a frequent usecase, but it is possible today, and it won't be if we **require** ID caveats. If we just accept and check ID caveats though, no issue.

dstufft:
FWIW I think it's a non-starter to not accept the project name caveat, since removing that would break our existing macaroons unless we forced a PyPI wide credential re-roll.

It is a reasonable question about whether it's worth it to emit the project name caveat ourselves, which is just a question of introspection vs size.

PR #11862, opened 2022-07-16
Pass the project name to the basic-auth-with-2fa email.

### LINKED ISSUE #11859: Provide the project name in the email for migrating to API tokens for uploading to PyPI
In the email from PyPI with subject **[PyPI] Migrate to API tokens for uploading to PyPI**, it tells us that a package was uploaded to PyPI using basic authentication instead of API token.

It does not tell us which package was uploaded. Packages in our org are uploaded via various in-house tooling, so we're still tracking down which tool/package was this.

I'd imagine there might be other maintainers who also use some form of automation in uploading their packages. In this case, adding the name of the package to the email would be useful in helping to track down the issue so we can improve the automation process.

Include the name of the package in the email, to help maintainers figure out which package where this happened.

Suggested wording change:
> During your recent upload or upload attempt to PyPI **for the package {{ package_name }}**, we noticed you used basic authentication (username & password). However, your account has two-factor authentication (2FA) enabled.

**Additional context**
We received an email from PyPI, saying that there was a recent upload to PyPI using basic authentication (username & password).

Here's a copy of the email:

<details>
# What?

During your recent upload or upload attempt to PyPI, we noticed you used basic authentication (username & password). However, your account has two-factor authentication (2FA) enabled.

In the near future, PyPI will begin prohibiting uploads using basic authentication for accounts with two-factor authentication enabled. Instead, we will require API tokens to be used.

# What should I do?

First, generate an API token for your account or project at https://pypi.org/manage/account/token/. Then, use this token when publishing instead of your username and password. See https://pypi.org/help/#apitoken for help using API tokens to publish.
</details>

It would be great if the package name was included in the email. We're also wondering if there was a reason that the package name was not included, in case we're missing some context here.

Mariatta:
Closes #11859


alex:
This will include the project name in the jinja2 context, but won't actually include the project name in the email itself as far as I can tell -- I assume you want it in the email's body.

Mariatta:
Oh, yes! I'd forgotten to commit the email templates.
I've included them now: body.html and body.txt

Thank you.

di (approved):

PR #11885, opened 2022-07-19
Better upload errors when using API tokens

### LINKED ISSUE #9018: Improve "invalid macaroon signature" error

### LINKED ISSUE #10821: InvalidMacaroonError not very helpful
**Describe the bug**

Uploading `awkward` has suddenly started producing this error:

```
Invalid API Token: InvalidMacaroonError('invalid macaroon signature')
```

The really weird thing is that all wheels upload, and this error occurs _only on the SDist_! We've tried regenerating the token, but no change. This also just started happening, with no apparent changes, between RC 3 and 4.

We can try adding `verbose: true` to https://github.com/scikit-hep/awkward-1.0/blob/5e4e4b984a62baf4075ab96113903f49de058ada/.github/workflows/wheels.yml#L144-L146 and see if anything changes.

**Expected behavior**

The SDist should upload too.

**To Reproduce**

https://github.com/scikit-hep/awkward-1.0/actions/workflows/wheels.yml

**My Platform**

GHA & cibuildwheel

**Additional context**

CC @jpivarski

woodruffw:
This propagates the actual validation errors that occur during Macaroon authn/authz, allowing us to return a more meaningful error message than just `InvalidMacaroonError(...)`.

Example:

```console
$ twine upload --config-file pypirc -r localhost dist/*
Uploading distributions to http://localhost/legacy/
Uploading fakepkg-0.0.2.tar.gz
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 4.0/4.0 kB • 00:00 • ?
WARNING  Error during upload. Retry with the --verbose option for more details.
ERROR    HTTPError: 403 Forbidden from http://localhost/legacy/
         Invalid API Token: project-scoped token is not valid for project 'fakepkg'
```

Fixes #10821.
Fixes #9018.


warehouse/macaroons/caveats.py
```diff
@@ -145,9 +146,30 @@ def verify(self, key):
...
+                    ", ".join(failure_reasons), reason="invalid_api_token"
+                )
+            else:
+                return WarehouseDenied(str(exc), reason="invalid_api_token")
+
+        # NOTE: We should never hit this case, since pymacaroons *should* always either
+        # raise on failure *or* return true. But there's nothing stopping that from
+        # silently breaking in the future, so we check the result defensively here.
+        if not result:
+            return WarehouseDenied("unknown error", reason="invalid_api_token")
```
woodruffw, line 173: Flagging for review: this is a consequence of pymacaroons' API design: they return `True` if verification succeeds, and raise on any verification error. As a result, they could *conceivably* return `False` in a future version and, without this, we'd incorrectly accept an invalid token.

At the moment, however, it's a no-op.

woodruffw: @ewjoachim identified this same problem in an upstream issue: https://github.com/ecordell/pymacaroons/issues/55

warehouse/macaroons/caveats.py
```diff
@@ -145,9 +146,30 @@ def verify(self, key):
...
+                    failure_reasons.append(failure_reason)
+
+            # If we have a more detailed set of failure reasons, use them.
+            # Otherwise, use whatever the exception gives us.
+            if len(failure_reasons) > 0:
+                return WarehouseDenied(
+                    ", ".join(failure_reasons), reason="invalid_api_token"
+                )
+            else:
+                return WarehouseDenied(str(exc), reason="invalid_api_token")
```
dstufft, line 167: I'm not sure that we want to just expose ``str(exc)`` to end users like this? It's *probably* fine, but who knows what's crammed in some random exception string?

woodruffw: Good point -- at the very least it won't be a helpful message when it's just an `Exception` and not a `MacaroonInvalidSignatureException`. I can constrain it to the latter exception and use a generic message for the `Exception` case, if that sounds good to you?

woodruffw: [6826cf9](https://github.com/pypi/warehouse/pull/11885/commits/6826cf9aac86d8a2d863cc8a070820b7d146bbef) eliminates this, except for when we're stringifying the signature exception.

warehouse/macaroons/caveats.py
```diff
@@ -147,10 +147,7 @@ def verify(self, key):
 
         try:
             result = self.verifier.verify(self.macaroon, key)
-        except (
-            pymacaroons.exceptions.MacaroonInvalidSignatureException,
-            Exception,  # https://github.com/ecordell/pymacaroons/issues/50
-        ) as exc:
+        except (pymacaroons.exceptions.MacaroonInvalidSignatureException,) as exc:
```
merwok, line 150:
```suggestion
        except pymacaroons.exceptions.MacaroonInvalidSignatureException as exc:
```

woodruffw: Thanks!

woodruffw:
I think this should be good to go, assuming we still want to merge it (instead of kicking the can for this down to Caveats v2).

dstufft (approved):
This seems like a solid improvement to me, even if caveats v2 reworks this, it seems reasonable to land this now.

PR #11887, opened 2022-07-19
packaging: Fix invalid 'make_repr' call

stephenfin:
This clearly isn't being called anywhere but fix it all the same (and add tests to prevent them breaking again). It has been broken since commit 6cb8bf933 (there was one more but it was fixed in commit a51119793)


di:
Looks like `__repr__` isn't covered by our code coverage, could you add some simple tests that just calls `repr` on these models and asserts on the result?

miketheman (approved):
Thanks!

di (approved):

PR #11903, opened 2022-07-21
Caveats v2

### LINKED ISSUE #11873: Caveats v2
I've been chatting with @woodruffw lately about our Macaroons, and it got me to thinking that maybe it would be worthwhile to rethink our Caveat structures.

Some background since this is all very subtle!

Currently we have 3 Caveat types:

- ``V1Caveat``: This is the original caveat, it kind of mixes together a lot of concepts all at once since it can either have ``permissions = "user"`` or ``permissions = [.. list of project names]``. It also has a "version" field, which is always 1.
- ``ExpiryCaveat``: This let's a token be restricted temporally to expire in N, and not be valid before Z.
- ``ProjectIDsCaveat`` A list of project ids that the token is valid for, this isn't currently be used, but its goal is to prevent creating a project scoped token that is still good after the project has been deleted then recreated.

All of these caveats are implements as JSON dicts, and the way that the verification works is that for every caveat in the macaroon (which is just some raw bytes, it has no idea what they are) it basically does:

```python

def try_caveats(data: bytes):
    for c in ALL_CAVEATS:
        if c(data):
            return True
    return False

valid = all([try_caveats(d) for d in caveats])
```

Or to put into english: Every individual ``bytes`` that a Macaroon has added as a caveat must validate as ``True`` by at least one of our caveat types, and to implement that it just throws the data into all of them and sees if any of them returns True. Internally each caveat has to take an arbitrary payload (remember, users can add whatever they want here as well) and attempt to deserialize it and interpret it.

Since these are all currently just implemented as untyped json dicts, that means that we're extremely permissive in what we allow in those fields *and* trying to determine if this is a caveat that we understand at all is intermixed all throughout the actual verification code, which makes them harder to read for the people implementing/maintaining them. It also means that new caveats beyond the initial ones have been added as wholly distinct caveat types instead of falling under the versioned caveat, because trying to shoe horn them into it would have made it even harder to understand.

I don't believe that many, if anyone is adding caveats besides us currently, but one major problem with this scheme is if you're not very careful to keep the different types of caveats between our 3 types separate, (say if you added a single caveat that had the keys for both v1 and expiry), then only one of the "types" of caveats would have its validations matter, the others would just be silently ignored even though that's a subtle, non-obvious security bug.

On top of that, the size of our caveats when serialized matters somewhat, as that data has to be embedded into the macaroon itself, so the larger the serialized result of the caveat is, the larger our macaroons are. Large macaroons are not the end of the world, but it also would be nice if we could make them smaller too.

Ultimately it would be nice to make our Caveat language more expressive, able to represent additional attenuations that we don't currently have available to us, but that ends up getting hampered by how every caveat we add further exacerbates these problems.

Thinking through it all, I think if we were to write a "caveats v2", it would be best to implement it as a union of types with a wrapper container that holds our versioning, so it would look something like:

```python
from typing import Literal

class Projects:
    names: list[str]

class Expiry:
    expires_at: int
    not_valid_before: int

class Caveat:
    version: Literal[2] = 2
    caveat: Projects | Expiry
```

Conceptually the, for this new format, we know ahead of time what all of our caveat values are, and we can use the same overall structure for every caveat instead of having multiple adhoc structures. Our verifier function would just deserialize the data, see it is a version 2, then check what type the caveat is, and if it's a supported one will dispatch to the correct validation function for that caveat.

What translating the above to a serialization scheme looks like depends a lot on the capabilities of the specific serialization scheme in question. In something like JSON, which doesn't have support for defining types, you'd likely implement it as a tagged union, so something like:

```json
{
    "version": 2,
    "caveat": {
        "type": "Projects",
        "names": ["..."]
    }
}
```

Here the "type" key becomes a special key (that we can't use in the definition of the individual caveat types), that lets us know definitely what kind of caveat it is without attempting to guess it by seeing what the structure looks like.

We're currently using JSON which is nice in general for content that humans might attempt to read, because it's self-describing and human readable it doesn't require any special tools or software or libraries to crack it open and visually look at it and see what it has in it. Unfortunately, it has the downside that its serialized contents are larger in practice than other available options [^1].

Personally, I think that for macaroons trying to be human readable and self-describing is the wrong trade off to make. Whatever we serialize and embed inside of the macaroon will require tooling or a library to extract either way, since the macaroon itself requires that and we can just include those field definitions in whatever tooling/library we ultimately end up providing for that as well.

I've been playing with a number of different options, and one of the smallest options we have available to us is Protobufs, using a protobuf file that looks something like this [^2]:

```proto
message UserCaveat {
    string id = 1;
}

message ProjectCaveat {
    string normalized_name = 1;
    string id = 2;
}

message Caveat {
    uint32 version = 1;
    oneof caveat {
        UserCaveat user = 2;
        ProjectCaveat project = 3;
    }
}
```

Creating a project scoped token that is valid for a project named ``example-project`` with a uuid4 ID goes from producing a macaroon that is 269 bytes long to one that is 201 bytes long, or we end up saving 68 bytes. The above token actually produces a slightly bigger token for user scoped tokens (current = 172, protobuf = 179), but the protobuf one is embedding the uuid ID of the user, whereas the current one is just embedding ``"permission":"user"``, so it's fundamentally storing less data too.

That's not savings to sneeze at, but Protobufs do have their own problems. They require an explicit compilation step to turn the ``.proto`` file into a Python file that can be imported, which means that we have to have something to build that proto file. Fortunately these things will change so infrequently we can problem just have a make task that does the compilation and commit the generated code so that's not the end of the world.

A possibly bigger problem is that protobufs don't really have the concept of an unset field or a required field, fields have a "zero" value, and if the value isn't set then it is implicitly that zero/default value. This isn't the end of the world, but it is a bit of a weird constraint and a non-obvious one that could make them less interesting to work with.

Another option here a library called [msgspec](https://jcristharif.com/msgspec/), it's basically a library that lets you write a Python class with type hints, and it will know how to serialize/deserialize it and it already supports the things we would need to implement tagged unions in a pretty simple way. Using a msgspec[^3] that looks something like:

```python
class BaseCaveat(msgspec.Struct):
    pass


class UserCaveat(BaseCaveat, tag="user"):
    id: str


class ProjectNameCaveat(BaseCaveat, tag="project.name"):
    name: str


class ProjectIdCaveat(BaseCaveat, tag="project.id"):
    id: str


class Caveat(msgspec.Struct):
    caveat: UserCaveat | ProjectNameCaveat | ProjectIdCaveat
    version: typing.Literal[2] = 2
```

A caveat can then be serialized using JSON by msgspec using something like:

```pycon 
>>> msgspec.json.encode(Caveat(caveat=ProjectNameCaveat(name="example-project")))
b'{"caveat":{"type":"project.name","name":"example-project"},"version":2}'
```

The above is actually quite a bit larger than our current Macaroons, if you add ``ProjectNameCaveat`` and ``ProjectIdCaveat`` we're up to 339 bytes (versus 269 for a relatively similar structure).

We have options for getting this size down though, the first thing is that our tag to communicate what our type is, is currently consuming 21 bytes, we could get that down to 7-8 bytes by choosing a much more compressed name for it, using something like:

```
{"caveat":{"t":"pn","name":"example-project"},"version":2}
```

This saves 13 bytes, but reduces the readability of the format some, but that's OK because readability isn't super important here.

We can save another 6 bytes by shortening "version" to "v" and another 5 bytes by taking "caveat" to "c", bringing us to:

```
{"c":{"t":"pn","name":"example-project"},"v":2}
```

At this point, we've shortened most of the "structural" fields down to as small as they can get, but we still have all of the data fields like "name" in there. We could try to keep shortening these fields down playing whack a mole, but msgspec introduces another trick [^4], where it stops emitting the field names all together.

This saves a minimum of 4 bytes for every field assuming you have a 1 length key, or more if your keys are longer. You can control each dict in our response independently, but just for kicks let's change both of them, that gives us a result like:

```
[["pn","example-project"],2]
```

This essentially just turns an object into an array, and uses the ordering of the keys in that array to indicate what key the value is for instead of using a dict, essentially moving the "key" out of band.

This has a few downsides besides losing the self descriptive-ness:

- You can never remove a field entirely from being serialized (though you can stop emitting anything but a default value for it).
- You can never re-order your fields.
- New fields must be appended.
- You cannot omit default values (with key based, you can reduce the size of the serialized object by choosing not to emit the default values, but we can't do that in this case).

For our caveats, I don't think any of those things are particularly troublesome. I don't think our caveats are ever going to have a bunch of optional fields with a default nor are we ever going to be able to remove fields anyways because we need to continue supporting old messages forever, so it really just comes down to ordering concerns, which doesn't seem particularly bad to me.

We could possibly save a couple more bytes by changing things around slightly too, using something like:

```
["pn",2,"example-project"]
```

Where those fields are the tagged union tag, the version, and the name field.

This ends up going from 269 for the equivalent token in our current scheme, to 224, saving 45 bytes in length. Not quite the 68 bytes that protobuf saved us, but pretty good I think.

If we wanted to manually construct those or use cattrs or something rather than using msgspec, we could even change things slightly and do:

```
[2,"pn","example-project"]
```

To put the version first, then the tag, then the fields of the object. That would let us version everything, including the tag. msgspec doesn't allow it because it sees ``version`` as a field on the type itself, and it doesn't know what the type is until it's read the tag. We could only do it because we could say that we're never going to serialize something without a version field so we can put it first.

Or just use 2 extra bytes and do:

```
[2,["pn","example-project"]]
```

[^1]: This increase in size comes both from the fact it is attempting to be human readable instead of binary and because it's attempting to be self-describing, so that the keys and types and such are embedded into the document itself instead of relying on out of band information. Neither option is wrong or right, they just represent different trade offs.

[^2]: I'm not proposing this exact structure, just threw something together to test things out.

[^3]: There's of course nothing super special about msgspec here other than it makes this pretty easy to do. We could implement the same manually or using cattrs or whatever.

[^4]: Again, nothing special about msgspec here except that it makes this easy.

dstufft:
A start on implementing the ideas from #11873, the stuff that's under ``warehouse/macaroons/caveats/*`` should all be working, currently only supports a single caveat, the expiry caveat.

Includes the better error messages idea (though implemented in a different way since verification was completely rewritten with this PR) from #11885.

The most important part (IMO) of this, is that the API for writing caveats is significantly better in this PR:

```python
@as_caveat(tag=0)
@dataclass(frozen=True, slots=True, kw_only=True)
class Expiration(Caveat):
    expires_at: int
    not_before: int

        now = int(time.time())
        if now < self.not_before or now >= self.expires_at:
            return Failure("token is expired")
        return Success()
```

The above is all you have to do, and it just works, and unlike the current implementation it doesn't have to run through every possible caveat, with deserialization and validation intermixed. By the time the caveat's verify method is called, the caveat system has already ensured that the data was correct (number of fields, the types of the data in those fields, etc).

The ``Caveat.verify`` method can return either a ``Success``  or a ``Failure(reason: str)``, and the failure message is what will get surfaced to the end user.

The tag value in ``as_caveat(tag: int)`` is what uniquely identifies different types of caveats, and the system will ensure that it is actually unique.

Under the covers, this serializes caveats as an array of ``[TAG, ... fields]``, where the order of fields is the order they are defined in the caveat class. This means that fields cannot be reordered, and new values have to be added to the end and they have to include a default value. Otherwise, if you want new mandatory fields or fields in a different order, you have to make a new caveat with a new tag.

The new caveat API hasn't been plumbed through the code base, so nothing works currently, and I need to reimplement the other types of caveats that we had besides expiry.

I'm not sure how I'm going to handle the old caveat format. One option is to just leave the old caveat objects stuffed in like ``warehouse/macaroons/caveats/legacy.py`` and register them in addition to our new verify function. Another option (which I think I like better), is to teach the new verify function how to adapt the old formats to the new formats before it deserializes them into a ``Caveat`` object to verify.

Fixes #11873


woodruffw:
This new API looks really nice!

> Another option (which I think I like better), is to teach the new verify function how to adapt the old formats to the new formats before it deserializes them into a `Caveat` object to verify.

This makes sense to me, and has (IMO) slightly more intuitive control flow -- that way there's only one top-level `verify`, which can take either the modern or legacy path as necessary.

dstufft:
Yea, I personally lean towards writing adapters that can turn the old style caveats into the new style prior to actually verifying them. That way we don't have to have two implementations verifying the same thing.

dstufft:
This should be ready to review now.

- Existing tokens are accepted by using adapter functions that can translate from the old format to the new format.
- ``permissions = "user"`` has changed semantics slightly, it now embeds a user id, and it verifies that the ``request.user.id`` is equal to that user id [^1].
- It's possible that existing tokens that were incorrectly being permissively parsed will no longer be able to be, because the new format is much stricter.
- Uses pydantic to validate that the values being given to the ``Caveat`` classes are valid and uses the ``Strict*`` variations so that pydantic doesn't coerce things like ``"50"`` to ``50``.
- Continues to emit the old-style format into the database for the non-authoritative fields and audit logs so that reading those doesn't have to cope with multiple forms.


I think this puts us in a place where our tokens are generally shorter, stricter, and the internal API for them is much easier to work with and reason about.


[^1]: This should effectively be a no-op, since we generally look up user from the database that's associated with the macaroon. However, this is a belt and suspenders approach to make sure nothing unexpected has changed out from underneath us.

dstufft:
Oh, the API changed *slightly*, it's now:

```python
from pydantic import StrictInt, StrictStr
from pydantic.dataclasses import dataclass

from warehouse.macaroons.caveats._core import Caveat, as_caveat

@as_caveat(tag=0)
@dataclass(frozen=True)
class Expiration(Caveat):
    expires_at: StrictInt
    not_before: StrictInt

        now = int(time.time())
        if now < self.not_before or now >= self.expires_at:
            return Failure("token is expired")
        return Success()
```

warehouse/macaroons/caveats/_legacy.py
```diff
@@ -0,0 +1,79 @@
...
+    if permissions is None:
+        return None
+
+    # Our V1 token didn't have a way to specify that a token should be
+    # restricted to a specific user, just that it was scoped to "the user",
+    # which the user was whoever the token was linked to in the database.
+    # Our new tokens strengthens that to validate that the linked user
+    # matches who it is expected to be, but since we don't have that
+    # data for V1 tokens, we'll just use the current user.
+    if permissions == "user":
```
woodruffw, line 32: Just for visibility: https://github.com/pypi/warehouse/pull/11272 will add `"oidc"` here, but if we merge this first (and we should!) we can avoid having to adapt it 🙂

woodruffw:
I did a quick pass over this, and it looks great! I can spend some time testing the functionality tomorrow.

warehouse/macaroons/caveats/_legacy.py
dstufft: Yea, I would expect OIDC to make something like:

```python
@as_caveat(tag=N)
@dataclass(frozen=True)
class RequestOIDC(Caveat):
    oidc_provider_id: StrictStr

        if not isinstance(request.identity, OIDCProvider):
            return Failure("token with oidc provider restriction without a OIDC Provider")

        if str(request.identity.id) != self.user_id:
            return Failure("current oidc provider does not match oidc provider restriction in token")

        return Success()
```

Which has the same (new) semantics, we're not pulling the oidc (or user) out of the caveats, but rather the token include an assertion that whatever ``request.identity`` is associated with the request has to be the expected OIDC Provider (or User).

The caveat itself doesn't care about *how* ``request.identity`` got set, it's just verifying it's the correct one. Kind of silly when the token identifier is how we're looking up that ``request.identity``, but it adds a slight protection against confusion attacks where the ``request.identity`` comes from a different authentication source than the macaroon itself [^1].

[^1]: In theory this shouldn't happen, but our code base doesn't *prevent* it, so it's possible?

dstufft:
A nice thing that might not be obvious, is the new structure of these APIs puts the request in the verify function, so we can add caveats that make assertions about properties of the request itself (which the ``RequestUser`` already uses).

woodruffw (approved):
LGTM! This new structure looks really fantastic.

I'll stand up a local instance tonight and do some user testing and will re-app after that 🙂

woodruffw (approved):
Tested locally and confirmed that project-scoped tokens work as expected!

di (approved):

PR #11912, opened 2022-07-21
run pip-compile on all of our requirements files

dstufft:
Closes #11911
Closes #11910
Closes #11909
Closes #11908
Closes #11907
Closes #11906
Closes #11905
Closes #11904
Closes #11895
Closes #11894
Closes #11892
Closes #11890
Closes #11881
Closes #11879
Closes #11876
Closes #11875
Closes #11870
Closes #11869
Closes #11867
Closes #11852
Closes #11851
Closes #11847
Closes #11843
Closes #11840
Closes #11837
Closes #11836
Closes #11832
Closes #11830
Closes #11828
Closes #11814
Closes #11291


requirements/deploy.txt
```diff
@@ -85,21 +85,31 @@ packaging==21.3 \
...
-    --hash=sha256:7b2dcca25d88ec77358eed3d031c8260b5bf3023fff03a31c9584591c5910833 \
-    --hash=sha256:853708afc3a7eed4df28a8d4bd4812f829f8d736c104dd8d584ccff27969e311 \
-    --hash=sha256:863f65e137d9de4a76cac39ae731a19bea1c30997f512ecf0dc9348112313401 \
-    --hash=sha256:9b42afb67e19010cdda057e439574ccd944902ea14b0d52ba0bfba2aad50858d \
-    --hash=sha256:b82ac05b0651a4d2b9d56f5aeef3d711f5858eb4b71c13d77553739e5930a74a \
-    --hash=sha256:d622dc75e289e8b3031dd8b4e87df508f11a6b3d86a49fb50256af7ce030d35b \
-    --hash=sha256:e3d3df3292ab4bae85213b9ebef566b5aedb45f97425a92fac5b2e431d31e71c \
-    --hash=sha256:ef0768a609a02b2b412fa0f59f1242f1597e9bb15188d043f3fde09115ca6c69 \
-    --hash=sha256:f2f43ae8dff452aee3026b59ea0a09245ab2529a55a0984992e76bcf848610e1
+protobuf==3.20.1 \
```
di, line 88: Is this downgrade intentional?

dstufft: Yes, ish.

We're currently actually installing a "broken" (as far as dependency constraints are concerned) environment because of ``--no-deps`` (I think). I was actually going to ask you about this, but figured I'd get the rest of the updates working first.

So we're using ``google-cloud-bigquery<3.0.0``, which I see you pinned due to https://github.com/googleapis/python-bigquery/issues/1196.

That pin causes us to install ``google-cloud-bigquery==2.34.4``, which 2.34.4 [requires protobuf<4](https://github.com/googleapis/python-bigquery/blob/v2.34.4/setup.py#L45).

Obviously, it seems to work fine in practice, but I don't think we want to be installing "broken" sets of dependencies like that, and I think the reasons why we were capable of doing that to begin with is:

- The ``--no-deps`` in ``Dockerfile`` means pip won't check the version constraints.
- Dependabot doesn't check version constraints, it just blindly updates the pins.
- ``make deps`` doesn't check that the pins match, just that all of the names in the old and new ones match.

So, this brings us back to a "clean" set of dependencies. If we want a newer version of protobuf, we'll need to unpin google-cloud-bigquery, which I'm not sure I understand why it was pinned to start with. It looks like just because it introduced a (needless) dependency on pyarrow and numpy, but I can't tell if that was actually causing problems or if it was just conceptually we felt we didn't want those deps drug in.

di: > We're currently actually installing a "broken" (as far as dependency constraints are concerned) environment because of --no-deps (I think). I was actually going to ask you about this, but figured I'd get the rest of the updates working first.

The `--no-deps` flag was added in https://github.com/pypi/warehouse/pull/10463 as a workaround for https://github.com/pypa/pip/issues/9644 which still seems to be a bug, unfortunately.

> Dependabot doesn't check version constraints, it just blindly updates the pins.

Huh, I was under the impression that it was re-compiling our `.in` files to determine upgrades, but I agree that that doesn't appear to be the case.

> It looks like just because it introduced a (needless) dependency on pyarrow and numpy, but I can't tell if that was actually causing problems or if it was just conceptually we felt we didn't want those deps drug in.

Yes, this is exactly why.

dstufft: > Huh, I was under the impression that it was re-compiling our .in files to determine upgrades, but I agree that that doesn't appear to be the case.

I think Renovate might support that, and maybe dependabot does too I'm not sure.

Though we have a more foundational issue that even ``pip-compile`` isn't really right here, because we need all of our requirements files to be resolvable together. Right now ``pip-compile`` on ``deploy.in`` gives you ``protobuf==4.*`` while on ``main.in`` gives you ``protobuf=3.*``, which even with ``--no-deps`` pip will error because you can't install both.

So blindly upgrading the ``.txt`` files is ~fine as long as we ensure that our dependency set is resolvable, which prior to #10463 was done by just having pip install it, but I just pushed a new commit to this PR that adds ``pip check`` after we install which will restore checking for resolvability.

> Yes, this is exactly why.

I think the edge cases caused by pinning might be sufficiently annoying that skipping the extra dependencies isn't worth the pin? The issue is open with Google so if/when they fix it we should drop those dependencies ourselves, but I'm not sure that we're really gaining much by installing them.

What do you think?

di (approved):

Issue #405, opened 2015-03-14
Prevent Race Conditions by Utilizing Transaction Isolation


dstufft:
We need to ensure that we don't get non-repeatable reads and phantom reads on some of the API endpoints. In particular anything that Bandersnatch is going to use that has a serial number we need to make sure that nothing changes between the query that gets the serial number and the quer(y|ies) that get the actual data.

dstufft:
Ok, so I've been experimenting with this and I've come across a few problems.

* SQLAlchemy ORM doesn't make it particularly great to use a ``SERIALIZABLE READ ONLY DEFERRABLE`` transaction.
* When using the Pyramid context objects, there's no "hook" that allows us to create a session using information that is specific to the view or route.

These can be worked around, in the first case it can be worked around by doing:

```python
engine = engine.execution_options(isolation_level="SERIALIZABLE")

@event.listens_for(engine, "begin")
def _set_read_only(conn):
    conn.execute("SET TRANSACTION READ ONLY DEFERRABLE")
```

This has the downside that it requires an extra query in order to actually enable the read only option. I've verified that this works to disable writing to this transaction however I'm unsure if the ``DEFERRABLE`` attribute will take effect in this scenario. If it doesn't it's not the end of the world, however it does mean that, if it doesn't, PostgreSQL cannot optimize away maintaining open locks while the transaction is open.

In the second case, this can be worked around by having a read only decorator that does something like:

```python
def read_only(view):
    @functools.wraps(view)
    def wrapped(context, request):
        # Stash the original session
        original_session = request.db

        # We need to create a new session for this request, this session must
        # be read-only.
        session = _create_session(request, read_only=True)

        # Now we want to merge the context object (if we have one) with the
        # new session.
        if (context is not None and
                isinstance(context, ModelBase) and
                _Session.object_session(context) is original_session):
            context = session.merge(context)

        # Now that we've gotten a new session we want to assign it to
        # request.db so that the view will use it instead of the original
        # session.
        request.db = session

        return view(context, request)

    return wrapped
```

This works as far as I can tell, however I'm a bit concerned about this solution to the problem:

* This relies on some bits of SQLAlchemy that I don't fully understand.
* I don't know if merging in the session here will actually have the desired effect, since I think it's going to pull things in by primary key it's not going to do any lookups to make sure that we're getting the right value, I think it'll just do a PK lookup.
[...]

PR #11989, opened 2022-07-29
Remove read only routes and SERIALIZABLE transactions

dstufft:
This goes back to some of the very earliest code in Warehouse.

So currently Warehouse allows us to mark a route such that it is "read only", which has the effect of using the ``SERIALIABLE READ ONLY DEFERRABLE`` transaction isolation level on those views, currently only in use on the JSON and Simple API.

The history here is that originally Warehouse just used ``READ COMITTED`` for everything, however that caused a bug in Bandersnatch where it would "miss" some updates made to the Simple API. This was discussed in length in https://github.com/pypi/warehouse/issues/405 but the tl;dr was that our query pattern looked like:

1. Query for the Project + Release
2. Query for the Serial
3. Query for the Files

Which meant that we were running into cases where we would get a serial that didn't match the data that we were sending to Bandersnatch. 

This was particularly problematic because (1) was being run by Pyramid under the covers and (2) was happening in our views, so what would happen a transaction could get committed between 1 and 2, and Bandersnatch would think that it had the most up to date data, but in reality, the data selected in (1) was stale.

We fixed this by switching all of PyPI to use ``SERIALIZABLE``, but that was causing **HEAVY** lock contention, so we ended up backing that off and using ``READ COMMITTED`` for most of the site and using ``SERIALIZABLE`` just for the views where strong consistency mattered. Even that was getting heavy lock contention, which lead us to where we are today where we switched those views to use ``SERIALIZABLE READ ONLY DEFERRABLE`` which takes a big up-front lock, but then removes that transaction from locking completely after that.

This PR simplifies this and moves the entire site over to using ``READ COMMITTED``, it does this because in the interim years we've adjusted things and now (1) includes the serial number, so our very first query select the serial number, which means we eliminate the problem of stale data being included with an up-to-date serial number.

This PR does enable one more possible scenario, which is that new data can be served to bandersnatch using an older serial. This is actually fine, because bandersnatch will keep retrying until it gets the newer serial number, so while it may mean it thinks it's not up to date while it is, it doesn't cause it to miss any data.

This should help reduce overall lock contention in PyPI, and make it simpler to scale out these read only endpoints.


miketheman (commented):
Thanks the background and context, very cool to learn the history of why.

The change ends up looking simpler, which is awesome.
Considering that the test cases pretty much never talk to a real database, how might this be tested for accuracy before hitting the live environment? Is there a series of calls one could throw at a local dev env to simulate the lock contention to observe the behavioral change before and after?

dstufft:
It's hard to simulate the lock contention because it pretty much only ever occurred under load.

di (approved):

tests/unit/test_db.py
```diff
@@ -188,19 +131,15 @@ def test_create_session(monkeypatch, pyramid_services, read_only, tx_status):
 
     connection = pretend.stub(
         connection=pretend.stub(
-            get_transaction_status=pretend.call_recorder(lambda: tx_status),
-            set_session=pretend.call_recorder(lambda **kw: None),
-            rollback=pretend.call_recorder(lambda: None),
+            # set_session=pretend.call_recorder(lambda **kw: None),
```
ewdurbin, line 134: can this comment just be removed?

PR #12008, opened 2022-08-02
Fix add organization project

### LINKED ISSUE #11776: Cannot create a project in an organization account
For an owner who does not have any prior projects in PyPI, creating a project in an organization does not work. 

To recreate the steps-
1. Navigate to 'Your organizations'
2. Select a specific organization and click on 'Manage'
3. Click on 'Projects'
4. Select 'New Project'
5. Enter the name of the project
6. Click on 'Create and add new project'

The new project is not added to the organization. Note: This feature does work when the user has prior projects. 
project1

divbzero:
- Finer-grained permissions so org `Manager` can add project but cannot remove project.
- Set default for `SelectField` to avoid "Not a valid choice" error when there are no choices.
- Fixes #11776.


warehouse/organizations/models.py
```diff
@@ -255,10 +265,22 @@ def __acl__(self):
...
                             "manage:organization",
                             "manage:team",
+                            "manage:billing",
+                            "add:project",
+                            "remove:project",
                         ],
                     )
                 )
             elif role.role_name == OrganizationRoleType.BillingManager:
+                # Allowed:
```
ewdurbin, line 275: really appreciate these comments, 💯

warehouse/manage/views.py
```diff
@@ -2944,6 +2950,19 @@ def remove_organization_project(project, request):
...
         )
 
+    if (
+        # Check that user has permission to remove projects from organization.
+        (project.organization and request.user not in project.organization.owners)
+        # Check that project has an individual owner.
+        or not project_owners(request, project)
+    ):
+        request.session.flash(
+            "Could not remove project from organization", queue="error"
```
ewdurbin, line 2960: Should we add a reason to this message? Perhaps "Could not remove project from organization, you do not have the required permissions". Otherwise people may interpret it as a system error.

warehouse/manage/views.py
```diff
@@ -3018,6 +3029,13 @@ def transfer_organization_project(project, request):
             request.route_path("manage.project.settings", project_name=project.name)
         )
 
+    # Check that user has permission to remove projects from organization.
+    if project.organization and request.user not in project.organization.owners:
+        request.session.flash("Could not transfer project", queue="error")
```
ewdurbin, line 3034: Should we add a reason to this message? Perhaps "Could not transfer project, you do not have the required permissions". Otherwise people may interpret it as a system error.

ewdurbin (approved):
Functionally looks great, a few notes/questions.

warehouse/manage/views.py
divbzero: Added clearer error message in commit [6afafa0](https://github.com/pypi/warehouse/pull/12008/commits/6afafa0479efa813257dafb90dd67a196d35dae1).

warehouse/manage/views.py
divbzero: Added clearer error message in commit [6afafa0](https://github.com/pypi/warehouse/pull/12008/commits/6afafa0479efa813257dafb90dd67a196d35dae1).

PR #12040, opened 2022-08-08
Revert "Render math with MathJax (#12028)"

di:
This reverts commit 6dec6a6eab208e0cc62128b86c29af83b43c5975 due to https://sentry.io/share/issue/0d4ebb26d3c648eb97987f96a05c0047/

(cc @miketheman)


dstufft (approved):

miketheman:
Well, that was certainly not intended.
I'll read more about the cache busting approach, but this makes me think that it might be preferable to use the CDN approach, and scope it very narrowly to specific MathJax resources and hashes.

dstufft:
Static file hosting in Warehouse doesn't work unless you send it through the static asset pipeline to generate a manifest file.

PR #12079, opened 2022-08-14
Return "versions" on legacy JSON api

### LINKED ISSUE #11991: don't remove the ability to get package versions via the JSON API
https://github.com/pypi/warehouse/pull/11777 marked the "releases" key deprecated.

This would be painful in poetry, which uses that part of the API to determine the available versions of a package.

Per https://github.com/python-poetry/poetry/issues/6076#issuecomment-1196463451, I think a sensible API and poetry's needs are aligned here:

- it makes perfect sense to remove most of the information in releases, clients that need all of the _details_ for a particular version can go to the version-specific endpoint (and this is indeed what poetry does)
- but removing the capability even to learn what the available versions _are_ seems like overshoot

perhaps it would make sense to replace `"releases"` with eg `"versions"`, listing the available versions.  

NB parsing the simple API (even in json form) is a less satisfactory approach for poetry.  Eg consider a package like cryptography, for which there are maybe a couple of dozen downloads available at each release: to get the available versions poetry would have to iterate over each of those files and parse their names - rather than simply reading the answer.  Clearly this is much less efficient.

dimbleby:
Anticipating the removal of deprecated "releases": retain information
about the available versions of a package.

Fixes #11991

### DIFF
--- a/warehouse/legacy/api/json.py
+++ b/warehouse/legacy/api/json.py
@@ -173,6 +173,7 @@ def _json_data(request, project, release, *, all_releases):
 
     if all_releases:
         data["releases"] = releases
+        data["versions"] = sorted(releases)
 
     return data
 
--- a/docs/api-reference/json.rst
+++ b/docs/api-reference/json.rst
@@ -167,6 +167,7 @@ Project
                     "yanked_reason": null
                 }
             ],
+            "versions": ["1.0", "1.2.0"],
             "vulnerabilities": []
         }
 
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -326,6 +326,7 @@ class TestJSONProject:
                     }
                 ],
             },
+            "versions": ["0.1", "1.0", "2.0", "3.0"],
             "urls": [
                 {
                     "comment_text": None,


warehouse/legacy/api/json.py
```diff
@@ -173,6 +173,7 @@ def _json_data(request, project, release, *, all_releases):
 
     if all_releases:
         data["releases"] = releases
+        data["versions"] = sorted(releases)
```
dstufft, line 176: I guess a question here is whether we want to align the version-less and the versioned urls and include the `versions` key on all responses like we used to include the `releases` key.

dimbleby: The way I'm thinking about it is that the version-less endpoint helps you to find details for particular versions (and also provides details for whatever the latest version is).  Then the versioned endpoints just give versioned information.

Certainly poetry doesn't have a use case for getting the full version list from a versioned endpoint, so I'm comfortable leaving it off.

Say if you want it in, presumably it's just a case of unindenting this line...

pradyunsg: I reckon it only makes sense for the unversioned URL.

dstufft:
See https://github.com/pypi/warehouse/issues/11991#issuecomment-1558548733

Issue #12095, opened 2022-08-17
Add project last serial number to simple index "all projects" JSON output


pfmoore:
**What's the problem this feature will solve?**
At the moment, the only way to get a list of all projects with their "last updated" serial number is via the XMLRPC API.

**Describe the solution you'd like**
Add an extra field, `_last-serial`, to the PEP 691 JSON output for the individual elements of the `projects` array on the root URL.

**Additional context**
Currently, the `meta` element on both the root page and the individual project pages contain the `_last-serial` key, but there is no way to get a list of all projects and the last time they were updated in a single API call.

Having this information would be useful when mirroring data - by getting the list of all projects and their last update serial, it is possible to check what has been updated since the last mirror run.

**Alternatives Considered**
It is, of course, possible to continue using the XMLRPC API, but that is significantly slower (over 10 seconds vs 0.25 sec). Of course, if the performance difference is the result of extracting the serial numbers, this idea is a non-starter and it's better to keep the simple index fast.

Using `ETag` values to avoid fetching unneeded data is reasonable (and probably should be done in any case) but given that the response body is often not much larger than the header, and even getting a "Not Modified" response needs a network round trip, it's not obvious that this is a significant saving (it's certainly not as good as never calling the server at all...)

pfmoore:
I've created a basic PR, #12134, that adds this, to aid with review.

dstufft:
Sorry I meant to respond then it slipped my mind.

I don't oppose this idea in theory (in fact I think it's a good idea, and is part of what will allow bandersnatch to stop using the XMLRPC API).

My biggest concern is whether we should standardize it through a PEP or not. I know that `_last_serial` is on the project specific pages already without a standard, I did that mostly because the existing HTML pages had that because it forces the ETag to change when the serial changes (which is important for our CDN). The big thing that confuses things is I'm not sure if anyone else but PyPI even has the idea of a serial or if that concept is useful at all except outside of PyPI and whether there is value in standardizing it at all [^1].

[^1]: Although I guess the value would be it gives things like bandersnatch a standard way to work?

pfmoore:
IMO, we should do this now as a Warehouse-specific change, then think about standardising later. I can see the value of standardising *something* to let mirrors say "is this stale?" but does anyone mirror any index other than PyPI? I'd be inclined to wait for a use case before standardising.

[...]

PR #12134, opened 2022-08-26
Add last-serial to the simple index root

pfmoore:


pfmoore:
Fixes #12095

I've manually tested this locally, but not added tests or anything else yet (this is my first attempt at contributing to warehouse). I can look at adding these if the change is acceptable in principle.

dstufft (approved):

pfmoore:
Awesome, thanks! One thing I didn't do is add a test that if we have a project with a last serial number, it shows up in the index. I wasn't 100% sure how to write such a test, and I can't run the tests on Windows for some reason I don't fully understand (I can run make from git-bash but I get an error about bin/tests not existing) so I'd have to experiment via CI.

I won't bother unless you want me to, but if you do, just let me know and I'll have a go.

dstufft:
I think the current state is OK.

Historically nobody working on Warehouse has had a Windows computer so I don't think anyone has ever really tried in earnest to make sure the experience works fine there. I actually switched to Windows for my primary PC a few months ago, but I use WSL2 with Ubuntu and do all my development (unless it's something windows specific) inside that, so while I'm technically still on Windows, not in a particularly meaningful way other than "it mostly works via WSL2".

Our tests run everything inside of docker containers, so I suspect there's some mismatch between paths or something when running via docker via native windows (vs WSL2).

pfmoore:
> I use WSL2 with Ubuntu

Can you use docker from within WSL2? I never tried that, if it works that's awesome.

I'm very definitely not a Unix user, but WSL2 lets me pretend I am :-) One day, I'll do the work to actually set up a comfortable Ubuntu environment for development, but it's really hard to find a good guide for experienced developers who know the tools, but just don't know Unix. (For example, is pyenv still what people use to manage Python on Unix? I heard about asdf but it looks like it goes way beyond what I'd need...)

Sorry, way off topic at this point...

dstufft:
> Can you use docker from within WSL2? I never tried that, if it works that's awesome.

Yes, I have Docker Desktop installed in Windows and I'm able to use it from within WSL2.

> it's really hard to find a good guide for experienced developers who know the tools, but just don't know Unix

Off topic and all that, but maybe I'll try to write up a guide for how I setup my Windows machine for development via WSL2, maybe that would be interesting or useful for you.

PR #12149, opened 2022-08-30
feat: support markdown style math tags

### LINKED ISSUE #12009: Support for mathematical expressions in Markdown
GitHub has [started supporting mathematical expressions in Markdown](https://github.blog/2022-05-19-math-support-in-markdown/).
It would be nice to have the same possibility in PyPi, since oftentimes the `README.md` on GitHub is used as `long_description` in `setup.py`. 

Implement the same syntax used by GitHub, described [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions).

**Additional context**
Example PyPi package for which math expressions are not rendered: https://pypi.org/project/sourcespec/ (section: "Theoretical background").
The original `README.md` on GitHub: https://github.com/SeismicSource/sourcespec/blob/master/README.md

miketheman:
We can configure MathJax to detect the convention of single dollar signs
in addition to its other methods, as it's not on by default.

The inline script needs to be supported by our CSP, and if the stanza
ever changes, we'll need to recalculate the hash and update the CSP
accordingly.

Resolves #12009
Resolves pypa/readme_renderer#214

### DIFF
--- a/tests/unit/test_csp.py
+++ b/tests/unit/test_csp.py
@@ -229,6 +229,7 @@ def test_includeme():
                         "'sha256-U3hKDidudIaxBDEzwGJApJgPEf2mWk6cfMWghrAa6i0='",
                         "https://cdn.jsdelivr.net/npm/mathjax@3.2.2/",
                         "'sha256-1CldwzdEg2k1wTmf7s5RWVd7NMXI/7nxxjJM2C4DqII='",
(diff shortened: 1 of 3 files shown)


miketheman:
Example of using both the block and inline:

Screen Shot 2022-08-29 at 8 15 12 PM

ewdurbin (approved):

PR #12213, opened 2022-09-14
Add 'created' column to all DB models

di:
We've got a fair amount of duplication on the `created` column for models that support it, and a number of tables that are missing a `created` column entirely.

This PR attempts to ensure all current and future models/rows have a `created` field/column by default.

Note that this unsets `nullable=False` as we can't backfill creation dates for models that don't already have the field, but also sets a default, so these should never be null going forward.

### DIFF
--- a/warehouse/db.py
+++ b/warehouse/db.py
@@ -19,7 +19,7 @@ import sqlalchemy
 import venusian
 import zope.sqlalchemy
 
-from sqlalchemy import event, inspect
+from sqlalchemy import Column, DateTime, event, inspect, sql
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlalchemy.ext.declarative import declarative_base  # type: ignore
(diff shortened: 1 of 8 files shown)


warehouse/db.py
```diff
@@ -90,6 +90,11 @@ class Model(ModelBase):
         primary_key=True,
         server_default=sqlalchemy.text("gen_random_uuid()"),
     )
+    created = Column(
+        DateTime(timezone=False),
```
miketheman, line 94: Curious, not material to the refactor, why do we not store the timezone as part of the timestamp?

miketheman:
Should we also consider another column?
```python
updated_at = Column(DateTime, default=datetime.now, onupdate=datetime.now)
```

warehouse/db.py
di: 🤷 This predates me: https://github.com/pypi/warehouse/blame/ead69a5793eb98452f7da7dc90bceda29b44c7f3/warehouse/packaging/models.py#L114

firefly9966 (approved):

PR #12222, opened 2022-09-16
Fix transfer project to organization

divbzero:
- Restrict transfer project choices to active orgs
- Avoid double <hr> when organizations are disabled

### DIFF
--- a/warehouse/manage/views.py
+++ b/warehouse/manage/views.py
@@ -2748,18 +2748,24 @@ class ManageProjectSettingsViews:
             # Disable transfer of project to any organization.
             organization_choices = set()
         else:
-            # Allow transfer of project to organizations owned or managed by user.
+            # Allow transfer of project to active orgs owned or managed by user.
             all_user_organizations = user_organizations(self.request)
-            organizations_owned = set(
+            active_organizations_owned = set(
                 organization.name
                 for organization in all_user_organizations["organizations_owned"]
+                if organization.is_active
             )
-            organizations_managed = set(
+            active_organizations_managed = set(
                 organization.name
                 for organization in all_user_organizations["organizations_managed"]
+                if organization.is_active
             )
-            organization_choices = (organizations_owned | organizations_managed) - (
+            current_organization = (
                 {self.project.organization.name} if self.project.organization else set()
+            )
+            organization_choices = (
+                active_organizations_owned
+                | active_organizations_managed - current_organization
             )
 
         return {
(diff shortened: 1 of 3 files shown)


divbzero:
A couple fixes for the project "Settings" page — not related to any active issue.

warehouse/manage/views.py
```diff
@@ -2748,19 +2748,25 @@ def manage_project_settings(self):
...
+                if organization.is_active
             )
-            organization_choices = (organizations_owned | organizations_managed) - (
+            current_organization = (
                 {self.project.organization.name} if self.project.organization else set()
             )
+            organization_choices = (
+                active_organizations_owned
+                | active_organizations_managed - current_organization
+            )
```
miketheman, line 2769: **thought:** Is the actual behavior of this covered by tests? The one I could find is:
https://github.com/pypi/warehouse/blob/debf2b7db71b36a884c005d92e8dc200ed222d24/tests/unit/manage/test_views.py#L5574-L5599
which has empty inputs, so none of the set union or subtraction is exercised here. Is that tested somewhere else?

I raise it, since it looks like the current would only be excluded from the `managed` portion, not the `owned`, so current would likely still end up in the choices this way, unless I'm misreading something.

```pycon
>>> choices = ( {1,2,3} | {1,2} - {1} )
>>> choices
{1, 2, 3}

# vs

>>> choices = ( ( {1,2,3} | {1,2} ) - {1} )
>>> choices
{2, 3}
```

warehouse/manage/views.py
```diff
@@ -3188,17 +3194,22 @@ def transfer_organization_project(project, request):
...
         for organization in all_user_organizations["organizations_managed"]
+        if organization.is_active
     )
-    organization_choices = (organizations_owned | organizations_managed) - (
+    current_organization = (
         {project.organization.name} if project.organization else set()
     )
+    organization_choices = (
+        active_organizations_owned | active_organizations_managed - current_organization
+    )
```
miketheman, line 3212: Same comment about sets as above.

warehouse/manage/views.py
divbzero: @miketheman Thanks, `-` does take precedence over `|` as indicated in [Python docs](https://docs.python.org/3/reference/expressions.html#operator-precedence). The fact I had to look that up to confirm means I should have used parentheses in the first place. Fixed in commit [d628d10](https://github.com/pypi/warehouse/pull/12222/commits/d628d10d24cd558c242ec895c2112594004224bc) and tested in commit [1533540](https://github.com/pypi/warehouse/pull/12222/commits/1533540e3a77fadf8f672d05558407400b2ad795).

warehouse/manage/views.py
divbzero: Fixed and tested as described above.

miketheman (approved):
Thanks, selection logic looks right!

I found a testing-related issue that seems unrelated to this specific change, so I'll open a different issue.

ewdurbin (approved):

PR #12223, opened 2022-09-16
Sort teams alphabetically

### LINKED ISSUE #12044: Sorting team names
When teams are created in an organization, the teams are listed in the insert order. Ideally, it should be listed in alpha order or the owner should be able to sort the team names.

To recreate the issue-

- Select an organization
- Click on 'Teams'

The teams in the org will be listed in the order they were created. It is not possible to sort the teams. 
test5

divbzero:
- Closes #12044.

### DIFF
--- a/warehouse/templates/manage/organization/teams.html
+++ b/warehouse/templates/manage/organization/teams.html
@@ -24,7 +24,7 @@
   <h2>{% trans team_count=organization.teams|length %}Teams ({{ team_count }}){% endtrans %}</h2>
 
   <div class="team-list">
-    {% for team in organization.teams|sort(attribute="name") %}
+    {% for team in organization.teams %}
     <div class="team-snippet">
       <div class="split-layout split-layout--no-wrap split-layout--wrap-on-tablet">
         <div>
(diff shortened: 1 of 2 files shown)


warehouse/templates/manage/organization/teams.html
```diff
@@ -24,7 +24,7 @@
   <h2>{% trans team_count=organization.teams|length %}Teams ({{ team_count }}){% endtrans %}</h2>
 
   <div class="team-list">
-    {% for team in organization.teams %}
+    {% for team in organization.teams|sort(attribute="name") %}
```
miketheman, line 27: **thought:** ‏ While this works, is there a way to send more of the responsibility upstream to the model, and keep the template "simpler"? Likely not a real tradeoff until any organization has a large volume of teams that would cost more to sort in Python than sorting in the Postgres query that returns them already-in-order. Not blocking by any means.

miketheman (approved):

warehouse/templates/manage/organization/teams.html
divbzero: @miketheman Shifted sorting responsibility upstream to `Organization` model in commit [7365596](https://github.com/pypi/warehouse/pull/12223/commits/7365596d52a72ab487770a0a07fab73add6297af).

miketheman (approved):
Nice!

ewdurbin (approved):

Issue #12039, opened 2022-08-08
Email generated after changing user role shows enum


s-mm:
Once a user has joined an org, the role of the user can be changed. When the role is changed, the email generated lists the new role as the enum value.

to recreate the steps-

- Click on 'Your organization'
- Select an org
- Scroll down to view the members in the org
- Change the role of any member and click on 'Save'

The following emails are generated

![test2](https://user-images.githubusercontent.com/37237726/183418403-8fdf722b-2629-48ff-bdff-f6fbf994a510.png)

![test3](https://user-images.githubusercontent.com/37237726/183418413-b7b15d59-d0f5-4926-a830-d467bf8f748a.png)

ewdurbin:
The template variable values in https://github.com/pypi/warehouse/blob/637435d22451494c83a5c404c409211b25df7c13/warehouse/manage/views.py#L2022-L2037 need to be updated to `role.role_name.value`.

Resolving this should also include checking for this in other places where `role.role_name` or similar is called.

miketheman:
Related https://github.com/pypi/warehouse/issues/6636

divbzero:
Fixed by #12231.

PR #12231, opened 2022-09-17
Fix email role enum issue (#12039)

sterbo:
- Fix bug where email generated after changing user role shows enum
- Fix similar issue with Team Project Role email

### DIFF
--- a/warehouse/manage/views.py
+++ b/warehouse/manage/views.py
@@ -2172,7 +2172,7 @@ def change_organization_role(
                 user=role.user,
                 submitter=request.user,
                 organization_name=organization.name,
-                role=role.role_name,
+                role=role.role_name.value,
             )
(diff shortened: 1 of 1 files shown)


miketheman (changes requested):
Looks correct to me! There's a couple other cases that don't appear to be handled, unless they are?
https://github.com/pypi/warehouse/blob/debf2b7db71b36a884c005d92e8dc200ed222d24/warehouse/manage/views.py#L4330-L4345

miketheman:
You’re right! I missed the variable set above this section 

    role.role_name = form.role_name.data
    
That’s what we get for reusing common names in the same module. 😬

miketheman (approved):

ewdurbin (approved):

Issue #12055, opened 2022-08-09
Removing a project from an organization leads to 403 error


s-mm:
If the user is not an Owner of the project, removing a project from an organization leads to an error

To recreate the issue-

- Click on Your Projects
- Select a Project and click on Manage
- Click on Settings
- Click on Remove Project
- Enter the Project name
- Click on Remove Project

The confirmation goes to /manage/project/project_name/settings/ which the user does not have access to as the user is not an Owner of the project anymore. 

![test11](https://user-images.githubusercontent.com/37237726/183537333-67d475d1-5ada-428e-ba29-bfeb5ea5189b.png)

The URL will have to be changed to the home page to see the confirmation message that the project has been removed from the organization.
![test12](https://user-images.githubusercontent.com/37237726/183537338-2a019437-e898-4dfe-8b6e-8e013521d3bc.png)

s-mm:
@sterbo Your suggestion is fine.

PR #12237, opened 2022-09-19
Fix redirect removing project from an org (#12055)

sterbo:
- Redirect to org project management page instead of project settings


ewdurbin:
Looks good once translations are updated.

PR #12232, opened 2022-09-18
chore(deps): upgrade gifsicle transitive dep

miketheman:
When previously attempting a `buster` to `bullseye` upgrade, [we observed a failure to install `gifsicle`](https://github.com/pypi/warehouse/pull/11472#issuecomment-1144089564), a build-time dependency of `gulp-image`.

Instead of upgrading the entire `gulp-image` stack right now, upgrade the transitive dependencies by running this command *inside* the `static` container:

    npm remove gulp-image && npm add gulp-image@6.1.0

The resulting changes to `package-lock.json` confirm that `gifsicle` has been upgraded from 5.0.0 to 5.3.0, and now installs cleanly without needing compilation.

Added benefit is some of the `integrity` hashes have been migrated from SHA1 to SHA512.

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

PR #11472, opened 2022-05-28
chore: upgrade underlying debian version

miketheman:
Debian `stable` has moved from buster (10) to bullseye (11).

Swapping out the underlying Debian version provides us with acces to
newer package dependencies, specifically: gcc 8.3 => 10.2

Refs: https://www.debian.org/releases/bullseye/
Refs: https://www.debian.org/releases/bullseye/amd64/release-notes/ch-whats-new.en.html#newdistro

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

PR #12262, opened 2022-09-24
chore: update static base image to bullseye

miketheman:
As previously attempted in #11472 and reverted in #11493

After isolating the condition that was observed and removing the calls to compile `gifsicle` in #12232, upgrade the `static` containers only with this changest.


miketheman:
@ewdurbin This is similar to last time. Is there a way to test this commit in the Cabotage build/deploy process prior to merging to see if it raises any issues?

ewdurbin (approved):

ewdurbin:
All seems well!

miketheman:
Thanks! We did it!

PR #12263, opened 2022-09-24
Introduce a gate/check GHA job

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

I was curious about the spread of use. And I've just discovered that it is now in use in aiohttp (and other aio-libs projects), CherryPy, some of the Ansible repositories, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some PyCQA, PyCA and pytest projects, a few AWS Labs projects. Admittedly, I maintain a few of these but it seems to address some of the pain folks have: https://github.com/jaraco/skeleton/pull/55#issuecomment-1106638475.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,3 +99,17 @@ jobs:
         run: npm ci
       - name: Run ${{ matrix.name }}
         run: ${{ matrix.command }}
+
+  check:  # This job does nothing and is only used for the branch protection
+    if: always()
+
+    needs:
+      - test
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}


.github/workflows/ci.yml
```diff
@@ -99,3 +99,17 @@ jobs:
...
+    if: always()
+
+    needs:
+      - test
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
```
webknjaz, line 113: Optionally, pin for stability/security:
```suggestion
        uses: re-actors/alls-green@198badcb65a1a44528f27d5da555c4be9f12eac6
```

.github/workflows/ci.yml
```diff
@@ -99,3 +99,17 @@ jobs:
         run: npm ci
       - name: Run ${{ matrix.name }}
         run: ${{ matrix.command }}
+
+  check:  # This job does nothing and is only used for the branch protection
```
webknjaz, line 103: @di this would need to go to branch protection.

miketheman:
I'm going to close this PR, as we've since expanded to using multiple workflows, and this approach for gated checks appears useful for single-workflow files.
It's a good idea, just not exactly for this repo workflow right now.

webknjaz:
@miketheman 😂 I'm having flashbacks!
Exactly the same happened to my CPython PR — Łukasz rejected my PR but then, he found me at the PyCon Sprints claiming he saw the light. Apparently, they were looking into enabling auto-merge and the docs workflow being separate, messed up things. So Łukasz asked me to resurrect the PR and I'm now working on that. We were lucky to have a chance to talk in person and I explained the implications of different approaches.
Finally, I proposed to make use of in-repo reusable workflows as a way of modularizing the conditional bits that Łukasz liked — this is something I've been planning to do for my other projects for a while now.
If this sounds compelling, I can look into doing the same for the Warehouse repo... It'd be a single PR workflow runtime-wise but then definitions could remain spread across multiple files.

webknjaz:
@miketheman just wanted to make sure you saw my previous suggestion/message?

PR #12273, opened 2022-09-26
Organization rename help text

### LINKED ISSUE #12042: Changing the organization account name should be intuitive
When changing the org account name, it is not clear that the user has to scroll down to see the option to change the org account name. 

To recreate this issue-

- Click on 'Your organizations'
- Select an organization
- Click on 'Settings'

test4
The divider in between the sections implies that there is nothing below. Adding a message next to org account name that says ' To change the org account name, scroll below' would also be useful.

divbzero:
- Closes #12042.


warehouse/templates/manage/organization/settings.html
```diff
@@ -28,6 +28,12 @@ <h2>{% trans %}Organization settings{% endtrans %}</h2>
     <div class="form-group">
       <span class="form-group__label">{% trans %}Organization account name{% endtrans %}</span>
       <p class="form-group__text">{{ organization.name }}</p>
+      <p class="form-group__help-text">
+        {% trans %}This account name is used in URLs on PyPI.{% endtrans %}<br>
+        {% trans %}Change the organization account name <a href="#change-organization-account-name">below</a>.{% endtrans %}
```
miketheman, line 33: Suggestion: extract url to a variable, per translation docs guidelines 
https://warehouse.pypa.io/translations.html#passing-non-translatable-values-to-translated-strings

miketheman (commented):
Looks like this would provide enough context to highlight to the user where this option lives.

warehouse/templates/manage/organization/settings.html
ewdurbin: +1

ewdurbin (changes requested):

miketheman (approved):

ewdurbin (approved):

PR #12285, opened 2022-09-28
Add functionality to hide user avatar

di:
Since Gravatar seems to be completely unresponsive to moderation requests, and since this is 'user-published content' on PyPI that we don't have any control over, this PR gives admins the ability to hide a user's avatar on their profile.

### DIFF
--- a/warehouse/migrations/versions/d0f67adbcb80_add_avatar_hidden_column_to_user_model.py
+++ b/warehouse/migrations/versions/d0f67adbcb80_add_avatar_hidden_column_to_user_model.py
@@ -1,0 +1,38 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


warehouse/migrations/versions/d0f67adbcb80_add_avatar_hidden_column_to_user_model.py
```diff
@@ -0,0 +1,38 @@
...
+
+revision = "d0f67adbcb80"
+down_revision = "fe2e3d22b3fa"
+
+
+def upgrade():
+    op.add_column(
+        "users",
+        sa.Column(
+            "hide_avatar", sa.Boolean(), server_default=sa.text("false"), nullable=False
```
ewdurbin, line 32: If intuition serves me, the server_default will allow this migration to be fast, regardless of the nullable=False, right?

ewdurbin (approved):

warehouse/migrations/versions/d0f67adbcb80_add_avatar_hidden_column_to_user_model.py
di: I think so, yes.

PR #12288, opened 2022-09-29
chore: Upgrade NodeJS to 16.x

miketheman:
Node 14 ends support in April 2023.
Node 16 will remain supported until September 2023.

Upgrading to Node 16 allows removal of the `node-webcrypto-ossl` dependency used in tests.

`package-lock.json` updated for the new lockfile format with npm 8 and removes some no longer needed dependencies.


miketheman:
I decided to not include the removal of the referenced test library in this changeset, rather let's let this one go out with no other changes and confirm things are happy before further changes.

miketheman:
The lockfile grew significantly in size, as it's now version 2 of the lockfile with speed improvements.
https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json

di (approved):
Here's the full diff: https://gist.github.com/di/8a396464d5b600cda5df2124edb1ba2b

This is somewhat hard to review due to the change in `package-lock.json` versions, but looks right to me.

We should probably wait until Monday to merge this so we can babysit it in case it fails for some reason.

di:
This rolled out successfully, thanks @miketheman!

PR #12333, opened 2022-10-08
feat: add notice for missing source distributions

### LINKED ISSUE #11060: Add warnings about package missing sdists.
As raised in #7980, sometimes packages lack sdists, instead uploading only wheels. This causes issues for some users of PyPI. Rather than enforce uploading sdists (which is a major breaking change), warning users of PyPI that a package lacks sdists increases the visibility of missing sdists (whether accidental or on purpose). This allows users to be able to make more informed choices about dependencies, and to more easily spot any automation-related issues.

There should be clear warnings about missing sdists. A first pass could be a warning on the release history page on any releases missing sdists, and a warning on release pages similar to the "newer version available" that already exists. This could then be expanded by having said warning link to documentation about uploading sdists and why they are needed (in case the reason is a misunderstanding by uploaders), and by adding metadata which allows uploader to specify why sdists are not uploaded.

miketheman:
In the event that the release has binary distributions but no source distributions, add a notice for the viewer that the files are missing.

Example rendering:

Screen Shot 2022-10-07 at 8 56 33 PM

Resolves #11060

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -529,6 +529,7 @@ msgstr ""
 #: warehouse/templates/manage/project/releases.html:140
 #: warehouse/templates/manage/project/releases.html:179
 #: warehouse/templates/packaging/detail.html:350
+#: warehouse/templates/packaging/detail.html:369
 #: warehouse/templates/pages/classifiers.html:25
 #: warehouse/templates/pages/help.html:20
 #: warehouse/templates/pages/help.html:212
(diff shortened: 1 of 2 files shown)


miketheman:
Makes sense, thanks for the clarification. I'll poke at this some more in a week or two and come up with another changeset and screenshot to share.

miketheman:
Updated to reflect the feedback.

Screen Shot 2022-10-21 at 6 56 16 PM

Note: there might be a pre-existing bug with the pluralization of the header, haven't dug into that yet, beyond scope for this change.

ewdurbin (approved):

di (approved):

PR #12351, opened 2022-10-12
Catalog event tags

### LINKED ISSUE #7119: Catalog and enforce ProjectEvent and UserEvent tags

divbzero:
- Catalog event tags as `EventTag.*` enums.
- Standardize tag names for role events.
- Record missing role events.
- Update journal entries (cherry-picked from #11779).
- Closes #7119.


divbzero:
This pull request should be reviewed and merged first.

warehouse/accounts/views.py
```diff
@@ -1069,13 +1089,13 @@ def _error(message):
     request.db.add(
         JournalEntry(
             name=project.name,
-            action=f"accepted {desired_role} {user.username}",
+            action=f"add {desired_role} {user.username}",
```
ewdurbin, line 1092:
```suggestion
            action=f"accepted {desired_role} {user.username}",
```

We should leave the legacy Journals as they are.

ewdurbin (approved):

PR #12395, opened 2022-10-19
[WIP] feature admin 2fa reset

sijis:
I recently had a 2fa issues with warehouse and i noticed a [pinned issue](https://github.com/pypa/pypi-support/issues/796) around the problem. 

I thought i would give it an initial attempt on realizing the feature, noticing there are 100's of request in the repository.

There are a few things, I have not tackled yet
* storing the generated token created (idea would be admin could press a button and "check" that branch was created)
* validating email message
* Completely validating the logic
* tests

I'm also not sure if that location of the reset is in the right correct location, or should it be in some other place.

Screenshot from 2022-10-19 10-56-59

### DIFF
--- a/warehouse/admin/templates/admin/users/detail.html
+++ b/warehouse/admin/templates/admin/users/detail.html
@@ -147,6 +147,40 @@
                 </div>
               </div>
             </li>
+            <li>
+              <button type="button" class="btn btn-danger" data-toggle="modal" data-target="#reset2FA" {{ "disabled" if not request.has_permission('admin') }}>
+                <i class="icon fa fa-bomb"></i> Initiate 2FA Reset
+              </button>
+              <div class="modal fade" id="reset2FA" tabindex="-1" role="dialog">
+                <form method="POST" action="{{ request.route_path('admin.user.reset_2fa', user_id=user.id) }}">
+                  <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+                  <div class="modal-dialog" role="document">
+                    <div class="modal-content">
+                      <div class="modal-header">
+                        <h4 class="modal-title" id="exampleModalLabel">Initiate 2FA reset for {{ user.username }}?</h4>
+                        <button type="button" class="close" data-dismiss="modal">
+                          <span>&times;</span>
+                        </button>
+                      </div>
+                      <div class="modal-body">
+                        <p>
+                          This will initiate resetting the user's 2FA.
+                        </p>
+                        <hr>
+                        <p>
+                          Type the Gitub Issue URL to begin:
+                        </p>
+                        <input type="text" name="issue-url" width="100%" placeholder="Github Issue URL">
+                      </div>
+                      <div class="modal-footer">
+                          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+                          <button type="submit" class="btn btn-danger" title="Reset 2fa" if not request.has_permission('admin') }}" {{ "disabled" if not request.has_permission('admin') }}>Reset 2FA</button>
+                        </div>
+                    </div>
+                  </form>
+                </div>
+              </div>
+            </li>
           </ul>
         </div>
       </div>
--- a/warehouse/admin/templates/admin/users/detail.html
+++ b/warehouse/admin/templates/admin/users/detail.html
@@ -281,6 +281,35 @@
             </table>
           </div>
         </div>
+      </div>
+      <div class="box">
+        <div class="box-header with-border">
+          <h3 class="box-title">2FA - Reset</h3>
+        </div>
+        <div>
+          <div class="box-body">
+            <form class="form-horizontal" method="POST" action="{{ request.route_path('admin.user.reset_2fa', user_id=user.id) }}">
+                {{ render_field("Github Issue", form.name, "user-name", class="form-control", placeholder="Github Issue URL") }}
+                <table class="table table-hover">
+                <tr>
+                    <th>Project</th>
+                    <th>{% if user.has_recovery_codes %}<i class="fa fa-check text-green"></i>{% else %}<i class="fa fa-times text-red"></i>{% endif %}</th>
+                </tr>
+                <tr>
+                    <th>Token</th>
+                    <th>{% if user.has_recovery_codes %}<i class="fa fa-check text-green"></i>{% else %}<i class="fa fa-times text-red"></i>{% endif %}</th>
+                </tr>
+                <tr>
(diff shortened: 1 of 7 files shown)


warehouse/admin/views/users.py
```diff
@@ -182,6 +185,51 @@ def user_add_email(request):
...
+                    try:
+                        status_code = requests.get(v).status_code
+                    except:
+                        status_code = 404
+
+                    if status_code == 200:
+                        valid_sites.add(project.name)
+
+            if valid_sites:
+                choice = random.choice(list(valid_sites))
```
di, line 216: We probably don't want this to be random, but rather leave it up to the admin to choose. That probably means we need a two-step process, one that generates a list of URLs and another that takes a selection and sends the email.

warehouse/admin/templates/admin/users/detail.html
```diff
@@ -281,6 +281,35 @@ <h3 class="box-title">2FA</h3>
...
+                </tr>
+                <tr>
+                    <th>Token</th>
+                    <th>{% if user.has_recovery_codes %}<i class="fa fa-check text-green"></i>{% else %}<i class="fa fa-times text-red"></i>{% endif %}</th>
+                </tr>
+                <tr>
+                    <th>Reset 2FA</th>
+                    <th>
+                        <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+                        <button type="submit" class="btn btn-danger" title="Reset 2fa" if not request.has_permission('admin') }}" {{ "disabled" if not request.has_permission('admin') }}>Reset 2FA</button>
```
di, line 306: Probably better to change this to something like 'Initiate 2FA reset' or 'Send 2FA reset email', since in the future we _will_ want a button that says 'Reset 2FA' that wipes all 2FA methods from the accounts.

warehouse/templates/email/admin-reset-2fa/body.html
```diff
@@ -0,0 +1,27 @@
...
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+{% block content %}
+
+<p>An account recovery request was filed at <a href="{{ gh_link }}">{{ gh_link }}</a> for the PyPI username "{{username}}".</p>
+
+<p>In order to verify the request, please push a branch with the name "{{token}}" to the public source code repository for the {{project_name}} project.</p>
```
di, line 20: This doesn't seem to handle the case where `project_name` is `None`.

warehouse/templates/email/admin-reset-2fa/body.txt
```diff
@@ -0,0 +1,27 @@
...
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.txt" %}
+
+
+{% block content %}
+An account recovery request was filed at {{gh_link}} for the PyPI username "{{username}}".
+
+In order to verify the request, please push a branch with the name "{{token}}" to the public source code repository for the {{project_name}} project.
```
di, line 20: Ditto here about `project_name` being `None`.

di (commented):
This is great @sijis, thank you for tackling this!

Some comments about the overall functionality but generally this looks good to me, cc @ewdurbin to take a look as well.

warehouse/admin/templates/admin/users/detail.html
sijis: Yeah, I was re-reviewing what I had put together and I do agree, renaming it makes sense.

warehouse/templates/email/admin-reset-2fa/body.html
sijis: Yup. I haven't really gone through the email templates fully.

warehouse/admin/views/users.py
sijis: I suppose that would work.

I don't know if the long term goal would be able to have the user be able to do reset it themselves. In that case, the system would be auto-picking the projects, right? I figured, if we are headed that way, maybe we should include that now.

sijis: I also wasn't sure if we should store the "token" (and probably selected project) used to validate the individual does have access to the code repository.

The reason is, I'm not sure how the warehouse admin would know what was sent to user, to have the admin validate everything matches.

di: The long term goal is to keep a PyPI admin in the loop, just make it easier on that human -- this won't ever be self-serve.

Yes, the token should either be stored somewhere, made more deterministic, or just turned into something else -- I don't quite see the benefit to this being a random string. The goal is to prove that the person we're interacting with has control over the account/repository in question, so asking them to do something like push a commit to a predictable branch like `pypi-account-recovery-YYYY-MM-DD` with a commit message like:

> I (pypi.org/user/<username>) have requested account recovery on PyPI, and this commit signifies that I am in control of this corresponding account.

should be just as meaningful as a random string -- it just needs to be some activity that someone who is being compromised on PyPI would be unlikely to do on a corresponding account.

di: Also summoning @ewdurbin for their thoughts here.

di: I'm wrong here - we do want some sort of random token, that is sent via the email, to demonstrate ownership of the email account as well.

We should store this as well as the choice that's being made here, so when we verify we can refer back to this. This should probably be stored in a new model/table that represents a reset request.

ewdurbin (commented):
This is headed in a good direction. I'll respond to some of the open questions on the discussion regarding picking/storing project for verification.

Otherwise what else is needed to be able to move forward?

warehouse/admin/views/users.py
ewdurbin: Indeed, a random secret allows us to validate that the actor who publishes a verification branch is the same as the one who controls the primary email address.

We'd need to store the verification code as well as the project(s) selected for validation. To proceed a single project is fine, but in some cases we may want someone to demonstrate ownership across certain boundaries (GitHub orgs or hosting services).

ewdurbin: Agreed, an administrator should review and select what project will be chosen for validation. Some key criteria we might look for which would be good to surface in the select view:

- Number of other maintainers/owners
- Is project marked as [critical](https://github.com/pypi/warehouse/blob/main/warehouse/packaging/models.py#L157)?
- Number of releases
- Most recent release
- # of valid/invalid URLs

di:
Done in https://github.com/pypi/warehouse/pull/13848.

PR #12398, opened 2022-10-20
Validation for recovery code field

### LINKED ISSUE #12397: Improvements to 'Login using recovery codes' input field
We've gotten a report that users are confused by the 'Login using recovery codes' input field, sometimes attempting to paste all their recovery codes into the field at once.

We should add some simple client-side and server-side verification of this field, to ensure that the contents are a) the length we expect for a recovery code b) a valid set of characters.

Additionally, we should ensure that the field does not autocomplete or retain previously entered data, which it does now.

ewdurbin:
We've had some user reports of confusion. Fixes #12397.


di (approved):
😗 👌

PR #12408, opened 2022-10-23
feat: wrap plaintext descriptions as preformatted

### LINKED ISSUE #12253: Monospaced font for text/plain long_description
Don't you think that would be nice to wrap project descriptions in text/plain with pre tag?
Close if duplicate. I'm really sorry, looking through over 400 issues of production system is beyond of my capabilities.

miketheman:
Instead of leveraging the `readme_renderer` for projects that explicitly upload as "text/plain" content types, use the raw content and wrap with the HTML `<pre>` tag to display correctly in browser.

Resolves #12253


tests/unit/packaging/test_views.py
```diff
@@ -186,7 +233,7 @@ def test_detail_rendered(self, db_request):
                 description=DescriptionFactory.create(
                     raw="unrendered description",
                     html="rendered description",
-                    content_type="text/plain",
+                    content_type="text/html",
```
miketheman, line 236: these changes are to make the behavior track the common use case - `text/html`.
I added a similar test for plain above these two - there wasn't enough overlap for the existing `_rendered/_renders` tests to refactor to a parameterize, but can do that if desired.

warehouse/packaging/views.py
```diff
@@ -84,11 +84,13 @@ def release_detail(release, request):
     if project.name != request.matchdict.get("name", project.name):
         return HTTPMovedPermanently(request.current_route_path(name=project.name))
 
+    # Prefer plain text if explicitly set, otherwise prefer html.
+    if release.description.content_type == "text/plain":
+        # Wrap as preformatted text to preserve whitespace.
+        description = f"<pre>{release.description.raw}</pre>"
```
di, line 90: I think we probably need to do this at render-time instead and store it in `release.description.html` (either when we're rendering inline, like on the lines below, or when we're doing it in a task)

warehouse/packaging/views.py
```diff
@@ -84,11 +84,13 @@ def release_detail(release, request):
...
     # Grab the rendered description if it exists, and if it doesn't, then we will render
     # it inline.
-    # TODO: Remove the fallback to rendering inline and only support displaying the
-    #       already rendered content.
```
di, line 90: The behavior mentioned in this comment is still happening, it probably shouldn't be removed.

warehouse/packaging/views.py
```diff
@@ -84,11 +84,13 @@ def release_detail(release, request):
...
-    if release.description.html:
+    elif release.description.html:
         description = release.description.html
     else:
         description = readme.render(
```
di, line 96: I think here (and anywhere else we call `readme.render` is where we need to wrap with `pre`.

warehouse/packaging/views.py
miketheman: Centralized the behavior in `readme.render()`.

warehouse/packaging/views.py
miketheman: Restored!

warehouse/packaging/views.py
miketheman: This was the key - instead of hunting down all those places, updated `render()` to do what we want.

di (approved):

PR #12427, opened 2022-10-25
Doom transaction if `request.user` is None

### LINKED ISSUE #12422: AdminFlag.READ_ONLY no longer works
**Describe the bug**

Enabling the `read-only` admin flag leads to multiple errors.

**Expected behavior**

Enabling the `read-only` admin flag should at least allow the basic web ui, simple api, and JSON api to function without error.

**To Reproduce**

- Login as an admin user (`ewdurbin`:`password` in dev)
- Navigate to http://localhost/admin/flags
- Enable `read-only` flag and save
- Attempt any request with http://localhost/

**Additional context**

I'd like to use the `read-only` flag to minimize the overall impact of upgrading our Postgres database from 12 -> 14. Read only mode would allow us to keep all services online and readable while dumping from 12 and loading into 14 which would minimize the impact/downtime.

di:
Fixes #12422.

### DIFF
--- a/tests/unit/test_db.py
+++ b/tests/unit/test_db.py
@@ -164,6 +164,7 @@ def test_create_session(monkeypatch, pyramid_services):
         (None, pretend.stub(is_superuser=False), []),
         (pretend.stub(enabled=False), pretend.stub(is_superuser=True), []),
         (pretend.stub(enabled=False), pretend.stub(is_superuser=False), []),
+        (pretend.stub(enabled=False), None, []),
         (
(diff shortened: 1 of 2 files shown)


ewdurbin:
This still raises a `RecursionError: maximum recursion depth exceeded while calling a Python object` when logged in and `RuntimeError: Cannot use request.session in a view without uses_session=True.` when unauth'd.

di:
@ewdurbin Try this again, if it works I'll add the necessary tests.

PR #12478, opened 2022-11-02
Quieter notdatadog

di:
Cuts down on the amount of output in local dev.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -159,7 +159,8 @@ services:
     build:
       context: .
     command: python /opt/warehouse/dev/notdatadog.py 0.0.0.0:8125
-    env_file: dev/environment
+    environment:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -159,6 +159,7 @@ services:
     build:
       context: .
     command: python /opt/warehouse/dev/notdatadog.py 0.0.0.0:8125
+    env_file: dev/environment
```
ewdurbin, line 162: I don't think we need to share the entire environment file with this container.

I'd suggest adding

```
    environment:
      METRICS_OUTPUT: "false"
```

di, line 162:
```suggestion
    environment:
      METRICS_OUTPUT: "false"
```

ewdurbin (approved):

Issue #6504, opened 2019-08-21
Restore dev/clean.sql and update for new security features


ewdurbin:
It appears the `dev/clean.sql` file referenced in https://warehouse.pypa.io/development/development-database/ was inadvertently removed in #5882.

We should restore this file, and also update it to remove:
  - User TOTP secrets: NULL out `users.totp_secret` column
  - User last TOTP values: NULL out `users.last_totp_value` column
  - User Security Keys: truncate `user_security_keys` table
  - Macaroons: truncate `macaroons` table
  - Audit tables: anonymize `project_events` and `user_events` tables

di:
While we're at it, we should also update the dev DB dump from TestPyPI, there's a lot of migrations that need to be applied since we last did this.

dstufft:
Alternatively, we can just run ``make initdb`` and stamp a new dev DB from that and not pull "production" data from TestPyPI.

ewdurbin:
that still has a huge downside of plopping another 100+MB blob in our git history :/

dstufft:
We could just use git LFS and/or host it on S3 and change ``make initdb`` fetch from that.

PR #12492, opened 2022-11-07
move development database out of the source tree

ewdurbin:
this also includes a dump/migrate/restore of the old DB being pushed to S3 which contains postgres14 pg_dump with all of the current migrations applied.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ docker-compose.override.yaml
 node_modules/
 
 dev/example.sql
+dev/example.sql.xz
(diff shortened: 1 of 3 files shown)


ewdurbin:
I'd like us to be able to have this replace the current `dev/example.sql.xz` if the shasum value in the Makefile changes... but I'm not smart enough at make.

ewdurbin:
toward(ish) #6504.

ewdurbin:
Going to go a different direction here and include it via a development dependency.

PR #12527, opened 2022-11-10
[WIP] Project reservations

### LINKED ISSUE #11296: "De novo" project creation via API tokens/OIDC
There's currently a little bit of friction around bootstrapping API tokens on a new PyPI project:

1. Prepare a package for a new project
2. Upload the first version of the project to PyPI using a user-scoped API token (not ideal for long-term actions) or user credentials (shouldn't be used in CI)
3. Log into PyPI and create a project-scoped token for the new project
4. Use the new project-scoped token for all future releases

A similar problem will happen with OIDC: we currently require a project to exist before an OIDC publisher can be registered against it, meaning that a user will have to upload at least one release with a user-scoped token or credentials before they can switch to a credential-less workflow.

Some potential solutions:

1. Allow users to pre-create projects on PyPI before uploading any releases.
    - Pros: Solves the "de novo" problem above.
    - Cons: Malicious users can abuse this as a low-friction way to squat names.
2. Allow users to pre-register OIDC publishers; on use, the project would be created for the first time.
    - Pros: Solves the "de novo" problem *and* minimizes squatting potential.
    - Cons: Complicates the OIDC data mode: pre-registering a publisher means either pre-creating the project (same problem as (1)) or creating a "symbolic" link to a nonexistent project (another column). We'd have to be careful about invalidating a publisher if someone else races/beats it to creating a project, so that someone can't quietly takeover a project's releases via OIDC.
3. Something else?

woodruffw:
This is still a WIP; just putting something up for visibility and early feedback.

Outstanding items:

- [x] `POST`ing the reservation form should actually reserve the project name
- [x] The user's `manage/projects` list should include a "reserved" chip on any reserved project(s)
- [x] Scheduled job for periodically removing reservations that have gone stale
- [ ] Unit tests
- [ ] The upload endpoint needs to remove `Project.reserved_until`, if present
- [ ] The project's own settings page should display the fact that it's reserved

Closes #11296.

cc @di


warehouse/accounts/models.py
```diff
@@ -184,6 +184,26 @@ def can_reset_password(self):
...
+            #  * Must have 2FA enabled
+            #  * Must be at least 90 days old
+            #  * Must not already have a project name reserved
+            has_already_reserved_project = any(p.is_reserved for p in self.projects)
+            ninety_days_ago = datetime.datetime.now() - datetime.timedelta(days=90)
+            return (
+                self.has_primary_verified_email
+                and self.has_two_factor
+                and self.date_joined <= ninety_days_ago
+                and not has_already_reserved_project
```
woodruffw, line 204: Flagging for review: do I need to check `User.is_frozen` here as well? I don't think I do since I believe that frozen users can't log in to begin with, but I could be wrong about that.

woodruffw:
> * Scheduled job for periodically removing reservations that have gone stale

Just making a note for myself: this'll go under `warehouse.packaging.tasks`, with the timer/cron configured in `warehouse.packaging` (i.e. `__init__.py`).

woodruffw:
@di given the change in design in https://github.com/pypi/warehouse/issues/11296, I think I'm going to close this out and start anew. Any objections to that? 🙂

di:
@woodruffw Whatever is easiest for you!

woodruffw:
(I'll leave this branch undeleted, since I think the groundwork here might be interesting/useful if the project reservation idea is ever revived.)

Issue #12484, opened 2022-11-05
reStructuredText markup broken for Robot Framework code blocks


mentalisttraceur:
**Describe the bug**
Project description written in reStructuredText; `.. code:: robotframework` blocks rendered with strikethrough on all the keyword arguments.

![image](https://user-images.githubusercontent.com/6531708/200101948-c0ec74d1-f25a-47a5-89c7-f4c1f097e0b7.png)

Example image from [`robotframework-dependencylibrary`](https://pypi.org/project/robotframework-dependencylibrary/).

**Expected behavior**
No strikethrough in code blocks.

**To Reproduce**
Make a package with a reStructuredText description containing `.. code:: robotframework` blocks with at least one test calling at least one keyword with at least one argument, I presume.

mentalisttraceur:
Note that it previously rendered correctly for years.

I've had that package with a lot of the same project description reStructuredText since December 2017, last looked at it around October 2021, and I think I would've noticed strikethrough in my code blocks.

mentalisttraceur:
The change to the generated HTML happened sometime between [January 28 2022](http://web.archive.org/web/20220128014519/https://pypi.org/project/robotframework-dependencylibrary/) and [October 31 2022](http://web.archive.org/web/20221031123817/https://pypi.org/project/robotframework-dependencylibrary/).

mentalisttraceur:
Looks like the HTML change was specifically `<span class='s'>...</span>` being replaced with `<s>...</s>`.

miketheman:
Thanks for the detailed report! I believe this is due to how the conversion from HTML4 to HTML5 elements worked along with the Pygments library setting for using short codes where 's' is used for a string, and it can also mean strike through.

Fixing this means a change to readme_renderer and the CSS here in warehouse to adapt.

PR #12530, opened 2022-11-11
fix: remove strikethrough from literal strings

miketheman:
When Pygments renders HTML, t uses the `<s>` tag for string literals, which most browsers will interpret as a Strikethrough, which is confusing.

Introduced during a fix for pypa/readme_renderer#262 and surfaced in project pages since.

Fixes: #12484

### DIFF
--- a/warehouse/static/sass/blocks/_project-description.scss
+++ b/warehouse/static/sass/blocks/_project-description.scss
@@ -258,6 +258,11 @@
     margin-right: auto;
   }
 
+  s { /* pygments Literal.String */
+    /* Override browser default behavior */
+    text-decoration: none;
+  }
+
 }
 
 


di (approved):

mentalisttraceur:
I appreciate the quick workaround, and it's definitely an improvement, but I worry about this causing problems for people who are deliberately using `<s>` in their docs.

Because `<s>` is [the official way](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-s-element) to say "this text is wrong/irrelevant" in HTML. The strikethrough is just the default way of visually displaying that meaning, but the meaning itself is standard.

So:

1. if someone has use in their documentation for saying "this text is no longer accurate or no longer relevant", `<s>` is the right way to say that; and
2. if someone is accessing the HTML in any way other than the normie visual way through a browser, the `<s>` tag would still be saying the wrong meaning to them.

Personally I'd consider this an upstream bug in the Pygments-to-HTML conversion - using `<s>` for "string literal" is wrong the same way that saying "this is invalid" when we really mean "this is a dog" is wrong.

But until that's fixed, thanks for working around it for the majority of people at the visual level.

miketheman:
@mentalisttraceur I appreciate the detailed explanation of the concern! Definitely something that I'd hope `pygments` fixes at their level.

I'll continue to explore whether we could leverage the non-short versions for pypi.org , and enable that output via configuration flag, since it would mean a breaking change if that's the default, at least for a while.

PR #12572, opened 2022-11-18
warehouse: add initial pending OIDC provider models

woodruffw:
See #11296.

This is going to be a relatively complex set of models (thanks to the mixins for shared functionality between "pending" and concrete OIDC providers), so I wanted to get them up and merged before beginning on actual routes/view logic.

~~Needs unit tests.~~

### DIFF
--- a/warehouse/oidc/models.py
+++ b/warehouse/oidc/models.py
@@ -182,6 +182,7 @@ class PendingOIDCProvider(OIDCProviderMixin, db.Model):
     A "pending" OIDC provider, i.e. one that's been registered by a user
     but doesn't correspond to an existing PyPI project yet.
     """
+
     __tablename__ = "pending_oidc_providers"
 
     discriminator = Column(String)
@@ -284,4 +285,6 @@ class PendingGitHubProvider(GitHubProviderMixin, PendingOIDCProvider):
         ),
     )
 
-    id = Column(UUID(as_uuid=True), ForeignKey(PendingOIDCProvider.id), primary_key=True)
+    id = Column(
+        UUID(as_uuid=True), ForeignKey(PendingOIDCProvider.id), primary_key=True
+    )
--- a/warehouse/oidc/models.py
+++ b/warehouse/oidc/models.py
@@ -17,8 +17,10 @@ import sentry_sdk
 
 from sqlalchemy import Column, ForeignKey, String, UniqueConstraint, orm
 from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.ext.declarative import declared_attr
 
 from warehouse import db
+from warehouse.accounts.models import User
 from warehouse.macaroons.models import Macaroon
 from warehouse.oidc.interfaces import SignedClaims
 from warehouse.packaging.models import Project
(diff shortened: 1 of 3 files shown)


warehouse/oidc/models.py
```diff
@@ -237,3 +258,35 @@ def job_workflow_ref(self):
...
+    __tablename__ = "github_oidc_providers"
+    __mapper_args__ = {"polymorphic_identity": "github_oidc_providers"}
+    __table_args__ = (
+        UniqueConstraint(
+            "repository_name",
+            "repository_owner",
+            "workflow_filename",
+            name="_github_oidc_provider_uc",
+        ),
+    )
```
woodruffw, line 273: Flagging for review: I had to dupe this `UniqueConstraint` between `GitHubProvider` and `PendingGitHubProvider`, even though they're functionally equivalent, thanks to the unique `name` requirement. There might be a better way to do this, though.

miketheman: I think this is due to not having set any naming_conventions for this project in the global metadata setup https://github.com/pypi/warehouse/blob/d58e6b241ce6fd2d65da469c32189b51a5572ff8/warehouse/db.py#L77
Alembic docs go into it here: https://alembic.sqlalchemy.org/en/latest/naming.html

I think we always named them manually, like here: https://github.com/pypi/warehouse/blob/d58e6b241ce6fd2d65da469c32189b51a5572ff8/warehouse/accounts/models.py#L197

So it might be a future refactoring to replace the manually-named constraints, but I’d approach that with some serious caution 😉

warehouse/oidc/models.py
```diff
@@ -161,28 +149,59 @@ def verify_claims(self, signed_claims: SignedClaims):
...
+
+class PendingOIDCProvider(OIDCProviderMixin, db.Model):
+    """
+    A "pending" OIDC provider, i.e. one that's been registered by a user
+    but doesn't correspond to an existing PyPI project yet.
+    """
+
+    __tablename__ = "pending_oidc_providers"
+
+    discriminator = Column(String)
```
di, line 188: This is duplicated in both `OIDCProvider` and `PendingOIDCProvider`, should it be in the `OIDCProviderMixin`?

di (approved):
LGTM overall, just the one comment about what should be duplicated vs. shared via mixins.

warehouse/oidc/models.py
woodruffw: Yeah, I think so! I'll do that.

woodruffw: Oh, I remember why I didn't do this -- `discriminator` is needed for `polymorphic_on`, which is specialized for each separate polymorphic hierarchy.

di: Doesn't it just need to be different (in the database) to support polymorphism? Why does it need to be on both classes?

woodruffw: Yeah, I overthought it -- I thought for some reason this wouldn't work:

```
    __mapper_args__ = {
        "polymorphic_identity": "oidc_providers",
        "polymorphic_on": OIDCProviderMixin.discriminator,
    }
```

...but it does, so I'll make these changes 🙂

woodruffw: ef67edd3

woodruffw:
Should be good to go now!

PR #12618, opened 2022-11-29
Make admin User pages route by username

di:
Fixes https://sentry.io/organizations/python-software-foundation/issues/3735347627/


ewdurbin:
I'm good with this, but could you add the user_id as a field in the detail view with this PR? When I need to reference that, I generally look at the URL.

PR #12626, opened 2022-11-30
Add Mastodon to custom icons for project_urls

adamchainz:
Use the [mastodon icon](https://fontawesome.com/icons/mastodon?s=&f=brands) for links titled “Mastodon”.


adamchainz:
I haven’t tested this but I did check that the icon exists in the version of fontawesome in the repo.

ewdurbin (approved):

PR #12664, opened 2022-12-10
chore(deps): pin packaging to upper boundary

miketheman:
`packaging` 22.0 was released recently, removing a dependency on `pyparsing`, which is cool.

Since `google-cloud-bigquery` pins to pre-22, we cannot update our dependencies until that is allowed and released, as we will have conflicts between what is installed in non `main.txt` requirements.

Refs: https://github.com/googleapis/python-bigquery/issues/1435


miketheman:
Python Dep PRs will currently fail a `make deps` check due to the mismatch in finding that `pyparsing` disappears from the installed scope without the version pin, blocking the overall pipeline of merging Python dependencies.

di (approved):

PR #12672, opened 2022-12-12
Upgrade to Python 3.11

ewdurbin:


ewdurbin:
While this is probably good to go for production we may want to wait for https://github.com/grpc/grpc/issues/30927 to be resolved, which will ship grpcio aarch64 wheels for Python 3.11 and improve experience for dev running on ARM machines.

ewdurbin:
Just for reference, dependency installation on my M1 went from 50.5s to 235.9s without grpcio wheels.

Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN gulp dist
 
 # Now we're going to build our actual application, but not the actual production
 # image that it gets deployed into.
-FROM python:3.10.4-slim-buster as build
+FROM python:3.11.0-slim-bullseye as build
```
miketheman, line 47: 🎯

Issue #12392, opened 2022-10-19
PEP 621-style license field shows the entire license text


GergelyKalmar:
**Describe the bug**
When using PEP 621-style project metadata with a license file (see https://peps.python.org/pep-0621/#license), the entire license text is shown in the Meta -> License section of a package on pypi. For an example see https://pypi.org/project/pytest-logikal/ and the relevant pyproject.toml file https://github.com/logikal-io/pytest-logikal/blob/main/pyproject.toml#L10. Showing the entire license text makes the Meta section almost useless by pushing the rest of the metadata quite far to the bottom.

**Expected behavior**
The license text should be hidden by default and perhaps shown with a modal upon clicking "see full text". Note that the issue is similar to https://github.com/pypi/warehouse/issues/1354, however, it is a little different due to the PEP 621 standard causing this behavior. With more and more projects adopting PEP 621 I would expect this to be a bigger issue over time.

di:
Some related discussion is here: https://github.com/FFY00/meson-python/issues/129

GergelyKalmar:
Thanks for the context! I think that the current UX needs to be improved regardless of how PEP 639 goes, the "Meta" section is really messed up for anyone who follows PEP 621 right now (which is the accepted standard at the moment). I think showing the (short) license name from the trove classifier there is more than enough (when available) with a fallback to the first line of the license text or so. PEP 639 would use different fields anyways, so even if that gets accepted you need to be able to deal with all the "legacy" approaches like PEP 639 or anything from before.

It seems to me that that area in the PyPI UI is not meant for displaying anything more than a short-ish line anyhow, so at the very least the excess text should be truncated/hidden.

GergelyKalmar:
I must add that currently setuptools just copies the contents of the referenced license file into the "License" field in PKG-INFO, so it can be interpreted that the contents of the field are the actual license text of the package. I can imagine that displaying the entire license on PyPI would be potentially useful, just not in the way it is done at the moment.

GergelyKalmar:
Lastly, I see what you mean by https://packaging.python.org/en/latest/specifications/core-metadata/#license and so I believe setuptools does the right thing, we should probably just not use the license field together with the trove classifiers (in light of the core metadata specification). Nonetheless, even if someone used the license field according to the PyPA specification and added a longer custom license text there, the Meta section on PyPI would be still messed up, so I think this is indeed a valid bug, regardless of the slightly confusing specifications surrounding license information.

[...]

PR #12692, opened 2022-12-18
docs: update JSON API response with current shape

miketheman:
As the API has evolved over time, update the example in the docs to reflect current keys.

Closes #8011


miketheman:
@di I don't know how "true" you wanted the response to be - it actually highlighted the "license" issue noted in #12392 and I manually trimmed the result for the docs.

PR #12702, opened 2022-12-19
chore: upgrade python syntax

miketheman:
Not super glamorous, but updated the codebase to take advantage of Python 3.11 syntax.

Things like:

- explicit `object` subclass
- f-string syntax
- `yield from`
- `set()` => `{}`


warehouse/forklift/legacy.py
```diff
@@ -948,8 +948,8 @@ def file_upload(request):
         if request.user:
             msg = (
                 (
-                    "The user '{0}' isn't allowed to upload to project '{1}'. "
-                    "See {2} for more information."
+                    "The user '{}' isn't allowed to upload to project '{}'. "
+                    "See {} for more information."
```
di, line 952: Should this become an f-string? I'm surprised it didn't catch this use of `.format`.

warehouse/forklift/legacy.py
```diff
@@ -961,8 +961,8 @@ def file_upload(request):
         else:
             msg = (
                 (
-                    "The given token isn't allowed to upload to project '{0}'. "
-                    "See {1} for more information."
+                    "The given token isn't allowed to upload to project '{}'. "
+                    "See {} for more information."
```
di, line 965: Ditto here.

di (approved):

warehouse/forklift/legacy.py
miketheman: I suspect it has something to do with the complex usage of `request.help_url(_anchor="project-name"),` and quotations - there's a test that shows this should be a noop: 
https://github.com/asottile/pyupgrade/blob/ff43f9a6ab4c37d0a89a4bb5ae338c2d0233b5eb/tests/features/fstrings_test.py#L34-L37

PR #12715, opened 2022-12-22
chore(deps): upgrade webpack to 5.x and webpack-stream to 7.x

miketheman:
A three-step process to get the latest versions of webpack available, so we can start to consider replacing aged gulp plugins with more modern webpack ones.

Removes ~130 total packages from the total installation, and removes 4 high vulnerabilities.


ewdurbin (approved):

som-design (commented):
ขอบคุณ​ครับ

PR #12733, opened 2022-12-30
Fix 11 test failures revealed by pytest-randomly

### LINKED ISSUE #12227: Implied state between tests, failing isolated tests
Found while testing #12222 

The entire test suite passes successfully, however isolating a single test out of the modified module fails, likely due to some sort of statefulness between tests. Example:

```shellsession
$ T=tests/unit/manage/test_views.py::TestManageOrganizationTeams::test_create_team_invalid make tests
...
>       assert form.name.errors == [
            "This team name has already been used. Choose a different team name."
        ]
E       AssertionError: assert [<warehouse.i...7f431df2dcf0>] == ['This team n...t team name.']
E         At index 0 diff: <warehouse.i18n.LazyString object at 0x7f431df2dcf0> != 'This team name has already been used. Choose a different team name.'
E         Use -v to get more diff

tests/unit/manage/test_views.py:3576: AssertionError
```
I confirmed this happens on `main` branch as well, so it wasn't part of that specific changeset.

In other projects I've included https://pypi.org/p/pytest-randomly/ early to suss out statefulness between tests, and running tests randomly on `warehouse` produced:

```shellsession
# after pytest-randomly is added to the deps:
$ make tests
...
Using --randomly-seed=3479063486
...
========================================================================== short test summary info ==========================================================================
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[True-eligible_projects5-None-None-False-You are not a collaborator on any critical projects]
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[False-eligible_projects1-None-user1-False-At this time there are no keys available] - Asse...
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_form - assert False
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_codes_available - assert False is True
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_eligible_projects_no_user - AssertionError: assert {'foo'} == set()
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[True-eligible_projects0-None-False-True-None] - AssertionError: assert {'eligible': ... 54...
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[True-eligible_projects3-None-user3-False-You already have two-factor authentication enabled with a hardware security key]
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_eligible_projects_pypi_mandates_2fa - AssertionError: assert {'foo'} == {'gsjwWwohYwmH'}
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_eligible_projects_owners_require_2fa - AssertionError: assert {'foo'} == {'VIykPxRsCel...eFAIJBqYwEvI'}
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_promo_code - AssertionError: assert None == TitanPromoCode(id=None, user_id=UUID('b8e37906-0036-4226-b3be-9...
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_codes_available_no_codes - assert True is False
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[True-eligible_projects4-promo_code4-user4-False-Promo code has already been generated] - A...
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_codes_available_no_unused_codes - assert True is False
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_promo_code_no_codes - assert <stub()> is None
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_promo_code_no_user - assert <stub()> is None
FAILED tests/unit/test_views.py::TestSecurityKeyGiveaway::test_default_response[True-eligible_projects2-None-user2-False-You are not a collaborator on any critical projects]
FAILED tests/unit/malware/test_checks.py::test_checks_subclass_base - AssertionError: assert 4 == 5
FAILED tests/unit/manage/test_views.py::TestManageOrganizationTeams::test_create_team_invalid - AssertionError: assert [<warehouse.i...7efe07867d90>] == ['This team n...t...
FAILED tests/unit/manage/test_views.py::TestManageTeamSettings::test_save_team_validation_fails - AssertionError: assert [<warehouse.i...7efe088bd810>] == ['This team n.....
FAILED tests/unit/manage/test_views.py::TestManageTeamRoles::test_create_team_role_not_a_member - AssertionError: assert [<warehouse.i...7efe0748b7f0>] == ['No organiza.....
========================================================= 20 failed, 3117 passed, 562 warnings in 139.58s (0:02:19) =========================================================
```

divbzero:
- Resolves #12227.

### DIFF
--- a/tests/unit/malware/test_checks.py
+++ b/tests/unit/malware/test_checks.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 import inspect
 
 import pretend
@@ -78,5 +79,9 @@ def test_base_prepare_nonfile_hooked(db_session):
         hooked_object = "Project"
 
     kwargs = FakeProjectCheck.prepare(request, file.id)
+
+    del FakeProjectCheck
+    gc.collect()
+
     assert request.route_url.calls == []
     assert "file_url" not in kwargs
--- a/tests/unit/malware/test_checks.py
+++ b/tests/unit/malware/test_checks.py
@@ -33,7 +33,7 @@ def test_checks_subclass_base():
         cls.__name__: cls for cls in MalwareCheckBase.__subclasses__()
     }
 
-    assert len(all_checks) <= len(subclasses_of_malware_base)
+    assert len(all_checks) == len(subclasses_of_malware_base)
 
     for check_name, check in all_checks:
         assert subclasses_of_malware_base[check_name] == check
--- a/tests/unit/malware/test_checks.py
+++ b/tests/unit/malware/test_checks.py
@@ -33,7 +33,7 @@ def test_checks_subclass_base():
         cls.__name__: cls for cls in MalwareCheckBase.__subclasses__()
     }
 
-    assert len(all_checks) == len(subclasses_of_malware_base)
+    assert len(all_checks) <= len(subclasses_of_malware_base)
 
     for check_name, check in all_checks:
         assert subclasses_of_malware_base[check_name] == check
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -628,7 +628,7 @@ class TestSecurityKeyGiveaway:
         with pytest.raises(HTTPNotFound):
             SecurityKeyGiveaway(request).security_key_giveaway()
 
-    def test_security_key_giveaway(self, monkeypatch):
+    def test_security_key_giveaway(self):
         request = pretend.stub(
             registry=pretend.stub(
                 settings={"warehouse.two_factor_mandate.available": True}
--- a/tests/unit/test_views.py
+++ b/tests/unit/test_views.py
@@ -628,13 +628,12 @@ class TestSecurityKeyGiveaway:
         with pytest.raises(HTTPNotFound):
             SecurityKeyGiveaway(request).security_key_giveaway()
 
-    def test_security_key_giveaway(self):
+    def test_security_key_giveaway(self, monkeypatch):
         request = pretend.stub(
             registry=pretend.stub(
                 settings={"warehouse.two_factor_mandate.available": True}
             )
         )
-        default_response = pretend.stub()
-        SecurityKeyGiveaway.default_response = default_response
+        view = SecurityKeyGiveaway(request)
 
-        assert SecurityKeyGiveaway(request).security_key_giveaway() == default_response
+        assert view.security_key_giveaway() == view.default_response
--- a/requirements/main.txt
+++ b/requirements/main.txt
@@ -378,8 +378,7 @@ dnspython==2.2.1 \
 docutils==0.19 \
     --hash=sha256:33995a6753c30b7f577febfc2c50411fec6aac7f7ffeb7c4cfe5991072dcf9e6 \
     --hash=sha256:5e1de4d849fee02c63b040a4a3fd567f4ab104defd8a5511fbbc24a8a017efbc
-    # via
-    #   readme-renderer
+    # via readme-renderer
 elasticsearch==7.10.1 \
     --hash=sha256:4ebd34fd223b31c99d9f3b6b6236d3ac18b3046191a37231e8235b06ae7db955 \
     --hash=sha256:a725dd923d349ca0652cf95d6ce23d952e2153740cf4ab6daf4a2d804feeed48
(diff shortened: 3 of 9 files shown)


tests/unit/test_views.py
```diff
@@ -628,13 +628,12 @@ def test_security_key_giveaway_not_found(self):
         with pytest.raises(HTTPNotFound):
             SecurityKeyGiveaway(request).security_key_giveaway()
 
-    def test_security_key_giveaway(self):
+    def test_security_key_giveaway(self, monkeypatch):
```
miketheman, line 631: **question:** ‏is `monkeypatch` needed here? I don't think it's being used, is it?

tests/unit/malware/test_checks.py
```diff
@@ -33,7 +33,7 @@ def test_checks_subclass_base():
         cls.__name__: cls for cls in MalwareCheckBase.__subclasses__()
     }
 
-    assert len(all_checks) == len(subclasses_of_malware_base)
+    assert len(all_checks) <= len(subclasses_of_malware_base)
```
miketheman, line 36: **thought:** ‏I'm not sure this is the desirable fix - since we currently do only have 4 real subclasses, at least until the `test_base_prepare_nonfile_hooked` creates one later on. Would it be better to update the selector to reject `FakeProjectCheck` from the list if found, or something else like removing the subclass? 

For reference, here's what `subclasses_of_malware_base` looks like when the order is flipped:

```python
{'ExampleHookedCheck': <class 'tests.common.checks.hooked.ExampleHookedCheck'>,
 'ExampleScheduledCheck': <class 'tests.common.checks.scheduled.ExampleScheduledCheck'>,
 'FakeProjectCheck': <class '__main__.FakeProjectCheck'>,
 'PackageTurnoverCheck': <class 'warehouse.malware.checks.package_turnover.check.PackageTurnoverCheck'>,
 'SetupPatternCheck': <class 'warehouse.malware.checks.setup_patterns.check.SetupPatternCheck'>}
```

miketheman (changes requested):
This is really cool! I checked out the code and ran the tests. I had to revert the last commit to enable randomly - curious why that is removed from scope?

I dropped some inline comments and thoughts - let me know if I can clarify any of them!

tests/unit/test_views.py
divbzero: You’re right, `monkeypatch` is not being used here. Must have been something else I was trying before I worked out this fix.

tests/unit/malware/test_checks.py
divbzero: In commit f36294c I tried the approach of cleaning up `FakeProjectCheck` after it is used for the test. Would that work alright or is one of the alternative approaches better?

divbzero:
> I had to revert the last commit to enable randomly - curious why that is removed from scope?

@miketheman My original plan was to leave `pytest-randomly` in, but as I was debugging the failures it started feeling weird to me that the test suite could succeed or fail randomly. To me it feels more solid if I can run `make tests` locally and have confidence that it will run exactly the same on GitHub.

That said, I also see the benefits of keeping `pytest-randomly` so I could be easily persuaded… Thoughts?

di:
I'm on the fence. On one hand, it would be nice to be able to find these types of issues in CI, but on the other hand, it would be annoying if the CI (and thus deployment pipeline) could randomly fail.

Perhaps as a compromise, we could add `pytest-randomly` as a dependency but configure it to be disabled in a normal run of the test suite, and instead set up a scheduled workflow to run the suite with it enabled?

miketheman (approved):
I'm good with the changes, thanks @divbzero ! I think we can merge this as is.

For future: I'll agree that the randomly-seeded testing can be confusing at first, but it'll definitely surface issues like these earlier, and at least broach the topic so we can use the random seed used in the test run to reproduce the issue.

@di I like the idea of running it periodically to get comfortable, with the regular suite using `pytest -p no:randomly` to disable the behavior. What level of confidence would you be looking for before adding it to the default test suite?

di:
I think that since any given PR could introduce tests that fail randomly (but might not fail in the PR introducing them?) I'd want to keep this in a scheduled job that runs ~daily, which should be enough to catch these sufficiently quickly.

di (approved):

PR #12751, opened 2023-01-04
Ordering API tokens by date created

ristomcgehee:
This change will make the API tokens on the Account Settings page be displayed in descending order of when they were created. That ordering makes the most sense to me, and that is how GitHub orders personal access tokens (as best I can tell).

Here's a screenshot of how the new ordering will look:
image

Closes https://github.com/pypi/warehouse/issues/9674


di (approved):

di:
Thanks!

PR #12755, opened 2023-01-04
Fixup: remove '.git' suffix from GitHub repository names as included in GitHub API requests

### LINKED ISSUE #12753: GitHub repository URLs that contain a '.git' suffix in the repository name segment are not normalized
**Describe the bug**
Python packages may include a list of project URLs in their metadata, and when PyPi detects a GitHub-related project URL, it attempts to retrieve statistical information (number of stars, forks, and so on) from the repository by using the GitHub API. 

It is valid for a GitHub repository to specify a URL that contains a `.git` suffix in the repository name part -- for example, `https://github.com/pypi/warehouse.git/`.

This is most commonly found in `git` `remote` references in cloned GitHub repositories, where the  remote reference typically includes the `.git` repository name suffix (it's also part of the HTTPS clone URL that GitHub suggests when using their web UI).

However, the GitHub API expects the repository name to be 'normalized' -- that is, repository type suffixes should not appear.  So to follow the previous example, the repository name `warehouse` (part of organization name `pypi`) should be provided as the repository name when retrieving statistics for this repository.

PyPi doesn't currently handle project URLs that include the `.git` suffix gracefully, and may display broken links and empty statistical fields for affected packages:

image

**Expected behavior**
The `.git` suffix in GitHub repository names could be detected and removed so that statistics can be retrieved for packages that include it within a project URL.

**To Reproduce**
The problem is currently visible following the latest release of `arcana-core` at v0.8.6 here:

- https://pypi.org/project/arcana-core/

And the repository URL configured for that release is visible here:

- https://github.com/ArcanaFramework/arcana-core/blob/238aa5abc46d613b0c5f12fcade354ee0db9feba/pyproject.toml#L66

(note the `.git` suffix in the repository name)

**My Platform**

N/A

**Additional context**
Similar to (and may include changes in a similar part of the code) to #12725.

jayaddison:
Resolves #12753.


tests/unit/packaging/test_models.py
```diff
@@ -430,6 +430,14 @@ def test_acl(self, db_session):
             ("https://google.com/pypi/warehouse/tree/main", None),
             ("https://google.com", None),
             ("incorrect url", None),
+            (
+                "https://www.github.com/pypi/warehouse.git",
+                "https://api.github.com/repos/pypi/warehouse",
+            ),
+            (
+                "https://www.github.com/pypi/warehouse.git/",
```
Julian, line 438: This duplicates the new parametrized example twice, no?

Julian: Oh never mind I see, trailing slash, ignore me.

tests/unit/packaging/test_models.py
```diff
@@ -430,6 +430,14 @@ def test_acl(self, db_session):
...
+                "https://api.github.com/repos/pypi/warehouse",
+            ),
```
di, line 440:
```suggestion
```

tests/unit/packaging/test_models.py
di: Oof, I missed this too 😅

di (approved):

di:
Thanks @jayaddison!

tests/unit/packaging/test_models.py
jayaddison: I sorta wondered about adding additional incorrectly-formatted GitHub test URLs, but then I didn't want to overspecify or add test URLs that shouldn't occur as input.  The trailing slash was a subtle difference there though, yep :)

jayaddison:
Thanks @di @Julian!

PR #12756, opened 2023-01-04
Improve the autodetection of GitHub URLs slightly

Julian:
- [x] strip off `.git` from the ends of any detected URL (#12753)
- [ ] skip URLs which aren't real repository links (e.g. `github.com/sponsors/`) when trying to find one in the list of links


Julian:
I just noticed #12755, so whoops, never mind.

Issue #12725, opened 2022-12-28
GitHub statistics (for 1 package) are using the wrong base URI / lead to 404s


Julian:
**Describe the bug**
The sidebar of https://pypi.org/project/jsonschema-specifications/ has links for GitHub statistics, but they lead to 404s.

The Stars link is currently trying to send someone to https://pypi.org/project/jsonschema-specifications/undefined/stargazers, and the Forks and Open Issues links are similarly somehow thinking that those links should redirect to somewhere under https://pypi.org/project/jsonschema-specifications/undefined.

All numbers are similarly not rendering (so the actual counts don't appear either).

**Expected behavior**
Similar counts and links as those which appear on other pages.

**To Reproduce**
https://pypi.org/project/jsonschema-specifications/ is the page I noticed this on.

Looking at another package I recently created (https://pypi.org/project/referencing/) shows that one seems fine, as do random other packages I've just browsed to check.

Not sure if the above is transient behavior that I've just never noticed, but filing since I don't see any existing issues mentioning it (unless I missed em).

**My Platform**

Firefox 109.0b2 (64-bit) on macOS Ventura 13.1

**Additional context**

Clearly this isn't hugely impactful, at least to me, filing mostly to track and in case it's wider spread and I just can't tell. Thank you all for everything!

<img width="386" alt="Screenshot 2022-12-27 at 19 51 11" src="https://user-images.githubusercontent.com/329822/209740980-ebb1e261-e733-4839-999b-ab05670af196.png">

di:
Looks like we're interpreting the URL `https://github.com/sponsors/Julian` as the repository URL, and that's causing the requests to the GitHub API to fail.

Two things I think we should do here:

a) fail more gracefully when the GitHub API gives us a 404
b) specifically exclude `/sponsors` (and maybe other things?) from this somewhat naive function: https://github.com/pypi/warehouse/blob/b06c866306d639c895e74d79b200dcefd9d435f4/warehouse/packaging/models.py#L587-L594

jayaddison:
This seems to affect more than GitHub-sponsor-path-URLs - the [`arcana-core`](https://pypi.org/project/arcana-core/) package similarly displays empty statistical information:

![image](https://user-images.githubusercontent.com/55152140/210579958-b58e7f73-0e9f-4cac-9575-3a1e8da7bc69.png)

[...]

PR #12757, opened 2023-01-04
Skip non-user URLs when autodetecting a GitHub repository

Julian:
Uses a hardcoded list of known examples (from a new dependency).

Closes: #12725

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ module = [
     "citext.*",
     "disposable_email_domains.*",
     "elasticsearch_dsl.*", # https://github.com/elastic/elasticsearch-dsl-py/issues/1533
+    "github_reserved_names.*",
     "google.cloud.*",
     "IPython.*",
     "mistune.*",
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -14,6 +14,7 @@ disposable-email-domains
 elasticsearch>=7.0.0,<7.11.0
 elasticsearch_dsl>=7.0.0,<8.0.0
 first
+github-reserved-names>=1.0.0
 google-cloud-bigquery
 google-cloud-storage
 hiredis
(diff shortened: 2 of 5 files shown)


Julian:
(Leaving as draft -- I didn't investigate improving the 404 behavior generally, will see if I get any additional time to do so, though what's here should fix the specific example(s) already)

jayaddison:
I might be able to help out with the error-response handling side of this: after looking at how the frontend handles stats display currently, I've opened #12759, which displays some of the existing code (I'm not sure yet whether the suggested fix is a good one though).

Julian:
Ah amazing, well done, yeah that looks nice!

jayaddison:
Thanks!  The name-filtering approach here looks good, too.  Something probably worth checking: is the `COPYING` file from `github-reserved-names` included when it's built into a package?

Julian:
> is the COPYING file from github-reserved-names included when it's built into a package?

Looks like it is yeah, let me know if you were just asking or if you'd seen it not be included somehow.

jayaddison:
Thanks - I'd taken a look at the [`pyproject.toml` license configuration](https://github.com/Julian/github-reserved-names/blob/228f0fc3c0cb9c6d83c4a6177aab9fd170fb9374/pyproject.toml#L12) and wasn't sure whether the `COPYING` file would be bundled when a `text` license is specified.

The reason I asked is that I noticed you credit the NPM source package -- the correct thing to do -- and I wanted to check that that work and reference is included in the distributed package (and indirectly in `pypi/warehouse` as well).

(I've recently worried about making mistakes related to copyright and attribution in my own projects - and so it was on my mind to check for similar problems elsewhere; even at the risk of seeming like asking difficult questions.  on balance I figured it was better to ask.  in retrospect I probably should've just downloaded the package and checked myself)

Julian:
Ah no worries at all, it's a helpful question! And even with the laziness, better to ask and have me check than stay quiet :D -- yeah don't worry bout that, I'm lazy enough for all of us so I don't mind offloading a bit of it hah.

Anyways, now I don't really understand why CI is complaining about dependencies, so trying to figure out what it expects me to run beyond `pip-compile`.

di:
@Julian Might want to check the version of `pip-tools` you have, I see some extraneous changes here.

PR #12758, opened 2023-01-04
chore(deps): cleanup some babel

miketheman:
As of Babel 7.4.0, the polyfill package is deprecated.

ECMA2015 (ES6) is also now widely-enough in use that we can leverage the
syntax directly, and cut ~50% of the overall JS bundle size.

Refs: https://babeljs.io/docs/en/babel-polyfill
Refs: https://caniuse.com/es6

A couple of other commits with descriptions.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,5 @@
 {
   "presets": [
     "@babel/preset-env"
-  ],
-  "plugins": [
-    "@babel/plugin-proposal-class-properties"
   ]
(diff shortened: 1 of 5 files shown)


miketheman:
Bundle sizes, before ~207KiB and after ~120KiB:

image

image

(imagery thanks to Chrome's Lighthouse)

di (approved):

PR #12759, opened 2023-01-05
Frontend: GitHub repo info controller: hide statistics when API responses have an error status

jayaddison:
Relates to #12725.

### DIFF
--- a/tests/frontend/github_repo_info_controller_test.js
+++ b/tests/frontend/github_repo_info_controller_test.js
@@ -69,6 +69,32 @@ describe("GitHub Repo Info controller", () => {
     });
   });
 
+  it("not-found response hides", (done) => {
+    fetch.mockResponse(
+      JSON.stringify({
+        html_url: "https://github.com/ghost/none",
+        stargazers_count: 100,
+        forks_count: 200,
+        open_issues_count: 300,
+      }),
+      { status: 404 }
+    );
+
+    startStimulus();
+    mountDom();
+
+    setTimeout(() => {
+      try {
+        const el = document.getElementById("github-repo-info");
+        expect(el).toHaveClass("hidden");
+        expect(fetch.mock.calls.length).toEqual(2);
+        done();
+      } catch (error) {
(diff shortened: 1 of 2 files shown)


Julian:
Ah amazing, well done, yeah that looks nice!

tests/frontend/github_repo_info_controller_test.js
```diff
@@ -69,6 +69,32 @@ describe("GitHub Repo Info controller", () => {
     });
   });
 
+  it("not-found response hides", (done) => {
+    fetch.mockResponse(
+      JSON.stringify({
+        html_url: "https://github.com/ghost/none",
+        stargazers_count: 100,
+        forks_count: 200,
+        open_issues_count: 300,
```
jayaddison, line 78: Note: this is mostly copied from the mock valid API response in the same file.  Perhaps there'd be alternative, better test fixture data to return from here.

miketheman: The current 404 response looks like this:
```json
{
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/reference/repos#get-a-repository"
}

miketheman (changes requested):
This is a good enhancement! Please update the mock response to reflect the current API behavior.

miketheman (approved):

jayaddison:
> We could theoretically make these server-side, but would add overhead, and likely need to create a GitHub app to allow for the API traffic.

Yep, I think client-side makes sense here.

I'm wondering whether hiding the stats panel entirely is appropriate, though, or whether some other status-related information should be displayed (pros: more transparent, and makes use of the work already-performed by making the request -- cons: additional implementation work, and a risk that the visual space used might not be justified by the value of the error messaging).

di:
I think it's OK to hide the panel for now.

We make these requests on the client side intentionally, because otherwise we would quickly exhaust API quota to retrieve them.

di (approved):

PR #12771, opened 2023-01-06
issue two purges for fastly surrogate keys

ewdurbin:
https://developer.fastly.com/learning/concepts/purging/#race-conditions states there are some race conditions we may be running into with shielding enabled.

ref: #12214


ewdurbin:
We're still working with Fastly on a more elegant solution, but this may improve things in the meantime at the cost of using up lots more celery worker time.

PR #12773, opened 2023-01-06
Frontend: GitHub repo info controller: debounce API requests within a one-second time window

### LINKED ISSUE #12763: Frontend: GitHub API requests may be invoked twice on package details page
**Describe the bug**
On the package details page (for example, https://test.pypi.org/project/example/ ), the frontend JS application may make multiple requests to the GitHub API if the package contains URLs that look like links to GitHub repositories.

**Expected behavior**
A single network call to the GitHub API should occur for statistics retrieval.

**To Reproduce**
- Monitor outbound HTTP requests (for example, by enabling built-in web browser developer tools)
- Visit https://test.pypi.org/project/example/ (a package that contains a GitHub URL reference)
- Observe that multiple requests to the host `api.github.com` occur

image

**My Platform**
Firefox 102, Linux

**Additional context**
The controller code that manages this appears to [load twice](https://github.com/pypi/warehouse/blob/f89c8cbeb999458baa9fadf43ee920b659dbe99b/warehouse/static/js/warehouse/controllers/github_repo_info_controller.js#L27-L30) - maybe that could be the cause?

Also: multiple-`connect`-calls appears to be an expected behaviour of the [`stimulus` library](https://github.com/hotwired/stimulus/) that the controller uses: https://github.com/hotwired/stimulus/issues/215#issuecomment-450552862 - perhaps migrating to the [`initialize` function](https://github.com/hotwired/stimulus/issues/215#issuecomment-450561022) could help?

jayaddison:
May resolve #12763.

### DIFF
--- a/warehouse/static/js/warehouse/controllers/github_repo_info_controller.js
+++ b/warehouse/static/js/warehouse/controllers/github_repo_info_controller.js
@@ -12,6 +12,9 @@
  * limitations under the License.
  */
 import { Controller } from "@hotwired/stimulus";
+import { debounce } from "debounce";
+
+const debouncedFetch = debounce(fetch, 1000);
 
 export default class extends Controller {
   static targets = [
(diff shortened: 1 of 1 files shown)


jayaddison:
I don't think that this is a great approach, since it doesn't follow patterns commonly found/recommended by the framework.  Closing pending better alternatives.

PR #12774, opened 2023-01-06
Replace deprecated sass-lint for stylelint

miketheman:
[sass-lint](https://github.com/sasstools/sass-lint#readme) has been deprecated since 2017, here's what's on their README:

Screenshot 2023-01-06 at 1 50 02 PM

I appreciate they are trying to get their message across. Out with the old, in with the new!

[Stylelint](https://stylelint.io/) started in 2014, and has become a common utility for Sass authors.

The resulting config rules is an attempt to preserve current structure and minimize changes,

See commits for specific changes to make review easier.

Resolves some open vulnerability reports. Results in `npm audit` from:
```
30 vulnerabilities (1 low, 14 moderate, 14 high, 1 critical)
```
to:
```
18 vulnerabilities (1 low, 9 moderate, 8 high)
```

**Note**: I attempted to also add the https://www.npmjs.com/package/stylelint-itcss package, but our ITCSS structure is not 100% standard, and raised a fair amount of linter issues on its own, so I'll leave that off for another time.


warehouse/static/sass/blocks/_vertical-tabs.scss
```diff
@@ -183,7 +183,7 @@
     }
 
     @media only screen and (max-width: $tablet) {
-      margin: 25px 10px 10px 10px;
+      margin: 25px 10px 10px;
```
ewdurbin, line 186: kinda gross that the new once prefers this way less obvious three value version just cause r/l are same margin, but OK

di (approved):

warehouse/static/sass/blocks/_project-description.scss
```diff
@@ -55,16 +55,6 @@
 
   // Remove top margin on elements at the top of the description
   > div:first-child > *:first-child,
-  > *:first-child,
```
ewdurbin, line 58: I do not understand this... do we know why it was repeated before?

warehouse/static/sass/blocks/_project-description.scss
```diff
@@ -277,23 +268,22 @@
...
+$light-blue:    #0078bd;
 $dark-blue:     darken($primary-color, 10);
 $darkest-blue:  darken($primary-color, 20);
 $green:         darken($success-color, 5);
 $light-green:   $success-color;
 $red:           $danger-color;
 $dark-red:      darken($danger-color, 20);
-$purple:        #bb0066;
-$orange:        #A36200;
+$purple:        #b06;
```
ewdurbin, line 278: is there a rule to prefer "long" hex codes? this seems unnecessarily golfy.

warehouse/static/sass/blocks/_project-description.scss
miketheman: Originated during this change: https://github.com/pypi/warehouse/commit/a5be8a7682e735a7e7391d582461c470a06de711

PR #12777, opened 2023-01-07
Frontend: GitHub repo info controller: refactor to use stimulus outlets

### LINKED ISSUE #12763: Frontend: GitHub API requests may be invoked twice on package details page

jayaddison:
This changeset updates the frontend application by introducing a singleton `github-repo-stats` stimulus controller, which is connected to the existing two `github-repo-info` controllers using `outlet` functionality that was added to [`stimulus` in version v3.2.0](https://github.com/hotwired/stimulus/tree/3353b84210ce3e203cddb5a4df3517ff181e5176).

With these changes in place, the singleton controller is responsible for calling the GitHub API, and then it sends the resulting extracted statistics to each connected info outlet controller for display.

This avoids duplicate GitHub API requests on each package detail page view.

Resolves #12763.


warehouse/static/js/warehouse/controllers/github_repo_stats_controller.js
```diff
@@ -0,0 +1,44 @@
...
+          bugs_url: data.html_url + "/issues",
+          followers: data.stargazers_count,
+          followers_url: data.html_url + "/stargazers",
+          forks: data.forks_count,
+          forks_url: data.html_url + "/network/members",
+        };
+        this.githubRepoInfoOutlets.forEach(outlet => outlet.updateStats(stats));
+      })
+      // swallow errors, we don't want to show them to the user
+      .catch(() => {});
```
jayaddison, line 42: Development note: I think that this `catch` statement may be masking a failure that has appeared at some point in the test code (mocking a `null` fetch response body fails if this catch is removed).

tests/frontend/github_repo_info_controller_test.js
```diff
@@ -15,18 +15,23 @@
 
 import { Application } from "@hotwired/stimulus";
 import GitHubRepoInfoController from "../../warehouse/static/js/warehouse/controllers/github_repo_info_controller";
+import GitHubRepoStatsController from "../../warehouse/static/js/warehouse/controllers/github_repo_stats_controller";
```
jayaddison, line 18: I'm trying to decide if it makes sense to split this file into two separate test modules (this existing file for `github_repo_info_controller`, and potentially a separate file for `github_repo_stats_controller`).

Their behaviour is fairly tightly-coupled though - I think by necessity, since they need to agree on the format of the `stats` object used during communication.

miketheman: Thanks for calling this out - it's making me wonder if this is worth it.

We're adding another controller to handle the existence of two controllers in the HTML and synchronize their output via the new Outlets API. That is indeed clever, and I've confirmed that it works.

I feel that the "right" solution would be to reduce the double-inclusion of the controller in the HTML, this could be considered a `TODO` to resolve/remove at a later date.

BUT if we're to proceed with this for now, the test should be updated to reflect the behavior we're changing - notably that there are two `github-repo-info` elements, and assert in the successful test that we only make a single fetch call (i.e. leave `expect(fetch.mock.calls.length).toEqual(3);` should not change as a result of the second HTML fragment being included.

tests/frontend/github_repo_info_controller_test.js
```diff
@@ -15,18 +15,23 @@
...
 const mountDom = async () => {
   document.body.innerHTML = `
+    <div id="github-repo-stats"
+          data-controller="github-repo-stats"
+          data-github-repo-stats-github-repo-info-outlet="#github-repo-info">
+          data-github-repo-stats-url-value="https://api.github.com/repos/pypi/warehouse">
     <div id="github-repo-info"
           class="hidden"
           data-controller="github-repo-info"
-          data-github-repo-info-url-value="https://api.github.com/repos/pypi/warehouse">
```
miketheman, line 29: **nitpick:** ‏Removing this line removed the closing `>` for this `div`

warehouse/templates/packaging/detail.html
```diff
@@ -392,6 +392,13 @@ <h3>
 
         {% endif %}
       </div>
+    {% if release.github_repo_info_url %}
+      <div class="hidden"
+        data-controller="github-repo-stats"
+        data-github-repo-stats-github-repo-info-outlet=".github-repo-info"
+        data-github-repo-stats-url-value="{{release.github_repo_info_url}}">
+      </div>
+    {% endif %}
```
miketheman, line 401: **issue:** ‏This change breaks the visual layout on desktop - resulting on placing the content in the wrong place.

Screenshot 2023-01-08 at 3 10 11 PM


I _think_ this wrapping `div` would have to come much earlier to apply to both controllers and not impact the layout, which adds to the complexity of this solution. 🤔

miketheman (changes requested):
Thanks for continuing to chip away at this, @jayaddison !

I added inline comments - let me know if I can clear anything up! There's a current breakage that would need to be addressed first as well.

warehouse/templates/packaging/detail.html
jayaddison: Hm, ok.  I'll try relocating this to near the beginning of the `content` template block, I think.

jayaddison: Relocated, but: I haven't yet confirmed that this displays correctly.  I'll verify that soon.

jayaddison: This should now be resolved (`div` relocated in 036dd5c55038e8aecba0cd850ecb5b4c15127bf7).

(I considered using a different type of HTML element (instead of `div`) to associate with the controller - but I didn't find one that seemed to be a precise match for the use-case, and continuing to use a `div` seems straightfoward)

jayaddison:
The pull request CI test failure on ab6afb34e6c385b9bfbaf9917e9f931a0b9bbbd8 seems to be due to a temporary network failure; I'll push either an empty commit or a merge-from-main in a while to re-run CI.

warehouse/templates/packaging/detail.html
miketheman: Looks like the change worked!

miketheman (approved):

jayaddison:
Thanks @miketheman!

jayaddison:
Although I'd be happy to see this merged, I'm also aware that the Outlets API in stimulus is a fairly recent addition -- and that the audience of the package warehouse is fairly large -- so I'd be totally OK with waiting for additional stability/maturity of that interface to improve.

(there are a [couple of open bugs](https://github.com/hotwired/stimulus/issues?q=is%3Aissue+is%3Aopen+outlets) related to outlets at the moment, although neither of them affect this changeset as far as I'm aware.  that and the potential for other bugs could be weighed up against the overheads, I suppose - I'm not really sure how to equate them in terms of cost/benefits)

di (approved):

di:
Thanks, nice job!

jayaddison:
Thank you!

Issue #6606, opened 2019-09-10
Fill API token form when user arrives from manage project page


nlhkabu:
## Current workflow:

1. User visits 'project settings' page in the PyPI admin
2. User clicks on a link to create an API token for the project in question
3. User arrives on 'create API token' page in their account settings

## Ideal workflow:

1. User visits 'project settings' page in the PyPI admin
2. User clicks on a link to create an API token for the project in question
3. User arrives on 'create API token' page in their account settings **with the project preselected in the 'create token' form**

Relevant comments:

- https://github.com/pypa/warehouse/pull/6586#issuecomment-528210853
- https://github.com/pypa/warehouse/pull/6586#issuecomment-528428456

PR #12022, opened 2022-08-05
Use POST-303-GET pattern for all manage views

divbzero:
- Searched for POST views using `awk` commands like the following:

      grep -r -l 'request_method="POST"' warehouse
      grep -r -l 'require_methods=\["POST"\]' warehouse
      grep -r -l 'request\.method .= "POST"' warehouse

      cat warehouse/manage/views.py | awk '/request_method="POST"/{flag=1}/(class|def) / && flag{print $2; flag=0}'
      cat warehouse/manage/views.py | 'awk /require_methods=\["POST"\]/{flag=1}/(class|def) / && flag{print $2; flag=0}'
      tac warehouse/manage/views.py | 'awk /request\.method .= "POST"/{flag=1}/(class|def) / && flag{print $2; flag=0}'

- Fixed POST-303-GET pattern for 12 of 45 POSTs in warehouse/manage/views.py (see commit a59d767 for details).

- Confirmed use of POST-303-GET pattern in the following files:

  - warehouse/accounts/views.py
  - warehouse/admin/views/banners.py
  - warehouse/admin/views/checks.py
  - warehouse/admin/views/emails.py
  - warehouse/admin/views/flags.py
  - warehouse/admin/views/organizations.py
  - warehouse/admin/views/prohibited_project_names.py
  - warehouse/admin/views/projects.py
  - warehouse/admin/views/sponsors.py
  - warehouse/admin/views/users.py
  - warehouse/admin/views/verdicts.py
  - warehouse/email/ses/views.py
  - warehouse/forklift/legacy.py
  - warehouse/integrations/github/views.py
  - warehouse/integrations/vulnerabilities/osv/views.py
  - warehouse/legacy/api/pypi.py
  - warehouse/legacy/api/xmlrpc/views.py
  - warehouse/manage/views.py

- Found exceptions to POST-303-GET pattern but chose not to fix:

  - `create_macaroon` in warehouse/manage/views.py
    https://github.com/pypi/warehouse/blob/a59d767/warehouse/manage/views.py#L984-L1051

  - `security_key_giveaway_submit` in warehouse/views.py
    https://github.com/pypi/warehouse/blob/a59d767/warehouse/views.py#L549-L574

- Fixes #11778.

PR #12778, opened 2023-01-08
Fill api token form when user arrives from manage project page

ristomcgehee:
This PR will make it so that when the user is on the manage project page, and they click "Create a token for <project_name>", it will have that project selected on the "Add API token" page.

I used query parameters to pass the project name, and I noticed the query parameters were not getting propagated through the reauth form, so I modified that to pass the query parameters.

Closes https://github.com/pypi/warehouse/issues/6606


di:
Thanks for working on this @ristomcgehee!

I think one slightly simpler way to go about this would be to make the "Create a token for ..." button submit a POST request to `/manage/account/token/` with the scope already set (but the token name missing).

That would result in the user landing on a page like this:

image

And we wouldn't need to add support for the query parameter, change the reauthentication form, etc.

ristomcgehee:
If we were to change the "Create a token for ..." button from a GET to a POST action, then the user would no longer be able to open that button in a new tab. This would slightly reduce the usability, I think. Also, I don't think it's very user friendly to show an error message to the user right when the page loads. With that said, do you still want the button to be a POST instead of GET?

By the way, I heard the episode you were in on the Sustain podcast. It was cool learning some of the history of PyPI!

di:
> If we were to change the "Create a token for ..." button from a GET to a POST action, then the user would no longer be able to open that button in a new tab. This would slightly reduce the usability, I think. Also, I don't think it's very user friendly to show an error message to the user right when the page loads. With that said, do you still want the button to be a POST instead of GET?

That's a fair point! Let me give this a closer review in that case.

> By the way, I heard the episode you were in on the Sustain podcast. It was cool learning some of the history of PyPI!

Thanks! If you want more history, this talk covers a lot of it: https://www.youtube.com/watch?v=AQsZsgJ30AE

ristomcgehee:
@di just wanted to give a friendly reminder to review this when you have the time.

ristomcgehee:
@miketheman or @ewdurbin, would one of you be able to review this PR?

miketheman (commented):
I was poking at the behavior of this code, and noted a couple of things:

1. The project dropdown gets populated correctly from the `?q=` param, cool!
1. The resulting URL after the user has POSTed create token remains with the `?q=` param - which is no longer correct - since now the user could have chosen a different project in the dropdown, yet their URL is still pointing to the previous project name

I didn't get into the reauth behaviors yet, but wanted to provide initial notes.

ristomcgehee:
Thanks for reviewing, Mike.

> The resulting URL after the user has POSTed create token remains with the `?q=` param - which is no longer correct - since now the user could have chosen a different project in the dropdown, yet their URL is still pointing to the previous project name

That's a good catch. I don't see an easy way of removing the `?q=` param from the URL after the user submits the POST create token request, unless we want to change [create_macaroon()](https://github.com/pypi/warehouse/blob/616130795e4fc5ed4bded04af9af8707ef6631a8/warehouse/manage/views.py#L1008) to redirect to a GET at the end of the method. I could change the query name from `q` to `initial_project_selection` to clarify its use. That way I don't think a user would be confused if they chose a different project and then see the previous project name in the URL. Would you like me to make that change?

miketheman:
> unless we want to change [create_macaroon()](https://github.com/pypi/warehouse/blob/616130795e4fc5ed4bded04af9af8707ef6631a8/warehouse/manage/views.py#L1008) to redirect to a GET at the end of the method.

Definitely a good callout - looking at that one, we can see that this was explicitly not converted in #12022 - so I'll consder that out of scope for this changeset.

@ristomcgehee Please resolve the merge conflicts - likely by merging upstream and re-running `make translations` to get the updated line references.

@di tagging you for review

miketheman (approved):

ristomcgehee:
@di I know this has been open for a while, but I'm still willing to work on this if you review it.

warehouse/accounts/views.py
```diff
@@ -1351,7 +1352,9 @@ def reauthenticate(request, _form_class=ReAuthenticateForm):
 
     if form.next_route.data and form.next_route_matchdict.data:
         redirect_to = request.route_path(
-            form.next_route.data, **json.loads(form.next_route_matchdict.data)
+            form.next_route.data,
+            **json.loads(form.next_route_matchdict.data),
+            _query=json.loads(form.next_route_query.data),
```
di, line 1357: I think we want to protect against `form.next_route_matchdict.data` potentially containing a `_query` key here. Something like:

```suggestion
            **json.loads(form.next_route_matchdict.data) | dict(_query=json.loads(form.next_route_query.data)),
```

ristomcgehee:
@di Thanks for the review. I've accepted your suggestions, so I'm ready for another review.

di (approved):

di:
Thanks for your patience here @ristomcgehee!

PR #12780, opened 2023-01-08
refactor: replace imagemin with sharp/svgo

miketheman:
The `imagemin` and packages bring in a lot of unpactched dependencies
and require C extensions to be pre-installed.

`sharp` is self-contained and can produce smaller, more efficient `webp`
formats.

After this change, we remain with:

    3 vulnerabilities (1 low, 2 high)

A second commit refactors the Dockerfiles to simplify our setup and maintain parity.


miketheman:
Gut-check the Dockerfile collapsing - it definitely works, but I'm not sure if it works for the workflows we intend to support.

ewdurbin (approved):

PR #12819, opened 2023-01-12
remove ddtrace

ewdurbin:
we added this back in July to hunt down performance issues, but it hasn't really been used much since.

seems the most recent release is causing some issues, combined with the rather large dependency tree it may make sense to reconsider how we collect these kinds of metrics. perhaps bringing it back in in its own worker class that we can send a portion of traffic to.


di (approved):

miketheman:
https://github.com/DataDog/dd-trace-py/pull/4895 Appears related.

ewdurbin:
I'll close this as we're not seeing as much trouble lately!

PR #12860, opened 2023-01-18
Origin Cache - Fastly: Make API endpoint configurable

ewdurbin:
In order to address #12214 we are being provided an endpoint that will allow us to issue our purges to a specific endpoint, so we'll support this then update the configuration to control behavior.


BinaryVixen899:
Excited to see this solved :)

di (approved):

ewdurbin:
Next step is to update the configuration. I checked and looks like the new endpoint isn't serving valid TLS.

PR #12872, opened 2023-01-19
feat(admin): surface email breach counts in UI

miketheman:
Based on https://www.troyhunt.com/pwned-or-bot/

Implement interface and service to retrieve breach counts for emails.

Previous HIBP API usage did not require an API Key, but the search by email account functionality requires both an API key and a custom header.

### DIFF
--- a/warehouse/accounts/services.py
+++ b/warehouse/accounts/services.py
@@ -840,6 +840,10 @@ class HaveIBeenPwnedEmailBreachedService:
         See https://haveibeenpwned.com/API/v3#BreachesForAccount
         """
 
+        # bail early if no api key is set, so we don't send failing requests
+        if not self.api_key:
+            return "!!!"
+
         try:
             resp = self._http.get(
                 urllib.parse.urljoin(self._api_base, email),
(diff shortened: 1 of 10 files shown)


warehouse/accounts/services.py
```diff
@@ -812,3 +814,58 @@ def check_password(self, password, *, tags=None):
...
+                urllib.parse.urljoin(self._api_base, email),
+                headers={"User-Agent": "PyPI.org", "hibp-api-key": self.api_key},
+            )
+            resp.raise_for_status()
+        except requests.RequestException as exc:
+            # 404 is expected if the email has **not** been breached
+            if exc.response.status_code == http.HTTPStatus.NOT_FOUND:
+                return 0
+            logger.warning("Error contacting HaveIBeenPwned: %r", exc)
+            return "?"
```
di, line 858: Feels a bit weird that this returns some semi-cryptic strings instead of integers when things are failing. I think if we ever unset the API key, we would just want this to stop attempting to hit the API and disappear from the UI rather than return `!!!` when it gets here for every email. Similarly for the request exceptions, maybe it'd be better to just return `-1`?

miketheman: Thanks for the critique!
I like that approach - so the return type changes to `int | None`. `None` would then be a signal to the UI to not display the column at all, and -1 is an error condition?

di (approved):

PR #12887, opened 2023-01-24
Update URL to public dataset

di:
Updating one last reference to our old BigQuery dataset in favor of the public dataset.


ewdurbin (approved):

kingjay66 (commented):
fix

PR #12888, opened 2023-01-24
Remove 'trending projects' and 'latest releases' from homepage

### LINKED ISSUE #3252: Assess libraries.io trending projects feed, consider replacing ours
The [Libraries.io list of trending PyPI projects](https://libraries.io/trending?platforms=PyPI) is very different from the one on the http://pypi.org/ front page. Our list just computes a zscore of downloads. We should check their data & methods, and if their results are more interesting or better than ours, we should [hit their API](https://libraries.io/api) and use their list instead of ours on our front page.

From [IRC today](http://kafka.dcpython.org/day/pypa/2018-03-14).

### LINKED ISSUE #3447: add "search by new release" and "trending" sort orders, link to them from homepage
Testing the new warehouse today at https://test.pypi.org, the list for "New Releases Hot off the press: the newest project releases" is only 5 items long. I can not find a link to view a longer list.

The older PyPI listing of 40 most recent packages was more usable given the rate of package uploads. A few projects upload many packages with each release.

A paged history of package uploads going back a day or two would be a welcome addition.

### LINKED ISSUE #8990: The `compute_trending` task is using the old download dataset
The `compute_trending` task is currently configured to use the old `the-psf.pypi.downloads*` dataset instead of the new clustered/partitioned `the-psf.pypi.file_downloads`.

https://github.com/pypa/warehouse/blob/aafc5185e57e67d43487ce4faa95913dd4573e14/warehouse/packaging/tasks.py#L25-L103

di:
Closes #8990, closes #3447, closes #3252.

### DIFF
--- a/tests/unit/packaging/test_init.py
+++ b/tests/unit/packaging/test_init.py
@@ -22,15 +22,13 @@ from warehouse.packaging.interfaces import IDocsStorage, IFileStorage, ISimpleSt
 from warehouse.packaging.models import File, Project, Release, Role
 from warehouse.packaging.tasks import (  # sync_bigquery_release_files,
     compute_2fa_mandate,
-    compute_trending,
     update_description_html,
 )
 
 
-@pytest.mark.parametrize("with_trending", [True, False])
 @pytest.mark.parametrize("with_bq_sync", [True, False])
 @pytest.mark.parametrize("with_2fa_mandate", [True, False])
-def test_includeme(monkeypatch, with_trending, with_bq_sync, with_2fa_mandate):
+def test_includeme(monkeypatch, with_bq_sync, with_2fa_mandate):
     storage_class = pretend.stub(
         create_service=pretend.call_recorder(lambda *a, **kw: pretend.stub())
     )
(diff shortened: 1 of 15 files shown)


warehouse/migrations/versions/f93cf2d43974_remove_project_zscore.py
```diff
@@ -0,0 +1,42 @@
...
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "f93cf2d43974"
+down_revision = "62601ddf674c"
+
+
+def upgrade():
+    op.drop_column("projects", "zscore")
```
ewdurbin, line 30: We should split this migration into a second PR that goes out after the code changes that stop referencing it.

ewdurbin (changes requested):

warehouse/migrations/versions/f93cf2d43974_remove_project_zscore.py
di: https://github.com/pypi/warehouse/pull/12889

dstufft (approved):

ewdurbin (changes requested):
I would have expected to see the zscore column removed from the Project model in this PR.

ewdurbin (approved):

PR #12978, opened 2023-02-10
Run CI workflow on merge_group

alex:


.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request:
+  merge_group:
```
dstufft, line 7: Should we change this to:

```
on:
  merge_group:
    types: [checks_requested]
```

The Github docs suggests doing that to protect against future other types of events?

dstufft (approved):

miketheman (approved):

Issue #12971, opened 2023-02-09
Split 'Open issues/PRs' to display the values separately 


HallvardMM:
**What's the problem this feature will solve?**
Change the GitHub statistics from:
 `Open issues/PRs: number`

Display it like this:
```
Open issues: number 
Open PRs: number
```

 Example:
![image](https://user-images.githubusercontent.com/49036938/217904590-1aadbb2c-dfe8-4402-888a-a0de4c953732.png)

It would help give a clearer understanding to users of how many issues and PRs are open for a package.
One example of why I argue today's display is somewhat problematic:
The [Django](https://pypi.org/project/Django/) package looks like it has 149 issues/PRs 

![image](https://user-images.githubusercontent.com/49036938/217906574-600e9172-da68-439e-b7ed-340e0785a9ce.png)

The [Pandas](https://pypi.org/project/pandas/) package looks like it has 3681 issues/PRs 

![image](https://user-images.githubusercontent.com/49036938/217906499-3dde7a9d-c613-4d91-8611-4c0363f13b75.png)

The difference makes Django seem way better maintained, however, the difference is linked to Django not displaying issues on GitHub. The total number of open PRs is about the same for the two projects.

**Additional context**
Would it be possible to use this solution?
https://github.com/isaacs/github/issues/536#issuecomment-532884919 

Use `https://api.github.com/repos/OWNER/REPO` to get open PRs and open Issues count from `"open_issues": number,` and
`https://api.github.com/search/issues?q=repo:OWNER/REPO+type:issue+state:open&per_page=1`
to get the open issues count from `"total_count": number,` and then subtract `open_issues`-`total_count` to get open PRs?

This is linked to #3697 and #3673

miketheman:
The current code for fetching and displaying the issues can be found here:

https://github.com/pypi/warehouse/blob/8d8cc16f0e0770156fb4c3648da64f2a7d5bc21f/warehouse/static/js/warehouse/controllers/github_repo_stats_controller.js
https://github.com/pypi/warehouse/blob/8d8cc16f0e0770156fb4c3648da64f2a7d5bc21f/warehouse/static/js/warehouse/controllers/github_repo_info_controller.js
https://github.com/pypi/warehouse/blob/8d8cc16f0e0770156fb4c3648da64f2a7d5bc21f/warehouse/templates/includes/packaging/project-data.html

We derive the API URL here:
[...]

PR #12985, opened 2023-02-12
Split issues and prs #12971

HallvardMM:
This PR is related to the issue reported in #12971 about showcasing open issues and PRs as separate values. 

Major changes:
- Showcase data for Issues and PRs as separate values
- Makes two calls to GitHub in parallel instead of one

I created some tests, but they are very similar to what was already done.

I had issues visually verifying the changes locally as it was not displaying the GitHub data (this happened before making changes).
image


tests/unit/packaging/test_models.py
```diff
@@ -445,6 +445,45 @@ def test_github_repo_info_url(self, db_session, home_page, expected):
...
+                "https://www.github.com/pypi/warehouse.git",
+                "https://api.github.com/search/issues?q=repo:pypi/warehouse+type:issue+state:open&per_page=1",
+            ),
+            (
+                "https://www.github.com/pypi/warehouse.git/",
+                "https://api.github.com/search/issues?q=repo:pypi/warehouse+type:issue+state:open&per_page=1",
+            ),
+        ],
+    )
+    def test_github_open_issue_info_url(self, db_session, home_page, expected):
```
miketheman, line 483: Excellent work on adding these tests!

warehouse/packaging/models.py
```diff
@@ -596,6 +596,19 @@ def github_repo_info_url(self):
...
+    def github_open_issue_info_url(self):
+        for url in self.urls.values():
+            parsed = urlparse(url)
+            segments = parsed.path.strip("/").split("/")
+            if parsed.netloc in {"github.com", "www.github.com"} and len(segments) >= 2:
+                user_name, repo_name = segments[:2]
+                if user_name in GITHUB_RESERVED_NAMES:
+                    continue
+                if repo_name.endswith(".git"):
+                    repo_name = repo_name.removesuffix(".git")
```
miketheman, line 609: **thought:** ‏not a blocker, but there seems to be a large amount of overlap between this function and `github_repo_info_url()` - might be worth refactoring for reuse.

tests/frontend/github_repo_info_controller_test.js
```diff
@@ -41,13 +41,19 @@ const mountDom = async () => {
...
+        </a>
+      </li>
     </div>
   `;
   document.body.innerHTML = `
     <div id="github-repo-stats"
           data-controller="github-repo-stats"
           data-github-repo-stats-github-repo-info-outlet=".github-repo-info">
           data-github-repo-stats-url-value="https://api.github.com/repos/pypi/warehouse">
+          data-github-repo-stats-issueUrl-value="https://api.github.com/search/issues?q=repo:pypi/warehouse+type:issue+state:open&per_page=1">
```
miketheman, line 56: **suggestion:** ‏This URL prefix does not currently exist in the [Content Security Policy](https://github.com/pypi/warehouse/blob/8218d5d9e6f72d90bd34cc64ca2cf25aff87a471/warehouse/csp.py#L88) so it will be blocked by browsers. Please add it to the scope (and update associated tests).

warehouse/static/js/warehouse/controllers/github_repo_stats_controller.js
```diff
@@ -15,28 +15,45 @@ import { Controller } from "@hotwired/stimulus";
...
-          forks: data.forks_count,
-          forks_url: data.html_url + "/network/members",
+          issues_url: res[0].html_url + "/issues",
+          PRs_url: res[0].html_url + "/pulls",
+          followers: res[0].stargazers_count,
+          followers_url: res[0].html_url + "/stargazers",
+          forks: res[0].forks_count,
+          forks_url: res[0].html_url + "/network/members",
+          issues: res[1].total_count,
+          PRs: res[0].open_issues_count - res[1].total_count,
```
miketheman, line 52: **question:** ‏Now that both API calls are performed concurrently in the Promise, what happens if either of them return `null`? Will calls to `res[N].some_property` fail?

miketheman (changes requested):
Thanks for working on this!

I've added some suggestions, questions in the format of [Conventional Comments](https://conventionalcomments.org/). Let me know if I can help elaborate on anything further.

There's a few lint issues that need to be handled before this can go through, and whenever we change strings we also run `make translations` to update the translation source, so please add that when you're done with other changes.

warehouse/static/js/warehouse/controllers/github_repo_stats_controller.js
HallvardMM: Based on the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) if one fails the whole expression throws and we would be in the catch(). The catch() does nothing, which was the behavior prior to this PR. However, if there is interest in changing this it could be possible to use [promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) and handle the errors separately.

warehouse/csp.py
```diff
@@ -86,6 +86,7 @@ def includeme(config):
                 "connect-src": [
                     SELF,
                     "https://api.github.com/repos/",
+                    "https://api.github.com/search/issues",
```
HallvardMM, line 89: Is it correct to not include a trailing "/"?
When the URL to be called is on this form `https://api.github.com/search/issues?q=repo:pypi/warehouse+type:issue+state:open&per_page=1`

HallvardMM:
@miketheman I think the commit should address your comments 😄

warehouse/csp.py
miketheman: Yes, my understanding is that the CSP rules are evaluated as direct string matches, and only allow wildcards `*` - so this should be fine.

warehouse/static/js/warehouse/controllers/github_repo_stats_controller.js
miketheman: No further change needed, thanks for clarifying the behaviors.

miketheman (approved):

PR #12996, opened 2023-02-15
Migrate keywords data structure to Postgres ARRAY type

miketheman:
🚨 **Note**: Data migration contained within

Does not drop existing column, but removes references to it.

Once merged, no new Releases will populate their `keywords` column, only the `keywords_array` column will be filled.

### DIFF
--- a/warehouse/migrations/versions/677b8c232e17_add_release_keywords_array_column.py
+++ b/warehouse/migrations/versions/677b8c232e17_add_release_keywords_array_column.py
@@ -1,0 +1,54 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add Release.keywords_array column
+
+Revision ID: 677b8c232e17
+Revises: f93cf2d43974
+Create Date: 2023-02-14 19:32:12.553294
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "677b8c232e17"
+down_revision = "f93cf2d43974"
+
+
+def upgrade():
+    op.add_column(
+        "releases",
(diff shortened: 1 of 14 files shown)


warehouse/utils/release.py
```diff
@@ -0,0 +1,27 @@
...
+    Split keywords on commas and strip whitespace, remove empties.
+    Useful to cleanse user input prior to storing in Release.keywords_array.
+    """
+    try:
+        split_keywords = keyword_input.split(",")
+    except AttributeError:
+        return None
+
+    stripped_keywords = [keyword.strip() for keyword in split_keywords]
+    slimmed_keywords = [keyword for keyword in stripped_keywords if keyword]
```
miketheman, line 25: I thought about making this a nested comprehension to save a variable, but figured that might hamper readability so did this instead.

warehouse/migrations/versions/677b8c232e17_add_release_keywords_array_column.py
```diff
@@ -0,0 +1,54 @@
...
+        SET keywords_array = (
+            SELECT ARRAY(
+                SELECT TRIM(
+                    UNNEST(
+                        STRING_TO_ARRAY(keywords, ',')
+                    )
+                )
+            )
+        )
+        WHERE keywords IS NOT NULL AND keywords != ''
```
miketheman, line 48: This is the data migration I'm a little wary about - since it'll need to update somewhere around 4M records (depending on the `WHERE` clause in prod). Any timing estimates that could be run on TestPyPI DB would be helpful to gauge production impact.

dstufft:
I haven't read the entire PR yet, but you probably want this split over multiple deploys, currently what's going to happen is:

1. Column will be added, data migration will be run.
2. New deployment will occur with the new code.
3. Old deployment will be spun down, shutting down old code.

The important bit of that is that the old code will still be running after the data migration has completed, so anything uploaded in that window of time via the old code will have ``keywords`` populated, but not ``keywords_array``.

So I would refactor this into multiple PRs, that we can deploy in separate deployments:

1. Adds the `keywords_array` column, sets up the upload code to write *both* columns.
2. Does the data migration, guarding not to run against rows that (1) would have already written, sets up the code to only use the `keywords_array` column.
    - Optional: This could also drop the `keywords` column from the code, but not from the database, making SQLAlchemy completely ignore that column.
3. If the `keywords` column has been dropped from the code, drop it from the database as well.

warehouse/migrations/versions/677b8c232e17_add_release_keywords_array_column.py
miketheman: Perfect, thanks for the example.

miketheman:
> probably want this split over multiple deploys

Yep! Going to dismantle and rebuild, now that all the pieces are known.

PR #13001, opened 2023-02-15
feat: introduce Release.keywords_array

miketheman:
Meant to replace `Release.keywords`, add the column and populate with newly uploaded data.

Part 1 of a few - see #12996 for background.


warehouse/utils/release.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    try:
+        split_keywords = keyword_input.split(",")
+    except AttributeError:
+        return None
+
+    stripped_keywords = [keyword.strip() for keyword in split_keywords]
+    slimmed_keywords = [keyword for keyword in stripped_keywords if keyword]
+
+    return slimmed_keywords if slimmed_keywords else None
```
dstufft, line 27: Do we want to use `None` instead of `[]` for no keywords? Using an empty array would make it easier to differentiate between unset, and set but empty.

dstufft: Note: I don't have a strong preference, just raising the question.

miketheman: It's a fair question! I tried looking at it from whether or not we make that distinction today.
Do we care if someone sets an empty keywords in their metadata on upload, and do we need to reflect that back?

I was lucky enough to stumble upon an example through some create GitHub searches:

- Explicitly setting to an empty string: https://github.com/mailgun/expiringdict/blob/db04243ceb6d2d8d616abba21caa7484c78e4d7d/setup.py#L36
- Doesn't show up in generated metadata v2.1: https://inspector.pypi.io/project/expiringdict/1.2.2/packages/cb/84/a04c59324445f4bcc98dc05b39a1cd07c242dde643c1a3c21e4f7beaf2f2/expiringdict-1.2.2-py3-none-any.whl/expiringdict-1.2.2.dist-info/METADATA
- JSON api returns `"keywords: ""` for this - the same as any package with `keywords` not set: https://pypi.org/pypi/expiringdict/json

If you wouldn't mind running these two commands to show how that is currently stored ion the prod DB, we could see if the current behavior is to store an empty string or leave it null:

```sql
\pset null 'Ø'
select keywords from releases r join projects p on p.id = r.project_id where p.name = 'expiringdict';
```
The `pset` command should make a null visible for the psql session.

The outcome of that command could guide us to how it's being done today

ewdurbin:
```
warehouse=> \pset null 'Ø'
Null display is "Ø".
warehouse=> select keywords from releases r join projects p on p.id = r.project_id where p.name = 'expiringdict';
 keywords 
----------
 
 
 Ø
 Ø
 Ø
 Ø
 Ø
 
 
(9 rows)
```

miketheman: Thanks @ewdurbin ! That's helpful, as it shows that there's both nulls and empty strings. I guess I should have added the `r.created` to the select to see when these happened - since this project dates back to 2013 - predating warehouse itself.

Could you run this query instead?

```sql
\pset null 'Ø'
select keywords, r.created
from releases r join projects p on p.id = r.project_id
where p.name = 'expiringdict'
order by r.created;
```

Essentially, if there's a clear divide between the 2014-and-before releases and the 2017-and-later releases, then we'll have an answer on how to best represent empty keywords metadata.

firefly9966 (approved):

warehouse/utils/release.py
miketheman: Query results, updated to show specifically empty strings as well:

```sql
warehouse=> select to_json(keywords), r.created
from releases r join projects p on p.id = r.project_id
where p.name = 'expiringdict'
order by r.created;
 to_json |          created
---------+----------------------------
 Ø       | 2013-09-21 00:01:44
 Ø       | 2014-02-20 02:23:08.252373
 Ø       | 2014-03-06 00:10:32.327305
 Ø       | 2014-03-07 21:42:35.350158
 Ø       | 2014-08-14 00:43:48.253346
 ""      | 2017-09-11 10:46:48.526709
 ""      | 2019-11-03 07:34:03.558161
 ""      | 2020-06-02 21:43:48.035501
 ""      | 2022-06-21 09:12:28.652017
(9 rows)
```

So we can see that when the `keywords=""` was set circa 2017, the column is now an empty string, vs null.

Seems likely to me to create an empty array if a client submits an empty string for keywords vs null if it wasn't submitted at all.

warehouse/migrations/versions/208d494aac68_add_release_keywords_array_column.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def upgrade():
+    op.add_column(
+        "releases",
+        sa.Column(
+            "keywords_array",
+            postgresql.ARRAY(sa.String()),
+            nullable=True,
+            comment="Array of keywords. Null indicates no keywords were supplied by the uploader.",
```
di, line 36:
```suggestion
            comment=(
                "Array of keywords. Null indicates no keywords were supplied by "
                "the uploader."
            ),
```

warehouse/packaging/models.py
```diff
@@ -561,6 +561,10 @@ def __table_args__(cls):  # noqa
     license: Mapped[str | None]
     summary: Mapped[str | None]
     keywords: Mapped[str | None]
+    keywords_array: Mapped[list[str] | None] = mapped_column(
+        ARRAY(String),
+        comment="Array of keywords. Null indicates no keywords were supplied by the uploader.",
```
di, line 566:
```suggestion
        comment=(
            "Array of keywords. Null indicates no keywords were supplied by "
            "the uploader."
        ),
```

warehouse/utils/release.py
```diff
@@ -0,0 +1,19 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def split_and_strip_keywords(keyword_input: str) -> list[str]:
+    """
+    Split keywords on commas and strip whitespace, remove empties.
+    Useful to cleanse user input prior to storing in Release.keywords_array.
+    """
+    return [keyword.strip() for keyword in keyword_input.split(",") if keyword.strip()]
```
di, line 19:
```suggestion
    return [
        stripped
        for keyword in keyword_input.split(",")
        if (stripped := keyword.strip())
    ]
```

miketheman: Nice walrus usage - I often overlook that one.

warehouse/migrations/versions/208d494aac68_add_release_keywords_array_column.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add Release.keywords_array column
+
+Revision ID: 208d494aac68
+Revises: bb6943882aa9
```
miketheman, line 16:
```suggestion
Revises: fd06c4fe2f97
```

warehouse/migrations/versions/208d494aac68_add_release_keywords_array_column.py
```diff
@@ -0,0 +1,45 @@
...
+Create Date: 2024-08-02 19:02:01.760253
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "208d494aac68"
+down_revision = "bb6943882aa9"
```
miketheman, line 26:
```suggestion
down_revision = "fd06c4fe2f97"
```

PR #13018, opened 2023-02-17
warehouse/packaging: initial ProjectService

woodruffw:
~~WIP; I need to remove the current `utils` code and update the tests.~~

Replaces the `add_project` helper with a new service `ProjectService` with constituent functionality. Also promotes some other functionality in the `file_upload` handler into `ProjectService.create_project`, like the role handling (since it's directly in-line with project creation).

### DIFF
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -60,7 +60,6 @@ from warehouse.organizations.models import (
     TeamProjectRoleType,
     TeamRoleType,
 )
-from warehouse.packaging.interfaces import IProjectService
 from warehouse.packaging.models import (
     File,
     JournalEntry,
@@ -3893,11 +3892,6 @@ class TestManageOrganizationProjects:
         validate_project_name = pretend.call_recorder(lambda *a, **kw: True)
         monkeypatch.setattr(views, "validate_project_name", validate_project_name)
 
-        # project_service = pretend.stub(
-        #     create_project=pretend.call_recorder(lambda *a: None)
-        # )
-        # db_request.find_service = pretend.call_recorder(lambda *a: project_service)
-
         def add_organization_project(*args, **kwargs):
             OrganizationProjectFactory.create(
                 organization=organization, project=project
@@ -3922,10 +3916,6 @@ class TestManageOrganizationProjects:
         assert isinstance(result, HTTPSeeOther)
         assert result.headers["Location"] == db_request.path
         assert validate_project_name.calls == [pretend.call(project.name, db_request)]
-        # assert db_request.find_service.calls == [pretend.call(IProjectService)]
-        # assert project_service.create_project.calls == [
-        #     pretend.call(project.name, db_request.user)
-        # ]
         assert len(organization.projects) == 2
         assert send_organization_project_added_email.calls == [
             pretend.call(
--- a/tests/unit/manage/test_views.py
+++ b/tests/unit/manage/test_views.py
@@ -3866,20 +3866,14 @@ class TestManageOrganizationProjects:
         db_request.help_url = lambda *a, **kw: ""
 
         organization = OrganizationFactory.create()
-        OrganizationProjectFactory(
-            organization=organization, project=ProjectFactory.create()
-        )
-
-        project = ProjectFactory.create()
 
         OrganizationRoleFactory.create(
             organization=organization, user=db_request.user, role_name="Owner"
         )
-        RoleFactory.create(project=project, user=db_request.user, role_name="Owner")
 
         add_organization_project_obj = pretend.stub(
             add_existing_project=pretend.stub(data=False),
-            new_project_name=pretend.stub(data=project.name),
+            new_project_name=pretend.stub(data="fakepackage"),
             validate=lambda *a, **kw: True,
         )
         add_organization_project_cls = pretend.call_recorder(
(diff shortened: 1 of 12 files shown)


woodruffw:
NB: There are a whole bunch of other project utils/helpers that can and probably should go under `ProjectService`, but I'll limit it to just project creation in the interest of expediency.

woodruffw:
One last test failure here, in the organization project creation code:

```
    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; 
consider using a session.no_autoflush block if this flush is occurring prematurely)
E       (psycopg2.errors.UniqueViolation) duplicate key value violates unique constrain
t "projects_normalized_name_key"
E       DETAIL:  Key (normalized_name)=(mocwyidmwksm) already exists.
E       
E       [SQL: INSERT INTO projects (name, has_docs, upload_limit, total_size_limit) VAL
UES (%(name)s, %(has_docs)s, %(upload_limit)s, %(total_size_limit)s) RETURNING projects
.id]
E       [parameters: {'name': 'MocwyiDMWKsM', 'has_docs': None, 'upload_limit': None, '
total_size_limit': None}]
E       (Background on this error at: https://sqlalche.me/e/14/gkpj)

../lib/python3.11/site-packages/sqlalchemy/engine/default.py:736: IntegrityError
```

I think this is happening because the test in question (`test_add_organization_project_new_project`) creates a project to test with using `ProjectFactory.create()`, but the actual code under test assumes that a project hasn't been created yet. The test itself originally mocked out the project creation code, but the implementation implies that a project _should_ be created.

Making this test pass would be easy enough, but I wasn't sure what the actual intended behavior is here (I'm not very familiar with the organizations feature): are organizations allowed to create projects? Or was this accidentally stubbed out, and so nobody noticed that they can when they shouldn't be able to?

ewdurbin:
> Making this test pass would be easy enough, but I wasn't sure what the actual intended behavior is here (I'm not very familiar with the organizations feature): are organizations allowed to create projects? Or was this accidentally stubbed out, and so nobody noticed that they can when they shouldn't be able to?

Organization accounts are supposed to be able to [create new Projects via the management form if the name is valid](https://github.com/pypi/warehouse/blob/8675183ef856bb19f723adb1e5f14afb5b177dc0/warehouse/manage/views.py#L1909-L1916).

ewdurbin:
Yeah, on closer inspection it looks like the ProjectService.create_project might need to be refactored to either "opt out" of creating the associated Role, or accept an "owned_by" to have different behavior if a User or an Organization object is supposed to be the owner.

woodruffw:
Thanks! 

>  either "opt out" of creating the associated Role, or accept an "owned_by" to have different behavior if a User or an Organization object is supposed to be the owner.

I think opt-out makes sense for now -- it matches what the current organization code does (no owner roles for the newly created project).

As far as I can tell, there's no dedicated "owner organization" role, right? When a pre-existing project gets added to an org, it seems to remove the requesting user (an owner) as a role from the project, and doesn't add any additional roles.

ewdurbin:
> As far as I can tell, there's no dedicated "owner organization" role, right? When a pre-existing project gets added to an org, it seems to remove the requesting user (an owner) as a role from the project, and doesn't add any additional roles.

Correct, there's no role. There is a relationship object (OrganizationProject) which implicitly gives Users with OrganizationRoleType.Owner Owner level permissions on Projects owned by an Organization (similar to GitHub).

woodruffw:
Awesome! I've made the "opt-out" changes and tweaked the test to avoid pre-creating a project, although now I'm getting another error that doesn't make a ton of sense to me:

```
>       assert len(organization.projects) == 2                                         
E       AssertionError: assert 0 == 2                                                  
E        +  where 0 = len([])                                                          
E        +    where [] = Organization(name='BpCIFcyTxLUL').projects                    
                                                                                       
tests/unit/manage/test_views.py:3904: AssertionError   
```

(In my reading, `test_add_organization_project_new_project` should now be creating exactly 1 project, rather than the 2 it did before. But it seems to create none instead.)

woodruffw:
The project under the org is definitely created, per this assert: https://github.com/pypi/warehouse/pull/13018/files#diff-f18da9d03b157bfb5c0bdd8572661923f3b0f0cd79f58809a010787e68887fa2R3900-R3901

`organization_service.add_organization_project` is also definitely called, since the tests observe the side effects of functions called directly after it.

woodruffw:
The plot thickens:

```
        # The project was created.
        project = (
            db_request.db.query(Project).filter_by(name="fakepackage").one_or_none()
        )
        assert project is not None
        assert project.organization == organization

        assert result.headers["Location"] == db_request.path
        assert validate_project_name.calls == [pretend.call("fakepackage", db_request)]
        assert len(organization.projects) == 1
```

`assert project.organization == organization` passes, while `len(organization.projects) == 1` fails. I'm guessing this is an artifact of the test setup/faker models?

tests/unit/manage/test_views.py
```diff
@@ -3916,17 +3897,22 @@ def add_organization_project(*args, **kwargs):
...
+            db_request.db.query(Project).filter_by(name="fakepackage").one_or_none()
+        )
+        assert project is not None
+        assert project.organization == organization
+
         assert isinstance(result, HTTPSeeOther)
         assert result.headers["Location"] == db_request.path
-        assert validate_project_name.calls == [pretend.call(project.name, db_request)]
-        assert add_project.calls == [pretend.call(project.name, db_request)]
-        assert len(organization.projects) == 2
```
woodruffw, line 3923: Flagging for review: the asserts I've added should be equivalent to the ones I removed, which stopped working (probably because of stubbing/mocking, but I didn't fully root-cause it).

di (approved):
Perfect!

PR #13020, opened 2023-02-18
chore(dev): cache generated artifacts between runs

### LINKED ISSUE #13019: Use mypy's SQLite cache option?
Typechecking Warehouse with `make lint` can be pretty slow; `mypy`'s `sqlite-cache = true` setting would probably slow it up a bit.

I can do this, just creating an issue first for visibility/in case there's a known reason not to do this 🙂

miketheman:
Closes #13019

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,7 @@ volumes:
   sponsorlogos:
   policies:
   vault:
+  caches:
 
 services:
   vault:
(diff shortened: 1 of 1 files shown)


ewdurbin:
Non blocker, but is this cacheable in CI as well?

ewdurbin (approved):

miketheman:
> is this cacheable in CI as well?

It's entirely possible, but not through this mechanism, as our CI doesn't use Docker Compose.

Considering the CI runs slowest step is still `tests` due to parallelism, is it worth it?

woodruffw (approved):
Nice, thanks @miketheman!

ewdurbin:
> Considering the CI runs slowest step is still `tests` due to parallelism, is it worth it?

From a pure stewardship perspective, I think any reduction in execution time are worth seeing through!

miketheman:
> 

Fair enough - started https://github.com/pypi/warehouse/pull/13021

di:
Just noticed that this allows us to use pytest's `--last-failed` and `--failed-first` options! https://docs.pytest.org/en/7.1.x/how-to/cache.html

miketheman:
Yep, that was one of my use cases, sorry for not calling that out louder!

PR #13022, opened 2023-02-19
fix: prevent excessive 403 calls

miketheman:
When a project's details page is loaded, the CSI mechanism will request the "Manage project" (aka edit) button.

However, since most folks are unlikely to be logged in, most of these requests result in a 403 HTTP status code response.

Wrapping with an `if` statement, we should cut down the volume of unnecessary requests to this endpoint significantly.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -561,8 +561,8 @@ msgstr ""
 #: warehouse/templates/manage/project/release.html:194
 #: warehouse/templates/manage/project/releases.html:140
 #: warehouse/templates/manage/project/releases.html:179
-#: warehouse/templates/packaging/detail.html:361
-#: warehouse/templates/packaging/detail.html:380
+#: warehouse/templates/packaging/detail.html:363
+#: warehouse/templates/packaging/detail.html:382
 #: warehouse/templates/pages/classifiers.html:25
 #: warehouse/templates/pages/help.html:20
 #: warehouse/templates/pages/help.html:216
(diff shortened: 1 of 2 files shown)


miketheman:
From a numbers perspective using Datadog APM, we can see that in a 15 minute window there were ~5,500 403 vs 12 200 responses to this endpoint ([private access link](https://app.datadoghq.com/apm/traces?query=service%3Awarehouse%20env%3Aprod%20operation_name%3Apyramid.request%20%40_top_level%3A1%20resource_name%3A%22GET%20includes.edit-project-button%22&agg_m=count&agg_q=%40http.status_code&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=false&index=trace-search&messageDisplay=inline&sort=desc&sort_m=&sort_t=&spanType=service-entry&step=auto&top_n=10&top_o=top&viz=query_table&x_missing=true&start=1676817667187&end=1676818567187&paused=false)).

ewdurbin (approved):
smart!!!

ewdurbin:
Hmmm, on second thought I’m trying to think how this interacts with caching.

ewdurbin:
The reason we use a client side include for this is to make it so we can cache these pages independently of logged in status, so I’m not sure if this will work as expected.

ewdurbin:
Yeah. I don’t think we can solve this in the template alone. We’d need to somehow short circuit in the browser.

miketheman:
Oh, that's more interesting. Hadn't thought of that, thanks for the pointer. 🤔  
I'll go back to the drawing table.

ewdurbin:
I’m not sure if our JS knows wether a user is logged in or not, but if it did we could add a some bits to mark certain client side includes as “requiring login” and skip loading them in https://github.com/pypi/warehouse/blob/main/warehouse/static/js/warehouse/utils/html-include.js

miketheman:
Indeed, pretty much all our JS is nonspecific to user behavior.

In thinking about how the html-includes works today, we could look at the DOM for the non-Login link element as a proxy for a logged in session, and add another class to the csi tag in html to operate as a conditional, so that the html remains static.

miketheman:
Using the non-Login DOM element as a conditional will still fire 403s for logged in users who are looking at other projects, but the overall volume would come down. (That's the same as the current proposed implementation)

PR #13056, opened 2023-02-22
refactor: move permissions evaluation into template

miketheman:
When browsers load the page, they fire 403 responses when user is either not logged in, or doesn't have permissions to manage the project.

Instead, we allow the permissions on the view, and return an empty response if now allowed.

See #13022 for prior approach.


miketheman:
Note that this doesn't cut down the volume of the calls, only the volume of 403 responses.

di (approved):

miketheman:
Screenshot 2023-02-27 at 11 32 22 AM
Now 403s can be more valuable!

PR #13059, opened 2023-02-22
refactor(ci): extract node to own workflow

miketheman:
To simplify the configuration, extract the node-related steps to their own workflow file.

As the node jobs do not require platform dependencies, service containers like postgres or stripe, we can speed up those steps a little by splitting these apart.


miketheman:
Timings:

Before, the [combined CI workflow](https://github.com/pypi/warehouse/actions/runs/4246609174/usage) "costs" ~26  GitHub Actions minutes

Now with the split, Python ~16 min + Node ~2 min = ~18 minutes, saving 8 minutes of execution time overall

Example runs:
(Python) CI: https://github.com/pypi/warehouse/actions/runs/4246623547/usage
Node CI: https://github.com/pypi/warehouse/actions/runs/4246623544/usage

ewdurbin (approved):
Nice!

ewdurbin:
Doing a think on how (if at all) we need to adjust branch protection before/after merge.

di:
We could merge https://github.com/pypi/warehouse/pull/12263 which makes it so we don't have to adjust required checks when we add/remove them.

miketheman:
> Doing a think on how (if at all) we need to adjust branch protection before/after merge.

I'm _hoping_ that nothing has to change, since the **job** names were preserved

di:
> I'm hoping that nothing has to change, since the job names were preserved

Aha, yes this does seem to be the case.

miketheman:
> We could merge #12263 which makes it so we don't have to adjust required checks when we add/remove them.

I forgot about that one - I wonder how it works across multiple workflow files?

ewdurbin:
> > We could merge #12263 which makes it so we don't have to adjust required checks when we add/remove them.
> 
> I forgot about that one - I wonder how it works across multiple workflow files?

I remember squinting at that and getting confused even before we had multiple workflows.

di (approved):

miketheman:
hold the phone - I missed something

miketheman:
Due to a whoopsie, I had removed the actual test sommand, restored now.

Numbers are still better - https://github.com/pypi/warehouse/actions/runs/4246789167/usage

Python down to 21 minutes, so 4 minutes savings, not 8.

webknjaz:
> I forgot about that one - I wonder how it works across multiple workflow files?

@miketheman it is possible to make it work through some refactoring, like in this example PR: https://github.com/pypa/build/pull/618.

Issue #12965, opened 2023-02-08
Closed beta for OIDC publishing


woodruffw:
Once #12915 is merged, we should be in a good place to start a closed beta for OIDC publishing.

I'll use this issue to track adding the feature flags/ACL checks needed for that.

woodruffw:
Also needed for a closed beta: admin app changes to allow admins to manage pending/normal OIDC providers.

woodruffw:
CC @cedricvanrompay-datadog and @webknjaz as two early testers.

CC @alex as well, since I know you're interested 😉

nitzmahone:
Sign me up!

di:
In order to cut down on noise here, anyone interested, please fill out https://forms.gle/XUsRT8KTKy66TuUp7 and we'll be in touch once this is ready.

PR #13073, opened 2023-02-23
warehouse: OIDC beta flag/controls

woodruffw:
~WIP.~

See #12965.


woodruffw:
Just to summarize: this adds a new principal (`group:oidc_beta`).

Users are added to the `group:oidc_beta` principal under one of two conditions:

1. They're a superuser (`User.is_superuser`)
2. They've been explicitly added to the OIDC beta (`User.has_oidc_beta_access`)

Anybody in the `group:oidc_beta` principal has the new (temporary) OIDC permissions:

* `manage:user:oidc`: view and manage "pending" OIDC publishers;
* `manage:project:oidc`: view project-level OIDC publishers;
* `manage:project:oidc:modify`: manage project-level OIDC publishers;

The last two permissions are separate so that users who *aren't* in the OIDC beta but have access to projects with OIDC publishers can see but not modify those publishers.

woodruffw:
Just remembered another thing I need to do here: the admin panel should allow Warehouse's admins to enable/disable OIDC beta status for a user.

warehouse/config.py
```diff
@@ -63,6 +63,13 @@ class RootFactory:
         (Allow, "group:psf_staff", "psf_staff"),
         (Allow, "group:with_admin_dashboard_access", "admin_dashboard_access"),
         (Allow, Authenticated, "manage:user"),
+        # OIDC beta users can see their user-level OIDC settings.
+        # XXX: Can be removed once OIDC is removed from beta.
+        (
+            Allow,
+            "group:oidc_beta",
+            "manage:user:oidc",
+        ),
```
woodruffw, line 72: Flagging: this isn't 100% right, since it gives everybody in the OIDC beta the ability to modify every other OIDC beta users' OIDC providers. So maybe a principal is the wrong abstraction here.

woodruffw:
This should be good to go -- I've added the `User` feature flag plus an admin control for it, along with checks to each OIDC view (+ menu renderers).

tests/unit/manage/test_views.py
```diff
@@ -9538,8 +9538,11 @@ def find_service(iface, name=None, context=None):
             view._check_ratelimits()
 
     def test_manage_project_oidc_publishers(self, monkeypatch):
-        project = pretend.stub()
+        project = pretend.stub(oidc_providers=[])
```
di, line 9541:
```suggestion
        project = pretend.stub(oidc_publishers=[])
```

tests/unit/manage/test_views.py
```diff
@@ -9572,8 +9575,11 @@ def test_manage_project_oidc_publishers(self, monkeypatch):
         ]
 
     def test_manage_project_oidc_publishers_admin_disabled(self, monkeypatch):
-        project = pretend.stub()
+        project = pretend.stub(oidc_providers=[])
```
di, line 9578:
```suggestion
        project = pretend.stub(oidc_publishers=[])
```

tests/unit/manage/test_views.py
```diff
@@ -9628,6 +9634,21 @@ def test_manage_project_oidc_publishers_oidc_not_enabled(self):
         with pytest.raises(HTTPNotFound):
             view.manage_project_oidc_publishers()
 
+    def test_manage_project_oidc_publishers_not_in_beta(self):
+        project = pretend.stub(oidc_providers=[])
```
di, line 9638:
```suggestion
        project = pretend.stub(oidc_publishers=[])
```

warehouse/manage/views.py
```diff
@@ -3073,6 +3073,9 @@ def manage_project_oidc_publishers(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
+        if not self.request.user.in_oidc_beta and not self.project.oidc_providers:
```
di, line 3076:
```suggestion
        if not self.request.user.in_oidc_beta and not self.project.oidc_publishers:
```

di, line 3076: We need to make sure that in the case where the user isn't in the beta, but there are publishers, we aren't showing the "Add a new publisher" form (or that it's displayed as disabled at least). Same for the "Remove" button for existing publisher(s).

warehouse/accounts/models.py
```diff
@@ -96,6 +96,9 @@ class User(SitemapMixin, HasEvents, db.Model):
     totp_secret = Column(LargeBinary(length=20), nullable=True)
     last_totp_value = Column(String, nullable=True)
 
+    # XXX: Can be removed once OIDC is removed from beta.
+    has_oidc_beta_access = Column(Boolean, nullable=False, server_default=sql.false())
```
ewdurbin, line 100: Bleh, I'm -1 on a column directly on the User model, but won't block the PR on this.

The aspiration was a more extensible AdminFlags or similar to manage this kind of thing, but obviously that hasn't come to fruition yet.

warehouse/packaging/models.py
```diff
@@ -296,10 +296,12 @@ def __acl__(self):
             permissions |= {(role.user_id, "Administer") for role in query.all()}
 
         for user_id, permission_name in sorted(permissions, key=lambda x: (x[1], x[0])):
+            permissions = ["upload"]
```
ewdurbin, line 299: just a nit, but this could lead to some trouble if another if or for block or whatever is added after this down the line where we're replacing the value of the `permissions` set that we're looping over.

warehouse/accounts/models.py
di: AdminFlags are just global flags that can be set by admins -- they don't allow us to set per-user flags (unless we want to overhaul them, which is probably out of scope here).

We tried to do this with ACLs but given how this currently works with the User model, it was a bit more janky than this, which will be temporary.

warehouse/manage/views.py
woodruffw: Good catch! Yeah, those forms will currently 403 on submission for those users, but I'll explicitly disable them with CSS as well. That way only a motivated user would hit the 403.

warehouse/packaging/models.py
woodruffw: Yeah, good point -- this is a vestigial change from the original ACL approach that we've discarded, so I can drop it entirely.

warehouse/templates/manage/project/publishing.html
```diff
@@ -115,9 +116,10 @@ <h3>GitHub</h3>
         </div>
       </div>
       <div>
-        <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary">
+        <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary" {{ "" if request.user.in_oidc_beta else "disabled" }}>
```
di, line 119:
```suggestion
        <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary">
```
Since we won't be rendering this if the user isn't in the beta...

warehouse/packaging/models.py
di: This still needs an update.

di (commented):
LGTM aside from two tweaks

di (approved):

PR #13076, opened 2023-02-23
Don't specify webauthn transports

### LINKED ISSUE #13062: Webauthn allows me to register my phone as security key but not login with it
**Describe the bug**
Webauthn allows me to register my phone as security key but not login with it

**Expected behavior**
After closing the native is security key prompt, A chrome prompt like this should pop up and allows me to select my phone to use as a security key
ss

**To Reproduce**
Add a android phone as a security key by visiting your profile and clicking add security key and follow the expected behavior
Then logout and try to login with the same expected behaviour

**My Platform**
Windows 10 and chrome version Version 110.0.5481.177 (Official Build) (64-bit)

**Additional context**
None

di:
Fixes #13062.

Since we're not restricting the transport options at registration time, or storing the transport used during registration, we don't actually know what transports should be used at assertion time. Instead, we should just omit this restriction and allow all transport types.

Per https://github.com/duo-labs/py_webauthn/blob/c3c6c21f76937b08aed089897fe595f8b9232cdf/webauthn/helpers/structs.py#L271 these are optional.

### DIFF
--- a/warehouse/utils/webauthn.py
+++ b/warehouse/utils/webauthn.py
@@ -25,7 +25,6 @@ from webauthn.helpers.structs import (
     AttestationConveyancePreference,
     AuthenticationCredential,
     AuthenticatorSelectionCriteria,
-    AuthenticatorTransport,
     PublicKeyCredentialDescriptor,
     RegistrationCredential,
     UserVerificationRequirement,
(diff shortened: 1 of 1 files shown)


di:
This will present users with _all_ webauthn transports their browser supports, but if they try to use one they haven't registered previously, they'll see something like this dialog:

image

woodruffw (approved):
LGTM. I went through the blame and saw that this restricted set of transports was introduced with the `webauthn` bump in #10482, just in case we need to debug later :slightly_smiling_face:

ewdurbin:
that transport kwarg was just copy pasta from the library docs during the last major upgrade, and doesn’t serve a specific purpose as far as I recall. 👍

ewdurbin (approved):

PR #13097, opened 2023-02-27
configurable ratelimiting for project creation by user/ip

ewdurbin:
This gives us the ability to rate limit *new* project creation.

In general project creation is far less infrequent than uploads, so I think we could tune down the defaults significantly.

example output from twine:

```
ERROR    HTTPError: 429 Too Many Requests from http://localhost/legacy/                                                                 
         Too many new projects created    
```

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -36,6 +36,7 @@ from pyramid.httpexceptions import (
     HTTPForbidden,
     HTTPGone,
     HTTPPermanentRedirect,
+    HTTPTooManyRequests,
 )
 from pyramid.response import Response
 from pyramid.view import view_config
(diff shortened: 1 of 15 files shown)


warehouse/forklift/legacy.py
```diff
@@ -909,7 +911,11 @@ def file_upload(request):
             raise _exc_with_message(exc.__class__, exc.detail) from None
 
         project_service = request.find_service(IProjectService)
-        project = project_service.create_project(form.name.data, request.user)
+        try:
+            project = project_service.create_project(form.name.data, request.user)
+        except RateLimiterException:
+            msg = "Too many new projects created"
+            raise _exc_with_message(HTTPTooManyRequests, msg)
```
di, line 918: We probably need to do something similar in the other places where `create_project` is used.

di: That works too!

ewdurbin: see https://github.com/pypi/warehouse/pull/13097/commits/42693eb63f627611e59970886f345d7669f95b47

di (approved):

PR #13109, opened 2023-02-28
feat(ci): exclude any pulls that are `blocked`

miketheman:
We currently block dependencies if they fail checks, but we also occasionally apply a `blocked` label to mark a passing PR for exclusion.

### DIFF
--- a/.github/workflows/combine-prs.yml
+++ b/.github/workflows/combine-prs.yml
@@ -71,7 +71,7 @@ jobs:
                     }
                   }
                 }
-                if(${{ github.event.inputs.mustBeUnblocked }}) {
+                if(${{ github.event.inputs.mustBeUnblocked && statusOK }}) {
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-prs.yml
```diff
@@ -66,6 +71,24 @@ jobs:
...
+                    for (const label of labels) {
+                      const label_name = label['name'];
+                      console.log('Validating label: ' + label_name);
+                      if(label_name == 'blocked') {
+                        console.log('Discarding ' + branch + ' with blocked label on PR');
+                        statusOK = false;
+                      }
+                    }
+                  }
+                }
```
di, line 91: Might be good to shortcut this if `statusOK` is already false, so we don't consume too much API quota on a check that doesn't matter.

di (approved):

ewdurbin (approved):

PR #13110, opened 2023-02-28
Add some checkboxes to the "Delete Project" modal

di:
With any luck, this will slow down users who usually blow by the warning text here and then ask us to un-delete.

image

### DIFF
--- a/warehouse/static/sass/warehouse.scss
+++ b/warehouse/static/sass/warehouse.scss
@@ -344,3 +344,8 @@ time {
   clip: rect(0, 0, 0, 0);
   border: 0;
 }
+
+// Class for unstyled lists
+.no-bullets {
+  list-style-type: none;
+}
--- a/warehouse/templates/manage/project/settings.html
+++ b/warehouse/templates/manage/project/settings.html
@@ -234,7 +234,7 @@
   <hr>
   {% endif %}
 
-  <div class="callout-block callout-block--danger">
+  <div class="callout-block callout-block--danger" data-controller="delete-confirm">
     <h3>{% trans %}Delete project{% endtrans %}</h3>
     <p>
       <i class="fa fa-exclamation-triangle" aria-hidden="true"><span class="sr-only">{% trans %}Warning{% endtrans %}</span></i>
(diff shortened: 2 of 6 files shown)


warehouse/static/sass/warehouse.scss
```diff
@@ -344,3 +344,8 @@ time {
   clip: rect(0, 0, 0, 0);
   border: 0;
 }
+
+// Class for unstyled lists
+.no-bullets {
+  list-style-type: none;
+}
```
miketheman, line 351: Would `unstyled` work instead of adding a new class?
https://github.com/pypi/warehouse/blob/097fb304b9b9653210a4501c806f6eb88f59df31/warehouse/static/sass/warehouse.scss#L143-L147

miketheman (commented):
**suggestion:** ‏add a test case, similar to https://github.com/pypi/warehouse/blob/097fb304b9b9653210a4501c806f6eb88f59df31/tests/frontend/confirm_controller_test.js but likely simpler, but with multiple checkboxes

I tested out the code, and it behaves the way I'd expect it to - selecting and unselecting changes the disabled state correctly.

warehouse/static/sass/warehouse.scss
di: Unfortunately it doesn't look right because it lacks padding/margins.

miketheman: Fair enough. In that case, should it be moved to either https://github.com/pypi/warehouse/blob/main/warehouse/static/sass/base/_lists.scss or should it be located in a block?

warehouse/templates/manage/project/settings.html
```diff
@@ -265,7 +265,13 @@ <h3>{% trans %}Delete project{% endtrans %}</h3>
         </small>
       </li>
     </ul>
+    <ul class="no-bullets">
+      <li><input type="checkbox" data-action="input->delete-confirm#check" data-delete-confirm-target="input"> I understand that I will not be able to re-upload any deleted versions.</li>
```
ewdurbin, line 269: Maybe link to https://pypi.org/help/#file-name-reuse

warehouse/templates/manage/project/settings.html
```diff
@@ -265,7 +265,13 @@ <h3>{% trans %}Delete project{% endtrans %}</h3>
...
+      <li><input type="checkbox" data-action="input->delete-confirm#check" data-delete-confirm-target="input"> I understand that I am permanently deleting all releases for this project.</li>
+      <li><input type="checkbox" data-action="input->delete-confirm#check" data-delete-confirm-target="input"> I understand that I am releasing this project name for use by any other PyPI user.</li>
```
ewdurbin, line 271: Maybe add a "I understand that I may not be able to re-register the project name under some circumstances" and link to https://pypi.org/help/#project-name

warehouse/templates/manage/project/settings.html
```diff
@@ -265,7 +265,13 @@ <h3>{% trans %}Delete project{% endtrans %}</h3>
...
+      <li><input type="checkbox" data-action="input->delete-confirm#check" data-delete-confirm-target="input"> I understand that I cannot undo this.</li>
```
ewdurbin, line 272: Maybe add a "I understand that the PyPI administrators cannot undo this".

ewdurbin (approved):

di:
Seven checkboxes feels right:

image

ewdurbin (approved):

PR #13126, opened 2023-03-03
chore(dev): update references for compose v2

miketheman:
The command has been updated, and should behave the same way regardless of whether the developer has "V2" checked in Docker Desktop.

Refs: https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2


ewdurbin (approved):
I think we tried this a few years back, but I think we're well past time to see it through! I'd rather update docs/help text than keep dragging feet.

ewdurbin:
https://warehouse.pypa.io/development/getting-started.html#make-initdb-is-slow-or-appears-to-make-no-progress might need reviewed.

miketheman:
> [warehouse.pypa.io/development/getting-started.html#make-initdb-is-slow-or-appears-to-make-no-progress](https://warehouse.pypa.io/development/getting-started.html#make-initdb-is-slow-or-appears-to-make-no-progress) might need reviewed.

Nice catch, removed!

PR #13127, opened 2023-03-03
refactor(dev): include missing files

miketheman:
When modifying `pyproject.toml` the changes are not included unless a full rebuild is run.

I could have added another entry for the specific file, however I figured it was better to take advantage of anonymous volumes and be explicit about the behavior we want to prevent instead.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,29 +82,19 @@ services:
     command: gunicorn --reload -b 0.0.0.0:8000 warehouse.wsgi:application
     env_file: dev/environment
     volumes:
-      # We specify all of these directories instead of just . because we want to
-      # avoid having ./node_modules from the host OS being shared with the docker
-      # container, and since there's no way to exclude a directory, the only way
-      # to make this work is to share multiple, smaller directories. These cover
-      # the important things that we want to share, but changes to requirements
-      # or any file not in these directories will require a rebuild.
       # The :z option fixes permission issues with SELinux by setting a
       # permissive security context.
-      - ./dev:/opt/warehouse/src/dev:z
-      - ./docs:/opt/warehouse/src/docs:z
-      - ./warehouse:/opt/warehouse/src/warehouse:z
-      - ./tests:/opt/warehouse/src/tests:z
-      - ./htmlcov:/opt/warehouse/src/htmlcov:z
-      - .coveragerc:/opt/warehouse/src/.coveragerc:z
+      - .:/opt/warehouse/src:z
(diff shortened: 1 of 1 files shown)


ewdurbin:
Hmmm, why not just add

```
- ${PWD}/pyproject.toml:/opt/warehouse/pyproject.toml:z
```

to the list?

ewdurbin:
I think that comment is _fair_ but also I think I'd rather be explicit about what we _do_ put in rather than what we don't.

ewdurbin:
Yeah, I'd suggest updating it to include pyproject.toml explicitly (and gunicorn.conf.py if you wouldn't mind).

Then I'd just do away with the big comment, or replace it with something that states we map files in explicitly.

miketheman:
> I think that comment is _fair_ but also I think I'd rather be explicit about what we _do_ put in rather than what we don't.

Okay, will refactor.

> and gunicorn.conf.py if you wouldn't mind

The `gunicorn-*conf.py` files are only ever referenced in the `Procfile` and used for prod as far as I can tell - why would we include them in the dev setup?

ewdurbin:
Ah, sorry its for linter.

ewdurbin:
As is now, as you change them the linter keeps saying "everything is fine!" or "you suck at syntax" when it may not be, while black and isort can't find them either :)

miketheman:
Updated!

docker-compose.yml
```diff
@@ -82,19 +82,16 @@ services:
...
+      # We specify paths explicitly to only add what we need, and to avoid
+      # adding the entire repository.
       # The :z option fixes permission issues with SELinux by setting a
       # permissive security context.
       - ./dev:/opt/warehouse/src/dev:z
       - ./docs:/opt/warehouse/src/docs:z
       - ./warehouse:/opt/warehouse/src/warehouse:z
       - ./tests:/opt/warehouse/src/tests:z
       - ./htmlcov:/opt/warehouse/src/htmlcov:z
+      - ./pyproject.toml:/opt/warehouse/src/pyproject.toml:z
```
ewdurbin, line 94: Did you confirm that this works? I've seen issues when the path specified isn't absolute Ala `${PWD}/pyproject.toml`

miketheman: Yes - confirmed - the leading `./` supplies the path (see the next entry for `.coveragerc`

ewdurbin (approved):

docker-compose.yml
ewdurbin: cool, thanks. that might just be old memories.

PR #13148, opened 2023-03-07
OIDC: Fix JWT decode key type

woodruffw:
We've been passing in the `PyJWK` directly, when `pyjwt.decode(...)` doesn't support that -- it only supports PEM-encoded `str` or `bytes` objects, *or* a pre-constructed `cryptography` key type (like `_RSAPublicKey` from the OpenSSL backend).

The latter isn't included in the public type hints, but it's included in the library's official documentation.

xref https://github.com/jpadilla/pyjwt/issues/864


woodruffw:
This is a little annoying to test since it's entirely in third-party code, but I _could_ add a backstop test for it (e.g. ensuring we only get an invalid signature, never a `TypeError`, when we pass a `cryptography` key type into `decode`). Thoughts @di?

di (approved):

PR #13153, opened 2023-03-07
tests, warehouse: drop Macaroon desc length DB constraint

woodruffw:
This is still enforced at the form validation layer, but isn't relevant for Macaroons we create explicitly (e.g. via OIDC).

### DIFF
--- a/warehouse/migrations/versions/eb736cb3236d_drop_macaroon_description_limit.py
+++ b/warehouse/migrations/versions/eb736cb3236d_drop_macaroon_description_limit.py
@@ -1,0 +1,45 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


di (approved):

warehouse/migrations/versions/eb736cb3236d_drop_macaroon_description_limit.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def downgrade():
+    op.alter_column(
+        "macaroons",
+        "description",
+        existing_type=sa.String(),
+        type_=sa.VARCHAR(length=100),
+        existing_nullable=False,
+    )
```
woodruffw, line 45: NB: This downgrade probably isn't sound, unless it does some silent truncation. Maybe time for another Order 227?

di: Yep, I think so.

di (approved):

PR #13155, opened 2023-03-07
migrations/env: configure `compare_type`

woodruffw:
This should help `alembic` autogenerate more migrations in the future.


di:
Hmm, it appears that many of these enums already had names, I don't see a reason to change them from what they already were, so we should adjust the added names accordingly.

di:
The migration for `sa.TEXT` -> `sa.ENUM` changes are not correctly auto-generated due to https://github.com/sqlalchemy/alembic/issues/278. Given that, I'm going to close this for now, as this works fine as-is.

PR #13163, opened 2023-03-08
security_policy: handle OIDC idents in 2FA check

woodruffw:
WIP.


warehouse/accounts/security_policy.py
```diff
@@ -251,13 +252,31 @@ def permits(self, context, principals, permission):
...
+            elif isinstance(request.identity, OIDCPublisher):
+                # Otherwise, if it's an OIDC-identified request, then we can
+                # ask whether the context (a project) has users who all
+                # have 2FA enabled.
+                # NOTE: TwoFactorRequireable applies to models other than projects,
+                # but we can assume that the context is a project here because of
+                # our subpolicy: OIDC identities only appear with macaroons,
+                # and macaroons are only permitted on the `upload` permission
+                # with a project context.
+                has_two_factor = all(user.has_two_factor for user in context.users)
```
di, line 271: Won't this result in publishing failing for the entire project if one user removes 2FA? I think that would be an unexpected outcome.

Given that a Publisher doesn't represent a specific user, I _think_ I'm OK skipping the check for 2FA in this case, assuming we are appropriately limiting OIDC Publisher creation for 2FA-mandated projects to users with 2FA only.

di (approved):

Issue #5867, opened 2019-05-16
2FA: "trust this device for 30 days" option


brainwane:
**What's the problem this feature will solve?**
The multi-factor auth makes logging in a little less convenient, and the fact of a previous login to a user's account from the same device makes it more likely that the user auth is legit.

**Describe the solution you'd like**
Much as other 2FA implementations do, we could add a "trust this device for 30 days" (or some similar period) checkbox during the 2FA check (during login).

**Additional context**
Hat-tip to @ericholscher for asking for this ("don't bother me on this computer for 30 days") during the PyCon NA 2019 sprints.

ericholscher:
Thanks for writing this up @brainwane! 💯 🎉

trishankatdatadog:
@brainwane This is a great idea, I don't see any issue from a security point of view, many companies do it (including Amazon), so the major security issues will really be on the server-side, on how we handle sessions in Warehouse, but that's it!

di:
Some additional discussion is in https://github.com/pypa/warehouse/issues/8033

callmecampos:
There are a few ways to ensure persistence of session information, and they mostly depend on whether we care about the longevity of these sessions enough that we need to store them outside Redis.

If we don't care too much about the longevity of these sessions, all we need to do is extend the lifetime of the session cookie if the user logs in and selects the "Keep me signed in" checkbox (which can just be added to the [login.html](https://github.com/pypa/warehouse/blob/master/warehouse/templates/accounts/login.html) template). Currently session information is stored in Redis for max 12 hour sessions (see [warehouse/sessions.py](https://github.com/pypa/warehouse/blob/master/warehouse/sessions.py?rgh-link-date=2020-06-19T00%3A56%3A49Z#L201)).

If we want to ensure data persistence no matter what, it seems we'd have to use Postgres, since I don't think we can (or even want to) [configure Redis to have data persistence](https://redis.io/topics/persistence). One way would be to just store relevant session information in a Postgres database in addition to Redis, and if Redis doesn't have the session information (whether Redis lost the data or if the session actually expired), we can double check in Postgres. However, this seems redundant and I imagine there's a better way to correlate session information with Redis with information in Postgres.

@ewdurbin wondering if you have an opinion on how to best go about this?

ristomcgehee:
I think I'd like to work on this. Here's the approach I would take. Feedback is welcome.

[...]

PR #13166, opened 2023-03-09
Remember device for 30 days - Part 1

ristomcgehee:
I am breaking this feature up into 2 PRs to make reviewing easier. This PR adds the new db table and user_service functions for saving devices. Part 2 of this feature can be found in https://github.com/ristomcgehee/warehouse/pull/2 in case you want to see how this code is used, but you shouldn't need to review Part 2 to review this PR. This PR can be merged to main independently of Part 2.

Part of https://github.com/pypi/warehouse/issues/5867


warehouse/accounts/services.py
```diff
@@ -621,6 +626,79 @@ def get_password_timestamp(self, user_id):
...
+        Generates a unique device id for the given user.
+        """
+        attempts_max = 1000
+        attempts = 0
+        while attempts < attempts_max:
+            device_id = secrets.token_urlsafe(DEVICE_ID_BYTES)
+            if not any(device.device_id == device_id for device in user.user_devices):
+                return device_id
+            attempts += 1
+        raise RuntimeError("Failed to generate device id")
```
di, line 700: Can you explain why we need a device ID _and_ a device secret? My first thought is that it should be sufficient to generate a random token, store it in a cookie, and store the hash of that token along with a reference to the user in the database.

ewdurbin: We could even just model it as

```
class UserDevice(db.Model):
     __tablename__ = "user_devices"

     user_id = Column(
         UUID(as_uuid=True), ForeignKey("users.id"), nullable=False, index=True
     )
     device_id = Column(
          UUID(as_uuid=True), unique=True, nullable=False, index=True
     )
     saved_date = Column(
         DateTime(timezone=False), nullable=False, server_default=sql.func.now()
     )
     device_info = # ....maybe a JSONB column?
```

And instead of storing a secret at all, use the `TokenService` to send a signed token containing `{"username": User.username, "device_id": UserDevice.id}`.

We can then just see if the signed token validates.

ristomcgehee: Yes, I thought it would be important to use a slow hash with a salt for the device secret, so that if an attacker gained access to the database, they would have a very hard time reversing the hash and finding the original secret. Using a salt means we have to hash the secret for each value we want to compare in the database. Since we're using a slow hash, I didn't want to have to hash the secret multiple times; if a user had several devices, that could add noticeable time. This is why I introduced the device ID. 

Now, I'll admit, I might be going a bit overkill with the security here; a slow hash with a salt might not be necessary with a randomly-generated 128-bit string. It could be fine storing a plain SHA256 hash of the secret in the database. But since this is a highly security sensitive area, I figured I would rather err on the side of too much security.

In regards to sending a signed token instead, if we could store all the necessary information in the token and not need a new database table, I think that would be the better option. However, since we will probably want to give the user the ability to revoke devices, I think a database table is necessary. I think using a signed token or a hash of a secret are both equally good approaches. I'm inclined to leave it as a hash of a secret since that's what I already have, but if there are advantages to the signed token that I'm not thinking of, I am willing to switch to that.

ewdurbin: I agree keeping a record in the DB for the Device is necessary for revocation. Only reason I think I prefer the token approach is that it makes the data in the table a bit more transparent, I'm not committed to it. Definitely welcome other thoughts.

warehouse/accounts/services.py
```diff
@@ -621,6 +626,79 @@ def get_password_timestamp(self, user_id):
...
+        device_id = self._generate_device_id(user)
+        device_secret = secrets.token_urlsafe(DEVICE_SECRET_BYTES)
+
+        self.db.add(
+            UserDevice(  # type: ignore
+                user=user,
+                device_id=device_id,
+                device_secret=self.hasher.hash(device_secret),
+            )
+        )
```
ewdurbin, line 684: We should add an [EventTag](https://github.com/pypi/warehouse/blob/main/warehouse/events/tags.py#L60) and [record an event](https://github.com/pypi/warehouse/blob/main/warehouse/accounts/services.py#L568) here

ristomcgehee: I'll add that!

ristomcgehee: I'm noticing that the majority of calls to `record_event` happen in view classes and not in database service classes. Would it be better to add the event to `warehouse/accounts/views.py` when I call `generate_device_id_secret()` (which will be in https://github.com/ristomcgehee/warehouse/pull/2).

ristomcgehee: I've added the recording of the event to my [phase 2 PR](https://github.com/ristomcgehee/warehouse/pull/2). I believe this MR is ready to be merged; I don't think the readthedocs check is failing due to changes introduced in my PR.

tests/unit/accounts/test_services.py
```diff
@@ -886,6 +887,107 @@ def test_check_recovery_code(self, user_service, metrics):
...
+        # Make sure the code handles things properly when it cannot generate an unused
+        # secret
+        monkeypatch.setattr(
+            secrets, "token_urlsafe", lambda length: "same_thing_every_time"
+        )
+        user_service.generate_device_id_secret(user.id)
+        with pytest.raises(RuntimeError):
+            user_service.generate_device_id_secret(user.id)
+
+    def test_check_device_valid(self, user_service, metrics):
```
ewdurbin, line 914: Breaking this test up into separate cases would be beneficial. That's in particular to the assertions around metrics, but overall its good practice.

warehouse/accounts/services.py
```diff
@@ -618,6 +623,78 @@ def get_password_timestamp(self, user_id):
         user = self.get_user(user_id)
         return user.password_date.timestamp() if user.password_date is not None else 0
 
+    def check_device_valid(self, user_id: str, device_id_secret_b64: str) -> bool:
+        """
+        Checks whether the given device id and secret are valid for the given user.
+        """
+        self._metrics.increment("warehouse.authentication.check_device_valid.start")
+
+        if not device_id_secret_b64:
```
ewdurbin, line 632: I'm assuming this is a None check? If so,

```suggestion
        if device_id_secret_b64 is None:
```

warehouse/accounts/services.py
```diff
@@ -618,6 +623,78 @@ def get_password_timestamp(self, user_id):
...
+            return False
+        try:
+            device_id_secret = DeviceIdSecret.from_base64(device_id_secret_b64)
+        except (KeyError, ValueError):
```
ewdurbin, line 636: I can't off the top of my head remember what all the juicy Exceptions we could expect from `json.loads(base64.b64decode(b64).decode("utf-8"))` are. I think we should probably add a parameterized test with a bunch of garbage values to exercise this. Off the top of my head I think a string that has extended unicode would raise an exception we wouldn't catch here and lead to a 500.

See https://github.com/pypi/warehouse/blob/46a15cc902531e156afb29be5ad6463565eaec84/tests/unit/macaroons/test_services.py#L132-L145 for a concise example of a similar concept.

warehouse/accounts/services.py
ewdurbin: I heard you on the brute-force side of things, but I'm still leaning towards using a TokenService rather than storing the hashed secret in the database.

Not to rely on tokens themselves, but just for the handling. When a device is saved we still retain the data in the DB for revocation purposes, we just send a signed token to the client that they can provide at login for warehouse to validate.

Any thoughts from the other @pypi/warehouse-admins?

ristomcgehee: I can change this to use a TokenService instead. I'm thinking that to complete this issue (https://github.com/pypi/warehouse/issues/5867), we won't need to add the database table yet. When we do the followup work to add the revocation, we can add the database table then. For now, we would just store the cookie in the user's browser and then verify it on the server side. Does that sound reasonable?

warehouse/accounts/services.py
ristomcgehee: Good idea!

warehouse/accounts/services.py
ristomcgehee: @ewdurbin What are your thoughts on my previous question?

ristomcgehee: I'll plan on using a signed token that is verified server side without a database table at the current time. I'm going to close this PR and create a new one since this is a significant rework.

ristomcgehee:
Closing to rework this to use a signed token as ewdurbin suggested.

PR #12953, opened 2023-02-07
Create user facing documentation structure

ewdurbin:
Uses [mkdocs](https://www.mkdocs.org) for building https://docs.pypi.org/

This will serve as the location for user-facing documentation moving forward. We should begin separating developer docs from the /docs directory, and @s-mm will be adding documentation for organization accounts here.

PR #13174, opened 2023-03-09
fix(dev): update ignore for doc8

miketheman:
With the changes in #12953, the existing path is no longer valid, and new doc builds surface linter errors.


ewdurbin (approved):
good catch, thanks!

di (approved):

PR #13177, opened 2023-03-09
OIDC: Render publisher on each release in admin view

woodruffw:
~~Needs tests.~~

~~This required some more involved changes: each `Release` now tracks its OIDC publisher (if present) via `Release.oidc_publisher`, and I've added `request.oidc_publisher` as a shim to reduce the amount of inline `isinstance` checks necessary.~~

This adds two new `File` columns:

* `File.uploading_user`: the `User` who uploaded this `File`, if uploaded by a user identity;
* `File.oidc_publisher`: the `OIDCPublisher` who uploaded this `File`, if uploaded by an OIDC identity;

This in turn allows us to display each `File`'s upload information in the admin pane, including if it came from an OIDC publisher.

Historical `File`s are not backloaded.


warehouse/oidc/models.py
```diff
@@ -194,6 +194,12 @@ class OIDCPublisher(OIDCPublisherMixin, db.Model):
     macaroons = orm.relationship(
         Macaroon, backref="oidc_publisher", cascade="all, delete-orphan", lazy=True
     )
+    files = orm.relationship(
+        File,
+        backref="oidc_publisher",
+        cascade="all, delete-orphan",
```
woodruffw, line 200: Flagging for review: not sure this cascade is correct.

di: I believe this would delete all `Files` when the `OIDCPublisher` was deleted, so no :)

ewdurbin (commented):
Overall I dig this. I am pondering if we should include backfill here and add a constraints like:

```
CheckConstraint('NOT(oidc_publisher_id IS NULL AND uploading_user_id IS NULL)')
CheckConstraint('NOT(oidc_publisher_id IS NOT NULL AND uploading_user_id IS NOT NULL)')
```

If they're mutually exclusive it's always best to enforce that in the db.

woodruffw:
Adding those constraints makes sense to me! I can do that now.

I think a backfill also makes sense, although I wonder about potentially being misleading: it's possible that existing projects have releases created by one user, but release files uploaded by multiple. That isn't recorded anywhere at the moment, but the backfill will make it appear as if the release-creating user is responsible for all files.

That's a minor issue IMO, though 🙂

ewdurbin:
Theoretically we can do a best effort by finding the upload in the Journal?

ewdurbin:
@woodruffw I'm also OK seeing this go without constraints initially, followed by PR for backfill migration + constraint enforcement.

di:
Constraints make sense, but I think backfill is out of scope for this particular thing -- we should file an issue to address that.

ewdurbin:
I don't think we can have the constraint without the backfill?

woodruffw:
> I don't think we can have the constraint without the backfill?

Yeah, these would need to happen together, I think.

ewdurbin:
Ah, solid yeah. Let's see the "you can only set one" constraint in this PR and at least file the issue for backfill + "you must set one" constraint.

woodruffw:
> Ah, solid yeah. Let's see the "you can only set one" constraint in this PR and at least file the issue for backfill + "you must set one" constraint.

SGTM!

woodruffw:
Not needed for coverage, but I went ahead and added some unit tests for the new constraint as well.

di:
Chatted with William: I'm concerned about publisher details about files going away if publishers are deleted. We're going to rework this to use the corresponding security event instead, which should contain most of the detail we need here (now and in the future).

woodruffw:
Superseded by #13205, closing.

PR #13178, opened 2023-03-09
fix(ci): invert condition for check-runs

miketheman:
With the condition being `failed`, we will include in-flight PRs that have not succeeded, which is counter to the messaging we surface to the user.


ewdurbin (approved):

miketheman:
Test failure appears to be from https://github.com/pypi/warehouse/pull/13175 - it's non-deterministic on the sort order, so I'll likely sort something to make it deterministic. I have not been able to replicate locally.

PR #13182, opened 2023-03-09
feat(ci): trigger ci flows after combine prs

miketheman:
The default behavior of GitHub Actions is to prevent new workflows from being triggered after using a `GITHUB_TOKEN` secret, unless manually called.


ewdurbin (approved):
Huzzah. Thanks @miketheman.

di (approved):

PR #13202, opened 2023-03-13
User docs for organization account in PyPI

s-mm:
This PR will create user docs for organization accounts in PyPI.


ewdurbin (changes requested):
Thanks for getting this started @s-mm. I've gone through at a high level, mostly focused on the tool and conventions. There are a couple suggested copy changes for your consideration too. I didn't repeat specific feedback on syntax, nor did I call out every instance of where a copy change might apply, so if you accept a change in one place (like the note on being logged in) please apply it in the other instances.

The contents on the whole appears straight forward, so I think once we get some of the basics covered this can merge and be reviewed in user-testing.

ewdurbin (approved):
Thank you @s-mm this looks fantastic now.

I'm going to push some commits to your branch to do the following:

1) Restructure the directory to implement your initial navigation like-for-like
2) Implement a bunch of nitpicks around formatting
3) Add admonitions to all org pages saying "in closed beta" or similar

I'll ask for you to approve these changes then this will be set to ship!

ewdurbin:
so pleasant

PR #13205, opened 2023-03-14
warehouse: add File events, render

woodruffw:
NB: This has the same problem that other event types do, which is that deleting a `File` will also delete its events.

Supersedes #13177.


warehouse/accounts/__init__.py
```diff
@@ -60,6 +61,16 @@ def _user(request):
...
 
 
+def _oidc_publisher(request):
+    if request.identity is None:
+        return None
+
+    if not isinstance(request.identity, OIDCPublisher):
+        return None
+
+    return request.identity
```
di, line 71:
```suggestion
    return request.identity if isinstance(request.identity, OIDCPublisher) else None
```

warehouse/forklift/legacy.py
```diff
@@ -1345,6 +1347,21 @@ def file_upload(request):
         file_data = file_
         request.db.add(file_)
 
+        project.record_event(
```
di, line 1350: Instead of recording these as `ProjectEvents`, I think we'll instead want to extend the `HasEvents` mixin to the `File` model, so this would become `file_.record_event(...)` and we could later do a `joinedload` to call `file_.events` to get all events that belong to each file for a release.

We might also want to find a way to have `<Project>.events` eventually include events of sub-models like releases and files, but for now I think that would be sufficient.

woodruffw: SGTM! Any thoughts on how `File.events` should be rendered in the project's security history? I figure I can just return all event streams under the project and interleave them by timestamp, but not sure if you have something else in mind.

di: That could work if we want to keep`<Project>.events` to just `ProjectEvents`. I'm somewhat concerned that this could become an N+1 query, but as long as we can construct a query that does a joinedload for _all_ event types, I think it will be fine.

dstufft: I'm not sure if you can do a joined load like that, but you can do a union query for all event types.

warehouse/manage/views.py
```diff
@@ -4986,13 +4983,23 @@ def manage_project_history(project, request):
...
 
-    events_query = (
+    project_events_query = (
         request.db.query(Project.Event)
         .join(Project.Event.source)
         .filter(Project.Event.source_id == project.id)
         .order_by(Project.Event.time.desc())
     )
 
+    file_ids = [file.id for release in project.releases for file in release.files]
```
woodruffw, line 4993: Flagging for review: this is a pretty gnarly way to get all associated file events, and there's probably a better way 🙂

woodruffw:
These changes somehow affected the render behavior for pagination:

```
E       AssertionError: assert {'events': <p...FhunBmyckqD')} == {'events': <p...FhunBmyckqD')}
E         Omitting 2 identical items, use -vv to show                                          
E         Differing items:                                                                     
E         {'events': <paginate.Page: Page 1/2>} != {'events': <paginate.Page: Page 1/2>}       
E         Use -v to get more diff                                                              
                                                                                               
tests/unit/manage/test_views.py:9691: AssertionError 
```

(I'll root cause this, but pasting just in case anybody has any ideas.)

warehouse/manage/views.py
di: If we put a project ID column on the event as well, we wouldn't need to do this.

woodruffw:
This doesn't make a ton of sense to me: we query for the `UNION` of `File.Event` and `Project.Event`, but the events returned are all of type `ProjectEvent` even when their instance content is clearly a `File.Event`:

```
E         Full diff:
E           [
E            ProjectEvent(id=UUID('a77fd9af-872c-44f5-b962-4969e67dbe34'), tag='fake:event', time=datetime.datetime(2019, 2, 5, 
17, 18, 18, 462634), ip_address_string='0.0.0.0', additional=None, ip_address_id=UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), s
ource_id=UUID('7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de')),
E         -  FileEvent(id=UUID('8904bae0-6165-4bb8-a067-03c40be47e41'), tag='fake:event', time=datetime.datetime(2018, 2, 5, 17,
 18, 18, 462634), ip_address_string='0.0.0.0', additional={'project_id': '7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de'}, ip_address_id=
UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), source_id=UUID('bfcab812-96a9-4f12-8b10-157fa9f69f1e')),
E         ?  ^^^^
E         +  ProjectEvent(id=UUID('8904bae0-6165-4bb8-a067-03c40be47e41'), tag='fake:event', time=datetime.datetime(2018, 2, 5, 
17, 18, 18, 462634), ip_address_string='0.0.0.0', additional={'project_id': '7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de'}, ip_address_
id=UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), source_id=UUID('bfcab812-96a9-4f12-8b10-157fa9f69f1e')),
E         ?  ^^^^^^^
E            ProjectEvent(id=UUID('2f1fa0d4-9234-4d3c-b2f0-92c2e86840fc'), tag='fake:event', time=datetime.datetime(2017, 2, 5, 
17, 18, 18, 462634), ip_address_string='0.0.0.0', additional=None, ip_address_id=UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), s
ource_id=UUID('7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de')),
E         -  FileEvent(id=UUID('8ca74e1c-4d81-4192-ae40-8c52525c9498'), tag='fake:event', time=datetime.datetime(2016, 2, 5, 17,
 18, 18, 462634), ip_address_string='0.0.0.0', additional={'project_id': '7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de'}, ip_address_id=
UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), source_id=UUID('bfcab812-96a9-4f12-8b10-157fa9f69f1e')),
E         ?  ^^^^
E         +  ProjectEvent(id=UUID('8ca74e1c-4d81-4192-ae40-8c52525c9498'), tag='fake:event', time=datetime.datetime(2016, 2, 5, 
17, 18, 18, 462634), ip_address_string='0.0.0.0', additional={'project_id': '7c09ba5b-9c5d-4fc2-8341-afa3cc4fc5de'}, ip_address_
id=UUID('589f9951-1169-4563-9a97-e9ea21bdc5ba'), source_id=UUID('bfcab812-96a9-4f12-8b10-157fa9f69f1e')),
E         ?  ^^^^^^^
E           ]

tests/unit/manage/test_views.py:9688: AssertionError
```

I would have expected that to be caught by each `Event` subclasses' polymorphic identity, but maybe I'm missing something.

dstufft:
You have to query against `Event`, but you need to do some extra stuff to make it possible to query against it, see https://docs.sqlalchemy.org/en/14/orm/inheritance.html#abstract-concrete-classes

woodruffw:
> You have to query against `Event`, but you need to do some extra stuff to make it possible to query against it

Yep -- I think what I need is to have `Event` inherit `db.Model`, right?

The only problem then is that `Event` has a bunch of `declared_attr`s that aren't prepared, since the subclass init hook prepares the state they need:

```
warehouse-web-1            |   File "/opt/warehouse/src/warehouse/events/models.py", line 58, in source_id
warehouse-web-1            |     "%s.id" % cls._parent_class.__tablename__,
```

di:
Instead of re-architecting our event models, why don't we just issue 2 queries and stitch the results together where we need them?

woodruffw:
> Instead of re-architecting our event models, why don't we just issue 2 queries and stitch the results together where we need them?

I can do that instead -- the main reason I didn't is that the pagination mechanism is lazy and tied to the ORM layer, so bypassing it by making two queries means realizing their contents eagerly.

If that's okay, though, I'm happy to do that!

woodruffw:
> > Instead of re-architecting our event models, why don't we just issue 2 queries and stitch the results together where we need them?
> 
> I can do that instead -- the main reason I didn't is that the pagination mechanism is lazy and tied to the ORM layer, so bypassing it by making two queries means realizing their contents eagerly.
> 
> If that's okay, though, I'm happy to do that!

Actually, thinking about this more: I definitely don't have to eagerly load all events for the two queries. I can just issue ranged queries for both.

woodruffw:
This should be good to go -- I didn't fix the underlying `Event` model issue, since it's only surfacing as a type issue and not a "shape" one. I've added a backstop to the unit tests for when that changes, though.

warehouse/events/tags.py
```diff
@@ -123,7 +123,6 @@ class Project(EventTagEnum):
         OwnersRequire2FAEnabled = "project:owners_require_2fa:enabled"
         ProjectCreate = "project:create"
         ReleaseAdd = "project:release:add"
-        ReleaseFileRemove = "project:release:file:remove"
```
di, line 126: Hmm, I didn't realize we already had this. Trying to decide if it makes more sense for this to stay the same (to maintain consistency) or change to what you have (to be more clear on what the event is generated from).

di (approved):
LGTM aside from that one comment. Will merge after the weekend.

woodruffw:
RTD build failure? https://readthedocs.org/projects/blogpypiorg/builds/19842091/

PR #13206, opened 2023-03-14
fix(ci): correct combiner logic

miketheman:
- Update the initial `GET` call to only select Open PRs, instead of every PR ever, which is wasteful.
- Perform a secondary API call for the given PR and check it's `mergable` state (no merge conflicts)
- Update checks logic to check if all checks in the check run are green, removing any still in `pending` state

### DIFF
--- a/.github/workflows/combine-prs.yml
+++ b/.github/workflows/combine-prs.yml
@@ -39,9 +39,10 @@ jobs:
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const pulls = await github.paginate('GET /repos/:owner/:repo/pulls', {
+            const pulls = await github.paginate('GET /repos/{owner}/{repo}/pulls', {
               owner: context.repo.owner,
-              repo: context.repo.repo
+              repo: context.repo.repo,
+              state: 'open',
             });
             branches = [];
             prs = [];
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-prs.yml
```diff
@@ -52,20 +53,33 @@ jobs:
               if (pull['user']['login'].match(/^(dependabot\[bot\]|dependabot-preview\[bot\])$/)) {
                 console.log('Branch matched: ' + branch);
                 statusOK = true;
+                // if the PR is not mergeable, bail early.
+                const pull_request = await github.rest.pulls.get({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  pull_number: pull['number']
+                });
+                if(pull_request.data.mergeable != true) {
```
miketheman, line 62: Something I learned today is that the `github` client has a `rest` endpoint, and this returns objects that can be used instead of array accessors. If this works well, we could explore converting the rest of the code to use these accessors.

miketheman:
I tested this change locally with the [`act` tool](https://github.com/nektos/act) with this command, after commenting out the actual `Created combined branch` steps onwards to not actually do the thing.

`act workflow_dispatch -j combine-prs -s GITHUB_TOKEN=$MY_GITHUB_API_TOKEN -e payload.json`
with `payload.json`:
```json
{
  "inputs": {
    "mustBeGreen": "true",
    "mustBeUnblocked": "true"
  }
}
```

.github/workflows/combine-prs.yml
di, line 62: Does this only check for merge conflicts, or will a branch that is out-of-date also be "not mergeable"?

miketheman: After you asked this, I took some more time to create the conditions necessary, so I set up branch protection on a fork, and confirmed that needing to "Update branch" due to branch protection will **not** mark a branch as unmergeable, so we're good to go!

di (approved):

PR #13214, opened 2023-03-16
Don't retry exceptions that come from bad email formatting

di:
These are due to the emails being malformed, and will never succeed.


ewdurbin (approved):

ewdurbin:
I suppose we could also consider stripping line breaks?

di:
> I suppose we could also consider stripping line breaks?

I think we should do #13216 instead, there's other things that could go wrong here.

PR #13215, opened 2023-03-16
Remove newlines from OIDC email subjects

di:
These were being retried indefinitely due to #13214.


di:
This is the quick fix, https://github.com/pypi/warehouse/issues/13216 captures what we should do to ensure this doesn't happen again.

ewdurbin (approved):

PR #13219, opened 2023-03-16
Address some warnings surfaced in tests

miketheman:
See commit messages for specifics. Drops ~30 warnings from total count.


di (approved):

di:
@miketheman This PR added `warehouse/admin/static/dist/.gitkeep` and `warehouse/static/dist/.gitkeep`, was that intentional? These files keep getting deleted for me in local development now that this has been merged.

miketheman:
@di yes, they were intentionally added in https://github.com/pypi/warehouse/commit/bc8a290556896bcd9abafbd2145ac4795d58694e

miketheman:
When you say local development, is that inside the docker container or outside of it?

di:
Both, I think? I'm guessing something about the static asset generation is wiping these away inside the container, which results in them being deleted from my source tree, since they are mapped between the two. Maybe the fix is excluding these from getting mounted inside the container?

miketheman:
Hmmm. That's one way, I guess.

Maybe the better option is to approach this another way.
Since these directories are needed for the tests only, we can have a pytest autouse fixture create them at the beginning of the test suite, and thus remove the gitkeep files.

miketheman:
@di Now I think I'm seeing things. 😆  I removed the gitkeeps and directories, cleaned everythign out and ran the tests again, and see no `UserWarning`.
I'll remove them, and keep an eye out for if they resurface,, and if they do I have a better idea to solve with pytest fixtures.

PR #13221, opened 2023-03-16
fix: replace text with matrix

### LINKED ISSUE #12431: Display organization member roles in a matrix
To improve user experience, the roles within an organization can be displayed in a matrix with check marks to indicate what each type of user can and cannot do.

The description of billing manager should also say what that role cannot do.

miketheman:
Adds some context to the header, and replaces the lines with a table and
icons showing the specifics.

Reordered the lines to push org/team management to the top, and projects
follow.

Fixes #12431


miketheman:
Now looks like this:
Screenshot 2023-03-16 at 5 35 27 PM

di (approved):
I'll leave it to @ewdurbin to say whether the content accurate, but this is exactly what I was imagining in terms of structure. Thank you!

miketheman:
(sorry for the force-push, but I had missed a closing translations tag, and wanted to make it easy to re-run `make translations` if I get out order)

miketheman:
I checked the content against https://github.com/pypi/warehouse/blob/5046d0b747566dbbb9504e35c36359688b2d89a5/warehouse/organizations/models.py#L323-L405 - thankfully there's great descriptions there that helped.

ewdurbin (approved):
the *teensiest* nit: could we add a comment in the location you referenced to create this saying "don't forget to update the roles.html grid" or similar?

regardless, this is good to 🚢

miketheman:
> the _teensiest_ nit: could we add a comment in the location you referenced to create this saying "don't forget to update the roles.html grid" or similar?

Good call, added in 4094d3a58facccf08a90cadcb96303b8ab3136be

Issue #13233, opened 2023-03-17
Org tests that use `token_service.dumps()` are flakey


di:
As seen in https://github.com/pypi/warehouse/actions/runs/4451994819/jobs/7819250263, tests that depend on `token_service.dumps()` returning the same token across multiple calls can be flakey.

These tests should use:

```python
with freezegun.freeze_time(now):
    ...
```

as the context for anything that generates or asserts on tokens.

PR #13238, opened 2023-03-18
test: wrap resend test case with a time freeze

miketheman:
One of the newer test cases using the `token_service` fixture wasn't wrapped with the freeze decorator.
Other test cases in this module using the `token_service` **fixture** are already wrapped as such, this follows suit.

There are some tests that use a stub token_service with guaranteed `loads/dumps` behavior that don't use the fixture, so they don't need to be frozen.

Resolves: #13233


ewdurbin (approved):

ewdurbin:
🤦

PR #13250, opened 2023-03-20
events, migrations: fix event deletion cascades

woodruffw:
I _think_ this is correct -- if my understanding of SQLAlchemy is correct, this does a DB level `ON DELETE CASCADE` rather than requiring the ORM level to collect all objects to delete, which can be lossy.

Ref: https://docs.sqlalchemy.org/en/20/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships


di (approved):
Makes sense to me!

PR #13255, opened 2023-03-21
refactor(ci): rework combiner logic

miketheman:
With the updated approach, we make a single API call to fetch all the relevant data to an object, and iterate through it based on our conditions.

- update inputs to match upstream
- simplified logic, commented steps
- logged output is visually pleasing


di (approved):

miketheman:
The good news is that this worked to select the right PRs, but failed to create the right pull request. A fix should be incoming soon!

PR #13264, opened 2023-03-22
bullied into this

ewdurbin:


miketheman (commented):
Actually some bugs in the rendered xml.
https://blogpypiorg--13264.org.readthedocs.build/atom.xml

ewdurbin:
yeah, I don't like the atom plugin. probably going to rss.

ewdurbin:
hoping one day, you find me here. yes you. the feed fiend.

and you find yourself wondering why the branch is called atom. but its an rss feed.

it is because you were unkind.

PR #13272, opened 2023-03-22
warehouse: optional OIDC environment constraint

### LINKED ISSUE #13270: [OIDC] Introduce additional (optional) trust constraints
Summarized from a private discussion on the OIDC beta.

At the moment, the OIDC trust relationship between GitHub and PyPI is defined by three variables: the GitHub org name, the repo name, the workflow within the repository. If PyPI receives an OIDC token that satisfies these three variables, it considers that token valid for exchange/minting purposes.

This is sufficient for many security models, but may not be sufficient for projects with maintainers who can commit and/or make branches but who should **not** be able to make releases to PyPI. An example "malicious maintainer" scenario:

* `foo/bar @ release.yml` is the trusted OIDC configuration, with `release.yml` configured to trigger on `release: published` events;
* Mallory has commit and branch access to `foo/bar`, but not release or environment modification access;
* Mallory creates a `malicious` branch on `foo/bar` and modifies `release.yml` to accept the `workflow_dispatch` trigger;
* Mallory manually dispatches `release.yml` using the GitHub API or Web UI, resulting in her ability to publish to PyPI without ordinary release permissions.

Critically, this can happen even when `release.yml` constrains the publishing job to a specific environment (e.g. `environment: release`), since Mallory can simply strip the release off of the workflow before triggering it.

To prevent this, we need to introduce additional trust constraints: if the PyPI OIDC configuration includes the GitHub Actions environment name, Mallory won't be able to strip it off. From there, the environment can enable "manual approval," meaning that Mallory won't be able to trigger it unless she's explicitly trusted as a workflow runner by whoever configured OIDC publishing (i.e., a project owner on PyPI). 

CC @di

woodruffw:
Closes #13270.

### DIFF
--- a/tests/common/db/oidc.py
+++ b/tests/common/db/oidc.py
@@ -26,6 +26,7 @@ class GitHubPublisherFactory(WarehouseFactory):
     repository_owner = factory.Faker("pystr", max_chars=12)
     repository_owner_id = factory.Faker("pystr", max_chars=12)
     workflow_filename = "example.yml"
+    environment = "production"
 
 
 class PendingGitHubPublisherFactory(WarehouseFactory):
@@ -38,3 +39,4 @@ class PendingGitHubPublisherFactory(WarehouseFactory):
     repository_owner = factory.Faker("pystr", max_chars=12)
     repository_owner_id = factory.Faker("pystr", max_chars=12)
     workflow_filename = "example.yml"
+    environment = "production"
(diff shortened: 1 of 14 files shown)


di:
I think this is good to go now.

One thing I'll highlight is that the environment name shows up in both the new `environment` claim and the `sub` claim: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#filtering-for-a-specific-environment. I didn't see a reason why we need to verify the environment name in both claims, and opted to just verify `environment` (because parsing `sub` will be a pain).

Another thing I'll highlight is that this is the first claim we've verified which might not be present in the token: if the token didn't come from within an environment, the `environment` claim is not present.

@woodruffw, would appreciate if you'd review (since this is mostly my changes and not yours).

woodruffw (commented):
LGTM! I can't submit an explicit approval since I'm the original PR author 🙂

di (approved):

PR #13231, opened 2023-03-17
create a blog

ewdurbin:

PR #13273, opened 2023-03-23
dev UX fixups

ewdurbin:
- A couple missed `docker-compose` -> `docker compose`
- stripe-mock now has arm builds! https://github.com/stripe/stripe-mock/issues/281
- Build/run our docs stuff in their own image
- Remove some stray debugging that was causing dev-docs to open a python REPL 🙃
- Better mechanism for sharing container images between services
- Don't delete compose volumes on `make stop`, rather do that only when `make purge` is called

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,61 @@ RUN NODE_ENV=production npm run build
 
 
 
+# We'll build a light-weight layer along the way with just docs stuff
+FROM python:3.11.2-slim-bullseye as docs
+
+# Install System level build requirements, this is done before
+# everything else because these are rarely ever going to change.
+RUN set -x \
+    && apt-get update \
+    && apt-get install --no-install-recommends -y \
+        build-essential git libffi-dev libcairo2-dev libfreetype6-dev libpq-dev
+
+# We create an /opt directory with a virtual environment in it to store our
+# application in.
+RUN set -x \
+    && python3 -m venv /opt/warehouse
+
+# Now that we've created our virtual environment, we'll go ahead and update
+# our $PATH to refer to it first.
+ENV PATH="/opt/warehouse/bin:${PATH}"
+
+# Next, we want to update pip, setuptools, and wheel inside of this virtual
+# environment to ensure that we have the latest versions of them.
+# TODO: We use --require-hashes in our requirements files, but not here, making
+#       the ones in the requirements files kind of a moot point. We should
+#       probably pin these too, and update them as we do anything else.
+RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip setuptools wheel
+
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -186,22 +186,33 @@ services:
...
 
   blog:
-    image: warehouse:docker-compose
+    image: warehouse:docker-compose-docs
+    pull_policy: never
     command: mkdocs serve -a 0.0.0.0:8000 -f docs/blog.yml
     volumes:
-      - ./docs/blog.yml:/opt/warehouse/src/docs/blog.yml:z
-      - ./docs/blog:/opt/warehouse/src/docs/blog:z
+      - ./.git:/opt/warehouse/src/.git:z
```
miketheman, line 214: I had a similar note about git directory previously 
https://github.com/pypi/warehouse/pull/13231#discussion_r1142476044

If it needs to be mounted, should it be in ro mode?

Dockerfile
```diff
@@ -132,7 +187,7 @@ RUN set -x \
     && apt-get update \
     && apt-get install --no-install-recommends -y \
         libpq5 libxml2 libxslt1.1 libcurl4  \
-        $(if [ "$DEVEL" = "yes" ]; then echo 'bash libjpeg62 postgresql-client build-essential libffi-dev libxml2-dev libxslt-dev libpq-dev libcurl4-openssl-dev libssl-dev git libcairo2-dev libfreetype6-dev libjpeg-dev libpng-dev libz-dev'; fi) \
```
miketheman, line 135: When the blog was added, this is the list of deps added for it to work:
git libcairo2-dev libfreetype6-dev libjpeg-dev libpng-dev libz-dev

I think all of them should migrate to the new docs target instead, right?

docker-compose.yml
ewdurbin: oh heck yeah, good call. it's just for generating the timestamps on blogs. i missed it getting removed.

Dockerfile
ewdurbin: ty.

docker-compose.yml
```diff
@@ -186,22 +186,33 @@ services:
...
 
   blog:
-    image: warehouse:docker-compose
+    image: warehouse:docker-compose-docs
+    pull_policy: never
     command: mkdocs serve -a 0.0.0.0:8000 -f docs/blog.yml
     volumes:
-      - ./docs/blog.yml:/opt/warehouse/src/docs/blog.yml:z
-      - ./docs/blog:/opt/warehouse/src/docs/blog:z
+      - ./.git:/opt/warehouse/src/.git:ro
```
ewdurbin, line 214: This is needed for rss feedz

miketheman (approved):

PR #13295, opened 2023-03-24
Use recaptcha.net instead of google.com

di:
This makes our reCAPTCHA more likely to work for a global user base, per https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally


ewdurbin:
We probably need CSP updates for this, no?

di:
> We probably need CSP updates for this, no?

Yeah, that's the changes to `warehouse/recaptcha.py`

ewdurbin:
ah, sorry I expected that to be in the csp.py file. I understand now :)

ewdurbin (approved):

Issue #3081, opened 2018-02-27
Utilize SQLAlchemy to prevent N+1 queries on the deferred description column


dstufft:
It doesn't exist *yet*, but when https://gerrit.sqlalchemy.org/#/c/zzzeek/sqlalchemy/+/679/ and https://bitbucket.org/zzzeek/sqlalchemy/issues/3037/support-setting-load-options-on-instances are both done it would be a good idea to utilize them so that accessing the ``Release.description`` field without explicitly undefering it would raise an error.

This would prevent us from accidentally accessing it and having SQLAlchemy issue extras queries behind the scenes, causing N+1 queries.

miketheman:
Some N+1 query work:

- https://github.com/pypi/warehouse/pull/19593
- https://github.com/pypi/warehouse/pull/19617

PR #13311, opened 2023-03-27
Use `raiseload` option for all queries

di:
Sort of towards https://github.com/pypi/warehouse/issues/3081.

This PR turns on the `raiseload` option by default for all queries, essentially making N+1 queries impossible in our codebase, and requiring explicit loads instead, rather than per-query as documented in https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#preventing-unwanted-lazy-loads-using-raiseload.

Marking this as draft for now because there are a lot of test failures to address. I'm also not entirely sure yet if this also covers deferred columns as well, because that seems to be a separate setting: https://docs.sqlalchemy.org/en/20/orm/queryguide/columns.html#using-raiseload-to-prevent-deferred-column-loads


di:
@miketheman Do you think we're ready to revisit this?

miketheman:
@di I think this is worthwhile, but should probably come after #15634 - since I was trying to ensure a query count before/after to support this effort from a web-side, since many of our potential N+1 come from HTML template rendering iteration loops.

miketheman:
> OK! Let's say this is blocked on #15634.

unblocked now that we can count queries emitted when loading a page via `webtest`

di:
Rebased. Quite a number of test failures to work through here if anyone is feeling up for it 🙂

firefly9966 (approved):

PR #13325, opened 2023-03-29
templates/publishing: remove accidental rendering

### LINKED ISSUE #13298: [OIDC] Slightly incorrect rendering in providers table
Just noticed this while writing docs: there are some definitely-wrong-looking `None`s rendered as part of the table:

Screenshot 2023-03-24 at 8 54 33 PM

I'll root cause and fix this; just filing the issue for tracking + visibility.

woodruffw:
Closes #13298.

~~`or ""` is a bit of a hack, but it's apparently (from looking online) the most common way to suppress rendering of Python expressions with side effects in Jinja2.~~


warehouse/templates/manage/account/publishing.html
```diff
@@ -34,7 +34,8 @@ <h3>Projects with active publishers</h3>
   {% set projects = [] %}
   {% for project in user.projects %}
     {% if project.oidc_publishers %}
-      {{ projects.append(project) }}
+      {# NOTE: `or ""` to prevent rendering None #}
+      {{ projects.append(project) or "" }}
     {% endif %}
   {% endfor %}
```
di, line 40: I think the Jinja-ish way to do this would be something like:

{% set projects = user.projects | selectattr('oidc_publishers') %}

woodruffw: Whoops, you're completely right (and that's way nicer). I'll update now.

di (approved):

PR #13353, opened 2023-03-31
refactor: replace custom clipboard with stimulus

miketheman:
Replaces the `clipboard` dependency with a Stimulus controller.
Uses native `navigator` browser API.

Refs: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText


miketheman:
The `cli[board` dependency is still used in the admin interface, a subsequent PR will attempt to address that with behaviors available in AdminLTE/Bootstrap/jQuery and allow us to remove it completely.

di (approved):

PR #13391, opened 2023-04-07
Remove `record_event` from `UserService` and `OrganizationService`

di:
This removes the vestigial `record_event` function from the `UserService` and `OrganizationService` in favor of `User.record_event` and `Organization.record_event` functions provided by`HasEvents`.

This brings some additional consistency to how we record events, and also in _most_ cases, this saves us from having to do an unnecessary `UserService.get_user()` call to get the user by ID.


ewdurbin (approved):

ewdurbin:
divine.

PR #13392, opened 2023-04-08
Allow closed beta participants to access org features

ewdurbin:
This refactors the direct AdminFlag check to go through a request_method that allows access to org features if one has been assigned to them directly.

For closed-beta, we'll be onboarding users by creating an Organization object, and an OrganizationRole object directly, which will then enable access via this gate.

### DIFF
--- a/warehouse/predicates.py
+++ b/warehouse/predicates.py
@@ -82,11 +82,6 @@ class ActiveOrganizationPredicate:
         )
 
         if (
-            # Organization accounts are disabled.
-            request.flags.enabled(AdminFlagValue.DISABLE_ORGANIZATIONS)
-        ):
-            return False
-        elif (
             # Organization is active.
             organization.is_active
             # Organization has active subscription if it is a Company.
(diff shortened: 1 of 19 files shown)


warehouse/predicates.py
```diff
@@ -96,6 +91,11 @@ def __call__(self, context: Organization | Team, request):
             )
         ):
             return True
+        elif (
+            # Organization accounts are disabled.
+            request.flags.enabled(AdminFlagValue.DISABLE_ORGANIZATIONS)
+        ):
+            return False
```
di, line 98: Shouldn't this still come first?

di (approved):
LGTM aside from that one comment

warehouse/predicates.py
ewdurbin: No, we're not flipping the flag until we're ready for open signup.

ewdurbin: Same logic as is in https://github.com/pypi/warehouse/pull/13392/files#diff-fd646332e15b1650ac3a05f944bf6ae36335637b687d06676bb67f7d6f3b1fefR70-R79

This allows users who have an organization to bypass the flag

di: Got it!

Issue #12428, opened 2022-10-25
Information displayed when deleting an organization is not sufficient


s-mm:
When deleting an organization, there is a link to transfer ownership and then delete the project. The link should be changed to transfer ownership, remove a project or delete the project. 

To recreate this issue-

- Click on Your organizations
- Click on Manage for a specific organization
- Click on Settings
- Scroll to the bottom of the page and click on Delete organization

Transferring a project to an individual user is not sufficient. It should be removed from the organization or deleted.

s-mm:
Priority - high

ewdurbin:
ca2000c9a6fbb7d732536582cdfbe7178ecf018d: <img width="813" alt="transanddelete" src="https://user-images.githubusercontent.com/1200832/230731807-c01a879f-0a76-4d2b-963f-a2d0aa720a6a.png">

Issue #12426, opened 2022-10-25
Change in organization details is not reflected in the Stripe invoice


s-mm:
When an organization account name is updated, the change is not reflected in the Stripe invoice/receipt. The same issue occurs when billing address is updated.

1. Should we display the organization account name or organization display name?
2. The invoice/receipt should say PyPI instead of Warehouse
3. Any change to organization account/display name and billing address should show up in the invoice/receipt. 

![Image](https://user-images.githubusercontent.com/37237726/197813602-575481ed-8d64-4062-ad18-d1a9b8937a6b.png)

ewdurbin:
Seems we should definitely be sending an update to stripe when org names change. We can address the prefix "Warehouse Organization" at that time.

We can probably update the product names to reflect "PyPI Organization Account" and similar at that time.

I anticipate we'll also need to do some updates to capture a human name for billing before launch. We'll address this as part of that phase.

ewdurbin:
> 1. Should we display the organization account name or organization display name?

I'm gonna go with "both"

> 2. The invoice/receipt should say PyPI instead of Warehouse

"Warehouse" comes from the `site.name` setting, so it will automatically be "PyPI" for pypi.org. 

https://github.com/pypi/warehouse/blob/ff5db0f76133185a1735522f42eba3f99331ff8f/warehouse/manage/views/organizations.py#L489

> 3. Any change to organization account/display name and billing address should show up in the invoice/receipt.

I'll get an update method for the customer name put together, that shouldn't be a problem.

We don't touch billing address, that is all in the portal.

ewdurbin:
addressed in 089034d72 as part of #13395

Issue #12423, opened 2022-10-25
Project maintainer should not lose maintainer status


s-mm:
When a project is added to an organization, the individual owner/maintainer loses their project role status. This status should exist even after the transfer.

To recreate the issue-

- Click on Your organizations
- Click on Manage for a specific organization
- Click on Projects
- Scroll down to the bottom of the page and select a project from the drop down list. The projects associated with the user will appear in the list. 
- Click on Transfer existing project

The organization becomes the sole owner of the project.

![Image](https://user-images.githubusercontent.com/37237726/197813867-526f9d36-a48d-4b22-bccd-869fbaa65a8f.png)

ewdurbin:
I'd like to hear more of your thoughts here @di, since I've been told this feedback comes from you.

Since we require a User to be both owner of a project and owner of an organization to perform a transfer, they retain their role implicitly, but we don't leave the dangling Role associated.

di:
It's information loss, basically. There's no option to maintain the current list of collaborators, so transferring a project to an organization would require who were the original collaborators and re-adding them to the organization.

This is going to be very tedious for a large organization with lots of projects and lots of unique collaborators on those projects, who likely want to maintain the existing collaborator structure.

di:
> Since we require a User to be both owner of a project and owner of an organization

Why do we require them to be organization owners? IMO this should be like regular collaborator role invites, where anyone can invite an organization to accept a project transfer, but only organization owners can accept the transfer.

I'm imagining it's going to be a common use case where it doesn't make sense for any project collaborators to be organization owners.

ewdurbin:
It's a bit too much for our timeline to implement invitations. Added notes in 0e34a0e25 and filed #13394 as a feature request.

di:
Is invitation strictly necessary? My main concern is losing the lost of roles for all the collaborators that didn't initiate the transfer. Can their roles get turned into non-owner roles on the organization instead of getting dropped?

ewdurbin:
No one except the transferring owner has their permissions changed

PR #13395, opened 2023-04-08
Organization launch priorities

### LINKED ISSUE #12432: Deleting an organization should add a warning
When deleting an organization, a warning should let the owner know to download invoices/receipts from Stripe. It should also indicate what will happen to the subscription plan - whether the plan will be cancelled automatically or if the user has to cancel the plan before deleting the organization.

ewdurbin:
Closes #12432 #12423 #12426 #12428

### DIFF
--- a/warehouse/organizations/models.py
+++ b/warehouse/organizations/models.py
@@ -418,7 +418,7 @@ class Organization(HasEvents, db.Model):
             return None
 
     def customer_name(self, site_name="PyPI"):
-        (f"{site_name} Organization -" f"self.display_name ({self.name})"),
+        return f"{site_name} Organization - {self.display_name} ({self.name})"
 
 
 class OrganizationNameCatalog(db.Model):
--- a/warehouse/organizations/models.py
+++ b/warehouse/organizations/models.py
@@ -417,6 +417,9 @@ class Organization(HasEvents, db.Model):
         else:
             return None
 
+    def customer_name(self, site_name="PyPI"):
+        (f"{site_name} Organization -" f"self.display_name ({self.name})"),
+
 
 class OrganizationNameCatalog(db.Model):
 
--- a/warehouse/templates/manage/organization/settings.html
+++ b/warehouse/templates/manage/organization/settings.html
@@ -235,7 +235,7 @@
         {% for project in active_projects %}
         <li>
           <strong>{{ project.name }}</strong> -
-          {% trans transfer_href=request.route_path('manage.project.roles', project_name=project.name), delete_href=request.route_path('manage.project.settings', project_name=project.name) %}
+          {% trans transfer_href=request.route_path('manage.project.settings', project_name=project.name), delete_href=request.route_path('manage.project.settings', project_name=project.name) %}
             <a href="{{ transfer_href }}">transfer ownership</a>
             or
             <a href="{{ delete_href }}">delete project</a>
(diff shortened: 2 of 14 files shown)


warehouse/templates/manage/organization/settings.html
```diff
@@ -225,20 +225,21 @@ <h3>{% trans %}Cannot delete organization{% endtrans %}</h3>
...
-          You must transfer ownership or delete this project before you can delete your organization.
-        {% pluralize %}
-          You must transfer ownership or delete these projects before you can delete your organization.
-        {% endtrans %}
+        {% trans %}For each project, you must either:{% endtrans %}
+        <ul>
+          <li>{% trans %}Transfer ownership and remove the project from your organization{% endtrans %}</li>
+          <li>{% trans %}Delete the project{% endtrans %}</li>
+        </ul>
+        {% trans %}Before you can delete your organization.{% endtrans %}
```
miketheman, line 233: **question:** ‏non-blocking, but is there a way to support translators in that this sentence is part of the one preceding the list items?

warehouse/organizations/models.py
```diff
@@ -417,6 +417,9 @@ def active_subscription(self):
         else:
             return None
 
+    def customer_name(self, site_name="PyPI"):
+        (f"{site_name} Organization -" f"self.display_name ({self.name})"),
```
miketheman, line 421: **suggestion:** ‏Correct f string representation here
```suggestion
        (f"{site_name} Organization - {self.display_name} ({self.name})"),
```

Curious why this wasn't caught by test - do we not assert on the outcome of the call of this function somewhere?

miketheman (approved):
One change requested with a fix. I'll approve so as to not block when corrected, since I'll be offline by then.

warehouse/organizations/models.py
ewdurbin: will add an assertion. good catch(es)

warehouse/templates/manage/organization/settings.html
ewdurbin: I pondered this as well, and tried

```
diff --git a/warehouse/locale/Makefile b/warehouse/locale/Makefile
index 844ea9f9e..6f511c2f3 100644
--- a/warehouse/locale/Makefile
+++ b/warehouse/locale/Makefile
@@ -5,6 +5,7 @@ compile-pot:
        PYTHONPATH=$(PWD) pybabel extract \
                -F babel.cfg \
                --omit-header \
+                --add-comments="i18n" \
                --output="warehouse/locale/messages.pot" \
                warehouse
 
diff --git a/warehouse/templates/manage/organization/settings.html b/warehouse/templates/manage/organization/settings.html
index d383b2c72..41d0e90f4 100644
--- a/warehouse/templates/manage/organization/settings.html
+++ b/warehouse/templates/manage/organization/settings.html
@@ -225,11 +225,13 @@
         {% pluralize %}
           Your organization currently owns {{ count }} projects.
         {% endtrans %}
+        {# i18n: Continued with "Before you can delete your organization." after a list of requirements #}
         {% trans %}For each project, you must either:{% endtrans %}
         <ul>
           <li>{% trans %}Transfer ownership and remove the project from your organization{% endtrans %}</li>
           <li>{% trans %}Delete the project{% endtrans %}</li>
         </ul>
+        {# i18n: Continuation of a sentence "For each project, you must either:" before a list of requirements #}
         {% trans %}Before you can delete your organization.{% endtrans %}
       </p>
       <ul class="no-bottom-margin">
```

But the comments weren't picked up.

PR #13396, opened 2023-04-08
sending emails: handle case where we want to email confirming account deletion.

ewdurbin:
followup to a followup to #13391 .

Resolves https://python-software-foundation.sentry.io/issues/4072548346/

### DIFF
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -288,7 +288,8 @@ class TestSendEmailToUser:
 
 
 class TestSendEmail:
-    def test_send_email_success(self, db_session, monkeypatch):
+    @pytest.mark.parametrize("delete_user", [True, False])
+    def test_send_email_success(self, delete_user, db_session, monkeypatch):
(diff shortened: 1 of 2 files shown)


miketheman (commented):
I was just looking at this, and didn't understand why the `user_id` wasn't populating into the `success_event`. This seems like it was working correctly with the `UserService` before, right? 

Wrapping this _will_ likely work, but is it the correct fix? Where were previously deleted user events being recorded?

ewdurbin:
Relevant diffs:

https://github.com/pypi/warehouse/pull/13391/files#diff-43d0c472ea153e3fb6daafdedf32184357a81c6bce3ed3e8fc3069fc14ddc0c5L68-L75

https://github.com/pypi/warehouse/pull/13391/files#diff-4d7e23e1c19fbedae06550a287ef846d8f8efe27ee3c54f1a7ec3f4f8b9c5af6L568-L579

I'm not sure how we got new behavior.

miketheman (approved):

PR #13402, opened 2023-04-10
Make OIDC columns non-nullable

di:
Extracted from https://github.com/pypi/warehouse/pull/13272


ewdurbin (approved):
only a question, but if we are enforcing these in DB shouldn't we be at least attempting to enforce them in the application?

di:
They are all required fields for the forms that create these models. Is that what you mean?

ewdurbin:
> They are all required fields for the forms that create these models. Is that what you mean?

Yah, just curious what spurred this I guess. Did nulls trickle through?

di:
No, just being explicit, because we're about to add a nullable field to this model.

ewdurbin:
👍👍👍

PR #13411, opened 2023-04-11
[oidc] Unify publisher removal messages

di:
This was originally printing the `Publisher` repr for one message, which was including a lot of unnecessary information in the flash mesage. Instead, we unify the message to be the same for both pending and established publishers.

### DIFF
--- a/warehouse/manage/views/__init__.py
+++ b/warehouse/manage/views/__init__.py
@@ -1384,7 +1384,8 @@ class ManageOIDCPublisherViews:
             )
 
             self.request.session.flash(
-                f"Removed {publisher} from {self.project.name}", queue="success"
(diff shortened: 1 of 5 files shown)


woodruffw (approved):
LGTM!

warehouse/manage/views/__init__.py
```diff
@@ -1384,7 +1384,8 @@ def delete_oidc_publisher(self):
             )
 
             self.request.session.flash(
-                f"Removed {publisher} from {self.project.name}", queue="success"
+                f"Removed trusted publisher for project {self.project.name!r}",
```
woodruffw, line 1387: For my own benefit: why are we taking the `repr` of `self.project.name` here? Just to intelligently quote it?

PR #13417, opened 2023-04-12
oidc/forms: turn empty environments into None

woodruffw:
This fixes the immediate bug (which was caused by a comparison that began failing due to dual state between `''` and `None` columns), but should be further improved: the unique constraint between `GitHubProvider`s needs to be expanded to include the environment name as well.


woodruffw:
I haven't added any unit tests (since this shouldn't produce any branch changes, but I can also add some tests to ensure that the `normalized_environment` behavior is now as we expect.

di (approved):

PR #13419, opened 2023-04-12
[oidc] Handle verification of optional claims when not present

di:
Support for verifying the `environment` claim was added in https://github.com/pypi/warehouse/pull/13272, however currently if this claim is missing, verification fails even if the Publisher is not enforcing that claim, because we originally made the assumption that all claims would always be present.

Fixes https://python-software-foundation.sentry.io/share/issue/9dd92bdf158a44eea94104c259517124/

### DIFF
--- a/warehouse/oidc/models.py
+++ b/warehouse/oidc/models.py
@@ -185,7 +185,7 @@ class OIDCPublisherMixin:
 
         # Finally, perform the actual claim verification.
         for claim_name, check in self.__required_verifiable_claims__.items():
-            # All verifiable claims are mandatory. The absence of a missing
+            # All required claims are mandatory. The absence of a missing
             # claim *is* an error with the JWT, since it indicates a breaking
             # change in the JWT's payload.
             signed_claim = signed_claims.get(claim_name)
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
LGTM! 

For consistency we could rename `__verifiable_claims__` to `__required_verifiable_claims__`, but IMO that's pretty minor.

miketheman (approved):
LGTM

warehouse/oidc/models.py
```diff
@@ -180,8 +186,8 @@ def verify_claims(self, signed_claims: SignedClaims):
         # Finally, perform the actual claim verification.
         for claim_name, check in self.__verifiable_claims__.items():
             # All verifiable claims are mandatory. The absence of a missing
```
dstufft, line 188: The wording of this comment feels off, optional claims are also verifiable claims but are not mandatory right? I guess this is just using the name of the variable, but maybe that variable should be renamed too?

di: Updated in 882c3e4!

PR #13429, opened 2023-04-17
[oidc] More translated strings, and other fixups

di:
This PR started as an attempt to mark many strings used in OIDC-related views as translatable, which also added some strings in non-OIDC views (705f60bee7262e93b1e532ca7153801217f919d0).

This required updating many tests that weren't using `pyramid_request`/`db_request` and would have needed a stub localizer.

While rewriting the tests, I found a few that were doing a bit too much mocking/stubbing, so I refactored these to use our factories as much as possible.

Along the way, I found a few minor bugs in our OIDC implementation:
- `added_by_id` for pending publishers was nullable at the database level (e00e80008d32e13d789e30ec7cf03a64fb798004)
- `publisher_id` was not being validated as required in the `PendingGitHubPublisherForm` (94c6e4a5cc3a8460cd81d1c80dfda2ac10dc981a)
- We had a slightly different error message when the `DeletePublisherForm` form was invalid vs when the publisher ID was wrong, which could lead to this becoming an oracle (6e1b102fb097e669032c65ea837e2edd1b2a9995)
- We weren't flushing the DB after creating a pending publisher, so the ID in the recorded event was `None` (5da818e3a86157f10ea735970a6f73b8ae39ffa3)
- We weren't checking if a request to delete a pending publisher was being made by the user that created that publisher (2b0bbf21b6d389c0af28dd6f3608b211b37a0c3e)

I opted to just include the fixes for these in this PR rather than breaking out each into separate PRs because of the amount of changes I had already made to the tests, and because I'm lazy -- apologies to any reviewers.

### DIFF
--- a/tests/common/db/accounts.py
+++ b/tests/common/db/accounts.py
@@ -30,7 +30,6 @@ class UserFactory(WarehouseFactory):
     is_superuser = False
     is_moderator = False
     is_psf_staff = False
-    has_oidc_beta_access = True
     date_joined = factory.Faker(
         "date_time_between_dates",
         datetime_start=datetime.datetime(2005, 1, 1),
--- a/tests/common/db/accounts.py
+++ b/tests/common/db/accounts.py
@@ -30,6 +30,7 @@ class UserFactory(WarehouseFactory):
     is_superuser = False
     is_moderator = False
     is_psf_staff = False
+    has_oidc_beta_access = True
     date_joined = factory.Faker(
         "date_time_between_dates",
         datetime_start=datetime.datetime(2005, 1, 1),
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -1499,6 +1499,7 @@ class ManageAccountPublishingViews:
         )
 
         self.request.db.add(pending_publisher)
+        self.request.db.flush()  # To get the new ID
 
         self.request.user.record_event(
             tag=EventTag.Account.PendingOIDCPublisherAdded,
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -1566,6 +1566,13 @@ class ManageAccountPublishingViews:
         # futzes with the form.
         if pending_publisher is None:
             self.request.session.flash(
+                self.request._("Invalid publisher ID"),
+                queue="error",
+            )
+            return self.default_response
+
+        if pending_publisher.added_by != self.request.user:
+            self.request.session.flash(
                 "Invalid publisher ID",
                 queue="error",
             )
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -1550,46 +1550,50 @@ class ManageAccountPublishingViews:
 
         form = DeletePublisherForm(self.request.POST)
 
-        if form.validate():
-            pending_publisher = self.request.db.query(PendingOIDCPublisher).get(
-                form.publisher_id.data
-            )
-
-            # pending_publisher will be `None` here if someone manually
-            # futzes with the form.
-            if pending_publisher is None:
-                self.request.session.flash(
-                    "Invalid publisher for user",
-                    queue="error",
-                )
-                return self.default_response
-
+        if not form.validate():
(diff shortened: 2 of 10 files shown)


woodruffw:
> * We had a slightly different error message when the `DeletePublisherForm` form was invalid vs when the publisher ID was wrong, which could lead to this becoming an oracle ([6e1b102](https://github.com/pypi/warehouse/commit/6e1b102fb097e669032c65ea837e2edd1b2a9995))

No objection to this change, but IIRC isn't the plan to eventually surface trusted publishing status for each package as public information anyways? I remember one of the issues had a discussion around that, as part of helping guide users towards selecting packages with the strongest possible release security practices.

woodruffw (approved):
LGTM! One separate question about the oracle bug, but I don't think it's a blocker 🙂

warehouse/migrations/versions/75ba94852cd1_make_pendingoidcpublisher_added_by_id_.py
```diff
@@ -0,0 +1,42 @@
...
+down_revision = "f7cd7a943caa"
+
+
+def upgrade():
+    op.alter_column(
+        "pending_oidc_publishers",
+        "added_by_id",
+        existing_type=postgresql.UUID(),
+        nullable=False,
+    )
```
miketheman, line 33: **thought:** ‏This kind of `ALTER COLUMN` statement will check for validity immediately - is there some other process that guarantees all records are non-null before this migration runs?

warehouse/accounts/views.py
```diff
@@ -1550,46 +1550,50 @@ def delete_pending_oidc_publisher(self):
...
 
-        if form.validate():
-            pending_publisher = self.request.db.query(PendingOIDCPublisher).get(
-                form.publisher_id.data
+        if not form.validate():
+            self.request.session.flash(
+                "Invalid publisher ID",
+                queue="error",
             )
+            return self.default_response
```
miketheman: **praise:** ‏love early exits!

tests/common/db/accounts.py
```diff
@@ -30,6 +30,7 @@ class Meta:
     is_superuser = False
     is_moderator = False
     is_psf_staff = False
+    has_oidc_beta_access = True
```
miketheman, line 33: **thought:** ‏Looking at this pattern, it's unclear why these are necessary, since the defaults are false. Something to look into at some future testing and Factory behaviors effort.

miketheman (approved):
I got through most of this, and got a bit lost in the test changes, but I believe that changing to use non-stub methods surfaced the issues you've corrected, so that's awesome!

Some questions and thoughts inline, nothing blocking execution - but please note the migration question in particular.

warehouse/migrations/versions/75ba94852cd1_make_pendingoidcpublisher_added_by_id_.py
di: Great point: there is not, however it doesn't seem like it was previously possible to add a row with a null column here. I need to verify this in the database before merging.

di:
> No objection to this change, but IIRC isn't the plan to eventually surface trusted publishing status for each package as public information anyways? 

Correct, but the oracle would be for determining a valid publisher ID, not whether a file was published with/without OIDC.

woodruffw:
> Correct, but the oracle would be for determining a valid publisher ID, not whether a file was published with/without OIDC.

Gotcha, makes sense! I'm having trouble thinking of a scenario in which it's dangerous for a third party to know another user's pending publisher's ID, but removing any unnecessary oracle is always an improvement 🙂

di:
> I'm having trouble thinking of a scenario in which it's dangerous for a third party to know another user's pending publisher's ID

Combined with https://github.com/pypi/warehouse/commit/2b0bbf21b6d389c0af28dd6f3608b211b37a0c3e, I think it would have been technically possible to guess (via bruteforce) another user's publisher ID and then delete it. Not a huge issue to be fair, and mitigated with that change anyways.

PR #13460, opened 2023-04-20
[oidc] Make 'trusted publishers' generally available

di:
This takes the 'trusted publishers' feature out of beta. The only thing left to do is drop the `has_oidc_beta_access` column from the `Users` model, but will do that in a follow-up PR.

Fixes #12965.

PR #13461, opened 2023-04-20
[oidc] Remove `User.has_oidc_beta_access` column

di:
Follow-on to #13460.


warehouse/migrations/versions/d1771b942eb6_remove_user_has_oidc_beta_access_column.py
```diff
@@ -0,0 +1,45 @@
...
+
+from alembic import op
+
+revision = "d1771b942eb6"
+down_revision = "75ba94852cd1"
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute("SET statement_timeout = 120000")
```
miketheman, line 30: Do you expect a `DROP COLUMN` to take a long time? In a particularly busy table, I guess getting a lock might take a while, but is `users` updated that much?

miketheman (approved):
Non blocking comment. Good to go after the other PR is deployed fully.

warehouse/migrations/versions/d1771b942eb6_remove_user_has_oidc_beta_access_column.py
di: I don't, but I figure this doesn't hurt given recent lock contention.

PR #13474, opened 2023-04-21
fix: remove invalid unicode characters from xml

### LINKED ISSUE #13471: Invalid Unicode character in package name causing parsing error when trying to parse the RSS feeds
**Describe the bug**
When we tried to parse the [newest packages feed](https://warehouse.pypa.io/api-reference/feeds.html#newest-packages-feed) this morning Python raised an XML parsing exception because there was an invalid Unicode character in the project description of <https://pypi.org/project/dj-test-queries/>.

**Expected behavior**
The RSS feed to not include invalid Unicode characters.

**To Reproduce**
Publish a package with invalid Unicode characters in the description and then attempt to parse the newest packages feed.

**My Platform**
We're using https://pypi.org/project/aiohttp/ to read the feed and https://pypi.org/project/xmltodict/ to dump it to a `dict` for eaiser processing.

**Additional context**

miketheman:
Certain control characters are invalid in XML, so we should strip them out before adding them to an XML template.

Currently only targeting the `description` tag, as that's what was identified in the reported issue.

Fixes #13471

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -267,3 +267,20 @@ def test_format_author_email(meta_email, expected_name, expected_email):
     author_name, author_email = filters.format_author_email(meta_email)
     assert author_name == expected_name
     assert author_email == expected_email
+
+
+@pytest.mark.parametrize(
+    ("inp", "expected"),
+    [
+        ("foo", "foo"),  # no change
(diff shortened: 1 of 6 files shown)


miketheman:
cc @mistwire who wanted to see how this ended 😉

di (approved):

PR #13501, opened 2023-04-24
More circular import fixes

woodruffw:
WIP; still chasing these down.


miketheman:
@woodruffw I'm considering closing this since #13737 accomplished the "main" issue.

We should definitely reorg code to prevent circular imports, probably also with a detection/prevention mechanism.

woodruffw:
Sounds good; this fell off my stack and is pretty stale, so closing makes sense 🙂

PR #13511, opened 2023-04-25
chore: address sqlalchemy warnings

miketheman:
See commits for specifics.

Address ~500 warnings surfaced by setting `SQLALCHEMY_WARN_20=1` during the test suite.

There's still some 4,000 left, but I wanted to get these out.

Please review carefully!


miketheman:
> Just an idle thought: I bet we have a few instances of `db.query(Model).filter(Model.id == id).one()` that could be replaced with `db.get(Model, id)`

We absolutely do - I found a few in banner.py and sponsors.py - I might refactor those in the next round.

PR #13520, opened 2023-04-25
Guard against missing approval/decline event

di:
Fixes https://python-software-foundation.sentry.io/issues/4126563778/


di:
> This will solve the exception, but it does raise another question on whether we should be using events as a source of data with referential integrity like this.

Agreed. I don't think we should, but I'm currently just trying to do the minimal amount of work to make the admin UI usable.

PR #13527, opened 2023-04-26
Bugfix to match docs to behavior on necessary upload checksum field(s)

teruokun:
Currently the documentation mentions that only md5_digest is supported for the legacy upload API but according to [the validator](https://github.com/pypi/warehouse/blob/main/warehouse/forklift/legacy.py#L614) it accepts the valid SHA256 digest as a possible digest for uploaded objects.

There may be a follow-up PR that could include the new blake2b_256 checksum as a supported type given it is also a validated field, but that has enough other changes that this correction is different enough in scope to warrant separating


di (approved):

donatocayurin:
Thank you ❤️🙏

PR #13528, opened 2023-04-26
Add capability to use blake2_256_digest alone for validation and fix typo

teruokun:
This adds the ability for uploaders to use `blake2_256_digest` alone as a valid digest on publishing artifacts and fix a typo in the testing which was errantly passing due to the nature of `"sdist" != None` being logically equivalent to `"sdist" != "bdist_wheel"` which is the intended test.  Documentation updated to match.


teruokun:
Whoops, forgot to run the lint after my doc update.  Will be updating shortly

di:
Hey Jeff, thanks for the PR. Could you maybe add some detail about why uploaders would prefer to use `blake2_256_digest` instead?

Since our upload API is not standardized but sort of a de-facto standard, I'm generally hesitant to make changes here, but since this is a fully backwards-compatible change, I think we could add it if there was reasonable motivation for it.

teruokun:
Looked like an unintentional miss given the surrounding code, so it felt
like a reasonable low-impact first code change to warehouse.  I would
likely find it just as helpful to have a comment on why its support doesn't
match the other hash types if it was intentionally left out as I found it
while doing some diving prior to an attempt at PEP 658.  So I took it as
simply one of the many implementation/documentation gaps that I could also
use as a small but substantive change to test parts of my development setup
prior to potentially doing more substantial PRs in the future.

di (approved):
Thanks!

PR #13532, opened 2023-04-27
issue purge when projects are transferred to/between orgs

ewdurbin:


ewdurbin:
Thanks @nnja for figuring out how to force an object to purge even when it hasn't directly changed!

warehouse/organizations/services.py
```diff
@@ -396,6 +396,10 @@ def add_organization_project(self, organization_id, project_id):
         )
 
         self.db.add(organization_project)
+        self.db.flush() # Flush db so we can address the organization related object
```
chrysle, line 399:
```suggestion
        self.db.flush() # flush the db now so we can address the organization related object
```

The linter doesn't seem to recognize the comment. Maybe if it's more like in the example?

di (approved):

PR #13537, opened 2023-04-28
docs: internal docs for trusted publishers

### LINKED ISSUE #13470: Document why trusted publishing uses a "two phase" token exchange system
This is something that's come up (reasonably!) as a question from other package indices looking to integrate OIDC for trusted publishing.

woodruffw:
This adds an "Internals and Technical Details" page with both a high-level description of how trusted publishing works and a Q&A section with some pre-filled Q&As on the design of trusted publishing.

Closes #13470.

CC @di

### DIFF
--- a/docs/mkdocs-user-docs.yml
+++ b/docs/mkdocs-user-docs.yml
@@ -61,3 +61,4 @@ nav:
     - "trusted-publishers/using-a-publisher.md"
     - "trusted-publishers/security-model.md"
     - "trusted-publishers/troubleshooting.md"
+    - "trusted-publishers/internals.md"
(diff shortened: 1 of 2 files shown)


miketheman:
**question:** ‏seeing as these are internal docs, should they be placed in [a dev docs section](https://github.com/pypi/warehouse/tree/main/docs/dev/development) instead?

woodruffw:
> **question:** ‏seeing as these are internal docs, should they be placed in [a dev docs section](https://github.com/pypi/warehouse/tree/main/docs/dev/development) instead?

Possibly -- I wrote these more as "retrospective" details (for other package indices to benefit from) rather than concrete implementation details that would help a Warehouse contributor, but I have no strong opinion on where they actually belong :slightly_smiling_face:

makansissoko:
Le fonds extérieur

di (approved):
We've gotten enough questions about this from regular users that I'm OK with them living in the user docs.

PR #13540, opened 2023-04-30
Update using-a-publisher.md

tschm:
${oicd_token} comes already enclosed in " ". I have tested this in github.com/tschm/pyhrp
Your code as it stands won't work


Alexis4b:
Now what

Issue #13551, opened 2023-05-01
Trusted publishing: support for Google Cloud service accounts


woodruffw:
Support for these would allow PyPI users with Google Cloud-based publishing workflows to benefit from trusted publishing.

An example claim set from a Google Cloud service account, lightly anonymized:

```json
{
  "aud": "{SOME-AUDIENCE}",
  "azp": "{SOME-AZP}",
  "email": "{PROJECT-NUMBER}-compute@developer.gserviceaccount.com",
  "email_verified": true,
  "exp": 1682967007,
  "google": {
    "compute_engine": {
      "instance_creation_timestamp": 1674546966,
      "instance_id": "{INSTANCE-ID}",
      "instance_name": "dev",
      "project_id": "{PROJECT-ID}",
      "project_number": {PROJECT-NUMBER},
      "zone": "us-central1-a"
    }
  },
  "iat": 1682963407,
  "iss": "https://accounts.google.com",
  "sub": "{SOME-AZP}"
}
```

I've tried to keep the substitution names consistent above, to show where field values are duplicated.

Based on that claim set, it looks like the relevant uniquely identifying fields are:

1. `aud` (which should be `pypi`, similar to GitHub-issued JWTs)
2. `azp`: no idea what this is
3. `google.project_id`: presumably configured by a user
4. `google.project_number`: presumably a unique ID that prevents resurrection of `google.project_id`
5. `email`: presumably derivable consistently from `google.project_number`

So, my first educated guess is that we'll want to allow users to configure (3) and (4). Does that sound right to you @di?

---

- [x] Refactor `warehouse.oidc.models` to make adding new publisher models simpler (#13553)
- [x] Add OIDC provider models and services for Google Cloud
    - [x] Models: #13559
    - [x] Services https://github.com/pypi/warehouse/pull/13569
- [x] Support multiple publishers in the UI: https://github.com/pypi/warehouse/pull/13571
- [x] Expose Google Cloud providers through forms and views: https://github.com/pypi/warehouse/pull/15144
- [x] Update emails, etc. to be generic over OIDC provider kinds (xref #13667): https://github.com/pypi/warehouse/pull/13872
- [ ] Update user docs: https://github.com/pypi/warehouse/pull/15192

di:
> `azp`: no idea what this is

This is some reference to the user the token was issued to

> So, my first educated guess is that we'll want to allow users to configure (3) and (4).

[...]

PR #13553, opened 2023-05-01
warehouse, tests: devolve `oidc.models`

woodruffw:
This turns `oidc.models` into a directory, with `models.base` for the basic OIDC publisher classes and adjacent modules for each specific OIDC implementation (currently just GitHub).

This will make adding additional OIDC publishers easier and cleaner, as each will live in its own dedicated submodule.

See #13551.

CC @di


warehouse/oidc/models/base.py
woodruffw, line 1: Flagging for review: I named this `base` arbitrarily. It could also be `common` or similar, if there's more of a taste for that 🙂 

CC @miketheman in particular for opinions!

di: Fairly minor distinction but I think I'd prefer putting these in `warehouse/oidc/models/__init__.py` instead.

woodruffw: > Fairly minor distinction but I think I'd prefer putting these in `warehouse/oidc/models/__init__.py` instead.

Yep, I can do this -- the main reasons I didn't was to make the diff smaller and to make import cycles a little less likely (since with the bases in `__init__.py` there would be both "inwards" and "outwards" imports). 

If we're okay with that, though, I can move it!

woodruffw: Another idea: we could do what `warehouse.macaroons` does and put the core APIs in `_core.py`, which both avoids the circular import risk and is consistent with the rest of the codebase.

miketheman: Thanks for the tag! I'm also leaning towards `core` or something like it, since `base` to me denotes that these are likely to be subclassed or something.

@di we recently ran into circular import issues with keeping lotsa stuff in `__init__.py` files, since import tools traversing the package path will load these files, making it harder to reason about what's happening when.

di: Fair! We use `_core.py` for macaroons, let's do the same here.

woodruffw: Refactored, thanks both!

di (approved):

PR #13554, opened 2023-05-01
feat: add remote_addr_hashed to Request

miketheman:
As a precursor to implementing further IP Geo functionality, add a property to the `request` object as it flows through our system

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -105,6 +105,15 @@ def remote_addr():
 
 
 @pytest.fixture
+def remote_addr_hashed():
+    """
+    Static output of `hashlib.sha256(remote_addr.encode("utf8")).hexdigest()`
+    Created statically to prevent needing to calculate it every run.
+    """
+    return "6694f83c9f476da31f5df6bcc520034e7e57d421d247b9d34f49edbfc84a764c"
+
+
+@pytest.fixture
 def jinja():
     dir_name = os.path.join(os.path.dirname(warehouse.__file__))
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -104,6 +104,15 @@ def remote_addr():
     return "1.2.3.4"
 
 
+@pytest.fixture
+def remote_addr_hashed():
+    """
+    Static output of `hashlib.sha256(remote_addr.encode("utf8")).hexdigest()`
+    Created statically to prevent needing to calculate it every run.
+    """
+    return "6694f83c9f476da31f5df6bcc520034e7e57d421d247b9d34f49edbfc84a764c"
```
miketheman, line 113: **thought:** ‏it's not that much of a time-suck, might be able to run the real thing?

```shellsession
$ python -m timeit -c 'import hashlib;hashlib.sha256("1.2.3.4".encode("utf8")).hexdigest()'
500000 loops, best of 5: 758 nsec per loop
```

miketheman: I decided against it, since there's a test that manually sets `remote_addr` to `None` for some reason. I'll look into that is and when the test is re-examined for hashed values.

warehouse/utils/wsgi.py
```diff
@@ -43,11 +50,14 @@ def __call__(self, environ, start_response):
             remote_addr = _forwarded_value(
                 environ.get("HTTP_X_FORWARDED_FOR", ""), self.num_proxies
             )
+            remote_addr_hashed = ""
```
di, line 53: Should we be hashing `HTTP_X_FORWARDED_FOR` here instead?

miketheman: Good idea, will add!

miketheman: Added in f8d8061

di (approved):

PR #13559, opened 2023-05-02
oidc/models: add initial Google models

woodruffw:
~~WIP.~~

This adds `GooglePublisher` and `PendingGooglePublisher`, mirroring the GitHub models.

See https://github.com/pypi/warehouse/issues/13551.

### DIFF
--- a/warehouse/oidc/models/_core.py
+++ b/warehouse/oidc/models/_core.py
@@ -30,13 +30,27 @@ def _check_claim_binary(binary_func):
     ignoring the third.
 
     This is used solely to make claim verification compatible with "trivial"
-    checks like `str.__eq__`.
+    comparison checks like `str.__eq__`.
     """
 
     def wrapper(ground_truth, signed_claim, all_signed_claims):
         return binary_func(ground_truth, signed_claim)
 
     return wrapper
+
+
+def _check_invariant(value: Any):
+    """
+    Wraps a fixed value comparison into a three-argument function.
+
+    This is used solely to make claim verification compatible with "invariant"
+    comparison checks, like "claim x is always the literal `true` value".
+    """
+
+    def wrapper(ground_truth, signed_claim, all_signed_claims):
+        return value == signed_claim
+
+    return wrapper()
 
 
 class OIDCPublisherProjectAssociation(db.Model):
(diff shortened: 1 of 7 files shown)


woodruffw:
Still needs test coverage, but the overall structure here is good for review.

CC @di 🙂

warehouse/oidc/models/google.py
```diff
@@ -0,0 +1,98 @@
...
+    Common functionality for both pending and concrete Google OIDC
+    providers.
+    """
+
+    email = Column(String, nullable=False)
+    sub = Column(String, nullable=False)
+
+    __required_verifiable_claims__ = {
+        "email": _check_claim_binary(str.__eq__),
+        "email_verified": _check_claim_invariant(True),
```
woodruffw, line 36: Flagging for review: `email_verified` should always be `true`, so this just enforces that.

warehouse/oidc/models/google.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class GooglePublisherMixin:
+    """
+    Common functionality for both pending and concrete Google OIDC
+    providers.
+    """
+
+    email = Column(String, nullable=False)
+    sub = Column(String, nullable=False)
```
di, line 32: I'm thinking this might be `nullable`: https://github.com/pypi/warehouse/issues/13551#issuecomment-1531959766

di:
LGTM aside from that one comment!

warehouse/oidc/models/google.py
woodruffw: Sounds good! Will tweak.

woodruffw:
This should be fully good to go now!

warehouse/oidc/models/_core.py
```diff
@@ -39,6 +39,20 @@ def wrapper(ground_truth, signed_claim, all_signed_claims):
...
+def _check_claim_invariant(value: Any):
+    """
+    Wraps a fixed value comparison into a three-argument function.
+
+    This is used solely to make claim verification compatible with "invariant"
+    comparison checks, like "claim x is always the literal `true` value".
+    """
+
+    def wrapper(ground_truth, signed_claim, all_signed_claims):
+        return value == signed_claim
```
di, line 51: This is a little confusing because we need to define the `@property` for `ground_truth` to be present as an argument here, but then we don't actually use it when verifying the invariant.

I think this would make more sense if this was:

```suggestion
    def wrapper(ground_truth, signed_claim, all_signed_claims):
        return ground_truth == signed_claim == value
```

woodruffw: Yeah, thanks for catching that! I agree completely.

di (approved):

PR #13564, opened 2023-05-02
fix: update source sans font to latest

### LINKED ISSUE #8634: Italic font gets a fallback in Cyrillic texts

miketheman:
The 3.x series of Source Sans was released in 2020, and contains the desired italicized glyphs.
It was added to Google Fonts in 2021 in google/fonts#3856

The author dropped "Pro" from the name, and replaced with "3".

See: https://blog.adobe.com/en/publish/2020/11/30/whats-new-in-source-sans-3

Fixes #8634

### DIFF
--- a/warehouse/static/sass/settings/_fonts.scss
+++ b/warehouse/static/sass/settings/_fonts.scss
@@ -14,7 +14,7 @@
 
 // SETTINGS - FONTS
 
-$base-font-family: "Source Sans Pro", "Helvetica", arial, sans-serif;
+$base-font-family: "Source Sans 3", "Helvetica", arial, sans-serif;
 $code-font: "Source Code Pro", monospace;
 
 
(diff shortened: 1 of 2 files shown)


miketheman:
Here's a screenshot of the font in use with the help text discussed in the related Issue:

Screenshot 2023-05-02 at 17 41 25


cc @webknjaz for visual review

di (approved):

PR #13566, opened 2023-05-03
Fix `find_publisher_by_issuer` environment filter

di:
Previously, we weren't adequately filtering on `environment` in `find_publisher_by_issuer`, which resulted in this function raising a `MultipleResultsFound` exception if more than one publisher was configured that matched the claimset.

Instead, the behavior should be as follows:
* If no `environment` claim is present
  * If there's a publisher that doesn't restrict on `environment`, return it
  * or return `None`
* If an `environment` claim is present
  * If there's a publisher that restricts on this `environment`, return it
  * If there's a publisher that doesn't restrict on `environment`, return it
  * or return `None`

Fixes https://python-software-foundation.sentry.io/issues/4150013748/.

### DIFF
--- a/tests/unit/oidc/test_utils.py
+++ b/tests/unit/oidc/test_utils.py
@@ -12,10 +12,9 @@
 
 import pretend
 
-from sqlalchemy.sql.expression import func, literal
 
 from warehouse.oidc import utils
-from warehouse.oidc.models import GitHubPublisher
+from tests.common.db.oidc import GitHubPublisherFactory
 
 
 def test_find_publisher_by_issuer_bad_issuer_url():
@@ -27,14 +26,14 @@ def test_find_publisher_by_issuer_bad_issuer_url():
     )
 
 
-def test_find_publisher_by_issuer_github():
-    publisher = pretend.stub()
-    one_or_none = pretend.call_recorder(lambda: publisher)
-    filter_ = pretend.call_recorder(lambda *a: pretend.stub(one_or_none=one_or_none))
-    filter_by = pretend.call_recorder(lambda **kw: pretend.stub(filter=filter_))
-    session = pretend.stub(
-        query=pretend.call_recorder(lambda cls: pretend.stub(filter_by=filter_by))
+def test_find_publisher_by_issuer_github(db_request):
+    publisher = GitHubPublisherFactory(
+        repository_owner="foo",
+        repository_name="bar",
+        repository_owner_id="1234",
+        workflow_filename="ci.yml",
     )
+
     signed_claims = {
         "repository": "foo/bar",
         "job_workflow_ref": "foo/bar/.github/workflows/ci.yml@refs/heads/main",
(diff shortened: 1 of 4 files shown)


tests/unit/oidc/test_utils.py
```diff
@@ -27,46 +28,50 @@ def test_find_publisher_by_issuer_bad_issuer_url():
...
+)
+def test_find_publisher_by_issuer_github(db_request, environment, expected_id):
+    GitHubPublisherFactory(
+        id="aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
+        repository_owner="foo",
+        repository_name="bar",
+        repository_owner_id="1234",
+        workflow_filename="ci.yml",
+        environment=None,
+        # No environment
```
miketheman, line 47: nitpick: match style
```suggestion
        environment=None,  # No environment
```

tests/unit/oidc/test_utils.py
```diff
@@ -27,46 +28,50 @@ def test_find_publisher_by_issuer_bad_issuer_url():
...
-        .compare(
-            literal("ci.yml@refs/heads/main").like(
-                func.concat(GitHubPublisher.workflow_filename, "%")
-            )
-        )
-    )
 
-    assert one_or_none.calls == [pretend.call()]
+# def test_find_publisher_by_issuer_with_environment_finds_match(db_request)
+# def test_find_publisher_by_issuer_with_environment_finds_wildcard(db_request)
```
miketheman, line 77: **question:** ‏leftover aspirational tests?

warehouse/oidc/utils.py
```diff
@@ -47,20 +48,42 @@ def find_publisher_by_issuer(session, issuer_url, signed_claims, *, pending=Fals
...
-                repository_owner_id=signed_claims["repository_owner_id"],
+        # Try finding a publisher that matches the provided environment first.
+        try:
+            return (
+                session.query(publisher_cls)
+                .filter_by(
+                    repository_name=repository_name,
+                    repository_owner=repository_owner,
+                    repository_owner_id=signed_claims["repository_owner_id"],
+                    environment=signed_claims.get("environment"),
```
miketheman, line 59: **question:** It looks like `signed_claims` could come in without an `environment` key.

By my reading, the first query in the `try` block would run a `filter_by(..., environment=None` and try to get `one()`, and if nothing comes back, issue a second **identical** query with `one_or_none()`.


My question is:

- If there's 2 claim records, one with an environment, and one without, the first query will get the env-specific one. Yay, that's the intent.
- If there's 1 claim record, with **no environment**, why would the second query get run? Wouldn't the `filter_by` condition apply just the same?

woodruffw: > **question:** It looks like `signed_claims` could come in without an `environment` key.

Yep, that's correct: GitHub's OIDC JWTs don't contain an `environment` claim if one isn't explicitly configured.

I think the confusing bit here is the possible states:

1. OIDC token with an `environment`
2. OIDC token without an `environment`
3. Trusted publisher with an `environment` configured
4. Trusted publisher without an `environment` configured

State (3) **must** only match state (1), while state (4) **can** match either (1) or (2). So we _need_ to explicitly carve out an `environment=None` case.

(This would have been nicer if we'd added the `environment` claim without broadening the uniqueness constraint to include it, but that would have made it impossible to register separate trusted publishers for different environments under the same workflow...)

woodruffw: Oh, but I see what you mean -- I think the queries are slightly off here: the first should ensure that `signed_claims["environment"]` is actually present, while the second should remain explicitly `environment=None`.

tests/unit/oidc/test_utils.py
di:
```suggestion
```

warehouse/oidc/utils.py
di: > If there's 1 claim record, with no environment, why would the second query get run? Wouldn't the `filter_by` condition apply just the same?

The second query is to capture the case where the signed claims have an `environment` key, but a publisher is configured with `environment=None` -- essentially, we can always fall back on this "wildcard" publisher no matter what, if it's present, if there isn't a publisher configured with a matching `environment`.

I think we need two queries regardless: one to check if there's a publisher that matches the environment, and one to check for a "wildcard" publisher. Using `signed_claims.get("environment")` in the first query here allows us to satisfy the first case and also short-circuit and only run one query when there is no environment in the signed claims at all. If we didn't do that, we'd have to add some more branching here, like:

I agree this is a little confusing though, let me see if I can make this more clear with some conditionals instead.

di: 9cf3badab8e7244c4688b2df6f06b4d4c8650049

miketheman (approved):
Thanks for making the logic simpler, based on the environment existing.
Not we should only ever make a single DB call for the publisher.

I'm betting this query could be refactored a little more to construct the query body, and conditionally add the specifics to switch the `environment` flag, but that's fine to defer to another time.

PR #13569, opened 2023-05-03
oidc: Google JWKS/services scaffolding

woodruffw:
~~WIP.~~

This also needs some more changes, but I'll wait for #13566 to make them.

See #13551.


warehouse/oidc/utils.py
woodruffw, line 1: The diff is a little confusing here, but what I've done is break `find_publisher_by_issuer` into two sub-helpers: `_find_github_publisher` and `_find_google_publisher`.

warehouse/oidc/utils.py
```diff
@@ -40,54 +47,84 @@ def find_publisher_by_issuer(session, issuer_url, signed_claims, *, pending=Fals
...
+
+def _find_google_publisher(session, signed_claims, *, pending=False):
+    publisher_cls = GooglePublisher if not pending else PendingGooglePublisher
 
+    # The `sub` claim should always be present in the claim set, but is
+    # optional on the registered publisher side. We first check for a
+    # publisher that has both, and then for one with no `sub`.
+    # This has to happen as separate queries to prevent duplicate
+    # results: both variants can be registered, and we always want
+    # to return the more specialized one if possible.
```
woodruffw, line 116: Flagging for review: we have to do this double-query dance for the same reason that GitHub publishers need it (an optional, unique claim, in this case `sub`).

Since this is a brand new publisher rather than one we're augmenting, we could instead preempt this problem: we could make the unique constraint on `GooglePublisher`s _just_ on the `email` claim, meaning that it wouldn't be possible to have two unique publishers that only vary on `sub`.  This simplifies the code here, at the cost of _potentially_ (I'm not sure how Google's `sub` claim maps to identity) excluding some valid unique combinations of `(email, sub)`.

Thoughts @di?

woodruffw:
Needs tests and other work, but the basic structure here is good for an initial review!

warehouse/oidc/utils.py
di: I _think_ the only time we'd have the same `email` with a different `sub` claim is if a service account was deleted & recreated, and a new publisher was added. That probably means that the old `sub` claim will never occur after that point, so this would probably be fine, however I think the UX here would be a little confusing, because a user would consider these separate publishers. So I think the double-query is an OK tradeoff for that.

woodruffw: Makes sense! Resolving.

di:
This is getting a little bit unwieldy and duplicative -- given that we're likely going to add a third issuer soon, I think we should find a way to simplify this.

One thought is that we could a) have a lookup table that maps from (issuer, pending) -> `OIDCPublisher` b) store the filter expressions on the respective `OIDCPublisher` instead. So the function would be something like:

```python
publisher_classes = {
    GITHUB_OIDC_ISSUER_URL: [GitHubPublisher, PendingGitHubPublisher],
    GOOGLE_OIDC_ISSUER_URL: [GooglePublisher, PendingGooglePublisher],
}

def find_publisher_by_issuer(session, issuer_url, signed_claims, *, pending=False):
    try:    
        publisher_cls = publisher_classes[issuer_url][pending]
    except KeyError:
        return None
        
    # Check for a publisher with options
    publisher =  (
        session.query(publisher_cls)
        .filter_by(publisher_cls.filter_expr_with_options(signed_claims))
        .one_or_none()
    )
    # Check for a fallback publisher
    return (
        session.query(publisher_cls)
        .filter_by(publisher_cls.filter_expr_without_options(signed_claims))
        .one_or_none()
    )
```

woodruffw:
That sounds good to me! The class mapping part is certainly doable; storing the expressions might be slightly trickier (since e.g. the `GitHub` one contains an expression that gets built from a string match). 

Do you want me to look into that with this PR, or hold off until the next round of publishers (e.g. GitLab)?

woodruffw:
> I think I'd like to try to do it here, maybe take a crack at it and see how much of a time suck it'll be?

Let's find out 🙂

woodruffw:
One thought: how should this work if/when we have publishers with multiple optional claims? It should be fine if we don't include multiple optionals in the unique constraint set, but if we do it'll probably result in `N!` stored expressions for `N` possible optional claims.

di:
Maybe premature optimization? But I think a list of N!+1 increasingly loosened filter expressions would also satisfy the current two-query case as well, so might be worth doing. The alternative would be writing N!+1 queries, so I don't think there's any way around the general behavior here.

woodruffw:
Actually, thinking about it some more: these don't even need to be written out manually (which would be error-prone), since they're expressible via the existing `__required_verifiable_claims__` and `__optional_verifiable_claims__` on each publisher, whether pending or not!

I'll play around with that a bit 🙂

woodruffw:
[059b1e6](https://github.com/pypi/warehouse/pull/13569/commits/059b1e6d9c8612cefd639a423884fc6a97472ba5) is my attempt at this: publisher model is now devolved back to each publisher (or more accurately, the "base" mixin between pending and non-pending publishers of each type).

Each publisher type now supplies `__lookup_strategies__`, which provides a list of callables that return queries.

A bit of nastiness happens with these callables: conceptually they're `@classmethod`s, but class methods aren't actually callable (because they're indeterminate until their surrounding class is finalized), so they're _actually_ static methods that happen to take a class object as their first argument. I've declared them inside their relevant classes for the sake of locality, but this could be resolved by putting them at the module scope instead.

warehouse/oidc/utils.py
woodruffw: This is no longer relevant!

warehouse/oidc/models/github.py
```diff
@@ -125,6 +128,56 @@ class GitHubPublisherMixin:
         "enterprise",
     }
 
+    @staticmethod
+    def __lookup_all__(klass, signed_claims: SignedClaims) -> Query | None:
```
woodruffw, line 132: Flagging for review: this exemplifies the "nastiness" I was talking about: this is conceptually a `classmethod`, but it's registered as a `staticmethod` because it can't be made callable until its surrounding class is finalized.

woodruffw:
This should be good to go again.

warehouse/oidc/models/google.py
```diff
@@ -54,6 +56,21 @@ class GooglePublisherMixin:
...
+        )
+
+    @staticmethod
+    def __lookup_no_sub__(klass, signed_claims: SignedClaims) -> Query | None:
+        return Query(klass).filter_by(email=signed_claims["email"], sub=None)
+
+    __lookup_strategies__ = [
+        __lookup_all__,
+        __lookup_no_sub__,
+    ]
```
di, line 72: So the idea here is that the `email` claim is required and the `sub` claim is optional? Makes sense to me I think.

woodruffw: Yep, exactly 🙂

di (approved):

PR #13571, opened 2023-05-03
Add a horizontal tab picker

di:
This will be required as we add more OIDC providers. Publishing as draft for now to address any JS/CSS issues before plugging in forms for the additional provider (looking at you, @miketheman!)

May-03-2023 19-44-21

### DIFF
--- a/warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
+++ b/warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
@@ -1,0 +1,58 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["tab", "tabPanel"];
+
+  connect() {
+  }
+
+  initialize() {
+    this.showTab();
+  }
+
+  change(e) {
+    this.index = this.tabTargets.indexOf(e.target);
+    this.showTab(this.index);
+  }
+
+  showTab() {
+    this.tabPanelTargets.forEach((el, i) => {
+      if (i == this.index) {
+        el.classList.remove("is-hidden");
+      } else {
+        el.classList.add("is-hidden");
+      }
+    });
+
+    this.tabTargets.forEach((el, i) => {
+      if (i == this.index) {
+        el.classList.add("is-active");
+      } else {
+        el.classList.remove("is-active");
+      }
+    });
+  }
+
+  get index() {
+    return parseInt(this.data.get("index"));
+  }
+
+  set index(value) {
+    this.data.set("index", value);
+    this.showTab();
+  }
+}
--- a/warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
+++ b/warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
@@ -17,8 +17,6 @@ import { Controller } from "@hotwired/stimulus";
 export default class extends Controller {
   static targets = ["tab", "tabPanel"];
 
-  connect() {
-  }
 
   initialize() {
     this.showTab();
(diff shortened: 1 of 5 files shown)


warehouse/templates/manage/account/publishing.html
```diff
@@ -113,100 +209,17 @@ <h2>{% trans %}Add a new pending publisher{% endtrans %}</h2>
...
-      <p class="form-group__help-text">
-      {% trans %}The name of the GitHub repository that contains the publishing workflow{% endtrans %}
-      </p>
-      <div id="repository-errors">
-        {{ field_errors(pending_github_publisher_form.repository) }}
-      </div>
+  <div class="horizontal-tabs" data-controller="horizontal-tabs" data-horizontal-tabs-index="0">
+    <div class="tabbar">
+      <button data-target="horizontal-tabs.tab" data-action="horizontal-tabs#change" class="tab is-active">GitHub</button>
+      <button data-target="horizontal-tabs.tab" data-action="horizontal-tabs#change" class="tab">Google</button>
```
miketheman, line 215: Deprecation warnings in Console: `The data-target attribute is deprecated and will be removed in a future version of Stimulus.`

```suggestion
      <button data-horizontal-tabs-target="tab" data-action="horizontal-tabs#change" class="tab is-active">GitHub</button>
      <button data-horizontal-tabs-target="tab" data-action="horizontal-tabs#change" class="tab">Google</button>
```

warehouse/templates/manage/account/publishing.html
```diff
@@ -113,100 +209,17 @@ <h2>{% trans %}Add a new pending publisher{% endtrans %}</h2>
...
-        dedicated publishing environment is <strong>strongly</strong>
-        encouraged, <strong>especially</strong> if your repository has
-        maintainers with commit access who shouldn't have PyPI publishing
-        access.
-        {% endtrans %}
-      </p>
-      <div id="environment-errors">
-        {{ field_errors(pending_github_publisher_form.environment) }}
-      </div>
+    <div class="tabcontent is-hidden" data-target="horizontal-tabs.tabPanel">
```
miketheman, line 217:
```suggestion
    <div class="tabcontent is-hidden" data-horizontal-tabs-target="tabPanel">
```

warehouse/templates/manage/account/publishing.html
```diff
@@ -113,100 +209,17 @@ <h2>{% trans %}Add a new pending publisher{% endtrans %}</h2>
...
+      {{ github_form(request, pending_github_publisher_form) }}
     </div>
-    <div>
-      <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary {{ "button--disabled" if not pending_oidc_enabled else "" }}">
+    <div class="tabcontent" data-target="horizontal-tabs.tabPanel">
```
miketheman, line 220:
```suggestion
    <div class="tabcontent" data-horizontal-tabs-target="tabPanel">
```

warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
```diff
@@ -0,0 +1,58 @@
...
+        el.classList.add("is-active");
+      } else {
+        el.classList.remove("is-active");
+      }
+    });
+  }
+
+  get index() {
+    return parseInt(this.data.get("index"));
+  }
```
miketheman, line 52: **thought:** ‏this approach uses an index - where we're relying on the correct ordering of two elements to be always correct (which if we do it right, they should be).

One common way is to use matching `id` values on the buttons and the content, and thus match them up, like so:

```html
<button id="github" ...>
...
<div id="github" class="tabcontent ... >
```

And then you can match lookups by ID instead of index, and activate whichever one the user clicks - and not have to mutate the index value in the HTML.

miketheman (commented):
Generally looks good, a few notes inline on syntax, and one on the approach.

warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["tab", "tabPanel"];
+
+  connect() {
+  }
```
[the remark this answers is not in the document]
di:
```suggestion
```

warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
di: My understanding is that element IDs should be unique within the DOM, so I'm not sure this approach would be valid without adding some kind of predictable prefix/suffix to the IDs.

I think the risk of having buttons/tabs out of order is probably small, so I'm inclined to leave this as-is.

di:
This should be good to go. This will effectively be a no-op change from a user perspective until we add additional publishers.

miketheman (approved):

PR #13594, opened 2023-05-05
add hash and geoip_info to IpAddress model, expose on request

ewdurbin:
Adds `hased_ip_address` and a JSONB `geoip_info` to the `IpAddress` model.

Populates them as necessary and adds the `IpAddress` object to db session when `request.ip_address` is called. This stops us from needlessly querying the DB unless the codepath *needs* information from the object.


warehouse/utils/wsgi.py
```diff
@@ -16,10 +16,24 @@
...
     from pyramid.request import Request
 
 
+GEOIP_FIELDS = {
+    "continent_code": "CONTINENT",
+    "country_code": "COUNTRY_CODE",
+    "country_code3": "COUNTRY_CODE_3",
+    "country_name": "COUNTRY",
+    "region": "REGION",
+    "city": "CITY",
```
miketheman, line 33: **thought:** ‏we've munged these when setting the headers from upstream, and we're now un-munging them. Since we haven't started to use them yet, maybe this is the time to fix them upstream, and thus preserve the same names throughout? https://github.com/pypi/infra/blob/f909316bf75e10ecafad8d558e0859c3e12d623a/terraform/warehouse/vcl/main.vcl#L186-L191

warehouse/utils/wsgi.py
```diff
@@ -104,8 +132,31 @@ def _remote_addr_hashed(request: Request) -> str:
     return request.environ.get("REMOTE_ADDR_HASHED", "")
 
 
+def _ip_address(request):
+    """Return the IpAddress object for the remote address from the environment."""
+    try:
+        ip_address = (
+            request.db.query(IpAddress)
+            .filter(IpAddress.ip_address == request.remote_addr)
```
miketheman, line 140: We could write a simpler query:

```suggestion
            .filter_by(ip_address=request.remote_addr)
```

Both generate the same SQL query under the hood, just a matter of what looks more pleasing.

Same comment applies to queries in the test code.

warehouse/utils/wsgi.py
```diff
@@ -104,8 +132,31 @@ def _remote_addr_hashed(request: Request) -> str:
...
+    except NoResultFound:
+        ip_address = IpAddress(ip_address=request.remote_addr)
+        request.db.add(ip_address)
+
+    ip_address.hashed_ip_address = request.remote_addr_hashed
+    ip_address.geoip_info = {
+        k: request.environ[f"GEOIP_{v}"]
+        for k, v in GEOIP_FIELDS.items()
+        if f"GEOIP_{v}" in request.environ
+    }
```
miketheman, line 152: **question:** ‏How and when do these values get persisted to the db?

tests/unit/utils/test_wsgi.py
```diff
@@ -168,6 +176,39 @@ def test_passes_through_headers(self):
         assert app.calls == [pretend.call({"HTTP_X_FOOBAR": "wat"}, start_response)]
 
 
+def test_ip_address_exists(db_request):
```
miketheman, line 179:
```suggestion
def test_ip_address_exists(db_request, remote_addr_hashed):
```

tests/unit/utils/test_wsgi.py
```diff
@@ -168,6 +176,39 @@ def test_passes_through_headers(self):
...
+
+    ip_address = (
+        db_request.db.query(IpAddress)
+        .filter(IpAddress.ip_address == "192.0.2.69")
+        .one()
+    )
+    assert ip_address.ip_address == "192.0.2.69"
+    assert (
+        ip_address.hashed_ip_address
+        == "6694f83c9f476da31f5df6bcc520034e7e57d421d247b9d34f49edbfc84a764c"
```
miketheman, line 207:
```suggestion
        ip_address.hashed_ip_address == remote_addr_hashed
```
This has value is of `remote_addr` since that's being put into the test request over here: https://github.com/pypi/warehouse/blob/352fec979299c1d64551ca4b2c1622e896e2adb7/tests/conftest.py#L180

So there's something amiss about this specific test, since it's comparing a hash that isn't for the locally set IP address.

Likely needs an entry after `db_request.environ["REMOTE_ADDR"] =` to set the hashed value.

Alternately, replace `db_request` with `db_session` fixture, which has less request scaffolding applied if you can?

warehouse/utils/wsgi.py
ewdurbin: at the end of the request, if everything was successful. same story as last time you asked, the transaction manager picks up things added to the session (see line 145).

warehouse/utils/wsgi.py
ewdurbin: hm? I don't quite follow what you're suggesting we do.

miketheman: Following this: https://github.com/pypi/infra/pull/133 If that's merged and deployed, then we update these inbounds to match the headers exactly.

warehouse/ip_addresses/models.py
```diff
@@ -39,6 +48,9 @@ def __repr__(self):
         return self.ip_address
 
     ip_address = Column(INET, nullable=False, unique=True)
+    hashed_ip_address = Column(Text, nullable=True, unique=True)
+    geoip_info = Column(JSONB, nullable=True)
```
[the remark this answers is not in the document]
ewdurbin: https://github.com/pypi/warehouse/pull/13594/commits/d9357aaf68aed31877e9c717849dc72334599860

warehouse/utils/wsgi.py
ewdurbin: https://github.com/pypi/warehouse/pull/13594/commits/d9357aaf68aed31877e9c717849dc72334599860

tests/unit/utils/test_wsgi.py
ewdurbin: https://github.com/pypi/warehouse/pull/13594/commits/d9357aaf68aed31877e9c717849dc72334599860

tests/unit/utils/test_wsgi.py
ewdurbin: https://github.com/pypi/warehouse/pull/13594/commits/d9357aaf68aed31877e9c717849dc72334599860

miketheman (approved):

miketheman:
Needs an alembic rebase

Issue #8158, opened 2020-06-24
Geolocate user IP addresses when presenting them in UI


di:
**What's the problem this feature will solve?**
Currently in the PyPI logged-in UI, we show the IP address that performed certain actions to the user:

<img width="917" alt="Screen Shot 2020-06-24 at 6 37 55 PM" src="https://user-images.githubusercontent.com/294415/85638026-71b8e800-b64a-11ea-9487-87eb6108e8a8.png">

I don't know my own IP offhand. Especially if there are multiple different IPs listed here, I would need to manually look up the approximate location where these came from to get an idea of whether they were actually me or not.

**Describe the solution you'd like**
It would be nice if PyPI also showed me an (approximate) location for any given IP address as well, so I could easily visually filter ones that seem incorrect, e.g.:

| Event | Date / time | IP address |
| - | - |-|
| Logged in | less than 10 seconds ago | 11.22.11.22 (Austin TX USA)
| Logged in | June 22, 2020 | 22.33.22.33 (Austin TX USA)
| Logged in | June 19, 2020 | 44.55.44.55 (Timbuktu, Mali)
| Logged in | June 19, 2020 | 66.77.66.77 (Austin TX, USA)

**Additional context**
This shouldn't require external API calls. Using something like https://pypi.org/project/geoip2/ with an embedded database like https://dev.maxmind.com/geoip/geoip2/geolite2/ would probably work.

Ideally this would be determined on the fly and not stored anywhere (e.g. along with the IP address), so if we someday replaced the mechanism with something more precise (or just updated the embedded DB) the updates would be immediately reflected.

**Todo list**
- [x] Use [Fastly's geolocation services](https://developer.fastly.com/reference/vcl/variables/geolocation/) to determine geographic location at edge
- [x] Hash & salt IP addresses at edge, pass those to our backends/logs (populate a X-PyPI-Hashed-IP header.
- [x] Replace IP addresses in gunicorn logs
- [ ] Begin storing hashed IPs everywhere for all events
- [ ] Replace IP addresses in the user-facing UI (user/project events) with corresponding geolocation data
- [ ] Replace IP addresses in journals with corresponding hashed IP
- [ ] For all events tables, change IP columns to CIText, retroactively hash IP addresses (geoIP data will be missing)
- [ ] For Admin IP addresses table, data migration to retroactively hash the IP addresses.
- [ ] Drop X-Fastly-IP header at edge

sanjaysiddhanti:
[...]

PR #6339, opened 2019-08-01
Audit logging

woodruffw:
Adds `UserEvent` and `ProjectEvent` models, as well as corresponding helpers for recording events.

A shortlist of loggable events (not all of which need to be part of this PR):

Users: 

- [x] Login (+ whether a 2FA method was used)
- [x] Token creation/deletion
- [x] 2FA method creation/deletion
- [x] Email addition/removal/verification/deverification
- [x] Password change

Projects:

- [x] Version upload/removal
- [x] Change in owner/maintainers
- [x] Token creation by a user with the project in scope

TODO:

- [x] Add `assert`s to `record_event` calls in unit tests, where appropriate.
- [x] Restrict retrieval of user events to a recent timespan (last two weeks?)
- [x] Allow Warehouse admins to view the events for a user/project

Summary of events added:

Users:
* `account:email:add` (New email added to account)
* `account:email:remove` (Email removed from account)
* `account:email:primary:change` (Primary email changed or added)
* `account:email:reverify` (Email reverified)
* `account:password:change` (Account password changed)
* `account:two_factor:method_added` (Two factor method added)
* `account:two_factor:method_removed` (Two factor method removed)
* `account:api_token:added` (API token added)
* `account:api_token:removed` (API token removed)
* `account:create` (Account created)
* `account:password:reset:request` (Password reset request sent)
* `account:password:reset` (Password reset request completed)
* `account:email:verified` (Email address verified)
* `account:login:success` (Successful login)

Projects:
* `project:create` (Project created)
* `project:release:added` (New project release added)
* `project:release:removed` (Project release removed)
* `project:release:file:removed` (Project release file removed)
* `project:api_token:added` (API token added w/ project in scope)
* `project:api_token:removed` (API token removed w/ project in scope)
* `project:role:add` (New role added to project)
* `project:role:delete` (Role removed from project)
* `project:role:change` (Role changed in project)

cc @brainwane @ewdurbin @di @dstufft @nlhkabu 

Closes #5863.

PR #13595, opened 2023-05-05
refactor: store hashed remote address on journals

miketheman:
Another step towards #8158 .

Starts storing only hashed IP addresses in `JournalEntry.submitted_from` text column from merge-time forward for any new Journals.

Provides a CLI command:

```shellsession
$ python -m warehouse hashing journal-entry -h
Usage: python -m warehouse hashing journal-entry [OPTIONS]

  Hash `journals.submitted_from` column with salt

Options:
  -s, --salt TEXT            Pass value instead of prompting for salt
  -b, --batch-size INTEGER   Number of rows to hash at a time  [default:
                             10000]
  -st, --sleep-time INTEGER  Number of seconds to sleep between batches
                             [default: 1]
  --continue-until-done      Continue hashing until all rows are hashed
  -h, --help                 Show this message and exit.
```

### DIFF
--- a/warehouse/cli/hashing.py
+++ b/warehouse/cli/hashing.py
@@ -1,0 +1,135 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import hashlib
+import time
+
+import click
+
+from warehouse.cli import warehouse
+
+
+@warehouse.group()
+def hashing():
+    """
+    Run Hashing operations for Warehouse data
+    """
+
+
+@hashing.command()
+@click.option(
+    "-s",
+    "--salt",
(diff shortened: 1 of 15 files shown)


warehouse/cli/hashing.py
```diff
@@ -0,0 +1,135 @@
...
+    # Hash the value rows
+    click.echo(f"Hashing {how_many} rows...")
+    for row in unhashed_rows:
+        row.submitted_from = hashlib.sha256(
+            (row.submitted_from + salt).encode("utf8")
+        ).hexdigest()
+
+    # Update the rows
+    session.add_all(unhashed_rows)
+    session.commit()
```
ewdurbin, line 120: I'm thinking deeply about one thing, which is: Do we want to ensure that an IpAddress object and corresponding row exists in the DB before we obliterate the data?

*probably not* is my gut reaction. since for all data since https://github.com/pypi/warehouse/pull/6339 (August, 2019) there already exists a corresponding ProjectEvent that has it.

ewdurbin: hmmm now with the date in mind.... and the fact that the IP addresses exist on ProjectEvents... dropping may be the right call.

miketheman: As much as I enjoyed writing the CLI tool (and the tests for it!) I'd be much happier to remove the column instead. I'll work that up instead.

miketheman: https://github.com/pypi/warehouse/pull/13751

miketheman:
Closing, the alternative was to drop it like it was hot.

PR #11380, opened 2022-05-14
Add support for metadata 2.2 and 2.3

domdfcoding:
Fixes #9660 and #11526

I couldn't find any existing tests for the `Metadata-Version` validation so I haven't added any yet.
There are tests for the new validation functions for `Dynamic` and `Provides-Extra` though.

PR #13606, opened 2023-05-08
Metadata 2.2 and 2.3

### LINKED ISSUE #9660: Support Metadata Version 2.2
Since PEP 643 is accepted, PyPI should start accepting Metadata 2.2 distributions. See  PyO3/maturin#564.

Allow `Metadata-Version: 2.2` to pass the legacy upload endpoint’s form validation, and validate the corresponding `Dynamic` field.

The latter part could be a bit tricky since `Dynamic` is not allowed for wheels. Not sure how it could be best handled (I don’t have much experience with WTForm).

**Additional context**
https://www.python.org/dev/peps/pep-0643/

### LINKED ISSUE #11526: Add support for PEP 685 (Metadata Version 2.3)
[PEP 685](https://peps.python.org/pep-0685/) specifies rules for normalisation of extra names in core metadata, and in turn bumps the metadata version to 2.3. Currently it is not possible to upload wheels with metadata 2.3 because warehouse rejects them as having an unsupported metadata version.

The PEP suggests tools consuming metadata (like warehouse) emit warnings if invalid extra names are found for *older* metadata versions, although the tool may raise an error. I assume there's no mechanism in the upload API to return a warning, only an error? I don't think an error is appropriate here as it will likely break uploads of packages with previously valid metadata.

Tools should also warn (but may error) for invalid extras with metadata version 2.3. 

Therefore I propose leaving the validation unchanged for older versions, but returning an error if the metadata version is 2.3 (or later) and an extra name is invalid.

**Additional context**

I have an initial implementation here: https://github.com/pypa/warehouse/pull/11380/commits/48b5063a65a382d3bba622351858aea9c058f6c0

konstin:
This is an attempt to finish #11380 and unblock uploading metadata version 2.2 and 2.3, which i'd like to have for maturin and my resolver prototype. I've applied all review comments and added a check that dynamic isn't used <2.2.

Fixes #9660, fixes #11526, closes #11380.


warehouse/packaging/models.py
```diff
@@ -415,6 +415,45 @@ class ReleaseURL(db.Model):
     url = Column(Text, nullable=False)
 
 
+class Dynamic(db.Model):
```
ewdurbin, line 418: I don't think we need this to be a relation. I'm going to push a change to convert it to an Array containing the Enum

ewdurbin (dismissed):
The only thing I would want to see to call this shippable is a test that exercises the storage/retrieval of Dynamic fields.

I think this could be added to `tests.unit.forklift.test_legacy:test_successful_upload` or maybe more favorably built out as an additional test `tests.unit.forklift.test_legacy:test_successful_upload_metadata_2_3`.

Just some assertions of the resulting value of `Release.dynamic`

warehouse/forklift/legacy.py
```diff
@@ -402,6 +402,35 @@ def _validate_classifiers(form, field):
             )
 
 
+def _validate_dynamic(_form, field):
+    declared_dynamic_fields = set(map(str.title, field.data or []))
+    disallowed_dynamic_fields = {"Name", "Version", "Metadata-Version"}
```
ewdurbin, line 407: From https://github.com/pypi/warehouse/pull/11380#discussion_r1033894111:

PEP 643 only says:

> The fields `Name` and `Version` MUST NOT be marked as Dynamic.

Should `Metadata-Version` be included here? Or should the PEP be updated?

warehouse/forklift/legacy.py
```diff
@@ -571,6 +604,12 @@ class MetadataForm(forms.Form):
         description="Requires-Dist",
         validators=[wtforms.validators.Optional(), _validate_legacy_dist_req_list],
     )
+    provides_extra = (
```
ewdurbin, line 607: From https://github.com/pypi/warehouse/pull/11380#discussion_r1033923344

This needs to be added as a field of the Release model as well: https://github.com/pypi/warehouse/blob/main/warehouse/packaging/models.py#L423-L604

ewdurbin (commented):
Ported outstanding comments from #11380

warehouse/forklift/legacy.py
ewdurbin: I think this can be a generic `sqlalchemy.dialects.postgresql.ARRAY` of `sqlalchemy.Text` since the incoming are validated:

```
    provides_extra = Column(ARRAY(Text))
```

ewdurbin:
> The only thing I would want to see to call this shippable is a test that exercises the storage/retrieval of Dynamic fields.
> 
> I think this could be added to `tests.unit.forklift.test_legacy:test_successful_upload` or maybe more favorably built out as an additional test `tests.unit.forklift.test_legacy:test_successful_upload_metadata_2_3`.
> 
> Just some assertions of the resulting value of `Release.dynamic`

Missed that Provides-Extra also needs to be stored. Similarly a test for storage/retrieval of that data would be great to see.

warehouse/forklift/legacy.py
pfmoore: This is already the case in [the spec](https://packaging.python.org/en/latest/specifications/core-metadata/#dynamic-multiple-use). It was added in https://github.com/pypa/packaging.python.org/pull/1057 (I can't check this right now as github is playing up, but that's what git blame tells me locally!)

pfmoore: The principle is that the spec is the definitive document, and the PEP is historical, so there's no need to update the PEP.

ewdurbin: I'll defer to that!

dstufft:
We probably want to expose this value in the existing JSON API, XMLRPC API, and.. .maybe in the UI? Not sure.

ewdurbin:
Hi @konstin, wanted to check in and see if you planned to continue work on this. If not I'm happy to pick it up and continue on with this PR.

ewdurbin:
> We probably want to expose this value in the existing JSON API, XMLRPC API, and.. .maybe in the UI? Not sure.

Which value, Provides-Extra?

dstufft:
`Dynamic`, and whatever other fields are new

ewdurbin:
> We probably want to expose this value in the existing JSON API, XMLRPC API, and.. .maybe in the UI? Not sure.

I will implement this in JSON API, but won't touch XMLRPC. There's nowhere we would expose this on the XMLRPC that isn't currently slated for deprecation. If people want programatic access to this data, they should retrieve it from JSON.

I'll give a _smol_ attempt to render it in the web UI, but consider it beyond scope of this PR.

ewdurbin:
Low budget display if available on project pages:

low budget

ewdurbin:
@di @dstufft @miketheman, this is good for a review pass.

konstin:
Sorry for responding so slow and thank you for addressing all the review comments!

konstin:
Should i rebase this?

di:
@konstin Yes please!

konstin:
rebased and updated with the changes from main

warehouse/forklift/legacy.py
```diff
@@ -413,6 +414,40 @@ def _validate_classifiers(form, field):
             )
 
 
+def _validate_dynamic(_form, field):
+    declared_dynamic_fields = set(map(str.title, field.data or []))
```
di, line 418:
```suggestion
    declared_dynamic_fields = set(map(str.title, field.data or []))
    declared_dynamic_fields = {str.title(k) for k in field.data}
```

warehouse/forklift/legacy.py
```diff
@@ -413,6 +414,40 @@ def _validate_classifiers(form, field):
...
+    disallowed_dynamic_fields = {"Name", "Version", "Metadata-Version"}
+    if invalid := (declared_dynamic_fields & disallowed_dynamic_fields):
+        raise wtforms.validators.ValidationError(
+            f"The following metadata field(s) cannot be marked as dynamic: {invalid!r}",
+        )
+    allowed_dynamic_field = set(DynamicFieldsEnum.enums)
+    if invalid := (declared_dynamic_fields - allowed_dynamic_field):
+        raise wtforms.validators.ValidationError(
+            f"The following values are not valid for Dynamic: {invalid!r}"
+        )
```
di, line 428: I think what we actually want to do here is check if all the values in `Dynamic` are present as other metadata fields -- other verification checks will ensure that these fields are valid metadata fields for that metadata version. I'm not sure we need the enum at all.

warehouse/forklift/legacy.py
```diff
@@ -413,6 +414,40 @@ def _validate_classifiers(form, field):
...
+
+
+def _validate_extras(form, field):
+    try:
+        metadata_version = packaging.version.Version(form.metadata_version.data)
+    except packaging.version.InvalidVersion as e:
+        raise wtforms.validators.ValidationError(
+            f"The Metadata-Version is invalid: {e}"
+        )
+
```
di, line 441: This should already be getting validated elsewhere (and before this function is called) so it should be safe to assume that the metadata version is valid.

warehouse/forklift/legacy.py
```diff
@@ -582,6 +621,10 @@ class MetadataForm(forms.Form):
         description="Requires-Dist",
         validators=[wtforms.validators.Optional(), _validate_legacy_dist_req_list],
     )
+    provides_extra = ListField(
+        description="Provides-Extra",
+        validators=[wtforms.validators.Optional(), _validate_extras],
```
di, line 626:
```suggestion
        validators=[wtforms.validators.Optional(), _validate_provides_extras],
```

warehouse/forklift/legacy.py
```diff
@@ -413,6 +414,40 @@ def _validate_classifiers(form, field):
...
+    if invalid := (declared_dynamic_fields - allowed_dynamic_field):
+        raise wtforms.validators.ValidationError(
+            f"The following values are not valid for Dynamic: {invalid!r}"
+        )
+
+
+_extra_name_re = re.compile("^([a-z0-9]|[a-z0-9]([a-z0-9-](?!--))*[a-z0-9])$")
+
+
+def _validate_extras(form, field):
```
di, line 434:
```suggestion
def _validate_provides_extras(form, field):
```

warehouse/forklift/legacy.py
```diff
@@ -631,6 +674,13 @@ def full_validate(self):
                 "Include at least one message digest."
             )
 
+        # Dynamic is only allowed with metadata version 2.2+
+        if self.dynamic.data and self.metadata_version.data not in ["2.2", "2.3"]:
```
di, line 678:
```suggestion
        metadata_version = packaging.version.Version(form.metadata_version.data)
        if metadata_version >= packaging.version.Version("2.2")
```

warehouse/legacy/api/json.py
```diff
@@ -155,12 +155,16 @@ def _json_data(request, project, release, *, all_releases):
             "requires_dist": (
                 list(release.requires_dist) if release.requires_dist else None
             ),
+            "provides_extra": (
+                list(release.provides_extra) if release.provides_extra else None
+            ),
```
di, line 160: Should we make a distinction here between when the field is not provided, and when it is provided but empty? I'm not sure if the PEP addresses this or makes a distinction.

warehouse/legacy/api/json.py
```diff
@@ -155,12 +155,16 @@ def _json_data(request, project, release, *, all_releases):
...
             "docs_url": project.documentation_url,
             "bugtrack_url": None,
             "home_page": release.home_page,
             "download_url": release.download_url,
             "yanked": release.yanked,
             "yanked_reason": release.yanked_reason or None,
+            "dynamic": list(release.dynamic) if release.dynamic else None,
```
di, line 167: Ditto here.

warehouse/legacy/api/json.py
konstin: Both `Provides-Extra` and `Dynamic` are optional multiple use fields, is at least `METADATA` files can't express a difference between missing-empty and present-empty, so i'd say it's fine to treat them the same here.

warehouse/forklift/legacy.py
konstin: Note that this means that we allow 3.0 here (i deleted the test for that from the old PR)

konstin:
I tried to address all review comments, i had to change some of the suggested changes though to make it work

miketheman:
> Is there a way to tell the linters and formatters (or the `make` commands in general) to not lint my local `.venv`?

Tough to say exactly, since I don't use the same .venv style, as most development happens inside a container. You might want to add a .venv directory entry to https://github.com/pypi/warehouse/blob/main/.dockerignore so that it's not added to the container - give that a shot and if it works, feel free to add that to its own PR.

konstin:
Thanks, created https://github.com/pypi/warehouse/pull/14739

brettcannon:
Anything that can be done to help progress this? Metadata version 2.2 and newer are important for resolvers/installers as it lets them (potentially) skip having to download and build an sdist -- if the `PKG-INFO` is made available separately -- just to get the list of dependencies which is a massive compute, time, and bandwidth saver.

ewdurbin:
Rebased and ready for another review. I think the only outstanding questions are regarding how None versus empty are presented in JSON API: https://github.com/pypi/warehouse/pull/13606#discussion_r1261300580

@brettcannon if you could weigh in there I think we can probably move forward.

warehouse/legacy/api/json.py
pfmoore: I agree with @konstin - there's no meaningful way of interpreting "present but empty" *except* by not including the metadata item in the `METADATA` file, so the two are for all practical purposes equivalent.

warehouse/migrations/versions/778f1c01a019_release_add_provides_extra_and_dynamic_.py
```diff
@@ -0,0 +1,125 @@
...
+        "Classifier",
+        "Requires-Dist",
+        "Requires-Python",
+        "Requires-External",
+        "Project-URL",
+        "Provides-Extra",
+        "Provides-Dist",
+        "Obsoletes-Dist",
+        name="release_dynamic_fields",
+    ).drop(op.get_bind())
```
di, line 125: This migration should probably happen first, in a separate PR. It probably should also set a higher statement timeout than the default.

warehouse/forklift/legacy.py
```diff
@@ -394,6 +394,40 @@ def _validate_classifiers(form, field):
...
 
+def _validate_dynamic(_form, field):
+    declared_dynamic_fields = {str.title(k) for k in field.data or []}
+    disallowed_dynamic_fields = {"Name", "Version", "Metadata-Version"}
+    if invalid := (declared_dynamic_fields & disallowed_dynamic_fields):
+        raise wtforms.validators.ValidationError(
+            f"The following metadata field(s) cannot be marked as dynamic: {invalid!r}",
+        )
+    # "Supported-Platform" is missing in the metadata form since does not have any
+    # constraints
```
di, line 405:
```suggestion
    # "Supported-Platform" is missing in the metadata form since it does not have any
    # constraints
```

warehouse/forklift/legacy.py
```diff
@@ -394,6 +394,40 @@ def _validate_classifiers(form, field):
...
+    allowed_dynamic_fields = (
+        {field.description for field in MetadataForm() if field.description}
+        | {"Supported-Platform"}
```
di, line 408: I understand why this is here but this seems a bit weird, perhaps we should add the field to the form instead, even if we don't do anything with it.

It also seems like a bit of a hack that we're relying on the description for these values, should we be getting this from `DynamicFieldsEnum` instead?

warehouse/migrations/versions/778f1c01a019_release_add_provides_extra_and_dynamic_.py
konstin: What lock/statement timeouts do you suggest?

di: Adding this at the top of the migration should be sufficient:

```python
    conn = op.get_bind()
    conn.execute(sa.text("SET statement_timeout = 120000"))
```

warehouse/legacy/api/json.py
brettcannon: Since both results are functionally the same I think either approach is fine.

brettcannon:
> @brettcannon if you could weigh in there I think we can probably move forward.

Done!

warehouse/migrations/versions/778f1c01a019_release_add_provides_extra_and_dynamic_.py
konstin: I've created https://github.com/pypi/warehouse/pull/15452, i unfortunately can't retarget this PR onto the other branch since the branch is in my fork

warehouse/forklift/legacy.py
konstin: > I understand why this is here but this seems a bit weird, perhaps we should add the field to the form instead, even if we don't do anything with it.

Done

> It also seems like a bit of a hack that we're relying on the description for these values, should we be getting this from DynamicFieldsEnum instead?

I think we had that and changed it due to https://github.com/pypi/warehouse/pull/13606#discussion_r1261291228, or am i misreading your comment?

warehouse/forklift/legacy.py
di: What I meant by this was whether this check should ensure that all the fields listed in `Dynamic` are actually provided as metadata -- e.g., if we have `Dynamic: Author` but `Author` is not provided, should this fail? I think the PEP is unclear on this.

warehouse/forklift/legacy.py
di: What I meant there is that I don't think we don't need to check if `Dynamic` contains fields that aren't valid metadata fields (e.g. `Dynamic: ThisDoesntExist`), as those will fail when they're provided as metadata fields (e.g. `ThisDoesntExist: Foo`) *but* that assumes that the validator will always fail when `Dynamic` contains any value that isn't also present as a valid metadata field (e.g., `Dynamic: Author`, but `Author` isn't provided).

If the latter is true, we don't need the Enum or to check if `Dynamic` contains a valid metadata field name, we only need to:

1. Check that it doesn't contain a disallowed field name (`Dynamic: Metadata-Version` should fail);
2. Check that it only contains field names that are also provided as metadata (`Dynamic: Author` should fail if `Author` is not provided, `Dynamic: ThisDoesntExist` should fail because the form prevents `ThisDoesntExist` from being provided).

warehouse/forklift/legacy.py
konstin: I think both cases allowed, e.g. if you have `Dynamic: Requires-Dist`, you may not provide any requirements, but you still have to set it as `Dynamic` or consumers can assume it's a package with no dependencies. But you may give an incomplete/insufficiently constrained list and tighten it at build time (i have been told this is important for some packages that can build for different numpy version).

warehouse/forklift/legacy.py
di: OK, that makes sense too! Given that I think we do need the Enum but we can simplify this, I'll provide a suggestion in a moment.

warehouse/forklift/legacy.py
```diff
@@ -394,6 +394,37 @@ def _validate_classifiers(form, field):
...
+        raise wtforms.validators.ValidationError(
+            f"The following metadata field(s) cannot be marked as dynamic: {invalid!r}",
+        )
+    allowed_dynamic_fields = (
+        {field.description for field in MetadataForm() if field.description}
+    ) - disallowed_dynamic_fields
+    if invalid := (declared_dynamic_fields - allowed_dynamic_fields):
+        raise wtforms.validators.ValidationError(
+            f"The following values are not valid for Dynamic: {invalid!r}"
+        )
```
di, line 410:
```suggestion
    declared_dynamic_fields = {str.title(k) for k in field.data or []}
    disallowed_dynamic_fields = {"Name", "Version", "Metadata-Version"}
    if invalid := (declared_dynamic_fields & disallowed_dynamic_fields):
        raise wtforms.validators.ValidationError(
            f"The following metadata field(s) are valid, but cannot be marked as dynamic: {invalid!r}",
        )
    allowed_dynamic_fields = set(DynamicFieldsEnum.enums)
    if invalid := (declared_dynamic_fields - allowed_dynamic_fields):
        raise wtforms.validators.ValidationError(
            f"The following metadata field(s) are not valid and cannot be marked as dynamic: {invalid!r}"
        )
```
I think it's better to use an explicit Enum for this, because it's not actually guaranteed that `description` is set for all these fields. Also, ideally, we would reuse that same enum in the migration as well.

konstin: If we import the enum in the migration, wouldn't this break the migration if we ever update the enum? (Sorry if that's a stupid question, i don't have any experience with alembic)

di: It's a good question. It wouldn't necessarily break the migration itself (because it would always pull the latest enum when we were running the migration), but it might make the migration technically incorrect if that were ever to happen. I'm OK with not using the enum -- we can verify that they are the same before merging.

konstin: With that resolved, is there anything else you need to be changed to be able to merge #15452?

di: No, I think its fine!

di (approved):
LGTM, is blocked on merging #15452 first.

ewdurbin (approved):

PR #13646, opened 2023-05-11
Python CI speedups

ewdurbin:
does all the deps work once up front, caches the whole environment for downstream tests, only starts the postgres/stripe containers if necessary.

this should also retain a cache of built wheels for our deps thanks to https://github.com/pypa/pip/pull/11897, so incremental changes (dependabot or new dependency addition) shouldn't be too much of a penalty.

"wall clock" time for `Tests` isn't actually impacted *that much*, but all the other individual checks are **much** faster. so we'll fail faster when `Lint`, `Dependencies`, `Licenses` etc fail.

cuts actual compute time used in GHA from ~26m to ~11m in best case 💅🏼 

Compare https://github.com/pypi/warehouse/actions/runs/4946873280/usage with https://github.com/pypi/warehouse/actions/runs/4947718593/usage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:14.4
+        image: ${{ (matrix.name == 'Tests') && 'postgres:14.4' || '' }}
         ports:
           - 5432:5432
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,27 +83,14 @@ jobs:
...
-            requirements.txt
-            requirements/*.txt
-      - name: Cache common Python cache paths
-        uses: actions/cache@v3
+      - name: Restore built Python environment from deps
+        uses: actions/cache/restore@v3
         with:
-          path: |
-            .cache
-            .mypy_cache
```
miketheman, line 70: Question: are we intentionally removing the mypy cache?

ewdurbin: nope, I just goofed! will restore.

PR #13596, opened 2023-05-06
implement a generic "Observation" model mixin

ewdurbin:
Idea is to add a generic way to attach information to objects in the database.

Usecases:
  - Internal annotations to PyPI
  - User reported information via Web UI
  - Third-Party declarations about Projects/Releases/Files
    - Malware reports
    - Metadata/file validity

demo with "report project issue":
  - Logged in users see "Report Issue" button in project sidebar
  - Issues are reported/stored in db on submission

Further extensions could be done to accept these Observations via API.

PR #13647, opened 2023-05-11
Remove internal malware infrastructure/checks

### LINKED ISSUE #12412: Repurpose 'malware' checks / YARA scanning for secret scanning
While our existing 'malware' checks are not being used due to noise and false positives, the mechanics available via the checks are perfect for token/credential scanning.

We should be able to write a class of checks that:
- scans every file in a distribution for a given pattern
- calls an internal/external webhook with the result, in some specified format

Once this is implemented, we should support scanning for specific tokens:
- [ ] `pypi-` and `testpypi-` prefixed tokens, which should automatically be revoked when found
- [ ] 'reciprocal' scanning for GitHub tokens: https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning
- [ ] other platforms/ecosystems?

**Additional context**
Some other potential ecosystems that might want to integrate: https://docs.github.com/en/code-security/secret-scanning/secret-scanning-patterns

ewdurbin:
This languished.

There was some discussion of re-using this infrastructure for other use cases such as scanning uploads for secrets to report them to providers (similar to GitHub token scanning).

I propose #13596 combined with some form of "what's new!" feed to build out a more generic interface for internal and external "third parties" to perform these kinds of scans and report them to PyPI.

Closes #12412.


warehouse/migrations/versions/a81f6cb8fc31_remove_malware_infrastructure.py
```diff
@@ -0,0 +1,184 @@
...
+"""
+
+import citext
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "a81f6cb8fc31"
+down_revision = "fd0479fed881"
```
miketheman, line 27: **suggestion:** ‏either rebase alembic heads, or merge them with another migration:
```shell
docker compose run --rm web python -m warehouse db merge -m "merge c5f718 and a81f6c" c5f718 a81f6c
```
and add the new migration to commit scope.
Ref: https://alembic.sqlalchemy.org/en/latest/branches.html#merging-branches

warehouse/tasks.py
```diff
@@ -209,11 +209,8 @@ def includeme(config):
         task_default_queue="default",
         task_default_routing_key="task.default",
         task_queue_ha_policy="all",
-        task_queues=(
-            Queue("default", routing_key="task.#"),
-            Queue("malware", routing_key="malware.#"),
```
miketheman, line 214: **suggestion:** ‏with this removal, also remove the `worker-malware` entry from `Procfile`

miketheman (changes requested):
**nitpick:** ‏‏Looks to be a typo, triggered my `malware` searches: https://github.com/pypi/warehouse/blob/532091757513f1ecda747708861a834c47d09257/warehouse/cli/classifiers.py#L23


**suggestion:** ‏Remove leftovers from `tests/common/checks/__init__.py` and associated classes/modules


**thoughts:** ‏I didn't review the removed modules in depth, since they were removed in their entirety. Same applies for most of the test code. I figured with other protections in place, there's not a high degree of potential breakage. I focused predominantly on the integration parts of this code and the rest of the stack.

Other than a few smaller items included inline and above, this should be a relatively simple re-review once addressed.

docs/dev/development/malware-checks.rst
```diff
@@ -1,188 +0,0 @@
...
-      MALWARE_CHECK_BACKEND=warehouse.malware.services.DatabaseMalwareCheckService
-
-   In the development environment, Warehouse by default will only print the name
-   of the queued check instead of executing it.
-
-#. Add your new malware check to the database.
-
-    .. code-block:: bash
-
-      docker compose run web python -m warehouse malware sync-checks
```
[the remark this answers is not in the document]
ewdurbin: good catch.

warehouse/migrations/versions/a81f6cb8fc31_remove_malware_infrastructure.py
ewdurbin: https://github.com/pypi/warehouse/pull/13647/commits/980cbabe27b728a1c1c9244047d3c19ba5f7004d

miketheman (approved):
I added a commit removing test config I had missed earlier.

ewdurbin:
Gonna wait for a 👍🏼 from @di.

di (approved):
I think it's clear that supporting 3rd party reports from trusted reporters is a better path towards actually removing malware here.

docs/dev/development/malware-checks.rst
```diff
@@ -1,188 +0,0 @@
...
-`pull request 7377`_.
-
-
-.. |VerdictLifecycle| image:: ../_static/verdict-lifecycle.png
-.. _IMalwareCheckService: https://github.com/pypi/warehouse/blob/main/warehouse/malware/interfaces.py
-.. _DatabaseMalwareCheckService: https://github.com/pypi/warehouse/blob/main/warehouse/malware/services.py
-.. _celery crontab: https://docs.celeryq.dev/en/latest/reference/celery.schedules.html#celery.schedules.crontab
-.. _prepare classmethod:  https://github.com/pypi/warehouse/blob/main/warehouse/malware/checks/base.py
-.. _MalwareVerdict model: https://github.com/pypi/warehouse/blob/main/warehouse/malware/models.py
-.. |CheckLifecycle| image:: ../_static/check-lifecycle.png
```
merwok, line 177: I think the two images referenced here were not removed by the PR

PR #13651, opened 2023-05-12
Upload distribution files to archive first, then sync to cache

### LINKED ISSUE #13650: Convert B2 to be a service cache tier
We've experienced multiple business-day outages for operations that leverage file uploads to B2 Cloud, notably `forklift.legacy.file_upload`.

The Backblaze [Status Page indicates this is weekly scheduled maintenance window](https://status.backblaze.com/#:~:text=Scheduled%20Maintenance), and in chatting with one of their support reps even outside the window, their recommendation is to "try again, increase timeouts, or try again at the end of day"

We currently use B2 as the `files.backend` configured IFileStorage "primary" service, so we should consider what that means in the context of uploads.

One thought is to accept uploads to S3, and then kick off a task to replicate to B2 for cached serving. In the event that the task fails, apply some retry logic (expo backoff for N hours?).
In the interim, serve the file from S3 (which I think is already the case).

ewdurbin:
Resolves #13650

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -48,10 +48,7 @@ from warehouse.packaging.models import (
     Release,
     Role,
 )
-from warehouse.packaging.tasks import (
-    sync_file_to_archive,
-    update_bigquery_release_files,
-)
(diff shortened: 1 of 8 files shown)


miketheman (approved):
Let's go

PR #13652, opened 2023-05-12
ci: run postgres in a tmpfs for fast

ewdurbin:


ewdurbin:
hmmm, not *that* much faster but its hard to tell run-to-run. it did work though. perhaps worth merging and seeing if there's a consistent pickup.

ewdurbin:
Comparing runtimes over the past few days, this doesn't appear to have the speedup I hoped for.

Issue #13515, opened 2023-04-25
Trusted publishing: preserve OIDC claims for events


woodruffw:
As part of making trusted publisher events "richer" and more useful to users, we should figure out how to embed their associated OIDC claims.

This is currently impossible, due to how we perform the token exchange and request identity setup:

1. During token exchange, we take the OIDC token and exchange it for a PyPI API token; this API token is a macaroon that can contain arbitrary data, but we don't currently embed any of the OIDC claims in it;
2. During authentication with an OIDC-minted API token, we use `OIDCProvider` as our `request.identity` type, which are "stateless" with respect to the *specific* API token that produced them;
3. Consequently, there's no `request` state that links the current `request.identity` to either the underlying macaroon *or* OIDC token that produced it.

To fix this, we have to create links at (1) and (2): the minted API token should contain caveats (or other metadata?) for the OIDC token's claims, and `request.identity` should become a type that contains the claims extracted from the macaroon.

CC @di

woodruffw:
@tnytown will be looking at this as well.

woodruffw:
To summarize Slack chat, this should probably be done as two separate PRs:

1. Add the OIDC claims (or some reasonable subset of them?) to a new or existing Macaroon caveat (maybe the current `OIDCProvider` caveat?)
2. Find a way to link the request's macaroon's caveat with OIDC events

woodruffw:
> * Find a way to link the request's macaroon's caveat with OIDC events

We have a few options here:

* We could replace the current `request.identity` (`OIDCProvider`) with a wrapped type that includes both the `OIDCProvider` and the relevant claims.
* We could add another sidecar attribute to `request` or similar, e.g. `request._identity_oidc_claims`.

CC @di and @dstufft for opinions here -- I think modifying `request.identity` is probably the cleanest approach here, but I'm possibly missing something better as well 🙂

woodruffw:
Great work here @tnytown!

dstufft:
✨

PR #13668, opened 2023-05-15
Add OIDC claims to the `OIDCPublisher` caveat

tnytown:
See #13515.


warehouse/macaroons/caveats/__init__.py
```diff
@@ -105,6 +106,7 @@ def verify(self, request: Request, context: Any, permission: str) -> Result:
 @dataclass(frozen=True)
 class OIDCPublisher(Caveat):
     oidc_publisher_id: StrictStr
+    oidc_claims: SignedClaims | None = None
```
woodruffw, line 109: Embedding the whole `SignedClaims` is going to make this macaroon very large, and they're already very large.

Instead of doing the whole thing, we should probably limit it to the set of claims needed for the events we plan to emit. This will additionally prevent any unintentional forwarding of "secret" claims (there shouldn't be any currently, but future OIDC IdPs may choose to embed claims that shouldn't be forwarded).

tests/unit/oidc/test_views.py
```diff
@@ -396,7 +396,7 @@ def test_mint_token_from_oidc_no_pending_publisher_ok(monkeypatch):
     # NOTE: Can't set __str__ using pretend.stub()
     monkeypatch.setattr(publisher.__class__, "__str__", lambda s: "fakespecifier")
 
-    claims = pretend.stub()
```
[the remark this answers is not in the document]
woodruffw: Not that I can remember!

woodruffw:
Just thinking out loud: this _shouldn't_ break existing in-flight OIDC-minted tokens (and the lack of changed tests reflects that), but we should double-check that.

The scope of breakage would be relatively small in any case (since any in-flight tokens would only be alive for <15 minutes), but it should be entirely avoidable.

warehouse/oidc/views.py
```diff
@@ -191,7 +191,10 @@ def _invalid(errors):
             f"({datetime.fromtimestamp(not_before).isoformat()})"
         ),
         [
-            caveats.OIDCPublisher(oidc_publisher_id=str(publisher.id)),
+            caveats.OIDCPublisher(
+                oidc_publisher_id=str(publisher.id),
+                oidc_claims={"ref": claims["ref"], "sha": claims["sha"]},
```
woodruffw, line 196: One thing to note: `ref` should always be present, but the GitHub docs are a little bit slippery on whether `sha` always is.

For the time being, it might be worth making these fallible and treating them as progressive enhancements in the subsequent PR:

```suggestion
                oidc_claims={"ref": claims.get("ref"), "sha": claims.get("sha")},
```

woodruffw: (And correspondingly, let's add tests for when they're missing.)

woodruffw (approved):
LGTM, nice work!

CC @ewdurbin for review as well 🙂

PR #13680, opened 2023-05-16
Expose OIDC claims in `request` context from macaroon

### LINKED ISSUE #13515: Trusted publishing: preserve OIDC claims for events

tnytown:
WIP.

~~Blocked on #13668.~~
Resolves #13515.


warehouse/macaroons/caveats/__init__.py
```diff
@@ -176,3 +176,18 @@ def _verify_caveat(predicate: bytes):
     if not result:
         return WarehouseDenied("unknown error", reason="invalid_api_token")
     return Allowed("signature and caveats OK")
+
+
+def extract_oidc_claims(macaroon: Macaroon) -> SignedClaims | None:
+    for raw in macaroon.caveats:
+        try:
+            caveat = deserialize(raw.caveat_id)
```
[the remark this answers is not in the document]
woodruffw: Also, let's inline this function directly into `services.extract_oidc_claims`. No need for a discrete helper, unless I'm missing something.

warehouse/macaroons/interfaces.py
```diff
@@ -32,6 +32,12 @@ def find_userid(raw_macaroon):
         macaroon.
         """
 
+    def extract_oidc_claims(raw_macaroon):
```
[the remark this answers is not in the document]
woodruffw: IMO this is an acceptable place -- service implementations are meant to contain things like this 🙂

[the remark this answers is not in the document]
woodruffw: (It can really go either way, but it's where a lot of the Macaroon utility code already lives so IMO better to keep it all in one place.)

warehouse/oidc/utils.py
```diff
@@ -91,3 +94,9 @@ def find_publisher_by_issuer(session, issuer_url, signed_claims, *, pending=Fals
     else:
         # Unreachable; same logic error as above.
         return None  # pragma: no cover
+
+
+@dataclass
+class OIDCContext:
```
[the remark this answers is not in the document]
woodruffw: Here is okay, although IMO `oidc/security_policy.py` (i.e., a new module) would also be acceptable. But given that it'd be the only member, here is probably better.

(Let's add a docstring explaining what this class is with some references to other APIs that use it; it's a little bit incongruous otherwise.)

woodruffw:
Unblocked: #13668.

warehouse/forklift/legacy.py
```diff
@@ -1161,6 +1161,7 @@ def file_upload(request):
                 else "OpenID created token",
                 "canonical_version": release.canonical_version,
                 "publisher_url": request.oidc_publisher.publisher_url
+                + f"@{request.oidc_claims.get('sha', 'unknown')}"
```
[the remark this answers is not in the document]
woodruffw: The URL here should be a GitHub repository, so we can render a direct URL to the commit, like so: 

```
https://github.com/{user}/{repo}/commit/{sha}
```

In other words: if we're given the `sha`, we should link directly to the commit view for it. If we aren't given the `sha`, we should gracefully degrade to a URL for just the repo.

> It seems we don't have a call_recorder that checks record_event calls for the tests covering this. Do we want that?

Yeah, probably an oversight due to implicit coverage. Please add that!

woodruffw (approved):
LGTM overall! Nice work @tnytown.

Could you post some screenshots of what the event renders now look like with these changes?

I'll also do some user testing later today.

woodruffw:
> * Do we want make a link or template the URL in as plain text?

@di and @ewdurbin have the final say, but I'm personally OK with the current render -- IMO having the `sha` visible is useful.

> If we do make it a link, do we want to sanitize it somehow? Do we trust the value in sha?

Yeah, `sha` is trusted -- we could add some additional hardening around the `SignedClaims` if we're feeling paranoid, but the situation in which `sha` would be anything but a hex string is one in which an attacker could control a valid GitHub JWT's payload. And if that happens, we have much more serious things to worry about 🙂

tests/unit/forklift/test_legacy.py
```diff
@@ -3282,18 +3313,55 @@ def test_upload_succeeds_creates_release(
...
+                mock.ANY,
+                tag=EventTag.Project.ReleaseAdd,
+                ip_address=mock.ANY,
+                additional={
+                    "submitted_by": identity.username
+                    if test_with_user
+                    else "OpenID created token",
+                    "canonical_version": release.canonical_version,
+                    "publisher_url": identity.publisher.publisher_url
+                    + "/commit/somesha"
```
woodruffw, line 3341: Extreme nitpick since it's a test, but let's do an f-string here: 

```python
f"{identity.publisher.publisher_url}/commit/somesha"
```

(Similarly for elsewhere below.)

tests/unit/forklift/test_legacy.py
```diff
@@ -3282,18 +3313,55 @@ def test_upload_succeeds_creates_release(
...
+                tag=EventTag.File.FileAdd,
+                ip_address=mock.ANY,
+                additional={
+                    "filename": filename,
+                    "submitted_by": identity.username
+                    if test_with_user
+                    else "OpenID created token",
+                    "canonical_version": release.canonical_version,
+                    "publisher_url": identity.publisher.publisher_url
+                    + "/commit/somesha"
```
woodruffw, line 3357: Ditto.

warehouse/forklift/legacy.py
```diff
@@ -1161,6 +1161,7 @@ def file_upload(request):
                 else "OpenID created token",
                 "canonical_version": release.canonical_version,
                 "publisher_url": request.oidc_publisher.publisher_url
+                + f"/commit/{request.oidc_claims.get('sha', 'unknown')}"
```
woodruffw, line 1164: We should degrade more gracefully than this: if `sha` isn't present, then let's just render the bare `publisher_url` without any `/commit/...` path.

I'd be okay with abstracting that behind some kind of helper on `OIDCPublisher`, e.g.:

```
"publisher_url": request.oidc_publisher.publisher_url(request.oidc_claims)
```

In other words: `publisher_url` could go from being a `@property` to a method that progressively enhances based on the state in `request.oidc_claims`.

warehouse/macaroons/caveats/__init__.py
```diff
@@ -111,12 +110,12 @@ class OIDCPublisher(Caveat):
     def verify(self, request: Request, context: Any, permission: str) -> Result:
         # If the identity associated with this macaroon is not an OpenID publisher,
         # then it doesn't make sense to restrict it with an `OIDCPublisher` caveat.
-        if not isinstance(request.identity, oidc_models.OIDCPublisher):
+        if not request.oidc_publisher:
```
woodruffw, line 113: Good riddance to another `isinstance` check 🙂

woodruffw (commented):
A few more small nits, but this is looking good!

tests/unit/forklift/test_legacy.py
woodruffw: Gotcha. Given that it's just test code, this is fine then 🙂

warehouse/oidc/models/github.py
```diff
@@ -149,6 +145,12 @@ def job_workflow_ref(self):
     def sub(self):
         return f"repo:{self.repository}"
 
+    def publisher_url(self, claims=None):
+        sha = claims.get("sha") if claims else None
+        suffix = f"/commit/{sha}" if sha else ""
+
+        return f"https://github.com/{self.repository}{suffix}"
```
woodruffw, line 152: Avoiding ternaries where they aren't necessary:

```suggestion
        base = f"https://github.com/{self.repository}"
        sha = claims.get("sha") if claims else None
        
        if sha:
            return f"{base}/commit/{sha}"
        return base
```

woodruffw (approved):
LGTM, nice work @tnytown!

woodruffw:
This should be good to go again; made compatible with the IP address changes.

warehouse/macaroons/models.py
```diff
@@ -69,6 +69,10 @@ class Macaroon(db.Model):
     # body of the permissions ("V1") caveat.
     permissions_caveat = Column(JSONB, nullable=False, server_default=sql.text("'{}'"))
 
+    # OIDC claims associated with this macaroon. Should only exist for OIDC
+    # publisher macaroons.
```
woodruffw, line 73: We'll probably want to use this column for other things in the future (e.g. collapsing `permissions_caveat` into it), so let's tweak this:

```suggestion
    # Additional state associated with this macaroon.
    # For OIDC publisher-issued macaroons, this will contain a subset of OIDC claims.
```

warehouse/migrations/versions/646bc86a09b6_macaroon_oidc_claims.py
```diff
@@ -0,0 +1,63 @@
...
+#       prevent situations where a slow migration takes the entire site down.
+#
+#       If you need to increase this timeout for a migration, you can do so
+#       by adding:
+#
+#           op.execute("SET statement_timeout = 5000")
+#           op.execute("SET lock_timeout = 4000")
+#
+#       To whatever values are reasonable for this migration as part of your
+#       migration.
```
woodruffw, line 48: Safe to delete:

```suggestion
```

warehouse/migrations/versions/646bc86a09b6_macaroon_oidc_claims.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "macaroons",
+        sa.Column("additional", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
+    )
+    # ### end Alembic commands ###
```
woodruffw, line 57:
```suggestion
    op.add_column(
        "macaroons",
        sa.Column("additional", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
    )
```

warehouse/migrations/versions/646bc86a09b6_macaroon_oidc_claims.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("macaroons", "additional")
+    # ### end Alembic commands ###
```
woodruffw, line 63:
```suggestion
    op.drop_column("macaroons", "additional")
```

warehouse/oidc/views.py
```diff
@@ -194,12 +194,12 @@ def _invalid(errors):
         [
             caveats.OIDCPublisher(
                 oidc_publisher_id=str(publisher.id),
-                oidc_claims={"ref": claims.get("ref"), "sha": claims.get("sha")},
             ),
             caveats.ProjectID(project_ids=[str(p.id) for p in publisher.projects]),
             caveats.Expiration(expires_at=expires_at, not_before=not_before),
         ],
         oidc_publisher_id=publisher.id,
+        additional={"ref": claims.get("ref"), "sha": claims.get("sha")},
```
woodruffw, line 202: Since we'll want to use `additional` for other things as well, let's group this under a unique-looking key:

```suggestion
        additional={"oidc": {"ref": claims.get("ref"), "sha": claims.get("sha")}},
```

(That means an additional step during retrieval as well.)

warehouse/macaroons/security_policy.py
```diff
@@ -90,13 +91,15 @@ def identity(self, request):
 
         try:
             dm = macaroon_service.find_from_raw(macaroon)
+            oidc_claims = dm.additional if dm.oidc_publisher else None
```
woodruffw, line 94:
```suggestion
            oidc_claims = dm.additional.get("oidc") if dm.oidc_publisher else None
```

woodruffw (approved):
LGTM, nice work!

dstufft (approved):

PR #13703, opened 2023-05-20
Filter maintainer's email when presenting project data

jdknight:
When preparing to present the `maintainer_email` metadata for a project, format this content in the same way as `author_email`. This is to prevent rendering a maintainer's Email inside the mailto's link text. For [example](https://pypi.org/project/sphinxcontrib-confluencebuilder/):

image


ewdurbin (approved):

ewdurbin:
Thanks @jdknight, makes good sense.

PR #13706, opened 2023-05-21
Implement PEP 714

### LINKED ISSUE #13705: Wrong key name used for PEP 658 metadata files in the JSON index
**Describe the bug**

[PEP 691](https://peps.python.org/pep-0691/#project-detail) states that the key name for metadata files in the JSON index should be `dist-info-metadata`:

> `dist-info-metadata`: An optional key that indicates that metadata for this file is available, via the same location as specified in [PEP 658](https://peps.python.org/pep-0658) (`{file_url}.metadata`).

However, warehouse is providing it under the `data-dist-info-metadata` key instead:

```
$ curl -H 'Accept: application/vnd.pypi.simple.v1+json' https://pypi.org/simple/fluffy-server/ | jq .files
[...]
  {
    "data-dist-info-metadata": {
      "sha256": "4db99543165cbdeef42ccb6257545911ccd7865d65e304e3e056f383a25f309c"
    },
    "filename": "fluffy_server-1.39.2-py3-none-any.whl",
    [...]
```

This is causing pip to not use the metadata files as it is looking for the `dist-info-metadata` key only:
https://github.com/pypa/pip/blob/f25f8fffbbd16fdb13a4f8977946afe9a3248453/src/pip/_internal/models/link.py#L265

**Additional context**

There are two bugs discovered recently in pip which may make this tricky to fix:

* https://github.com/pypa/pip/issues/12042
* https://github.com/pypa/pip/issues/12038

I believe if we simply fix the key name in pypi.org, it will break existing pip versions as it will cause users to encounter these bugs. It may be necessary to coordinate this fix with fixes to the above bugs in pip to avoid disruption?

ewdurbin:
Closes #13705

Needs to coordinate with https://github.com/pypa/pip/issues/12042


dstufft (approved):

ewdurbin:
Blocked on https://github.com/pypa/pip/issues/12042

ewdurbin:
> Given [pypa/pip#12042 (comment)](https://github.com/pypa/pip/issues/12042#issuecomment-1579350451) we should probably close this and find an alternative.

Gonna implement [714](https://peps.python.org/pep-0714/) provisionally instead.

warehouse/templates/api/simple/detail.html
```diff
@@ -20,7 +20,7 @@
   <body>
     <h1>Links for {{ name }}</h1>
     {% for file in files -%}
-    <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}" {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}" {% endif %}{% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}" {% endif %}{% if file['data-dist-info-metadata'] %}data-dist-info-metadata="sha256={{ file['data-dist-info-metadata']['sha256'] }}"{% endif %}>{{ file.filename }}</a><br />
+    <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}" {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}" {% endif %}{% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}" {% endif %}{% if file['data-dist-info-metadata'] %}data-dist-info-metadata="sha256={{ file['data-dist-info-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['data-dist-info-metadata']['sha256'] }}"{% endif %}>{{ file.filename }}</a><br />
```
dstufft, line 23: These keys probably need updated to say ``file["core-metadata"]`` if I remember the way this works correctly.

ewdurbin: the object part comes back key'd as 'data-dist-info-metadata' which is internal to the implementation.

we can remap them internally as part of this PR!

ewdurbin: ope, we already did! we just send both from `warehouse.packaging.utils._simple_detail` we can reference either here. I'll update to the new canonical source :)

dstufft: Ah missed the previous key in `_simple_detail` was still there, probably worth to use the new name and just remove the old name completely to keep it from getting out of date or something?

ewdurbin: 9de3b7b2b76b4d480144de0b0604781ab025655d

dstufft (approved):

ewdurbin:
PEP 714 has been accepted https://github.com/python/peps/pull/3186!

edmorley:
Thank you for implementing this! 

Is there an issue filed for tracking the backfilling of metadata files for older package releases?

di:
@edmorley: https://github.com/pypi/warehouse/issues/8254

PR #13707, opened 2023-05-21
Remove support for PGP signatures

dstufft:
I'd like to propose removing PGP support from PyPI, which is something we've talked about in the past but have generally punted on it because of the idea that someone, somewhere might be validating these signatures and maybe they're providing some kind of use. Thus we went for a middle ground where we removed it from the front end, but still provided support for uploading them if someone felt so inclined to do so. The idea being that the only people who would upload them, are the people who really care and thus perhaps they'd form a stronger signal.

Unfortunately, that appears not to be the case. In a [recent post](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless), @woodruffw explores the signatures that exist on PyPI today (or at least, in the last 3 years). The full post is instructive, but it's shocking the results that he found.

Given that, I'd like to just remove the last remaining vestigial components of PGP support on PyPI. Hardly anyone is even using it at this point, so impact should be minimal, but even for those using it:

- Uploaders will just have signatures silently ignored, so their upload tasks will continue to work.
- Downloaders will just have the experience that nobody has any PGP keys for any new projects.
- Due to the way our file hosting works, all existing PGP signatures will continue to work, and metadata in BigQuery will continue to list `has_signature: True`, but metadata from PyPI itself will not. No future PGP signatures will be saved/emitted.


woodruffw:
> * Uploaders will just have signatures silently ignored, so their upload tasks will continue to work.
> 
> * Downloaders will just have the experience that nobody has any PGP keys for any new projects.
> 
> * Due to the way our file hosting works, all existing PGP signatures will continue to work, and metadata in BigQuery will continue to list `has_signature: True`, but metadata from PyPI itself will not. No future PGP signatures will be saved/emitted.

This seems like a very reasonable set of behaviors to me!

ewdurbin:
I think this is worth adding a blog post about. Would you mind writing something up as part of this PR?

dstufft:
@ewdurbin done I think? I haven't written one before.

dstufft:
It would be great to get @woodruffw to double check the blog post as well.

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,56 @@
...
+Given all of this, the continued support of uploading PGP signatures to PyPI is
+no longer defenfsible. While it doesn't represent a *massive* operational burden
+to continue to support it, it does require any new features that touch the
+storage of files to be made aware of and capable of handling these PGP
+signatures, which is a non zero cost on the maintainers and contributors of
+PyPI.
+
+---
+
+_Donald Stufft is a maintainer of the Python Package Index since 2013._
```
ewdurbin, line 50:
```suggestion
_Donald Stufft is a PyPI administrator and maintainer of the Python Package Index since 2013._
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,56 @@
...
+If you are someone who is currently uploading signatures, your uploads will
+continue to succeed, but any PGP signatures will be silently ignored. If you are
+someone who is currently downloading PGP signatures, existing signatures
+*SHOULD* continue to work [^1], but no new signatures will be made available.
+The related API fields such as `has_sig` have all been hardcoded to always be
+`False`.
+
+Historically PyPI has supported uploading PGP signatures alongside the release
+artifacts in an attempt to provide some level of package signing. However, the
+approach used long standing, [documented issues](https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/) which had lead us to demphasize the support
```
ewdurbin, line 22:
```suggestion
approach used long standing,
[documented issues](https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/)
which had lead us to demphasize the support
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,56 @@
...
+for PGP signatures over time by removing them from the UI.
+
+PyPI did still support uploading these signatures in the hope that there
```
ewdurbin, line 25:
```suggestion
PyPI continued to support uploading these signatures in the hope that there
```

ewdurbin (commented):
Just reviewing blog for now.

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,58 @@
...
+
+However, a
+[recent look at the signatures on PyPI ](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless)
+shows us that the current support for PGP signatures is not providing useful.
+
+In the last 3 years, about 50k signatures had been uploaded to PyPI by 1069
+unique keys. Of those 1069 unique keys, about 30% of them were not discoverable
+on major public keyservers, making it difficult or impossible to actually verify
+those signatures. Of the remaining 71%, nearly half of them were unable to be
+verified at the time of the audit (2023-05-19) [^2].
```
woodruffw, line 38: Just for clarity: "unable to be verified" refers to the keys themselves being expired or having no valid binding signature at all; the signatures themselves _could_ be verified with these keys but it's unclear what that verification would mean (since the key is either explicitly expired or not bound to any identity).

dstufft: I'm struggling to think of good way of wording that, and the best I'm coming up with is leaving it as is, or tacking on a "securely" or "meaningfully" before verified.

That's mostly because I personally consider verified to mean that we're checking the signature against all of the constraints of the key, and one of those constraints is expiry (for instance). I [looked around](https://gist.github.com/justinsteven/ac8bd55ada5e15be946ba9c5ac8f89fc) to see what gpg actually does here, and annoyingly it seems to call the signature verified but explicitly warns that the key is expired (I wonder how many tools actually surface that information...).

```
% cat <<EOF | gpg --verify
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

My message
-----BEGIN PGP SIGNATURE-----

iQEzBAEBCgAdFiEEVfpOheaJaGHOouQC9y10eKQylzsFAlrA5eAACgkQ9y10eKQy
lztvBQf+Ptxx3oiDnScJFClec2WNuYrDL3H4Tv63OIEeN8tiJItpCSSQYB2oX6MZ
nSegd6l5MHiBFEJ+lFi4JCqfUjoav/XhL9A94meCE9xI31B1Fo1yTWYob+8xLWZN
n0Tx0AtM0k+7mFl902r0Cu+e5DOxalTReVQp5IHqxK4u1g7KQjIGWYM+WTb96H/Z
RDapskb8T6zpN0D4IPZWPiUrLIKw+x3qSr8sNLGJMsKNRIyRLwYrSIDdVZIsmDzH
wStwZkAEWP1NNGVi8PXK2SD/lLmGhhMhJIj93ld8IL3fboQ28fOGys1o/Dm1RT/3
dfB3K1sM3dqxDa3WrJdk+pcZrcu4gQ==
=dYoC
-----END PGP SIGNATURE-----

EOF

gpg: Signature made Mon 02 Apr 2018 00:00:00 AEST
gpg:                using RSA key 55FA4E85E6896861CEA2E402F72D7478A432973B
gpg: Good signature from "Test Key <test@localhost>" [expired]
gpg: Note: This key has expired!
Primary key fingerprint: 55FA 4E85 E689 6861 CEA2  E402 F72D 7478 A432 973B
```

So maybe I'll figure out a way to qualify that verified in some way.

dstufft: I tried to do a better job of explaining this in the most recent commit, hopefully that sounds better!

woodruffw: Looks good to me!

ewdurbin (approved):
Code changes look good, and I support this deprecation.

Any @pypi/warehouse-admins have concerns?

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+[documented issues](https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/)
+which had lead us to demphasize the support
+for PGP signatures over time by removing them from the UI.
+
+PyPI continued to support uploading these signatures in the hope that there
+might be some systems out there that found them useful.
+
+However, a
+[recent look at the signatures on PyPI ](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless)
+shows us that the current support for PGP signatures is not providing useful.
```
str4d, line 32:
```suggestion
[recent look at the signatures on PyPI](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless)
shows us that the current support for PGP signatures is not proving useful.
```
(or `providing useful <SOMETHING>.`)

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+publish_date: 2023-05-21
+date: "2023-05-21 00:00"
+tags:
+  - security
+---
+
+PyPI has removed support for uploading PGP signatures.
+
+If you are someone who is currently uploading signatures, your uploads will
+continue to succeed, but any PGP signatures will be silently ignored. If you are
```
woodruffw, line 14: Just to disambiguate with "uploading PGP signatures" directly above 🙂 

```suggestion
If you are someone who is currently uploading signatures, your package uploads will
continue to succeed, but any PGP signatures will be silently ignored. If you are
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+someone who is currently downloading PGP signatures, existing signatures
+*SHOULD* continue to work [^1], but no new signatures will be made available.
```
woodruffw, line 16: Given the emphasis on them not actually working below:

```suggestion
*SHOULD* continue to be available [^1], but no new signatures will be made available.
```

woodruffw (approved):
LGTM! Two tiny nits, IMO optional 🙂

miketheman (commented):
couple of nits in the blog post, otherwise I trust other reviewers

di:
> Uploaders will just have signatures silently ignored, so their upload tasks will continue to work.

I think this is going to be fairly confusing to the any users who are uploading signatures and expecting PyPI to persist them. I think this assumes that nobody will notice, but I don't think that _everyone_ uploading signatures is doing so unknowingly. That said, I agree with the goal of not breaking anyone's uploads.

I think it would probably be best for us to let the user know about this somehow. Two ways we can do this, I think we should do both:

1) [Twine lets us display arbitrary messages in the response.](https://github.com/pypa/twine/blob/0bb428c410b8df64c04dc881ac1db37d932f3066/twine/commands/upload.py#L149-L151). We can still let the request succeed, but if `gpg_signature` is present, we send back a message indicating that the signature will not be stored (maybe also include a link to the blog post?)
2) Similarly, send an email to the uploading user at upload time if `gpg_signature` is present.

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Removing PGP from PyPI.
+description: PyPI has removed support for PGP signatures.
```
di, line 3:
```suggestion
description: PyPI has removed support for uploading PGP signatures with new releases.
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+author: Donald Stufft
+publish_date: 2023-05-21
+date: "2023-05-21 00:00"
+tags:
+  - security
+---
+
+PyPI has removed support for uploading PGP signatures.
+
```
di, line 12:
```suggestion
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+If you are someone who is currently uploading signatures, your package uploads will
+continue to succeed, but any PGP signatures will be silently ignored. If you are
+someone who is currently downloading PGP signatures, existing signatures
+*SHOULD* continue to be available [^1], but no new signatures will be made available.
+The related API fields such as `has_sig` have all been hardcoded to always be
+`False`.
+
+Historically PyPI has supported uploading PGP signatures alongside the release
```
di, line 20:
```suggestion
Historically, PyPI has supported uploading PGP signatures alongside the release
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+artifacts in an attempt to provide some level of package signing. However, the
+approach used had long standing,
+[documented issues](https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/)
+which had lead us to demphasize the support
```
di, line 24:
```suggestion
which had previously lead us to deemphasize the support
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+for PGP signatures over time by removing them from the UI.
```
di, line 25:
```suggestion
for PGP signatures over time by removing them from the PyPI web user interface.
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+
+PyPI continued to support uploading these signatures in the hope that there
+might be some systems out there that found them useful.
+
+However, a
+[recent look at the signatures on PyPI](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless)
+shows us that the current support for PGP signatures is not proving useful.
```
di, line 32:
```suggestion
PyPI has continued to support uploading these signatures in the hope that there
might be some systems out there that found them useful. Recently though,
[an examination of the signatures on PyPI](https://blog.yossarian.net/2023/05/21/PGP-signatures-on-PyPI-worse-than-useless)
has revealed to us that the current support for PGP signatures is not proving useful.
```

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
...
+
+In the last 3 years, about 50k signatures had been uploaded to PyPI by 1069
+unique keys. Of those 1069 unique keys, about 30% of them were not discoverable
+on major public keyservers, making it difficult or impossible to meaningfully
+verify those signatures. Of the remaining 71%, nearly half of them were unable
+to be meaningfully verified at the time of the audit (2023-05-19) [^2].
+
+In other words, out of all of the unique keys that had uploaded signatures to
+PyPI, only 36% of them were capable of being meaningfully verified [^3] at the
+time of audit.
```
di, line 42: Thirty-six percent sounds like a lot, I think we need an additional quantifier here, like "and those files that had verifiable signatures only represented 0.000X% of files on PyPI".

docs/blog/posts/2023-05-21-removing-pgp.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Removing PGP from PyPI.
+description: PyPI has removed support for PGP signatures.
+author: Donald Stufft
+publish_date: 2023-05-21
```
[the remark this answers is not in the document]
dstufft: I'll commit this now, but fix the filename right  before I publish.

docs/blog/posts/2023-05-21-removing-pgp.md
dstufft: To make it fair, I'll have to look up how many files have been released in the 3 year window that @woodruffw used in his blog post, but I can do that.

di: There were 6556154 files without signatures and 53030 with them, so 0.3% if my math is right.

dstufft: I've added some wording to mention this.

dstufft:
I've gone ahead and added (I believe) a warning when using twine and the email (time limited to once per day, though I'm tempted to do once per year or something).

I think if we add the email, we probably want to open an issue to remove the email at some point in the future. I don't think emailing forever is a useful thing... maybe 6 months? Maybe a year? Not sure.

warehouse/migrations/versions/d738a238d781_remove_pgp_signature_support.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from alembic import op
+
+revision = "d738a238d781"
+down_revision = "c5f718cb98ac"
+
+
+def upgrade():
+    op.drop_column("release_files", "has_signature")
```
dstufft, line 27: Note: This migration will cause failures when deployed until the old version of Warehouse is no longer serving traffic. I plan to drop it from the PR before merging, and move it to a follow up PR.

woodruffw:
> I don't think emailing forever is a useful thing... maybe 6 months? Maybe a year? Not sure.

A year makes sense to me -- I don't see any particular harm in it being longer rather than shorter, but agree completely about turning the emails off at some point.

woutdenolf:
The error clients get because of this it rather cryptic

```
         gpg_signature: Should not be a tuple.
```

dstufft:
Fixed in https://github.com/pypi/warehouse/pull/13738

Issue #13702, opened 2023-05-19
Test using redis as the celery broker


miketheman:
We're using Amazon SQS as our message broker today.

Celery support for SQS does not afford Monitoring nor Remote Control from within the context of Celery.

> Missing monitor support means that the transport doesn’t implement events, and as such Flower, celery events, celerymon and other event-based monitoring tools won’t work.
> Remote control means the ability to inspect and manage workers at runtime using the celery inspect and celery control commands (and other tools using the remote control API).
--https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/index.html

We can monitor items like queue depth for SQS via Cloudwatch Metrics (and possibly in Datadog) but we have little visibility as to **what** is in the queue itself.

@ewdurbin and I recently chatted about the notion of queue debouncing/superseding, and I raised the question of whether we might want to use Redis as the broker instead. We already use Redis as part of the stack, and we could leverage the same cluster/instance for the Celery broker.
This would do a couple of things:
1. Remove reliance on a provider's secret sauce service in favor of a portable technology
2. Increase visibility via tools like Flower
3. ~Remove some extra dependencies needed for SQS like `kombu`~
4. Increase usage of our current Redis cluster - both in terms of bandwidth and memory. TBD on how to look at this.

We also discussed using SQLAlchemy and Postgres as a backend, and rejected it. [The docs kinda support that decision](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/index.html#:~:text=Historically%2C%20SQLAlchemy%20has%20not%20been%20the%20most%20stable%20result%20backend%20so%20if%20chosen%20one%20should%20proceed%20with%20caution.).

The change "seems" easy, since most of our setup is already governed by a config var of `BROKER_URL`:
https://github.com/pypi/warehouse/blob/67d5b04228bd401eb75d29c3efec92383e51c9ad/warehouse/config.py#L180
https://github.com/pypi/warehouse/blob/38e0e0400f8585e382aa5d48836ef08fcfde742a/warehouse/tasks.py#L184

We already use Redis in aspects of our celery lifecycle:
https://github.com/pypi/warehouse/blob/67d5b04228bd401eb75d29c3efec92383e51c9ad/warehouse/config.py#L181-L182

[...]

PR #13714, opened 2023-05-22
chore: replace local development broker with redis

miketheman:
As a means to test and demonstrate that using Redis as a broker is an easy replacement for SQS.

Add flower at `http://localhost:5555` to play around with and observe behaviors.

Doesn't touch main runtime dependencies until we decide to retire SQS behaviors, and even then, not much changes other than minor tweaks to the depgraph - we still need all the packages we install anyhow.

Drops the largeish ~2GB localstack docker image

Relates to #13702

### DIFF
--- a/dev/environment
+++ b/dev/environment
@@ -6,7 +6,7 @@ WAREHOUSE_TOKEN=insecuretoken
 
 AWS_ACCESS_KEY_ID=foo
 AWS_SECRET_ACCESS_KEY=foo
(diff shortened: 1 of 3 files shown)


firefly9966 (approved):

miketheman:
Closing for now - we should address the shared Redis database (REDIS_UIRL and BROKER_URL being effectively the same) before looking at this for real production usage and deciding if we want to abandon SQS.

PR #13716, opened 2023-05-22
feat: pass the request to record_event

miketheman:
Add the ability to pass and consume `request` values for Events.

Not comprehensive for all events, start with Organization-and Team-related events as a way to ease ourselves into it.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -471,12 +471,12 @@ msgid ""
 msgstr ""
 
 #: warehouse/manage/views/__init__.py:2278
-#: warehouse/manage/views/organizations.py:882
+#: warehouse/manage/views/organizations.py:895
 msgid "User '${username}' already has an active invite. Please try again later."
 msgstr ""
 
 #: warehouse/manage/views/__init__.py:2344
-#: warehouse/manage/views/organizations.py:947
+#: warehouse/manage/views/organizations.py:962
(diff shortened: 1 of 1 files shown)


warehouse/events/models.py
```diff
@@ -129,6 +129,18 @@ def record_event(
     ):
         """Records an Event record on the associated model."""
         session = orm.object_session(self)
+
+        if request is not None:
+            # Add `request.ip_address` data to `Event.additional` as a nested key
+            additional = additional or {}
+            additional["ip_address"] = {}
```
ewdurbin, line 136: I'd remove these, and instead populate the `ip_address_obj` relation on the event if we have an `IPAddress`!

Something like

```
ip_address_obj = None
if request is not None:
    ip_address_obj = request.ip_address
    additional = additional or {}
    additional["geoip_info"] = (
        request.ip_address.geoip_info or None
    )

event = self.Event(
    source=self,
    tag=tag,
    ip_address=ip_address,
    ip_address_obj=ip_address_obj,
    additional=additional,
)

ewdurbin (changes requested):

warehouse/events/models.py
miketheman: We discussed in depth synchronously, and I think I got the gist of it.

A refactor commit was added - it's a bit more involved than I had hoped for, but I think it does the trick.

ewdurbin (approved):

PR #13717, opened 2023-05-22
chore(dev): update stripe-mock to latest

miketheman:


ewdurbin (approved):

ewdurbin (changes requested):
Wait... https://github.com/pypi/warehouse/blob/db21fc3f4d434771c273c74625612c3a4fd76dc4/.github/workflows/ci.yml#L73-L74

miketheman:
Thanks! What is that really doing for us?

miketheman:
Added in 8154b67

ewdurbin (approved):

ewdurbin:
> Thanks! What is that really doing for us?

The organization billing tests hit it.

miketheman:
> > Thanks! What is that really doing for us?
> 
> The organization billing tests hit it.

Derp, I probably should have phrased that differently. I meant to inquire what this interesting syntax was doing - and it looks like it's applying an `if` statement where an `if` isn't readily available, so it's creating an image string only when the condition applies.

ewdurbin:
> > > Thanks! What is that really doing for us?
> 
> > 
> 
> > The organization billing tests hit it.
> 
> 
> 
> Derp, I probably should have phrased that differently. I meant to inquire what this interesting syntax was doing - and it looks like it's applying an `if` statement where an `if` isn't readily available, so it's creating an image string only when the condition applies.

See https://github.com/pypi/warehouse/pull/13646/commits/4950f9465eb0c6c8c020b374dba470150e21becd

PR #13737, opened 2023-05-23
test: suppress warnings during test

miketheman:
We expressly know we're using the testing OIDC interface, so no need to
warn ourselves about it.


miketheman:
@woodruffw This is what I came up with - it's not awesome, since it pulls the class far away from who uses it, but it prevents the import-time cycle.

woodruffw:
> This is what I came up with - it's not awesome, since it pulls the class far away from who uses it, but it prevents the import-time cycle.

I'm okay with it if you are 🙂 -- I haven't had time to resume the other hacking on import cycles, so IMO this is expedient in terms of getting these suppressed.

di (approved):

PR #13744, opened 2023-05-24
move all record_event instances to send a request

ewdurbin:
One more step towards #8158, updates all `record_event` calls to use a `request`.

### DIFF
--- a/tests/unit/accounts/test_forms.py
+++ b/tests/unit/accounts/test_forms.py
@@ -259,7 +259,7 @@ class TestLoginForm:
             get_user=lambda _: user,
             check_password=lambda userid, pw, tags=None: True,
             disable_password=pretend.call_recorder(
-                lambda user_id, reason=None, request=None: None
+                lambda user_id, request, reason=None: None
             ),
             is_disabled=lambda userid: (False, None),
         )
@@ -278,8 +278,8 @@ class TestLoginForm:
         assert user_service.disable_password.calls == [
             pretend.call(
                 1,
+                request,
                 reason=DisableReason.CompromisedPassword,
-                request=request,
             )
         ]
         assert send_email.calls == [pretend.call(request, user)]
--- a/tests/unit/accounts/test_forms.py
+++ b/tests/unit/accounts/test_forms.py
@@ -203,6 +203,7 @@ class TestLoginForm:
             pretend.call(
                 tag=EventTag.Account.LoginFailure,
                 ip_address=request.remote_addr,
+                request=request,
                 additional={"reason": "invalid_password"},
             )
         ]
@@ -258,7 +259,7 @@ class TestLoginForm:
             get_user=lambda _: user,
             check_password=lambda userid, pw, tags=None: True,
             disable_password=pretend.call_recorder(
-                lambda user_id, reason=None, ip_address="127.0.0.1": None
+                lambda user_id, reason=None, request=None: None
             ),
             is_disabled=lambda userid: (False, None),
         )
@@ -276,7 +277,9 @@ class TestLoginForm:
         assert form.password.errors.pop() == "Bad Password!"
         assert user_service.disable_password.calls == [
             pretend.call(
-                1, reason=DisableReason.CompromisedPassword, ip_address="1.2.3.4"
+                1,
+                reason=DisableReason.CompromisedPassword,
+                request=request,
             )
         ]
         assert send_email.calls == [pretend.call(request, user)]
@@ -846,6 +849,7 @@ class TestTOTPAuthenticationForm:
             pretend.call(
                 tag=EventTag.Account.LoginFailure,
                 ip_address=request.remote_addr,
+                request=request,
                 additional={"reason": "invalid_totp"},
             )
         ]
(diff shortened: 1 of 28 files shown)


tests/unit/accounts/test_forms.py
```diff
@@ -203,6 +203,7 @@ def test_validate_password_notok(self, db_session):
             pretend.call(
                 tag=EventTag.Account.LoginFailure,
                 ip_address=request.remote_addr,
```
miketheman, line 205: Did you also want to expand scope to encompass these kinds of changes?
```suggestion
                ip_address=request.ip_address.ip_address,
```

This particular test case is providing a local stub request, so that would have to change as well.

A thought I've been having throughout these refactors is that we're pretty-heavily stubbing a lot of things that we likely should be using fixtures for.

warehouse/packaging/services.py
```diff
@@ -369,28 +369,26 @@ def create_service(cls, context, request):
 
 @implementer(IProjectService)
 class ProjectService:
-    def __init__(self, session, remote_addr, metrics=None, ratelimiters=None) -> None:
+    def __init__(self, session, metrics=None, ratelimiters=None) -> None:
```
ewdurbin, line 372: biggest refactor here. removed remote_addr on service instantiation and made it required when calling the one method that uses it.

ewdurbin:
**tempted** to remove `ip_address` from the `record_event` signature altogether... but feel like that is itself another herculean PR.

tests/unit/accounts/test_services.py
```diff
@@ -464,19 +470,29 @@ def test_disable_password(self, user_service, reason, expected):
         [(True, None), (True, DisableReason.CompromisedPassword), (False, None)],
     )
     def test_is_disabled(self, user_service, disabled, reason):
+        request = pretend.stub(
+            remote_addr="127.0.0.1",
+            ip_address=IpAddressFactory.create(ip_address="127.0.0.1"),
```
miketheman, line 475: If we never assert the value of the created IP, allow the factory to generate a random one?

```suggestion
            ip_address=IpAddressFactory.create(),
```

Similar note applies to other cases in this module.

di:
> tempted to remove ip_address from the record_event signature altogether... but feel like that is itself another herculean PR.

We should definitely do this, but in a separate PR.

ewdurbin:
yes, yes, I omitted "in this pr" from that sentence

warehouse/accounts/services.py
```diff
@@ -300,13 +300,14 @@ def update_user(self, user_id, **changes):
 
         return user
 
-    def disable_password(self, user_id, reason=None, ip_address="127.0.0.1"):
+    def disable_password(self, user_id, reason=None, request=None):
```
miketheman, line 303: **thought:** ‏the calls to `user.record_event` ought to fail  with an AttributeError of `request` is None here when calling for `request.remote_addr`

I _think_ that can only happen today where an Admin view doesn't pass `request` (or really `ip_address`) when calling this function. It appears that you've updated the caller, so maybe it's time to make `request` non-optional and remove the default?

di (approved):

warehouse/events/models.py
```diff
@@ -130,28 +130,19 @@ def record_event(
         """Records an Event record on the associated model."""
         session = orm.object_session(self)
 
-        if request is not None:
```
miketheman, line 133: **suggestion:** ‏update the function signature to remove `= None` from the request kwarg, and thus no longer allow non-request-passing calls to `record_event`

warehouse/packaging/services.py
```diff
@@ -369,28 +369,26 @@ def create_service(cls, context, request):
...
         self._metrics = metrics
 
-    def _check_ratelimits(self, creator):
+    def _check_ratelimits(self, request, creator):
         # First we want to check if a single IP is exceeding our rate limiter.
-        print(self.ratelimiters)
-        if self.remote_addr is not None:
-            if not self.ratelimiters["project.create.ip"].test(self.remote_addr):
+        if request.remote_addr is not None:
+            if not self.ratelimiters["project.create.ip"].test(request.remote_addr):
```
miketheman, line 383: The next refactor is likely to replace this with hashed_ip_address values, but that's next time around.

miketheman (approved):
Nothing blocking, but some improvements suggested.

tests/unit/accounts/test_forms.py
ewdurbin: next PR will probably be to remove ip_address from the call signature, so they'll go away soon. very soon.

warehouse/events/models.py
ewdurbin: b683ea4ad

warehouse/packaging/services.py
ewdurbin: we're a ways (couple prs and a big plan) off from the big switch, but yeah.

tests/unit/accounts/test_services.py
ewdurbin: 3e752256f

warehouse/accounts/services.py
ewdurbin: 6801f5541

PR #13745, opened 2023-05-24
feat: use geoip info for display

miketheman:
Implementation of pulling the GeoIP info off `Event.additional`

### DIFF
--- a/tests/unit/organizations/test_models.py
+++ b/tests/unit/organizations/test_models.py
@@ -183,14 +183,14 @@ class TestOrganization:
             key=lambda x: x[1],
         )
 
-    def test_record_event(self, db_request):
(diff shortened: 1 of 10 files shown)


tests/unit/organizations/test_models.py
```diff
@@ -183,14 +183,14 @@ def test_acl(self, db_session):
             key=lambda x: x[1],
         )
 
-    def test_record_event(self, db_request):
+    def test_record_event_with_geoip(self, db_request):
```
miketheman, line 186: **thought:** ‏for a future refactor, I'm thinking these test cases should be split out to events/test_models.py, but I wanted to defer that for now, since there's some complexity in creating a subclassed fakeish db model

ewdurbin (approved):

PR #13790, opened 2023-05-26
Call out deprecated fields of APIs

### LINKED ISSUE #13788: Call out deprecated fields of APIs
Users of warehouse's API will better understand how to handle the returned data.

The API reference on warehouse.pypa.io clearly calls out deprecated fields.

**Additional context**
With the change to `has_sig` this field is essentially deprecated, but the documentation doesn't show this. The only way to know that this field can safely be ignored by clients is to check the source.

Similarly, `downloads` can safely be ignored by clients, but there is no mention of this in the docs.

`releases` _is_ called out in the docs as deprecated, and that clients should begin to not rely on it even though it is still serving actionable data.

My open questions:
* these requested changes are geared toward the JSON API. Should the XML API receive similar call-outs? The fields also exist there and also do not return actionable data, but the entire API guide is cautioned as deprecated.
* are there any other fields that should be marked as deprecated that are not as obvious?

ucodery:
Fixes #13788

### DIFF
--- a/docs/dev/api-reference/json.rst
+++ b/docs/dev/api-reference/json.rst
@@ -14,6 +14,27 @@ Project
     In the future, the ``releases`` key may be removed from this response.
 
 
+.. attention::
+    The ``downloads`` key on this response should be considered deprecated.
+
+    It currently always returns ``-1`` and in the future, the ``downloads``
+    key may be removed from this response.
+
+
+.. attention::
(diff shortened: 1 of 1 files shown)


dstufft (approved):
Looks great, I'll merge this in a bit.

dstufft:
Thanks! ✨

PR #13793, opened 2023-05-27
Enrich security history displays with User-Agent summary.

ewdurbin:
Displays a summarize User-Agent if we have stored one along side the event.

Also harmonizes all the various user-facing event views.


ewdurbin:
additionalinfo

miketheman (approved):

Issue #2746, opened 2018-01-08
Allow search by username on collaborator input


nlhkabu:
In the 'collaborator' page, we have the ability to add another Warehouse user to your project via their usernname:

![screenshot from 2018-01-08 18-19-02](https://user-images.githubusercontent.com/3323703/34691286-14c570fa-f4b4-11e7-9ade-a28f30a056fb.png)

It would be awesome if we could make this search the database of existing users, as in GitHub:

![screenshot from 2018-01-08 18-15-53](https://user-images.githubusercontent.com/3323703/34691190-bfe90998-f4b3-11e7-80b5-a182da202985.png)

As discussed with @di, this is certainly not necessary for the MVP, so I will add it to the 'cool but not urgent' milestone.  If we want to move it up later, we can.

PR #13802, opened 2023-05-29
feat: support username autocomplete in forms

miketheman:
When adding a collaborator to a project or organization, support typing
in a few characters to get the top 10 results.

Uses https://www.npmjs.com/package/stimulus-autocomplete

Resolves https://github.com/pypi/warehouse/issues/2746

Includes lint command fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,8 @@
         "date-fns": "^2.30.0",
         "debounce": "^1.2.1",
         "jquery": "^3.7.0",
-        "normalize.css": "^8.0.1"
+        "normalize.css": "^8.0.1",
+        "stimulus-autocomplete": "^3.1.0"
       },
       "devDependencies": {
         "@babel/core": "^7.21.3",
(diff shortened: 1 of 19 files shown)


miketheman:
What it looks like (I'm no CSS wizard, but I did what I thought worked out)

https://github.com/pypi/warehouse/assets/529516/3bab0390-6be9-4c0d-ab73-62754f6d276c

di (commented):
Wow, this is super nice! I like how simple Stimulus makes this.

The only thing I'd suggest is adding a rate limiter around the `accounts_search` view, maybe 100 req/hr per IP? (Also, I'm curious whether the autocomplete will degrade gracefully with a `429` response)

di:
Also, looking at some prior art, we might want to crib adding avatars and full names (if available) from GitHub:

image

Maybe even shortening it to 5 users?

miketheman:
> The only thing I'd suggest is adding a rate limiter around the `accounts_search` view, maybe 100 req/hr per IP? (Also, I'm curious whether the autocomplete will degrade gracefully with a `429` response)

I'm not really sure how we do view-based rate limiters today - any pointers? It seems like the majority of the ratelimiters we have require some level of interaction with a service? 
We require the user to be logged in, so that's not *nothing*.

> Maybe even shortening it to 5 users?

I like adding names if they've got em. And our gravatar implementation calls our camo, so I'll add that, but can't confirm that the images are correct since my local camo doesn't work to proxy to their service for some reason? (haven't dug too far into it).

On the list design, GitHub fetches 10, but shows the first 5 and scrolls. I don't think I'm going to tackle building a popup/overlaid list with scroll. Happy to change to 5 instead if you prefer.

di:
Here's an example of how the ratelimiters could be used: https://github.com/pypi/warehouse/blob/d1b1581622090ea3fc60d777160547d5e0bfb284/warehouse/accounts/views.py#L1349-L1362

di (approved):

PR #13804, opened 2023-05-29
feat: add ip address backfill cli command

miketheman:
To support the eventual removal of `*Event.ip_address_string`, create a CLI command that provides the operator the ability to run the necessary queries to associate the relationship between any existing UserEvent and an IpAddress record.

Related to #8158

Once merged, relies on an operator to execute:

```shell
python -m warehouse hashing backfill-ipaddrs
```
which should backfill 10,000 events at a time.

Has toggles for `--batch-size` and `--sleep-time`, as well `--continue-until-done` once we're happy with the batch size and sleep time impact on the system overall.

**Note:** Need to update/modification for other Events: `File.Event`, `Project.Event`, `Organization.Event`, `Team.Event`

### DIFF
--- a/warehouse/cli/hashing.py
+++ b/warehouse/cli/hashing.py
@@ -29,13 +29,6 @@ def hashing():
 
 @hashing.command()
 @click.option(
-    "-s",
-    "--salt",
-    prompt=True,
-    hide_input=True,
-    help="Pass value instead of prompting for salt",
-)
-@click.option(
     "-b",
     "--batch-size",
     default=10_000,
@@ -58,7 +51,6 @@ def hashing():
 @click.pass_obj
 def backfill_ipaddrs(
     config,
-    salt: str,
     batch_size: int,
     sleep_time: int,
     continue_until_done: bool,
@@ -73,6 +65,8 @@ def backfill_ipaddrs(
     # This lives in the outer function so we only create a single session per
     # invocation of the CLI command.
     session = Session(bind=config.registry["sqlalchemy.engine"])
+
+    salt = config.registry.settings["warehouse.ip_salt"]
 
     _backfill_ips(session, salt, batch_size, sleep_time, continue_until_done)
 
--- a/warehouse/cli/hashing.py
+++ b/warehouse/cli/hashing.py
@@ -1,0 +1,152 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


warehouse/cli/hashing.py
```diff
@@ -0,0 +1,152 @@
...
+    session.add_all(no_ip_obj_rows)
+    session.commit()
+
+    # If there are more rows to backfill, recurse until done
+    if continue_until_done and how_many == batch_size:
+        click.echo(
+            f"Backfilled {batch_size} rows. Sleeping for {sleep_time} second(s)..."
+        )
+        time.sleep(sleep_time)
+        _backfill_ips(
```
dstufft, line 143: If we have more than 10,000,000 entries that need backfilled this will cause an error, but we've already committed the session so we won't lose progress, so can just run it again.

dstufft (approved):

warehouse/cli/hashing.py
```diff
@@ -0,0 +1,152 @@
...
+def hashing():
+    """
+    Run Hashing operations for Warehouse data
+    """
+
+
+@hashing.command()
+@click.option(
+    "-s",
+    "--salt",
```
ewdurbin, line 33: Isn't this already available in the configuration? I think its probably best to grab it from there rather than open the door for operator error by pasting the wrong thing.

ewdurbin (changes requested):

warehouse/cli/hashing.py
miketheman: Refactored in recent commit.

dstufft:
Python defaults to a max recursion depth of 1000, and this does batches of 10000. Sent from my iPhoneOn May 30, 2023, at 2:27 PM, Mike Fiedler ***@***.***> wrote:﻿
@miketheman commented on this pull request.



In warehouse/cli/hashing.py:
> +            )
+        # Associate the IPAddress object with the Event
+        row.ip_address_obj = ip_addr
+        session.add(ip_addr)
+
+    # Update the rows with any new IPAddress objects
+    session.add_all(no_ip_obj_rows)
+    session.commit()
+
+    # If there are more rows to backfill, recurse until done
+    if continue_until_done and how_many == batch_size:
+        click.echo(
+            f"Backfilled {batch_size} rows. Sleeping for {sleep_time} second(s)..."
+        )
+        time.sleep(sleep_time)
+        _backfill_ips(

Good note - where does the 10m entries limit come from?

warehouse/cli/hashing.py
ewdurbin: I am also curious of this.

ewdurbin (approved):

ewdurbin:
Just as a sanity check, I manually applied the python hashing to some records already populated in the DB via CDN and found the hashes to match! 🎉

PR #13823, opened 2023-05-31
feat(cli): add input to select event type

miketheman:
Now operator MUST pass `-e` or `--event-type` and one of the acceptable choices of `["user", "project", "file", "organization", "team"]`

### DIFF
--- a/warehouse/cli/hashing.py
+++ b/warehouse/cli/hashing.py
@@ -29,6 +29,13 @@ def hashing():
 
 @hashing.command()
 @click.option(
+    "-e",
+    "--event-type",
+    type=click.Choice(["user", "project", "file", "organization", "team"]),
+    required=True,
+    help="Type of event to backfill",
+)
+@click.option(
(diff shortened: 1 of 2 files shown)


warehouse/cli/hashing.py
```diff
@@ -28,6 +28,13 @@ def hashing():
 
 
 @hashing.command()
+@click.option(
+    "-e",
+    "--event-type",
+    type=click.Choice(["user", "project", "file", "organization", "team"]),
```
ewdurbin, line 34: question: how wild would it be to make this dynamic by seeing what inherits from HasEvents?

ewdurbin (approved):

warehouse/cli/hashing.py
miketheman: it'd be wild, since we'd need to know about what models have events, and our CLI will not know about models until execution, so we don't import any code during the cli resolution itself.

PR #13831, opened 2023-06-01
2FA basic auth notification email

di:
This updates #13830 so that the 2FA notification email on upload to still sends, with updated content.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -35,6 +35,7 @@ from wtforms.validators import ValidationError
 
 from warehouse.admin.flags import AdminFlag, AdminFlagValue
 from warehouse.classifiers.models import Classifier
+from warehouse.errors import BasicAuthTwoFactorEnabled
 from warehouse.forklift import legacy
 from warehouse.metrics import IMetricsService
 from warehouse.packaging.interfaces import IFileStorage, IProjectService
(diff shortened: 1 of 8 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -2593,8 +2594,17 @@ def test_upload_succeeds_with_2fa_enabled(
             IMetricsService: metrics,
         }.get(svc)
 
-        legacy.file_upload(db_request)
+        with pytest.raises(BasicAuthTwoFactorEnabled) as excinfo:
```
ewdurbin, line 2597: probably change name of this test now

ewdurbin (approved):
I'm iffy on breaking this out from BasicAuth policy, but for user ergonomics in the short term. Makes sense.

Approving conditionally. Let's open a PR that reestablishes enforcement via auth policy and drops the email. Merge it in a week or two?

tests/unit/forklift/test_legacy.py
di: Eh? Is the change to `test_basic_auth_upload_fails_with_2fa_enabled` not sufficient?

di:
> Let's open a PR that reestablishes enforcement via auth policy and drops the email. Merge it in a week or two?

Agreed this should eventually be part of the policy, but I think that should probably happen when enforcement is global and not conditional for some users/projects.

tests/unit/forklift/test_legacy.py
ewdurbin: image

I guess weird folding in UI

PR #13842, opened 2023-06-02
refactor: remove calls to ip_address_string

miketheman:
# Only Merge After Backfill Complete

Only merge this after backfill is complete on 5 tables:

- [x] `file_events`
- [x] `organization_events`
- [x] `project_events`
- [x] `team_events`
- [x] `user_events`

Confirm each case with query confirming no null values on `*_events.ip_address_id`
```sql
SELECT 'user_events' AS tbl, COUNT(*) FROM user_events WHERE ip_address_id IS NULL
UNION ALL
SELECT 'project_events' AS tbl, COUNT(*) FROM project_events WHERE ip_address_id IS NULL
UNION ALL
SELECT 'file_events' AS tbl, COUNT(*) FROM file_events WHERE ip_address_id IS NULL
UNION ALL
SELECT 'organization_events' AS tbl, COUNT(*) FROM organization_events WHERE ip_address_id IS NULL
UNION ALL
SELECT 'team_events' AS tbl, COUNT(*) FROM team_events WHERE ip_address_id IS NULL;
```
Output should resemble:
```sql
         tbl         | count
---------------------+-------
 user_events         |     0
 project_events      |     0
 file_events         |     0
 organization_events |     0
 team_events         |     0
(5 rows)
```

---

Now that every Event will have a backfilled IP Address object, we no
longer store, lookup any values from `ip_address_string`.

The logic for creating new IpAddress objects has been moved to
`request.ip_address` calls (inside `warehouse/utils/wsgi.py`).

A migration dropping the column will follow successful deployment.

- Updates tests to not pass `ip_address` to `*EventFactory`
- Removes backfill CLI command


miketheman:
Assigned to @ewdurbin to complete/confirm backfills completion.

tests/unit/accounts/test_models.py
```diff
@@ -102,23 +102,20 @@ def test_query_by_email_when_not_primary(self, db_session):
...
         stale_event = DBUserEventFactory(
             source=user,
             tag="bar",
-            ip_address="0.0.0.0",
             time=datetime.datetime.now() - datetime.timedelta(days=91),
         )
 
         assert user.events.all() == [recent_event, legacy_event, stale_event]
         assert user.recent_events.all() == [recent_event, legacy_event]
-        assert user.recent_events.all()[-1].ip_address == "0.0.0.0"
```
miketheman, line 121: This assertion was removed since the `db_session` doesn't have the `request` object, and we're also not using the `record_event()` method that would make the IP address available.
We _could_ also remove/rename the concept of `legacy_event`, since there's no real difference between the two structurally.

ewdurbin (approved):

PR #13844, opened 2023-06-02
Include all projects a User has access to in Add API token view

### LINKED ISSUE #13840: Cannot create scoped token to upload  organisation package, but personal general token works.
- IPython has recently been moved to Jupyter Org. 
- It now requires 2FA to publish. 
- I tried publishing with username password, got refused as I need token. Make sens. 
- Go to https://pypi.org/manage/project/ipython/settings/ click "create token for IPython". 
- Get redirected to "https://pypi.org/manage/account/token/"
- Give a names and try to select Scope (IPython not present). 

Question 1)
1) How am I supposed to create an IPython scope token ? Should I have been redirected to maybe `https://pypi.org/manage/org/token/` ? Or is a personal token used to upload and IPython scope should be present ?

- Curious I try to create a token will all permissions. 
- Try it. 
- twine upload works.
- `<pickachu face>`

Question 2) 
- Was i suppose to be able to upload if I can't create a scope token ? 
- Is some of the logic wrong ?

Also uploading to  https://upload.pypi.org/legacy/ returned HTML though it's an API endpoint, it maybe return something machine/terminal readable ?

ewdurbin:
resolves #13840 (in the ugliest way possible)

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -267,8 +267,8 @@ msgid "You are now ${role} of the '${project_name}' project."
 msgstr ""
 
 #: warehouse/accounts/views.py:1438 warehouse/accounts/views.py:1458
-#: warehouse/accounts/views.py:1593 warehouse/manage/views/__init__.py:1225
-#: warehouse/manage/views/__init__.py:1244
+#: warehouse/accounts/views.py:1593 warehouse/manage/views/__init__.py:1238
+#: warehouse/manage/views/__init__.py:1257
 msgid ""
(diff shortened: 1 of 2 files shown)


di:
I feel like there are probably other places where we use `user.projects` where we want the same behavior, should we fix this in a more broad way instead?

ewdurbin:
probably, I'm not married to this solution! just wanted to get something out there.

might be good to buff up `user_projects` from https://github.com/pypi/warehouse/blob/main/warehouse/manage/views/view_helpers.py

firefly9966 (approved):

Issue #11787, opened 2022-07-08
Implement admin tools for account recovery


di:
To make it easier for PyPI admins & moderators to process account recovery requests, there are few things we can do to make it possible for any admin/moderator to handle these requests and make it self-service as much as possible.

- [ ] Codify the initial process of sending an email to initiate the recovery process (as described [here](https://github.com/pypa/pypi-support/issues/796#issuecomment-1100081874))
- [ ] Summarizing recent actions on the User account and their projects
- [ ] Summarizing download counts for the user's project to determine criticality
- [ ] Some link to allow the affected user to notify admins that the initial steps have been completed
- [ ] Exposing the functionality for disabling 2FA in the admin (including issuing User events) so this doesn't require prod DB access
- [ ] Take into account critical projects & maintainers, possibly requiring admin intervention in these cases.

Regardless of improvements, we'll still require that a human in the loop does the final review before resetting access to the account. We'll also need to handle edge cases where a project doesn't have a public source repository.

Open questions:
- Keep the original requests in the GitHub tracker? This helps with associating w/ a GitHub account, but strongly ties us to GitHub
  - Possibly explore generating these in PyPI based on a GitHub action in that repository 
- Possibly: drop all roles that aren't sole owner, notify co-maintainers that they will need to vet and re-add?
- How do we establish & document a shared understanding of what requires further inspection
- Should we establish tiers? e.g. Tier 0: you have no projects. Tier 1: you have sole-owned projects that aren't widely used. Tier 2: you are co-maintainer on other projects. Tier 3: You are primary maintainer of a critical project.

ref: https://github.com/pypa/pypi-support/issues/796

di:
After chatting with @ewdurbin:
Two things that have the biggest impact:codifying the initial process of sending the email (with some human input to select the out-of-band URL used for proof of ownership), and an admin button for disabling all 2FA, recovery codes for the user.

di:
After chatting with @ewdurbin: Two things that have the biggest impact:

* codifying the initial process of sending the email (with some human input to select the out-of-band URL used for proof of ownership)
* an admin button for disabling all 2FA, recovery codes for the user.

woodruffw:
> an admin button for disabling all 2FA, recovery codes for the user.

Doing this now.

woodruffw:
> * [ ]  Notify co-maintainers, publish an event (possibly publicly)

Should this be another set of emails? If so, I can add it to https://github.com/pypi/warehouse/issues/13234.

[...]

PR #13848, opened 2023-06-03
admin: add a "wipe factors" button

woodruffw:
This behaves like the "reset password" button,
except that it additionally wipes the user's second factors and recovery codes.

Needs tests.

See https://github.com/pypi/warehouse/issues/11787#issuecomment-1279147003.

### DIFF
--- a/warehouse/admin/views/users.py
+++ b/warehouse/admin/views/users.py
@@ -296,6 +296,41 @@ def user_reset_password(user, request):
 
 
 @view_config(
+    route_name="admin.user.wipe_factors",
+    require_methods=["POST"],
+    permission="admin",
+    has_translations=True,
+    uses_session=True,
+    require_csrf=True,
+    context=User,
+)
+def user_wipe_factors(user, request):
+    if user.username != request.matchdict.get("username", user.username):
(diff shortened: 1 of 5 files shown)


warehouse/admin/views/users.py
```diff
@@ -295,6 +295,41 @@ def user_reset_password(user, request):
...
+            request.route_path("admin.user.detail", username=user.username)
+        )
+
+    user.totp_secret = None
+    user.webauthn = []
+    user.recovery_codes = []
+
+    login_service = request.find_service(IUserService, context=None)
+    send_password_compromised_email(request, user)
+    login_service.disable_password(
```
ewdurbin, line 323: I think we want to initiate a password reset flow (including sending the email).

We should be able to just call user_reset_password to accomplish that.

woodruffw: Yeah, I just copied this directly from the body of `user_reset_password`, since I wasn't sure about calling a `view_config`-decorated function within another. If that's okay though, I can replace it!

woodruffw: 0ba6bc02

ewdurbin (approved):

PR #13849, opened 2023-06-03
Refactor Authorization

### LINKED ISSUE #13690: AuthZ: remove deprecated Pyramid 1.0 policies
Warehouse currently uses Pyramid 2.0 style "security policies" at the top level (via `warehouse.utils.security_policy.MultiSecurityPolicy`). However, these are (currently) just a thin wrapper for a few 1.0-style authorization policies:

* `TwoFactorAuthorizationPolicy`
* `MacaroonAuthorizationPolicy`

These should be merged into their corresponding security policies, both simplifying the codebase and removing a handful of deprecated API sites.

(Assigning myself for triage, but will reassign.)

dstufft:
This refactors our use of the legacy `IAuthorizationPolicy` to migrate fully onto `ISecurityPolicy`.

Important things to note:

- The `MultiSecurityPolicy` remembers what policy provided an identity, and will *only* dispatch to that policy to call `permits()`.
    - This works using a `RequestLocalCache` which will store a weak reference to the request and return the same results each time, this keeps the logic encapsulated inside `MultiSecurityPolicy`.
    - All of our security policies ultimately dispatch to Pyramid's ACL system, so permissions are the same across all the security policies except for the policy specific bits (Macaroons, 2FA, etc).
    - This keeps  each security policy focused, you no longer have to figure out a complex interaction between multiple authorization policies that wrap each other.
- 2FA enforcement is no longer it's own AuthZ policy, but part of the Basic Auth and Session policies.
- Any object that we use for `request.identity` now knows how to generate it's own principals, this keeps ACLs and Principals both "near" the objects that implement them.
- ~~Removed some `group:with_admin_dashboard_access` group and related permission that, as far as I can tell, were not used for anything and were being granted to every logged in user, which felt wrong and confusing to me.~~
- Removed the `group:with_admin_dashboard_access` and granted the `admin_dashboard_access` permission to our 3 admin groups.

Making this a draft because I want to do some manual testing tomorrow when I'm more awake.

Thanks to @tnytown for a great start with #13754!

Closes #13754
Fixed #13690

### DIFF
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -122,8 +122,29 @@ def _basic_auth_check(username, password, request):
     return False
 
 
+class UserSecurityPolicyMixin:
+    def permits(self, request, context, permission):
+        # It should only be possible for request.identity to be a User object
+        # at this point, and we only a User in these policies.
+        assert isinstance(request.identity, User)
+
+        # Dispatch to our ACL
+        # NOTE: These parameters are in a different order
+        res = self._acl.permits(context, principals_for(request.identity), permission)
+
+        # If our underlying permits allowed this, we will check our 2FA status,
+        # that might possibly return a reason to deny the request anyways, and if
+        # it does we'll return that.
+        if isinstance(res, Allowed):
+            mfa = _check_for_mfa(request, context)
+            if mfa is not None:
+                return mfa
+
+        return res
+
+
 @implementer(ISecurityPolicy)
-class SessionSecurityPolicy:
+class SessionSecurityPolicy(UserSecurityPolicyMixin):
     def __init__(self):
         self._session_helper = SessionAuthenticationHelper()
         self._acl = ACLHelper()
@@ -187,28 +208,9 @@ class SessionSecurityPolicy:
         # Handled by MultiSecurityPolicy
         raise NotImplementedError
 
-    def permits(self, request, context, permission):
-        # It should only be possible for request.identity to be a User object
-        # at this point, and we only a User in these policies.
-        assert isinstance(request.identity, User)
-
-        # Dispatch to our ACL
-        # NOTE: These parameters are in a different order
-        res = self._acl.permits(context, principals_for(request.identity), permission)
-
-        # If our underlying permits allowed this, we will check our 2FA status,
-        # that might possibly return a reason to deny the request anyways, and if
-        # it does we'll return that.
-        if isinstance(res, Allowed):
-            mfa = _check_for_mfa(request, context)
-            if mfa is not None:
-                return mfa
-
-        return res
-
 
 @implementer(ISecurityPolicy)
-class BasicAuthSecurityPolicy:
+class BasicAuthSecurityPolicy(UserSecurityPolicyMixin):
     def __init__(self):
         self._acl = ACLHelper()
 
@@ -245,25 +247,6 @@ class BasicAuthSecurityPolicy:
     def authenticated_userid(self, request):
         # Handled by MultiSecurityPolicy
         raise NotImplementedError
-
-    def permits(self, request, context, permission):
-        # It should only be possible for request.identity to be a User object
-        # at this point, and we only a User in these policies.
-        assert isinstance(request.identity, User)
-
-        # Dispatch to our ACL
-        # NOTE: These parameters are in a different order
-        res = self._acl.permits(context, principals_for(request.identity), permission)
-
-        # If our underlying permits allowed this, we will check our 2FA status,
-        # that might possibly return a reason to deny the request anyways, and if
-        # it does we'll return that.
-        if isinstance(res, Allowed):
-            mfa = _check_for_mfa(request, context)
-            if mfa is not None:
-                return mfa
-
-        return res
 
 
 def _check_for_mfa(request, context) -> WarehouseDenied | None:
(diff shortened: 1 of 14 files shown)


warehouse/accounts/security_policy.py
```diff
@@ -16,13 +16,14 @@
     SessionAuthenticationHelper,
     extract_http_basic_credentials,
 )
+from pyramid.authorization import ACLHelper
 from pyramid.httpexceptions import HTTPUnauthorized
-from pyramid.interfaces import IAuthorizationPolicy, ISecurityPolicy
-from pyramid.threadlocal import get_current_request
```
ewdurbin, line 21: I bet this made you so happy @dstufft

warehouse/macaroons/security_policy.py
```diff
@@ -122,65 +125,40 @@ def authenticated_userid(self, request):
...
-        # allowed by a particular permission, they just restrict specific requests
-        # to not have that permission.
-        return self.policy.principals_allowed_by_permission(context, permission)
+        # It should not be possible to *not* have a macaroon at this point, because we
+        # can't call this function without an identity that came from a macaroon
+        assert isinstance(macaroon, str), "no valid macaroon"
+
+        # Check to make sure that the permission we're attempting to permit is one that
+        # is allowed to be used for macaroons.
+        # TODO: This should be moved out of there and into the acaroons themselves, it
```
ewdurbin, line 139:
```suggestion
        # TODO: This should be moved out of there and into the macaroons themselves, it
```

warehouse/macaroons/security_policy.py
```diff
@@ -122,65 +125,40 @@ def authenticated_userid(self, request):
...
+        try:
+            macaroon_service.verify(macaroon, request, context, permission)
+        except InvalidMacaroonError as exc:
+            return WarehouseDenied(
+                f"Invalid API Token: {exc}", reason="invalid_api_token"
+            )
+
+        # The macaroon is valid, so we can actually see if request.identity is
+        # authorized now or not.
+        # NOTE: These parameters are in a different order
```
ewdurbin, line 163:
```suggestion
        # NOTE: These parameters are in a different order than the signature of this method.
```

warehouse/accounts/security_policy.py
```diff
@@ -225,82 +229,77 @@ def authenticated_userid(self, request):
...
+    assert isinstance(request.identity, User)
 
-    def permits(self, context, principals, permission):
-        # The Pyramid API doesn't let us access the request here, so we have to pull it
-        # out of the thread local instead.
-        # TODO: Work with Pyramid devs to figure out if there is a better way to support
-        #       the worklow we are using here or not.
-        request = get_current_request()
+    # Dispatch to our ACL
+    # NOTE: These parameters are in a different order
```
ewdurbin, line 241:
```suggestion
    # NOTE: These parameters are in a different order than the signature of this method.
```

ewdurbin (approved):
Grossest diff ever, but this makes sense after four cups of coffee and a bit too much time switching between diffs and the files as they sit.

warehouse/accounts/security_policy.py
```diff
@@ -225,82 +229,77 @@ def authenticated_userid(self, request):
...
+        if mfa is not None:
+            return mfa
+
+    return res
+
+
+def _check_for_mfa(request, context) -> WarehouseDenied | None:
+    # It should only be possible for request.identity to be a User object
+    # at this point, and we only a User in these policies.
+    assert isinstance(request.identity, User)
```
miketheman, line 258: **thought:** ‏I thought using `assert` could be tricky, if the runtime ever chose to run under `-O` https://docs.python.org/3.11/using/cmdline.html#cmdoption-O
I don't think we're doing that today, but on the off chance we ever do run the warehouse with the flag (or envvar `PYTHONOPTIMIZE`), this would always return `None`. Wouldn't running under optimize mode effectively bypass this check?
`assert` is used another couple of times here, as well as other existing cases in `warehouse/macaroons/caveats/__init__.py` and `warehouse/http.py` so this isn't a new thing, but If You See Something, Say Something.

miketheman (approved):
I'm not a Pyramid security policy expert, but from what I could follow, this makes sense and simplifies the policy stack.

Should these changes (and the ones that follow) be represented somewhere in the dev docs? https://github.com/pypi/warehouse/blob/main/docs/dev/security.rst
Overall security approach/strategy for internal authN/authZ might be helpful to understand how the thing is meant to work. I think you wrote a fair amount already in Slack messages, and have some more ideas up your sleeve.

warehouse/accounts/security_policy.py
dstufft: It would technically bypass this check, but the outcome of doing that is we would just fail later with a more inscrutable error message. For example in this method we would fail trying to access `request.identity.has_two_factor``.

In all the places I used asserts, it should be fine to remove them because the structure of the code means that any other condition should not be possible, and would likely error out due to an assumption later.

So why do I have the assertions at all?

1. They serve as verifiable documentation for these functions that they are only valid to be called in cases where the assertion is true.
2. They teach type checkers like MyPy what type is being used, so it can do better linting than assuming `request.identity` is `Any`.
3. If we somehow get into a state where a non user is passed in, we'll just try to fail early on with a more reasonable error message.

(3) is the only one really affected by `PYTHONOPTIMIZE`, and it's the weakest reason of the bunch. It would also likely only be something we do in prod so if we started violating these assumptions we'd likely notice it in development first.

The alternatives here would be to do nothing and not assert (which is what most Python code bases do, they just assume that you've passed in a type that looks like what they expect, e.g. duck typing) or explicitly do type checking with a real error. The latter has all of the same benefits really, but we end up having to test impossible conditions then, for something that shouldn't ever happen.

PR #13853, opened 2023-06-04
fix: actually call for the publisher_url

miketheman:
Without the call, we surface broken links to the end user and admin interfaces.


miketheman:
FYI @tnytown

ewdurbin (approved):

PR #13856, opened 2023-06-04
Invalidate the security policy cache on login/logout

### LINKED ISSUE #13855: WebAuthn currently broken
#13849 introduced a bug in Two-factor signin with Security devices (WebAuthn).

dstufft:
Part of the changes in #13849 was that we would cache the result of trying to get the identity from our `ISecurityPolicy`, since the identity should not generally change mid-request, which allowed us to ensure that the same policy that fetched the identity was also the same policy that checks permissions on the identity.

Unfortunately there is a case where the identity *does* change mid request, when the user is being logged out or logged in, which we didn't handle.

This caused errors whenever, within the same request, we expected `request.identity` to return the now logged in or now logged out user. 

To fix this, we add an extension to the `ISecurityPolicy` class where the login and logout functions can reset the cache for a given request, similar to how it invalidates sessions when crossing the authentication boundaries.

It's possible that rather than an extension, we could piggyback off of `remember()` and `forget()`, but I don't feel comfortable doing that without sitting down and thinking through the implications.

Fixes #13855

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -401,6 +401,13 @@ class TestLogin:
         )
         pyramid_request.session.record_password_timestamp = lambda timestamp: None
 
+        security_policy = pretend.stub(
+            authenticated_userid=lambda r: None,
+            remember=lambda r, u, **kw: [],
+            reset=pretend.call_recorder(lambda r: None),
+        )
+        pyramid_request.registry.queryUtility = lambda iface: security_policy
+
         form_obj = pretend.stub(
             validate=pretend.call_recorder(lambda: True),
             username=pretend.stub(data="theuser"),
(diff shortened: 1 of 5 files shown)


dstufft:
I guess the ultimate question is whether we think that the assumption in https://github.com/pypi/warehouse/pull/13849 was generally correct, that the `request.identity` for a request does not change mid-request, and thus can be cached, and is still valid.

This PR assumes that it was, but with one edge case where it wasn't and fixes that one edge case.

If we think that assumption was wrong, then we would want to fix this a different way.

ewdurbin (approved):

ewdurbin:
Within the lifetime of a request, this is the only case I can think where it makes sense for the identity to change... yeah.

Issue #13234, opened 2023-03-17
Send more emails


di:
PyPI sends a fairly low volume of event-based emails to users, which results in the occasional spam report causing an abnormally high overall % of complaints.

Some ideas of things we could send emails for, to increase our overall email send, that would be a net positive user benefit and not just considered noise:

- [ ] New releases (sent to all maintainers)
- [ ] New files added to old releases (sent to all maintainers)
- [ ] Login from new IP addresses (sent to primary email)
- [ ] New email added (sent to any other emails on the account)
- [ ] New primary email (sent to the old primary email)

dstufft:
Other ideas:

- Emails whenever a release gets yanked or unyanked
- All of the new release/file/yanked/etc emails should be able to be subscribed to by any user for that project.

aganders3:
I added and deleted a few API tokens today and was surprised to not get any emails about it.

di:
@aganders3, what's your PyPI username?

aganders3:
I am also [@aganders3](https://pypi.org/user/aganders3/) there.

di:
@aganders3 Sorry, thought we already had that one! Great point, I've added it to the list.

woodruffw:
Another email idea: we should periodically notify project owners about stale/unused API tokens, e.g. if a project has had releases in the past `N` months that *don't* use a particular token.

As a conditional sequence:

1. If the project has API tokens;
2. _and_ those API tokens haven't been used in the last `N` months;
3. _and_ a release been has done in the last `N` months (e.g. with a different API token or trusted publishing);
4. _then_ send an email notifying the owner of the API token + project owners that they have one or more stale tokens still registered.

This will be particularly helpful/useful as more projects roll out trusted publishing, since they may forget to delete the old tokens that trusted publishing replaces.

di:
https://github.com/pypi/warehouse/issues/11524 would be related to that as well.

miketheman:
#997 would help increase volume

woodruffw:
Triaging: we're doing a couple of these as part of STF-funded work.

I've asked @xBalbinus to start with emails for yanking and unyanking of releases.

woodruffw:
>  Emails whenever a release gets yanked or unyanked

Looks like this one was already done ~3 years ago: https://github.com/pypi/warehouse/blame/7c6c4cf16d9f46660f687123a44766fb775dcea7/warehouse/email/__init__.py#L901-L935

[...]

PR #13869, opened 2023-06-05
Send emails on login from new IP address, API token creation

tnytown:
See #13234.

- [x] Login from new IP address
- [x] Added API tokens

#### HTML

image

#### Text

image

#### HTML

image

#### Text

image


warehouse/accounts/security_policy.py
```diff
@@ -95,11 +98,17 @@ def _basic_auth_check(username, password, request):
                 )
 
             login_service.update_user(user.id, last_login=datetime.datetime.utcnow())
-            user.record_event(
+            has_seen_ip_before = login_service.seen_from_ip_before(
```
ewdurbin, line 101: I'm worried doing this for BasicAuth (aka uploads) will be way too noisy. Consider a user that only uploads via GitHub Actions, they may get a new notice every time a release workflow runs.

ewdurbin: Correct, this should be the only place where HTTP Basic Auth is allowed and is used solely for the uploads endpoint. 2FA users will not be able to use the BasicAuth method due to the `_check_mfa` that is called by the permits part, they will have to use a macaroon directly or via trusted publisher.

firefly9966 (approved):

PR #13871, opened 2023-06-06
tests, warehouse: per-provider OIDC admin flags

### LINKED ISSUE #13841: Admin flags to disable individual Trusted Publisher providers
With #13551 we will support multiple IdPs for Trusted Publishers. PyPI admins should be able to disable individual providers via Admin flags.

woodruffw:
Closes #13841.

CC @di: This should be helpful in your form/view work on https://github.com/pypi/warehouse/issues/13551 🙂

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -1449,7 +1449,7 @@ class ManageAccountPublishingViews:
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc():
             self.request.session.flash(
                 self.request._(
                     "Trusted publishers are temporarily disabled. "
@@ -1469,7 +1469,7 @@ class ManageAccountPublishingViews:
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc(AdminFlagValue.DISALLOW_GITHUB_OIDC):
             self.request.session.flash(
                 self.request._(
                     "Trusted publishers are temporarily disabled. "
@@ -1604,7 +1604,7 @@ class ManageAccountPublishingViews:
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc():
             self.request.session.flash(
                 self.request._(
                     "Trusted publishers are temporarily disabled. "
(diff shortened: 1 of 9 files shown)


warehouse/admin/flags.py
```diff
@@ -47,6 +49,13 @@ def notifications(self):
             .all()
         )
 
+    def disallow_oidc(self, flag_member=None):
+        global_disallow_oidc = self.enabled(AdminFlagValue.DISALLOW_OIDC)
+        if flag_member is None:
+            return global_disallow_oidc
+
+        return self.enabled(flag_member) or global_disallow_oidc
```
di, line 57: This is fairly confusing, lots of double-negatives here. I think this could become:

```suggestion
    def disallow_oidc(self, flag_member=None):
        return self.enabled(flag_member) or self.enabled(AdminFlagValue.DISALLOW_OIDC)
```

we'd need to make the following change to `enabled`:

```diff
    def enabled(self, flag_member):
-       flag = self.request.db.get(AdminFlag, flag_member.value)
+       flag = self.request.db.get(AdminFlag, flag_member.value) if flag_member else None
        return flag.enabled if flag else False
```

warehouse/manage/views/__init__.py
```diff
@@ -1366,7 +1366,7 @@ def delete_oidc_publisher(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc():
```
di, line 1369: Should this handle publisher-specific deletions?

warehouse/accounts/views.py
```diff
@@ -1604,7 +1604,7 @@ def delete_pending_oidc_publisher(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc():
```
di, line 1607: Should this handle publisher-specific deletions?

warehouse/admin/flags.py
woodruffw: That's much cleaner, thanks!

warehouse/manage/views/__init__.py
```diff
@@ -1249,7 +1249,7 @@ def add_github_oidc_publisher(self):
         if not self.oidc_enabled:
             raise HTTPNotFound
 
-        if self.request.flags.enabled(AdminFlagValue.DISALLOW_OIDC):
+        if self.request.flags.disallow_oidc(AdminFlagValue.DISALLOW_GITHUB_OIDC):
```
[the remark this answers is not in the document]
woodruffw: Makes sense to me -- this particular callsite is GitHub only, so I can make it clear that GitHub-based trusted publishers are disabled here.

warehouse/admin/flags.py
woodruffw: 7dce63c3e..5a56bc025

woodruffw:
@di More generally, what do you think about removing the `warehouse.oidc.enabled` feature flag? It should be redundant with the admin flag now that the feature has been rolled out. I can do this in this PR or with a follow-on.

di (approved):

PR #13872, opened 2023-06-06
Generalize trusted publishing emails

woodruffw:
This makes the `trusted-publisher-added` and
`trusted-publisher-removed` email structures a little more
generic, allowing them to be re-used over both
GitHub and Google publishers. Future publishers will require
additional accommodations.

See https://github.com/pypi/warehouse/issues/13551.

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -3627,7 +3627,7 @@ class TestManageAccountPublishingViews:
 
         assert db_request.session.flash.calls == [
             pretend.call(
-                "Registered a new publishing publisher to create "
+                "Registered a new pending publisher to create "
                 "the project 'some-project-name'.",
                 queue="success",
             )
--- a/warehouse/oidc/models/google.py
+++ b/warehouse/oidc/models/google.py
@@ -82,7 +82,7 @@ class GooglePublisherMixin:
         return "Google"
 
     def publisher_url(self, claims=None):
-        return "https://accounts.google.com"
+        return None
 
     @property
     def email_verified(self):
--- a/warehouse/oidc/models/google.py
+++ b/warehouse/oidc/models/google.py
@@ -78,6 +78,13 @@ class GooglePublisherMixin:
     ]
 
     @property
+    def publisher_name(self):
+        return "Google"
+
+    def publisher_url(self, claims=None):
+        return "https://accounts.google.com"
+
+    @property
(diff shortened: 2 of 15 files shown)


tests/unit/accounts/test_views.py
```diff
@@ -3627,7 +3627,7 @@ def test_add_pending_github_oidc_publisher(self, monkeypatch, db_request):
 
         assert db_request.session.flash.calls == [
             pretend.call(
-                "Registered a new publishing publisher to create "
+                "Registered a new pending publisher to create "
```
woodruffw, line 3630: NB: Noticed this small typo while working on this.

warehouse/oidc/models/google.py
```diff
@@ -77,6 +77,13 @@ def __lookup_no_sub__(klass, signed_claims: SignedClaims) -> Query | None:
         __lookup_no_sub__,
     ]
 
+    @property
+    def publisher_name(self):
+        return "Google"
+
+    def publisher_url(self, claims=None):
+        return "https://accounts.google.com"
```
di, line 85: For GitHub, this ends up being a link to a specific repository or SHA. I'm not sure this is useful for Google accounts. If these were always true Google accounts, we could link to `https://accounts.google.com/?authuser=user@gmail.com`, but I don't think that will work for service accounts.

di (approved):
LGTM aside from that one comment.

warehouse/oidc/models/google.py
woodruffw: Yeah, there may not be a great URL here. Perhaps we could link to Google's OIDC docs instead?

di: I think we should really only provide this if it's specific to the identity used. I think it's probably fine for this to return `None` for now, and for the templates to handle this accordingly.

woodruffw: Sounds good, will update.

woodruffw: Okay, updated -- I updated a handful of the uses of `publisher_url()` to prepare them for providers where it'll be `None`, but some rendering uses will require a more in-depth look with the view/form changes (e.g. the trusted publisher table on each user/project).

di:
@woodruffw You've got a failing test here now.

woodruffw:
> @woodruffw You've got a failing test here now.

Thanks, fixed!

PR #13878, opened 2023-06-06
[mkdocs] Configure mkdocstrings plugin to enable intersphinx inventory

chrysle:
This will result in an intersphinx-like inventory being generated for the PyPI docs. See the [mkdocstrings documentation](https://github.com/chrysle/warehouse/pull/new/configure-mkdocstrings) for more information.


di:
> This will result in an intersphinx-like inventory being generated for the PyPI docs. See the [mkdocstrings documentation](https://github.com/chrysle/warehouse/pull/new/configure-mkdocstrings) for more information.

This link doesn't go to the mkdocstrings documentation. Could you add more details about the issue this is solving?

> It is needed to generate the docs-user.txt requirements file newly using pip-compile, but I'm not running Python 3.11. What can I do?

See https://warehouse.pypa.io/development/patterns.html#adding-new-dependencies

chrysle:
Sorry, this is the link: https://mkdocstrings.github.io/usage/#cross-references-to-any-markdown-heading

I aimed to enable [intersphinx linking](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) for the PyPI docs, which would have enabled Sphinx-based documentations to use intersphinx links (relative links) for any docs page. These are much easier to look after than absolute links.

But according to mkdocstrings/mkdocstrings#508, this isn't possibe yet with the plugin I added. I'll look if I can get that implemented, and I think close this pull request until then.

PR #13899, opened 2023-06-09
refactor: replace URLType with Text

miketheman:
The underlying datatype for URLType is `Text` and we don't use any of the functionality from the type at all.

Introduced in #9512

We don't need to carry the library any further.


ewdurbin (approved):
👍🏼 once you fight the dependency gods

di:
@miketheman I fought the dependency fight so you don't have to: https://github.com/pypi/warehouse/pull/13861

PR #13900, opened 2023-06-09
bow wow

ewdurbin:


warehouse/forklift/legacy.py
```diff
@@ -1512,7 +1512,7 @@ def file_upload(request):
         "uploaded_via": file_data.uploaded_via,
         "upload_time": file_data.upload_time,
     }
-    if not request.registry.settings.get("warehouse.release_files_table") is None:
+    if request.registry.settings.get("warehouse.release_files_table") is not None:
```
dstufft, line 1515: Already improving the code base

di (changes requested):
I'm in favor, speed seems to be about the same, if not a little faster.

If we merge this, this PR should also remove `flake8` from our dependencies and the `flake8` section from `setup.cfg`.

di (approved):

firefly9966 (approved):

PR #13913, opened 2023-06-11
Remove ip_address from record_event calls

ewdurbin:
Follows up on #13842, drops the columns it deprecates and removes ip_address from the record_event call altogether.


di (approved):
😗👌

ewdurbin:
`before-2023-06-12-drop-ip-address-string-from-events` created

PR #13914, opened 2023-06-12
Use explicit `formdata` for `ConfirmPasswordForm`

di:
Missed in #13828. This fixes a user report that account deletion was failing.

When using `InputRequired`, the validator only considers `form.field.raw_data`, but when initialized with kwargs, only `form.field.data` is set, and the validator fails.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -133,7 +133,7 @@ msgstr ""
 msgid "Successful WebAuthn assertion"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


miketheman (approved):
Seems reasonable.
The test that claims to cover this behavior is here, despite not calling it directly.

https://github.com/pypi/warehouse/blob/56c161583447b4811f8be49b297b96fd70c385a7/tests/unit/manage/test_forms.py#L258-L281

How was it passing before?

PR #13926, opened 2023-06-12
fix call to record_event in send_email

ewdurbin:
https://python-software-foundation.sentry.io/share/issue/448f599d818843448416cf1fa45e2f3d/


tests/unit/email/test_init.py
```diff
@@ -308,10 +306,9 @@ class FakeUser:
             def __init__(self):
                 self.events = []
 
-            def record_event(self, tag, ip_address, request=None, additional=None):
+            def record_event(self, tag, request=None, additional=None):
```
ewdurbin, line 309: got us again!

Issue #12933, opened 2023-02-04
Project.last_serial failed to update.


ewdurbin:
Reported by a bandersnatch user in #12214, the `last_serial` value for a Project was out of sync with the `journals` table after what appears to have been an automated mass removal of releases.

```
warehouse=> select id, name, version, action, submitted_date from journals where name='jarvis-ui' order by id desc limit 10;
    id    |   name    | version |     action     |       submitted_date       
----------+-----------+---------+----------------+----------------------------
 16626043 | jarvis-ui | 0.3.3   | remove release | 2023-01-30 23:30:11.147139
 16626042 | jarvis-ui | 0.3.8   | remove release | 2023-01-30 23:30:10.755916
 16626041 | jarvis-ui | 0.4.0   | remove release | 2023-01-30 23:30:10.732833
 16626040 | jarvis-ui | 0.3.5   | remove release | 2023-01-30 23:30:10.69155
 16626039 | jarvis-ui | 0.3.1   | remove release | 2023-01-30 23:30:10.640855
 16626038 | jarvis-ui | 0.3.2   | remove release | 2023-01-30 23:30:10.541308
 16626037 | jarvis-ui | 0.2.5   | remove release | 2023-01-30 23:30:10.50915
 16626036 | jarvis-ui | 0.3.0   | remove release | 2023-01-30 23:30:10.527536
 16626035 | jarvis-ui | 0.4.8   | remove release | 2023-01-30 23:30:10.537667
 16626034 | jarvis-ui | 0.4.3   | remove release | 2023-01-30 23:30:10.518038
(10 rows)

warehouse=> select last_serial from projects where name='jarvis-ui';
 last_serial 
-------------
    16626041
(1 row)

```

Note that the `last_serial` column on the project record is not actually the latest serial. Seems the trigger that calls `maintain_project_last_serial()` for the `journals` table failed to update the project. I'll file a separate issue for that.

It has been corrected by calling a no-op update on the row for now.

```
warehouse=> begin;
BEGIN
warehouse=*> update journals set name=name where id=16626043;
UPDATE 1
warehouse=*> select * from journals where name='jarvis-ui' order by id desc limit 10;
    id    |   name    | version |     action     |       submitted_date       | submitted_by  |             submitted_from              
----------+-----------+---------+----------------+----------------------------+---------------+-----------------------------------------
[...]

Issue #11918, opened 2022-07-21
Formally deprecate journal events?


woodruffw:
Just wanted to record this proposal somewhere: the "Project journal" view has had a "deprecation coming soon" warning for a few years now:

![Screen Shot 2022-07-21 at 2 00 03 PM](https://user-images.githubusercontent.com/3059210/180282405-27259d35-128c-4843-bc13-d9cd311c6bae.png)

Users are told to visit the security history view instead, which is both generally more detailed (per event) and has more event types. That view has existed since 2019 (https://github.com/pypi/warehouse/pull/6339).

Given that the replacement has been around for a couple of years and that we've been warning about deprecation for as long, does it make sense to formally deprecate journal events and set a timeline for their removal?

divbzero:
Deprecating journal entries seems to make sense to me. The deprecation warning and new events framework has been in place for about 3 years ago (from your PR #6339) and we also have generic events to add events to any model we would want in the future (from @di’s PR #8324).

What action would be taken to “formally” deprecating the project journal? Would it be announcing a specific date or version when the project journal view would be removed?

To help inform this discussion, do we have insight into how much the project journal page (`/manage/project/{project.name}/journal/`) is used?

woodruffw:
> What action would be taken to “formally” deprecating the project journal? Would it be announcing a specific date or version when the project journal view would be removed?

One of the PyPI maintainers/admins can correct me, but I think it would involve at minimum:

* Changing the current view's language from "will be deprecated" to "is deprecated", with a specific removal date.
* Making an announcement on discuss.python.org's [packaging forum](https://discuss.python.org/c/packaging/14) describing the deprecation

And then, in terms of the removal itself, it should mostly just be a matter of removing all uses of `JournalEvent`, deleting the routes and views, and maybe redirecting the current "project journal" route to the "security history" view instead.

dstufft:
Is ``JournalEvent`` different than ``JournalEntry``?

dstufft:
Ok I see the context now, ``JournalEntry`` is what you're actually talking about.

The big thing that relies on the journal is the serial number is actually just the journal entry's ID number, so we would need to add something new to give the behavior that Bandersnatch et al rely on.

woodruffw:
> Is `JournalEvent` different than `JournalEntry`?

Huh, my response via email never appeared here...

Yeah, this was just me mixing up words 😅 -- `JournalEntry` is what I meant.

woodruffw:
[...]

Issue #4892, opened 2018-10-18
Root cause Bandersnatch hitting Stale Cached JSON API Requests


cooperlees:
**Describe the bug**
Summary: Bandersnatch installs hitting stale JSON API Data
Full Details: https://github.com/pypa/bandersnatch/issues/56
Purge Hack: https://github.com/pypa/bandersnatch/pull/57

**Expected behavior**
Bandersnatch not having to PURGE the cache.

**To Reproduce**
Unknown here, but I’d suggest lets add logging / collection of more data about Bandersnatch user-agent calling purges

**My Platform**
Bandersnatch with requests

igotcha:
we have the same problem here:

2018-10-23 16:30:58,066 INFO: bandersnatch/3.0.0.dev0 (cpython 3.6.6-final0, Linux x86_64)
2018-10-23 16:30:58,123 INFO: Syncing with https://pypi.python.org.
2018-10-23 16:30:58,124 INFO: Current mirror serial: 4344762
2018-10-23 16:30:58,124 INFO: Resuming interrupted sync from local todo list.
2018-10-23 16:30:58,188 INFO: Trying to reach serial: 4346516
2018-10-23 16:30:58,188 INFO: 7 packages to sync.
2018-10-23 16:30:58,190 INFO: Syncing package: basetuyaapi (serial 4346273)
2018-10-23 16:30:58,191 INFO: Syncing package: disparitySMD (serial 4345765)
2018-10-23 16:30:58,192 INFO: Syncing package: nixcms (serial 4346344)
2018-10-23 16:31:00,187 INFO: disparitySMD no longer exists on PyPI
2018-10-23 16:31:00,187 INFO: Syncing package: pacifica-python-downloader (serial 4346407)
2018-10-23 16:31:00,260 INFO: basetuyaapi no longer exists on PyPI
2018-10-23 16:31:00,260 INFO: Syncing package: portus (serial 4345746)
2018-10-23 16:31:01,468 INFO: nixcms no longer exists on PyPI
2018-10-23 16:31:01,468 INFO: Syncing package: python-todict (serial 4346125)
2018-10-23 16:31:01,686 INFO: portus no longer exists on PyPI
2018-10-23 16:31:01,687 INFO: Syncing package: toil-ionox0 (serial 4346027)
2018-10-23 16:31:02,318 INFO: pacifica-python-downloader no longer exists on PyPI
2018-10-23 16:31:02,941 INFO: python-todict no longer exists on PyPI
[...]

PR #13936, opened 2023-06-14
Monotonic journals

dstufft:
Our mirroring support currently assumes that the `journals.id` field is a monotonically increasing integer.

However, the way that `SERIAL` is implemented in PostgreSQL, this isn't actually true. The fetching of the next value from a `SERIAL` ignores the transaction. In addition, we have no control over what order the transactions will commit in, this means something like:

TXN A does an `INSERT`, gets `journals.id = 100`.
TXN B does an `INSERT`, gets `journals.id = 101`.
TXN B `COMMIT`, makes last serial = 101.
Bandersnatch makes a request, and notes that the last serial is 101.
TXN A `COMMIT`, the last serial is still 101.
Bandersnatch never sees 100.

What this does is wherever we create a `JournalEntry`, we first take an advisory lock (scoped to the transaction). This will have the effect of forcing journal entry insertion to be serialized, without affecting our table locking otherwise.

The downside to this is it can cause scalability concerns, but AFAIK it's the only real way we have to make sure that our last serial is monotonically increasing.

To try and reduce the negative effects, we also go through and stop emitting any journal event that our mirroring support doesn't mean. Specifically this is any journal event that isn't correlated with a change in the output of the `/simple/$name/` page.

Each "type" of journal event that we stop emitting is in it's own commit, so if we disagree with any of them, it should be easy to revert.

This will fix (hopefully) https://github.com/pypi/warehouse/issues/12214, https://github.com/pypi/warehouse/issues/4892, and https://github.com/pypi/warehouse/issues/12933.

This will move us closer to https://github.com/pypi/warehouse/issues/11918.

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -758,6 +758,20 @@ class JournalEntry(db.ModelBase):
 
     @classmethod
     def create_with_lock(cls, session, *args, **kwargs):
+        # We rely on `journals.id` to be a monotonically increasing integer,
+        # however the way that SERIAL is implemented, it does not guarentee
+        # that is the case.
+        #
+        # Ultimately SERIAL fetches the next integer regardless of what happens
+        # inside of the transaction. So journals.id will get filled in, in order
+        # of when the `INSERT` statements were executed, but not in the order
+        # that transactions were committed.
+        #
+        # The way this works, not even the SERIALIZABLE transaction types give
+        # us this property. Instead we have to implement our own locking that
+        # ensures that each new journal entry will be serialized.
+        #
+        # See: https://mattjames.dev/auto-increment-ids-are-not-strictly-monotonic/
         hashed = hashlib.blake2b(b"table:journals").digest()[:8]
         key = int.from_bytes(hashed, "little", signed=True)
         session.execute(select(func.pg_advisory_xact_lock(cast(key, BigInteger))))
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import enum
+import hashlib
 
 from collections import OrderedDict
 from urllib.parse import urlparse
@@ -36,8 +37,10 @@ from sqlalchemy import (
     Table,
     Text,
     UniqueConstraint,
+    cast,
     func,
     orm,
+    select,
     sql,
 )
 from sqlalchemy.dialects.postgresql import UUID
@@ -752,6 +755,14 @@ class JournalEntry(db.ModelBase):
         nullable=True,
     )
     submitted_by = orm.relationship(User, lazy="raise_on_sql")
+
+    @classmethod
+    def create_with_lock(cls, session, *args, **kwargs):
+        hashed = hashlib.blake2b(b"table:journals").digest()[:8]
+        key = int.from_bytes(hashed, "little", signed=True)
+        session.execute(select(func.pg_advisory_xact_lock(cast(key, BigInteger))))
+
+        return cls(*args, **kwargs)
 
 
 class ProhibitedProjectName(db.Model):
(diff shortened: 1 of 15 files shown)


ewdurbin:
Only concern here is making sure we emit a matching ProjectEvent anywhere we are removing a JournalEntry.

It looks like that's the case everywhere but the admin view (which is a past oversight).

dstufft:
I can emit the event in the admin view to fix that oversight!

di (approved):

di:
> Only concern here is making sure we emit a matching ProjectEvent anywhere we are removing a JournalEntry.
> 
> It looks like that's the case everywhere but the admin view (which is a past oversight).

Did this in 308ba3ada7f49287becbfbcf35ce017f46203c7a!

warehouse/packaging/models.py
```diff
@@ -859,6 +870,31 @@ def __table_args__(cls):  # noqa
     )
     submitted_by: Mapped[User] = orm.relationship(lazy="raise_on_sql")
 
+    @classmethod
+    def create_with_lock(cls, session, *args, **kwargs):
```
miketheman, line 874: My issue with this is that now we must remember that the right way to create a `JournalEntry` _must_ use `create_with_lock()`, or we risk creating ids out of order, and there's nothing to prevent someone from doing that, is there?

Had we looked at using the [event listeners we use today on other models](https://github.com/pypi/warehouse/blob/2e4117d127315b42f833939936e6c2e1f84908e7/warehouse/packaging/__init__.py#L38-L59), in this case it would probably be `before_insert`?

If implemented over there, we would take the lock automatically on any insert, thus preserving the normal creation methods we use today.

miketheman (commented):
I like the cleanup, especially for items that aren't super journal-oriented like adding owners and roles.
(I completely acknowledge that I've used journals as the audit trail for packages that have been removed and re-uploaded, since we nuke Events, but that's a different problem, and not part of journal's responsibility.)

I had one implementation question inline.

Another question I think we should answer is how might we do lock monitoring, since I don't think we do that today. If memory serves, a Datadog Postgres monitor could emit those metrics, but I don't think we have that Agent configuration set today.

warehouse/packaging/models.py
di: I haven't explored this. Just a note that if we want to go down that route, we'll probably want to do `batch=false` to avoid these being grouped: https://docs.sqlalchemy.org/en/20/orm/events.html#sqlalchemy.orm.MapperEvents.before_insert:~:text=The%20event%20is%20often%20called%20for%20a%20batch%20of%20objects%20of%20the%20same%20class%20before%20their%20INSERT%20statements%20are%20emitted%20at%20once%20in%20a%20later%20step

dstufft: I switched to a `before_insert` event and verified that the advisory lock query is still being executed prior to the `INSERT`.

I don't believe we need `batch=False`, because we're not actually operating on the individual objects, we just need to insert a single unrelated query _before_ the `INSERT` happens so we can take the advisory lock.

dstufft:
I've updated this pull request to use a `before_insert` hook and verified that the query taking the advisory lock is still being emitted prior to an `INSERT` of a Journal.

@di had previously added the events that @ewdurbin had asked for to ensure we're not losing auditability.

I *think* this should be good for review now.

warehouse/packaging/models.py
```diff
@@ -74,6 +83,8 @@
 if typing.TYPE_CHECKING:
     from warehouse.oidc.models import OIDCPublisher
 
+_MONOTONIC_SEQUENCE = 42
```
miketheman, line 86: I suspect this is a Meaning of Life reference? A comment would make it clearer why 42 was chosen.

miketheman (approved):
Yay! I had one non-blocking comment.

I'm still curious how this will operate at scale, and how we are likely to determine if we're getting lock contention. Is that something you've considered already?

PR #13941, opened 2023-06-14
make the `invalid-publisher` err msg more informative

kemingy:
- refer to https://github.com/pypa/gh-action-pypi-publish/issues/138#issuecomment-1591064887


tests/unit/oidc/test_views.py
```diff
@@ -180,7 +180,11 @@ def test_mint_token_from_trusted_publisher_lookup_fails():
         "errors": [
             {
                 "code": "invalid-publisher",
-                "description": "valid token, but no corresponding publisher",
+                "description": (
+                    "valid token, but no corresponding publisher, "
+                    "either this publisher is not configured in PyPI "
+                    "or this publisher required environment is not met"
+                ),
```
woodruffw, line 187: I suspect that this description will mislead users:

* It implies that the token didn't match a publisher specifically because it's missing an `environment` setting on the PyPI side, but we don't actually know that (it could have failed to match due to another claim);
* It implies that the environment claim is required, when it's explicitly optional (but has to match if provided);
* It mentions a detail about a *specific* trusted publisher implementation (GitHub) that won't be true for other trusted publishers being added (like GitLab and Google).

The larger problem here is that we don't have all of the context we need to tell the user _which_ claim is wrong, in part because there's no decision function that gives us that (e.g., it's not clear that we actually _should_ encourage a user to fix a particular claim in a particular, since it might be giving a publisher they don't intend access to their project).

woodruffw (changes requested):
PyPI admins are the ultimate arbiters here, but I'm personally a -1 on the changes to the error messages here: I think they introduce more ambiguity than they resolve, unfortunately.

IMO a workable solution here would be to add another bulletpoint to the troubleshooting docs explicitly mentioning that any environment configured on PyPI needs to match the CI job's environment, and then include the troubleshooting URL in the error render on the `gh-action-pypi-publish` side.

In other words: IMO this change should be docs-only, and should be accompanied by a separate PR to the `gh-action-pypi-publish` action that updates the error message to include a link to the troubleshooting docs.

docs/user/trusted-publishers/troubleshooting.md
```diff
@@ -51,7 +51,9 @@ endpoint:
   is well-formed (and has a valid signature), but doesn't match any known
   (pending) OIDC publisher. This likely indicates a mismatch between the
   OIDC publisher specified in the user/project settings and the claims
-  represented in the actual OIDC token. Check for typos!
+  represented in the actual OIDC token. Check for typos! If you're using
+  GitHub actions, check if the workflow is using the same environment
```
di, line 55:
```suggestion
  GitHub Actions, check if the workflow is using the same environment
```

docs/user/trusted-publishers/troubleshooting.md
```diff
@@ -51,7 +51,9 @@ endpoint:
...
+  as configured in the PyPI publishers.
```
di, line 56:
```suggestion
  as configured when the publisher was configured on PyPI.
```

docs/user/trusted-publishers/using-a-publisher.md
```diff
@@ -37,6 +39,8 @@ flow, your diff might look like this:
    pypi-publish:
      name: upload release to PyPI
      runs-on: ubuntu-latest
++    # **strongly** encouraged to protect it with a GitHub environment
```
di, line 42:
```suggestion
+    # Specifying a GitHub environment is optional, but strongly encouraged
```

docs/user/trusted-publishers/using-a-publisher.md
```diff
@@ -19,6 +19,8 @@ jobs:
   pypi-publish:
     name: upload release to PyPI
     runs-on: ubuntu-latest
+    # **strongly** encouraged to protect it with a GitHub environment
```
di, line 22:
```suggestion
    # Specifying a GitHub environment is optional, but strongly encouraged
```

di (approved):

PR #10870, opened 2022-03-04
TUF Initialization using python-tuf 2.0.0 

kairoaraujo:
This work refactors the [Draft PR](https://github.com/pypa/warehouse/pull/7488) by
@woodruffw, to build a new repository tool on top of the Python-TUF
Metadata API, and use it instead of the Python-TUF repository tool
that was deprecated in v1.0.0.

Part of https://github.com/pypa/warehouse/issues/10672

**Note to reviewer**

The current implementation has some development-only components, and lacks a few services for full PEP458 compliance as well as extensive tests. However, it should qualify for a review of the overall architecture and flow (see details in 'Overview' below). Components and functionality that are planned for subsequent PRs are listed in 'Next steps' below.

### Overview

  ```mermaid
    classDiagram
      direction LR
      class `tuf.interfaces` {
        zope.interface.Interface
        IKeyService(Interface)
        IStorageService(Interface)
        IRepositoryService(Interface)
      }
      class `tuf.services` {
        IKeyService
        IRepositoryService
        IStorageService
        LocalKeyService(IKeyService)
        LocalStorageService(IStorageService)
        RepositoryService(IRepositoryService)
      }
      class `tuf.tasks` {
        init_repository
        init_targets_delegation
        bump_snapshot
        bump_bin_n_roles
        add_hashed_targets
      }

      class `cli.tuf`{
          dev keypairs
          dev init-repo
          dev init-delegations
          dev add-all-packages
          dev add-all-indexes
          dev bump-snapshot
          dev bump-bin-n-roles
      }


      `tuf.services` <|-- `tuf.interfaces`
      `tuf.tasks` -- `tuf.services`
      `cli.tuf` -- `tuf.tasks`
      warehouse -- `cli.tuf`
      warehouse -- `tuf.tasks`
  ```

#### warehouse.tuf.repository

- ~~``MetadataRepository`` implements a custom TUF metadata repository tool on top of
the new Python-TUF Metadata API to create and maintain (update, sign, sync with storage) TUF metadata for Warehouse.~~

#### warehouse.tuf.services

- ``LocalKeyService`` provides a local file storage backend for TUF role keys used by the repository tool (development only!!).
- ``LocalStorageService`` provides a local file storage backend for TUF role metadata used by the repository tool.
- ``RepositoryService`` provides methods for common Warehouse-TUF tasks, using the repository tool.

#### warehouse.tuf.tasks

Defines common Warehouse-TUF tasks that use the `RepositoryService` for
[...]

PR #13943, opened 2023-06-15
PEP 458: RSTUF Integration

kairoaraujo:
This PR implements [PEP 458](https://peps.python.org/pep-0458) by adding a setup for [Repository Service for TUF (RSTUF)](https://github.com/repository-service-tuf/repository-service-tuf) and connecting Warehouse to it.

#### Context

Unlike previous attempts (#7488, #10870) to implement PEP 458, this PR does not deeply integrate TUF repository management into Warehouse, but instead configures a stand-alone RSTUF service, which maintains the TUF metadata repository as a black box, and which accepts calls to a REST API, so that Warehouse can indicate addition or removal of release files, and trigger TUF metadata changes.

Additionally, RSTUF provides a CLI for root signing in order to initialize the RSTUF metadata repository. 

See [RSTUF docs](https://repository-service-tuf.readthedocs.io/en/v1.0.0a1-draft/) for details.

#### Description of this PR

* Configure RSTUF (development instance)
  - Add development dependencies
  - Add RSTUF worker services to `docker-compose.yml` 
	 - RSTUF uses the same Redis Server,  but unique Redis DB ids `1` and `2`
	 - RSTUF uses the same PostgreSQL, but a specific rstuf database
  - Add a pre-generated bootstrap file, with RSTUF config and root of trust
  - Add RSTUF commands `Makefile`
	  - `make tufinit` to bootstrap the RSTUF service
	  - `make tufimport` to create TUF metadata for all existing release files from the example Warehouse database
              it is a make command to the added Warehouse cli command (`warehouse tuf dev import-all`)

* Add calls to RSTUF API upon package addition and removal

#### Status of RSTUF

RSTUF is close to releasing a beta version, called [Minimum Working Version (MWV)](https://repository-service-tuf.readthedocs.io/en/v1.0.0a1-draft/devel/roadmap.html#goal-minimum-working-version). Actually, two of three components (RSTUF [Worker](https://github.com/repository-service-tuf/repository-service-tuf-worker) and [API](https://github.com/repository-service-tuf/repository-service-tuf-api)) are already tagged MWV. The third component (RSTUF [CLI](https://github.com/repository-service-tuf/repository-service-tuf-cli)) is missing one feature, which is not relevant for this PR, and not expected to break compatibility for the MWV release.


warehouse/packaging/tasks.py
```diff
@@ -508,3 +509,9 @@ def populate_data_using_schema(file):
                 json_rows, table_name, job_config=LoadJobConfig(schema=table_schema)
             ).result()
             break
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def generate_projects_simple_detail(request, projects):
```
kairoaraujo, line 515: About this task

1. I need it to have the `request` object to be used by the `render_simple_detail` function. 
I couldn't find a way to retrieve the `request` , only using tasks.
You can see it is used during the process of the `warehouse tuf dev import-all` command https://github.com/kairoaraujo/warehouse/blob/706478ac2f375e6f2d03259ee75cba67c91f21ae/warehouse/cli/tuf.py#L136

2. I also intended to use it later during the **Manage packages/project and simple details to TUF** commit. Still, it wasn't possible because I could not send the simple detail generated in the background. I need the response to use in the TUF metadata.

If I can retrieve the `request` object or generate one for use with the `render_simple_detail`, we can remove it.

kairoaraujo:
#### Open questions / Next steps

* How to test 
* Documentation about how to roll out

warehouse/packaging/tasks.py
```diff
@@ -508,3 +509,9 @@ def populate_data_using_schema(file):
                 json_rows, table_name, job_config=LoadJobConfig(schema=table_schema)
             ).result()
             break
+
+
+@tasks.task(ignore_result=True, acks_late=True)
```
kairoaraujo, line 514: About this task

1. I need it to have the `request` object to be used by the `render_simple_detail` function. 
I couldn't find a way to retrieve the `request` , only using tasks.
You can see it is used during the process of the `warehouse tuf dev import-all` command https://github.com/kairoaraujo/warehouse/blob/706478ac2f375e6f2d03259ee75cba67c91f21ae/warehouse/cli/tuf.py#L136

2. I also intended to use it later during the **Manage packages/project and simple details to TUF** commit. Still, it wasn't possible because I could not send the simple detail generated in the background. I need the response to use in the TUF metadata.

If I can retrieve the `request` object or generate one for use with the `render_simple_detail`, we can remove it.

warehouse/tuf/targets.py
```diff
@@ -0,0 +1,91 @@
...
+        },
+    }
+
+
+def _handle(response):
+    """Helper to handle http response for POST or DELETE targets request."""
+    if response.status_code != 202:
+        raise HTTPBadGateway(f"Unexpected TUF Server response: {response.text}")
+
+    return response.json()
```
kairoaraujo, line 45: RSTUF, by design, is asynchronous.
Every request to add or remove an artifact in TUF metadata generates a task id.
With this task id is possible to do call backs to receive the status. RSTUF gives the control to monitor this task to the requester.

Some integration design questions about Warehouse
* Should the Warehouse create an asynchronous monitoring/action for the tasks to take some action depending on the result?

* Should Warehouse store the task result in some Warehouse table as in RSTUF use the default [Celery result expires]?  Currently, we store the task id in the events, but the task result is not persistent as mentioned, for example: https://github.com/pypi/warehouse/blob/818210ad8647bad2f29f4494fc4a7cf13d63995b/warehouse/forklift/legacy.py#L1410-L1427

kairoaraujo:
I'm tagging here some TUF folks.

@lukpueh -- Thanks for pre-draft-PR review ❤️ 
@joshuagl -- It's following the RSTUF work from the very beginning

Note: Feel free to tag others

lukpueh:
cc @justincappos @mnm678 @adityasaky @trishankatdatadog  @brainwane

warehouse/config.py
```diff
@@ -241,6 +241,12 @@ def configure(settings=None):
         coercer=int,
         default=100,
     )
+    maybe_set(settings, "tuf.database.url", "TUF_DATABASE_URL")
+    maybe_set(settings, "tuf.metadata.url", "TUF_METADATA_URL")
+    maybe_set(settings, "tuf.api.url", "TUF_API_URL")
```
simi, line 246: Would it make sense to rename this to `RSTUF_API_URL` since RSTUF is providing the API in here?

ewdurbin:
Thanks for this PR @kairoaraujo.

I think to make progress, we would want to deliver chunks of this iteratively rather than all at once.

1) Break out a separate PR that introduces a development mode `rstuf` container even if nothing talks to it... ideally that can be treated as much like a "blackbox" as possible. The current setup with api/worker/worker adds a lot of complexity the makes review difficult. This PR could also remove the un-used `vault` service in dev intended for the original tuf implementation.

2) A follow on PR that demonstrates integration with the `rstuf` service locally. This will help in review so that we can focus on warehouse's concerns rather than RSTUFs.

miketheman:
Superseded by #15241

PR #13948, opened 2023-06-15
fix: update migrations code for sqlalchemy 2.0

miketheman:
Follows #13929

Some changes that were not observed during dev/test:

- raw SQL now needs to be wrapped in `text()`
- if using an existing Connection, it doesn't have a `connnect()`
  method, so we configure the existing context to use it.
  Follows pattern from Alembic Cookbook
  https://alembic.sqlalchemy.org/en/latest/cookbook.html#sharing-a-connection-across-one-or-more-programmatic-migration-commands

### DIFF
--- a/warehouse/migrations/versions/4490777c984f_migrate_existing_data_for_release_is_.py
+++ b/warehouse/migrations/versions/4490777c984f_migrate_existing_data_for_release_is_.py
@@ -55,7 +55,7 @@ def upgrade():
                 """
             )
         )
-        conn.execute("COMMIT")
+        conn.execute(sa.text("COMMIT"))
 
     op.alter_column(
         "releases",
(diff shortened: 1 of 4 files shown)


warehouse/migrations/versions/4490777c984f_migrate_existing_data_for_release_is_.py
```diff
@@ -55,7 +55,7 @@ def upgrade():
                 """
             )
         )
-        conn.execute("COMMIT")
+        conn.execute(sa.text("COMMIT"))
```
miketheman, line 58: This surfaced during a `make initdb` vs running migrations. I chose to preserve the text vs `conn.commit()` since the function has other semantics applicable to the running connection.

ewdurbin (approved):

Issue #13935, opened 2023-06-14
Add confirmation gauntlet to release file deletion


ewdurbin:
When deleting a single file from a release, we should include similar acknowledgments to what we do for releases and projects.

While reviewing #13930 I recognized that we have one more irreversible user initiated deletion that should have a similar set of acknowledgements:

- [ ] I understand that my users will no longer be able to install this file for this release of this project.
- [ ] I understand that I will <a href="https://pypi.org/help/#file-name-reuse">not be able to re-upload a using the same file name</a> {{ file.filename }}.
- [ ] I understand that I will not be able to undo this.
- [ ] I understand that the PyPI administrators will not be able to undo this.

_Originally posted by @ewdurbin in https://github.com/pypi/warehouse/issues/13930#issuecomment-1588301970_

ewdurbin:
cc @mm if you're interested in implementing.

mm:
back for more! we love a fun checkbox adventure 🎉  I have this more or less working, though may need some styling fixes:

https://github.com/pypi/warehouse/assets/639901/ccc898ff-65ad-4989-8c78-4ebd7acab4db

To accomplish that `If the checkboxes are checked, AND the input string matches, let us proceed` logic, I took the approach of modifying the ConfirmController that modal uses to accept listening on checkboxes (if they're marked as `checkbox` targets) and the input string as before:

https://github.com/mm/warehouse/blob/3168a1bcb666148f3171eb39ba98f747ad2a5547/warehouse/static/js/warehouse/controllers/confirm_controller.js#L43

It might be re-usable for other modals in the future and I can open a PR with the changes once I get the test coverage up, but was curious on initial thoughts on extending it that way 🙂

ewdurbin:
I think that makes good sense! Thanks for digging in.

PR #13952, opened 2023-06-15
Add confirmation checks when deleting a file from a release

mm:
This PR addresses #13935 by adding additional checks the user has to go through to confirm the intent of their actions when deleting a file from a specific release. With these updates, a user has to check all the checkboxes within the file deletion modal, **and** confirm the project name before being able to delete the file.

To accomplish this, the `ConfirmController` needed an update to be able to check multiple types of input (besides the existing text input field) Existing uses of the controller will work as intended. With this PR, if you mark a checkbox input element as a `checkbox` target within a modal connected to this controller, its `checked` property will be considered when allowing the button target to be enabled/disabled.

A lil video of this in action:

https://github.com/pypi/warehouse/assets/639901/cd5221ff-f1bf-48f6-bd6f-b2389fb4cb1f

### DIFF
--- a/tests/frontend/confirm_controller_test.js
+++ b/tests/frontend/confirm_controller_test.js
@@ -42,41 +42,129 @@ describe("Confirm controller", () => {
     application.register("confirm", ConfirmController);
   });
 
-  describe("initial state", function() {
-    describe("the button", function() {
-      it("is disabled", function() {
+  describe("initial state", function () {
+    describe("the button", function () {
+      it("is disabled", function () {
         const buttonTarget = document.getElementById("button-target");
         expect(buttonTarget).toHaveAttribute("disabled");
       });
     });
   });
 
-  describe("functionality", function() {
-    describe("entering expected text", function() {
-      it("enables the button", function() {
-        fireEvent.input(document.getElementById("input-target"), { target: { value: "package" } });
+  describe("functionality", function () {
+    describe("entering expected text", function () {
+      it("enables the button", function () {
+        fireEvent.input(document.getElementById("input-target"), {
+          target: { value: "package" },
+        });
 
         const buttonTarget = document.getElementById("button-target");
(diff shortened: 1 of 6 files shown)


warehouse/static/js/warehouse/controllers/confirm_controller.js
```diff
@@ -12,18 +12,36 @@
...
+  }
+
+  checkInputString() {
+    // Check if the current value of the input text field matches what we expect it to. If so, this
+    // will return true
+    return (
+      this.inputTarget.value.toLowerCase() ===
+      this.buttonTarget.dataset.expected.toLowerCase()
+    );
+  }
```
miketheman, line 41: **praise:** ‏really nice clear refactor. Makes reading `check()` that much clearer.

miketheman (approved):
A question that doesn't block merging inline.

tests/frontend/confirm_controller_test.js
```diff
@@ -42,41 +42,129 @@ describe("Confirm controller", () => {
     application.register("confirm", ConfirmController);
   });
 
-  describe("initial state", function() {
-    describe("the button", function() {
-      it("is disabled", function() {
+  describe("initial state", function () {
+    describe("the button", function () {
+      it("is disabled", function () {
```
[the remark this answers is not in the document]
mm: yeah awesome point, it came after I ran `prettier` on it locally (muscle memory in VS Code) 😞 It may make sense to have a style plugin for stuff like this if folks want it (maybe a local Prettier config), but the initial formats to these files will cause a lot of diff noise like this

warehouse/static/js/warehouse/controllers/confirm_controller.js
mm: thanks so much! ❤️

PR #13953, opened 2023-06-15
fix: remove explicit commit from task

miketheman:
We have implicit commits, and calling this directly fails with:

    AssertionError: Transaction must be committed using the transaction manager

Resolves WAREHOUSE-PRODUCTION-1DS

### DIFF
--- a/warehouse/config.py
+++ b/warehouse/config.py
@@ -259,7 +259,12 @@ def configure(settings=None):
     )
 
     # Pythondotorg integration settings
-    maybe_set(settings, "pythondotorg.host", "PYTHONDOTORG_HOST", default="python.org")
+    maybe_set(
+        settings,
+        "pythondotorg.host",
+        "PYTHONDOTORG_HOST",
+        default="https://www.python.org",
+    )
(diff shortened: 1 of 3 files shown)


warehouse/config.py
```diff
@@ -259,7 +259,12 @@ def configure(settings=None):
     )
 
     # Pythondotorg integration settings
-    maybe_set(settings, "pythondotorg.host", "PYTHONDOTORG_HOST", default="python.org")
+    maybe_set(
+        settings,
+        "pythondotorg.host",
+        "PYTHONDOTORG_HOST",
+        default="https://www.python.org",
```
miketheman, line 266: Updated this value to match production, as discovered during reproduction

ewdurbin (approved):

ewdurbin:
Did you check to ensure this succeeds inside a task execution? I think there was a split-behavior where it worked in one and not the other.. but its been a while since I squinted at it.

miketheman:
> Did you check to ensure this succeeds inside a task execution? I think there was a split-behavior where it worked in one and not the other.. but its been a while since I squinted at it.

Yes, my test was to run the worker, and confirm that I got updated DB records after it successfully completed.

PR #13954, opened 2023-06-15
fix: add an email length validator

miketheman:
Currently if an email is too long it'll fall through to the database and
raise a `DataError` since the column max is 254.

By validating in the form, we prevent that from leaking through.

fixes WAREHOUSE-PRODUCTION-1EG


warehouse/accounts/forms.py
```diff
@@ -232,6 +232,9 @@ class NewEmailMixin:
             wtforms.validators.Regexp(
                 r".+@.+\..+", message=_("The email address isn't valid. Try again.")
             ),
+            wtforms.validators.Length(
+                max=254, message=_("The email address is too long. Try again.")
```
miketheman, line 236: Note that [the RFC allows for maximum length is 320](https://www.rfc-editor.org/rfc/rfc3696#page-5:~:text=In%20addition%20to%20restrictions,they%20are%20rarely%0A%20%20%20encountered.), but our DB is set to 254 [since the beginning](https://github.com/pypi/warehouse/blame/82965a56fc67d0c75fce32a2320d2fdb5899f51c/warehouse/accounts/tables.py#L60) [of time](https://github.com/pypi/legacy/blob/a418131180a14768328713df5533c600c425760d/pkgbase_schema.sql#L28).

ewdurbin (approved):

Issue #13962, opened 2023-06-17
PyPI is now requiring wheels to be too much compressed


moriyoshi:
**Describe the bug**

The change ([#13877](https://github.com/pypi/warehouse/pull/13877)) done in last week made my valid wheels that contain some native shared objects totally unuploadable. IMO requiring compression ratio < 0.1 is pretty rigorous.

**Expected behavior**

The threshold is too heuristic. Mitigation for zipbombs should be done in another way.

**To Reproduce**

Build any wheel of the following project and upload it to pypi:

https://github.com/opencollector/jntajis-python

**My Platform**

I am using cibuildwheel to build wheels.

**Additional context**

N/A

miohtama:
I encountered this issue with [web3-ethereum-defi](https://github.com/tradingstrategy-ai/web3-ethereum-defi) and can no longer not upload new versions of the package produced with Poetry.

[Poetry discussion here](https://github.com/orgs/python-poetry/discussions/8115).

di:
Should be fixed by #13967.

PR #13967, opened 2023-06-20
Fix ZIP bomb threshold too low

miohtama:
The current 10x ratio check prevents uploading legit wheels containing files like JSON, interface description and binaries.

See https://github.com/pypi/warehouse/issues/13962

Bumping from 10x compression threshold to 100x as I assume "100x should be enough for everyone". Note that this should be a temporary solution, as the ZIP bomb check could be more robust.


warehouse/forklift/legacy.py
```diff
@@ -83,7 +83,13 @@
...
 
 COMPRESSION_RATIO_MIN_SIZE = 64 * ONE_MB
-COMPRESSION_RATIO_THRESHOLD = 10
+
+# If the zip file decompressed to 100x more space 
+# than it is uncompressed, consider it a ZIP bomb.
+# Note that packages containing interface descriptions, JSON,
+# such resources can compress really well.
+# See discussion here: https://github.com/pypi/warehouse/issues/13962
+COMPRESSION_RATIO_THRESHOLD = 100
```
ewdurbin, line 92:
```suggestion
COMPRESSION_RATIO_THRESHOLD = 50
```

ewdurbin (approved):

PR #13993, opened 2023-06-22
docs: correct environment variable name

LawnGnome:
The getting started guide consistently refers to `WEB_HOST`, but the environment variable used in `docker-compose.yml` is `WEB_PORT`.


miketheman (approved):
Thanks so much!

PR #13994, opened 2023-06-22
Bump celery, kombu, urllib3

di:
Now that https://github.com/sibson/redbeat/issues/249 is resolved and the upgrade added in https://github.com/pypi/warehouse/pull/13927, we can un-revert #13861.

Closes https://github.com/pypi/warehouse/pull/13947.

PR #13995, opened 2023-06-22
Revert "Bump celery, kombu, urllib3 (#13994)"

di:
Just in case. This reverts commit 299f7b1a4589877996f3d62f6ba750d0ef060afd.


miketheman (approved):
Approving just in case you need it.

PR #14001, opened 2023-06-22
feat(ui): align sponsor logos to top

miketheman:
Visually looks better.

Using a mix of flex and inline-grid we can align items a little more cleanly.


miketheman:
Approximation of new look (had to hack directives into browser to view prod imagery:

image

di (approved):

Issue #7811, opened 2020-04-18
No validation of sdist filenames allows uploading files pip won't choose


jamadden:
**Describe the bug**
In response to https://github.com/pypa/warehouse/issues/3962#issuecomment-614515258, a user uploaded an sdist-that-wasn't with the file name `data-importer-3.1.1.linux-x86_64.tar.gz`.  I thought at first it might truly contain non-manylinux compiled binaries so I wanted to check its contents; turns out it didn't, but pip won't install it anyway unless you force it to.

```console
$ wget https://files.pythonhosted.org/packages/65/8d/.../data-importer-3.1.1.linux-x86_64.tar.gz
…
$ tar -tf ./data-importer-3.1.1.linux-x86_64.tar.gz
./
./mnt/
./mnt/d/
./mnt/d/Sandbox/
./mnt/d/Sandbox/huru-server/
./mnt/d/Sandbox/huru-server/.venv/
./mnt/d/Sandbox/huru-server/.venv/lib/
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer/
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer/admin.py
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer/admin.pyc
...
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer/__init__.pyc
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer-3.1.1-py2.7.egg-info/
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer-3.1.1-py2.7.egg-info/dependency_links.txt
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer-3.1.1-py2.7.egg-info/not-zip-safe
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer-3.1.1-py2.7.egg-info/PKG-INFO
./mnt/d/Sandbox/huru-server/.venv/lib/python2.7/site-packages/data_importer-3.1.1-py2.7.egg-info/requires.txt
[...]

Issue #13991, opened 2023-06-22
Replace deprecated usage of `pkg_resources`


miketheman:
`setuptools` has deprecated their `pkg_resources` sub-package in favor of stdlib packages. See https://setuptools.pypa.io/en/latest/pkg_resources.html

We appear to use it on only a few places:

```shellsession
$ git grep pkg_resources
tests/unit/forklift/test_legacy.py:import pkg_resources
tests/unit/forklift/test_legacy.py:                project.name, pkg_resources.safe_name(project.name).lower()
warehouse/forklift/legacy.py:import pkg_resources
warehouse/forklift/legacy.py:    prefix = pkg_resources.safe_name(project.name).lower()
warehouse/forklift/legacy.py:    if not pkg_resources.safe_name(filename).lower().startswith(prefix):
warehouse/utils/readme.py:import pkg_resources
warehouse/utils/readme.py:    return pkg_resources.get_distribution("readme-renderer").version
```

I'm not certain where to best find a replacement for [`safe_name()`](https://setuptools.pypa.io/en/latest/pkg_resources.html#parsing-utilities:~:text=on%20other%20lines.-,safe_name(name),-Return%20a%20%E2%80%9Csafe) yet. We might have existing comparable functionality in warehouse already.
**Feel free to omit this from the refactor if it's too complex**.

A similar refactor took place in `readme-renderer`, see https://github.com/pypa/readme_renderer/pull/281 for details.

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypi/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

If you are working on this issue and have questions, feel free to ask them here, in the [`#pypa-dev` chat channel on Libera](https://web.libera.chat/#pypa-dev), the [PyPA Discord](https://discord.gg/pypa) or on the [Discourse](https://discuss.python.org/c/packaging/14).

**Screenshot Required**: *If your pull request makes a visual change*, include a screenshot of your update. This helps our team give you feedback faster.

di:
[...]

PR #14027, opened 2023-06-26
Make filename check more strict

di:
This ensures that the filenames of uploaded files only have the actual project name where the project name should be. Previously, this allowed filenames that started with the project name, so a file `no-way-{version}.tar.gz` could be uploaded to the `no` project, effectively making the filename unusable for the `no-way` project due to our policy on filename reuse.

Towards https://github.com/pypi/warehouse/issues/7811, but doesn't totally fix it.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2311,15 +2311,21 @@ class TestFileUpload:
         )
 
     @pytest.mark.parametrize(
-        "filename_prefix, project_name",
+        "filename, project_name",
         [
-            ("nope", "something_else"),  # completely different
-            ("nope", "no"),  # starts with same prefix
-            ("no-way", "no"),  # starts with same prefix with hyphen
+            # completely different
+            ("nope-{version}.tar.gz", "something_else"),
+            ("nope-{version}-py3-none-any.whl", "something_else"),
+            # starts with same prefix
+            ("nope-{version}.tar.gz", "no"),
+            ("nope-{version}-py3-none-any.whl", "no"),
+            # starts with same prefix with hyphen
+            ("no-way-{version}.tar.gz", "no"),
+            ("no_way-{version}-py3-none-any.whl", "no"),
         ],
     )
     def test_upload_fails_with_wrong_filename(
-        self, pyramid_config, db_request, metrics, filename_prefix, project_name
+        self, pyramid_config, db_request, metrics, filename, project_name
     ):
         user = UserFactory.create()
         pyramid_config.testing_securitypolicy(identity=user)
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -1208,10 +1208,20 @@ def file_upload(request):
...
     _validate_filename(filename)
 
+    # Extract the project name from the filename and normalize it.
+    filename_prefix = pkg_resources.safe_name(
+        # For wheels, the project name is normalized and won't contain hyphens, so
+        # we can split on the first hyphen.
+        filename.partition("-")[0]
+        if filename.endswith(".whl")
+        # For source releases, we know that the version should not contain any
+        # hypens, so we can split on the last hypen to get the project name.
```
miketheman, line 1218:
```suggestion
        # hyphens, so we can split on the last hyphen to get the project name.
```

warehouse/forklift/legacy.py
```diff
@@ -1208,10 +1208,20 @@ def file_upload(request):
     # Ensure the filename doesn't contain any characters that are too 🌶️spicy🥵
     _validate_filename(filename)
 
+    # Extract the project name from the filename and normalize it.
+    filename_prefix = pkg_resources.safe_name(
```
miketheman, line 1212: **question:** ‏this is using the deprecated `pkg_resources` package - should we take the time now to replace, or do you want for forge ahead and let this get bundled into #13991 (or another)?

tests/unit/forklift/test_legacy.py
```diff
@@ -2310,31 +2310,53 @@ def test_upload_fails_with_diff_filename_same_blake2(
...
+        }.get(svc)
 
         db_request.POST = MultiDict(
             {
                 "metadata_version": "1.2",
                 "name": project.name,
                 "version": release.version,
                 "filetype": "sdist",
-                "md5_digest": "nope!",
+                "md5_digest": _TAR_GZ_PKG_MD5,
```
miketheman, line 2351: **question:** ‏now that we're paramtrizing .whl files in the test case, does the same TAR_GZ behavior name apply to those files as well? Or is this 100% a testing-only behavior and we're unlikely to hit a whl-specific condition on prod? I suspect it's fine, but wanted to ask.

miketheman (approved):
Typos suggested for fixin' and a couple of non-blocking questions.

warehouse/forklift/legacy.py
di: Yeah, I'd like it to get bunded into the fix for #7811, this is sort of orthogonal.

tests/unit/forklift/test_legacy.py
di: The file contents don't really matter here as long as the hashes match, but I agree this could be a little confusing to future travelers.

PR #14044, opened 2023-06-30
Add metrics for projects using OIDC publishers

sethmlarson:
Track the uptake of OIDC publishers on PyPI.

### DIFF
--- a/tests/unit/oidc/test_tasks.py
+++ b/tests/unit/oidc/test_tasks.py
@@ -1,3 +1,15 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import pretend
 
 from warehouse.oidc.tasks import compute_oidc_metrics
@@ -6,7 +18,7 @@ from ...common.db.oidc import GitHubPublisherFactory
 from ...common.db.packaging import ProjectFactory
 
 
-def test_compute_oidc_metrics(db_request, monkeypatch):
+def test_compute_oidc_metrics(db_request, metrics):
     # Projects with OIDC
     critical_project_oidc = ProjectFactory.create(
         name="critical_project_oidc", pypi_mandates_2fa=True
@@ -31,12 +43,9 @@ def test_compute_oidc_metrics(db_request, monkeypatch):
     )
     oidc_publisher_2 = GitHubPublisherFactory.create(projects=[critical_project_oidc])
 
-    gauge = pretend.call_recorder(lambda metric, value: None)
-    db_request.find_service = lambda *a, **kw: pretend.stub(gauge=gauge)
-
     compute_oidc_metrics(db_request)
 
-    assert gauge.calls == [
+    assert metrics.gauge.calls == [
         pretend.call("warehouse.oidc.total_projects_using_oidc_publishers", 2),
         pretend.call("warehouse.oidc.total_critical_projects_using_oidc_publishers", 1),
     ]
--- a/tests/unit/oidc/test_tasks.py
+++ b/tests/unit/oidc/test_tasks.py
@@ -10,12 +10,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
+
 import pretend
 
 from warehouse.oidc.tasks import compute_oidc_metrics
 
 from ...common.db.oidc import GitHubPublisherFactory
-from ...common.db.packaging import ProjectFactory
+from ...common.db.packaging import (
+    FileEventFactory,
+    FileFactory,
+    ProjectFactory,
+    ReleaseFactory,
+)
 
 
 def test_compute_oidc_metrics(db_request, metrics):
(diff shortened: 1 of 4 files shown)


tests/unit/oidc/test_tasks.py
```diff
@@ -0,0 +1,42 @@
+import pretend
+
+from warehouse.oidc.tasks import compute_oidc_metrics
+
+from ...common.db.oidc import GitHubPublisherFactory
+from ...common.db.packaging import ProjectFactory
+
+
+def test_compute_oidc_metrics(db_request, monkeypatch):
```
miketheman, line 9: **question:** ‏Looks like `monkeypatch` fixture isn't used in the test, should it be removed?
```suggestion
def test_compute_oidc_metrics(db_request):
```

tests/unit/oidc/test_tasks.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Create OIDC publishers. One OIDCPublisher can be shared by multiple
+    # projects so 'oidc_publisher_1' represents that situation. Verify
+    # that metrics don't double-count projects using multiple OIDC publishers.
+    oidc_publisher_1 = GitHubPublisherFactory.create(
+        projects=[critical_project_oidc, non_critical_project_oidc]
+    )
+    oidc_publisher_2 = GitHubPublisherFactory.create(projects=[critical_project_oidc])
+
+    gauge = pretend.call_recorder(lambda metric, value: None)
```
miketheman, line 34: **suggestion:** ‏we have a general-purpose `metrics` fixture which you can use instead of creating one. 
https://github.com/pypi/warehouse/blob/2241171cf130b853dd534653c0d932dbf5a0db6e/tests/conftest.py#L71C1-L82

tests/unit/oidc/test_tasks.py
```diff
@@ -0,0 +1,42 @@
...
+    db_request.find_service = lambda *a, **kw: pretend.stub(gauge=gauge)
+
+    compute_oidc_metrics(db_request)
+
+    assert gauge.calls == [
```
miketheman, line 39: Using using the fixture, this ought to become:
```suggestion
    assert metrics.gauge.calls == [
```

warehouse/oidc/tasks.py
```diff
@@ -0,0 +1,24 @@
...
+from warehouse.metrics import IMetricsService
+from warehouse.packaging.models import Project
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def compute_oidc_metrics(request):
+    metrics = request.find_service(IMetricsService, context=None)
+
+    projects_using_oidc = (
+        request.db.query(Project).distinct().join(Project.oidc_publishers)
```
miketheman, line 11: **question:** ‏The query as is is likely to perform sequential scans on the `projects` table which is now in the 400k+ range - we may need to add an index. Have you looked into that already?

miketheman (commented):
Looks generally okay, a couple of suggestions and questions inline.

warehouse/oidc/tasks.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+
+    # Metric for count of all projects using OIDC
+    metrics.gauge(
+        "warehouse.oidc.total_projects_using_oidc_publishers",
+        projects_using_oidc.count(),
```
miketheman, line 17: Also noticed the usage of `.count()` - which is slower than postgres' `count` function (I know, it's confusing...)

Since we don't intend to use the returned objects in any way, we can use `func.count()` to add the `COUNT(*)` to the SQL query. See https://docs.sqlalchemy.org/en/20/core/functions.html#sqlalchemy.sql.functions.count

sethmlarson:
Thanks for the review @miketheman! I took my inspiration from the [2FA metrics task](https://github.com/pypi/warehouse/commit/0c2ff6d74aa97e31a3d40365fe84f1e491f82dea), I'll get those comments resolved that you mentioned.

warehouse/oidc/tasks.py
miketheman: This is not a hard requirement - we use count() elsewhere already with no material negative impact observed thus far. I saw something, so I said something. 😉

warehouse/oidc/tasks.py
di: Also, technically, this is just counting projects that have a publisher configured, but that doesn't mean they're actually _using_ it.

We might want to count `FileEvents` from OIDC publishers instead:

```sql
select count(*) from file_events where additional->>'publisher_url' is not null;
```

and if we want to count # of projects, we can do the necessary joins from that.

sethmlarson: Per @di's suggestion I've distinguished between projects that have configured OIDC and have successfully published a release with OIDC. Hopefully the join-chain is what you were expecting?

@miketheman Do we need to change either of these queries to optimize them further? My SQLAlchemy-fu is a bit lacking so any suggestions would be helpful :pray:

warehouse/oidc/tasks.py
```diff
@@ -0,0 +1,59 @@
...
+from warehouse.metrics import IMetricsService
+from warehouse.packaging.models import File, Project, Release
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def compute_oidc_metrics(request):
+    metrics = request.find_service(IMetricsService, context=None)
+
+    projects_configured_oidc = (
+        request.db.query(Project.id).distinct().join(Project.oidc_publishers)
```
miketheman, line 23: **thought:** ‏I'm having some difficulty in getting this query to run in a shell:

```shellsession
$ make shell
>>> from warehouse.packaging.models import File, Project, Release
>>> db.query(Project.id).distinct().join(Project.oidc_publishers)
  File "/usr/local/lib/python3.11/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
AttributeError: type object 'Project' has no attribute 'oidc_publishers'
```

I'm not entirely certain why this relationship isn't being expressed. 🤔 

Were you able to run this query?

miketheman: Never mind, had to call `sqlalchemy.orm.configure_mappers()` after import to create the mappings, something that happens automatically in production.

warehouse/oidc/tasks.py
miketheman: > Do we need to change either of these queries to optimize them further?

I'm gonna say "not yet" for now - let's let the task run and see how it does.

miketheman (approved):

miketheman:
@sethmlarson Any objections about shipping this Monday instead of today, just in case the queries tax the db too heavily?

sethmlarson:
@miketheman No objections here! We should maybe even wait for Wednesday since Monday and Tuesday are a holiday for many folks in the US so I and Ee won't be working that day.

warehouse/oidc/tasks.py
sethmlarson: Glad you figured that out, I wouldn't have been much help there :P All I had was "unit tests seem to work!"

sethmlarson:
@miketheman Looks like CI can't proceed without maintainer approval? Also I'm cool to deploy this now, I can create a revert PR in case we need to quickly revert the commit.

PR #14046, opened 2023-06-30
Separate pending org applications and add banner

di:


di:
Looks like this:

image

warehouse/templates/manage/organizations.html
```diff
@@ -132,6 +132,19 @@ <h3 class="organization-snippet__title">
         <p>{% trans %}You have not joined any organizations on PyPI, yet.{% endtrans %}</p>
       </div>
     {% endfor %}
+  </div>
+
+  {% if organization_applications|selectattr('is_approved', 'equalto', None)|list|length > 0 %}
```
miketheman, line 137: **question:** ‏is this attempting to to perform the equivalent of `if orgapps.is_approved is None`? 
If so, the jinja of `{% if organization.is_approved is none ... %}` ought to work.

I hit this before, and this helped: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-tests.none:~:text=The%20special%20constants,the%20lowercase%20versions

miketheman (approved):
Generally fine if it works, a suggestion for improvement vs selectattr if that's what's intended.

warehouse/templates/manage/organizations.html
di: It's checking if there are _any_ organizations for which `is_approved` is `None`, not sure your example would work there?

I'll change the casing to `none` for consistency though:

```suggestion
  {% if organization_applications|selectattr('is_approved', 'equalto', none)|list|length > 0 %}
```

PR #14055, opened 2023-07-03
chore(deps): update all js deps with `npm update`

miketheman:
Two stylelint plugins are left out, as they exceeded the semver range allowed.


miketheman:
Closing for now. Attempting to excise babel from the stack first.

di (approved):

PR #14062, opened 2023-07-05
Revert "Add metrics for projects using OIDC publishers (#14044)"

sethmlarson:
In case of an issue with the new queries we can revert #14044.
This reverts commit cabce6426362d52879738fb55b60cac1e45b9ba3.


di:
This seems to have gone out just fine, closing.

PR #13980, opened 2023-06-21
OIDC provider ActiveState

th3coop:
1 or 2 PRs to add ActiveState as an OIDC provider.  This PR adds the backend features:
- DB migration
- associated modeling
- tests
- `disallow-activestate-oidc` admin flag is `True`

As the OIDC code seems to be evolving, I explicitly avoided making type changes outside the new code as I was getting a lot of merge conflicts otherwise. 

ActiveState Internal work tracking ID: DS-1723

PR #14063, opened 2023-07-05
Refactor OIDC mint token endpoint

th3coop:
Make it easier to mint tokens with additional publishers

This PR splits warehouse.oidc.view.mint_token_from_oidc into two functions.  1, `mint_token`, is OIDC provider specific.  The other, `mint_token_from_oidc_*` that is OIDC provider agnostic.  This should make it easier to add more OIDC provider token handlers.

ActiveState Internal work tracking ID: DS-1722

### DIFF
--- a/tests/unit/oidc/test_views.py
+++ b/tests/unit/oidc/test_views.py
@@ -18,15 +18,13 @@ import pretend
 import pytest
 
 from tests.common.db.accounts import UserFactory
-from tests.common.db.oidc import PendingGitHubPublisherFactory
+from tests.common.db.oidc import GitHubPublisherFactory, PendingGitHubPublisherFactory
 from tests.common.db.packaging import ProjectFactory
 from warehouse.events.tags import EventTag
 from warehouse.macaroons import caveats
 from warehouse.macaroons.interfaces import IMacaroonService
 from warehouse.oidc import errors, views
 from warehouse.oidc.interfaces import IOIDCPublisherService
-from warehouse.oidc.models import github
-from warehouse.packaging.models import Project
 from warehouse.rate_limiting.interfaces import IRateLimiter
 
 
@@ -111,7 +109,7 @@ def test_mint_token_from_github_oidc_not_enabled():
         {"token": {}},
     ],
 )
-def test_mint_token_from_github_oidc_invalid_payload(body):
+def test_mint_token_oidc_invalid_payload(body):
     class Request:
         def __init__(self):
             self.response = pretend.stub(status=None)
@@ -202,7 +200,9 @@ def test_mint_token_from_trusted_publisher_lookup_fails():
 
 def test_mint_token_from_oidc_pending_publisher_project_already_exists(db_request):
     project = ProjectFactory.create()
-    pending_publisher = PendingGitHubPublisherFactory.create(project_name=project.name)
+    pending_publisher = PendingGitHubPublisherFactory.create(
+        project_name=project.name,
+    )
 
     db_request.flags.disallow_oidc = lambda f=None: False
     db_request.body = json.dumps({"token": "faketoken"})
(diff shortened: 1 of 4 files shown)


tests/unit/oidc/test_views.py
```diff
@@ -383,18 +385,21 @@ def test_mint_token_from_oidc_no_pending_publisher_ok(
...
+    monkeypatch.setattr(
+        project, "record_event", pretend.call_recorder(lambda **kw: None)
     )
-    publisher = pretend.stub(
-        id="fakepublisherid",
-        projects=[project],
-        publisher_name="fakepublishername",
-        publisher_url=lambda x=None: "https://fake/url",
+
+    publisher = github.GitHubPublisher(
```
th3coop, line 393: I had to do this change because I had to add [this change](https://github.com/pypi/warehouse/pull/14063/files#diff-00352684c58d85c29397c3a8b7a868b5a7846e75a3dfb6f67af1ab141483ff0bR204) to stop linting errors from occurring [here](https://github.com/pypi/warehouse/pull/14063/files#diff-00352684c58d85c29397c3a8b7a868b5a7846e75a3dfb6f67af1ab141483ff0bR236) and anywhere else a prop was accessed on the OIDCPublisher object.

warehouse/oidc/models/_core.py
```diff
@@ -211,7 +212,9 @@ def publisher_name(self) -> str:  # pragma: no cover
         # Only concrete subclasses are constructed.
         raise NotImplementedError
 
-    def publisher_url(self, claims=None) -> str | None:  # pragma: no cover
+    def publisher_url(
+        self, claims: SignedClaims | None = None
+    ) -> str:  # pragma: no cover
```
th3coop, line 217: I've been meaning to make this change for sometime so adding it to this PR so it's upstream sooner.  This should never be `None` right?  Or shouldn't, so types and tests should reflect that expectation?  If that's not true, I'm happy to revert.

th3coop:
I wanted to address the comment [here](https://github.com/pypi/warehouse/pull/14063/files#diff-00352684c58d85c29397c3a8b7a868b5a7846e75a3dfb6f67af1ab141483ff0bR119) in this PR but I ended up breaking things up into the known route and the publisher we know that route handles.  So, no looking at the request to see which publisher to load.  Does this make sense?

th3coop:
Ok, this is ready for review.  @di, is it appropriate to ping you in that case?  I've had a triple check look at the [submitting patches](https://warehouse.pypa.io/development/submitting-patches.html) docs but want to make sure I'm not bypassing a step.  The actions on this repo keep you pretty locked in to that pages processes which is great.

th3coop:
I noticed that quite a few of the tests in *warehouse/tests/unit/oidc/test_views.py* are implicitly testing the `GitHubPublisher` and `PendingGitHubPublisher` classes functionality when it doesn't need to as that functionality is tested in `oidc/models/test_github.py`.

I went down a bit of a rabbithole to try to abstract things away a bit more in *warehouse/tests/unit/oidc/test_views.py* so that the new `oidc.views::mint_token()` can really be tested "publisher agnostically".

To achieve that I had to add two new factories to warehouse/tests/common/db/oidc.py, `PendingOIDCPublisherFactory` and `OIDCPublisherFactory`.  It's all working quite nicely except for one hitch.

`OIDCPublisher.publisher_name` is a `@property` so I can't "stub" it's value.  If I remove the `@property` decorator though, then I can stub it as a function.

Before I put in time to wrap up these changes, I wanted to run the idea of `OIDCPublisher.publisher_name` being a function, not a property of the class.

If I were able to make this change and introduce these new Factories, the only function in *oidc/test_views.py* that needs to know about a specific publisher is the `test_mint_token_from_github_oidc_not_enabled` method in this PR, which doesn't actually need to construct a Publisher.

If it made it easier to review and give feedback/answer this question, I could put up a PR that contains only the change to `property_name`, the new factories, and 1 example for each factory where they would be used in the existing tests.

warehouse/oidc/models/_core.py
di: The original intention here is that not all publishers would necessarily have a URL to link to, so this could be `None`, although that's not currently the case.

th3coop: Ahh ok.  Then my change doesn't make sense.  Will revert.

th3coop:
@di, if this PR overlaps with work you already have on the go, I don't mind dropping this PR.  I should have asked before going this route in the first place.  Let me know what your preference is.

di:
@th3coop I think this pairs nicely actually, let me get my branch up and we can figure out next steps.

th3coop:
Works for me, @di.

th3coop:
https://github.com/pypi/warehouse/pull/14063#issuecomment-1681278695
Any luck with that work @di?  I've got the majority of implemenation work done for adding ActiveState but waiting on my PR here to wrap up to go through finishing/fixing tests and manually testing of my other PRs.

th3coop:
Hi @di, We're hoping to move this stuff along over the next few weeks.  Any chance the branch you mentioned above is ready?

tests/unit/oidc/test_views.py
```diff
@@ -392,18 +456,16 @@ def test_mint_token_from_oidc_no_pending_publisher_ok(
         id="fakeprojectid",
         record_event=pretend.call_recorder(lambda **kw: None),
     )
-    publisher = pretend.stub(
```
th3coop, line 395: This was changed to use the factory due to the type check [here](https://github.com/pypi/warehouse/pull/14063/files#diff-00352684c58d85c29397c3a8b7a868b5a7846e75a3dfb6f67af1ab141483ff0bR146) and [here](https://github.com/pypi/warehouse/pull/14063/files#diff-00352684c58d85c29397c3a8b7a868b5a7846e75a3dfb6f67af1ab141483ff0bR220) which was added because of mypy errors  in the code after that type check.

tests/unit/oidc/test_views.py
```diff
@@ -374,6 +371,73 @@ def test_mint_token_from_pending_trusted_publisher_invalidates_others(
     ]
 
 
+def test_mint_token_from_oidc_only_pending_publisher_fail(monkeypatch, db_request):
```
th3coop, line 374: This test tests an impossible state to reach as laid out in [this comment](https://github.com/pypi/warehouse/pull/14063/files#r1406874749) and I'm not sure what to do about it.

th3coop:
Closing loops on [this commend](https://github.com/pypi/warehouse/pull/14063#issuecomment-1633304299): I ended up following the existing pattern of stubbing away the code that shouldn't be tested, to focus on the code being test. So this new classes was removed from my changes.

th3coop:
@di I saw your Google PR seems to have been merged which I believe is what I was waiting on for this PR so I've gone through this one again to get it ready for review.  It's now ready for review.

warehouse/oidc/views.py
```diff
@@ -71,96 +96,106 @@ def oidc_audience(request):
     require_csrf=False,
     has_translations=True,
 )
-def mint_token_from_oidc(request):
-    def _invalid(errors):
-        request.response.status = 422
-        return {"message": "Token request failed", "errors": errors}
-
+def mint_token_from_oidc_github(request: Request):
```
di, line 99: I think instead we should probably make a generic non-GitHub-specific endpoint that accepts any OIDC token in the body and attempts to mint a token for it based on properties of the token (specifically, the `iss` claim should allow us to pick the right publisher type)

I don't see a great reason to maintain a separate endpoint for every OIDC provider we want to support here, that was probably not necessary in the GitHub implementation.

(cc @woodruffw for any thoughts as well)

warehouse/oidc/views.py
```diff
@@ -178,14 +213,30 @@ def _invalid(errors):
...
+        # point.
+        return _invalid(
+            errors=[
+                {
+                    "code": "invalid-publisher",
+                    "description": "valid token, but no corresponding publisher",
+                }
+            ],
+            request=request,
+        )
```
di, line 232: This seems weird, does typing the `find_publisher` call fix this? Can we `assert isinstance(publisher, OIDCPublisher)` in the `try`/`except` instead?

tests/unit/oidc/test_views.py
```diff
@@ -392,18 +456,16 @@ def test_mint_token_from_oidc_no_pending_publisher_ok(
...
+    publisher = GitHubPublisherFactory()
+    monkeypatch.setattr(publisher.__class__, "projects", [project])
+    publisher.publisher_url = pretend.call_recorder(lambda **kw: "https://fake/url")
     # NOTE: Can't set __str__ using pretend.stub()
     monkeypatch.setattr(publisher.__class__, "__str__", lambda s: "fakespecifier")
 
     def _find_publisher(claims, pending=False):
         if pending:
-            raise errors.InvalidPublisherError
+            return None
```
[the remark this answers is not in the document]
di: I agree, this seems wrong, I think the line you're having trouble with is probably unnecessary, see the comment there.

di (commented):
Hi @th3coop, thanks for your extreme patience here. I've given this a review and I think it's mostly good to go aside from some small comments. Thank you!

warehouse/oidc/views.py
woodruffw: Yes, agreed -- the main reason I did that originally was to avoid snooping the unverified JWT claims before verifying them, but that ship has sailed elsewhere in the implementation IIRC (and is not a security risk, so long as we perform proper key segregation and add both positive and negative tests).

We'll probably have to keep the GitHub-specific endpoint around for a while because of existing users (including some people who have reimplemented the token exchange flow), but it can be a thin shim over the "generic" endpoint :slightly_smiling_face:

woodruffw:
@th3coop JFYI: I'm going to make a new PR off of your branch here, to address the comments/suggestions above. I'll preserve your history though, so it'll be your authorship in the merge 🙂

warehouse/oidc/views.py
th3coop: Yeah, this was icky.  Good idea to use `assert`.

th3coop:
@di thanks for the review and no sweat.  I appreciate any of your time.  @woodruffw much appreciated!  I was just about to get to this but thanks for taking it.  Do you need anything from me in #15148?  I really should have thought of using a tokens `iss` field to establish what publisher it was 🤦

woodruffw:
No problem! And nope, nothing needed at the moment -- I'll give you a ping there if I have any questions. Thanks a ton for your hard work here!

di:
Thanks @th3coop, I wasn't sure what your availability was and you've already put in a lot of work here so I asked @woodruffw to work on keeping this moving. We'd definitely appreciate you getting https://github.com/pypi/warehouse/pull/13980 ready to merge once this/#15148 are merged if you're able!

woodruffw:
#15148 is undrafted, so closing here!

th3coop:
No problem, I'll resolve the merge conflict in #13980.  Clarification: #13980 is 1 of 3 PRs that add the full OIDC treatment for ActiveState.
#13980 adds the DB modeling and most of the validation (as you've seen).

The other two are PRs in the ActiveState fork:
1. https://github.com/ActiveState/warehouse/pull/3 adds the form in the account settings to configure an ActiveState Publisher.
2. https://github.com/ActiveState/warehouse/pull/1 adds the handler endpoint

I'll create subsequent PRs against pypi/warehouse:main as the respective "parent" PR gets merged.

Hopefully this makes sense for chunking up the changes.  I didn't want to throw a PR at y'all with that includes ALL of the changes.

PR #14082, opened 2023-07-07
legacy: improve error msg for project mismatches

woodruffw:
WIP; I'll update this to add a troubleshooting section (and link to it) as well.


webknjaz (approved):
It'd be nice to come up with some linter-like error codes and permanent links so that the error message could also link the troubleshooting section..

woodruffw:
> linter-like error codes and permanent links so that the error message could also link the troubleshooting section..

We could probably use the HTML fragment here, but I figured it wasn't worth it in this case (especially since the longer URL will probably reflow on smaller terminals). Happy to change, though!

miketheman (approved):

PR #14086, opened 2023-07-07
feat(admin): add re-render mechanism for releases

miketheman:
Add an admin view for Release Detail to trigger a request for description re-rendering, instead of waiting for an updated readme-renderer version to trigger a site-wide update.


warehouse/admin/templates/admin/projects/release_detail.html
```diff
@@ -229,7 +229,15 @@ <h3>Description Rendering</h3>
...
-          <td><code>{{ release.description.content_type }}</code></td>
+          <td>
+            {% if release.description.content_type is none %}
+            <code>-NULL-</code>
+            {% elif release.description.content_type == "" %}
+            <code>-BLANK-</code>
+            {% else %}
+            <code>{{ release.description.content_type }}</code>
+            {% endif %}
+          </td>
```
miketheman, line 240: I realized it was tricky to determine if the content type was null or black string, so I added a carveout here for that.

I tried to use [jinja's `default()`](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.default), but got a bit lost and couldn't make it work. Happy to change if someone has a better idea!

di (approved):

PR #14113, opened 2023-07-12
chore: remove view transcript for 404 pages

Abdur-rahmaanJ:
This read transcript link on 404 pages http://www.montypython.net/scripts/cheese.php leads to a casino site.

Proposing to remove it as it might be seen as the pypi promotes gambling.

Selection_004

I haven't found a replacement yet.


di:
How odd. Agree we should remove this, perhaps we can replace the link with https://people.cs.rutgers.edu/~mdstone/class/440-fall-99/cheeseshop.html instead to maintain accessibility?

Abdur-rahmaanJ:
Cool, let me do this.

Abdur-rahmaanJ:
Done

miketheman (approved):

PR #14126, opened 2023-07-13
feat: disallow management actions without a verified primary email

miketheman:
As a way to continue on the path of making user accounts more secured,
require a user to have verified the primary email address prior to performing any
other management actions.

Preserves their ability to manage their own account.

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -134,6 +134,22 @@ def forbidden(exc, request):
 
     # Check if the error has a "result" attribute and if it is a WarehouseDenied
     if hasattr(exc, "result") and isinstance(exc.result, WarehouseDenied):
+        # If the forbidden error is because the user does not have a verified
+        # email address, redirect them to their account page for email verification.
+        if exc.result.reason == "unverified_email":
+            request.session.flash(
+                request._(
+                    "You must verify your **primary** email address before you "
+                    "can perform this action."
+                ),
+                queue="error",
+            )
+            url = request.route_url(
+                "manage.account",
+                _query={REDIRECT_FIELD_NAME: request.path_qs},
+            )
+            return HTTPSeeOther(url)
+
         # If the forbidden error is because the user doesn't have 2FA enabled, we'll
         # redirect them to the 2FA page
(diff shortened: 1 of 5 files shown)


warehouse/views.py
```diff
@@ -134,6 +134,22 @@ def forbidden(exc, request):
 
     # Check if the error has a "result" attribute and if it is a WarehouseDenied
     if hasattr(exc, "result") and isinstance(exc.result, WarehouseDenied):
+        # If the forbidden error is because the user does not have a verified
+        # email address, redirect them to their account page for email verification.
+        if exc.result.reason == "unverified_email":
+            request.session.flash(
+                request._(
+                    "You must verify your email address before you can perform "
+                    "this action."
```
di, line 143: I think we can be more specific here: the _primary_ email address needs to be verified. It's possible the user has a verified email, but it's not their primary email.

miketheman: Excellent point, will reword!

warehouse/views.py
```diff
@@ -134,6 +134,22 @@ def forbidden(exc, request):
 
     # Check if the error has a "result" attribute and if it is a WarehouseDenied
     if hasattr(exc, "result") and isinstance(exc.result, WarehouseDenied):
+        # If the forbidden error is because the user does not have a verified
+        # email address, redirect them to their account page for email verification.
+        if exc.result.reason == "unverified_email":
```
di, line 139: I realize you're just following the pattern already established here, but I'm wondering if these should be subclasses of `WarehouseDenied` that we check with `isinstance` instead of comparing the reason here.

di (approved):
LGTM with or without the change

warehouse/views.py
miketheman: It's a good thought - I tinkered around with that, but didn't understand exactly what I was doing, since these aren't exceptions, rather subclasses of `pyramid.security.Denied`. I'll leave this as is for now, especially since there's other work in play about restructuring the ACLs and separating authentication from authorization a little better.

PR #14127, opened 2023-07-13
OIDC forms folder/file structure allows inclusion of new Publishers

th3coop:
There should be no functional change in this PR.  This PR is restricted to moving the OIDC forms files into a folder structure that will make it easier or more clear where to add future OIDC publisher forms.

Based on this folder structure, I'll be adding ActiveState forms to *warehouse/oidc/forms/activestate.py*.

Note worth as it's deviating from what might be a exisiting naming scheme patter:  I couldn't name the test file *oidc/forms/test_github.py*.  Pytest complains that the file name isn't unique.


woodruffw:
> Note worth as it's deviating from what might be a exisiting naming scheme patter: I couldn't name the test file _oidc/forms/test_github.py_. Pytest complains that the file name isn't unique.

FWIW, this usually happens because there's an `__init__.py` missing in a directory somewhere. 

In this case, it looks like `tests/unit/oidc/models` is missing an `__init__.py`, so adding one there should allow you to resolve + rename here 🙂

th3coop:
@woodruffw thanks for that.  I'll make that change!

th3coop:
I'm not sure why that test is failing.  It passes locally and any code I look at that's associated with that test hasn't changed for quite a while, let alone from this PR.

th3coop:
Hi @di, is there anything I can do to make this PR easier to reivew?  Could I assist in untangling this from up coming work you having going on?

di (approved):
LGTM, can you bring this branch up to date?

th3coop:
Thank a lot @di.

PR #14142, opened 2023-07-17
Improve caching for local docker builds

ewdurbin:
- Caches apt metadata and downloads
- Caches downloaded/built wheels
- Caches npm

This will primarily be a speedup for local dev builds, as deployment builds using ephemeral buildkitd instances and relies on the layer cache.

Supersedes #10004 and #10021


Dockerfile
```diff
@@ -106,9 +117,18 @@ ARG DEVEL=no
 # i.e. 'docker compose run --rm web python -m warehouse shell --type=ipython')
 ARG IPYTHON=no
 
+# By default, Docker has special steps to avoid keeping APT caches in the layers, which
+# is good, but in our case, we're going to mount a special cache volume (kept between
+# builds), so we WANT the cache to persist.
+RUN set -eux; \
```
dstufft, line 123: Did you mean to do this twice? It looks like this modifies the state of the container so that it's *always* caching, so I suspect we don't need it a second time?

ewdurbin: once for docs target and the other for the build target.

ewdurbin: Since build starts `FROM python:3.11.4-slim-bullseye` we need to re do this. It re-uses the cache and adds to it.

When we `FROM python:3.11.4-slim-bullseye` for the final image, we once again just use the cache, but let it get obliterated.

miketheman (approved):
I tested this a little bit, and confirm I don't see any detrimental behavior. I wasn't able to confirm the speedups yet, but hope that they materialize as I touch different parts of the filesystem.

PR #14158, opened 2023-07-18
Trusted publishing: Enforce strict audience checking

### LINKED ISSUE #13887: Trusted publishing: ensure `aud` _only_ contains our audience.
This was raised as a result of [CircleCI's OIDC implementation](https://circleci.com/docs/api/v2/index.html#tag/OIDC-Token-Management) and related discussion here: https://github.com/sigstore/fulcio/issues/591#issuecomment-1579417445.

Currently we accept OIDC tokens for which the `aud` claim contains our required audiences (`pypi` or `testpypi` respectively), however this doesn't preclude the `aud` claim from containing other audiences, because PyJWT will verify any token that simply contains the requested audience: https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L335-L336.

The potential issue is that for a provider like CircleCI, the OIDC token could be shared across multiple 3rd party services with multiple audiences. This means that a CircleCI use that wants to use a token for both `aud=pypi` and `aud=some_other_provider` only gets a single token that would be valid for either audience, which means the other provider technically gains the ability to publish to PyPI as a result.

To mitigate this, we should restrict our verification of OIDC tokens to those that _only_ specify our audience for the `aud` claim.

(While we're at it, we should also update the tests around our JWT verification to do less mocking/stubbing of the 3P library we're using here)

woodruffw:
This bumps `pyjwt` to 2.8.0 and uses the newly-added `strict_aud` option to ensure that we reject JWTs that don't contain exactly a single string `aud` claim.

Closes #13887.


requirements/main.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.11
 # by the following command:
 #
-#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements/main.txt requirements/main.in
+#    pip-compile --allow-unsafe --config=pyproject.toml --generate-hashes --output-file=requirements/main.txt requirements/main.in
```
di, line 5:
```suggestion
#    pip-compile --allow-unsafe --generate-hashes --output-file=requirements/main.txt requirements/main.in
```
I think we need to do a bigger change to our lockfiles which will include this.

woodruffw: Makes sense -- that got added when I ran `make requirements/main.txt`, so that target might need to be tweaked slightly in the mean time.

di:
@woodruffw A test to ensure this doesn't regress would be nice!

woodruffw:
> @woodruffw A test to ensure this doesn't regress would be nice!

Sounds good, doing now.

woodruffw:
Done: I've added a test on both our own JWT helper and a "backstop" test on `pyjwt` itself.

tests/unit/oidc/test_services.py
```diff
@@ -937,22 +963,20 @@ class TestPyJWTBackstop:
     "Backstop" tests against unexpected PyJWT API changes.
     """
 
-    def test_decodes_token_bare_key(self):
-        privkey = rsa.generate_private_key(65537, 2048)
-        pubkey = privkey.public_key()
+    def __init__(self) -> None:
+        self._privkey = rsa.generate_private_key(65537, 2048)
+        self._pubkey = self._privkey.public_key()
```
woodruffw, line 968: Diff is messy here, but I took the opportunity to optimize this test group a bit (RSA keygen is pretty slow and the keys don't need to be unique per-test, so I'm having them reuse the same keypair rather than regenerating each time.)

di (approved):

Issue #14156, opened 2023-07-18
Reject filenames with `_` in project name replaced with `.`, `-` (PEP 427 / PEP 625)


di:
Currently, we are not fully compliant with PEP 427 ([and the updates to it](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode)) or [PEP 625](https://peps.python.org/pep-0625/) because we don't reject filenames like the following:

| Currently allowed | Should be |
|--------|--------|
| `flufl.enum-6.0.1-py3-none-any.whl` | `flufl_enum-6.0.1-py3-none-any.whl` |
| `foo-bar-1.0.0.tar.gz` | `foo_bar-1.0.0.tar.gz` | 

We should have a deprecation period (with sufficient warnings via email on upload) once we've confirmed that the latest versions of the most widely used build tools are producing the correct filenames.

Fixing this is a step towards #12245.

PR #14172, opened 2023-07-20
Handle runs of delimiters in filenames

### LINKED ISSUE #13991: Replace deprecated usage of `pkg_resources`

di:
Fixes the issue raised here: https://github.com/pypi/warehouse/pull/14155#issuecomment-1644195831

Fixes #13991.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2371,10 +2371,7 @@ class TestFileUpload:
         assert resp.status == (
             "400 Start filename for {!r} with {!r}.".format(
                 project.name,
-                pkg_resources.safe_name(project.name)
-                .lower()
-                .replace(".", "_")
-                .replace("-", "_"),
+                project.normalized_name.replace("-", "_"),
             )
         )
 
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -20,7 +20,6 @@ import zipfile
 from cgi import FieldStorage
 from unittest import mock
 
-import pkg_resources
 import pretend
 import pytest
 
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2810,7 +2810,7 @@ class TestFileUpload:
         project = ProjectFactory.create(name=project_name)
         RoleFactory.create(user=user, project=project)
 
-        filename = filename_prefix + "-1.0.0-py3-none-any.whl"
+        filename = f"{filename_prefix}-1.0.0-py3-none-any.whl"
         filebody = _get_whl_testdata(name=filename_prefix, version="1.0.0")
 
         @pretend.call_recorder
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2324,6 +2324,8 @@ class TestFileUpload:
             ("no-way-{version}.tar.gz", "no"),
             ("no_way-{version}-py3-none-any.whl", "no"),
             ("no_way-{version}-py3-none-any.egg", "no"),
+            # multiple delimiters
+            ("foo__bar-{version}-py3-none-any.whl", "foo-.bar"),
         ],
     )
     def test_upload_fails_with_wrong_filename(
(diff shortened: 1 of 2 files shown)


di:
(This also removes one of our usages of `pkg_resources.safe_name`, towards #13991)

takluyver:
I think this probably works, but I don't know precisely what rules packages generating wheels (principally setuptools) have been using for the filename, e.g. if they might have `..` or `._` in the filename.

The [name normalization](https://packaging.python.org/en/latest/specifications/name-normalization/) spec has a one line function to fully normalize a name from any valid starting point. Is there any reason to do it differently for the filename prefix?

```python
def normalize(name):
    return re.sub(r"[-_.]+", "-", name).lower()
```

di:
We don't actually _want_ to normalize the filename prefix, as it should already be normalized. But to handle backwards compatibility with tools that might not be replacing `-` and `.` with `_`, or lowercasing the prefix, we have to do a minimal normalization here, which we'll eventually want to remove via #14156:

```python
filename_prefix = filename_prefix.lower().replace(".", "_").replace("-", "_")
```

If we remove consecutive runs when normalizing the prefix, this would mean any filenames that have any amount of consecutive runs of delimiters in the filename would become valid, because the comparison between the prefix and the project name would not take repeating delimiters into account. Since these filenames are currently invalid, we haven't had an upload with a `._`, `_.`, or `..` in the filename for quite some time:

```
warehouse=> select upload_time, filename from release_files where filename ilike '%\.\_%.whl' order by upload_time desc limit 1;
        upload_time         |                   filename
----------------------------+-----------------------------------------------
 2021-11-03 20:54:45.340789 | hydrotools._restclient-3.0.5-py3-none-any.whl
(1 row)

warehouse=> select upload_time, filename from release_files where filename ilike '%\_\.%.whl' order by upload_time desc limit 1;
        upload_time         |                     filename
----------------------------+---------------------------------------------------
 2021-05-03 00:44:10.904378 | cmc_csci046_.yilinli_trees-1.0.1-py3-none-any.whl
(1 row)

warehouse=> select upload_time, filename from release_files where filename ilike '%\.\.%.whl' order by upload_time desc limit 1;
        upload_time         |                       filename
----------------------------+-------------------------------------------------------
 2020-11-06 20:00:06.720522 | example_..._pkg_megankuooo1234-0.0.1-py3-none-any.whl
(1 row)
```

So TL;DR: I'm not sure we want to become less restrictive by normalizing out character runs in the filename before comparison to the project name.

takluyver:
Thanks, that makes sense. In any case, my main interest in the original
issue was for pre-normalised names, which should work now.

tests/unit/forklift/test_legacy.py
```diff
@@ -2791,16 +2789,29 @@ def storage_service_store(path, file_path, *, meta):
...
     ):
         user = UserFactory.create()
         EmailFactory.create(user=user)
-        project = ProjectFactory.create(name="flufl.enum")
+        project = ProjectFactory.create(name=project_name)
         RoleFactory.create(user=user, project=project)
 
-        filename = "flufl_enum-1.0.0-py3-none-any.whl"
-        filebody = _get_whl_testdata(name="flufl_enum", version="1.0.0")
+        filename = filename_prefix + "-1.0.0-py3-none-any.whl"
```
miketheman, line 2813: f-strings!
```suggestion
        filename = f"{filename_prefix}-1.0.0-py3-none-any.whl"
```

miketheman (approved):
Good to go. Accept the suggestion if you like, it's okay if you don't!

PR #14194, opened 2023-07-23
Remember device for 30 days option

ristomcgehee:
This gives the user the option to not have to provide 2nd factor authentication on a device for 30 days. A new cookie `remember_device` will contain a signed token that allows the user to skip two-factor authentication (assuming the token is valid). To minimize data being sent by the browser, this cookie is only sent for requests with the path `accounts/login`.

This is how the 2FA page looks:
<details>
<summary>With authenticator app and webauthn</summary>

image
</details>

<details>
<summary>With only webauthn</summary>

image
</details>

By default "Remember this device for 30 days" will not be checked. 

I originally created a different PR for this issue (https://github.com/pypi/warehouse/pull/13166) but since we decided to take a substantially different approach, I figured this merited a new, clean PR.

This change requires adding a `TOKEN_REMEMBER_DEVICE_SECRET` environment variable to the production configuration. 

Closes https://github.com/pypi/warehouse/issues/5867

I'll create a new issue for how these tokens can be revoked. One approach could be to invalidate the tokens when the user changes their password.

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -46,7 +46,11 @@ from warehouse.accounts.interfaces import (
     TooManyFailedLogins,
     TooManyPasswordResetRequests,
 )
-from warehouse.accounts.views import two_factor_and_totp_validate
+from warehouse.accounts.views import (
+    REMEMBER_DEVICE_COOKIE,
+    REMEMBER_DEVICE_SECONDS,
+    two_factor_and_totp_validate,
+)
 from warehouse.admin.flags import AdminFlag, AdminFlagValue
 from warehouse.events.tags import EventTag
 from warehouse.metrics.interfaces import IMetricsService
@@ -435,8 +439,14 @@ class TestLogin:
         assert result.headers["Location"] == "/the-redirect"
 
     @pytest.mark.parametrize("redirect_url", ["test_redirect_url", None])
-    def test_two_factor_auth(self, pyramid_request, redirect_url, token_service):
+    def test_two_factor_auth(
+        self, monkeypatch, pyramid_request, redirect_url, token_service
+    ):
         token_service.dumps = lambda d: "fake_token"
+
+        monkeypatch.setattr(
+            views, "_check_remember_device_token", lambda *a, **kw: False
+        )
 
         user_service = pretend.stub(
             find_userid=pretend.call_recorder(lambda username: 1),
@@ -696,11 +706,20 @@ class TestTwoFactor:
 
     @pytest.mark.parametrize("redirect_url", ["test_redirect_url", None])
     @pytest.mark.parametrize("has_recovery_codes", [True, False])
+    @pytest.mark.parametrize("remember_device", [True, False])
     def test_totp_auth(
-        self, monkeypatch, pyramid_request, redirect_url, has_recovery_codes
+        self,
+        monkeypatch,
+        pyramid_request,
+        redirect_url,
+        has_recovery_codes,
+        remember_device,
     ):
         remember = pretend.call_recorder(lambda request, user_id: [("foo", "bar")])
         monkeypatch.setattr(views, "remember", remember)
+
+        _remember_device = pretend.call_recorder(lambda *a, **kw: None)
+        monkeypatch.setattr(views, "_remember_device", _remember_device)
 
         query_params = {"userid": str(1)}
         if redirect_url:
@@ -755,6 +774,7 @@ class TestTwoFactor:
         form_obj = pretend.stub(
             validate=pretend.call_recorder(lambda: True),
             totp_value=pretend.stub(data="test-otp-secret"),
+            remember_device=pretend.stub(data=remember_device),
         )
         form_class = pretend.call_recorder(lambda d, user_service, **kw: form_obj)
         pyramid_request.route_path = pretend.call_recorder(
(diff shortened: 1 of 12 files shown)


warehouse/templates/accounts/two-factor.html
```diff
@@ -112,6 +120,13 @@ <h2>{% trans %}Authenticate with an app{% endtrans %}</h2>
             <input type="submit" value="{% trans %}Verify{% endtrans %}" class="button button--primary">
           </div>
 
+          <div class="form-group">
+            <label>
+              <input type="checkbox" name="remember_device" value="true" />
```
miketheman, line 125: **question:** ‏does this input also need an `id`? I noticed it differed from the webauthn one

warehouse/config.py
```diff
@@ -238,6 +239,13 @@ def configure(settings=None):
         coercer=int,
         default=300,
     )
+    maybe_set(
+        settings,
+        "token.remember_device.max_age",
+        "TOKEN_REMEMBER_DEVICE_MAX_AGE",
+        coercer=int,
+        default=2592000,  # 30 days,
```
miketheman, line 247: I've started to want to migrate "magic seconds" to more readable values.

```python
from datetime import timedelta
...
        default=timedelta(days=30).total_seconds(),
...
```

warehouse/accounts/views.py
```diff
@@ -91,6 +91,8 @@
 from warehouse.utils.http import is_safe_url
 
 USER_ID_INSECURE_COOKIE = "user_id__insecure"
+REMEMBER_DEVICE_COOKIE = "remember_device"
+REMEMBER_DEVICE_SECONDS = 2592000  # 30 days
```
miketheman, line 95: **question:** ‏shouldn't this be retrieved from config?

warehouse/accounts/views.py
```diff
@@ -449,6 +456,47 @@ def webauthn_authentication_validate(request):
...
+    remember_device_data = {"user_id": str(userid)}
+    token_service = request.find_service(ITokenService, name="remember_device")
+    token = token_service.dumps(remember_device_data)
+    response.set_cookie(
+        REMEMBER_DEVICE_COOKIE,
+        token,
+        max_age=REMEMBER_DEVICE_SECONDS,
+        httponly=True,
+        secure=request.scheme == "https",
+        samesite=b"lax",
```
miketheman, line 488: **question:** ‏can you help explain why we need to allow for lax (the default), vs strict? I think you're following the pattern in sessions.py, but I think we might be able to tighten this one up a bit more.

tests/unit/accounts/test_views.py
```diff
@@ -1065,12 +1098,86 @@ def test_webauthn_validate(self, monkeypatch, pyramid_request, has_recovery_code
...
+        token_service = pretend.stub(loads=lambda *a: {"user_id": str(999)})
+        request = pretend.stub(
+            cookies=pretend.stub(get=lambda *a, **kw: "token"),
+            find_service=lambda interface, **kwargs: {
+                ITokenService: token_service,
+            }[interface],
+        )
+        assert not views._check_remember_device_token(request, 1)
+
+    def test_remember_device(self, monkeypatch, pyramid_request):
```
miketheman, line 1152: **question:** ‏looks like you're recreating a `pyramid_request` inline of the test function, so you don't need the fixture, and you're not using the `monkeypatch` fixture either - should they be removed?

If the only part of `pyramid_request` you want to override is the `find_service`, you could express like:

```python
pyramid_request.find_service = lambda interface, **kwargs: {
    ITokenService: token_service
}[interface]
```

And remove the other parts that aren't necessary if the test works without them.

miketheman (commented):
Added comments inline - overall looks like a nice patch.

warehouse/config.py
ristomcgehee: Good idea!

warehouse/accounts/views.py
ristomcgehee: Your right that I was following the pattern from session.py. For this cookie, we don't need lax, but I also don't think we would benefit from strict. Since this cookie is restricted to `account/login`, if we set it to lax and the user navigates to `https://pypi.org/account/login` from another page, this cookie would be sent, but it wouldn't be used the server since we only use it on POST requests. If we set it to strict, the only difference is that it wouldn't be sent when the user first navigates to `https://pypi.org/account/login` from another page. I'll change it to strict because in general I think one should use strict unless one has a need for lax.

warehouse/accounts/views.py
ristomcgehee: Yeah, I can get this from config.

warehouse/templates/accounts/two-factor.html
ristomcgehee: It don't think it needs an `id` because I'm not referencing it in JS or CSS or anywhere else.

ristomcgehee: The `remember_device_webauthn` input is being referenced in `webauthn.js`.

warehouse/config.py
```diff
@@ -238,6 +241,20 @@ def configure(settings=None):
         coercer=int,
         default=300,
     )
+    maybe_set(
+        settings,
+        "remember_device.days",
+        "REMEMBER_DEVICE_DAYS",
+        coercer=int,
+        default=30,
```
miketheman, line 249: **praise:** ‏I like this even more, since then we don't even need to set a value unless we want to override the default of 30 days.

miketheman (approved):
I'm pretty good with this, but would appreciate a second review from @di as well.

ewdurbin:
random `TOKEN_REMEMBER_DEVICE_SECRET` values generated and set for both pypi and testpypi deployments, so this can be shipped whenever reviews are complete.

ristomcgehee:
@di It looks like this PR is waiting for your review. Btw, if you re-run the failed checks, they should pass (temporary github issue).

di (approved):
Thanks! FYI, translations need to be updated.

Issue #13771, opened 2023-05-26
Prevent removing the last 2FA method associated with an account


dstufft:
2FA will be required on PyPI, and for some projects is already required. We should prevent an account from regressing in their 2FA-ness and prevent the last 2FA method from being removed before another method is added.

ristomcgehee:
Hi, I'd be happy to work on this one. Here are the implementation details I'd likely follow. Feel free to voice suggested changes:
* On the Account Settings page, if there is only one 2FA method, the Remove button will be disabled and greyed out.
* If you mouse over the disabled Remove button, the cursor will turn into the "not" symbol and a tool tip will say "Not allowed to remove last 2FA method" (similar behavior to the Your Projects page with the Manage and View buttons).
* The backend will also check that at least two 2FA methods are present before removing a method.

PR #14195, opened 2023-07-24
Prevent removing the last 2FA method associated with an account

ristomcgehee:
With this change, a user will only be allowed to remove a 2FA method if there are 2 or more 2FA methods on the account. When there is only 1, the "Remove" button will be disabled. The backend also checks if there are 2 or more 2FA methods when trying to remove.

I noticed that when the "Remove" button is disabled, the cursor doesn't change to "not-allowed" when hovering, nor does the title appear in a tooltip. That's because https://github.com/pypi/warehouse/pull/13110, added `pointer-events: none` to [disabled buttons](https://github.com/pypi/warehouse/blob/fe6455c0a946e81f61d72edc1049f536d8bba903/warehouse/static/sass/blocks/_button.scss#L121). On the "Your Projects" page, disabled buttons show the title and the "not-allowed" cursor because [_snippet.scss](https://github.com/pypi/warehouse/blob/fe6455c0a946e81f61d72edc1049f536d8bba903/warehouse/static/sass/blocks/_snippet.scss#L92) sets `pointer-events: auto`. I wasn't sure the reason behind setting `pointer-events: none` in `_button.scss`, but if this behavior change is unintended, we could address this in another PR.

Closes https://github.com/pypi/warehouse/issues/13771

### DIFF
--- a/warehouse/manage/views/__init__.py
+++ b/warehouse/manage/views/__init__.py
@@ -592,6 +592,10 @@ class ProvisionTOTPViews:
             )
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
+        if not self.request.user.has_multiple_2fa:
+            self.request.session.flash("Cannot remove last 2FA method", queue="error")
+            return HTTPSeeOther(self.request.route_path("manage.account"))
+
         form = DeleteTOTPForm(
             formdata=MultiDict(
                 {
@@ -721,6 +725,10 @@ class ProvisionWebAuthnViews:
             self.request.session.flash(
                 "There is no security device to delete", queue="error"
             )
+            return HTTPSeeOther(self.request.route_path("manage.account"))
+
+        if not self.request.user.has_multiple_2fa:
+            self.request.session.flash("Cannot remove last 2FA method", queue="error")
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
         form = DeleteWebAuthnForm(
--- a/warehouse/manage/views/__init__.py
+++ b/warehouse/manage/views/__init__.py
@@ -592,7 +592,7 @@ class ProvisionTOTPViews:
             )
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
-        if not self.request.user.has_multiple_2fa:
+        if self.request.user.has_single_2fa:
             self.request.session.flash("Cannot remove last 2FA method", queue="error")
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
@@ -727,7 +727,7 @@ class ProvisionWebAuthnViews:
             )
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
-        if not self.request.user.has_multiple_2fa:
+        if self.request.user.has_single_2fa:
             self.request.session.flash("Cannot remove last 2FA method", queue="error")
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
--- a/warehouse/templates/manage/manage_base.html
+++ b/warehouse/templates/manage/manage_base.html
@@ -61,16 +61,16 @@
         <tr>
           <th scope="row">{% trans %}Authentication application (<abbr title="time-based one-time password">TOTP</abbr>){% endtrans %}</th>
           <td class="table__align-right">
-            {% if user.has_multiple_2fa %}
+            {% if user.has_single_2fa %}
+            <button class="button" disabled title="{% trans %}Cannot remove last 2FA method{% endtrans %}" >
+            {% trans %}Remove{% endtrans %}
+            </button>
+            {% else %}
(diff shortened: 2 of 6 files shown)


warehouse/manage/views/__init__.py
```diff
@@ -592,6 +592,10 @@ def delete_totp(self):
             )
             return HTTPSeeOther(self.request.route_path("manage.account"))
 
+        if not self.request.user.has_multiple_2fa:
```
di, line 595:
```suggestion
        if self.request.user.has_single_2fa:
```

warehouse/manage/views/__init__.py
```diff
@@ -723,6 +727,10 @@ def delete_webauthn(self):
...
```
di, line 730:
```suggestion
        if not self.request.user.has_single_2fa:
```

warehouse/templates/manage/manage_base.html
```diff
@@ -61,11 +61,17 @@ <h2 class="sub-title">{% trans %}Two factor authentication (2FA){% endtrans %}</
         <tr>
           <th scope="row">{% trans %}Authentication application (<abbr title="time-based one-time password">TOTP</abbr>){% endtrans %}</th>
           <td class="table__align-right">
+            {% if user.has_multiple_2fa %}
```
di, line 64:
```suggestion
            {% if user.has_multiple_2fa %}
            <button class="button" disabled title="{% trans %}Cannot remove last 2FA method{% endtrans %}" >
            {% trans %}Remove{% endtrans %}
            </button>
            {% else %}
```

warehouse/templates/manage/manage_base.html
```diff
@@ -61,11 +61,17 @@ <h2 class="sub-title">{% trans %}Two factor authentication (2FA){% endtrans %}</
...
             <a href="#disable-totp" class="button button--secondary">{% trans %}Remove{% endtrans %}</a>
             {% set title=gettext("Remove authentication application") %}
             {% set confirm_button_label=gettext("Remove application") %}
             {% set action=request.route_path('manage.account.totp-provision') %}
             {{ confirm_password_modal(title=title, action=action, slug="disable-totp", warning=False, confirm_button_label=confirm_button_label) }}
+            {% else %}
+            <button class="button" disabled title="{% trans %}Cannot remove last 2FA method{% endtrans %}" >
+            {% trans %}Remove{% endtrans %}
+            </button>
+            {% endif %}
```
di, line 74:
```suggestion
            {% endif %}
```

warehouse/templates/manage/manage_base.html
```diff
@@ -84,6 +91,11 @@ <h2 class="sub-title">{% trans %}Two factor authentication (2FA){% endtrans %}</
               <input type="hidden" name="label" value="{{ credential.label }}" autocomplete="off">
             {% endset %}
             {{ confirm_modal(title=title, label=gettext("Device name"), confirm_name="device_name", confirm_string=credential.label, confirm_button_label=confirm_button_label, slug=slug, extra_fields=extra_fields, action=action, warning=False, confirm_string_in_title=False) }}
+            {% else %}
+            <button class="button" disabled title="{% trans %}Cannot remove last 2FA method{% endtrans %}" >
+            {% trans %}Remove{% endtrans %}
+            </button>
+            {% endif %}
```
di, line 98:
```suggestion
            {% endif %}
```

di:
> I noticed that when the "Remove" button is disabled, the cursor doesn't change to "not-allowed" when hovering, nor does the title appear in a tooltip. That's because https://github.com/pypi/warehouse/pull/13110, added pointer-events: none to [disabled buttons](https://github.com/pypi/warehouse/blob/fe6455c0a946e81f61d72edc1049f536d8bba903/warehouse/static/sass/blocks/_button.scss#L121). On the "Your Projects" page, disabled buttons show the title and the "not-allowed" cursor because [_snippet.scss](https://github.com/pypi/warehouse/blob/fe6455c0a946e81f61d72edc1049f536d8bba903/warehouse/static/sass/blocks/_snippet.scss#L92) sets pointer-events: auto. I wasn't sure the reason behind setting pointer-events: none in _button.scss, but if this behavior change is unintended, we could address this in another PR.

I _think_ I just did this in an effort to not have it change to the more active pointer for the button, without realizing that there was a better option. I think that that setting `pointer-events: auto` across all buttons probably makes more sense!

ristomcgehee:
> I think it's a little cleaner if we invert the logic here. (I'll let you update the tests accordingly!)

Agreed :+1:

ristomcgehee:
> I think that that setting `pointer-events: auto` across all buttons probably makes more sense!

I'll open a separate PR for this.

di (approved):

PR #14234, opened 2023-07-31
Change wording for API Token

### LINKED ISSUE #11874: Unclear wording for creating API tokens
The text "Add an API token" is misleading. A better wording would be "Create an API token" as the token is created by the platform, not uploaded by the user.

Also: the help message "What is this token for?" under "Token name" is not useful. What are we supposed to call our tokens ? I called mine "API Token". Maybe I should have called it "First token" ?

Even better: generate the name automatically unless the user knows what they want (e.g. "API token for Maria", etc.).

arshm06:
I changed the wording so that it says "Create API Token". I also edited the help message so that it actually helps the user.

Fixes #11874

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -3167,7 +3167,6 @@ msgid "Active API tokens for this account"
 msgstr ""
 
 #: warehouse/templates/manage/account.html:449
-#: warehouse/templates/manage/account/token.html:17
 msgid "Add API token"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


di:
Hi @arshm06, this looks good but you will need to [update translations](https://warehouse.pypa.io/development/submitting-patches.html#translations) as well before we can merge.

arshm06:
@di I have updated translations. Thank you!

di (approved):
Thanks!

PR #14241, opened 2023-07-31
docs: add system architecture diagrams

miketheman:
Not 100% comprehensive, but a good starting point.

Related to #2794 

To see generated HTML, run:

```shell
make dev-docs && python -m http.server -d docs/dev/_build/html/
```

and browse to http://localhost:8000/architecture.html


di:
The docs are also previewed here: https://warehouse--14241.org.readthedocs.build/architecture.html

di:
The only thing missing here I think is that there should probably be a "Redis" component for uploads as well, because this is how we handle rate-limiting for all endpoints, including uploads.

miketheman:
> The only thing missing here I think is that there should probably be a "Redis" component for uploads as well, because this is how we handle rate-limiting for all endpoints, including uploads.

Added!

miketheman:
> would it make more sense for this to be added to our docs at [docs.pypi.org](https://docs.pypi.org/) instead?

I considered this, and didn't know the right answer. I thought about how most of the information is relevant to a warehouse developer (and others). The docs.pypi.org site seems more end-user-focused. I'm happy to put it whereever it makes most sense - but this felt more correct to me.

di:
I think that delineation is good enough for me, for now. Thanks!

di (approved):

PR #14244, opened 2023-08-01
Bring back disabled cursor and tooltip for disabled buttons

ristomcgehee:
I noticed that when most buttons are disabled, the cursor doesn't change to "not-allowed" when hovering, nor does the title appear in a tooltip. This change happened in https://github.com/pypi/warehouse/pull/13110 from adding `pointer-events: none` to [disabled buttons](https://github.com/pypi/warehouse/blob/fe6455c0a946e81f61d72edc1049f536d8bba903/warehouse/static/sass/blocks/_button.scss#L121). In https://github.com/pypi/warehouse/pull/14195#issuecomment-1648501825, Dustin commented saying that it probably makes sense to have `pointer-events: auto`.

### DIFF
--- a/warehouse/static/sass/blocks/_button.scss
+++ b/warehouse/static/sass/blocks/_button.scss
@@ -118,7 +118,7 @@
     border-color: darken($background-color, 5%);
     color: darken($background-color, 12%);
     text-decoration: line-through;
-    pointer-events: none;
+    pointer-events: auto;
 
     &:focus,
     &:hover,


miketheman:
This change actually re-enabled the wrong behavior, in that a user may now click the disabled "button", and effectively bypass the required checkboxes.

Reasoning here is that the `disabled` attribute does nothing on a `a href` element, only works on a `button` element.

di:
Aha, that's why it was set to `pointer-events: none`. Seems like maybe we should be using `button` elements everywhere instead of `a` elements?

miketheman:
> Seems like maybe we should be using `button` elements everywhere instead of `a` elements?

Yes, we likely should be using `button` since then we can take advantage of that.

That's a larger change - and would need changes in blocks/_buttons.scss as well as the `macro confirm_button/confirm_password_button` generators in manage_base.html

PR #14258, opened 2023-08-02
Clarify how to replace last 2FA

### LINKED ISSUE #14233: Indicate that a second 2FA device must be added before the first can be removed
As a follow-up to #13771, we should make it more clear in the UI that when there is only one 2FA device on an account, a new device must be added before the existing one can be removed.

Currently the button is just disabled, with no tooltip or other message to guide the user for what to do next:

image

(cc @ristomcgehee)

ristomcgehee:
Two UI changes in this PR to clarify how you can replace your last 2FA method:
- Adds "Once 2FA is enabled, it cannot be disabled." This displays both before and after 2FA is enabled.
- Adds "To remove this 2FA method, you must first add another method". This only displays when the user has a single 2FA method.

image

Closes #14233

### DIFF
--- a/warehouse/templates/manage/manage_base.html
+++ b/warehouse/templates/manage/manage_base.html
@@ -56,6 +56,15 @@
         </tr>
       </thead>
       <tbody>
+        {% if user.has_single_2fa %}
+        <tr>
+          <td colspan="2" class="table__no-bottom-border">
+            <div class="callout-block callout-block--warning no-top-margin no-bottom-margin">
+              {% trans %}To remove this 2FA method, you must first add another method.{% endtrans %}
+            </div>
+          </td>
+        </tr>
+        {% endif %}
         {% if user.totp_secret %}
         <tr>
           <th scope="row">{% trans %}Authentication application (<abbr title="time-based one-time password">TOTP</abbr>){% endtrans %}</th>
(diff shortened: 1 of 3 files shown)


warehouse/templates/manage/manage_base.html
```diff
@@ -102,6 +101,11 @@ <h2 class="sub-title">{% trans %}Two factor authentication (2FA){% endtrans %}</
 
       </tbody>
     </table>
+    {% if user.has_single_2fa %}
+    <div class="callout-block">
```
di, line 105:
```suggestion
    <div class="callout-block callout-block--warning">
```

warehouse/templates/manage/manage_base.html
```diff
@@ -102,6 +101,11 @@ <h2 class="sub-title">{% trans %}Two factor authentication (2FA){% endtrans %}</
...
+        <p>{% trans %}To remove this 2FA method, you must first add another method.{% endtrans %}</p>
+    </div>
+    {% endif %}
```
di, line 108: I think I'd prefer to move this up under the horizontal line, but before the table of two factor methods.

ristomcgehee: I've made this change. I'm not great with CSS, so I hope my approach isn't too hacky. I've updated the pic in the PR desc.

di (approved):
Thanks!

Issue #14257, opened 2023-08-02
Allowing editing existing 2FA application


ristomcgehee:
**What's the problem this feature will solve?**

When you have a TOTP application enabled on your account and you want to replace it with a different TOTP application, you have to remove the existing TOTP app and add a new one. This is especially problematic now that we [prevent removing the last 2FA method](https://github.com/pypi/warehouse/issues/13771) as reported in https://github.com/pypi/warehouse/issues/14233#issuecomment-1658502542.

**Describe the solution you'd like**

- When a user has TOTP configured, instead of the button saying "Add 2FA with authentication application", it says "Edit 2FA with authentication application".
- When the user clicks the button, the "Set up 2FA with an authentication application (TOTP)" view comes up and the user can scan the QR code and enter a TOTP code.
- On the Account Settings page, there will still be a Remove button next to the TOTP option. If the user has no other 2FA, that button will be disabled.
- If the user has TOTP and 1 or more webauthn devices, the button will still be "Edit 2FA with authentication application".

**Additional context**

I'll work on this issue.

PR #14260, opened 2023-08-02
Allowing editing existing 2FA TOTP application

ristomcgehee:
Previously, if a user wanted to replace their existing TOTP application, they would have had to remove it and add a new one. This is no longer possible if the user doesn't have webauthn set up also, due to https://github.com/pypi/warehouse/pull/14195. Now, users will be able to edit their existing TOTP application:
- When a user has TOTP configured, instead of the button saying "Add 2FA with authentication application", it says "Edit 2FA with authentication application".
- When the user clicks the button, the "Set up 2FA with an authentication application (TOTP)" view comes up and the user can scan the QR code and enter a TOTP code.
- On the Account Settings page, there is still a Remove button next to the TOTP option. If the user has no other 2FA, that button is disabled.
- If the user has TOTP and 1 or more webauthn devices, the button still is "Edit 2FA with authentication application".

Closes https://github.com/pypi/warehouse/issues/14257, closes https://github.com/pypi/warehouse/issues/14274.


warehouse/manage/views/__init__.py
```diff
@@ -544,27 +536,21 @@ def validate_totp_provision(self):
...
             )
             self.request.session.clear_totp_secret()
             self.request.user.record_event(
-                tag=EventTag.Account.TwoFactorMethodAdded,
+                tag=EventTag.Account.TwoFactorMethodEdited
+                if old_totp_secret
+                else EventTag.Account.TwoFactorMethodAdded,
                 request=self.request,
                 additional={"method": "totp"},
             )
```
di, line 556: Instead of introducing a new event type, I think we should just emit two back-to-back events, "Removed" + "Added".

warehouse/manage/views/__init__.py
```diff
@@ -527,12 +523,8 @@ def totp_provision(self):
...
         totp_secret = self.user_service.get_totp_secret(self.request.user.id)
         if totp_secret:
-            self.request.session.flash(
-                "Account cannot be linked to more than one authentication "
-                "application at a time",
-                queue="error",
-            )
-            return HTTPSeeOther(self.request.route_path("manage.account"))
+            # User wants to replace their existing TOTP application.
+            self.request.session.clear_totp_secret()
```
di, line 527: Is clearing the secret from the session here strictly necessary? I would expect us to be able to remove this entire section.

ristomcgehee: Strictly necessary, no, but I think it's the right behavior. If we don't clear the session here, the user will still be able to view the QR code and update their TOTP settings, however they would end up with the same TOTP secret as they had before. I think it's best to change to a new TOTP secret because the reason the user might be replacing their TOTP is because they lost their old phone and don't want someone else to be able to use it to authenticate as them. I can update the comment in the code explaining this more.

di (approved):

warehouse/manage/views/__init__.py
di: Thanks, it was unclear to me that this secret will be used to render the page if it's available, I agree it should be cleared. Made a slight simplification to just always clear this before rending the page instead.

miketheman (approved):

Issue #12243, opened 2022-09-21
[FR] Multiline (and rich?) yank reason


webknjaz:
**What's the problem this feature will solve?**

Having a comprehensive explanation for yanking. I think it is important not to limit the maintainers with a very small input box. It would be useful to be able to show a descriptive message to the end-users too (related to FR #9837).

**Describe the solution you'd like**

I want `<textarea>` in the UI, not an `<input>`.
It could then also be displayed on the individual release pages and only the first line (maybe) would appear in badge tooltips (as suggested in #9837).

Not sure if allowing renderable Markdown would be safe but, I'd at least make the links clickable there too.

**Additional context**

N/A

di:
We likely need to set a maximum allowable length for this field on https://peps.python.org/pep-0592/ before being able to support this, as we likely don't want this to be completely unbounded.

arshm06:
I have submitted my pull request. Thank you!

PR #14261, opened 2023-08-02
Made yank reason multiline and possible richer

arshm06:
Issue #12243

I have changed <input> to <textarea> so that users could have a larger input box.


miketheman (changes requested):
Thanks for the contribution, however this should not proceed forward until the other aspect of this has been handled.
Did you see the comment here? https://github.com/pypi/warehouse/issues/12243#issuecomment-1256392902
The data type should be constrained before we allow the user interface to publish more data.

firefly9966 (approved):

PR #14270, opened 2023-08-03
Migrate admin_flags and ses_{emails,events} tables to modern sqlalchemy

ewdurbin:

PR #14271, opened 2023-08-04
Migrate vulnerabilities models to declarative sqlalchemy

di:
Follow-on to #14266 and #14270.


warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,64 @@
...
 
     # The URL for the vulnerability report at the source
     # e.g. "https://osv.dev/vulnerability/PYSEC-2021-314"
-    link = mapped_column(String)
+    link: Mapped[str]
 
     # Alternative IDs for this vulnerability
     # e.g. "CVE-2021-12345"
-    aliases = mapped_column(ARRAY(String))  # type: ignore[var-annotated]
+    aliases: Mapped[list[str]] = Column(ARRAY(String))
```
di, line 54: Not sure this is right, I thought this just would be:

```python
aliases: Mapped[list[str]]
```

but it complains:

> `sqlalchemy.exc.ArgumentError: Could not locate SQLAlchemy Core type for Python type list[str] inside the 'aliases' attribute Mapped annotation`

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
 
     # The URL for the vulnerability report at the source
     # e.g. "https://osv.dev/vulnerability/PYSEC-2021-314"
-    link = mapped_column(String)
+    link: Mapped[str]
 
     # Alternative IDs for this vulnerability
     # e.g. "CVE-2021-12345"
-    aliases = mapped_column(ARRAY(String))  # type: ignore[var-annotated]
+    aliases = mapped_column(ARRAY(String))
```
di, line 55: Still not sure what to do with this.

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
+    vulnerability_source: Mapped[str]
+    vulnerability_id: Mapped[str]
 
 
 class VulnerabilityRecord(db.ModelBase):
     __tablename__ = "vulnerabilities"
 
-    source = mapped_column(String, primary_key=True)
-    id = mapped_column(String, primary_key=True)
+    source: Mapped[str] = mapped_column(String, primary_key=True)
```
miketheman, line 46: With passing `str`, we don't need to specify the column type unless it's special somehow (like with a specific length)

```suggestion
    source: Mapped[str] = mapped_column(primary_key=True)
```

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
+    id: Mapped[str] = mapped_column(primary_key=True)
 
     # The URL for the vulnerability report at the source
     # e.g. "https://osv.dev/vulnerability/PYSEC-2021-314"
-    link = mapped_column(String)
+    link: Mapped[str]
```
miketheman, line 51: Since the default column behavior is implicit `nullable=True`, we need to annotate the type that this could be null:
```suggestion
    link: Mapped[str | None]
```

Similar annotations should be made for the other implicit nullables in this class

warehouse/integrations/vulnerabilities/models.py
miketheman: Totally fair! Since ARRAY is not part of the SQL standard, we'll still need to provide the `mapped_column` for this case. The correct way to annotate this line is:

```suggestion
    aliases: Mapped[list[str] | None] = mapped_column(ARRAY(String))
```

Same logic applies for `fixed_in`

warehouse/packaging/models.py
```diff
@@ -49,7 +49,6 @@
 from warehouse.accounts.models import User
 from warehouse.classifiers.models import Classifier
 from warehouse.events.models import HasEvents
-from warehouse.integrations.vulnerabilities.models import VulnerabilityRecord
```
miketheman, line 52: **question:** ‏was this to avoid a circular import, or for some other reason? When we come to annotate this line, we'll likely need to bring the import back.

miketheman (changes requested):

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
-    details = mapped_column(String)
+    details: Mapped[str]
 
     # A short, plaintext summary of the vulnerability
-    summary = mapped_column(String)
+    summary: Mapped[str]
 
     # Events of introduced/fixed versions
-    fixed_in = mapped_column(ARRAY(String))  # type: ignore[var-annotated]
+    fixed_in = mapped_column(ARRAY(String))
```
di, line 64:
```suggestion
    fixed_in: Mapped[list[str] | None] = mapped_column(ARRAY(String))
```

warehouse/packaging/models.py
di: Yes, this was to avoid a circular import, wasn't really clear to me how we'd handle circular imports with types + two way relationships.

warehouse/integrations/vulnerabilities/models.py
di: Wait, since this didn't originally have `nullable=True`, shouldn't this be `Mapped[str]` and not `Mapped[str | None]`? Doesn't "nullable" imply `Optional`?

miketheman: `Column.nullable` Defaults to `True`, which implies `Optional` aka `foo | None` (these are the same as of Python 3.10 / [PEP-604](https://peps.python.org/pep-0604/))

`Mapped[str]` == `nullable=False` (aka required), so to preserve the implicit `nullable=True`, we make denote it optional.

warehouse/packaging/models.py
miketheman: Hmmm I don't know the answer to that yet - something to ruminate on.

I wonder if there's a need to express the relationship on both sides if we use back_populates - we should dig into that at some point, but definitely not a blocker.

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
+    link: Mapped[str | None]
 
     # Alternative IDs for this vulnerability
     # e.g. "CVE-2021-12345"
-    aliases = mapped_column(ARRAY(String))  # type: ignore[var-annotated]
+    aliases: Mapped[list[str] | None] = mapped_column(ARRAY(String))
 
     # Details about the vulnerability
-    details = mapped_column(String)
+    details: Mapped[str]
```
miketheman, line 58:
```suggestion
    details: Mapped[str | None]
```

warehouse/integrations/vulnerabilities/models.py
```diff
@@ -10,71 +10,65 @@
...
 
     # A short, plaintext summary of the vulnerability
-    summary = mapped_column(String)
+    summary: Mapped[str]
```
miketheman, line 61:
```suggestion
    summary: Mapped[str | None]
```

miketheman (changes requested):
Two more that need handling

warehouse/integrations/vulnerabilities/models.py
di: Aha, got it!

miketheman (approved):

Issue #13766, opened 2023-05-26
Require 2FA for creating new Organizations


dstufft:
Ratchet more into into 2FA by requiring 2FA to create/request an organization.

ristomcgehee:
I can work on this one. Here's the approach I think I would take:
On the organization page, under "Request a new organization", I would put a box saying "You must first enable 2FA on your account before requesting an organization". Part of that text would be a hyperlink to the 2FA section on the Account Settings page. I would also disable the "Create" button.

di:
@ristomcgehee Sounds like a good plan!

PR #14280, opened 2023-08-06
Require 2FA for requesting new organization

ristomcgehee:
If user does not have 2FA enabled on their account, a call-out block states that they need to enable it. The form for creating a new org is hidden (I thought that was better than merely disabling the form).

image

Closes https://github.com/pypi/warehouse/issues/13766

### DIFF
--- a/warehouse/templates/manage/organizations.html
+++ b/warehouse/templates/manage/organizations.html
@@ -176,6 +176,7 @@
   {{ form_error_anchor(create_organization_application_form) }}
   <section id="create-organization">
     <h2 class="sub-title">{% trans %}Request a new organization{% endtrans %}</h2>
+    {% if request.user.has_two_factor %}
     <form method="POST">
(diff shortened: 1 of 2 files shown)


warehouse/templates/manage/organizations.html
```diff
@@ -283,6 +284,13 @@ <h2 class="sub-title">{% trans %}Request a new organization{% endtrans %}</h2>
 
       <input value="{% trans %}Create{% endtrans %}" class="button button--primary" type="submit">
     </form>
+    {% else %}{# user has not enabled 2FA #}
+    <div class="callout-block no-top-margin">
+      {% trans href=request.route_path('manage.account.two-factor') %}
+      You must first enable <a href="{{ href }}">two-factor authentication</a> on your account before requesting an organization.
```
di, line 290:
```suggestion
      You must first enable <a href="{{ request.route_path("manage.account.two-factor") }}">two-factor authentication</a> on your account before requesting an organization.
```

di (approved):

PR #14292, opened 2023-08-07
chore: update dependabot config for groups

miketheman:
Inform Dependabot that some deps should move together.
See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

Bring back updates to `boto*` on a less frequent cadence.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,11 +13,25 @@ updates:
   ignore:
   # Always ignore elasticsearch, future versions are always incompatible with our provider
   - dependency-name: "elasticsearch"
-  # These update basically every day, and 99.9% of the time we don't care
-  - dependency-name: "boto3"
-  - dependency-name: "boto3-stubs"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -13,11 +13,25 @@ updates:
...
+  schedule:
+    interval: monthly
+  groups:
+    # Items in these groups are combined in a single PR
+    boto3:
+      patterns:
+        - "boto3*"
+    botocore:
+      patterns:
+        - "botocore*"
```
di, line 33:
```suggestion
    boto-deps:
      patterns:
        - "boto3*"
        - "botocore*"
```
Would it make sense to tie both of these together as well?

miketheman: Yeah, I'm fine with that too!

di (approved):

Issue #13762, opened 2023-05-26
Require 2FA enrollment for all new users


ewdurbin:
**What's the problem this feature will solve?**

Since we will require it for all uploads beginning in 2024, we should add 2FA enrollment for all new users as soon as possible.

**Describe the solution you'd like**

Add 2FA enrollment to routine new user registration flow.

miketheman:
## User Registration experience today:

1. Visit registration page
2. Complete registration signup form
3. User is now logged in, and an email has been sent for verification

At this point, there's nothing preventing the user from **never** verifying their email address, other than wanting to publish a package - Create API Token requires a verified email.

4. User verifies email, they can create an API token, and also start 2FA enrollment

## The question is

At what step of the flow should we **require** the user to enroll in 2FA?

- Should we perform a similar check as email verification for whether 2FA is enrolled prior to creating an API token?
- Should we require a verified email address/2FA before allowing logged-in actions?

Basically, something about the user registration flow has to change, and today we allow unverified registrations to continue to use the logged-in aspects of PyPI - manage collaborators, yank/delete releases, setup trusted publishing, etc.

di:
> Should we require a verified email address/2FA before allowing logged-in actions?

I think this makes sense.

ristomcgehee:
How does this sound for a new user registration flow?

Instead of having a single view for registering a user, the user goes through these 5 steps, each in its own view:
1. Basic Info - Enters name, email, username, and password. This view will be just like the current registration view, except it will say "Continue" or "Next" instead of "Create Account".
2. Choose 2FA Method - Inform the user that new user registration requires setting up 2FA. Include the link to the help page about 2FA. Have the two buttons: "Add 2FA with authentication application" and "Add 2FA with security device (e.g. USB key)".
3. Set Up 2FA - Show either the TOTP or webauthn configuration page, based on what they picked on the previous step.
4. Display Recovery Codes - Basically the same as the current recovery codes page: Explain the reason for the recovery codes, show the 10 codes, have copy and save buttons.
5. Burn Recovery Code - Same as current page, maybe change button to say "Verify & Create Account".

Then the user will be taken to the manage account page where they'll be more or less forced to verify their email address (nice improvement, Mike).

[...]

PR #14294, opened 2023-08-07
feat: require 2fa for new users

miketheman:
Hooks into security policy to require all non-account management actions
to have a form of 2FA set up.

Adds a time-based restriction for new users, which we can remove when we
want to enforce for everyone.

Resolves https://github.com/pypi/warehouse/issues/13762

Includes refactor adf056ea5656c20432662d351538cd1de6b22f1b

### DIFF
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -184,16 +184,7 @@ class SessionSecurityPolicy:
         raise NotImplementedError
 
     def permits(self, request, context, permission):
-        res = _permits_for_user_policy(self._acl, request, context, permission)
-        # Verify email before you can manage account/projects.
-        if (
-            isinstance(res, Allowed)
-            and not request.identity.has_primary_verified_email
-            and request.matched_route.name.startswith("manage")
-            and request.matched_route.name != "manage.account"
-        ):
-            return WarehouseDenied("unverified", reason="unverified_email")
-        return res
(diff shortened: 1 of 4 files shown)


miketheman:
The current user experience behaves the same for anyone prior to 2023-08-01 regardless of 2FA status.

For anyone past that date, they will still have the top banner prompt (unless they dismiss it) but then any management action they attempt without 2FA enabled will flash a red banner and dump them on the 2FA enrollment page.

warehouse/accounts/security_policy.py
```diff
@@ -309,4 +309,33 @@ def _check_for_mfa(request, context) -> WarehouseDenied | None:
...
+        and request.matched_route.name != "manage.account"
+        and not any(
+            request.matched_route.name.startswith(route) for route in _exempt_routes
+        )
+        and not request.identity.has_two_factor
+    ) and (
+        # Start enforcement from 2023-08-01, but we should remove this check
+        # at the end of 2023.
+        request.identity.date_joined
+        and request.identity.date_joined > datetime.datetime(2023, 8, 1)
```
ewdurbin, line 334: we're ~90 minutes from 2023-08-08 UTC... lets go for that.

ewdurbin (approved):

warehouse/accounts/security_policy.py
miketheman: 098370316916783cf3ada8aa73e1679abadb3c0e

PR #14308, opened 2023-08-08
Improve OIDC error messages

di:
This switches our OIDC logic to raising exceptions with specific error messages when publishers are invalid, rather than returning `None` for multiple failure cases, which resulted in one error message covering a large number of possible failures.

### DIFF
--- a/warehouse/oidc/models/_core.py
+++ b/warehouse/oidc/models/_core.py
@@ -23,6 +23,7 @@ from sqlalchemy.orm import mapped_column
 
 from warehouse import db
 from warehouse.macaroons.models import Macaroon
+from warehouse.oidc.errors import InvalidPublisherError
 from warehouse.oidc.interfaces import SignedClaims
 from warehouse.packaging.models import Project
 
@@ -136,7 +137,7 @@ class OIDCPublisherMixin:
 
             if publisher := query.with_session(session).one_or_none():
                 return publisher
-        return None
+        raise InvalidPublisherError("All lookup strategies exhausted")
 
     @classmethod
     def all_known_claims(cls) -> set[str]:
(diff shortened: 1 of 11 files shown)


warehouse/oidc/models/_core.py
```diff
@@ -189,10 +190,12 @@ def verify_claims(self, signed_claims: SignedClaims):
                         f"JWT for {self.__class__.__name__} is missing claim: "
                         f"{claim_name}"
                     )
-                return False
+                raise InvalidPublisherError(f"Missing claim {claim_name!r}")
 
             if not check(getattr(self, claim_name), signed_claim, signed_claims):
```
woodruffw, line 195: Nonblocker: we could probably push the context even further here by having each `check` also raise rather than return; just a thought in case we need to in the future 🙂

warehouse/oidc/models/_core.py
```diff
@@ -136,7 +137,7 @@ def lookup_by_claims(cls, session, signed_claims: SignedClaims):
 
             if publisher := query.with_session(session).one_or_none():
                 return publisher
-        return None
+        raise InvalidPublisherError("All lookup strategies exhausted")
```
woodruffw, line 140: Nice! This will be especially handy in terms of distinguishing logic errors in lookup vs. claim verification.

woodruffw (approved):
LGTM!

warehouse/oidc/models/_core.py
di: I am reserving that as an option if we need it, but didn't want to have to write a bunch of new claim-specific tests if I didn't have to. 🙂

ewdurbin (approved):

PR #14317, opened 2023-08-09
Add `NullFastlyCache` service, use it in development

di:
This adds a origin cache service that mimics our Fastly service in local development, and outputs the requests it would issue in production.

### DIFF
--- a/tests/unit/cache/origin/test_fastly.py
+++ b/tests/unit/cache/origin/test_fastly.py
@@ -610,3 +610,28 @@ class TestFastlyCache:
 
         assert cacher._purge_key.calls == _purge_key_calls
         assert metrics.increment.calls == []
+
+
+class TestNullFastlyCache:
+    def test_purge_key_prints(self, capsys, metrics):
+        purge_key = pretend.stub(delay=pretend.stub())
+        request = pretend.stub(
+            registry=pretend.stub(
+                settings={
+                    "origin_cache.api_endpoint": "https://api.example.com",
+                    "origin_cache.api_key": "the api key",
+                    "origin_cache.service_id": "the service id",
+                }
+            ),
+            task=lambda f: purge_key,
+        )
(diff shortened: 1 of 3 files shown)


warehouse/cache/origin/fastly.py
```diff
@@ -143,3 +143,23 @@ def purge_key(self, key, metrics=None):
...
+        url = urllib.parse.urljoin(self.api_endpoint, path)
+        headers = {
+            "Accept": "application/json",
+            "Fastly-Key": self.api_key,
+            "Fastly-Soft-Purge": "1",
+        }
+
+        print("Origin cache purge issued:")
+        print(f"* URL: {url!r}")
+        print(f"* Headers: {headers!r}")
```
miketheman, line 165: **question:** ‏should we use print or logging?

miketheman (approved):
Looks good - a couple of non-blockers, so you choose if you want.

tests/unit/cache/origin/test_fastly.py
```diff
@@ -610,3 +610,28 @@ def test_purge_key_no_fallback(
...
+        )
+        cacher = fastly.NullFastlyCache.create_service(None, request)
+        cacher.purge_key("one", metrics=metrics)
+
+        captured = capsys.readouterr()
+        expected = """
+Origin cache purge issued:
+* URL: 'https://api.example.com/service/the service id/purge/one'
+* Headers: {'Accept': 'application/json', 'Fastly-Key': 'the api key', 'Fastly-Soft-Purge': '1'}
+"""  # noqa
```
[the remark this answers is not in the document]
di: The `noqa` here is because line 635 is too long

warehouse/cache/origin/fastly.py
di: Given that we use `print` in other `Null`-services in local dev, I think this is probably fine, not sure if we'd get anything by using logging.

PR #14320, opened 2023-08-10
OIDC: explicitly require that `ref` claim is present

di:
We implicitly require this via the `_job_workflow_ref` claim check, but this PR makes it explicit that this claim is required.

Since we can't predict what the value of this will be, and thus can't verify the value against anything we know, this PR modifies the mechanics a bit to permit specifying required claims that have no check function.

This PR also changes the OIDC token validation flow a bit, to first check for the presence of all required claims, and then check for the validity of all required claims, rather than checking both back to back, per claim.

Tokens missing this claim will error with now error with "Missing claim 'ref'" and we'll get a Sentry notification as well.

### DIFF
--- a/warehouse/oidc/models/_core.py
+++ b/warehouse/oidc/models/_core.py
@@ -177,7 +177,8 @@ class OIDCPublisherMixin:
                     f"{unaccounted_claims}"
                 )
 
-        # Finally, perform the actual claim verification.
+        # Finally, perform the actual claim verification. First, verify that
+        # all requred claims are present.
         for claim_name, check in self.__required_verifiable_claims__.items():
             # All required claims are mandatory. The absence of a missing
             # claim *is* an error with the JWT, since it indicates a breaking
@@ -192,6 +193,9 @@ class OIDCPublisherMixin:
                     )
                 raise InvalidPublisherError(f"Missing claim {claim_name!r}")
 
+        # Now that we've verified all claims are present, verify each claim is correct
+        for claim_name, check in self.__required_verifiable_claims__.items():
+            signed_claim = signed_claims.get(claim_name)
             if not check(getattr(self, claim_name), signed_claim, signed_claims):
                 raise InvalidPublisherError(
                     f"Check failed for required claim {claim_name!r}"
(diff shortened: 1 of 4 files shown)


warehouse/oidc/models/_core.py
```diff
@@ -88,10 +88,10 @@ class OIDCPublisherMixin:
 
     # A map of claim names to "check" functions, each of which
     # has the signature `check(ground-truth, signed-claim, all-signed-claims) -> bool`.
-    __required_verifiable_claims__: dict[str, CheckClaimCallable[Any]] = dict()
+    __required_verifiable_claims__: dict[str, CheckClaimCallable[Any] | None] = dict()
 
     # Simlar to __verificable_claims__, but these claims are optional
-    __optional_verifiable_claims__: dict[str, CheckClaimCallable[Any]] = dict()
+    __optional_verifiable_claims__: dict[str, CheckClaimCallable[Any] | None] = dict()
```
woodruffw, line 95: This is a lot cleaner than my idea :slightly_smiling_face:

warehouse/oidc/models/github.py
```diff
@@ -205,6 +210,10 @@ def job_workflow_ref(self):
     def sub(self):
         return f"repo:{self.repository}"
 
+    @property
+    def ref(self):
+        return unpredictable
```
woodruffw, line 215: Now that the `check` callable is optional, this can be removed, right? If I'm reading the validation logic correctly, we should never access it.

di: Probably, although I'd need to rework some tests that expect all required claims to have properties on the publisher.

woodruffw: Ah yeah, maybe not worth it then.

ewdurbin: I don't understand the sentinel `unpredictable` here, is it just to return an object that is not `None`?

ewdurbin (approved):
one curiosity, but otherwise 👍🏼

warehouse/oidc/models/github.py
woodruffw: Yeah, I think that's mostly to prevent someone from accidentally comparing on it.

di: What @woodruffw said, but I decided that it'd be preferable for this property to just not exist instead: 
6a3ddfa

PR #14335, opened 2023-08-10
OIDC: Allow SHA in `job_workflow_ref` claim.

di:
Resolves https://github.com/pypa/gh-action-pypi-publish/issues/173.

Sometimes we receive an OIDC token where the `ref` claim is an empty string and the `job_workflow_ref` claim takes the form `{owner}/{repo}/.github/workflows/{filename}@{sha}` rather than `{owner}/{repo}/.github/workflows/{filename}@{ref}`. It's currently unclear why/when this happens.

Since the `job_workflow_ref` claim can still be considered valid if it contains the SHA, we can expand our validation of it to check for the form containing the SHA.

This PR also makes the `sha` claim a required unverifiable claim, and removes the Sentry message introduced in #14333.

### DIFF
--- a/warehouse/oidc/models/github.py
+++ b/warehouse/oidc/models/github.py
@@ -113,7 +113,7 @@ class GitHubPublisherMixin:
         "job_workflow_ref": _check_job_workflow_ref,
     }
 
-    __required_unverifiable_claims__: set[str] = {"ref"}
+    __required_unverifiable_claims__: set[str] = {"ref", "sha"}
 
     __optional_verifiable_claims__: dict[str, CheckClaimCallable[Any]] = {
         "environment": _check_environment,
@@ -123,7 +123,6 @@ class GitHubPublisherMixin:
         "actor",
         "actor_id",
         "jti",
-        "sha",
         "run_id",
         "run_number",
         "run_attempt",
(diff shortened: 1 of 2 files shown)


warehouse/oidc/models/github.py
```diff
@@ -39,21 +37,17 @@ def _check_job_workflow_ref(ground_truth, signed_claim, all_signed_claims):
...
+    sha = all_signed_claims.get("sha")
+    if not (ref or sha):
+        raise InvalidPublisherError("The ref and sha claims are empty")
 
-    if not (expected := f"{ground_truth}@{ref}") == signed_claim:
+    expected = {f"{ground_truth}@{_ref}" for _ref in [ref, sha] if _ref}
+    if signed_claim not in expected:
         raise InvalidPublisherError(
-            f"The claim does not match, expecting {expected!r}, got {signed_claim!r}"
+            f"The claim does not match, expecting one of {sorted(expected)!r}, got "
```
woodruffw, line 49: Small nitpick, noticed that "the claim" isn't super clear in the error messages we return:

```suggestion
            f"job_workflow_ref does not match, expecting one of {sorted(expected)!r}, got "
```

di:
Huh, was not expecting to be able to accidentally push to `main` with this...

Issue #13773, opened 2023-05-26
Require 2FA to add a Trusted Publisher


dstufft:
Help ratchet more people onto 2FA by requiring 2FA to add a trusted publisher.

ristomcgehee:
Hey, I'll start on working on this one. I'll do a similar approach as in https://github.com/pypi/warehouse/pull/14280 and hide the form and display a call-out block when 2FA is not enabled.

PR #14360, opened 2023-08-16
Require 2FA to add a Trusted Publisher

ristomcgehee:
If a user does not have 2FA enabled, they will not be able to add a trusted publisher. The form to add will be hidden and a call-out block will be in its place:

**Project publishing page:**

image

**Account publishing page:**

image

Closes https://github.com/pypi/warehouse/issues/13773


warehouse/templates/manage/account/publishing.html
```diff
@@ -225,4 +226,12 @@ <h2>{% trans %}Add a new pending publisher{% endtrans %}</h2>
     {% endfor %}
   </div>
 
+  {% else %}{# user has not enabled 2FA #}
+  <div class="callout-block">
```
di, line 230:
```suggestion
  <div class="callout-block callout-block--warning">
```

warehouse/templates/manage/project/publishing.html
```diff
@@ -145,6 +146,14 @@ <h3>GitHub</h3>
         <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary">
       </div>
     </form>
+
+    {% else %}{# user has not enabled 2FA #}
+    <div class="callout-block">
```
di, line 151:
```suggestion
    <div class="callout-block callout-block--warning">
```

di (approved):

PR #14372, opened 2023-08-16
Make `environment` column non-nullable for OIDC publishers

di:
Fixes WAREHOUSE-PRODUCTION-1KM (https://python-software-foundation.sentry.io/issues/4376572408/).

### DIFF
--- a/warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
+++ b/warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
@@ -65,9 +65,9 @@ def downgrade():
 
     # Data migration
     op.execute(
-        "UPDATE github_oidc_publishers SET environment IS NULL where environment = ''"
+        "UPDATE github_oidc_publishers SET environment = NULL where environment = ''"
     )
     op.execute(
         "UPDATE pending_github_oidc_publishers "
-        "SET environment IS NULL where environment = ''"
+        "SET environment = NULL where environment = ''"
     )
--- a/warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
+++ b/warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
@@ -1,0 +1,73 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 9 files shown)


warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
```diff
@@ -0,0 +1,73 @@
...
+from alembic import op
+
+revision = "a8ebe73ccaf2"
+down_revision = "a2af745511e0"
+
+
+def upgrade():
+    # Data migration
+    op.execute(
+        "UPDATE github_oidc_publishers SET environment = '' where environment = NULL"
```
miketheman, line 31: Use `IS NULL` instead of equality.

> Do not write expression = NULL because NULL is not “equal to” NULL. (The null value represents an unknown value, and it is not known whether two unknown values are equal.)

https://www.postgresql.org/docs/14/functions-comparison.html


```sql
warehouse=> select count(*) from pending_github_oidc_publishers where environment = NULL;
 count
-------
     0

warehouse=> select count(*) from pending_github_oidc_publishers where environment IS NULL;
 count
-------
   873
```

miketheman (changes requested):
Other than the NULL quirks in the data migration, looks good.

ewdurbin (approved):

warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
di: fool me once, shame on me, fool me twice, shame on... shame on nulls.

warehouse/migrations/versions/a8ebe73ccaf2_make_githubpublisher_environment_non_.py
```diff
@@ -0,0 +1,73 @@
...
+    op.alter_column(
+        "github_oidc_publishers",
+        "environment",
+        existing_type=sa.VARCHAR(),
+        nullable=True,
+    )
+
+    # Data migration
+    op.execute(
+        "UPDATE github_oidc_publishers SET environment IS NULL where environment = ''"
```
miketheman, line 68: I don't think setting a value via `IS NULL` works, does it?

miketheman (changes requested):

miketheman (approved):

webknjaz:
Possible regression: https://github.com/pypa/gh-action-pypi-publish/issues/180

Issue #14264, opened 2023-08-02
`release_classifiers` table contains duplicates


miketheman:
Interesting behavior observed when looking at converting the current table to a Declarative ORM style.

We allow for duplicate `release_classifier` entries, since the table is a joiner between `trove_classifiers` and `releases` and has no primary key / unique constraint. Table defined here:
https://github.com/pypi/warehouse/blob/881782a62b96323ce97528c75508056e27d2cea2/warehouse/packaging/models.py#L715-L726

This query returns a list of duplicates:
```sql
SELECT trove_id, release_id, COUNT(*)
FROM release_classifiers
GROUP BY trove_id, release_id
HAVING COUNT(*) > 1 ORDER BY 3 DESC;
```
We have some 4k duplicate entries in a table of 30,175,758 rows - so it's not endemic, but is definitely odd.

One such entry is: https://pypi.org/project/django-supertools/0.1/
Which has 3 records linking to `Programming Language :: Python` (trove_id: 241)

This was released back in 2013, so it seems like there might have been tooling/data constraints then?

I don't yet know how this could be a runtime problem - but if nothing else it's table bloat.

We also cannot convert the table to declarative without either a real primary key or [a mapper candidate key](https://docs.sqlalchemy.org/en/20/faq/ormconfiguration.html#how-do-i-map-a-table-that-has-no-primary-key) - so instead of choosing a candidate I figured let's discuss if we should remove the duplicates and enforce database-level uniqueness across the pairings.

di:
Seems like this is an artifact of "PyPI legacy" as it hasn't happened since 2017:

```
warehouse=> SELECT rc.trove_id, rc.release_id, r.created, COUNT(*) as count
FROM release_classifiers rc
JOIN releases r ON rc.release_id = r.id
GROUP BY rc.trove_id, rc.release_id, r.created
HAVING COUNT(*) > 1
ORDER BY created DESC limit 10;
 trove_id |              release_id              |          created           | count
----------+--------------------------------------+----------------------------+-------
      156 | f533bf17-fc32-4ed8-8e35-884816959b4d | 2017-07-03 12:54:25.017221 |     2
      423 | 4f8e58b9-2e4e-416b-aef4-448db0e8642b | 2017-07-02 05:10:16.030547 |     2
      423 | 1742ac12-0ba1-4abd-8192-d99ad85b4a3f | 2017-07-02 04:59:52.272168 |     2
[...]

PR #14373, opened 2023-08-16
chore: remove duplicates from release_classifiers

miketheman:
Part 1 - data migration to remove the duplicates.

Part 2 will follow by converting the table to Declarative ORM and setting the two columns as primary keys and adding a migration to adjust the table definition.

Refs: #14264

### DIFF
--- a/warehouse/migrations/versions/7f6bed4f4345_remove_duplicates_from_release_.py
+++ b/warehouse/migrations/versions/7f6bed4f4345_remove_duplicates_from_release_.py
@@ -1,0 +1,47 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


warehouse/migrations/versions/7f6bed4f4345_remove_duplicates_from_release_.py
```diff
@@ -0,0 +1,47 @@
...
+    op.execute("SET statement_timeout = 60000")  # 60s
+    op.execute("SET lock_timeout = 60000")  # 60s
+
+    op.execute(
+        """
+        DELETE FROM release_classifiers a USING (
+            SELECT MIN(ctid) as ctid, release_id, trove_id
+            FROM release_classifiers
+            GROUP BY release_id, trove_id HAVING COUNT(*) > 1
+            LIMIT 4453 -- 4453 is the number of duplicates in production
```
di, line 36: Is this true for TestPyPI as well?

di (approved):

PR #14384, opened 2023-08-17
Check for disabled/frozen users when authorizing via macaroons

ewdurbin:
I noticed some frozen malware uploaders were still able to get things shipped via API tokens (Macaroons) minted before being frozen.

Tracked it down to not checking is_frozen via is_disabled in the MacroonSecurityPolicy.

I'm not _that_ happy we have to duplicate this check across files... so I'm not that committed to this as a fix.

### DIFF
--- a/tests/unit/macaroons/test_security_policy.py
+++ b/tests/unit/macaroons/test_security_policy.py
@@ -131,7 +131,7 @@ class TestMacaroonSecurityPolicy:
         assert add_vary_cb.calls == [pretend.call("Authorization")]
         assert request.add_response_callback.calls == [pretend.call(vary_cb)]
 
-    def test_identity_frozen_user(self, monkeypatch):
+    def test_identity_disabled_user(self, monkeypatch):
         policy = security_policy.MacaroonSecurityPolicy()
 
         vary_cb = pretend.stub()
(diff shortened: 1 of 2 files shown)


ewdurbin:
Note, I specifically don't bother with this when going via OIDC since 1) Trusted Publishers are bound to a single Project 2) We destroy Projects when taking them down for malware 3) Frozen Users cannot create a new Trusted Publisher.

tests/unit/macaroons/test_security_policy.py
```diff
@@ -130,6 +131,53 @@ def test_identity_no_db_macaroon(self, monkeypatch):
         assert add_vary_cb.calls == [pretend.call("Authorization")]
         assert request.add_response_callback.calls == [pretend.call(vary_cb)]
 
+    def test_identity_frozen_user(self, monkeypatch):
```
miketheman, line 134: **nitpick:** ‏we're evaluating `is_disabled` instead of `is_frozen` - which _should_ bubble up correctly in `is_disabled`. but I'd be remiss if I didn't point that out.

miketheman (approved):

tests/unit/macaroons/test_security_policy.py
ewdurbin: https://github.com/pypi/warehouse/pull/14384/commits/1643d6d1f1d3f4a8d5c52a61f9d5456c27ddc6eb

We have coverage for frozen -> disabled elsewhere so I just updated the test name.

PR #14388, opened 2023-08-17
feat: nudge users to 2fa after email verification

miketheman:
Prior to this change, after a user has registered, and verified their email address, they end up on their own account management page. If they try to take any non-account actions, they get a red bar telling them to set up 2FA.

Since we will already show users a blue notice that they _can_ set up 2FA, why not send them there immediately after confirming their email to complete the task.

The template change is to prevent adding the blue bar at the top of the session, since they are already on the right page. We can deduce that from the `Referrer` header.

Follows #14294

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -874,6 +874,11 @@ def verify_email(request):
         ),
         queue="success",
     )
+
+    # If they've already set up a 2FA app, send them to their account page.
+    if request.user.has_two_factor:
+        return HTTPSeeOther(request.route_path("manage.account"))
+    # Otherwise, send them to the two-factor setup page.
     return HTTPSeeOther(request.route_path("manage.account.two-factor"))
 
 
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -874,7 +874,7 @@ def verify_email(request):
         ),
         queue="success",
     )
-    return HTTPSeeOther(request.route_path("manage.account"))
+    return HTTPSeeOther(request.route_path("manage.account.two-factor"))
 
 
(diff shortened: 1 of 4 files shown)


miketheman:
Post-email verification, the user should see something very similar to this:

Screenshot 2023-08-17 at 17 28 12

ewdurbin (approved):

miketheman:
I'll add a conditional for the yellow bar as well - since the URL it provides redirects to the exact same page.

di (approved):
I think this a good cheap & easy fix, but long term when this becomes a requirement we should think about reworking the registration flow to have a very clear set of steps (signup -> email verification -> 2FA setup) prior to "success", such that a user that is at a point in this process will always return to that stage when visiting `pypi.org` until they complete it.

warehouse/accounts/views.py
```diff
@@ -874,7 +874,7 @@ def _error(message):
         ),
         queue="success",
     )
-    return HTTPSeeOther(request.route_path("manage.account"))
+    return HTTPSeeOther(request.route_path("manage.account.two-factor"))
```
ewdurbin, line 877: Hmmm, should this be wrapped in a conditional? If I already have 2FA and add/verify a new address I’ll still be thrown to the 2FA page.

miketheman: Nice catch, will handle that behavior.

miketheman:
> 

I'm in full agreement - but I'd defer a bit to a better-thought-out UX/UI flow before undertaking that.

warehouse/accounts/views.py
miketheman: Added!

PR #14394, opened 2023-08-18
drop support for SNS SignatureVersion 1

ewdurbin:
Now that [both testpypi and pypi queues are configured with version 2](https://github.com/pypi/infra/pull/158), we should drop support to ensure we aren't handling old SHA1 hashes.


miketheman (approved):
Looks great, nice and clean.

ewdurbin:
Pre validated this a bit by deploying this branch to test.pypi.org and ensuring that SNS notifications were working. Pretty confident it is good to go.

PR #14415, opened 2023-08-22
More OIDC metrics

di:
This adds some additional OIDC metrics: total # of files published via OIDC, and number of publishers per publisher type.

This also adds some basic stats on the total number of projects, releases and files so we can calculate percentages easily.

### DIFF
--- a/warehouse/oidc/tasks.py
+++ b/warehouse/oidc/tasks.py
@@ -71,8 +71,9 @@ def compute_oidc_metrics(request):
     for t in request.db.query(OIDCPublisher.discriminator).distinct().all():
         discriminator = t[0]
         metrics.gauge(
-            f"warehouse.oidc.{discriminator}.publishers",
+            "warehouse.oidc.publishers",
             request.db.query(OIDCPublisher)
             .where(OIDCPublisher.discriminator == discriminator)
             .count(),
+            tag=discriminator,
         )
--- a/warehouse/oidc/tasks.py
+++ b/warehouse/oidc/tasks.py
@@ -12,6 +12,7 @@
 
 from warehouse import tasks
 from warehouse.metrics import IMetricsService
+from warehouse.oidc.models import OIDCPublisher
 from warehouse.packaging.models import File, Project, Release
 
 
@@ -57,3 +58,21 @@ def compute_oidc_metrics(request):
         "warehouse.oidc.total_critical_projects_published_with_oidc_publishers",
         projects_published_with_oidc.where(Project.pypi_mandates_2fa.is_(True)).count(),
     )
+
+    # Metric for total number of files published via OIDC
+    metrics.gauge(
+        "warehouse.oidc.total_files_published_with_oidc_publishers",
+        request.db.query(File.Event)
+        .where(File.Event.additional.op("->>")("publisher_url").is_not(None))
+        .count(),
+    )
+
+    # Number of publishers for specific publishers
+    for t in request.db.query(OIDCPublisher.discriminator).distinct().all():
+        discriminator = t[0]
+        metrics.gauge(
+            f"warehouse.oidc.{discriminator}.publishers",
+            request.db.query(OIDCPublisher)
+            .where(OIDCPublisher.discriminator == discriminator)
+            .count(),
+        )
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -21,7 +21,6 @@ import pip_api
 
 from google.cloud.bigquery import LoadJobConfig
 from packaging.utils import canonicalize_name
-from sqlalchemy import func
 from sqlalchemy.orm import joinedload
 
 from warehouse import tasks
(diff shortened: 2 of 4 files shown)


warehouse/oidc/tasks.py
```diff
@@ -57,3 +58,21 @@ def compute_oidc_metrics(request):
...
+        request.db.query(File.Event)
+        .where(File.Event.additional.op("->>")("publisher_url").is_not(None))
+        .count(),
+    )
+
+    # Number of publishers for specific publishers
+    for t in request.db.query(OIDCPublisher.discriminator).distinct().all():
+        discriminator = t[0]
+        metrics.gauge(
+            f"warehouse.oidc.{discriminator}.publishers",
```
miketheman, line 74: Question: is the discriminator better provided as a tag instead of the name, so we can sum by all, and split by discriminator?

warehouse/packaging/tasks.py
```diff
@@ -56,6 +56,21 @@ def sync_file_to_cache(request, file_id):
         file.cached = True
 
 
+@tasks.task(ignore_result=True, acks_late=True)
+def compute_packaging_metrics(request):
+    metrics = request.find_service(IMetricsService, context=None)
+
+    metrics.gauge(
+        "warehouse.packaging.total_projects", request.db.query(Project).count()
```
miketheman, line 64: The .count() operator isn't as efficient as using the SQL func.count() - since this query selects all fields and then runs a count on them.
See this example for a more efficient approach:
https://github.com/pypi/warehouse/blob/6be6bccf07dace18784ea8aeac7906903fdbcf3a/warehouse/accounts/tasks.py#L33

warehouse/oidc/tasks.py
```diff
@@ -57,3 +58,21 @@ def compute_oidc_metrics(request):
         "warehouse.oidc.total_critical_projects_published_with_oidc_publishers",
         projects_published_with_oidc.where(Project.pypi_mandates_2fa.is_(True)).count(),
     )
+
+    # Metric for total number of files published via OIDC
+    metrics.gauge(
+        "warehouse.oidc.total_files_published_with_oidc_publishers",
+        request.db.query(File.Event)
+        .where(File.Event.additional.op("->>")("publisher_url").is_not(None))
```
miketheman, line 66: I can't shake the feeling that there's a better way to perform this query, but I don't have it at hand. I'll continue to muse on it.

miketheman (changes requested):
Requesting changes for the count() calls. Other comments are non-blocking.

warehouse/packaging/tasks.py
dstufft: Even better is to use the `RowCount` table :) 

https://github.com/pypi/warehouse/blob/main/warehouse/views.py#L255

warehouse/oidc/tasks.py
dstufft: I guess the question is whether we want the total number of files published via OIDC, or whether we want the total number of files published via oidc in some time frame.

If it's the latter, we can emit metrics for each publish event and let Datadog do the rollup for us, if it's the former... well that's much harder to do without a gauge like this.

miketheman: I think this is what I was looking for:

```suggestion
        .where(File.Event.additional["publisher_url"].is_not(None))
```

The underlying query generates the same syntax, and now we don't have to call the `.op("->>")` semantic, rather we treat the jsonb column like a dict.

If we wanted to use the return value, then we'd pass `.additional["foo"].astext...` but since we're only comparing for non-nulls, and want a count, then we don't need the cast.

miketheman: I completely recognize that we have prior art in the codebase that uses the `op.("->>")` semantic, and we should refactor those to use this, but that's not 100% necessary for this changeset.

miketheman (approved):

warehouse/packaging/tasks.py
```diff
@@ -56,6 +57,35 @@ def sync_file_to_cache(request, file_id):
         file.cached = True
 
 
+@tasks.task(ignore_result=True, acks_late=True)
+def compute_packaging_metrics(request):
```
miketheman, line 61: **question:** ‏what schedules this task?

warehouse/oidc/tasks.py
di: I tried that, it doesn't work:

```
>>> db.query(File.Event).where(File.Event.additional.op("->>")("publisher_url").is_not(None)).count()
25618

>>> db.query(File.Event).where(File.Event.additional["publisher_url"].is_not(None)).count()
1174355

>>> db.query(File.Event).count()
1174355
```

miketheman: 🤔  I guess this is where `.astext` would be the right change, too.

warehouse/packaging/tasks.py
di: https://github.com/pypi/warehouse/pull/14417

PR #14432, opened 2023-08-25
Compute OIDC metrics hourly

di:
Rather than daily.


warehouse/oidc/__init__.py
```diff
@@ -50,4 +50,4 @@ def includeme(config):
     config.add_route("oidc.github.mint_token", "/_/oidc/github/mint-token", domain=auth)
 
     # Compute OIDC metrics periodically
-    config.add_periodic_task(crontab(minute=0, hour=3), compute_oidc_metrics)
+    config.add_periodic_task(crontab(hour="*"), compute_oidc_metrics)
```
miketheman, line 53: I think we need to preserve minute=0 otherwise it'll run every minute.

miketheman (changes requested):

warehouse/oidc/__init__.py
di, line 53:
```suggestion
    config.add_periodic_task(crontab(minute=0, hour="*"), compute_oidc_metrics)
```

miketheman (approved):

PR #14442, opened 2023-08-28
docs: add troubleshooting for repo rename

miketheman:


docs/user/trusted-publishers/troubleshooting.md
```diff
@@ -54,6 +54,11 @@ endpoint:
   represented in the actual OIDC token. Check for typos! If you're using
   GitHub Actions, check if the workflow is using the same environment
   as configured when the publisher was configured on PyPI.
+* `invalid-publisher` for a previously-working project: this usually indicates
+  that something has changed on either side. One example we've seen in when a
+  GitHub repository is renamed, and the configration on PyPI continues to use
+  the old repository name. Check that the `repository` value is the same on
+  both sides.
```
di, line 61:
```suggestion
* `invalid-publisher` for a previously-working project: this usually indicates
  a typo or that something has changed on either side. One example we've seen
  is when a source repository is renamed, and the configration on PyPI
  continues to use the old repository name. For GitHub, check that the
  `repository_owner`, `repository` and workflow filename values are the same on
  both sides.
```

di (approved):

PR #14443, opened 2023-08-28
feat: emit metrics on email schedule

miketheman:
As a way for us to gain more visibility into which email templates are being sent and when, add a metrics emitter at the last place we know about the template name.

### DIFF
--- a/docs/dev/development/email.rst
+++ b/docs/dev/development/email.rst
@@ -40,6 +40,8 @@ Calling a function with the ``_email`` decorator does the following:
   response
 - A security log is added to the user's account
 - The email is sent using Amazon SES (on production environment)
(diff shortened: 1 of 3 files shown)


ewdurbin (approved):
approved, knowing it will pass ci/coverage, but do we want at least one test?

miketheman:
> approved, knowing it will pass ci/coverage, but do we want at least one test?

I couldn't find a decorator-specific test, but I'll add an assertion or two to confirm the metrics behavior.

miketheman:
@ewdurbin see test assertions added in [a879eff](https://github.com/pypi/warehouse/pull/14443/commits/a879effd0e1f88cd1b11f27e8d06153203cd2c45)

PR #14449, opened 2023-08-29
Update order of operations in password checking

ewdurbin:
Updates order of operations when checking user passwords and frozen/disabled status to avoid leaking account status pre-auth.

Also along the way realized we don't invalidate user sessions once they've been frozen/disabled, since we block login we should probably also block them from doing anything.

### DIFF
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -164,6 +164,13 @@ class SessionSecurityPolicy:
         if user is None:
             return None
 
+        # User may have been frozen or disabled since the session was created.
+        is_disabled, _ = login_service.is_disabled(userid)
+        if is_disabled:
+            request.session.invalidate()
+            request.session.flash("Session invalidated", queue="error")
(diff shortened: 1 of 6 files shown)


warehouse/accounts/security_policy.py
```diff
@@ -59,28 +59,31 @@ def _basic_auth_check(username, password, request):
     if userid is not None:
         user = login_service.get_user(userid)
         is_disabled, disabled_for = login_service.is_disabled(user.id)
```
di, line 61: I think this line can move in as well?

ewdurbin: https://github.com/pypi/warehouse/pull/14449/commits/aaa641314a0edb7ae8a7b1c3d6436b5286c8753d

di (approved):

miketheman (approved):
I've confirmed that the tests fail if the code isn't changed, yay!
The logical nesting and reordering makes sense.

PR #14475, opened 2023-09-01
chore: revert celery and kombu upgrade

miketheman:
We're seeing some slowdowns and startup time alerts with the latest
versions.

Also Revert "chore: increase worker startup timeout (#14474)"

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -5,7 +5,7 @@ b2sdk
 Babel
 bcrypt
 boto3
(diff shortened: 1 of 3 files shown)


ewdurbin (commented):
5.3.3 was deleted after the initial yank.

ewdurbin (approved):

miketheman:
Looks related: https://github.com/celery/kombu/issues/1783

PR #14476, opened 2023-09-01
doom transactions for re-uploaded files

ewdurbin:
This stops us from issuing all the purges related to the file/release in question when we no-op.


miketheman (approved):
image

ewdurbin:
> but can it run doom?

PR #14477, opened 2023-09-01
chore: enable tcp keepalive for celery/kombu/sqs

miketheman:
We perform a number of calls to SQS as part of normal operations - sometimes on the order of 10 calls during an upload request.

Using TCP Keep-Alives is one way to speed up subsequent connections, and the default is `False`.
Refs: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#:~:text=respective%20regional%20endpoint.-,tcp_keepalive,-Toggles%20the%20TCP

Passing this config to Celery, which in turn sends it to Kombu, which in turn sends it to boto3, which sends to botocore, which sends to urllib3, enables it.

That's quite a few turtles :turtle:

### DIFF
--- a/tests/unit/test_tasks.py
+++ b/tests/unit/test_tasks.py
@@ -454,49 +454,85 @@ def test_make_celery_app():
             "amqp://guest@rabbitmq:5672//",
             {},
         ),
-        (Environment.development, False, "sqs://", "sqs://", {}),
-        (Environment.production, True, "sqs://", "sqs://", {}),
+        (
+            Environment.development,
+            False,
+            "sqs://",
+            "sqs://",
+            {
+                "client-config": {"tcp_keepalive": True},
+            },
+        ),
+        (
+            Environment.production,
+            True,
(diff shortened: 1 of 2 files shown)


miketheman:
I guess it comes down to how the sqs client is instantiated - is it in the context of the server startup, or per-request?

If it's per-request, then keep alive might not help much. But if it's per server-startup (web worker) - and I think it is - then we would benefit from having the keepalive around for subsequent calls to the endpoint.

I'm game to give it a shot - we have the necessary observability via Datadog Traces to see if there's an impact.

ewdurbin:
@dstufft might be able to confirm, but my understanding is that since we construct the celery app in the "include" at https://github.com/pypi/warehouse/blob/1cca80952114b6a50ceb5ca0b7657d0f7c2777d3/warehouse/tasks.py#L205-L207, that is _per startup_. so yeah! let's roll.

ewdurbin (approved):

Issue #13768, opened 2023-05-26
Require 2FA to accept an invite to join an organization


dstufft:
Ratchet more people onto 2FA by requiring 2FA to accept an invite to join an organization.

PR #14485, opened 2023-09-04
Require 2FA to accept an org invite

ristomcgehee:
This PR will require a user to have 2FA enabled in order to accept an invitation to join an organization.

Here's how it looks when they click on the link to join:
image

Closes https://github.com/pypi/warehouse/issues/13768


ristomcgehee:
The failed check appears to be just an intermittent build issue.

warehouse/templates/accounts/organization-invite-confirmation.html
```diff
@@ -17,8 +17,16 @@
     {% trans %}Confirm Organization Invite{% endtrans %}
 {% endblock %}
 
-{% block main %}
+{% block prompt %}
+  {% if request.user.has_two_factor %}
+    {{ super() }}
+  {% else %}
+    <!-- Content hidden -->
+  {% endif %}
```
miketheman, line 25: **praise:** ‏I hadn't seen this pattern used, nice way to hide the confirmation.

miketheman (approved):

miketheman:
Thanks @ristomcgehee !

di:
I'm not sure this is sufficient, a user could still manually POST to the endpoint that corresponds to this form and bypass the restriction.

Instead, the pattern I think we should generally follow is: still show the form, but make the "submit" button disabled and have a secondary button/direction to perform the blocking action, and also add a similar check on the view to prevent manual posting from succeeding.

ristomcgehee:
I haven't been adding server-side checks for these "require 2fa" PRs because I don't see this as a security control but rather as a nudge to enable 2fa. If you like, I can add server-side checks to these places. 

What do you mean by this part:

> have a secondary button/direction to perform the blocking action

PR #14491, opened 2023-09-05
chore: update base debian image to bookworm

miketheman:
Debian 12 released in June. Updates any of our base-level deps, since Python is already selected via Docker image tag.

Refs: https://www.debian.org/News/2023/20230610


ewdurbin (approved):

ewdurbin:
image

PR #14497, opened 2023-09-06
refactor: update url parsing

miketheman:
- Make urllib3 a direct dependency
  We're already installing it as part of requests, boto, and more.
  Make it explicit as we'll be importing it directly now for parsing.
- Add a local flake8 plugin to detect usage of `urllib.parse.urlparse`
- Refactor usages

### DIFF
--- a/tests/unit/utils/test_http.py
+++ b/tests/unit/utils/test_http.py
@@ -43,6 +43,7 @@ class TestIsSafeUrl:
             "\njavascript:alert(x)",
             "\x08//example.com",
             "\n",
+            "view/?param=//example.com",
         ],
     )
     def test_rejects_bad_url(self, url):
@@ -54,7 +55,6 @@ class TestIsSafeUrl:
             "/view/?param=http://example.com",
             "/view/?param=https://example.com",
             "/view?param=ftp://exampel.com",
-            "view/?param=//example.com",
             "https://testserver/",
             "HTTPS://testserver/",
             "//testserver/",
--- a/dev/flake8/checkers.py
+++ b/dev/flake8/checkers.py
@@ -1,0 +1,60 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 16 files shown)


ewdurbin (approved):
the linter rule was more straight forward than I would have imagined, nice!

tests/unit/utils/test_http.py
```diff
@@ -43,6 +43,7 @@ class TestIsSafeUrl:
             "\njavascript:alert(x)",
             "\x08//example.com",
             "\n",
+            "view/?param=//example.com",
```
di, line 46: I guess there's no harm in considering _more_ URLs to be unsafe, but the change here is somewhat curious.

miketheman: Indeed it is - it's a difference between stdlib and urllib3 handling the same string.

```pycon
>>> from urllib.parse import urlparse
>>> from urllib3.util import parse_url
>>> url = "view/?param=//example.com"
>>> std_lib = urlparse(url)
>>> new_hotness = parse_url(url)
>>> repr(std_lib)
ParseResult(scheme='', netloc='', path='view/', params='', query='param=//example.com', fragment='')
>>> repr(new_hotness)
Url(scheme=None, auth=None, host='view', port=None, path='/', query='param=//example.com', fragment=None)
```

We can see that the parser now places the `host` (aka `netloc`) as `view`. Arguably the correct intended behavior?

The new parser trips up the final evaluation here:

https://github.com/pypi/warehouse/blob/72fbc6f9ce90db0a332616062aaabf2508e6d190/warehouse/utils/http.py#L50

Since the test suite is explicitly passing a host value of "testserver", the conditions no longer match.

@di do you think this is cause for concern?

cc @sethmlarson - is this parsing difference expected, and correct?

di: I defer to @sethmlarson 🙂

sethmlarson: This is "expected" by urllib3 standards because we "handle" scheme-less URLs (which I dislike very much, and [we've deprecated in the latest version](https://github.com/urllib3/urllib3/pull/2950)). In a future version of urllib3 that URL will parse identically (and correctly according to RFC 3986) to urlparse.

For now the best way to protect against this is to assert a truthy scheme and host are parsed anywhere you're expecting an absolute URL.

miketheman: Thanks @sethmlarson ! It'll be some time before we're on the 2.x series and will handle the deprecation warning when we have it (which might mean moving the test example back to the original "allowed").
For now, I think we're good.

PR #14499, opened 2023-09-06
chore: surface event if proxy token mismatches

miketheman:
If someone passes the proxy token and it doesn't match, we default to using the other block that pulls values off other headers.

Instead of allowing that, send a sentry ~and dump the request~.

### DIFF
--- a/warehouse/utils/wsgi.py
+++ b/warehouse/utils/wsgi.py
@@ -16,6 +16,8 @@ import hmac
 
 from typing import TYPE_CHECKING
 
+import sentry_sdk
+
 from sqlalchemy import type_coerce
 from sqlalchemy.dialects.postgresql import INET
 from sqlalchemy.exc import NoResultFound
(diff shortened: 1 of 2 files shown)


ewdurbin (approved):

warehouse/utils/wsgi.py
```diff
@@ -66,11 +68,24 @@ def __call__(self, environ, start_response):
             }
 
             host = environ.get("HTTP_WAREHOUSE_HOST", "")
+
         # If we're not getting headers from a trusted third party via the
         # specialized Warehouse-* headers, then we'll fall back to looking at
         # X-Forwarded-* headers, assuming that whatever we have in front of us
         # will strip invalid ones.
         else:
+            # If there IS a token, but it doesn't match, then tell us about it.
```
dstufft, line 77: You probably don't need the compare here, there's no possible way for the compare digest to return `True`.

ewdurbin:
That was me validating ^, sorry.

PR #14505, opened 2023-09-07
feat: require 2fa for new user to upload as well

miketheman:
Expand the policy to include file upload actions.

Follows #14294
Refs #13762


miketheman:
The response message isn't perfect, since it was geared towards management actions, but a twine upload will show:

```
ERROR    HTTPError: 403 Forbidden from http://localhost/legacy/
         You must enable two factor authentication to manage other settings
```

ewdurbin:
See https://github.com/miketheman/warehouse/pull/2048/files for a proposed refactor to improve the error message for uploads.

ewdurbin (approved):

miketheman:
wonder-twins-powers

PR #14515, opened 2023-09-08
refactor: extract package type to enum

miketheman:
An update to alembic's autogenerate detection started to barf on the unnamed enum.

The ALTER TYPE migration is to make the custom-built type conform to the alembic autogenerated name, since it's used in a custom index (which really ought to be a Check Constraint, but that's a problem for another time).


ewdurbin (approved):
✨

PR #14516, opened 2023-09-08
refactor: update more text to string

miketheman:
Since Postgres treats `text` and `varchar` (with no limit) the same under the hood, we don't need to make a distinction unless there's a limit.

Note: There's still `Text` usage in `warehouse.packaging.models` which has yet to be fully converted. Once it is, expect to see a similar migration.

Refs: https://www.postgresql.org/docs/current/datatype-character.html


miketheman:
No idea what's going on with RTD builds.

ewdurbin (approved):

PR #14518, opened 2023-09-08
fix: call for the enum's value in xmlrpc

miketheman:
`xmlrpc.client.dumps()` cannot marshall a `str, Enum` to the string value like `json.dumps()` can.

Refs: #14515 #14517


ewdurbin (approved):
so tired of xmlrpc

PR #14550, opened 2023-09-13
wip feat: add ratelimit view deriver

miketheman:
To support a more generic use case of rate limiting views by IP.

### DIFF
--- a/warehouse/rate_limiting/__init__.py
+++ b/warehouse/rate_limiting/__init__.py
@@ -21,6 +21,7 @@ from first import first
 from limits import parse_many
 from limits.storage import storage_from_string
 from limits.strategies import MovingWindowRateLimiter
+from pyramid.httpexceptions import HTTPTooManyRequests
 from zope.interface import implementer
 
 from warehouse.metrics import IMetricsService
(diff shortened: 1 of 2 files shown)


warehouse/rate_limiting/__init__.py
```diff
@@ -162,7 +163,82 @@ def __eq__(self, other):
...
+
+        def wrapper_view(context, request):
+            limit_attr = getattr(request, limit_by)
+
+            ratelimiter = request.find_service(
+                IRateLimiter, name="rate_limiting.client"
+            )
+            # TODO: Is this Kosher?
+            # Override the default limits with the ones specified in the view
+            ratelimiter._limits = parse_many(limit_str)
```
miketheman, line 194: I don't know how I feel about this. I was trying to find a way to pass `limit_str` as a way to modify the Service's constructor call, but I can't figure out how to make it work with a pre-registered service factory other than doing it like this.

warehouse/rate_limiting/__init__.py
```diff
@@ -162,7 +163,82 @@ def __eq__(self, other):
...
+
+            metrics = request.find_service(IMetricsService, context=None)
+
+            request_route = request.matched_route.name
+
+            ratelimiter.hit(limit_attr)
```
miketheman, line 200: What's pretty odd is that this works but doesn't appear to add entries to redis as far as I can tell. The `ratelimiter._storage` is a `RedisStorage`, and I can get rate limited just fine, but I never see a key inserted into redis.

warehouse/rate_limiting/__init__.py
```diff
@@ -162,7 +163,82 @@ def __eq__(self, other):
...
+ratelimit_view_deriver.options = {"ratelimit", "ratelimit_by"}  # type: ignore[attr-defined] # noqa: E501
+
+
 def includeme(config):
     config.registry["ratelimiter.storage"] = storage_from_string(
         config.registry.settings["ratelimit.url"]
     )
+
+    config.register_service_factory(
+        RateLimit(limit="100 per second"),
```
miketheman, line 239: I also couldn't figure out a way to not pass a `limit` arg here, so I picked something. Since any users of the view deriver have to override the string anyhow, is that a problem?

miketheman:
Going to close this out. It's a good experiment, and I learned a few things, but since many of our ratelimiting semantics are often more specific, this broader use case isn't the right one for now.

PR #14563, opened 2023-09-14
chore: code quality cleanups

miketheman:
See each commit for specifics.


warehouse/accounts/views.py
```diff
@@ -793,7 +794,18 @@ def _error(message):
...
+            "user",
+            {
+                "username": user.username,
+                "last_login": user.last_login,
+                "token_last_login": last_login,
+            },
+        )
+        sentry_sdk.capture_message(
+            "Password reset token used after user logged in",
+            level="warning",
```
di, line 807: Should this include a fingerprint or is the context sufficient to report different messages per user (assuming we want that).

miketheman: What kind of fingerprint are you thinking of?

di: Like this: https://github.com/pypi/warehouse/blob/e7e9d4ccadc852d33b7c475eaa2f749086227a09/warehouse/oidc/models/_core.py#L178-L182

di: Maybe unique on the username?

miketheman: Oh, I see! I'll update the message to include username in it, and that ought to trigger an issue per username. No need to much with custom scope or fingerprinting for that.

PR #14568, opened 2023-09-15
Refactor usage of IFileStorage to deal in file-like-objects only

ewdurbin:


warehouse/packaging/services.py
```diff
@@ -229,14 +244,19 @@ def get_checksum(self, path):
...
 
         path = self._get_path(path)
 
-        self.bucket.upload_file(file_path, path, ExtraArgs=extra_args)
+        _initial_pos = fileobj.tell()
+        try:
+            fileobj.seek(0)
+            self.bucket.upload_fileobj(fileobj, path, ExtraArgs=extra_args)
+        finally:
+            fileobj.seek(_initial_pos)
```
ewdurbin, line 259: 🙄 boto3 helpfully closes the file object as a side effect of `upload_fileobj`

warehouse/packaging/services.py
```diff
@@ -179,13 +183,24 @@ def get_checksum(self, path):
...
-        )
+        _initial_pos = fileobj.tell()
+
+        def _stream_opener():
+            fileobj.seek(0)
+            return fileobj
+
+        try:
+            self.bucket.upload(
+                upload_source=b2sdk.transfer.outbound.UploadSourceStream(
```
ewdurbin, line 196: Validated this works locally.

firefly9966 (approved):

PR #14553, opened 2023-09-13
chore: add datadog tracing to tasks

miketheman:
Previously attempted in #12981 and reverted in #13003

ddtrace shipped a fix recently after in 1.9.0
Refs: https://github.com/DataDog/dd-trace-py/pull/4969

This time we also add a distinct name for the worker class (thanks, Karl) as Cabotage doesn't provide a facility yet to set environment variables per-Procfile process.

PR #14572, opened 2023-09-16
Revert "chore: add datadog tracing to tasks (#14553)"

ewdurbin:
This reverts commit a8a83bba2397a47a1c326dd437910be0259fe886.

Workers have been stalling out like the last time we tried to enable this.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 release: bin/release
 web: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-uploads.conf.py warehouse.wsgi:application
-worker: env DD_SERVICE=warehouse-worker bin/start-worker ddtrace-run celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
-worker-beat: env DD_SERVICE=warehouse-worker-beat bin/start-worker ddtrace-run celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
+worker: bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
+worker-beat: bin/start-worker celery -A warehouse beat -S redbeat.RedBeatScheduler -l info


ewdurbin:
data seems to back up the prognosis.

since the point in time that dd-trace was implemented, we slowly lost workers from our worker pool over time until a single worker was holding the whole house up. when it froze we stopped processing tasks.

dd-trace-cant-melt-kombu-sheets

after revert we haven't seen a single stalled worker.

Procfile
```diff
@@ -1,5 +1,5 @@
 release: bin/release
 web: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-uploads.conf.py warehouse.wsgi:application
-worker: env DD_SERVICE=warehouse-worker bin/start-worker ddtrace-run celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
-worker-beat: env DD_SERVICE=warehouse-worker-beat bin/start-worker ddtrace-run celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
+worker: bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
+worker-beat: bin/start-worker celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
```
P403n1x87, line 5: Would it be cool to try this again with
```
worker: env DD_SERVICE=warehouse-worker DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE=1 bin/start-worker ddtrace-run celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
worker-beat: env DD_SERVICE=warehouse-worker-beat DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE=1 bin/start-worker ddtrace-run celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
```
by any chance? This should hopefully create more "isolation" between ddtrace and the celery application 🤞 .

P403n1x87: More details here: https://ddtrace.readthedocs.io/en/stable/configuration.html#DD_UNLOAD_MODULES_FROM_SITECUSTOMIZE

ewdurbin: We've shipped this in #14583 and will keep an eye on things over the next few hours.

ewdurbin: Same behavior. But we hadn't yet merged https://github.com/pypi/warehouse/pull/14585 so we were running 1.18.5, I noticed the env var you asked us to set in the release notes under known issues, does 1.18.5 respect this flag?

P403n1x87: Ah shame 🙁 . That env var was introduced around 1.8 so it's definitely respected in 1.18. Any logs that might give us a hint as to why the workers are timing out would be much appreciated (e.g. do they hang on SSL requests?)

ewdurbin: Unfortunately the workers just stop silently. We have the latest dd-trace out and running right now. I will try to make a close inspection if/when a worker stalls.

ewdurbin: Sorry. No more additional info was available in logs and we’re down to one worker now. A revert has been set on course for deployment, and there’s no obvious way for us to glean more information from running containers (run as non root user and we can’t add new utilities for inspection after deployment)

PR #14583, opened 2023-09-18
try this option for dd-trace on celery

ewdurbin:
Ref: https://github.com/pypi/warehouse/pull/14572/files#r1328371500


miketheman (approved):
Does this work without needing to set `env` before the second option?

PR #14587, opened 2023-09-18
chore: convert packaging models to PEP 484 Declarative

miketheman:
A behavior uncovered here that hasn't been dealt with before in this codebase.
If the column in question is used in an existing trigger, we cannot modify it. A possible change to support that would be to:

- drop the trigger
- perform the alter table
- recreate the trigger

I elected to not pursue this course right now, preferring to get as many model fields in declarative format as possible.


ewdurbin (approved):
I guess its unsurprising how many packing related models we have 😵‍💫.

I still don't quite understand the whole sa.String -> sa.TEXT churn, but it seemed fine the last round.

miketheman:
> I still don't quite understand the whole sa.String -> sa.TEXT churn, but it seemed fine the last round.

The underlying issue is that some database engines treat `VARCHAR` differently from `TEXT` (MySQL, Maria, et al) - where VARCHAR can limit the length to max 65k chars, and TEXT is limited to <something much larger>.

With Postgres they behave the same - the only time VARCHAR differs from TEXT is when there's a limit applied (e.g. `VARCHAR(40)` ). Otherwise they are the same under the hood.

However, when using SQLAlchemy `String()` or Declarative `Mapped[str]`, the engine sets the column type to VARCHAR, vs `Text()`, which does ... TEXT.

Since the declarative syntax of `foo: Mapped[str]` is more succinct than `foo: Mapped[str] = mapped_column(Text)` and we ultimately get the same outcome from the database engine, I elected to take advantage of the terser syntax, and update the column definitions to match, hence the migrations.

Does that clear it up?

PR #14594, opened 2023-09-19
helpscout app: load all users for + style emails

ewdurbin:
When a user with email address `user+pypi@example.com` emails us from `user@example.com` the sidebar finds no users.

This will make it load all users that match the base email.


warehouse/admin/templates/admin/helpscout/app.html
```diff
@@ -11,6 +11,7 @@
  # See the License for the specific language governing permissions and
  # limitations under the License.
 -#}
+{%- for user in users -%}
```
miketheman, line 14: **thought (non-blocking):** ‏I recall some setting we could use to auto-set the `{%-` tag strip behaviors for all templates, we might wanna look that up at some point.

warehouse/admin/views/helpscout.py
```diff
@@ -45,16 +46,22 @@ def helpscout(request):
     email = (
         request.db.query(Email)
         .where(
-            Email.email.ilike(request.json_body.get("customer", {}).get("email", ""))
+            Email.email.ilike(
+                re.sub(
+                    r"\+[^)]*@",
+                    "@",
+                    request.json_body.get("customer", {}).get("email", ""),
+                )
```
miketheman, line 54: **question:** ‏I know this is only used in admin views, on already-sanitized email address strings, but is there any other weird injection risk here?

ewdurbin: hm, I would hope that sqlalchemy would guard us against SQL injection...

I suppose if someone sent an email from an address that would case re.sub to freak out, and helpscout didn't barf, and we loaded the page... it would make it here.

warehouse/admin/templates/admin/helpscout/app.html
ewdurbin: just did this to make the existing test pass 😂, its HTML in the end so I'm not sure how important setting such a default makes.

miketheman (approved):

Issue #13979, opened 2023-06-21
Admin UI for viewing/revoking API tokens


di:
**What's the problem this feature will solve?**
Provide an Admin UI for viewing/revoking API tokens.

**Describe the solution you'd like**
Given an API token, the admin UI can deserialize/inspect the contents of the token, link to a `Macaroon` in the database (if it is present) as well as correlated events, and provide a UI for revoking the token (including an email to the token owner).

PR #14597, opened 2023-09-19
refactor: extract helper functions from service

miketheman:
Extract two static functions from `DatabaseMacaroonService`.

De-serializing a token involves no database access, so we can use these functions directly without instantiating the class.

I intend to use this function as part of #13979 

Best reviewed with ignore whitespace.


miketheman:
Tagging @woodruffw os the original author

ewdurbin (approved):

PR #14616, opened 2023-09-21
feat: allow bearer type for macaroons

miketheman:
Since Macaroons are often considered bearer tokens, allow them to be used in the authorization header scheme.

### DIFF
--- a/warehouse/macaroons/security_policy.py
+++ b/warehouse/macaroons/security_policy.py
@@ -66,7 +66,7 @@ def _extract_http_macaroon(request):
 
     if auth_method.lower() == "basic":
         return _extract_basic_macaroon(auth)
-    elif auth_method.lower() == "token":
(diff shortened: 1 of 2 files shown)


miketheman:
https://explore.org/livecams/brown-bears/brown-bear-salmon-cam-brooks-falls

ewdurbin:
🐻

warehouse/macaroons/security_policy.py
```diff
@@ -66,7 +66,7 @@ def _extract_http_macaroon(request):
 
     if auth_method.lower() == "basic":
```
ewdurbin, line 67: just an idea. as I'm unsure how often we're using anything other than basic...

should we emit a lil metric here? just like `warehouse.macaroon.auth` with tag `auth_method:{auth_method.lower()` maybe done better as an entire PR instrumenting the session auth as well.

ewdurbin (approved):

Issue #8462, opened 2020-08-23
Only show GitHub stars, if it is the same project


MartinThoma:
**What's the problem this feature will solve?**

I've recently written an article about pytest plugins. I manually checked a lot of them and had to judge quickly which ones are worth a shot. So the Github stars displayed on PyPI help a bit by filtering. There was one plugin I only mentioned because it had many stars: https://pypi.org/project/pytest-xdist-debug-for-graingert/ 

Having a second look at it, I noticed that it simply linked to a different project.

This is misleading at the very least, but can support typosquatting at worst.

**Describe the solution you'd like**

There are multiple solutions:

1. Don't show github stars at all.
2. Only show them, if the linked page has a back-link to the package.
3. Check if other packages link to the same repository already

I would prefer (2). We could also provide a badge like https://badge.fury.io/for/py

miketheman:
Now that Trusted Publishing exists, we might consider this for only repos that have published with GitHub Actions, since then we have something to compare against.

miketheman:
Also, today I learned of https://capec.mitre.org/data/definitions/693.html

miketheman:
Before we try to tackle this, we need to understand the flow of how trusted publishing treats multiple packages from a single github repo (a la monorepo)

MartinThoma:
I'd rather completely remove the feature than having such easy starjacking. The only value this gives to the user is to know if the package is trustworthy. As the current mechanism of retrieving the number of stars is so easy to fake, it gives a false sense of trustworthiness.

MartinThoma:
@yeraydiazdiaz I think you missed the point. The point is that the Github stars belong to another project. It doesn't matter that it is a copy/fork. The copy might introduce bugs or even be malicious.

As it is, the "Github stars" feature of PyPI warehouse makes the live of attackers easier and provides no value to end users.

di:
Fixed by #16532.

PR #14657, opened 2023-09-29
feat: use trusted publishing to adapt UI

miketheman:
Projects using Trusted Publishing from GitHub have a strong attestation that the repo is the one they came from.

Using the metadata we have from the upload events, we can determine how the file got there, and bubble that up to the Release to control some aspects of the UI.
We could also display something in the Download Files if we wanted to.

Wrapping the sidebar statistics with the HTML `<details>` accomplishes preserving the sidebar metadata for projects, but making it more prominent for those using a Trusted Publishing workflow.

Refs: [StarJacking](https://capec.mitre.org/data/definitions/693.html)
Addresses #8462 (partially) and probably replaces #10917

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -603,6 +603,17 @@ class Release(db.Model):
             ]
         )
 
+    @property
+    def trusted_published(self) -> bool:
+        """
+        A Release can be considered published via a trusted publisher if
+        **all** the Files in the release are published via a trusted publisher.
+        """
+        files = self.files.all()  # type: ignore[attr-defined]
+        if not files:
+            return False
+        return all(file.uploaded_via_trusted_publisher for file in files)
+
 
 class PackageType(str, enum.Enum):
     bdist_dmg = "bdist_dmg"
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -670,6 +670,16 @@ class File(HasEvents, db.Model):
         comment="If True, the object has been archived to our archival bucket.",
     )
 
+    @property
+    def uploaded_via_trusted_publisher(self) -> bool:
+        """Return True if the file was uploaded via a trusted publisher."""
+        return (
+            self.events.where(
+                self.Event.additional.has_key("publisher_url")  # type: ignore[attr-defined] # noqa E501
+            ).count()
+            > 0
+        )
+
     @hybrid_property
     def metadata_path(self):
         return self.path + ".metadata"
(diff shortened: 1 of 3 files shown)


miketheman:
Looks a little something like this:

https://github.com/pypi/warehouse/assets/529516/d617ee03-4fc1-4917-8913-79ab4d131a66

warehouse/packaging/models.py
```diff
@@ -603,6 +603,17 @@ def has_meta(self):
             ]
         )
 
+    @property
+    def trusted_published(self) -> bool:
+        """
+        A Release can be considered published via a trusted publisher if
+        **all** the Files in the release are published via a trusted publisher.
+        """
+        files = self.files.all()  # type: ignore[attr-defined]
```
miketheman, line 612: Since `self.files` returns an `AppenderQuery`, it returns True until we resolve the query by fetching the rows. If there's a better way to do this, I'm all ears!

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -69,6 +70,7 @@ <h3 class="sidebar-section__title">{% trans %}Statistics{% endtrans %}</h3>
       </li>
     </ul>
   </div>
+  {% if not release.trusted_published %}</details>{% endif %}
```
di, line 73: I think this is probably not worth doing this way, for a few reasons, and we should Do It Right™ instead.

As-is, we're not actually checking that the URL used for getting repo statistics is the same as the URL used for trusted publishing, so this can be subverted easily by using trusted publishing for something that is attempting to do "starjacking".

Additionally, it also doesn't indicate to the user or maintainer why sometimes statistics are hidden, and why sometimes they are not, which will make a more confusing UI.

I think we should remove this from the PR for now, with the plan to eventually replace it with a new statistics sidebar that does the following:

* if trusted publishing from GitHub has been used for the release:
  * include the URL of the upstream repository (https://github.com/pypi/warehouse/issues/10917)
  * fetch statistics for the repository that trusted publishing is configured for
  * mark the statistics as "verified via Trusted Publishing", with a link
* if trusted publishing was not used, and a GitHub URL was provided in the metadata (e.g., we would currently show statistics):
  * don't fetch the statistics
  * instead, include a message + provide a link that indicates that trusted publishing should be used to enable GitHub statistics

miketheman: @di Thanks for the feedback!

I've gone ahead and removed this change from scope so we can ship the other enhancements in the meantime.

I'll move the rest of the notes to a new issue so we can discuss the stats sidebar.

Issue #13767, opened 2023-05-26
Require 2FA to be accept an invitation to join a project


dstufft:
Ratchet more users onto 2FA by requiring 2FA to accept an invitation to join a project.

PR #14659, opened 2023-09-30
Require 2FA to join project

ristomcgehee:
This PR will require a user to have 2FA enabled in order to accept an invitation to join a project.

Here's how it looks when they click on the link to join:
image

I'm not adding a server-side check because I don't see this as a security control but rather as a nudge to enable 2FA. My expectation is that when we globally require 2FA, we will have 1 central location on the server side where we'll enforce this.

Closes https://github.com/pypi/warehouse/issues/13767


miketheman (approved):
Thanks! We're continuing to nudge.

PR #14691, opened 2023-10-03
views: don't splat request.POST

### LINKED ISSUE #14690: 2FA with USB key always fails with "This field is required."
**Describe the bug**

I registered a USB Titan security key to my pypi account. It seemed to go smoothly. When I attempt to use it to log in, I get the following:

image

I have tried this with another Titan key with the same result.

**Expected behavior**

I expect to log in via USB titan key.

**To Reproduce**

I cannot determine if there are steps to reproduce or if it is isolated to my account.

**My Platform**

Browser is Chrome `Version 117.0.5938.132 (Official Build) (x86_64)`

**Additional context**

Recovery codes work, and I did get an authenticator app to work as well, FWIW. I've removed and re-added the security key(s) a few times just to be sure.

woodruffw:
This needs to be a `MultiDict`; by turning it into kwargs, we fail the `InputRequired` validator because it doesn't treat the values as actual user-supplied inputs.

Fixes #14690.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -422,7 +422,7 @@ def webauthn_authentication_validate(request):
 
     user_service = request.find_service(IUserService, context=None)
     form = WebAuthnAuthenticationForm(
-        **request.POST,
+        request.POST,
         request=request,
         user_id=userid,
         user_service=user_service,


miketheman (approved):

woodruffw:
NB: This only fails on login and not token registration because `validate_webauthn_provision` does this:

```
        form = ProvisionWebAuthnForm(
            MultiDict(**self.request.POST),
            user_service=self.user_service,
            user_id=self.request.user.id,
            challenge=self.request.session.get_webauthn_challenge(),
            rp_id=self.request.domain,
            origin=self.request.host_url,
        )
```

However that `MultiDict(**...)` pattern isn't strictly necessary, and should probably just be refactored to `self.request.POST` wherever it appears.

miketheman:
Thanks for the speedy fix @woodruffw !

Issue #14700, opened 2023-10-05
Unsupported platform tag macosx_14_0_arm64


bsrdjan:
**What's the problem this feature will solve?**
Twine upload of binary wheel built with Python 12 on ARM MacOS 14 is rejected because of unsupported platform tag 

```shell
Binary wheel 'pyrfc-3.3-cp312-cp312-macosx_14_0_arm64.whl' has an unsupported platform tag            
         'macosx_14_0_arm64'. 
```

**Describe the solution you'd like**
The platform tag should be supported I suppose?

PR #14702, opened 2023-10-06
fix: Unsupported platform tag macosx_14_0_arm64 (#14700) 

bsrdjan:
fix: Unsupported platform tag macosx_14_0_arm64 (#14700)


miketheman:
Thanks! Looks like we need the test updated as well.
Would you mind updating this line to major version 15?

https://github.com/pypi/warehouse/blob/main/tests/unit/forklift/test_legacy.py#L3143

miketheman (approved):

PR #14709, opened 2023-10-06
chore: strip out whitespace added by jinja blocks

miketheman:
When looking at source of HTML, we emit whitespace around Jinja blocks.

While individual blocks can opt out of emitting this whitespace with `-` control characters (like `robots.txt` did), it is simpler to apply this as a global setting.

On pages like `/help/`, these settings remove some 500 lines of whitespace from the output.

Refs: https://jinja.palletsprojects.com/en/3.0.x/templates/#whitespace-control


ewdurbin (approved):

di:
FYI, looks like this is causing some missing spaces in places where we previously expected newlines to get converted to whitespace:

image


This example is from https://github.com/pypi/warehouse/blob/b2a306fbc6e1087d484f422863b31d569cf09d8b/warehouse/templates/packaging/detail.html#L382C1-L383

Issue #14710, opened 2023-10-06
Unable to provision TOTP


patel999jay:
> Hey -- i seem to have the same issue... I've tested with both authy, google authenticator + Microsoft authenticator
> 
> ![Screenshot from 2023-10-06 13-45-59](https://user-images.githubusercontent.com/15642823/273075546-5f18b782-81b5-4235-aa55-544e403377b3.png)
> 
> This is my user account https://pypi.org/user/avi_vajpeyi/
> 
> Any suggestions?
> 
> (I tried the 'Time sync' option for the google authenticator app, i got a message stating "... in sync with Google servers")

i have the same issue, could not do it with authy, google authenticator and/or Microsoft authenticator.

_Originally posted by @patel999jay in https://github.com/pypi/warehouse/issues/7385#issuecomment-1751389413_

miketheman:
@patel999jay I've double-checked the behavior for TOTP enrollment, and it still works as expected.
Are you certain you're not adding any other characters to the input?
Input is also time-sensitive.

patel999jay:
@miketheman : I am certain I am not adding any other character to the input and also input the code in timely fashion. 

**To Reproduce**

- Create New Account
- Try To Enable 2FA
- Every time getting this : "_Invalid TOTP code. Try again?_"

miketheman:
@patel999jay I am unable to reproduce the issue. I've also just confirmed the behavior is working on both pypi.org and test.pypi.org .

And you have confirmed that your app has been time-synced?

patel999jay:
@miketheman : yes it is sync and i tried using both android and apple device, not working on any of them.

![image](https://github.com/pypi/warehouse/assets/5512610/7451b4b2-92dc-43cd-af93-121c5e7623dd)

avivajpeyi:
Hey -- I'm having the same issue, tested with authy, Microsoft authenticator and the Google authenticator apps. :(

di:
For folks having issues adding TOTP, can you visit https://time.is/ on your **mobile** device (the same one with your authenticator app) and confirm that it says "Your time is exact!"?

patel999jay:
Take a look at this:

![image](https://github.com/pypi/warehouse/assets/5512610/05309bba-54df-49f5-8523-07ff3c0909af)

di:
Thanks @patel999jay. Can you confirm that in Google Authenticator, you're using the code from an entry that says `PyPI: <username>`?

patel999jay:
[...]

PR #14720, opened 2023-10-09
Emit unique error when TOTP is skewed

di:
This PR adds a unique error message when a TOTP is valid, but outside the time range of what we consider valid. This should help identify issues to add TOTP devices due to clock skew. 

Related to #14710.

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -21,6 +21,7 @@ import markupsafe
 import wtforms
 import wtforms.fields
 
+import warehouse.utils.otp as otp
 import warehouse.utils.webauthn as webauthn
 
 from warehouse import forms, recaptcha
@@ -38,7 +39,6 @@ from warehouse.email import (
 )
 from warehouse.events.tags import EventTag
 from warehouse.i18n import localize as _
-from warehouse.utils.otp import TOTP_LENGTH
 
 # Taken from passlib
 MAX_PASSWORD_SIZE = 4096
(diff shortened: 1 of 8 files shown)


ewdurbin (approved):

ewdurbin (commented):
the only concern is if this would aid in brute force attacks. i forget if we rate limit TOTP specifically.

this effectively gives 20 + 3 “checks” per submission. probably not a concern but wanted to raise.

warehouse/accounts/forms.py
```diff
@@ -415,7 +415,19 @@ class TOTPAuthenticationForm(TOTPValueMixin, _TwoFactorAuthenticationForm):
...
+            user = self.user_service.get_user(self.user_id)
+            user.record_event(
+                tag=EventTag.Account.LoginFailure,
+                request=self.request,
+                additional={"reason": "out_of_sync_totp"},
+            )
+            raise wtforms.validators.ValidationError(
+                _("Invalid TOTP code. Your device time may be out of sync.")
+            )
+        except otp.InvalidTOTPError:
```
miketheman, line 430: I like this, but is this form used during OTP provisioning? I think it's only for authentication post-provisioning.

di:
@ewdurbin Fair point, I think it'd be OK to just limit this to views where the user is adding a new TOTP device, and not when they're logging in (since we haven't had any reports to that effect). I'll adjust accordingly.

miketheman (approved):

warehouse/accounts/forms.py
di: aa40787

Issue #8090, opened 2020-06-10
Fix how Warehouse stores metadata (per-file, not per-release)


brainwane:
Task list:

- [ ] Add a `FileMetadata` model and an optional 1-1 relationship to the `File` model. The fields of this model should correspond to the available [Core Metadata](https://packaging.python.org/en/latest/specifications/core-metadata/) fields.
- [ ] Add a `metadata_source` field to `FileMetadata` that lets us determine if the metadata is "provided" (i.e. from the `POST`) or "extracted" (i.e. from the artifact itself). 
- [ ] Modify the upload endpoint to start creating `FileMetadata` objects on new uploads, discerning how we got the metadata (wheels should always be "extracted", source distributions will be "provided")
- [ ] Write a task to backfill `FileMetadata` files for wheels that have already been uploaded with data from our metadata backfill files (#15526 will be helpful here)
- [ ] Create a way to select the "metadata source" for a release (the oldest file for that release)
- [ ] Drop all the various metadata fields on the `Release` model and update the UI to get them from the "metadata source" instead.

---
Original issue

**Describe the bug**
Warehouse's API gives the user sometimes inaccurate dependency metadata for a project's release, because:

* we only store the dependency information per Release, but it can vary per file.
* first file uploaded creates the Release object. which is also problematic -- if the user uploads a source distribution first, no dependency information is encoded for the Release

**Expected behavior**
As I understand it, we should change how we store and provide dependency information, recording and storing it per file instead of per release. I presume this means that the `requires_dist` field within [the release endpoint](https://warehouse.readthedocs.io/api-reference/json/#release) would move from the "info" value to the individual "releases" values.

**To Reproduce**
Sorry, I don't have one to hand.

**Additional context**
Quoting [a conversation in IRC today](http://kafka.dcpython.org/day/pypa-dev/2020-06-10) between @dstufft and @techalchemy (quotes are condensed and edited for grammar):

@dstufft said of the current [Warehouse JSON API](https://warehouse.readthedocs.io/api-reference/json/), "I don't think it's usable in it's current form for resolving dependencies". Regarding the metadata available, which clients would otherwise need to download packages to acquire, 
"the data is wrong is the main thing ... for dep info .... because warehouse (and originally PyPI's data model) is wrong. We only store the dependency information per Release, but it can vary per file."

@techalchemy asked: "so which file do you pick for parsing dependencies? the first wheel that gets uploaded? Or the last one?"
[...]

PR #14834, opened 2023-10-31
feat: add Observers and Observations models

miketheman:
Here's a draft of Observers, Observations.

I've applied the mixin to Project, Release models so far, and only added Admin visibility into Projects.
There's a bunch of `TODO`s in there, and any pointers for concrete solutions would be much appreciated!

Take it for a whirl, try it out, let me know what you think.


warehouse/admin/views/projects.py
```diff
@@ -136,9 +144,108 @@ def project_detail(project, request):
...
+    if not request.user.observer:
+        request.user.observer = Observer()
+
+    project.record_observation(
+        request=request,
+        kind=kind,
+        observer=request.user.observer,
+        summary=summary,
+        payload=payload,
+    )
```
di, line 239: Wondering if we're going down the same path as https://github.com/pypi/warehouse/issues/8090 here by storing these on the project (or release) and not on individual files. I see how it would be tedious to report N observations for N files on a release, or M releases in a project, but this sort of prevents us from being able to make observations about individual artifacts.

miketheman: Excellent observation (pun totally intended).

The overall intent is that the mixin be flexible enough to record Observations at any level - Project, Release, File, User, Organization, etc.

I've only added the mixin so far to the Project and Releases, but only added the Admin UI for Projects.

Any mixed-in model can then call `<model>.record_observation(...)` to add a record.

miketheman: @di I've added an example of adding observations to Releases via Admin as well to hopefully make it clearer.

di (approved):
LGTM!

PR #14860, opened 2023-11-03
Collection of nits

miketheman:
Annoyances collected for the past few weeks, none of which justified their own PR at the time.

Removes ~100 warnings from test suite.

See commits for the different pieces.

### DIFF
--- a/warehouse/utils/__init__.py
+++ b/warehouse/utils/__init__.py
@@ -14,7 +14,7 @@ import datetime
 
 
 def now() -> datetime.datetime:
-    return datetime.datetime.now(datetime.timezone.utc)
+    return datetime.datetime.now(datetime.UTC)
 
 
 def dotted_navigator(path):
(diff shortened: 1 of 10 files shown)


ewdurbin (approved):
only one I was sketched out by was the parse_raw -> model_validate_json, but a deprecation makes sense.

warehouse/utils/__init__.py
```diff
@@ -13,8 +13,8 @@
 import datetime
 
 
-def now():
-    return datetime.datetime.utcnow()
+def now() -> datetime.datetime:
+    return datetime.datetime.now(datetime.UTC)
```
di, line 17: @miketheman I think this might be causing https://python-software-foundation.sentry.io/issues/4601709612/

miketheman: @di indeed, looks like it. 3.12 deprecates `utcnow()`, which is what prompted this.
Fix incoming!

PR #14871, opened 2023-11-06
fix: return a naive UTC object

miketheman:
When used in Jinja filters, we compare against other naive time-like objects.

Introduced in #14860
Fixes https://python-software-foundation.sentry.io/share/issue/6f07733076404820b4e87c3fad7cb896/

Add docstring and test for explicit behavior.

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -22,6 +22,15 @@ import pytest
 from urllib3.util import parse_url
 
 from warehouse import filters
+from warehouse.utils import now
+
+
+def test_now():
+    assert isinstance(now(), datetime.datetime)
+    assert now().tzinfo is None
+    with pytest.raises(TypeError) as excinfo:
+        _ = now() < datetime.datetime.now(datetime.UTC)
+    assert "can't compare offset-naive and offset-aware datetimes" in str(excinfo.value)
 
 
 def test_camo_url():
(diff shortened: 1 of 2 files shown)


tests/unit/test_filters.py
```diff
@@ -22,6 +22,15 @@
...
 from warehouse import filters
+from warehouse.utils import now
+
+
+def test_now():
+    assert isinstance(now(), datetime.datetime)
+    assert now().tzinfo is None
+    with pytest.raises(TypeError) as excinfo:
+        _ = now() < datetime.datetime.now(datetime.UTC)
+    assert "can't compare offset-naive and offset-aware datetimes" in str(excinfo.value)
```
di, line 33: Don't we want to test that `now()` _can_ be compared instead?

miketheman: Added on next line. That way we get both assertions.

warehouse/utils/__init__.py
```diff
@@ -14,7 +14,8 @@
 
 
 def now() -> datetime.datetime:
-    return datetime.datetime.now(datetime.UTC)
+    """Return the current datetime in UTC without a timezone."""
+    return datetime.datetime.now(datetime.UTC).replace(tzinfo=None)
```
di, line 18: This is now just the same as `datetime.datetime.now()`, unless I'm missing something.

miketheman: Does this help explain the difference?
```python
>>> import datetime
>>> print(datetime.datetime.now())
2023-11-06 17:09:32.924917
>>> print(datetime.datetime.now(datetime.UTC).replace(tzinfo=None))
2023-11-06 22:09:45.453366
```

miketheman: (and yes, [time is hard](https://zachholman.com/talk/utc-is-enough-for-everyone-right))

di (approved):
Time is hard!

PR #14905, opened 2023-11-13
Fix German pronounciation explanation for "PyPI"

cmarqu:


ewdurbin:
Thank you for this @cmarqu! Our translations are not updated directly like this. Please see https://warehouse.pypa.io/translations.html for how to submit changes!

PR #14906, opened 2023-11-13
improve stop time for docker compose services

ewdurbin:
signal handling is mixed bag, for most of the Python processes setting SIGINT is good enough

for localstack and static, just use SIGKILL since its all that seems to work and that is effectively what we have been doing forever... just without waiting 10s


di (approved):
💯

PR #14923, opened 2023-11-16
Dismissable banners

### LINKED ISSUE #10806: Sticky banners that cannot be dismissed are annoying
Every page on https://pypi.org/ has this sticky banner and it cannot be dismissed:

image

di:
This PR makes banners created via the admin UI optionally dismissable.

Fixes #10806.

### DIFF
--- a/warehouse/banners/models.py
+++ b/warehouse/banners/models.py
@@ -15,6 +15,7 @@ from sqlalchemy.orm import Mapped, mapped_column
 
 from warehouse import db
 from warehouse.utils.attrs import make_repr
+from warehouse.utils.db.types import bool_false
 
 
 class Banner(db.Model):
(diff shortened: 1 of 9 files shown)


warehouse/banners/models.py
```diff
@@ -31,6 +31,7 @@ class Banner(db.Model):
     link_url: Mapped[str]
     link_label: Mapped[str] = mapped_column(default=DEFAULT_BTN_LABEL)
     fa_icon: Mapped[str] = mapped_column(default=DEFAULT_FA_ICON)
+    dismissable: Mapped[bool] = mapped_column(default=True)
```
miketheman, line 34: question: we have `bool_true` as a type in `warehouse.utils.db.types` that uses `server_default` which would set the value server-side if we used an `insert` directly, vs `default` here that sets the value in Python-land. Was this choice intentional?
If changed, need to re-generate the migration so that the value is passed to the database.

miketheman (approved):
Looks good - one question inline. Good to go if that's intended.

warehouse/banners/models.py
di: Interestingly I actually meant to set the default to `false` here as I was going back and forth on what the default should be, so nice catch! I'm OK with setting this via `server_default`!

miketheman (approved):

PR #14949, opened 2023-11-21
refactor: replace identity lookups

miketheman:
When determining whether a request is from a given user, we should check the `request.user` (where applicable) / `request.identity` instead of the `authenticated_userid` string.


miketheman:
Starts to address the [TODO here](https://github.com/pypi/warehouse/blob/636dc2d54bc0f825bc21edcf2f8692de8d661be2/warehouse/utils/security_policy.py#L79-L84) by replacing any direct usages of `MultiSecurityPolicy.authenticated_userid`. The TODO and other pieces may be removed later.

tests/unit/accounts/test_views.py
```diff
@@ -162,7 +162,7 @@ def test_returns_users_with_prefix(self, db_session, user_service):
         ]
 
         request = pretend.stub(
-            authenticated_userid=1,
+            identity=1,
```
woodruffw, line 165: Tiny thing, but you might want to make this `identity=pretend.stub()` here and below to emphasize that `identity` is expected to be some kind of object model, not just an identifier 🙂

warehouse/accounts/views.py
```diff
@@ -178,7 +178,7 @@ def accounts_search(request) -> dict[str, list[User]]:
     Used with autocomplete.
     User must be logged in.
     """
-    if request.authenticated_userid is None:
+    if request.identity is None:
```
woodruffw, line 181: Nitpick: maybe use `request.user` here and below? The AuthN/Z layers should prevent `request.identity` from being anything except a `User` in this context, but accessing through `user` is slightly more self-describing.

woodruffw (commented):
LGTM overall, two small nitpicks!

warehouse/accounts/views.py
miketheman: I like that, and will give it a shot. However, if I'm reading the intent correctly, this kind of change **should not** be applied to the Macaroon policy, since that isn't always a User, correct?

woodruffw: Correct!

miketheman: I was able to convert over most usages to `request.user` except for some of the totp/webauthn because of how the tests are currently constructed.

PR #14989, opened 2023-11-29
Deprecate 'changelog' XML-RPC endpoint

di:
This endpoint is being deprecated due to low usage and high resource consumption.

### DIFF
--- a/docs/dev/api-reference/feeds.rst
+++ b/docs/dev/api-reference/feeds.rst
@@ -48,33 +48,19 @@ then.
 
 Example usage::
 
+  >>> import time
   >>> import xmlrpc.client
-  >>> import arrow
-  >>> client = xmlrpc.client.ServerProxy('https://test.pypi.org/pypi')
-  >>> latefeb = arrow.get('2018-02-20 10:00:00')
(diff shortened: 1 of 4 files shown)


miketheman (approved):

ewdurbin (approved):

tylfin:
Hey there, I noticed failures from this API today. I know it's deprecated, but the warning on the Warehouse API says:

> Users of this API are strongly encouraged to subscribe to the [pypi-announce](https://mail.python.org/mailman3/lists/pypi-announce.python.org/) mailing list for notices as we begin the process of removing XML-RPC from PyPI.

Could future changes go through the mailing list to notify the community ahead of time?

tylfin:
@di Looking into this more, I don't think `changelog_since_serial` is an equivalent API for consuming mirrors. 

For example, if we need to find all changes since X date, we could get the (limited) 50,000 changes, and make another request starting at the most recent date.

Now, as far as I understand, using the serial will always limit you to the 50k most recent changes, and it's impossible to get earlier changes? I think `changelog_since_serial` needs proper pagination before this can be removed

PR #15003, opened 2023-12-02
 #6192 - Document the package size limit

sourabhninawe:


di:
Hi, thanks for the PR. We already have a PR open for this that will include the actual limit into the documentation, so we won't need to update the docs if/when it changes: #14478

di:
Closing in favor of https://github.com/pypi/warehouse/pull/14478.

PR #15004, opened 2023-12-02
forklift/legacy: fallback on all project users for GPG email

### LINKED ISSUE #14928: Sending "GPG signature uploaded" email fails when using Trusted Publishing
Currently we return a 500-level failure if a user attempts to upload GPG signatures via Trusted Publishing, because we assume there is a `request.user`:

https://github.com/pypi/warehouse/blob/d5ee54d0deb193c01edf3de3243d8d2027ea0501/warehouse/forklift/legacy.py#L1144-L1146

However this might be `None` if Trusted Publishing is used, in which case the email should probably go to all owners instead.

Reported in #14927, this is https://python-software-foundation.sentry.io/share/issue/4680dcbcc9d348109307d6dc0d03f900/.

woodruffw:
If a distribution is uploaded via Trusted Publishing and is accompanied by a PGP signature, send the GPG deprecation email to all project members, rather than exploding due to `request.user` not being present.

Fixes #14928.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1142,7 +1142,9 @@ def file_upload(request):
                 "provided signature has been discarded."
             )
             send_gpg_signature_uploaded_email(
-                request, request.user, project_name=project.name
+                request,
+                request.user if request.user else project.users,
+                project_name=project.name,
             )
 
         # TODO: This should be handled by some sort of database trigger or


warehouse/forklift/legacy.py
```diff
@@ -1142,7 +1142,9 @@ def file_upload(request):
                 "provided signature has been discarded."
             )
             send_gpg_signature_uploaded_email(
-                request, request.user, project_name=project.name
+                request,
+                request.user if request.user else project.users,
```
woodruffw, line 1146: NB: Passing in either a scalar `user` or `list[users]` here is allowed by `@email`-decorated APIs, like `send_gpg_signature_uploaded_email`.

woodruffw:
Done!

woodruffw:
> Oops, one more time please, [#14949](https://github.com/pypi/warehouse/pull/14949) just got merged (not sure why I'm missing the ability to do this on your fork).

This is a GitHub bug 🙃 -- the fork is on an org account (`woodruffw-forks`), and GitHub doesn't let me share "upstream" push rights to branches on org account forks.

(I can manually make you a maintainer on my fork however, one sec.)

PR #15020, opened 2023-12-06
Always remove pending publishers when project is created

### LINKED ISSUE #14389: `invalid-pending-publisher` when using OIDC to upload 2nd version of package
**Describe the bug**
Not sure if this is a bug, but at least it was unexpected, and the [OIDC Troubleshooting guide](https://docs.pypi.org/trusted-publishers/troubleshooting/) wasn't very helpful.

In [this CI run](https://github.com/NickCrews/feco3/actions/runs/5899323313), I got `Token request failed: the server refused the request for the following reasons: invalid-pending-publisher: valid token, but project already exists`

So, I went ahead and downloaded the .whl files generated from that CI run, and manually uploaded them from my laptop using twine. That worked. I haven't tried again using OIDC to see if it will work again now that I did a fresh upload.

**Expected behavior**
Upload to PyPI should work: The project exists, and I have the OIDC configured correctly.

**To Reproduce**
You can see the workflow file used at https://github.com/NickCrews/feco3/actions/runs/5899323313/workflow.

Screenshot of OIDC config:
image

Here is the complete security audit log for my project (screenshot taken Aug 17, 2023)
image

Was this something to do with enabling 2FA?

Also, this doesn't show up in the security log, but I thought I remember back in June adding GitHub as a "pending trusted publisher", and then a very similar thing happening, and I think that is why I uploaded v0.3.0 of the project using Twine back in June. Perhaps this was bugging out because of some stale state where PyPI thought that my OIDC registration was pending, when now I don't even see an option int the PyPI webUI to to add pending publishers?

di:
~Draft PR while I try to figure out why this is causing organization tests to fail.~

Fixes #14389.

### DIFF
--- a/tests/unit/manage/views/test_organizations.py
+++ b/tests/unit/manage/views/test_organizations.py
@@ -1592,9 +1592,8 @@ class TestManageOrganizationProjects:
 
         # The project was created
         project = (
-            db_request.db.query(Project).filter_by(name="fakepackage").one_or_none()
+            db_request.db.query(Project).filter_by(name="fakepackage").one()
         )
-        assert project is not None
 
         # Refresh the project in the DB session to ensure it is not stale
         db_request.db.refresh(project)
--- a/tests/unit/manage/views/test_organizations.py
+++ b/tests/unit/manage/views/test_organizations.py
@@ -1590,11 +1590,16 @@ class TestManageOrganizationProjects:
         view = org_views.ManageOrganizationProjectsViews(organization, db_request)
         result = view.add_organization_project()
 
-        # The project was created, and belongs to the organization.
+        # The project was created
         project = (
             db_request.db.query(Project).filter_by(name="fakepackage").one_or_none()
         )
         assert project is not None
+
+        # Refresh the project in the DB session to ensure it is not stale
+        db_request.db.refresh(project)
+
+        # The project belongs to the organization.
         assert project.organization == organization
 
         assert isinstance(result, HTTPSeeOther)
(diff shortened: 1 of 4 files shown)


tests/unit/manage/views/test_organizations.py
```diff
@@ -1590,11 +1590,16 @@ def test_add_organization_project_new_project(
         view = org_views.ManageOrganizationProjectsViews(organization, db_request)
         result = view.add_organization_project()
 
-        # The project was created, and belongs to the organization.
+        # The project was created
         project = (
             db_request.db.query(Project).filter_by(name="fakepackage").one_or_none()
         )
         assert project is not None
```
miketheman, line 1597: If we now expect that the project can never be None, we can remove one assert and update this assignment.

```suggestion
        project = (
            db_request.db.query(Project).filter_by(name="fakepackage").one()
        )
```

tests/unit/manage/views/test_organizations.py
```diff
@@ -1590,11 +1590,16 @@ def test_add_organization_project_new_project(
...
+
+        # Refresh the project in the DB session to ensure it is not stale
+        db_request.db.refresh(project)
```
miketheman, line 1600: What I don't fully understand is how this is possible, if the line above this is the first time we're querying for this object from the DB.

Is it possible that these lines are somehow causing this?
https://github.com/pypi/warehouse/blob/6e3fb21a454f26de9205817ea976beb7cdf1e51d/warehouse/organizations/services.py#L520-L523

miketheman (commented):
Generally supporting this change!

I had a couple of questions inline - hopefully easy enough tio answer.

tests/unit/manage/views/test_organizations.py
di: These lines have no effect on the test here, and neither does additional flushes elsewhere. I don't have an answer for why this is happening in this test 🤷

PR #15025, opened 2023-12-07
blog: replace external plugin with built-in plugin

### LINKED ISSUE #14112: Migrate to the mkdocs-material blog plugin
This is now generally available: https://squidfunk.github.io/mkdocs-material/insiders/#12000-piri-piri

More details here: https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/

miketheman:
Resolves #14112 

There's 
some very interesting stuff here - read through and make sure that it makes sense.

Notable:

- tags are prettier. 🤔 we could explore [Categories](https://squidfunk.github.io/mkdocs-material/setup/setting-up-a-blog/#adding-categories) vs Tags - but I don't think that's necessary
- posts must now include an excerpt marker `<!-- more -->` to denote the cutoff point in the index
- search!
- navigation improvements (next & previous at bottom of post, "back to top" when you start scrolling up from the bottom)
- `.authors.yml` is a central way of making authors, and it looks nicer.

Be aware: The sort order for the audit posts isn't correct. Bug filed and fixed, ~~but not yet released~~ released and ready in https://github.com/pypi/warehouse/pull/15029 .

# Example post

## Before
Screenshot 2023-12-07 at 17 33 12

## After
Screenshot 2023-12-07 at 17 33 40


ewdurbin:
> * posts must now include an excerpt marker `<!-- more -->` to denote the cutoff point in the index

If they do not, what happens? Does some validation fail in CI?

docs/mkdocs-blog.yml
```diff
@@ -59,14 +61,25 @@ plugins:
...
+      link: https://fosstodon.org/@pypi
     - icon: fontawesome/brands/twitter
       link: https://twitter.com/pypi
     - icon: material/rss
       link: https://blog.pypi.org/feed_rss_created.xml
+#   TODO: Get a property ID and uncomment the below block
+#  See https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/
+#  analytics:
+#    provider: google
+#    property: G-XXXXXXXXXX
```
ewdurbin, line 82: G-4L9L0H79CY

miketheman:
> > * posts must now include an excerpt marker `<!-- more -->` to denote the cutoff point in the index
> 
> If they do not, what happens? Does some validation fail in CI?

Yes, a configuration parameter `post_excerpt: required` will fail the build.

ewdurbin (approved):

docs/mkdocs-blog.yml
miketheman: Thanks! e6b5675af522090539c52f713a7fe2f8975ce087

PR #15078, opened 2023-12-17
Show yanked reason on release page

cofiem:
Releases that have been yanked can include a reason. This reason is not currently displayed on the PyPI website.

For releases that have been yanked, show the yanked reason in a callout box on the project's release page.

Closes #9837

See screenshot for an example.
image

PR #15115, opened 2023-12-27
Add yank reason to release history page

### LINKED ISSUE #9837: Public release history and release page do not show yanked reason
For yanked versions with a reason provided, PyPI shows this reason on neither the public release history nor the public release page.

I don't know where exactly it should be shown but I think it should be shown somewhere *at least* on the release page.

**Additional context**
See version 3.4.8 that *has* a yanked reason but it is not shown on the page:
https://pypi.org/project/Red-DiscordBot/#history
https://pypi.org/project/Red-DiscordBot/3.4.8/

cofiem:
Adds the yank reason to the release history page.

Notes:

- I'm happy to change the text colour to blue instead of red, but I thought the red matches the callout on the release page.
- I've added the existing `@include ellipsis;` to ensure the yank reason is always restricted to one line.
- As mentioned in the issue, the full (possibly multi-line) reason is on the release page for those who want to see it.
- I've included  `Reason this release was yanked:` with the screen reader-only class, because it's obvious looking at the release box, but a screenreader might need some context.

Closes #9837

image


warehouse/templates/packaging/detail.html
```diff
@@ -357,6 +357,12 @@ <h2 class="page-title split-layout">
                 <p class="release__version-date">
                   {{ humanize(hversion.created) }}
                 </p>
+                {% if hversion.yanked and hversion.yanked_reason %}
+                <p class="release__yanked-reason">
+                  <span class="sr-only">{% trans %}Reason this release was yanked:{% endtrans %}</span>
+                  {{ hversion.yanked_reason }}
+                </p>
+                {% endif %}
```
woodruffw, line 365: What do you think about making this a `callout-block callout-block--warning` instead? That'll mirror what other views on PyPI do for a warning block. 

For example, here's the warning callout on user-scoped API token creation:

Screenshot 2024-01-31 at 2 49 38 PM

woodruffw (commented):
This approach LGTM overall, thank you @cofiem! I left one thought about presentation.

warehouse/templates/packaging/detail.html
cofiem: Thanks for the review! On the release page, the callout is 'danger' / red. Should the yank reason callout here match the release page? https://github.com/pypi/warehouse/pull/15078#pullrequestreview-1787211887

woodruffw: Yeah, makes sense to me!

miketheman (changes requested):
Looking good - awaiting the final change for the callout block styling.

cofiem:
Yanked reason styling updated.
image

woodruffw (approved):
LGTM, thanks @cofiem!

miketheman:
@cofiem Thanks for working through this! I manually purged one example page to confirm, and it looks good: https://pypi.org/project/Red-DiscordBot/#history

The rest of pages should naturally expire from cache in a day, or as they yank/unyank their HTML will be refreshed.

PR #15130, opened 2024-01-02
warehouse/help: fixup API token guidance

woodruffw:
Now that 2FA is mandatory, references to passwords as an alternative for package uploads are incorrect.

Additionally, I've added a guidance point about creating a minimally scoped API token, as well as a point for using Trusted Publishing if the project's workflow/CI permits it.

See: https://discuss.python.org/t/announcement-2fa-now-required-for-pypi/42251/7?u=woodruffw


jeanas (approved):
Thanks!

miketheman (approved):
Thanks @woodruffw !

PR #15131, opened 2024-01-02
Remove per-project 2FA requirements

di:
Now that [2FA is required for all users](https://blog.pypi.org/posts/2024-01-01-2fa-enforced/), we can remove the per-project and per-organization 2FA settings, as well as remove the "criticality" designation (which was being used to enforce 2FA for a subset of critical projects).


warehouse/views.py
```diff
@@ -151,26 +151,6 @@ def forbidden(exc, request):
             )
             return HTTPSeeOther(url)
 
-        # If the forbidden error is because the user doesn't have 2FA enabled, we'll
-        # redirect them to the 2FA page
-        if exc.result.reason in {
-            "owners_require_2fa",
-            "pypi_mandates_2fa",
-            "manage_2fa_required",
```
miketheman, line 159: **question:** ‏With this line removed, how will a user without 2FA know what to do next? This reason is set in https://github.com/pypi/warehouse/pull/15131/files#diff-60b1172fcc3ed0f44469ecfff8ba1b5021c2fca3686eb4837522291a722ef624L358-L361

warehouse/packaging/tasks.py
```diff
@@ -211,130 +208,10 @@ def reconcile_file_storages(request):
...
-    new_projects = request.db.query(Project).filter(
-        Project.normalized_name.in_(project_names), Project.pypi_mandates_2fa.is_(False)
-    )
-
-    # Add them to the mandate
-    new_projects.update({Project.pypi_mandates_2fa: True})
-
-
 @tasks.task(ignore_result=True, acks_late=True)
 def compute_2fa_metrics(request):
```
miketheman, line 212: **thought (non-blocking):** ‏This is likely to break some of the metrics on our publicly-shared 2FA Dashboard. Not a problem per se, but something we should be aware of.

warehouse/packaging/models.py
```diff
@@ -148,18 +148,7 @@ def __contains__(self, project):
...
 
-class TwoFactorRequireable:
-    # Project owner requires 2FA for this project
-    owners_require_2fa: Mapped[bool_false]
-    # PyPI requires 2FA for this project
-    pypi_mandates_2fa: Mapped[bool_false]
-
-    @hybrid_property
-    def two_factor_required(self):
-        return self.owners_require_2fa | self.pypi_mandates_2fa
```
miketheman, line 159: **question:** ‏With this removal, I'd expect to see an associated migration in scope dropping the columns. Is that intentionally excluded from this changeset, and will follow?

miketheman (changes requested):

warehouse/views.py
```diff
@@ -151,26 +151,6 @@ def forbidden(exc, request):
             )
             return HTTPSeeOther(url)
 
-        # If the forbidden error is because the user doesn't have 2FA enabled, we'll
-        # redirect them to the 2FA page
-        if exc.result.reason in {
-            "owners_require_2fa",
```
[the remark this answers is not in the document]
di: These are left intentionally because they are required to continue rendering these events in the event history for a project: https://github.com/pypi/warehouse/blob/bcb3fec6d622a3b79116bd085fdd0a50bb37f7db/warehouse/templates/manage/project/history.html#L283-L293

warehouse/views.py
di: Thanks, good catch: f16fd1c

warehouse/packaging/tasks.py
di: I've already updated the dashboard accordingly.

warehouse/packaging/models.py
di: Nope, just forgot to run it: 214fbac

warehouse/views.py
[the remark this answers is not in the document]
miketheman: Makes sense - I'll leave this comment unresolved for our future selves if we ever want to entertain removing those tags and database events.

miketheman (approved):

Issue #13770, opened 2023-05-26
Remove support for passwords in upload API


dstufft:
With the [push to 2FA](https://blog.pypi.org/posts/2023-05-25-securing-pypi-with-2fa/) we should eventually just remove the ability to authenticate with username/password in uploads all together and just require an API token.

Ideally this comes after https://github.com/pypi/warehouse/issues/13769.

This differs from https://github.com/pypi/warehouse/issues/7265 in that https://github.com/pypi/warehouse/issues/7265 only applies to people who have 2FA on their account, this applies to *everyone*.

ewdurbin:
duplicate of #7265

ewdurbin:
ope sorry :)

miketheman:
A message will now be returned to users attempting to upload with basic auth.

We should eventually remove most of the associated code in the policy and simplify the logic to return the message regardless of whether the account works or not. Leaving this open as a reminder for that.

PR #15132, opened 2024-01-02
chore: return message on basic auth usage

miketheman:
~As upload is the only endpoint that allows for Basic Auth, remove it from the policy stack to move all calls to Macaroons.~

Return a message to users informing them to migrate.

Refs: #13770


miketheman:
This does not complete the related issue, as the code remains. I left a TODO to cleanup after we're happy with the change.

ewdurbin (approved):

di (approved):

PR #15137, opened 2024-01-03
refactor: make captcha a real service

miketheman:
Best reviewed commit by commit, since there's some that are not too interesting (example: mass rename).

Paves the way for a future captcha service via interface.


dev/environment
```diff
@@ -70,3 +70,6 @@ OIDC_AUDIENCE=pypi
 # Default to the reCAPTCHA testing keys from https://developers.google.com/recaptcha/docs/faq
 RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
 RECAPTCHA_SECRET_KEY=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+
+# Example of Captcha backend configuration
+# CAPTCHA_BACKEND=warehouse.captcha.recaptcha.Service
```
miketheman, line 75: A value isn't needed for deployment, since the default is currently `recaptcha.Service`

ewdurbin:
Looks great, my only concern was CSP but it appears we are ahead of that!

ewdurbin (approved):

PR #15143, opened 2024-01-05
Genericize OIDC templates, forms and views 

di:
This PR makes the OIDC templates, forms and views more "generic" and less GitHub-specific. This is a precursor to the fix for #13551 (which I have drafted). This PR doesn't introduce any functional changes.

After this, we should get https://github.com/pypi/warehouse/pull/14063 merged as well, and then we can begin adding other trusted publishers.

(cc @woodruffw)

### DIFF
--- a/warehouse/manage/views/__init__.py
+++ b/warehouse/manage/views/__init__.py
@@ -1131,6 +1131,10 @@ class ManageOIDCPublisherViews:
         self.request = request
         self.project = project
         self.metrics = self.request.find_service(IMetricsService, context=None)
+        self.github_publisher_form = GitHubPublisherForm(
+            self.request.POST,
+            api_token=self.request.registry.settings.get("github.token"),
+        )
 
     @property
     def _ratelimiters(self):
@@ -1161,13 +1165,6 @@ class ManageOIDCPublisherViews:
                     self.request.remote_addr
                 )
             )
-
-    @property
-    def github_publisher_form(self):
-        return GitHubPublisherForm(
-            self.request.POST,
-            api_token=self.request.registry.settings.get("github.token"),
-        )
 
     @property
     def default_response(self):
--- a/warehouse/oidc/forms/__init__.py
+++ b/warehouse/oidc/forms/__init__.py
@@ -9,18 +9,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import wtforms
 
-from warehouse import forms
-from warehouse.i18n import localize as _
+from warehouse.oidc.forms._core import DeletePublisherForm
+from warehouse.oidc.forms.github import GitHubPublisherForm, PendingGitHubPublisherForm
 
-
-class DeletePublisherForm(forms.Form):
-    __params__ = ["publisher_id"]
-
-    publisher_id = wtforms.StringField(
-        validators=[
-            wtforms.validators.InputRequired(message=_("Specify a publisher ID")),
-            wtforms.validators.UUID(message=_("Publisher must be specified by ID")),
-        ]
-    )
+__all__ = [
+    "DeletePublisherForm",
+    "GitHubPublisherForm",
+    "PendingGitHubPublisherForm",
+]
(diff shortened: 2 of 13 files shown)


woodruffw (approved):
LGTM structurally! I'm standing these changes up locally in a moment as well.

Edit: Confirmed that it works as expected locally as well :slightly_smiling_face:

warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
```diff
@@ -19,6 +19,14 @@ export default class extends Controller {
 
 
   initialize() {
+    // Select the tab with errors if there are any
+    this.tabPanelTargets.forEach((target, index) => {
+      const aElement = target.querySelector("div#errors");
+      if (aElement) {
+        this.index = index;
+      }
+    });
```
miketheman, line 28: **question:** ‏Should this be accompanied by a js test?

warehouse/oidc/forms/__init__.py
```diff
@@ -9,18 +9,12 @@
...
-
-class DeletePublisherForm(forms.Form):
-    __params__ = ["publisher_id"]
-
-    publisher_id = wtforms.StringField(
-        validators=[
-            wtforms.validators.InputRequired(message=_("Specify a publisher ID")),
-            wtforms.validators.UUID(message=_("Publisher must be specified by ID")),
-        ]
-    )
```
miketheman, line 26: **praise:** ‏nice move - I've started to really like `__init__.py` files that do nothing other than expose other imports.

warehouse/manage/views/__init__.py
```diff
@@ -1131,6 +1131,10 @@ def __init__(self, project, request):
         self.request = request
         self.project = project
         self.metrics = self.request.find_service(IMetricsService, context=None)
+        self.github_publisher_form = GitHubPublisherForm(
+            self.request.POST,
+            api_token=self.request.registry.settings.get("github.token"),
+        )
```
miketheman, line 1137: **question:** ‏can you help me understand this change? Before, it was a property, and it's still a property, and doesn't the same thing as before?

miketheman (approved):
Nice work, took me a bit to read through and figure this out.

A couple of questions and notes inline - nothing particularly blocking.

warehouse/manage/views/__init__.py
woodruffw: That's my understanding as well -- I think it's a purely lateral change (it's notionally slightly faster as well, although I don't think that'll be perceivable in this context).

warehouse/templates/manage/project/publishing.html
```diff
@@ -146,7 +105,64 @@ <h3>GitHub</h3>
...
+    </table>
+    {% else %}
+    <p class="no-bottom-padding">{% trans %}No publishers are currently configured.{% endtrans %}</p>
+    {% endif %}
+
+    <h2 class="no-bottom-padding">{% trans %}Add a new publisher{% endtrans %}</h2>
+
+    {% if request.user.has_two_factor %}
+
+    {% set publishers = [("GitHub", github_form(request, github_publisher_form))] %}
```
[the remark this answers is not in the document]
woodruffw: Agreed, although I'm not sure where best this would live on the Python side. Maybe in the default render for the associated view?

(I can do this in follow-up if the two of you would prefer.)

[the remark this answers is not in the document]
di: This is a mapping of the publisher name to the result of a Jinja macro that takes the form as input, not the form itself. I also don't really see where this would live on the Python side unfortunately, but since we'll always need some custom per-provider HTML to render the form, I think this is acceptable.

warehouse/manage/views/__init__.py
di: I would call these attributes now, not properties. They are functionally the same in usage but this ensures that they are only defined once rather than every time they are accessed.

I started this PR several months ago and I don't remember exactly I made this change but I suspect the properties were causing problems with testing, but also there was no reason for them to be properties in the first place.

warehouse/static/js/warehouse/controllers/horizontal_tabs_controller.js
di: Grumble grumble, ac4f804.

miketheman (approved):

PR #15148, opened 2024-01-05
Refactor OIDC mint token endpoint (cont. #14063)

woodruffw:
See #14063.

For merge: please preserve the original author, since this is just me getting things deconflicted/resolved.


tests/conftest.py
```diff
@@ -379,6 +380,57 @@ def oidc_service(db_session):
     )
 
 
+@pytest.fixture
+def dummy_oidc_jwt():
```
woodruffw, line 384: NB: Added this to de-dupe a bunch of the hard-coded test JWTs in individual tests.

tests/unit/oidc/test_utils.py
```diff
@@ -115,3 +115,21 @@ def test_oidc_context_principals():
...
+    for iss in utils.OIDC_ISSUER_URLS:
+        # Each issuer should be present in each mapping.
+        assert iss in utils.OIDC_ISSUER_SERVICE_NAMES
+        assert iss in utils.OIDC_ISSUER_ADMIN_FLAGS
+        assert iss in utils.OIDC_PUBLISHER_CLASSES
+
+    for class_map in utils.OIDC_PUBLISHER_CLASSES.values():
+        # The class mapping for pending and non-pending publisher models
+        # should be distinct.
+        assert class_map[True] != class_map[False]
```
woodruffw, line 135: Technically unrelated to the rest of these changes, but I've added this as another "backstop" to ensure these various constants remain consistent with each other.

warehouse/oidc/views.py
```diff
@@ -68,81 +96,119 @@ def oidc_audience(request):
...
-    def _invalid(errors):
-        request.response.status = 422
-        return {"message": "Token request failed", "errors": errors}
+@view_config(
+    route_name="oidc.mint_token",
+    require_methods=["POST"],
+    renderer="json",
+    require_csrf=False,
+    has_translations=True,
+)
```
woodruffw, line 105: NB: The diff makes it a bit hard to read here, but this view now has two configured routes (each with its own `@view_config`): `oidc.github.mint_token` *and* `oidc.mint_token`.

warehouse/oidc/views.py
```diff
@@ -68,81 +96,119 @@ def oidc_audience(request):
...
+    try:
+        unverified_claims = jwt.decode(
+            unverified_jwt, options=dict(verify_signature=False)
+        )
+        unverified_issuer: str = unverified_claims["iss"]
+    except Exception:
+        return _invalid(
+            errors=[{"code": "invalid-payload", "description": "malformed JWT"}],
+            request=request,
+        )
```
woodruffw, line 129: NB: This is an inherently risky thing to do, but is also unfortunately the only way to get from a specific JWT to the service capable of verifying it.

This is safe *so long as* the `OIDCPublisherService` lookup that happens from this unverified claim never yields a service that no-ops the actual JWT signature and claim verification. That being said, this deserves multiple eyes on it to make sure I haven't missed something important here 🙂

warehouse/oidc/views.py
th3coop: Ahhhh THIS is how you do this part!  This was actually something I meant to ask specifically about as I couldn't figure it out from the applicable docs.

warehouse/oidc/views.py
```diff
@@ -68,81 +96,119 @@ def oidc_audience(request):
...
-def mint_token_from_oidc(request):
-    def _invalid(errors):
-        request.response.status = 422
-        return {"message": "Token request failed", "errors": errors}
+@view_config(
+    route_name="oidc.mint_token",
+    require_methods=["POST"],
+    renderer="json",
+    require_csrf=False,
+    has_translations=True,
```
di, line 104: Unrelated to the PR, but I think `has_translations=True` here is unnecesary.

warehouse/oidc/views.py
```diff
@@ -68,81 +96,119 @@ def oidc_audience(request):
...
+    # We currently have an **unverified** JWT. To verify it, we need to
+    # know which OIDC service's keyring to check it against.
+    # To do this, we gingerly peek into the unverified claims and
+    # use the `iss` to key into the right `OIDCPublisherService`.
+    try:
+        unverified_claims = jwt.decode(
+            unverified_jwt, options=dict(verify_signature=False)
+        )
+        unverified_issuer: str = unverified_claims["iss"]
+    except Exception:
```
di, line 125: Can we make this a more specific set of exceptions?

warehouse/oidc/views.py
```diff
@@ -68,81 +96,119 @@ def oidc_audience(request):
...
+
+    if request.flags.disallow_oidc(OIDC_ISSUER_ADMIN_FLAGS[unverified_issuer]):
         return _invalid(
             errors=[
                 {
                     "code": "not-enabled",
-                    "description": (
-                        "GitHub-based trusted publishing functionality not enabled"
-                    ),
+                    "description": "Trusted publishing functionality not enabled",  # noqa
```
di, line 149: We're sort of losing some specificity here, at the very least we can use `OIDC_ISSUER_SERVICE_NAMES` to make this more specific.

warehouse/oidc/views.py
woodruffw: Yeah, I think we could probably make this `(KeyError, PyJWTError)`, although we'll probably have the same exception leakiness we saw in `verify_jwt_signature`:

https://github.com/pypi/warehouse/blob/de4062b6b27f9386e038ee8777e2afbbfdf2cca6/warehouse/oidc/services.py#L223-L240

warehouse/oidc/views.py
woodruffw: Yeah, this is not super well documented anywhere in Pyramid that I can tell 🙂 -- I only know this trick from seeing it used elsewhere in Warehouse.

warehouse/oidc/views.py
di: @woodruffw I think you forgot to address this.

di: (Or maybe your comment was actually a question about whether we should do that instead, IMO yes)

woodruffw: Yep, was meant to be a comment, sorry for the ambiguity! I'll do this now.

woodruffw: Done in [c6a8d96](https://github.com/pypi/warehouse/pull/15148/commits/c6a8d965ef9b367dfabf53d898cf0f3dbac6b604) -- we still need to catch the generic `Exception` because `jwt.decode` might be leaky, but we specialize its handling in the same way mentioned above.

woodruffw:
I believe this is good to go -- @di are you thinking this should get merged first, or after #15144?

di (approved):

di:
I think we can merge this first!

PR #15168, opened 2024-01-08
OIDC active state form

th3coop:
This PR follows from #13980.  It adds form management portion of the ActiveState Publisher in PyPI Account settings.

Without #13980 merged, this PR have about 1400 more changed lines than it should.  It'll look more like this when 13980 is merged: https://github.com/ActiveState/warehouse/pull/3

### DIFF
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -544,29 +544,68 @@ class TestActiveStatePublisherForm:
     @pytest.mark.parametrize(
         "data",
         [
-            {"organization": None, "project": "good", "actor": "good"},
-            {"organization": "", "project": "good", "actor": "good"},
+            # Too short
+            # Too long
+            # Invalid characters
+            # No leading or ending -
+            # No double --
+            # Missing
+            # Empty
+            #
+            # organization
+            {"organization": "AB", "project": "good", "actor": "good"},
+            {
+                "organization": "abcdefghojklmnopqrstuvwxyz123456789012345",
+                "project": "good",
+                "actor": "good",
+            },
             {
                 "organization": "invalid_characters@",
                 "project": "good",
                 "actor": "good",
             },
-            {"project": None, "organization": "good", "actor": "good"},
-            {"project": "", "organization": "good", "actor": "good"},
+            {"organization": "-foo-", "project": "good", "actor": "good"},
+            {"organization": "---", "project": "good", "actor": "good"},
+            {"organization": "", "project": "good", "actor": "good"},
+            {"organization": None, "project": "good", "actor": "good"},
+            # actor
+            {"actor": "AB", "project": "good", "organization": "good"},
             {
-                "project": "$invalid#characters",
+                "actor": "abcdefghojklmnopqrstuvwxyz123456789012345",
+                "project": "good",
                 "organization": "good",
-                "actor": "good",
             },
-            {"actor": None, "project": "good", "organization": "good"},
+            {
+                "actor": "invalid_characters@",
+                "project": "good",
+                "organization": "good",
+            },
+            {"actor": "-foo-", "project": "good", "organization": "good"},
+            {"actor": "---", "project": "good", "organization": "good"},
             {"actor": "", "project": "good", "organization": "good"},
-            {"actor": "$invalid#characters", "project": "good", "organization": "good"},
+            {"actor": None, "project": "good", "organization": "good"},
+            # project
+            {"project": "AB", "actor": "good", "organization": "good"},
+            {
+                "project": "abcdefghojklmnopqrstuvwxyz123456789012345",
+                "actor": "good",
+                "organization": "good",
+            },
+            {
+                "project": "invalid_characters@",
+                "actor": "good",
+                "organization": "good",
+            },
+            {"project": "-foo-", "actor": "good", "organization": "good"},
+            {"project": "---", "actor": "good", "organization": "good"},
+            {"project": "", "actor": "good", "organization": "good"},
+            {"project": None, "actor": "good", "organization": "good"},
         ],
     )
     def test_validate_basic_invalid_fields(self, monkeypatch, data):
+        print(data)
         form = activestate.ActiveStatePublisherForm(MultiDict(data))
 
-        # We're testing only the basic validation here.
         monkeypatch.setattr(form, "_lookup_actor", lambda o: fake_user_info)
         monkeypatch.setattr(form, "_lookup_organization", lambda o: None)
 
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -540,6 +540,16 @@ class TestActiveStatePublisherForm:
     @pytest.mark.parametrize(
         "data",
         [
+            # Organization
+            # Missing
+            # Empty
+            {"organization": "", "project": "good", "actor": "good"},
+            # Actor
+            # Missing
+            # Empty
+            {"actor": "", "project": "good", "organization": "good"},
+            {"actor": None, "project": "good", "organization": "good"},
+            # Project
             # Too short
             # Too long
             # Invalid characters
@@ -547,40 +557,6 @@ class TestActiveStatePublisherForm:
             # No double --
             # Missing
             # Empty
-            #
-            # organization
-            {"organization": "AB", "project": "good", "actor": "good"},
-            {
-                "organization": "abcdefghojklmnopqrstuvwxyz123456789012345",
-                "project": "good",
-                "actor": "good",
-            },
-            {
-                "organization": "invalid_characters@",
-                "project": "good",
-                "actor": "good",
-            },
-            {"organization": "-foo-", "project": "good", "actor": "good"},
-            {"organization": "---", "project": "good", "actor": "good"},
-            {"organization": "", "project": "good", "actor": "good"},
-            {"organization": None, "project": "good", "actor": "good"},
-            # actor
-            {"actor": "AB", "project": "good", "organization": "good"},
-            {
-                "actor": "abcdefghojklmnopqrstuvwxyz123456789012345",
-                "project": "good",
-                "organization": "good",
-            },
-            {
-                "actor": "invalid_characters@",
-                "project": "good",
-                "organization": "good",
-            },
-            {"actor": "-foo-", "project": "good", "organization": "good"},
-            {"actor": "---", "project": "good", "organization": "good"},
-            {"actor": "", "project": "good", "organization": "good"},
-            {"actor": None, "project": "good", "organization": "good"},
-            # project
             {"project": "AB", "actor": "good", "organization": "good"},
             {
                 "project": "abcdefghojklmnopqrstuvwxyz123456789012345",
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -91,7 +91,11 @@ class TestActiveStatePublisherForm:
             content=b"fake-content",
         )
         requests = pretend.stub(
-            post=pretend.call_recorder(lambda o, **kw: response), HTTPError=HTTPError
+            post=pretend.call_recorder(lambda o, **kw: response),
+            expception=_requests.exceptions,
+            Timeout=Timeout,
+            HTTPError=HTTPError,
+            ConnectionError=ConnectionError,
         )
 
         monkeypatch.setattr(activestate, "requests", requests)
@@ -120,7 +124,10 @@ class TestActiveStatePublisherForm:
         )
         requests = pretend.stub(
             post=pretend.call_recorder(lambda o, **kw: response),
+            expception=_requests.exceptions,
+            Timeout=Timeout,
             HTTPError=HTTPError,
+            ConnectionError=ConnectionError,
         )
         monkeypatch.setattr(activestate, "requests", requests)
 
@@ -143,9 +150,7 @@ class TestActiveStatePublisherForm:
         ]
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState actor lookup: " "b'fake-content'"
-            )
+            pretend.call("Unexpected 422 error from ActiveState API: b'fake-content'")
         ]
 
     def test_lookup_actor_http_timeout(self, monkeypatch):
@@ -165,7 +170,7 @@ class TestActiveStatePublisherForm:
             form._lookup_actor(fake_username)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call("Timeout from ActiveState actor lookup API (possibly offline)")
+            pretend.call("Connection error from ActiveState API")
         ]
 
     def test_lookup_actor_connection_error(self, monkeypatch):
@@ -185,9 +190,7 @@ class TestActiveStatePublisherForm:
             form._lookup_actor(fake_username)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Connection error from ActiveState actor lookup API (possibly offline)"
-            )
+            pretend.call("Connection error from ActiveState API")
         ]
 
     def test_lookup_actor_non_json(self, monkeypatch):
@@ -213,7 +216,7 @@ class TestActiveStatePublisherForm:
             form._lookup_actor(fake_username)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call("Unexpected error from ActiveState actor lookup: b''")  # noqa
+            pretend.call("Unexpected error from ActiveState API: b''")
         ]
 
     def test_lookup_actor_gql_error(self, monkeypatch):
@@ -248,9 +251,7 @@ class TestActiveStatePublisherForm:
             )
         ]
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState actor lookup: ['some error']"
-            )
+            pretend.call("Unexpected error from ActiveState API: ['some error']")
         ]
 
     def test_lookup_actor_gql_no_data(self, monkeypatch):
@@ -288,7 +289,11 @@ class TestActiveStatePublisherForm:
             json=lambda: fake_gql_user_response,
         )
         requests = pretend.stub(
-            post=pretend.call_recorder(lambda o, **kw: response), HTTPError=HTTPError
+            post=pretend.call_recorder(lambda o, **kw: response),
+            HTTPError=HTTPError,
+            requests=_requests.exceptions,
+            Timeout=_requests.Timeout,
+            ConnectionError=_requests.ConnectionError,
         )
         monkeypatch.setattr(activestate, "requests", requests)
 
@@ -305,7 +310,6 @@ class TestActiveStatePublisherForm:
                 timeout=5,
             )
         ]
-        assert response.raise_for_status.calls == [pretend.call()]
         assert info == fake_user_info
 
     # _lookup_organization
@@ -319,6 +323,8 @@ class TestActiveStatePublisherForm:
             post=pretend.call_recorder(lambda o, **kw: response),
             HTTPError=HTTPError,
             exceptions=_requests.exceptions,
+            Timeout=_requests.Timeout,
+            ConnectionError=_requests.ConnectionError,
         )
 
         monkeypatch.setattr(activestate, "requests", requests)
@@ -348,6 +354,9 @@ class TestActiveStatePublisherForm:
         requests = pretend.stub(
             post=pretend.call_recorder(lambda o, **kw: response),
             HTTPError=HTTPError,
+            exceptions=_requests.exceptions,
+            Timeout=_requests.Timeout,
+            ConnectionError=_requests.ConnectionError,
         )
         monkeypatch.setattr(activestate, "requests", requests)
 
@@ -370,10 +379,7 @@ class TestActiveStatePublisherForm:
         ]
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState organization lookup: "
-                "b'fake-content'"
-            )
+            pretend.call("Unexpected 422 error from ActiveState API: b'fake-content'")
         ]
 
     def test_lookup_organization_http_timeout(self, monkeypatch):
@@ -393,9 +399,7 @@ class TestActiveStatePublisherForm:
             form._lookup_organization(fake_org_name)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Timeout from ActiveState organization lookup API (possibly offline)"
-            )
+            pretend.call("Connection error from ActiveState API")
         ]
 
     def test_lookup_organization_connection_error(self, monkeypatch):
@@ -415,9 +419,7 @@ class TestActiveStatePublisherForm:
             form._lookup_organization(fake_org_name)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Connection error from ActiveState organization lookup API (possibly offline)"  # noqa
-            )
+            pretend.call("Connection error from ActiveState API")
         ]
 
     def test_lookup_organization_non_json(self, monkeypatch):
@@ -443,9 +445,7 @@ class TestActiveStatePublisherForm:
             form._lookup_organization(fake_org_name)
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState organization lookup: b''"  # noqa
-            )
+            pretend.call("Unexpected error from ActiveState API: b''")
         ]
 
     def test_lookup_organization_gql_error(self, monkeypatch):
@@ -481,9 +481,7 @@ class TestActiveStatePublisherForm:
             )
         ]
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState organization lookup: ['some error']"
-            )
+            pretend.call("Unexpected error from ActiveState API: ['some error']")
         ]
 
     def test_lookup_organization_gql_no_data(self, monkeypatch):
@@ -518,7 +516,6 @@ class TestActiveStatePublisherForm:
     def test_lookup_organization_succeeds(self, monkeypatch):
         response = pretend.stub(
             status_code=200,
-            raise_for_status=pretend.call_recorder(lambda: None),
             json=lambda: fake_gql_org_response,
         )
         requests = pretend.stub(
@@ -539,7 +536,6 @@ class TestActiveStatePublisherForm:
                 timeout=5,
             )
         ]
-        assert response.raise_for_status.calls == [pretend.call()]
 
     @pytest.mark.parametrize(
         "data",
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -1,0 +1,518 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pretend
+import pytest
+import wtforms
+
+from requests import ConnectionError, HTTPError, Timeout
+from webob.multidict import MultiDict
+
+from warehouse.oidc.forms import activestate
+
+fake_username = "some-username"
+fake_org_name = "some-org"
+fake_user_info = {"user_id": "some-user-id"}
+fake_org_info = {"added": "somedatestring"}
+fake_gql_org_response = {"data": {"organizations": [fake_org_info]}}
+fake_gql_user_response = {"data": {"users": [fake_user_info]}}
+
+
+class TestPendingActiveStatePublisherForm:
+    def test_validate(self, monkeypatch):
+        project_factory = []
+        data = MultiDict(
+            {
+                "organization": "some-org",
+                "project": "some-project",
+                "actor": "someuser",
+                "project_name": "some-project",
+            }
+        )
+        form = activestate.PendingActiveStatePublisherForm(
+            MultiDict(data), project_factory=project_factory
+        )
+
+        # Test built-in validations
+        monkeypatch.setattr(form, "_lookup_actor", lambda *o: {"user_id": "some-id"})
+
+        monkeypatch.setattr(
+            form, "_lookup_organization", lambda *o: {"added": "somedatestring"}
+        )
+
+        assert form._project_factory == project_factory
+        assert form.validate()
+
+    def test_validate_project_name_already_in_use(self):
+        project_factory = ["some-project"]
+        form = activestate.PendingActiveStatePublisherForm(
+            project_factory=project_factory
+        )
+
+        field = pretend.stub(data="some-project")
+        with pytest.raises(wtforms.validators.ValidationError):
+            form.validate_project_name(field)
+
+
+class TestActiveStatePublisherForm:
+    def test_validate(self, monkeypatch):
+        data = MultiDict(
+            {
+                "organization": "some-org",
+                "project": "some-project",
+                "actor": "someuser",
+            }
+        )
+        form = activestate.ActiveStatePublisherForm(MultiDict(data))
+
+        org_info = {"added": "somedatestring"}
+        monkeypatch.setattr(form, "_lookup_organization", lambda o: org_info)
+        owner_info = {"user_id": "some-user-id"}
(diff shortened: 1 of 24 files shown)


di:
I'll review this once #13980 is merged and conflicts are resolved here!

th3coop:
@di, @woodruffw.  Before you give this PR a super thorough review, I'll point out that I've included quite a bit of duplication tests/unit/accounts/test_views.py.  I think I would be able to avoid this if I changed all of the validation tests in that file to call `_add_pending_oidc_publisher` and perhaps parameterize passing in the respective services values so I can check stuff like [this](https://github.com/ActiveState/warehouse/blob/main/warehouse/accounts/views.py#L1533) has the right service name in the string?

That could result in quite large or complex parameterize lists though.

Any thoughts?

di:
@th3coop We already have parametrized tests for these which were added recently, e.g. for that specific example: https://github.com/pypi/warehouse/blob/e62049b47483464239fc35df01854274da16c091/tests/unit/accounts/test_views.py#L3442-L3526

I think if you get this branch up to date, you should just be able to add a new ActiveState parameter to these existing tests, rather than duplicating them.

th3coop:
@di, i was expecting these merge conflicts to go away when https://github.com/pypi/warehouse/pull/13980 was merged since I did a fresh rebase yesterday but I still see them now that that code is in `main`.  I'll rebase again.  I am now realizing that those are the params you already told me about a couple days ago.

di:
@th3coop Is it possible you're rebasing on the out-of-date `main` from the Activestate fork rather than from PyPI's remote?

th3coop:
@di, that is possible.  I often pull in the upstream `main` to my local environment though.  And I was certain I had done that before doing my last series of rebases but I could easily forgotten.  Sorry for the delay on this one.  I really thought it was ready for review already.  I'm fixing these merge conflicts right now.

th3coop:
OK.  @di,I got things fixed up.  Again, sorry about that mixup.  Also, I'm working on getting **Allowing changes to a pull request branch created from a fork** enabled on this PR.  I don't appear to have permissions to do that.

th3coop:
Hey @di.  It appears that we can't turn on **Allowing changes to a pull request branch created from a fork
**.  The option to enable that doesn't show up for myself or any admins of our GitHub account and I can't find anything on the internet as to why that might be.  I'm like 75% sure that it's because our fork is in the ActiveState org and not my personal GitHub account.  That's my best guess based on step 4 [from the docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) where it says **On user-owned forks**.

If not having this turned on is a huge pain for y'all, let me know and I can move the remaining PRs to my personal account.

woodruffw:
Yeah, that's an unfortunate known long-standing bug with GitHub orgs and forks. The hacky workaround is to manually give the person commit rights on your fork, but that's not always desirable :upside_down_face:

th3coop:
I'm fine with that! Thanks for the workaround @woodruffw.  Invites sent to yourself and DI.  I'll remove you both when we get the PRs merged.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+import requests
+import sentry_sdk
+import wtforms
+
+from warehouse import forms
+from warehouse.i18n import localize as _
+from warehouse.utils.project import PROJECT_NAME_RE
+
+_VALID_ORG_URL_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
+_VALID_PROJECT_NAME = re.compile(r"^[.a-zA-Z0-9-]{3,40}$")
```
di, line 26: Just sanity checking these... `---` or `-foo-` is a valid "org url" name? `...` or `-.-` is a valid project name?

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+        ]
+    )
+
+    actor = wtforms.StringField(
+        validators=[
+            wtforms.validators.InputRequired(
+                message=("Specify the ActiveState actor username")
+            ),
+            wtforms.validators.Regexp(
+                _VALID_ORG_URL_NAME,
```
di, line 73: Should the org regex be reused here?

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+                    "Try again in a few minutes"
+                )
+            )
+        except requests.Timeout:
+            sentry_sdk.capture_message(
+                "Timeout from ActiveState organization lookup API (possibly offline)"
+            )
+            raise wtforms.validators.ValidationError(
+                _("Unexpected timeout from ActiveState. Try again in a few minutes")
+            )
```
di, line 118: Seems like we could probably condense these into:

```python
except (requests.HTTPError, requests.ConnectionError, request.Timeout) as e:
    sentry_sdk.capture_message(
        f"Unexpected error {e!r} from ActiveState organization lookup: {response.content!r}"
    )
```

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+        except requests.exceptions.JSONDecodeError:
+            sentry_sdk.capture_message(
+                f"Unexpected error from ActiveState organization lookup: {response.content!r}"  # noqa
+            )
+            raise wtforms.validators.ValidationError(
+                _("Unexpected error from ActiveState. Try again")
+            )
+
+    def validate_organization(self, field):
+        self._lookup_organization(field.data)
```
di, line 146: Any reason not to just put the body of this function here instead?

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+                f"Unexpected error from ActiveState actor lookup: {response.content!r}"
+            )
+            raise wtforms.validators.ValidationError(
+                _("Unexpected error from ActiveState. Try again")
+            )
+
+    def validate_actor(self, field):
+        actor = field.data
+
+        actor_info = self._lookup_actor(actor)
```
di, line 214: Ditto here.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+        super().__init__(*args, **kwargs)
+        self._project_factory = project_factory
+
+    def validate_project_name(self, field):
+        project_name = field.data
+
+        if project_name in self._project_factory:
+            raise wtforms.validators.ValidationError(
+                _("This project name is already in use")
+            )
```
di, line 241: This should use the `PendingPublisherMixin` instead.

warehouse/oidc/forms/activestate.py
th3coop: They are not.  I've missed a few validations here.  Fixing and adding a couple more test cases.

warehouse/oidc/forms/activestate.py
th3coop: Yes, we use the same validation on usernames.  Maybe I should make that explicit and name this in a way that it's clear that it's meant to be for both.  Would you prefer that?

warehouse/oidc/forms/activestate.py
th3coop: This is so nice.  Thanks!

warehouse/oidc/forms/activestate.py
th3coop: No dice on this.  Some step in the code path of looking up validation functions based on the field name causes stubbing not to work.  When I try to stub `validate_organization`, the original code runs instead of the stubbed function.  If I keep `_lookup_organization` the stub works.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def _lookup_organization(self, org_url_name: str) -> None:
+        """Make gql API call to the ActiveState API to check if the organization
+        exists"""
+        try:
+            response = requests.post(
+                f"{ACTIVESTATE_GRAPHQL_API_URL}",
```
woodruffw, line 87: Nitpick: unnecessary f-string:

```suggestion
                ACTIVESTATE_GRAPHQL_API_URL,
```

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def validate_organization(self, field):
+        self._lookup_organization(field.data)
+
+    def _lookup_actor(self, actor: str) -> UserResponse:
+        """Make gql API call to the ActiveState API to check if the actor/username
+        exists and return the associated user id"""
+        try:
+            response = requests.post(
+                f"{ACTIVESTATE_GRAPHQL_API_URL}",
```
woodruffw, line 153:
```suggestion
                ACTIVESTATE_GRAPHQL_API_URL,
```

warehouse/oidc/forms/activestate.py
woodruffw, line 87: Nitpick, nonblocking: given that this `POST` is made both here and in `_lookup_actor` and has mostly identical error handling in both cases, maybe we can abstract it into a helper function? Something like:

```python
def _activestate_request(query: dict, variables: dict) -> dict: ...
```

(OTOH, this might make the error states harder to handle, so maybe not worth it.)

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+
+_VALID_ORG_URL_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
+_VALID_PROJECT_NAME = re.compile(r"^[.a-zA-Z0-9-]{3,40}$")
+
+ACTIVESTATE_GRAPHQL_API_URL = "https://platform.activestate.com/graphql/v1/graphql"
+GRAPHQL_GET_ORGANIZATION = "query($orgname: String) {organizations(where: {display_name: {_eq: $orgname}}) {added}}"  # noqa
+GRAPHQL_GET_ACTOR = (
+    "query($username: String) {users(where: {username: {_eq: $username}}) {user_id}}"
+)
+TIMEOUT = 5
```
woodruffw, line 33: Nitpick, nonblocking: consider making these `_ACTIVESTATE_GRAPHQL_API_URL`, etc. since they only get used in this module :slightly_smiling_face:

woodruffw (commented):
A couple of small nitpicks; otherwise this is looking really good to me! I'm going to stand it up locally tomorrow and do some test-driving as well.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,245 @@
...
+        try:
+            response = requests.post(
+                f"{ACTIVESTATE_GRAPHQL_API_URL}",
+                json={
+                    "query": GRAPHQL_GET_ORGANIZATION,
+                    "variables": {"orgname": org_url_name},
+                },
+                timeout=TIMEOUT,
+            )
+            response.raise_for_status()
```
th3coop, line 94: I've removed this and instead am doing a check like this:
`if response.status_code >= 400:`
because I'm "UnboundLocalError" when a TimeOut error occurs. This arose when i collapsed all the HTTPError handling into 1 `except` line.

warehouse/oidc/forms/activestate.py
th3coop: I made it clearer what this can be used for.

warehouse/oidc/forms/activestate.py
th3coop: Done locally.

th3coop: This worked out nicely.  I'm using a "response handler" function to do handle the unique form code paths where a clear validation error needs to be displayed.

th3coop:
@woodruffw.  All set.

warehouse/oidc/forms/activestate.py
```diff
@@ -12,31 +12,100 @@
...
 
 from warehouse import forms
 from warehouse.i18n import localize as _
-from warehouse.utils.project import PROJECT_NAME_RE
+from warehouse.oidc.forms._core import PendingPublisherMixin
 
-_VALID_ORG_URL_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
+_VALID_ORG_URL_NAME_AND_ACTOR_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
 _VALID_PROJECT_NAME = re.compile(r"^[.a-zA-Z0-9-]{3,40}$")
+_DOUBLE_DADHES = re.compile(r"--+")
```
woodruffw, line 28: Typo?

```suggestion
_DOUBLE_DASHES = re.compile(r"--+")
```

th3coop: Doh...Indeed. Fixed locally to update where it's used as well.

th3coop: Pushed.

tests/unit/oidc/forms/test_activestate.py
```diff
@@ -143,9 +150,7 @@ def test_lookup_actor_other_http_error(self, monkeypatch):
         ]
 
         assert sentry_sdk.capture_message.calls == [
-            pretend.call(
-                "Unexpected error from ActiveState actor lookup: " "b'fake-content'"
-            )
+            pretend.call("Unexpected 422 error from ActiveState api: b'fake-content'")
```
th3coop, line 153: Patched this lower case "API" to be upper case.

tests/unit/oidc/forms/test_activestate.py
```diff
@@ -371,8 +380,7 @@ def test_lookup_organization_other_http_error(self, monkeypatch):
 
         assert sentry_sdk.capture_message.calls == [
             pretend.call(
-                "Unexpected error from ActiveState organization lookup: "
-                "b'fake-content'"
+                f"Unexpected {422} error from ActiveState api: b'fake-content'"
```
th3coop, line 383: Remove unnecesssary formatting string and uppercased API

woodruffw (approved):
LGTM overall! I didn't get to the local deployment today, but will test tomorrow.

warehouse/accounts/views.py
```diff
@@ -1514,13 +1520,17 @@ def default_response(self):
...
             "disabled": {
                 "GitHub": self.request.flags.disallow_oidc(
                     AdminFlagValue.DISALLOW_GITHUB_OIDC
                 ),
                 "Google": self.request.flags.disallow_oidc(
                     AdminFlagValue.DISALLOW_GOOGLE_OIDC
                 ),
+                "ActiveState": self.request.flags.disallow_oidc(
+                    AdminFlagValue.DISALLOW_ACTIVESTATE_OIDC
+                ),
```
woodruffw, line 1533: No action required, but this is making me realize that we should probably get rid of these hardcoded provider names and replace them with something in `oidc.utils`.

warehouse/manage/views/__init__.py
```diff
@@ -1421,6 +1432,118 @@ def add_google_oidc_publisher(self):
...
+                "specifier": str(publisher),
+                "url": publisher.publisher_url(),
+                "submitted_by": self.request.user.username,
+            },
+        )
+
+        self.request.session.flash(
+            f"Added {publisher} "
+            + (f"in {publisher.publisher_url()}" if publisher.publisher_url() else "")
+            + f" to {self.project.name}",
```
woodruffw, line 1537: I think `publisher_url()` can never be falsey here, right? Based on:

https://github.com/pypi/warehouse/blob/30c8720951db492dbd800bb5433edbdf1fc6085f/warehouse/oidc/models/activestate.py#L125-L128

So, I think this can be:

```suggestion
            f"Added {publisher} "
            + f"in {publisher.publisher_url()} "
            + f"to {self.project.name}",
```

(and then maybe reduce the lines?)

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,213 @@
...
+_VALID_ORG_URL_NAME_AND_ACTOR_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
+_VALID_PROJECT_NAME = re.compile(r"^[.a-zA-Z0-9-]{3,40}$")
+_DOUBLE_DASHES = re.compile(r"--+")
+
+_ACTIVESTATE_GRAPHQL_API_URL = "https://platform.activestate.com/graphql/v1/graphql"
+_GRAPHQL_GET_ORGANIZATION = "query($orgname: String) {organizations(where: {display_name: {_eq: $orgname}}) {added}}"  # noqa
+_GRAPHQL_GET_ACTOR = (
+    "query($username: String) {users(where: {username: {_eq: $username}}) {user_id}}"
+)
+_TIMEOUT = 5
```
woodruffw, line 35: Nitpick, nonblocking: I think `_TIMEOUT` is now only used in one place, so we can just drop the constant.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,213 @@
...
+        raise wtforms.validators.ValidationError(
+            _("Double dashes are not allowed in the name")
+        )
+
+
+def _no_leading_or_trailing_dashes(form, field):
+    if field.data.startswith("-") or field.data.endswith("-"):
+        raise wtforms.validators.ValidationError(
+            _("Leading or trailing dashes are not allowed in the name")
+        )
```
woodruffw, line 58: Thought, nonblocking: I think this could be worked into the regexps themselves, rather than having a separate validator. But IMO this is not worth trying to do now, something we can iterate on in the future 🙂

woodruffw (commented):
One change request and a few nitpicks that I leave up to you!

This is looking really good!

warehouse/manage/views/__init__.py
th3coop: Oh man, good eye.

warehouse/accounts/views.py
th3coop: I'm going to leave this one out as there's a bit (very little) of nuance to what the change could be exactly.

warehouse/oidc/forms/activestate.py
th3coop: I think I'm going to skip this as I actually prefer if we had more isolated validators that can give specific feedback.  The current regex, if it were to give helpful feedback, needs to state multiple things as the possible fix.  With this specific validator, it can state exactly what's wrong.

warehouse/manage/views/__init__.py
th3coop: Having a bit of trouble with this change.  It looks like the [Google Publisher Model](https://github.com/ActiveState/warehouse/blob/main/warehouse/oidc/models/google.py#L84) returns None for `publisher_url` and I'm not, if that can even be changed, what it would be.

Mind if I leave this change out as well?

th3coop: Actually, this points out an issue in my code.  In my attempted to be type safe, i'm breaking the interface we're using for the PublisherMixin.  I'm going to change `publisher_url` in the AtiveState Publisher mixin to not have the type declarations I added.  This can be iterated on later and done across all published to make sure they match.

th3coop: OK, sorry for the noise.  I DID make this change but I also backed out my additional type declarations that make the `ActiveStatePublisherMixin.publisher_url` not match what the others `*PublisherMixin.publisher_url` have.

th3coop:
Okie dokie.  All changes pushed, @woodruffw.

woodruffw:
Thanks! I'm taking another look today.

(You'll need to update against `main` for the merge as well.)

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,212 @@
...
+                _VALID_ORG_URL_NAME_AND_ACTOR_NAME,
+                message=("Invalid ActiveState username"),
+            ),
+            _no_double_dashes,
+            _no_leading_or_trailing_dashes,
+        ]
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
woodruffw, line 158: I think this can be removed outright, since we're not specializing the ctor in any way:

```suggestion
```

warehouse/manage/views/__init__.py
```diff
@@ -1421,6 +1432,116 @@ def add_google_oidc_publisher(self):
...
+        )
+
+        self.request.session.flash(
+            f"Added {publisher} in {publisher.publisher_url()} to {self.project.name}",
+            queue="success",
+        )
+
+        self.metrics.increment(
+            "warehouse.oidc.add_publisher.ok", tags=["publisher:ActiveState"]
+        )
```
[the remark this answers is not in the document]
th3coop: There's been a couple of these "great but not right now" comments.  How do you all track things like this?  Should I file Enhancement request Issue reports in the Warehouse repo or do you have some other method of tracking possible future work?

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,209 @@
...
+            _ACTIVESTATE_GRAPHQL_API_URL,
+            json={
+                "query": query,
+                "variables": variables,
+            },
+            timeout=5,
+        )
+        if response.status_code >= 400:
+            sentry_sdk.capture_message(
+                f"Unexpected {response.status_code } error from ActiveState API: {response.content!r}"  # noqa
```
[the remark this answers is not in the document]
th3coop: This `noqa` can probably go away.  I'm fairly certain I had added it due to a much longer msg string.  I'll remove.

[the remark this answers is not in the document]
th3coop: Done.  Also reformatted the line to split the line up but keep it as one line when output.

th3coop:
Thanks so much for all your time @woodruffw !

woodruffw (approved):

th3coop:
@miketheman will you have time to review this PR this week?

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                raise wtforms.validators.ValidationError(
+                    _("ActiveState actor not found")
+                )
+
+        return _activestate_gql_api_call(
+            _GRAPHQL_GET_ACTOR, {"username": actor}, process_actor_response
+        )
+
+    def validate_actor(self, field):
```
miketheman, line 193: I was reading more about wtforms validators, and I _think_ that if there's a named validator function, it fires **before** the validators on the field, so `validate_actor()` would perform API calls before the other validators like regex/dashes are in play.

It's definitely worth re-verifying - since if our validators operate in that order, then we're performing an API call regardless. I hope I'm wrong! If that _is_ true, then it's not only this form - so I don't think it's immediately actionable, but if you can test it, that'd be great.

warehouse/accounts/views.py
```diff
@@ -1514,13 +1520,17 @@ def default_response(self):
         return {
             "pending_github_publisher_form": self.pending_github_publisher_form,
             "pending_google_publisher_form": self.pending_google_publisher_form,
+            "pending_activestate_publisher_form": self.pending_activestate_publisher_form,  # noqa
```
miketheman, line 1523: **suggestion:** use the code we're disabling so that we have a better idea ofwhy a rule was disabled.‏

```suggestion
            "pending_activestate_publisher_form": self.pending_activestate_publisher_form,  # noqa: E501
```

Similar note applies to other `noqa` instances.

miketheman (changes requested):
Thanks to the heavy lifting the other folks did before, this was relatively easy to review.

I had a few notes, questions, comments inline - the only real blocker are with the linter disables which should be an easy fix.

There's other questions and comments inline, feel free to address them as you wish.
The branch should also be rebased/merged once ready.

th3coop:
Thanks for the review, @miketheman!  Great notes in there.  I'll work through them either later today or tomorrow.  The GQL API is available again if you wanted to, or have time to take it for a manual test drive.  Again, sorry for that terrible timing.

warehouse/templates/manage/account/publishing.html
```diff
@@ -188,6 +188,77 @@
...
+    <div class="form-group">
+      <label for="actor" class="form-group__label">
+        {% trans %}Actor Username{% endtrans %}
+        {% if pending_activestate_pubisher_form.actor.flags.required %}
+        <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+        {% endif %}
+      </label>
+      {{ pending_activestate_pubisher_form.actor(placeholder=gettext("my-username"), class_="form-group__field", autocomplete="off", **{"aria-describedby":"actor-errors"}) }}
+      <p class="form-group__help-text">
+        {% trans %}The ActiveState user that will trigger the build of your Python artifact.{% endtrans %}
```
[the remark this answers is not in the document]
th3coop: Would it be more clear if I changed the wording slightly to something like:
 
> The username of the user that triggered the build in the ActiveState Platform that will publish your package
 
There is certainly contextual wording in the help text here but it's context I would expect someone that is configuring this to have.  eg. someone that uses the ActiveState Platform to manage their Python installs knows what it means for a user to trigger a build of their project, or even what a project is.

The ActiveState docs that explain the steps to build your package will elucidate this as well.

To actually answer your question: The only links I can think of that would clarify this help string would be a link to our "Getting Started" docs here: https://docs.activestate.com/platform/start/getting-started/

I think the help link you'd be interested would be much more focused than that.

warehouse/oidc/models/activestate.py
```diff
@@ -122,7 +122,7 @@ def publisher_name(self) -> str:
     def project(self) -> str:
         return self.activestate_project_name
 
-    def publisher_url(self, claims: SignedClaims | None = None) -> str:
+    def publisher_url(self, claims=None):
```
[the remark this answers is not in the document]
th3coop: Curious about why I removed the type declarations?  Or why claims are passed to this function at all?  

The former I did because, at the time, `oidc.models._core.OIDCPublisherMixin.publisher_url` didn't have the types.  Now that I look again though, they DO have them, though slightly different.  I'll bring them back.

The latter:  To fulfill the "interface" of `OIDCPublisherMixin` and override `pubslisher_url` there.  I could technically leave it out but we could run into issues some day if someone is iterating over publishers and calling their `publisher_url` and it fails on the ActiveState publisher because `publisher_url` only takes `self` and the calling code is trying to pass it `claims`.

I'll push a changes to bring back the typing.  Let me know if my explanation of including `claims` as a method arg makes sense.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,210 @@
...
+
+_VALID_ORG_URL_NAME_AND_ACTOR_NAME = re.compile(r"^[a-zA-Z0-9-]{3,40}$")
+_VALID_PROJECT_NAME = re.compile(r"^[.a-zA-Z0-9-]{3,40}$")
+_DOUBLE_DASHES = re.compile(r"--+")
+
+_ACTIVESTATE_GRAPHQL_API_URL = "https://platform.activestate.com/graphql/v1/graphql"
+_GRAPHQL_GET_ORGANIZATION = "query($orgname: String) {organizations(where: {display_name: {_eq: $orgname}}) {added}}"  # noqa
+_GRAPHQL_GET_ACTOR = (
+    "query($username: String) {users(where: {username: {_eq: $username}}) {user_id}}"
+)
```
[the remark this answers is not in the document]
th3coop: Looks like I was using `Try again in a few minutes`, like the GitHub form, before.  I'll bring that back in the contexts that make sense.

warehouse/oidc/forms/activestate.py
th3coop: I don't think I totally agree here.  I'm providing the granular regex's checks to provide a better user experience in this particular context but we could get away with "That organization does not exist".  Then the user would double check what they entered.

Maybe that's the better solution.  The user isn't trying to come up with an org name, they know the org.  They just need feedback that they entered the wrong thing, ie. "Not found"

So I would remove the regex checks on the format of the org name, then it falls to `_lookup_organization` then the empty list response is caught there.

What do you think @miketheman or @woodruffw ?

warehouse/oidc/forms/activestate.py
th3coop: If you like the idea here https://github.com/pypi/warehouse/pull/15168#discussion_r1480490211. I would do that change for this validator as well.  We really just need a "you entered the wrong text" kind of feedback here.

warehouse/accounts/views.py
th3coop: Done this in a local change.  Nice catch.  Didn't notice that used elsewhere in the code.  I like this much better!

warehouse/templates/manage/account/publishing.html
```diff
@@ -188,6 +188,77 @@
...
+    <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+    {{ form_errors(pending_activestate_pubisher_form) }}
+    <div class="form-group">
+      <label for="project_name" class="form-group__label">
+        {% trans %}PyPI Project Name{% endtrans %}
+        {% if pending_activestate_pubisher_form.project_name.flags.required %}
+        <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+        {% endif %}
+      </label>
+      {{ pending_activestate_pubisher_form.project_name(placeholder=gettext("project name"), autocomplete="off", autocapitalize="off", spellcheck="false", class_="form-group__field", aria_describedby="project_name-errors") }}
```
miketheman, line 203: I hate to do this, but I just merged #10003 that uses newlines to make these longer lines more readable. Would you mind reflowing these in the rest of the changed template code?

```suggestion
      {{ pending_activestate_pubisher_form.project_name(
          placeholder=gettext("project name"),
          autocomplete="off",
          autocapitalize="off",
          spellcheck="false",
          class_="form-group__field",
          aria_describedby="project_name-errors"
      ) }}
```

warehouse/templates/manage/account/publishing.html
```diff
@@ -188,6 +188,77 @@
...
+      </div>
+    </div>
+    <div class="form-group">
+      <label for="project" class="form-group__label">
+        {% trans %}ActiveState Project name{% endtrans %}
+        {% if pending_activestate_pubisher_form.project.flags.required %}
+        <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+        {% endif %}
+      </label>
+      {{ pending_activestate_pubisher_form.project(placeholder=gettext("my-project"), autocomplete="off", autocapitalize="off", spellcheck="false", class_="form-group__field", **{"aria-describedby":"project-errors"}) }}
```
miketheman, line 233: nit: move `aria-describedby` out of a splat at to a proper keyword `aria_describedby`

Similar note applies to the rest of the templates.

miketheman (changes requested):
@th3coop we're almost there! I dropped a couple of notes inline, should be straightforward changes.

Then a rebase and `make translations` update, and should be good to merge!

warehouse/templates/manage/account/publishing.html
th3coop: MIIIiiiIIiiIIIiIIiiIIiiIIIIIke (pouty face emoji and said with a whiny lilt)

th3coop: Kidding.  I prefer the new format muchly.  I'll do that for sure.

warehouse/manage/views/__init__.py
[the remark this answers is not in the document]
th3coop: @miketheman, I'm not sure if you're saying we should do that in this PR or should do that after this PR.  If it's the former, I'd prefer not to mix those changes into this PR.  If it's the latter, I'd be interested in doing that for your team with a follow up PR.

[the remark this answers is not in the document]
miketheman: Sorry for any confusion! Leave it alone for now.

warehouse/templates/manage/account/publishing.html
th3coop: oh wow....good eye!

miketheman:
> I'm leaning towards removing them though.

@th3coop go ahead and remove them - less is more 😁

th3coop:
> go ahead and remove them - less is more 😁

Agreed and done.  Fixed the tests and took it for a spin and it works as I'd expect it to

miketheman (changes requested):
Almost there! just a couple last bits, and rebase against `main`, and I think we're good to do.

warehouse/oidc/forms/activestate.py
```diff
@@ -0,0 +1,205 @@
...
+                message=_("Specify ActiveState project name")
+            ),
+            wtforms.validators.Regexp(
+                _VALID_PROJECT_NAME,
+                message=_("Invalid ActiveState project name"),
+            ),
+            wtforms.validators.Regexp(
+                _VALID_PROJECT_NAME,
+                message=_("Invalid ActiveState project name"),
+            ),
```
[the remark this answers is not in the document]
th3coop: Huh, those are both moderately embarrassing.  Sorry about those misses, Mike.

miketheman (approved):
Yay! I'll approve now, and merge in the AM.

th3coop:
> Yay! I'll approve now, and merge in the AM.

Hurray!  Thanks so much Mike, Will and Dustin!

miketheman:
@th3coop I've gone ahead and enabled the OIDC provider on TestPYI so you can test run it over there.

PR #15182, opened 2024-01-10
chore: remove most of basic auth code

### LINKED ISSUE #13770: Remove support for passwords in upload API

miketheman:
Now that the active policies are Session (for web) and Macaroon (for upload), remove the bulk of the BasicAuth policy.

Move the logic farther up the chain, so that there's no extra work put in before telling the user that they should migrate to API Tokens.

Resolves #13770


miketheman:
P.S. We can potentially also remove `commit_veto`, since nothing can raise the BasicAuth errors any longer.
https://github.com/pypi/warehouse/blob/52616573d872b66eceb040377f3815af4582e1d4/warehouse/config.py#L101-L111

I elected to defer until this is reviewed and approved, in case there's some other tricky reason to keep the `commit_veto`.

miketheman:
P.P.S. There's also some basic auth email sending code in the legacy upload endpoint that would be another deferral.

woodruffw (approved):
LGTM! Negative diffs are the best diffs 🙂

di:
I think we can also remove this: https://github.com/pypi/warehouse/blob/74d8013c6d76a4af702624ad0745f23c0a1233c1/warehouse/forklift/legacy.py#L996-L1013

miketheman:
> I think we can also remove this:

Agreed - I noted that I'm deferring that https://github.com/pypi/warehouse/pull/15182#issuecomment-1885840565

di:
Sorry, missed your P.P.S. 😂

di (approved):

PR #15203, opened 2024-01-12
Trusted Publishing: Google: The _check_sub verifier should check if empty

di:
Follow on to #15202: We should also short-circuit the verification when the ground truth is the empty string, not `None`.

### DIFF
--- a/warehouse/oidc/models/google.py
+++ b/warehouse/oidc/models/google.py
@@ -31,7 +31,7 @@ def _check_sub(
 ) -> bool:
     # If we haven't set a subject for the publisher, we don't need to check
     # this claim.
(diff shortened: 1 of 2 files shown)


warehouse/oidc/models/google.py
```diff
@@ -31,7 +31,7 @@ def _check_sub(
 ) -> bool:
     # If we haven't set a subject for the publisher, we don't need to check
     # this claim.
-    if ground_truth is None:
+    if ground_truth is "":
```
miketheman, line 34:
```suggestion
    if ground_truth == "":
```

miketheman (approved):

PR #15205, opened 2024-01-13
refactor: replace permission strings with enums

miketheman:
~~Not comprehensive!~~

Demonstrates using fine-grained permission strings that have more meaningful names than the principals they currently represent.

Currently converts all Admin-level strings to permissions, as a way to ease into this change before converting all others. If we break it, we break it for ourselves before others.


warehouse/admin/templates/admin/banners/list.html
```diff
@@ -22,7 +22,7 @@
 {% block content %}
 
 <div class="card card-primary card-outline">
-  {% if request.has_permission('psf_staff') %}
+  {% if request.has_permission(Permissions.AdminBannerWrite) %}
```
miketheman, line 25: This is probably the simplest example where we currently have a principal used in the permission string. Instead of using `"psf_staff"` as a permission, use a more accurate string `"admin:banner:write"`.

Thus we convey what is needed, instead of who we expect to have it - decoupling the conceptual group from the permission.

warehouse/config.py
```diff
@@ -60,12 +61,44 @@ class RootFactory:
...
+            Allow,
+            "group:psf_staff",
+            (
+                Permissions.AdminBannerRead,
+                Permissions.AdminBannerPreview,
+                Permissions.AdminBannerWrite,
+                Permissions.AdminSponsorsRead,
+                Permissions.AdminSponsorsWrite,
+            ),
+        ),
```
miketheman, line 101: There's a possibility that we overload ourselves with too many fine-grained permissions - I don't know how many is too many.

warehouse/authnz/_permissions.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Permissions can be specified in an ACL (`__acl__`) or `@view_config(permission=...)`
+    instead of using a string literal, minimizing the chance of typos.
+
+    They are also disconnected from Principals (users, groups, etc.), so they can be
+    used in a more generic way. For example, a permission could be used to allow a user
+    to edit their own profile, or to allow a group to edit any profile.
+
+    Naming should follow the format:
+        <optional scope>:<resource>:<action>
```
miketheman, line 26: After writing this, I'm realizing that after all is said and done, the actual strings themselves don't matter, as long as they are consistently used across the codebase.

If that holds true, then we can further simplify by using the StrEnum feature `= auto()` and not have to worry what the string is. We could likely change the format to `ScopeResourceAction = auto()`.

warehouse/authnz/__init__.py
miketheman, line 1: I considered an initial path of `warehouse.permissions.Permissions` but then figured that if I ever want other authN/Z stuff, this could be the general package for it.

ewdurbin (commented):
I don't see any drawbacks other than the callout on potential sprawl of fine grained permissions. As is, removing a bunch of duplicated strings across the codebase is strictly a benefit.

I'm +1 to moving forward with this comprehensively. It helps to model what is going on in my head at least :)

miketheman:
There's also some TODOs I left that I don't think are worth tacking right now, unless someone thinks otherwise and has a good solution.

ewdurbin (approved):

PR #15207, opened 2024-01-13
Miscellaneous OIDC fixes

di:
Some small things I found while working on testing the Google publisher.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1150,6 +1150,7 @@ def file_upload(request):
                 )
                 if request.oidc_publisher
                 else None,
+                "uploaded_via_trusted_publisher": str(bool(request.oidc_publisher)),
             },
         )
 
@@ -1425,6 +1426,7 @@ def file_upload(request):
                 if request.oidc_publisher
                 else None,
                 "project_id": str(project.id),
+                "uploaded_via_trusted_publisher": str(bool(request.oidc_publisher)),
             },
         )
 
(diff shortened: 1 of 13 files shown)


warehouse/forklift/legacy.py
```diff
@@ -1150,6 +1150,7 @@ def file_upload(request):
                 )
                 if request.oidc_publisher
                 else None,
+                "uploaded_via_trusted_publisher": str(bool(request.oidc_publisher)),
```
woodruffw, line 1153: I might be missing it, but what does this event attribute do that `publisher_url is not None` doesn't? The answer might just be "it's simpler", but just wanted to make sure I understood :slightly_smiling_face: 

Separately: instead of storing the strings `True` and `False` in the JSONB here, could this store bools instead? I think that'd result in slightly smaller rows + more idiomatic accesses (if it works).

woodruffw (approved):
LGTM overall, one question about the new event attribute!

warehouse/forklift/legacy.py
di: The Google publisher doesn't have a `publisher_url`, so it's always `None` which breaks a lot of logic where we assume having one means something is a publisher.

I'm open to alternatives here (I don't love the stringified bool) but this does work.

di: I tried bools at first, but the ORM didn't seem to like it.

woodruffw: I _think_ `.astext.cast(sqlalchemy.Boolean).is_(True)` in the comparison would work -- I can experiment with that!

PR #15225, opened 2024-01-16
Miscellaneous OIDC fixes (from #15207)

woodruffw:
This is #15207, plus an ORM simplification in https://github.com/pypi/warehouse/pull/15207#discussion_r1451222301 and an additional test to ensure the column typing works as we expect.


warehouse/packaging/models.py
```diff
@@ -717,7 +718,12 @@ def uploaded_via_trusted_publisher(self) -> bool:
         """Return True if the file was uploaded via a trusted publisher."""
         return (
             self.events.where(
-                self.Event.additional.op("->>")("publisher_url").is_not(None)  # type: ignore[attr-defined] # noqa E501
+                or_(
+                    self.Event.additional[  # type: ignore[attr-defined]
+                        "uploaded_via_trusted_publisher"
+                    ].as_boolean(),  # noqa E501
+                    self.Event.additional.op("->>")("publisher_url").is_not(None),  # type: ignore[attr-defined] # noqa E501
```
miketheman, line 725: I wonder if this should follow the `as_*()` approach, since then we're not asking for PG-specific operations, and following the new pattern for consistency.

```suggestion
                    self.Event.additional["publisher_url"].as_string().is_not(None)  # type: ignore[attr-defined] # noqa E501
```

warehouse/packaging/models.py
```diff
@@ -717,7 +718,12 @@ def uploaded_via_trusted_publisher(self) -> bool:
         """Return True if the file was uploaded via a trusted publisher."""
         return (
             self.events.where(
-                self.Event.additional.op("->>")("publisher_url").is_not(None)  # type: ignore[attr-defined] # noqa E501
+                or_(
+                    self.Event.additional[  # type: ignore[attr-defined]
+                        "uploaded_via_trusted_publisher"
+                    ].as_boolean(),  # noqa E501
```
miketheman, line 724: What happened here with E501? Wouldn't that have remained on a single line like so?

```suggestion
                    self.Event.additional["uploaded_via_trusted_publisher"].as_boolean(),  # type: ignore[attr-defined] # noqa E501
```

warehouse/templates/manage/project/history.html
```diff
@@ -41,7 +41,7 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
     </strong>
     <small>
       {% trans %}Added by:{% endtrans %}
-      {% if event.additional.publisher_url %}
+      {% if event.additional.uploaded_via_trusted_publisher or event.additional.publisher_url %}
```
miketheman, line 44: **thought (non-blocking):** ‏Seems like we ought to have one or the other, but not both. Probably only `uploaded_via_trusted_publisher` and do a data migration?

miketheman (commented):
Some comments inline - nothing particularly concerning, but an opportunity to gain consistency. Let me know if anything is unclear!

warehouse/forklift/legacy.py
```diff
@@ -1150,6 +1150,7 @@ def file_upload(request):
                 )
                 if request.oidc_publisher
                 else None,
+                "uploaded_via_trusted_publisher": bool(request.oidc_publisher),
```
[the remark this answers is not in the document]
woodruffw: Yeah, this isn't fantastic (and is the problem with JSON types in DBs -- it's too tempting to put yet another key in!). I don't have a better idea here, unfortunately, except maybe to offload this onto the `File` model itself.

Thoughts on that? In that case, we'd have something like `File.trusted_published` which we could both initialize for new uploads and backfill from events for old ones.

warehouse/packaging/models.py
woodruffw: Makes sense to me! I was having trouble with this one for brainfart reasons.

warehouse/packaging/models.py
woodruffw: Yeah, I believe it's now non-operative (since the line is below the `E501` limit). I *think* I can just remove the `noqa` entirely, unless we'd prefer to intentionally push the line over the lint limit?

warehouse/templates/manage/project/history.html
woodruffw: A minor hiccup here is that we want to render `publisher_url` in some scenarios (e.g., for GitHub, it'll point to the specific commit of the workflow that was run). So I think we need to preserve both JSON keys, unfortunately.

(Maybe there's a better way to do this, though?)

warehouse/templates/manage/project/history.html
```diff
@@ -237,9 +237,11 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
       {% trans %}Creator{% endtrans %}:
       {% if event.additional.publisher_url %}
       {{ event.additional.publisher_name }}:
-      <a href="{{ event.additional.publisher_url }}">
-        {{ event.additional.publisher_url }}
-      </a>
+        {% if event.additional.publisher_url %}
```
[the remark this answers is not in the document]
woodruffw: I think so! Will fix.

warehouse/packaging/models.py
woodruffw: I dropped the E501s here by reflowing.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
miketheman: Agreed - this was definitely a good placeholder while we learned about the utility of the data and how we would use it. `File.uploaded_via_trusted_publisher` grew out of me wanting to explore how it might be used - currently only in Admin UI via [`Release.trusted_published`](https://github.com/pypi/warehouse/blob/5466d3a4ae6f1b2d6c81f7ba91d313175d1d0233/warehouse/packaging/models.py#L636-L645).

A bool lookup ought to be faster, for sure.

I'm happy to proceed with this merge and get this going, if you wanted to work through a data migration and refactor apart, so we can benefit from these changes already. Your call!

miketheman (approved):
I'm fine with the open conversation direction. We can merge with these as they are now - say the word, and I'll press the button.

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
woodruffw: > I'm happy to proceed with this merge and get this going, if you wanted to work through a data migration and refactor apart, so we can benefit from these changes already. Your call!

Yep, that SGTM! I can work on that after this is merged.

warehouse/templates/manage/project/history.html
woodruffw: Yep, that summary is correct! We effectively have one old variable (`publisher_url`) that we also want to render, one new variable (`uploaded_via_trusted_publisher`) which is just a bool for checking, and the two only partially overlap 🙃 

A data migration for the older events makes sense to me. I can work on that in a follow-up.

woodruffw: (Or alternatively, the column addition would fix all of this.)

woodruffw:
> I'm fine with the open conversation direction. We can merge with these as they are now - say the word, and I'll press the button.

Same -- merge at your convenience!

Issue #14503, opened 2023-09-07
[DRAFT] PyPI Observation Reporting Payload


miketheman:
# Background
As part of a broader malware handling project, we’re looking at creating the first implementation by which Reporters may submit Observations about packages on PyPI.
(Specific terminology may change as we learn more.)

We’re still working out other parts of the API infrastructure, but wanted to get a conversation started on what the API **transaction payload** between Reporters and PyPI might look like.

As a refresher, our currently [security reporting process](https://pypi.org/security/) requests via email:
- A URL to the project in question
- An explanation of what makes the project a security issue
- If applicable: a link to the problematic lines in the project's distributions via [inspector.pypi.io](https://inspector.pypi.io/)

We have engaged a variety of security researchers and reporters who have been reporting malicious packages to PyPI Security to understand what aspects they prefer. We’ve learned more about their processes and use cases, and believe we’ve come up with something that is more automation-friendly, and leverages an evolving standard.

# Proposal

We’ve learned that there’s a general desire for more standards in the overall security ecosystem, and the [OSV project](https://osv.dev/) has defined a machine-friendly format for collecting published advisories.
The [OSV Schema 1.6.0](https://ossf.github.io/osv-schema/) is used for advisory databases.

While PyPI isn’t an advisory database, we thought using a format similar to OSV schema for an **inbound payload format** would be more sustainable long term, as we don’t invent our own standard, rather layer some extras on top of the existing one.

## Minimal Example

A Terse, Minimal Example, that expresses only the absolutely required keys:

```json
{
  "schema_version": "1.6.0+pypi",
  "modified": "2021-01-01T00:00:00Z",
  "summary": "During installation of pacakge, BitCoin miner installed and activated",
  "affected": [
    {
      "package": {
        "name": "request3",
        "ecosystem": "PyPI"
      },
      "versions": ["2.19.5"]
    }
  ],
  "references": [
    {
      "type": "INSPECTOR_URL",
      "url": "https://inspector.pypi.io/project/request3/2.19.5/..."
    }
  ]
}

```

## Changes from OSV schema

- remove `id`, as required top-level key
	- As a reporter, you are unlikely to have an ID yet when reporting. We could preserve the requirement, and allow any string, and discard it?
- add `schema_version`, `summary`, `affected`, `references` as required top-level keys
	- These already exist in the schema, we're making them required for transacting with our (future) API
- add some validations of:
	- at least 1 package
	- `ecosystem` is `PyPI`
[...]

PR #15228, opened 2024-01-18
feat: working POST API for Project Observations

miketheman:
Please review commit-by-commit to make reviewing easier.

Notes:

- There's a bunch of TODOs to highlight conversations or choices.
- Does have a `GET /danger-api/echo` endpoint
- Does have a `POST /danger-api/projects/{name}/observations` endpoint that validates the inbound payload (kinda sorta) and persists in the database
- Does **NOT** include the database migration to add the new column on the User table - I'll submit that as a separate PR if we want to proceed this way (it's a kludge). Here: https://github.com/pypi/warehouse/pull/15238
- Does **NOT** leverage Macaroon Caveats - so a User-scoped API Token from a user that `is_observer` will be allowed to POST. This is not a good long-term solution, but can likely work well enough to gather some data.
- ~~Does **NOT** include any tests yet - I wanted to get some structural approaches done here first.~~ Tests! But they rely on the migration PR being merged first.

The migration is basically:

```sql
ALTER TABLE users ADD COLUMN is_observer boolean NOT NULL DEFAULT false;
-- and then for fun:
UPDATE users SET is_observer = true WHERE username = '<MY USERNAME>';
```

A couple of `curl` commands to test. Provision yourself a User-scoped API Token:

```bash
curl -X GET \
    -H "Authorization: Bearer pypi-<API-TOKEN-HERE>" \
    -H "Content-Type: application/json" \
    http://localhost/danger-api/echo

curl -X POST \
    -H "Authorization: Bearer pypi-<API-TOKEN-HERE>" \
    -H "Content-Type: application/json" \
    -d '{"kind":"is_spam", "summary":"we found something wrong...", "extra_stuff": "not required or used yet"}' \
    http://localhost/danger-api/projects/foo/observations
```


warehouse/api/echo.py
```diff
@@ -0,0 +1,126 @@
...
+    return _wrapper
+
+
+@api_v0_view_config(
+    route_name="api.echo",
+    permission=Permissions.APIEcho,
+)
+def api_echo(request: Request):
+    return {
+        "accept": "I accept version 1 of the API",
```
ewdurbin, line 66: thought: use our view_config to bind an `api_version` string to the `Request` that would allow this to be

```
        "accept": f"I accept version {request.api_version} of the API",
```

warehouse/routes.py
```diff
@@ -539,6 +539,19 @@ def includeme(config):
         domain=warehouse,
     )
 
+    config.add_route(
+        "api.echo",
+        "/api/echo",
```
ewdurbin, line 544: thought: do we want to use `/api-unstable/` or something as a prefix?

its one more way for us to make clear that this is a no-commitment zone.

ewdurbin (commented):
This is a really nice and straightforward PR thanks to a lot of previous legwork. Thanks for that @miketheman.

Overall I will be confident in approving this for merge once we have a couple of things inplace:

* Sufficiently annoying/obvious signs to users that we are not committing to any API ergonomics or features. Think "danger" and "unstable" and "preview" and "dragons here" type verbiage in versions and or urls.
* A confidence in our malware observation payloads such that we feel comfortable that they will be actionable immediately, and not _too_ likely to be discarded or useless long term. I see that #14503 has some good updates on approaches here. Thanks for that.
* A side-effect to committing a malware observation that creates a conversation in the PyPI Security inbox. This ensures that any reporters that adopt the API won't have their reports "lost". It will integrate this effort into our current pipeline fairly seamlessly! I recommend attaching the observation as a `.json` attachment given...
* Some documentation that observations associated with deleted objects will also be deleted until we implement soft-deletes.

miketheman (commented):
Thanks for the review!

I've addressed the first couple of bullets so far, and will continue to work on the next two.

warehouse/routes.py
miketheman: `/danger-api/` is the prefix for now, and we'll swap it back when we're ready for v1.

miketheman:
This is pretty much ready to go, while I work out some more details in a follow up, unless reviewers see them as merge blockers:

- add danger-api endpoint for Releases
- add Observations and API to Files, Users

Keep in mind: Only superusers or folks we've set `User.is_observer` flag will be able to post with a user-scoped API Token, until we do more on the caveats/permissions front (soon).

warehouse/api/echo.py
```diff
@@ -0,0 +1,144 @@
...
+if typing.TYPE_CHECKING:
+    from pyramid.request import Request
+
+    from warehouse.packaging.models import Project
+
+
+# ### DANGER ZONE ### #
+# These views are a v0, danger, all-bets-are-off version of the API.
+# We may change the API at any time, and we may remove it entirely
+# without notice, but we'll try to tell the folks we know are testing
```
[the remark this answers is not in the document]
miketheman: Observations making more Observations, it's 🐢🐢🐢 all the way down!

We discussed this and decided it's a cool idea, but not a blocker for this change.

warehouse/observations/tasks.py
```diff
@@ -0,0 +1,146 @@
...
+    # Fetch the observations from the session.
+    observations = session.info.pop("warehouse.observations.new", set())
+    for obj in observations:
+        # We pass the ID of the Observation, not the Observation itself,
+        #  because the Observation object is not currently JSON-serializable.
+        config.task(report_observation_to_helpscout).delay(obj.id)
+
+
+@tasks.task(bind=True, ignore_result=True, acks_late=True)
+def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> None:
```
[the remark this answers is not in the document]
miketheman: A simple bail-out has been added.

ewdurbin (approved):

PR #15235, opened 2024-01-18
fix: use correct constraint on Observations

### LINKED ISSUE #15227: Deleting a record with Observations raises exception
**Describe the bug**
When a Project or Release has an associated Observation record, trying to remove the parent raises an exception.

```pycon
sqlalchemy.exc.IntegrityError: (psycopg.errors.NotNullViolation) null value in column "related_id" of relation "project_observations" violates not-null constraint
DETAIL:  Failing row contains (null, 46baa7c4-e218-4513-9e55-cb8324e68b6c, 2024-01-16 18:36:10.998765, is_malicious, asdfasdf, {}, ed662f70-daa9-4ea4-bc98-10609950955b).
[SQL: UPDATE project_observations SET related_id=%(related_id)s::UUID WHERE project_observations.id = %(project_observations_id)s::UUID]
[parameters: [{'related_id': None, 'project_observations_id': UUID('ed662f70-daa9-4ea4-bc98-10609950955b')}, {'related_id': None, 'project_observations_id': UUID('f561848c-f990-4c16-891d-cdf3ff1f85dd')}, {'related_id': None, 'project_observations_id': UUID('f9444b22-41e0-41a7-ab59-5ac3c86979df')}]]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```

**Expected behavior**
The project record should be removed, and no exception should be raised.
The related Observation records should be removed as well to prevent orphans.

**To Reproduce**
- Using local dev, log in as an admin
- Add an Observation to a project
- Delete the project

This should definitely get resolved before we add more ways to create Observations.

miketheman:
Fixes #15227

### DIFF
--- a/warehouse/migrations/versions/812e14a4cddf_cascade_observation_delete.py
+++ b/warehouse/migrations/versions/812e14a4cddf_cascade_observation_delete.py
@@ -1,0 +1,81 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+cascade observation delete
+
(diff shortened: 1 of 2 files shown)


ewdurbin (approved):
half of me wants to just denormalize the reference so the observation can live on... but the other just wants soft-deletes.

miketheman:
I hear you on the soft-deletes for sure. But until then, at least we won't be blocking ourselves if we remove things.

Issue #15224, opened 2024-01-16
The link to Python Packaging User Guide is currently broken


willmcgugan:
The link to Python Packaging User Guide is currently 404ing.

https://pypi.org/manage/project/PROJECT/releases/

<img width="1158" alt="Screenshot 2024-01-16 at 11 03 07" src="https://github.com/pypi/warehouse/assets/554369/fb159697-2266-4950-b24e-cff4e88db6b4">

woodruffw:
Confirmed that this is broken (this only appears when a project has no releases, which is pretty rare).

Bugpoint: 

https://github.com/pypi/warehouse/blob/30c8720951db492dbd800bb5433edbdf1fc6085f/warehouse/templates/manage/project/releases.html#L179

The correct link is probably this one: <https://packaging.python.org/en/latest/guides/section-build-and-publish/>

woodruffw:
@willmcgugan would you be willing to do a quick PR for this? Otherwise, I'll try and file one later today 🙂

woodruffw:
Doing this now.

JohannesEbke:
> Confirmed that this is broken (this only appears when a project has no releases, which is pretty rare).

Maybe this is unnecessary, but I came here to report the same issue; and I got the link in a project with three releases: [https://pypi.org/manage/project/aws-close-account/releases/](https://pypi.org/manage/project/aws-close-account/releases/)

I do appreciate the link being there, though!

PR #15236, opened 2024-01-18
docs/packaging: fix the link to correct page

davidbgk:
https://packaging.python.org/tutorials/distributing-packages/ returns a 404

### DIFF
--- a/docs/dev/api-reference/legacy.rst
+++ b/docs/dev/api-reference/legacy.rst
@@ -120,7 +120,7 @@ The API endpoint served at `upload.pypi.org/legacy/
 <https://upload.pypi.org/legacy/>`_ is Warehouse's emulation of the
 legacy PyPI upload API. This is the endpoint that tools such as `twine
 <https://twine.readthedocs.io/>`_ use to `upload distributions to PyPI
-<https://packaging.python.org/tutorials/distributing-packages/>`_.
+<https://packaging.python.org/guides/distributing-packages-using-setuptools/>`_.
 
 The upload api can be used to upload artifacts by sending a multipart/form-data
 POST request with the following fields:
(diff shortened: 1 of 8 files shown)


woodruffw (approved):
LGTM in terms of fixing the links; two thoughts:

1. Maybe we need a linkchecker or similar integration test to catch these link breakages proactively?
2. We should confirm that this is the link we want to use: it looks like the appropriate successor page, but it also uses `setup.py` instead of a more modern packaging flow (with e.g. `pyproject.toml`). But maybe that can be a follow-up, since it's good to just fix the link for now.

woodruffw:
@davidbgk Thanks for the fixes here! Could you bring this branch up-to-date with `main`?

miketheman:
> * Maybe we need a linkchecker or similar integration test to catch these link breakages proactively?

Good idea - `make -C docs/dev/ linkcheck` from the repo root runs Sphinx's built-in link checker. We already have a directive for links we expressly do NOT want to check, see here:
https://github.com/pypi/warehouse/blob/55476f39266473e49d21a3a8e64b7cda4d71fdbf/docs/dev/conf.py#L56-L61

miketheman:
Another general thought specific to the packaging docs site is that they also use [interspinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html), so in theoiry we could as well. Here's an example config from twine: https://github.com/pypa/twine/blob/fe1885f2bf896c1852dedea7733a582c5718bbbc/docs/conf.py#L279-L282
and an example of usage: https://github.com/pypa/twine/blob/fe1885f2bf896c1852dedea7733a582c5718bbbc/docs/index.rst?plain=1#L180

If we use intersphinx references, and the remote **references** remain stable, then as URLs change our would update as well.

woodruffw:
This is now a subset of #15265 (sorry for the confusion here -- I missed that #15224 was essentially a duplicate report).

woodruffw:
Done with #15265, sorry again for the confusion here!

davidbgk:
Not super contributor friendly but glad it's fixed 👍

PR #15242, opened 2024-01-21
Record the attached caveats during Macaroon creation

dstufft:
We currently have `Macaroon.permission_caveat` which originally recorded the "v1" permission caveat that the Macaroon was created with, to allow differentiating between a "user scoped" macaroon and a "project scoped" macaroon.

This was "OK" at first, but in the interim things have changed:

- We no longer emit these now legacy "v1" permission caveats, but we instead fake a structure that looks like they used to look to continue to allow us to determine if a Macaroon is user or project scoped. While this works, there's a risk that in the future our caveats don't map 1:1 back to this old permission caveat and we fail to record the true scope of the macaroon.
- We have more caveats than just the old v1 permission caveat (and adding more: https://github.com/pypi/warehouse/pull/15240), which also affect the scope of the macaroon that may be worthwhile to present to the user or to interpret for other reasons (for instance, we may want to delete Macaroons that have expired, or set their secret key to null or something).

So what this PR does is rather than try and decide which caveats are important enough to save or not, we just stash all of the caveats that were attached to the Macaroon when we created it. This means that if at any point in the future we decide that the information contained in some caveat is useful to know, we always have it at hand even for previously generated macaroons.

Much like the existing `permissions_caveat`, this is not, and should not be used as an authoritative source, the ones in the Macaroon should still be used for that, but it can be used for informational purposes to know what the scope is of this macaroon.

We reuse the (de)serialization framework that we use for embedding our caveats into the Macaroon itself to convert our caveats into a JSON compatible data structure that can be saved into a PostgreSQL `jsonb` column, which affords us the full querying power of PostgreSQL if the need arises (though these are the _serialized_ form, so it optimizes for space efficiency not for readability, so these queries may not be the nicest looking queries-- but it's possible and will work well).

We attempt to back-fill this data for existing Macaroons using the `permissions_caveat` data, however the back-fill is going to be imperfect (for instance the implicit `ProjectID` caveat wasn't recorded in any way), but there's not really a better option available to us.

We don't currently use this information anywhere and we're just recording it, but ideally in the future we can stop using the `Macaroons.permissions_caveat` and drop it and shift to using `Macaroons.caveats` and pulling out the information that is most useful to us for whatever we're doing. Doing that would allow us to drop one of the places we're synthesizing the legacy v1 caveat for informational purposes.


warehouse/migrations/versions/be62a4cd76e3_add_a_column_to_store_the_entire_set_of_.py
```diff
@@ -0,0 +1,66 @@
...
+    # is the attached user_id for this Macaroon.
+    op.execute(
+        """ UPDATE macaroons
+            SET caveats = jsonb_build_array(jsonb_build_array(3, user_id::text))
+            WHERE
+                caveats IS NULL
+                AND user_id IS NOT NULL
+                AND permissions_caveat->>'permissions' = 'user'
+        """
+    )
```
miketheman, line 46: This is going to hit some ~80k rows in production.
For a SELECT to find them, the planner chose to do a Parallel Seq Scan that takes ~320ms to select (without the new `caveats` column, since it doesn't exist yet).

```sql
warehouse=> explain analyze select * from macaroons where user_id is not null AND permissions_caveat->>'permissions' = 'user';
                                                          QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------
 Gather  (cost=1000.00..7224.11 rows=725 width=222) (actual time=1.364..313.752 rows=88549 loops=1)
   Workers Planned: 2
   Workers Launched: 2
   ->  Parallel Seq Scan on macaroons  (cost=0.00..6151.61 rows=302 width=222) (actual time=0.414..297.411 rows=29516 loops=3)
         Filter: ((user_id IS NOT NULL) AND ((permissions_caveat ->> 'permissions'::text) = 'user'::text))
         Rows Removed by Filter: 30226
 Planning Time: 0.497 ms
 Execution Time: 319.197 ms
(8 rows)
```

I suspect that both the UPDATEs will happen in under our statement timeout of 4 seconds, so I wanted to note the timing I saw in case we hit a migration problem.

warehouse/macaroons/models.py
```diff
@@ -71,6 +76,38 @@ class Macaroon(db.Model):
...
+    #       those additional caveats to us, so those would not be reflected in
+    #       this data.
+    #
+    # NOTE: Originally we only stored the "permissions" caveat, which maps to
+    #       the RequestUser and ProjectName caveat, while that data was mapped
+    #       into these caveats, there may have been additional caveats (such as
+    #       expiration) that were attached to the Macaroon during generation
+    #       that is missing from this column for older Macaroons.
+    #
+    _caveats: Mapped[list | None] = mapped_column("caveats", JSONB)
```
miketheman, line 96: There's something bothering me about this data type, and I'm not sure why, so let me write out some thoughts.

While a `json(b)`-type columns can absolutely contain a list instead of a dict, to date we've used only dict (json object). This sets the column to be a list (json array), and in my head that would be a postgres `jsonb[]` data type instead - as that would constrain the type to only ever allow arrays of json objects at the DB level. 

I think the right way to express that is:

```python
from sqlalchemy.dialects.postgresql import ARRAY

_caveats: Mapped[list | None] = mapped_column("caveats", ARRAY(JSONB))
```
and the resulting migration gives us a db-level `jsonb[]` type.

I don't know if there's DB-space efficiency or index/query time efficiency, but I think having a column type constrained to what is expected to be there is better.

There's an imperfect example showing some differences here: https://stackoverflow.com/a/56625335/244037 (reference is from 2019, things may have changed a bit, so worth revalidating).

But if that materially changes the query-time ergonomics, or makes things even more confusing, then I'm happy to pass on this.

miketheman (commented):
Really liking the direction this is taking, especially with the other PR changes we want.

💯 👍 on replacing `permissions_caveat` with the caveats this Macaroon was initially serialized with.
Might be a good idea to take some of the `NOTE` details and put in `_caveats` `mapped_column(comment="...")` so we're super clear when looking at the DB itself, or using tools that extract DB metadata (a la DBML).

Other than the outstanding question about the jsonb column type, this looks pretty awesome!

warehouse/macaroons/models.py
dstufft: I considered doing `jsonb[]` but ended up not doing that because it means that to query the column you need to mix PostgreSQL array functions with jsonb functions, and I thought that the combination of those two would be more confusing than just using a `jsonb` as an array.

I'm happy to swap this to a `jsonb[]` though if people prefer, I don't really have a preference one way or another, I was on the fence and just sorta went with what felt right.

dstufft: I've gone ahead and done this.

dstufft: Actually, I'm going to undo it because it makes SQLAlchemy barf when you have multiple caveats.

The problem is that SQLAlchemy ends up generating SQL like this:

```
 "{{4,"\"7804c7db-c87d-4fee-a254-e0baafd06c59\"","null"},{2,"[\"2767a5b8-85f6-4bf4-9adc-6ed507df6c57\"]"},{0,900,0}}"
 ```

And PostgreSQL sees that as a multi dimensional array, not as an array of JSONB.

**Comment by dstufft** [comment]
Ok, this should be r4r now

File: warehouse/macaroons/models.py (hunk 2 above)
Comment (miketheman) [reply warehouse/macaroons/models.py]: Okay, that's icky. I wonder if there's a better way to express the concept in SQLAlchemy to have the queries make more sense? In any case, no need to pursue this further - thanks for trying it out!

File: warehouse/migrations/versions/be62a4cd76e3_add_a_column_to_store_the_entire_set_of_.py [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+                AND jsonb_typeof(
+                    permissions_caveat->'permissions'->'projects'
+                ) = 'array'
+        """
+    )
+
+    # Set our column to not nullable
+    op.alter_column(
+        "macaroons", "caveats", server_default=sa.text("'{}'"), nullable=False
+    )
```
miketheman, line 80: This is another part of the migration that I'd be a little cautious about. There's ~180k records in this table that will be forced to run a check constraint which takes an `ACCESS EXCLUSIVE` lock on the table.
Since this table's current major driver of activity is Trusted Publishers via CI (on the order of a couple minute), I don't think this will block this migration, but somethign we should keep an eye on.

miketheman (approved):
Nice work, let's go!

PR #15258, opened 2024-01-24
Improve error msg when creating pending publisher for a user's existing project

### LINKED ISSUE #14232: Improve PendingPublisher UX when project already exists
Currently, when an owner attempts to create a "Pending Publisher" for a project that already exists and that they already own, they get the same error as when they _don't_ own the project:

https://github.com/pypi/warehouse/blob/97ff3692081c621c8ccc4906c6d0e228bd5225ae/warehouse/oidc/forms.py#L190

Ideally, we would just create a regular publisher for them instead, or at the very least provide a more meaningful error message in this case.

---

_Original issue below:_

It currently isn't possible to have a separate Trusted Publisher workflow for development and official workflows because more than one workflow is not permitted from the same project repository.

I have a Github Actions workflow `publish_package_dev` which publishes alpha and beta releases to `test.pypi.org`, and another workflow `publish_package` which publishes non development tagged releases to `pypi.org`.

However, when setting up a Trusted Publisher workflow for the test repository then it is no longer possible to add a Trusted Publisher for the non-test version of PyPi because it complains that "This project name is already in use".

tetsuo-cpp:
Closes #14232

CC: @woodruffw


warehouse/oidc/forms/_core.py
```diff
@@ -30,10 +30,21 @@ class PendingPublisherMixin:
...
+            project = self._project_factory[project_name]
+        except KeyError:
+            # If the project doesn't exist, we're ok to proceed
+            return
+
+        if self._current_user in project.owners:
+            error_msg = _(
+                "Project already exists, create an ordinary "
+                "trusted publisher instead"
             )
```
woodruffw, line 43: Error message LGTM, but it's a bit unfortunate that we have to plumb the current user all the way into the form logic like this.

Thinking about this a bit: maybe rather than specializing on `user in project.owners`, we can just unconditionally spit out this message. That'll make it slightly wrong for users who don't have permission on the requested project, but that should be a pretty rare case anyways (and will avoid the need to check the current user).

tetsuo-cpp: Sure, I don't mind either way. That would turn this PR into a one-liner more or less.

Does anyone else have opinions on it?

woodruffw:
I'm going to re-do this PR, per https://github.com/pypi/warehouse/pull/15258#discussion_r1470185089.

Issue #13575, opened 2023-05-04
Trusted publishing: support for GitLab CI


woodruffw:
Per https://github.com/pypi/warehouse/issues/12465#issuecomment-1532965483: GitLab now supports a customizable `aud` for their CI-issued identity tokens, meaning that it should be possible to integrate them as a provider of trusted publishers!

Some initial tasks:

- [x] Extract some GitLab OIDC tokens and inspect their claim set
- [x] Determine the appropriate set of fields/claim constraints to expose to users

From there, the actual development tasks on this should look similar to the tasks enumerated in #13551.

facutuesca:
I'm taking a look at this

woodruffw:
Thanks @facutuesca! I've assigned you.

facutuesca:
>  Extract some GitLab OIDC tokens and inspect their claim set

Here's one:
```json
{
  "namespace_id": "12885807",
  "namespace_path": "user",
  "project_id": "53804090",
  "project_path": "user/oidc_test",
  "user_id": "9420787",
  "user_login": "user",
  "user_email": "user@example.com",
  "pipeline_id": "1136338956",
  "pipeline_source": "push",
  "job_id": "5919171663",
  "ref": "main",
  "ref_type": "branch",
  "ref_path": "refs/heads/main",
  "ref_protected": "true",
  "runner_id": 12270807,
  "runner_environment": "gitlab-hosted",
  "sha": "0a037033c5d88a5e5fd4d0658740c7a952b4f317",
  "project_visibility": "private",
  "ci_config_ref_uri": "gitlab.com/user/oidc_test//.gitlab-ci.yml@refs/heads/main",
  "ci_config_sha": "0a037033c5d88a5e5fd4d0658740c7a952b4f317",
  "jti": "1a78bc43-842b-42bc-994c-25c12124af5e",
  "iss": "https://gitlab.com",
  "iat": 1705066514,
  "nbf": 1705066509,
  "exp": 1705070114,
  "sub": "project_path:user/oidc_test:ref_type:branch:ref:main",
  "aud": "pypi"
}
```

The fields are documented in https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload. 

Since the `iss` will be whatever the URL of the GitLab instance is, I'm assuming that for this provider we're not supporting self-hosted instances, right? That is, for now only `"iss": "https://gitlab.com"` will be supported

[...]

PR #15275, opened 2024-01-25
Initial implementation of GitLab OIDC trusted publisher

facutuesca:
## Description
This PR includes the initial implementation of the GitLab trusted publisher. It's mostly based on the existing GitHub implementation, with a few modifications (mentioned below). Unit tests have been included, and a separate PR for the docs is available [here](https://github.com/pypi/warehouse/pull/15283). This is part of [Trusted publishing: support for GitLab CI](https://github.com/pypi/warehouse/issues/13575).

The corresponding GitLab feature (along with all the claims provided in the token) is documented [here](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html).

## Differences with GitHub OIDC
- GitLab uses the term "project" for a repository.
- While GitHub uses the term "username" or "organization" for the `org` part of `github.com/org/repo`, GitLab uses the term "namespace". A namespace can be either a username or a group. Groups can be nested, and so for `https://gitlab.com/my_org/my_suborg/my_repo`, the namespace would be `my_org/my_suborg`.
- For GitHub publishers, PyPI checks that the username/organization provided exists. For GitLab we don't do this, since namespaces can be private and therefore we can't check if they exist or not. See [private groups](https://docs.gitlab.com/ee/user/public_access.html#private-projects-and-groups).
- The above makes this provider less secure that GitHub's, in the sense that it doesn't protect against account resurrection attacks. This is mentioned in the `security-model.md` doc.
- Environments are case insensitive in GitHub, but are case sensitive in GitLab.
- Workflow files (e.g: `workflow.yml`) can be anywhere in the GitLab repo. Since the OIDC claims include the path of the workflow relative to the root of the repo, when setting up a GitLab publisher we also ask for this path (as opposed to GitHub, where we only ask for the file name).

## TODO
- [ ] Once final changes are made, add the screenshots missing to the docs

## Screenshot
image

### DIFF
--- a/warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
+++ b/warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
@@ -13,7 +13,7 @@
 Add GitLab OIDC models
 
 Revision ID: 81f9f9a60270
-Revises: be62a4cd76e3
+Revises: 4d1b4fcc4076
 Create Date: 2024-01-16 17:46:16.443395
 """
 
@@ -22,7 +22,7 @@ import sqlalchemy as sa
 from alembic import op
 
 revision = "81f9f9a60270"
-down_revision = "be62a4cd76e3"
+down_revision = "4d1b4fcc4076"
 
 
 def upgrade():
@@ -32,7 +32,7 @@ def upgrade():
         VALUES (
             'disallow-gitlab-oidc',
             'Disallow the GitLab OIDC provider',
-            FALSE,
+            TRUE,
             FALSE
         )
         """
--- a/warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
+++ b/warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
@@ -1,0 +1,85 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add GitLab OIDC models
+
+Revision ID: 81f9f9a60270
+Revises: be62a4cd76e3
+Create Date: 2024-01-16 17:46:16.443395
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "81f9f9a60270"
+down_revision = "be62a4cd76e3"
+
+
+def upgrade():
+    op.execute(
+        """
+        INSERT INTO admin_flags(id, description, enabled, notify)
+        VALUES (
+            'disallow-gitlab-oidc',
+            'Disallow the GitLab OIDC provider',
+            FALSE,
+            FALSE
+        )
+        """
+    )
+    op.create_table(
+        "gitlab_oidc_publishers",
+        sa.Column("id", sa.UUID(), nullable=False),
+        sa.Column("namespace", sa.String(), nullable=False),
+        sa.Column("project", sa.String(), nullable=False),
+        sa.Column("workflow_filepath", sa.String(), nullable=False),
+        sa.Column("environment", sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["id"],
+            ["oidc_publishers.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "namespace",
+            "project",
+            "workflow_filepath",
+            "environment",
+            name="_gitlab_oidc_publisher_uc",
+        ),
+    )
+    op.create_table(
+        "pending_gitlab_oidc_publishers",
+        sa.Column("id", sa.UUID(), nullable=False),
+        sa.Column("namespace", sa.String(), nullable=False),
+        sa.Column("project", sa.String(), nullable=False),
+        sa.Column("workflow_filepath", sa.String(), nullable=False),
+        sa.Column("environment", sa.String(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["id"],
+            ["pending_oidc_publishers.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "namespace",
+            "project",
+            "workflow_filepath",
+            "environment",
+            name="_pending_gitlab_oidc_publisher_uc",
+        ),
+    )
+
+
+def downgrade():
+    op.execute("DELETE FROM admin_flags WHERE id = 'disallow-gitlab-oidc'")
+    op.drop_table("pending_gitlab_oidc_publishers")
+    op.drop_table("gitlab_oidc_publishers")
--- a/warehouse/oidc/forms/gitlab.py
+++ b/warehouse/oidc/forms/gitlab.py
@@ -18,11 +18,9 @@ from warehouse import forms
 from warehouse.i18n import localize as _
 from warehouse.oidc.forms._core import PendingPublisherMixin
 
-# TODO Gitlab also accepts emojis
-# see:
 # https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-project-and-group-names
-_VALID_GITLAB_PROJECT = re.compile(r"^[a-zA-Z0-9_][a-zA-Z0-9-_.+\- ]*$")
-_VALID_GITLAB_NAMESPACE = re.compile(r"^[a-zA-Z0-9_][a-zA-Z0-9_.()\- /]*$")
+_VALID_GITLAB_PROJECT = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]*$")
+_VALID_GITLAB_NAMESPACE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]*$")
 _VALID_GITLAB_ENVIRONMENT = re.compile(r"^[a-zA-Z0-9\-_/${} ]+$")
 
 
--- a/warehouse/oidc/forms/gitlab.py
+++ b/warehouse/oidc/forms/gitlab.py
@@ -1,0 +1,106 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+import wtforms
+
+from warehouse import forms
+from warehouse.i18n import localize as _
+from warehouse.oidc.forms._core import PendingPublisherMixin
+
+# TODO Gitlab also accepts emojis
+# see:
+# https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-project-and-group-names
+_VALID_GITLAB_PROJECT = re.compile(r"^[a-zA-Z0-9_][a-zA-Z0-9-_.+\- ]*$")
+_VALID_GITLAB_NAMESPACE = re.compile(r"^[a-zA-Z0-9_][a-zA-Z0-9_.()\- /]*$")
+_VALID_GITLAB_ENVIRONMENT = re.compile(r"^[a-zA-Z0-9\-_/${} ]+$")
+
+
+class GitLabPublisherBase(forms.Form):
+    __params__ = ["namespace", "project", "workflow_filepath", "environment"]
+
+    namespace = wtforms.StringField(
+        validators=[
+            wtforms.validators.InputRequired(
+                message=_("Specify GitLab namespace (username or group/subgroup)"),
+            ),
+            wtforms.validators.Regexp(
+                _VALID_GITLAB_NAMESPACE,
+                message=_("Invalid GitLab username or group/subgroup name."),
+            ),
+        ]
+    )
+
+    project = wtforms.StringField(
+        validators=[
+            wtforms.validators.InputRequired(message=_("Specify project name")),
+            wtforms.validators.Regexp(
+                _VALID_GITLAB_PROJECT, message=_("Invalid project name")
+            ),
+        ]
+    )
+
+    workflow_filepath = wtforms.StringField(
+        validators=[
+            wtforms.validators.InputRequired(message=_("Specify workflow filepath"))
+        ]
+    )
+
+    environment = wtforms.StringField(
+        validators=[
+            wtforms.validators.Optional(),
+            wtforms.validators.Regexp(
+                _VALID_GITLAB_ENVIRONMENT, message=_("Invalid environment name")
+            ),
+        ]
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def validate_workflow_filepath(self, field):
+        workflow_filepath = field.data
+
+        if not (
+            workflow_filepath.endswith(".yml") or workflow_filepath.endswith(".yaml")
+        ):
+            raise wtforms.validators.ValidationError(
+                _("Workflow file path must end with .yml or .yaml")
+            )
+        if workflow_filepath.startswith("/") or workflow_filepath.endswith("/"):
+            raise wtforms.validators.ValidationError(
+                _("Workflow file path cannot start or end with /")
+            )
+
+    @property
+    def normalized_environment(self):
+        # NOTE: We explicitly do not compare `self.environment.data` to None,
+        # since it might also be falsey via an empty string (or might be
+        # only whitespace, which we also treat as a None case).
+        return (
+            self.environment.data
+            if self.environment.data and not self.environment.data.isspace()
+            else ""
+        )
+
+
+class PendingGitLabPublisherForm(GitLabPublisherBase, PendingPublisherMixin):
+    __params__ = GitLabPublisherBase.__params__ + ["project_name"]
+
+    def __init__(self, *args, project_factory, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._project_factory = project_factory
+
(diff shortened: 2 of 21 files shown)


warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,111 @@
...
+#       prevent situations where a slow migration takes the entire site down.
+#
+#       If you need to increase this timeout for a migration, you can do so
+#       by adding:
+#
+#           op.execute("SET statement_timeout = 5000")
+#           op.execute("SET lock_timeout = 4000")
+#
+#       To whatever values are reasonable for this migration as part of your
+#       migration.
```
woodruffw, line 47: Nitpick: safe to remove 🙂 

```suggestion
```

docs/user/trusted-publishers/adding-a-publisher.md
```diff
@@ -43,6 +43,26 @@ each.
         your trusted workflow, such as requiring manual approval on each run
         by a trusted subset of repository maintainers.
 
+=== "GitLab CI/CD"
```
woodruffw, line 46: I *think* we want to exclude this from this PR, and instead bundle it with https://github.com/pypi/warehouse/pull/15192 -- that way we can make these docs public on the same "flag" day as the other publishers being added.

woodruffw: (And ditto for all the other doc pages -- I'll see if I can give you push rights to that PR and, if not, I'll re-push it from the ToB fork.)

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,111 @@
...
+        INSERT INTO admin_flags(id, description, enabled, notify)
+        VALUES (
+            'disallow-gitlab-oidc',
+            'Disallow the GitLab OIDC provider',
+            FALSE,
+            FALSE
+        )
+        """
+    )
+    # ### commands auto generated by Alembic - please adjust! ###
```
woodruffw, line 62:
```suggestion
```

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,111 @@
...
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint(
+            "namespace",
+            "project",
+            "workflow_filepath",
+            "environment",
+            name="_pending_gitlab_oidc_publisher_uc",
+        ),
+    )
+    # ### end Alembic commands ###
```
woodruffw, line 103:
```suggestion
```

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def downgrade():
+    op.execute("DELETE FROM admin_flags WHERE id = 'disallow-gitlab-oidc'")
+    # ### commands auto generated by Alembic - please adjust! ###
```
woodruffw, line 108:
```suggestion
```

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,111 @@
...
+    op.drop_table("pending_gitlab_oidc_publishers")
+    op.drop_table("gitlab_oidc_publishers")
+    # ### end Alembic commands ###
```
woodruffw, line 111:
```suggestion
```

warehouse/oidc/forms/gitlab.py
```diff
@@ -0,0 +1,111 @@
...
+
+import wtforms
+
+from warehouse import forms
+from warehouse.i18n import localize as _
+from warehouse.oidc.forms._core import PendingPublisherMixin
+
+# TODO Gitlab also accepts emojis
+# see:
+# https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-project-and-group-names
```
woodruffw, line 23: Oof. I think we can either say "emoji are not supported, regardless of what GitLab lets you do" or use a property escape like `\p{Emoji_Presentation}`, although I don't know if Python's `re` supports that.

But probably just the first idea 🙂

warehouse/oidc/forms/gitlab.py
```diff
@@ -0,0 +1,111 @@
...
+        super().__init__(*args, **kwargs)
+
+    def validate_namespace(self, field):
+        namespace = field.data
+
+        # We pre-filter namespace with a regex
+        if not _VALID_GITLAB_NAMESPACE.match(namespace):
+            raise wtforms.validators.ValidationError(
+                _("Invalid GitLab username or group/subgroup name.")
+            )
```
woodruffw, line 74: Can this be an inline `validators=[]` member instead? Might be missing something.

warehouse/oidc/models/gitlab.py
```diff
@@ -0,0 +1,288 @@
...
+    CheckClaimCallable,
+    OIDCPublisher,
+    PendingOIDCPublisher,
+    check_claim_binary,
+)
+
+
+def _check_ci_config_ref_uri(ground_truth, signed_claim, all_signed_claims):
+    # We expect a string formatted as follows:
+    #   gitlab.com/OWNER/REPO//WORKFLOW_PATH/WORKFLOW_FILE.yml@REF
```
woodruffw, line 32: Is `REPO//WORKFLOW_PATH` right here?

woodruffw (commented):
This is great! I did a very fast initial pass; will do a more intensive review after.

docs/user/trusted-publishers/adding-a-publisher.md
facutuesca: I created a PR for the docs, with the base branch being the PR you mentioned: https://github.com/pypi/warehouse/pull/15283

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
facutuesca: fixed

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
facutuesca: fixed

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
facutuesca: fixed

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
facutuesca: fixed

warehouse/oidc/forms/gitlab.py
facutuesca: Yeah I'd also go with not supporting emojis for now. We can always add support for them later if needed

warehouse/oidc/forms/gitlab.py
facutuesca: yup, you're right, the separate function was a leftover from how GitHub does it. Fixed.

warehouse/oidc/models/gitlab.py
facutuesca: Yup, from the [`ci_config_ref_uri`](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload) field docs:

> The ref path to the top-level pipeline definition, for example, gitlab.example.com/my-group/my-project//.gitlab-ci.yml@refs/heads/main. [Introduced](https://gitlab.com/gitlab-org/gitlab/-/issues/404722) in GitLab 16.2. This claim is null unless the pipeline definition is located in the same project.

jku:
> For GitHub publishers, PyPI checks that the username/organization provided exists. For GitLab we don't do this, since namespaces can be private and therefore we can't check if they exist or not.

I suppose it would not be an impossible requirement to say only public users/orgs can be Trusted Publishers.

This would however have a practical problem when an existing Trusted Publisher decides to make their org private. Dealing with this might not be worth the trouble.

woodruffw:
Most of the ActiveState provider has landed, so this will probably need a decent bit of deconflicting 🙂

facutuesca:
> Most of the ActiveState provider has landed, so this will probably need a decent bit of deconflicting 🙂

Fixed!

warehouse/oidc/models/gitlab.py
woodruffw: Gotcha. That's a bit weird, but I guess it is what it is 🙂 

(We should confirm with GitLab that the double-slash here won't change in the future.)

facutuesca: I think it makes sense considering you can have nested groups/subgroups. So your repo might be located at `gitlab.com/group/subgroup1/subgroup2/repo`. Then, the double-slash is useful to disambiguate between the repo name and the folder. For example:
`gitlab.com/group/subgroup1/subgroup2/repo//folder/file.yml`. Without the double-slash, `repo` might be a folder and `subgroup2` the name of the repo.

di:
> For GitLab we don't do this, since namespaces can be private and therefore we can't check if they exist or not.

Am I correct in assuming that we _can_ do this for public namespaces? If so, we should probably attempt to do this, determine if the namespace is public or not, and verify this only for public namespaces.

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def upgrade():
+    op.execute(
+        """
+        INSERT INTO admin_flags(id, description, enabled, notify)
+        VALUES (
+            'disallow-gitlab-oidc',
+            'Disallow the GitLab OIDC provider',
+            FALSE,
```
di, line 35:
```suggestion
            TRUE,
```

I think we want this to be disabled by default, similar to:

https://github.com/pypi/warehouse/blob/09ea793f35a14bc29e5f05fe93486a8ce2cd615c/warehouse/migrations/versions/9a0ed2044b53_add_activestate_oidc_publisher.py#L69-L80

warehouse/oidc/forms/gitlab.py
di: I'm a bit confused by this, because glancing at https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/path_regex.rb (referenced from this link) it doesn't seem like Gitlab permits emoji either.

I think it's fine for us to not support emoji for now.

warehouse/oidc/models/gitlab.py
```diff
@@ -0,0 +1,288 @@
...
+    if not signed_claim:
+        raise InvalidPublisherError("The ci_config_ref_uri claim is empty")
+
+    # We need at least one of these to be non-empty
+    ref = all_signed_claims.get("ref_path")
+    sha = all_signed_claims.get("sha")
+    if not (ref or sha):
+        raise InvalidPublisherError("The ref_path and sha claims are empty")
+
+    expected = {f"{ground_truth}@{_ref}" for _ref in [ref, sha] if _ref}
```
di, line 46:
```suggestion
    ref_path = all_signed_claims.get("ref_path")
    sha = all_signed_claims.get("sha")
    if not (ref_path or sha):
        raise InvalidPublisherError("The ref_path and sha claims are empty")

    expected = {f"{ground_truth}@{_ref}" for _ref in [ref_path, sha] if _ref}
```
A bit of a nit, but since there's also a `ref` claim, let's be more clear here.

warehouse/oidc/models/gitlab.py
```diff
@@ -0,0 +1,288 @@
...
+    def publisher_name(self):
+        return "GitLab"
+
+    def publisher_url(self, claims=None):
+        base = f"https://gitlab.com/{self.project_path}"
+        sha = claims.get("sha") if claims else None
+
+        if sha:
+            return f"{base}/commit/{sha}"
+        return base
```
di, line 218: Should we do something else if `sha` isn't present but `ref_path` is?

woodruffw:
> Am I correct in assuming that we _can_ do this for public namespaces? If so, we should probably attempt to do this, determine if the namespace is public or not, and verify this only for public namespaces.

Yeah, I think so (@facutuesca can confirm) -- we should be able to do this for public namespaces.

warehouse/oidc/models/gitlab.py
```diff
@@ -0,0 +1,288 @@
...
+    workflow_filepath = mapped_column(String, nullable=False)
+    environment = mapped_column(String, nullable=False)
+
+    __required_verifiable_claims__: dict[str, CheckClaimCallable[Any]] = {
+        "sub": _check_sub,
+        "project_path": check_claim_binary(str.__eq__),
+        "ci_config_ref_uri": _check_ci_config_ref_uri,
+    }
+
+    __required_unverifiable_claims__: set[str] = {"ref_path", "sha"}
```
di, line 115: This has the effect of requiring both claims to be present and not `None`: https://github.com/pypi/warehouse/blob/fd35a53dfc4b2aae994aa9368908097ec5b15552/warehouse/oidc/models/_core.py#L189-L204

But based on _check_ci_config_ref_uri, it seems like at least one of these could be non-empty? We might need to special-case the validation here for these.

warehouse/templates/manage/manage_base.html
```diff
@@ -545,6 +545,15 @@ <h3 class="modal__title">{{ title }}</h3>
       {% else %}
         <i>({% trans %}Any{% endtrans %})</i>
       {% endif %}
+    {% elif publisher.publisher_name == "GitLab" %}
+      <b>Repository:</b> <a href="{{ publisher.publisher_url() }}">{{ publisher.project_path }}</a><br>
```
di, line 549:
```suggestion
      <b>Project:</b> <a href="{{ publisher.publisher_url() }}">{{ publisher.project_path }}</a><br>
```
Per your description for the PR, should we use GitLab's terminology here instead?

woodruffw:
We could also do this in a TOFU-ish manner for private namespaces, since the OIDC claim set includes the `namespace_id`: when the namespace is a private one, we can populate the ID on first sight from the claim set.

(I'm not sure if this is worth doing, however, since it complicates the relationship between the publisher model and token consumption.)

warehouse/templates/manage/project/publishing.html
```diff
@@ -124,6 +124,93 @@
...
+      <div class="form-group">
+        <label for="workflow_filepath" class="form-group__label">
+          {% trans %}Workflow file path{% endtrans %}
+          {% if gitlab_publisher_form.workflow_filepath.flags.required %}
+          <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+          {% endif %}
+        </label>
+        {{ gitlab_publisher_form.workflow_filepath(placeholder=gettext(".gitlab-ci.yml"), class_="form-group__field", autocomplete="off", **{"aria-describedby":"workflow_filepath-errors"}) }}
+        <p class="form-group__help-text">
+          {% trans %}The file path of the publishing workflow, relative to the repository's root. This file should exist in the repository configured above (external workflows are not supported).{% endtrans %}
```
di, line 177: Similar question here about whether we should be calling this a repository.

warehouse/templates/manage/project/publishing.html
```diff
@@ -124,6 +124,93 @@
...
+        <p class="form-group__help-text">
+          {% trans href="https://docs.gitlab.com/ee/ci/environments/" %}
+          The name of the <a href="{{ href }}">GitLab CI/CD environment</a>
+          that the above workflow uses for publishing.  This should be
+          configured under the repository's settings. While not required, a
+          dedicated publishing environment is <strong>strongly</strong>
+          encouraged, <strong>especially</strong> if your repository has
+          maintainers with commit access who shouldn't have PyPI publishing
+          access.
+          {% endtrans %}
```
di, line 202: Similar question here about whether we should be calling this a repository.

di (commented):
Overall LGTM, just some minor comments.

warehouse/templates/manage/manage_base.html
woodruffw: I think so -- we should probably consistently refer to the repo as a project and the workflow as a "job name" or similar.

facutuesca: I agree with being consistent. The reason I used "Repository" here instead of "Project" is that in PyPI's UI, the term project is already used for PyPI Projects (see screenshot). We could disambiguate by saying "PyPI Project" and "GitLab Project"
image

facutuesca: (deleted my previous comment, I was misremembering. I'll fix this to consistently use "project" or "Gitlab project", in order to not confuse it with the "PyPI project")

warehouse/templates/manage/project/publishing.html
facutuesca: Fixed

warehouse/templates/manage/project/publishing.html
facutuesca: Fixed (changed to project)

warehouse/templates/manage/manage_base.html
facutuesca: Done, changed to "Project"

warehouse/oidc/models/gitlab.py
facutuesca: This was due to using the Github code as a base, which assumes either `sha` or `ref_path` can be empty. GitLab's documentation specifies that both should be present, so I changed `_check_ci_config_ref_uri` to reflect that.

warehouse/oidc/models/gitlab.py
facutuesca: This code actually only checks if `claims` is present. We assume that both `sha` and `ref_path` should be present (see my comment below)

warehouse/oidc/models/gitlab.py
facutuesca: fixed

warehouse/migrations/versions/81f9f9a60270_add_gitlab_oidc_models.py
facutuesca: fixed

warehouse/oidc/models/gitlab.py
di: I think this is the only thing still outstanding:

```suggestion
    def publisher_url(self, claims=None):
        base = f"https://gitlab.com/{self.project_path}"
        return f"{base}/commit/{claims['sha']}" if claims else base
```

di (approved):

PR #15320, opened 2024-02-02
feat(admin): display observations in aggregate

miketheman:
As a method for us to learn more about their frequency, shape, utility.


ewdurbin:
Is there a reasonable way to populate a bunch of observations for locally exercising this?

miketheman:
> Is there a reasonable way to populate a bunch of observations for locally exercising this?

Here's a sample script that you could tweak and run in a `make shell` context.

```python
import random
from sqlalchemy import func
from warehouse.accounts.models import User
from warehouse.packaging.models import Project
from warehouse.observations.models import ObservationKind, Observer

users = db.query(User).order_by(func.random()).limit(10).all()
projects = db.query(Project).order_by(func.random()).limit(10).all()

for i in range(20):
    user = random.choice(users)
    project = random.choice(projects)
    # kind is an enum, pick a random one
    kind = random.choice(list(ObservationKind))
    # create the observer for the user if it doesn't exist
    if not user.observer:
        user.observer = Observer()
        db.add(user.observer)
    # create the observation
    observation = project.Observation(
        related=project,
        observer=user.observer,
        kind=kind.value[0],
        summary=f"Random observation {i}",
        payload={"random": "data"},
    )
    db.add(observation)
db.commit()
```

ewdurbin (approved):
As a basis for initial work, I think this is good enough to see "oh we have some!"

I was thinking along the lines of a view that was more specific than just "Observations", but rather for a classification of "Actionable Junk" (working title). I also would have hoped to see observations on the same project grouped somehow.

In my mental model this would allow me to see "oh here are all the current Malware reports that haven't been acted on", click on a group of reports for a given Project, where i'd view the Observations, and have a single button that would allow me to "remove as malware" or "mark as invalid" which would notify the reporter(s).

miketheman:
All good points, definitely we need to design a workflow, which is much larger in scope than "let's see them for now". This was never intended to be a full workflow solution, rather a place to surface them for now.

PR #15337, opened 2024-02-06
chore: cache calls to signing key

### LINKED ISSUE #4463: Add resilience to the network call to fetch the SNS signing certificate
Whenever we're verifying a SNS message, we have to fetch the public certificate from an HTTP url provided to us by Amazon. If fetching this fails for any reason, we will error and will rely on SNS retrying the request to get it accurately recorded.

We can do better!

There are two possible strategies I can think of here, and the right answer might be to use one or the other, or both.

- Cache the public key.
  - The HTTP response at the URL does not indicate that it can be cached, however on the [AWS forums](https://forums.aws.amazon.com/thread.jspa?threadID=174730) AWS has indicated that if/when they change the certificate they will use a different URL. That means one option here is to just cache the signing certificate for a long time. This could either just be a simply in memory cache (in which case we will refetch it anytime we restart the process) or utilizing redis to store the cached signing URL so that the cache survives restarts, is shared amongst processes etc.
  - This cache *should* expire some how, probably some sort of LRU that keeps some number of keys but will evict older ones when needed.
- Add retries.
  - Whenever we get an error, simply try fetching it again! This will make the HTTP request take longer and it's possible that whatever network error is effecting us will last longer then we're willing to have a single request take, so it doesn't eliminate the problem, but makes us survive momentary blips better.

My opinion is I'd start with caching, ideally with a redis based cache and see where that leaves us. It will likely make the failures infrequent enough as to not be worth worrying about, and will make verifying the signature faster as well.

miketheman:
With this simple caching mechanism, each running instance should only have to make a single call at their first instantiation, and cache the result for the lifetime of the process.

This call rarely fails, and adds ~200ms of each inbound hook, so caching across requests should cut down the time it takes to complete the processing.

Instead of using a Redis cache and worrying about cache expiration strategies, if this ever fails a restart should evict the in-memory cache and trigger a new HTTP call for the key.

Resolves #4463

### DIFF
--- a/warehouse/utils/sns.py
+++ b/warehouse/utils/sns.py
@@ -12,6 +12,7 @@
 
 import base64
 import datetime
+import functools
 import re
 
 import requests
@@ -72,6 +73,7 @@ class MessageVerifier:
         except _InvalidSignature:
             raise InvalidMessageError("Invalid Signature") from None
 
+    @functools.lru_cache(maxsize=1)
     def _get_pubkey(self, cert_url):
         # Before we do anything, we need to verify that the URL for the
         # signature matches what we expect.


ewdurbin (commented):
from a pure "ops" perspective i'm -1 on the assumed recovery scenario of restarting the process. while it is straightforward, how are we to recall that this could be the necessary action if it happened?

ewdurbin:
our gunicorn configuration simultaneously makes recovery a non-issue and reduces the impact of this cache:

https://github.com/pypi/warehouse/blob/95c1b6e0ebaf57afb6fe40d5a542469411cc9c77/gunicorn-prod.conf.py#L16-L17

i'm now -0

miketheman:
That **is** interesting, didn't know that. Does this mean we are recycling worker processes every few minutes now?
If we run a single web worker per instance, since we don't set the [`workers` value](https://docs.gunicorn.org/en/stable/settings.html#worker-processes), we default to 1.

We have 40 deployed instances of `web` workers right now, and with a request rate of ~500 requests per second (imprecise, but demonstrative), and each request takes ~150ms to complete (again, using estimations here) - so they'd all exhaust and restart after ~2.5 minutes or so.

Does that align with your understanding of our current universe? If it does, should we consider the `max_requests` value from 2018?

I couldn't find a Redis-backed cache pattern yet in the codebase, I guess we don't do that very much yet, unless I'm looking in the wrong place?

ewdurbin:
Yes, this is effectively a diaper for memory leaks. Each worker gracefully reloads after 1920-2176 requests. This was put in place as our web processes slowly leaked memory over time leading to _non graceful_ restarts.

Ideally we just don't leak memory! I'd reconsider the value if we had demonstrably stable memory usage without.

ewdurbin:
Note: We also tune worker count by the `WEB_CONCURRENCY` env var rather than hardcoding so we don't have to push code to tune it.

miketheman:
> Ideally we just don't leak memory! I'd reconsider the value if we had demonstrably stable memory usage without.

Makes sense! Did you have a way to test/verify memory leakage/profiling prior to production changes in mind, or is this a "let's double `max_requests` and observe" kind of thing?

> Note: We also tune worker count by the `WEB_CONCURRENCY` env var rather than hardcoding so we don't have to push code to tune it.

`WEB_CONCURRENCY` was removed circa 2015 in  https://github.com/pypi/warehouse/pull/741/files#diff-0a99231995da379e7aebabe76c9d849a23737a42c3b3a8994043e2aa80958424 and I can't find another tunable for gunicorn

ewdurbin:
WEB_CONCURRENCY is native to gunicorn: https://docs.gunicorn.org/en/stable/settings.html#workers

miketheman:
> WEB_CONCURRENCY is native to gunicorn: [docs.gunicorn.org/en/stable/settings.html#workers](https://docs.gunicorn.org/en/stable/settings.html#workers)

🤯 That'll learn me to read the docs more

ewdurbin:
FWIW the leak appears to still exist, here's a sample trace that appears as i recall, a sudden spike in memory usage that remains until the worker is reaped.
Screenshot 2024-02-08 at 9 58 09 AM

firefly9966 (approved):

miketheman:
Gonna close this out based on the frequency of restarts. We may have to consider another solution if this is a problem.

PR #15365, opened 2024-02-12
tests: run postgres via pytest

miketheman:
Instead of running a separate container, allow the pytest plugin to
bootstrap its own postgres process.

Increases the development image by adding the postgres server package.

Changes the overall Docker runtime to run under a non-root user.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -225,6 +225,7 @@ COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehous
 COPY --from=build /opt/warehouse/ /opt/warehouse/
 COPY . /opt/warehouse/src/
 
-RUN adduser warehouse
-RUN chown -R warehouse /opt/warehouse/src
+# We cannot run `postgres` as root, so add a user to run the application
+# The user is replaced by `nobody` in the production image by cabotage.
+RUN adduser warehouse && chown -R warehouse /opt/warehouse/src
 USER warehouse
--- a/Dockerfile
+++ b/Dockerfile
@@ -212,7 +212,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && apt-get update \
     && apt-get install --no-install-recommends -y \
         libpq5 libxml2 libxslt1.1 libcurl4  \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -224,3 +224,7 @@ COPY --from=static /opt/warehouse/src/warehouse/static/dist/ /opt/warehouse/src/
 COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehouse/src/warehouse/admin/static/dist/
 COPY --from=build /opt/warehouse/ /opt/warehouse/
 COPY . /opt/warehouse/src/
+
+RUN adduser warehouse
+RUN chown -R warehouse /opt/warehouse/src
+USER warehouse
```
miketheman, line 230: This is the one thing I'm not the most certain about, since it changes real runtime behavior to execute all processes under this user, vs previously implict `root`.

It's a requirement for postgres, as pg's built-in logic will actively deny it from being started as `root` for safety reasons.

di (approved):

Dockerfile
ewdurbin: cabotage currently forces built containers to `nobody` [here](https://github.com/cabotage/cabotage-app/blob/2d7c0262ec347734adfde86e859a2a27b18e3964/cabotage/utils/release_build_context.py#L3-L10).

i'm not sure how using a named user like `warehouse` will interact.

i recommend remove the `adducer` and `chown` parts and just replacing it with `USER nobody`

miketheman: I tried replacing, and it failed immediately, since it cannot write coverage.

Considering that cabotage adds the last layer of nobody and sets an entrypoint, and we need the ability to write local files for coverage, or creating db revisions during development, I'm electing to keep this user as is, and add a comment above.

ewdurbin (approved):
now that it's `nobody` i'm 👍🏼

just had to configure coverage to write to a directory owned by the user.

confirmed creating migrations works.

miketheman:
Looks like `nobody` kills `/opt/warehouse/src/.pytest_cache/v/randomly_seed` for the same reason of not being able to write to the homedir, which is a warning, not an error.

I'll look for toggles that move that file

PR #15366, opened 2024-02-12
oidc/forms: improve project exists error

### LINKED ISSUE #14232: Improve PendingPublisher UX when project already exists

woodruffw:
This will hopefully offset some user confusion around pending versus "normal" publishers.

Towards #14232.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -570,14 +570,14 @@ msgid "Invalid project name"
 msgstr ""
 
 #: warehouse/oidc/forms/_core.py:35
-msgid "This project name is already in use"
+msgid "This project already exists, create an ordinary Trusted Publisher instead"
 msgstr ""
 
-#: warehouse/oidc/forms/_core.py:44
+#: warehouse/oidc/forms/_core.py:46
 msgid "Specify a publisher ID"
 msgstr ""
 
-#: warehouse/oidc/forms/_core.py:45
+#: warehouse/oidc/forms/_core.py:47
 msgid "Publisher must be specified by ID"
 msgstr ""
 
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -573,11 +573,11 @@ msgstr ""
 msgid "This project already exists, create an ordinary Trusted Publisher instead"
 msgstr ""
 
-#: warehouse/oidc/forms/_core.py:46
+#: warehouse/oidc/forms/_core.py:47
 msgid "Specify a publisher ID"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


warehouse/oidc/forms/_core.py
```diff
@@ -32,7 +32,9 @@ def validate_project_name(self, field):
 
         if project_name in self._project_factory:
             raise wtforms.validators.ValidationError(
-                _("This project name is already in use")
+                _(
+                    "This project already exists, create an ordinary Trusted Publisher instead"
```
di, line 36:
```suggestion
                    "This project already exists, create an ordinary Trusted "
                    "Publisher instead"
```

warehouse/locale/messages.pot
```diff
@@ -570,14 +570,14 @@ msgid "Invalid project name"
 msgstr ""
 
 #: warehouse/oidc/forms/_core.py:35
-msgid "This project name is already in use"
+msgid "This project already exists, create an ordinary Trusted Publisher instead"
 msgstr ""
 
-#: warehouse/oidc/forms/_core.py:44
+#: warehouse/oidc/forms/_core.py:46
```
di, line 576:
```suggestion
#: warehouse/oidc/forms/_core.py:47
```

warehouse/locale/messages.pot
```diff
@@ -570,14 +570,14 @@ msgid "Invalid project name"
...
 msgid "Specify a publisher ID"
 msgstr ""
 
-#: warehouse/oidc/forms/_core.py:45
+#: warehouse/oidc/forms/_core.py:47
```
di, line 580:
```suggestion
#: warehouse/oidc/forms/_core.py:48
```

di (approved):

PR #15368, opened 2024-02-13
Add a metadata backfill task

di:
Towards resolving #8254, this adds a periodic task that backfills wheel metadata in reverse chronological order.

### DIFF
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -11,13 +11,19 @@
 # limitations under the License.
 
 import datetime
+import hashlib
 import logging
+import os
 import tempfile
 
 from collections import namedtuple
 from itertools import product
+from pathlib import Path
 
 from google.cloud.bigquery import LoadJobConfig
+from pip._internal.network.lazy_wheel import dist_from_wheel_url
+from pip._internal.network.session import PipSession
+from sqlalchemy import desc
 from sqlalchemy.orm import joinedload
 
 from warehouse import tasks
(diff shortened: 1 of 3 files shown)


warehouse/packaging/tasks.py
```diff
@@ -54,6 +60,67 @@ def sync_file_to_cache(request, file_id):
         file.cached = True
 
 
+@tasks.task(ignore_result=True, acks_late=True)
+def backfill_metadata(request):
+    metrics = request.find_service(IMetricsService, context=None)
+    is_pypi = request.registry.settings.get("warehouse.domain") == "pypi.org"
+    subdomain = "files" if is_pypi else "test-files"
+    base_url = f"https://{subdomain}.pythonhosted.org/packages"
```
ewdurbin, line 68: Should be able to use config "files.url"?

ewdurbin (approved):
One nit regarding available config versus hard-coding subdomains. Otherwise, should be fine?

warehouse/packaging/tasks.py
di: Nice catch

warehouse/packaging/tasks.py
```diff
@@ -54,6 +60,65 @@ def sync_file_to_cache(request, file_id):
...
+                hashlib.sha256(wheel_metadata_contents).hexdigest().lower()
+            )
+            file_.metadata_file_blake2_256_digest = (
+                hashlib.blake2b(wheel_metadata_contents, digest_size=256 // 8)
+                .hexdigest()
+                .lower()
+            )
+
+            # Store the metadata file via our object storage backend
+            storage.store(
```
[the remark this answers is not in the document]
ewdurbin: well cheap in dev time. it could be expensive on reads from S3 glacier.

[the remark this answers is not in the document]
di: I did an even cheaper fix: c8a6a88

warehouse/packaging/tasks.py
```diff
@@ -54,6 +60,77 @@ def sync_file_to_cache(request, file_id):
...
+
+            # Hash the metadata and add it to the File instance
+            file_.metadata_file_sha256_digest = (
+                hashlib.sha256(wheel_metadata_contents).hexdigest().lower()
+            )
+            file_.metadata_file_blake2_256_digest = (
+                hashlib.blake2b(wheel_metadata_contents, digest_size=256 // 8)
+                .hexdigest()
+                .lower()
+            )
```
edmorley, line 102: I presume the digests here are written to the DB immediately? (Rather than when `_file`'s destructor is called)

If so, it seems like there is a failure mode here that could leave files without their metadata file?

ie: If the backfill task's worker process gets killed/restarted between writing the digests and storing the metadata file in the storage backend (or if the write to the storage backend fails). In such a case, the next time the backfill tasks runs, the wheel in question wouldn't be returned by the DB query since the digests would then be set - even though the metadata file was never uploaded.

If setting the digests was moved after writing to the storage backend, then in the case of an interrupted backfill task, the inconsistency between storage backend (which would have the uploaded file) and DB digest state (which wouldn't yet have the digest) would instead self-heal on the next run.

ewdurbin: this is a non-issue, the changes to the database aren't persisted until the entire request has completed and a transaction is committed.

edmorley: Ah good to know. (The fact that everything was configured to use transactions isn't obvious from code inspection of this single task.)

PR #15373, opened 2024-02-13
Downgrade pytest from 8.0.0 to 7.4.4

di:
Attempting to resolve hung runs of our test workflow, e.g.:

https://github.com/pypi/warehouse/actions/runs/7884312889/job/21513166129 https://github.com/pypi/warehouse/actions/runs/7884312873/job/21513149554


miketheman:
Without an associated pin in tests.in, dependabot will suggest an upgrade - is that intended?

di:
> Without an associated pin in tests.in, dependabot will suggest an upgrade - is that intended?

It won't, because we closed https://github.com/pypi/warehouse/pull/15287, but it will suggest new releases, which I think is what we want (maybe this will get fixed in the next release, whenever that happens).

miketheman (approved):

PR #15374, opened 2024-02-13
Handle invalid wheels (Part 1): Add `File.unbackfillable` column

di:
This adds a temporary column `File.unbackfillable` which will be set when we determine we can't backfill metadata for a wheel.

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -719,6 +719,9 @@ class File(HasEvents, db.Model):
     archived: Mapped[bool_false] = mapped_column(
         comment="If True, the object has been archived to our archival bucket.",
     )
+    unbackfillable: Mapped[bool_false] = mapped_column(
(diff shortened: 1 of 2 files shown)


ewdurbin:
We have so many files. I'm worried using nullable=False here will make this migration a pain. Please either bump timeout ahead or consider if we can handle a null case

warehouse/packaging/models.py
```diff
@@ -719,6 +719,9 @@ def __table_args__(cls):  # noqa
     archived: Mapped[bool_false] = mapped_column(
         comment="If True, the object has been archived to our archival bucket.",
     )
+    unbackfillable: Mapped[bool_false] = mapped_column(
```
ewdurbin, line 722: metadata_file_unbackfillable?

ewdurbin (approved):

PR #15377, opened 2024-02-14
Fix simple json requires-python is empty string

BSathvik:
Very much a nit. The jinja template for simple [project detail](https://github.com/pypi/warehouse/blob/facbd56b4810e17ee2182aab64eeee1f84851998/warehouse/templates/api/simple/detail.html#L23) doesn't produce `data-requires-python` with empty string, but the json does. I ran the change locally to test it out

```
$ curl -s -H "Accept: application/vnd.pypi.simple.v1+html" https://pypi.org/simple/backoff/ | grep -o 'data-requires-python=""'
$ curl -s -H "Accept: application/vnd.pypi.simple.v1+json" https://pypi.org/simple/backoff/ | grep -o '"requires-python":""'
"requires-python":""
"requires-python":""
"requires-python":""
```


di (approved):
This is aligned with what we are doing [here](https://github.com/pypi/warehouse/pull/13606#discussion_r1261300580) as well so LGTM, thanks for keeping us consistent.

PR #15385, opened 2024-02-14
refactor: replace manage:user with Permissions

miketheman:
Anywhere we had `manage:user`, use a more specific Permissions entry.

A couple of non-Account-related permissions were added here, as the
previous use of `manage:user` was put in place of a more specific
permission, since that's all we had.

### DIFF
--- a/tests/unit/macaroons/test_security_policy.py
+++ b/tests/unit/macaroons/test_security_policy.py
@@ -20,6 +20,7 @@ from pyramid.security import Denied
 from zope.interface.verify import verifyClass
 
 from warehouse.accounts.interfaces import IUserService
+from warehouse.authnz import Permissions
 from warehouse.macaroons import security_policy
 from warehouse.macaroons.interfaces import IMacaroonService
 from warehouse.macaroons.services import InvalidMacaroonError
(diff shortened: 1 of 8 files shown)


warehouse/config.py
```diff
@@ -135,7 +135,21 @@ class RootFactory:
...
-        (Allow, Authenticated, "manage:user"),
+        (
+            Allow,
+            Authenticated,
+            (
+                Permissions.Account2FA,
+                Permissions.AccountAPITokens,
+                Permissions.AccountManage,
+                Permissions.AccountManagePublishing,
+                Permissions.AccountVerifyEmail,
```
ewdurbin, line 146: just a thought, but some of these feel a bit too granular but overall seem reasonable to me. i think the general question i'd ask when granularizing is "would another entity do this on the behalf of another". in this instance i think that most of the Verify* will only be acted on by a user so in my (very negotiable) mental model we could just have "Permissions.AccountVerify" that covers all of these.

probably no action here just writing down the thought provoked.

ewdurbin (approved):
forward! i really dig the outcomes here and am excited to see us closer to the end goal.

tests/unit/macaroons/test_security_policy.py
```diff
@@ -316,7 +317,7 @@ def test_permits_valid_macaroon(self, monkeypatch, principals, expected):
 
     @pytest.mark.parametrize(
         "invalid_permission",
-        ["admin", "moderator", "manage:user", "manage:project", "nonexistent"],
```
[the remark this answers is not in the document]
miketheman: That's correct - `moderator` is gone as well. I expect to remove `manage:project` when I refactor it to oblivion.

warehouse/config.py
miketheman: I hear you - I was going through and breaking things down to the smallest unit first, and then seeing where things can maybe be combined. I'll keep these in mind when I take a more holistic view of account provisioning, and may bundle many of them together then.

PR #15408, opened 2024-02-16
Nitpick and cleanups

miketheman:
See each commit for reasoning.


warehouse/forklift/legacy.py
```diff
@@ -73,7 +73,6 @@
 ONE_GB = 1 * 1024 * 1024 * 1024
 
 MAX_FILESIZE = 100 * ONE_MB
-MAX_SIGSIZE = 8 * 1024
```
ewdurbin, line 76: at first i was surprised pylint didn't catch this but then i guess it just looks file by file?

thought: is there any way to get pylint to consider if a symbol isn't used anywhere in a codebase? seems kinda extreme and would make no sense for libraries... but 🤔

ewdurbin (approved):

warehouse/forklift/legacy.py
miketheman: We don't run pylint, and I don't think flake8 can detect unused constants

PR #15430, opened 2024-02-20
test: missed one level of nesting

miketheman:
Started failing the `dev env` daily tests.

PR #15444, opened 2024-02-22
Revert "tests: run postgres via pytest (#15365)"

di:
This reverts #15365, due to issues with having to set the `nobody` user with local development on Linux.

Closes #15430.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,7 +85,6 @@ services:
       - ./htmlcov:/opt/warehouse/src/htmlcov:z
       - ./setup.cfg:/opt/warehouse/src/setup.cfg:z
       - ./pyproject.toml:/opt/warehouse/src/pyproject.toml:z
-      - .coveragerc:/opt/warehouse/src/.coveragerc:z
       - packages:/var/opt/warehouse/packages
       - packages-archive:/var/opt/warehouse/packages-archive
       - sponsorlogos:/var/opt/warehouse/sponsorlogos
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -115,6 +115,16 @@ services:
       stripe:
         condition: service_started
 
+  tests:
+    image: warehouse:docker-compose
+    pull_policy: never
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -311,8 +313,6 @@ def app_config(database):
         "docs.backend": "warehouse.packaging.services.LocalDocsStorage",
         "sponsorlogos.backend": "warehouse.admin.services.LocalSponsorLogoStorage",
         "billing.backend": "warehouse.subscriptions.services.MockStripeBillingService",
-        "billing.api_base": "http://stripe:12111",
-        "billing.api_version": "2020-08-27",
```
miketheman, line 315: This was a nice change that allowed us to ignore `dev/environment` for tests, which is restored by using the same compose service `base` vs the new and now removed `tests`. Not a blocker.

docker-compose.yml
```diff
@@ -85,6 +85,7 @@ services:
       - ./htmlcov:/opt/warehouse/src/htmlcov:z
       - ./setup.cfg:/opt/warehouse/src/setup.cfg:z
       - ./pyproject.toml:/opt/warehouse/src/pyproject.toml:z
+      - .coveragerc:/opt/warehouse/src/.coveragerc:z
```
miketheman, line 88: We don't need to restore this - it was left over from #14758 
```suggestion
```

docker-compose.yml
```diff
@@ -114,14 +115,6 @@ services:
       stripe:
         condition: service_started
 
-  tests:
-    image: warehouse:docker-compose
-    pull_policy: never
-    volumes: *base_volumes
-    depends_on:
-      stripe:
-        condition: service_started
```
miketheman, line 123: I think there's still value in having this split out to it's own compose service, if nothing else to prevent needing to share `dev/environment and redis servisce for tests. Not a blocker for now, but a good testing improvement for the future.

miketheman (approved):
One change, but otherwise 👍

Issue #15384, opened 2024-02-14
Upgrade runtime to Python 3.11.8


miketheman:
The current stack runs on Python 3.11.7, and 3.11.8 has been released with some fixes.
See https://docs.python.org/release/3.11.8/whatsnew/changelog.html#python-3-11-8

Places to update:

- `Dockerfile` (at least 3 spots)
- `.python-version`

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypi/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

PR #15446, opened 2024-02-22
chore(deps): Upgrade Python runtime from 3.11.7 to 3.11.8

Mr-Sunglasses:
fix: #15384


Dockerfile
```diff
@@ -82,10 +82,10 @@ COPY requirements /tmp/requirements
...
     set -x \
     && pip --disable-pip-version-check \
-            install --no-deps \
-            -r /tmp/requirements/docs-dev.txt \
-            -r /tmp/requirements/docs-user.txt \
-            -r /tmp/requirements/docs-blog.txt \
+    install --no-deps \
+    -r /tmp/requirements/docs-dev.txt \
+    -r /tmp/requirements/docs-user.txt \
+    -r /tmp/requirements/docs-blog.txt \
```
miketheman, line 88: **question:** can you explain why the leading whitespace was removed here? We have it here and in other spots to increase readability of multi-line commands.‏

miketheman (commented):
Thanks @Mr-Sunglasses ! Generally looks good.

I asked a question inline that should be answered before we proceed.

Dockerfile
Mr-Sunglasses: I think it is done by `prettier`, Let me fix them.

miketheman (approved):
Thanks @Mr-Sunglasses !

PR #15450, opened 2024-02-22
Handle InvalidWheel for metadata backfill

di:
Fixes https://python-software-foundation.sentry.io/issues/5002343865/.


miketheman (approved):

miketheman:
For once this is done, it might be an interesting blog post to look at the unbackfillable as a "here's some weird stuff that we've accumulated over 20 years of existence"

Issue #14171, opened 2023-07-20
Warn about API key usage/presence when Trusted Publishing is configured


di:
Currently in our Trusted Publishing flow, we don't direct the user to remove/delete API tokens that were previously being used for pushing, potentially leading to the long-lived tokens remaining available for publishing.

As a result, we should do the following:
- We should warn (via email) whenever an API key is used for a project where Trusted Publishing is configured, suggesting that the maintainer remove the API token.
- We should additionally send a one-time or occasional email (every 1 month?) whenever an API token exists for a project where Trusted Publishing is configured, regardless of whether it's been used fr publishing.
- We should update the flow to immediately identify long-lived tokens that are no longer necessary, and prompt the user to remove them.

sethmlarson:
Should we wait until after trusted publishing has succeeded once to email them? I suspect it would be frustrating to get an email immediately if you're evaluating trusted publishers.

di:
Yes, my thought is that the first bullet above would only happen after a publish event with a trusted publisher, and the second would only happen after some period of time after the publisher was configured.

ristomcgehee:
Instead of prompting to remove old API tokens, I think it would be better to have a project setting that disables uploading via API token, as discussed in https://github.com/pypi/warehouse/issues/14300. Removing all API tokens for a project might not be straightforward, because a user is not able to see a list of outstanding tokens that are scoped to a project since there might be tokens that belong to other users. There also might be tokens scoped to all projects for a user, and some of those projects might not be set up to use Trusted Publishing.

It could still be worthwhile sending an email when an API token is used if Trusted Publishing is enabled, and also if API tokens exist and the project has not disabled publishing with API tokens.

What do y'all think?

woodruffw:
Triaging: @facutuesca and I will be working on this next!

> * We should warn (via email) whenever an API key is used for a project where Trusted Publishing is configured, suggesting that the maintainer remove the API token.

This sounds good to me! We might want to cap this on "once per release," to avoid sending N emails for N files uploaded at a time.

> Should we wait until after trusted publishing has succeeded once to email them? I suspect it would be frustrating to get an email immediately if you're evaluating trusted publishers.

Yeah, this makes sense to me -- there's a `Release.trusted_published` property that we could key on for this.

To summarize, I think we can do this in two parts:

- [ ] Email notifications, both on release creation and periodically (30 days to start)
[...]

PR #15453, opened 2024-02-22
Warn users when API token is used in Trusted Publishing project

facutuesca:
First part of https://github.com/pypi/warehouse/issues/14171

If a user uploads files using an API token to a project that has a Trusted Publisher configured, send an email to all the project's users notifying them of the API token use, suggesting they remove it and use Trusted Publishing instead.

This warning is only sent if the upload is for a new release of the project. Uploads of files for existing releases do not trigger the warning.

### DIFF
--- a/tests/unit/email/test_init.py
+++ b/tests/unit/email/test_init.py
@@ -5850,3 +5850,123 @@ class TestTrustedPublisherEmails:
                 },
             )
         ]
+
+    def test_api_token_warning_with_trusted_publisher_emails(
+        self, pyramid_request, pyramid_config, monkeypatch
+    ):
+        template_name = "api-token-used-in-trusted-publisher-project"
+        # We set up two users to receive the email. The owner of the API token
+        # will be stub_user, their username should be the one mentioned in the
+        # email body.
+        stub_user = pretend.stub(
+            id="id",
+            username="username",
+            name="",
+            email="email@example.com",
+            primary_email=pretend.stub(email="email@example.com", verified=True),
+        )
+        stub_user_maintainer = pretend.stub(
+            id="id_maintainer",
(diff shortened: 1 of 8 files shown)


di:
> This warning is only sent if the upload is for a new release of the project. Uploads of files for existing releases do not trigger the warning.

Per discussion in #14171, I think this should:
* send the email on the first file uploaded, regardless of whether it is creating a new release or not
* only send one email per API token, ever

di:
That all sounds okay, except:

> We could even relax the requirement of only warning about API tokens scoped to a single project. If a user uses a token scoped for multiple projects to upload to one of them, with a trusted publisher already existing, they'll still get the warning only once.

I don't think we could relax this, because we want to send this to all owners/maintainers of the project, not just the user who owns the token. So a user might use a account-wide token to upload to two different projects: we'd want to send a notification for each usage in that case.

And we probably shouldn't exclude account-wide API tokens from this notification, so an association table between tokens <-> projects seems to still make the most sense.

One other thing: we should make sure these notifications are immediately actionable. It would be ideal if they linked to a URL where the user could immediately revoke the token in question.

facutuesca:
@di
Alright, then I'll summarize what we've got so far, because we changed a lot from the previous discussion. Before we had:
> 1. Also, we only care about API tokens scoped only for that single project
> 2. only send one email per API token, ever
> 3. Yep: we want to email the user who can take the action, not other users

And now we have:
1. Send warnings for API tokens both project-scoped and account-level scoped.
2. Send only one email per `(API token, Project)` combination
3. Send the email to all of the project's maintainers

Is that accurate, or did I miss something?

facutuesca:
Closing this in favor of https://github.com/pypi/warehouse/pull/15641

Issue #15356, opened 2024-02-09
Trim leading and trailing whitespace from username in forms


miketheman:
**Describe the bug**
When using a mobile keyboard for input, auto complete may add a trailing space to a username field in a form, making the submission fail due to mismatch.

**Expected behavior**

Any trailing or leading whitespace is removed either from the form, the view, or the database query, allowing the user to log in

**To Reproduce**

Try logging in with your username and add a space to the end.

HimanshuSinghGH:
Hi @miketheman, I would like to work on this. I reproduced the problem on the login page and understand how to solve this. However, when I try to configure the development environment on my system it fails after the `make initdb` step with an error saying **"warehouse-db-1 is unhealthy"**

Is this a common issue and is there a known fix available ?

![db_1](https://github.com/pypi/warehouse/assets/20122511/ea80fff4-f162-49bc-84de-0f1ef7519a6d)

miketheman:
Hi @HimanshuSinghGH ! Thanks for taking a pass at this.

Can you share some more of your computer environment? It's hard to diagnose from a screenshot.

Did you do these steps in order?
- stop everything with `make stop`
- start the stack with `make serve` and wait a few minutes until you see something like: 
    > static-1         |   admin (webpack 5.88.2) compiled successfully in 5226 ms
- in another terminal session, execute `make initdb`

PR #15455, opened 2024-02-22
fix: Add Trim leading and trailing whitespace from username in forms

Mr-Sunglasses:
fix; #15356 

https://github.com/pypi/warehouse/assets/81439109/fc69b528-6c33-4ca4-b298-0ca8b0180627

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -78,7 +78,8 @@ class UsernameMixin:
         validators=[
             wtforms.validators.InputRequired(),
             PreventNullBytesValidator(message=INVALID_USERNAME_MESSAGE),
-        ]
+        ],
+        filters=[lambda x: x.strip() if x else None]
     )
 
     def validate_username(self, field):


di:
@Mr-Sunglasses You'll want to update the translations, see https://warehouse.pypa.io/development/submitting-patches.html#translations

miketheman (commented):
Nice work so far @Mr-Sunglasses !

Since we try to add tests for functionality and guarantee that the behavior is tested when someone else changes the code, please look at the existing tests for the Login form. Here's one:
https://github.com/pypi/warehouse/blob/26a3446ada6c2db27e6e608d81508ca25018f389/tests/unit/accounts/test_forms.py#L81-L92

You could either:
- add a new test like `test_validate_username_strips_whitespace()` and have the input have some whitespace and confirm that the submitted form passes validation and calls the user_service with the trimmed username, or
- use `@pytest.mark.parametrize()` ([documentation here](https://docs.pytest.org/en/stable/how-to/parametrize.html)) and update the existing `test_validate_username_with_user()` test function to accept the existing `my_username` and expect that in the service call, and add another entry with whitespace and expect the resulting call to not have the whitespace.

The second option is more advanced, so either test solution would be acceptable.

Mr-Sunglasses:
> Nice work so far @Mr-Sunglasses !
> 
> Since we try to add tests for functionality and guarantee that the behavior is tested when someone else changes the code, please look at the existing tests for the Login form. Here's one:
> 
> https://github.com/pypi/warehouse/blob/26a3446ada6c2db27e6e608d81508ca25018f389/tests/unit/accounts/test_forms.py#L81-L92
> 
> You could either:
> 
> * add a new test like `test_validate_username_strips_whitespace()` and have the input have some whitespace and confirm that the submitted form passes validation and calls the user_service with the trimmed username, or
> * use `@pytest.mark.parametrize()` ([documentation here](https://docs.pytest.org/en/stable/how-to/parametrize.html)) and update the existing `test_validate_username_with_user()` test function to accept the existing `my_username` and expect that in the service call, and add another entry with whitespace and expect the resulting call to not have the whitespace.
> 
> The second option is more advanced, so either test solution would be acceptable.

Done, Added tests Thanks a lot @miketheman 🙏🏻

miketheman (approved):
Nice work. Glad to see that the tests surfaced an even simpler solution.
Thanks for working through this issue!

Issue #15432, opened 2024-02-20
Periodically purge OIDC-exchanged API tokens from the DB


woodruffw:
From https://github.com/pypi/warehouse/pull/15240#discussion_r1496243886: PyPI is currently accumulating "orphaned" API tokens in the macaroons table, since Trusted Publishing's OIDC exchange produces (by design) a new self-expiring API token each time. 

These don't cause any real issues on their own, but they might contribute to unnecessarily long/contentious locks during data migrations. So a periodic job to purge them would probably be ideal 🙂 

In terms of how best to do that: `Macaroon.created` reflects the Macaroon's *database* creation time, which should be nearly identical to the interior `Expiration` caveat on the user-held Macaroon. OIDC-minted Macaroons expire 15 minutes after their creation, so it'd be more than safe to have a once-daily job that deletes every DB macaroon that satisfies `Macaroon.oidc_publisher_id is not None and Macaroon.created > 15 minutes ago`.

woodruffw:
(Triaging: this falls into the scope of the OIDC maintenance that @facutuesca and I are doing, so we'll work on this as well.)

facutuesca:
PR created here: https://github.com/pypi/warehouse/pull/15463

PR #15463, opened 2024-02-23
Add daily task to purge expired OIDC macaroons

facutuesca:
This PR adds a daily task to delete all OIDC-minted macaroons that are no longer valid. Since these OIDC-minted tokens are meant to be short-lived (15 minutes) and are generated automatically by CI/CD processes, it makes sense to not let them accumulate once they have expired.

The task deletes all Macaroons that have an `oidc_publisher_id` field and that have a creation date older than 15 minutes in the past (meaning that they have already expired).

Fixes https://github.com/pypi/warehouse/issues/15432.

### DIFF
--- a/tests/unit/oidc/test_tasks.py
+++ b/tests/unit/oidc/test_tasks.py
@@ -101,8 +101,9 @@ def test_compute_oidc_metrics(db_request, metrics):
 
 
 def test_delete_expired_oidc_macaroons(db_request, macaroon_service, metrics):
-    # We'll create 3 macaroons:
+    # We'll create 4 macaroons:
     # - An OIDC macaroon with creation time of 1 day ago
+    # - An OIDC macaroon with creation time of 1 hour ago
     # - An OIDC macaroon with creation time now
     # - A non-OIDC macaroon with creation time of 1 day ago
     # The task should only delete the first one
@@ -120,6 +121,18 @@ def test_delete_expired_oidc_macaroons(db_request, macaroon_service, metrics):
         additional={"oidc": publisher.stored_claims(claims)},
     )
     old_oidc_macaroon.created -= datetime.timedelta(days=1)
+
+    # Create an OIDC macaroon and set its creation time to 1 hour ago
+    macaroon_service.create_macaroon(
+        "fake location",
+        "fake description",
+        [
+            caveats.OIDCPublisher(oidc_publisher_id=str(publisher.id)),
+        ],
+        oidc_publisher_id=publisher.id,
+        additional={"oidc": publisher.stored_claims(claims)},
+    )
+    old_oidc_macaroon.created -= datetime.timedelta(hours=1)
 
     # Create a normal OIDC macaroon
     macaroon_service.create_macaroon(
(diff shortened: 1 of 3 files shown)


warehouse/oidc/tasks.py
```diff
@@ -65,3 +68,21 @@ def compute_oidc_metrics(request):
...
+    # the last time this task was run.
+    (
+        request.db.query(Macaroon)
+        .filter(Macaroon.oidc_publisher_id.isnot(None))
+        .filter(
+            # The expiration time (created + 15 mins) is in the past
+            Macaroon.created + timedelta(minutes=15)
+            < datetime.now(tz=timezone.utc)
+        )
+        .delete(synchronize_session=False)
```
miketheman, line 87: **suggestion:** capture the result of the delete operation and emit as a metric from [`result.rowcount`](https://docs.sqlalchemy.org/en/20/tutorial/data_update.html#getting-affected-row-count-from-update-delete) so we have an idea of how many are being removed each run.‏

warehouse/oidc/tasks.py
```diff
@@ -65,3 +68,21 @@ def compute_oidc_metrics(request):
...
             tags=[f"publisher:{discriminator}"],
         )
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def delete_expired_oidc_macaroons(request):
+    # Purge all API tokens minted using OIDC Trusted Publishing with a creation time
+    # more than 15 minutes ago. Since OIDC-minted macaroons expire 15 minutes after
+    # creation, this task cleans up all the expired tokens that have accumulated since
+    # the last time this task was run.
```
miketheman, line 78: **suggestion:** use a docstring‏

facutuesca: fixed

warehouse/oidc/tasks.py
facutuesca: Added

warehouse/oidc/__init__.py
```diff
@@ -78,3 +78,7 @@ def includeme(config):
 
     # Compute OIDC metrics periodically
     config.add_periodic_task(crontab(minute=0, hour="*"), compute_oidc_metrics)
+
+    # Daily purge expired OIDC-minted API tokens. These tokens are temporary in nature
+    # and expire after 15 minutes of creation.
+    config.add_periodic_task(crontab(minute=0, hour=0), delete_expired_oidc_macaroons)
```
woodruffw, line 84: Extreme nitpick: there are probably other periodic tasks that run around midnight daily, so we could shift this to `hour=6` or similar to spread the (admittedly very small) load. In practice however, I don't think this will matter much.

facutuesca: Changed to `hour=6`

warehouse/oidc/tasks.py
```diff
@@ -65,3 +68,28 @@ def compute_oidc_metrics(request):
...
+    the last time this task was run.
+    """
+    rows_deleted = (
+        request.db.query(Macaroon)
+        .filter(Macaroon.oidc_publisher_id.isnot(None))
+        .filter(
+            # The expiration time (created + 15 mins) is in the past
+            Macaroon.created + timedelta(minutes=15)
+            < datetime.now(tz=timezone.utc)
+        )
```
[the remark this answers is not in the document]
facutuesca: Fixed, I changed it to 1+ days.

tests/unit/oidc/test_tasks.py
```diff
@@ -95,3 +98,62 @@ def test_compute_oidc_metrics(db_request, metrics):
...
         ),
     ]
+
+
+def test_delete_expired_oidc_macaroons(db_request, macaroon_service, metrics):
+    # We'll create 3 macaroons:
+    # - An OIDC macaroon with creation time of 1 day ago
+    # - An OIDC macaroon with creation time now
+    # - A non-OIDC macaroon with creation time of 1 day ago
+    # The task should only delete the first one
```
woodruffw, line 108: Let's add another OIDC macaroon here, one that's expired by has been expired for `< 1` day 🙂

woodruffw (commented):
LGTM overall, one last test request! Great work here @facutuesca!

woodruffw (approved):

di (approved):
I confirmed that this will only delete tokens generated for OIDC publishers.

PR #15466, opened 2024-02-23
Add unchecked GitLab claims

di:
Fixes WAREHOUSE-STAGING-Q8 (https://python-software-foundation.sentry.io/issues/5005548900/).


miketheman (approved):

woodruffw (approved):
LGTM!

(Thinking out loud: the TOFU idea for `namespace_id` would also probably work for `project_id`: leave them unpopulated at publisher creation time, and then populate them based on claim values on first use. But that's an additional complexity.)

PR #15467, opened 2024-02-23
Disable ddtrace in tests

di:
We don't currently use `ddtrace` in our test suite, and the library's plugin automatically instruments tests, so disable it for now to see if it's the cause of our test suite hanging.


ewdurbin (approved):

di:
This hung too, so seems like `ddtrace` is not the culprit.

PR #15470, opened 2024-02-25
refactor(IP Address): Update all IP Addresses in tests to use RFC-5735 TEST-NET-* addresses.

### LINKED ISSUE #13186: Update all IP Addresses in tests to use RFC-5735 TEST-NET-* addresses.

Mr-Sunglasses:
fix: #13186


Mr-Sunglasses:
Hey @miketheman , as far as I've researched on the issue #13186 
The IP address ranges for *RFC-5735 TEST-NET-**

```
TEST-NET-1: 192.0.2.0/24
TEST-NET-2: 198.51.100.0/24
TEST-NET-3: 203.0.113.0/24
```

I've updated all the IP Addresses in `test/`

miketheman (commented):
Nice work, a couple of comments inline to think about.

tests/conftest.py
```diff
@@ -111,7 +111,7 @@ def metrics():
 
 @pytest.fixture
 def remote_addr():
-    return "1.2.3.4"
+    return "192.0.2.1"
```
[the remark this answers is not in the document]
Mr-Sunglasses: Hey @miketheman I've some my research and changed them to [scope="session"](https://docs.pytest.org/en/8.0.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) , so they are dynamic now.

Mr-Sunglasses:
@miketheman I've made the changes you suggested me, let me know if there are any more changes to be made in this PR :)

tests/unit/utils/test_wsgi.py
```diff
@@ -130,13 +138,14 @@ def test_accepts_x_forwarded_headers(self, remote_addr_salted):
 
         resp = wsgi.ProxyFixer(app, token=None, ip_salt="pepa")(environ, start_response)
 
+        expected_remote_addr_hashed = calculate_hashed_value("1.2.3.4", "pepa")
```
miketheman, line 141: **question:** Can you help me understand why we're back to using a 1.2.3.4 here?
In my head, all the `REMOTE_ADDR` entries above could be replaced by the `remove_addr` fixture, and the `REMOTE_ADDR_HASHED` below could be replaced with the `remote_addr_hashed` fixture, and the `calculate_hashed_value()` function could go away.‏

tests/conftest.py
```diff
@@ -109,26 +110,21 @@ def metrics():
...
 
 
-@pytest.fixture
-def remote_addr_hashed():
-    """
-    Static output of `hashlib.sha256(remote_addr.encode("utf8")).hexdigest()`
-    Created statically to prevent needing to calculate it every run.
-    """
-    return "6694f83c9f476da31f5df6bcc520034e7e57d421d247b9d34f49edbfc84a764c"
+@pytest.fixture(scope="function")
```
miketheman, line 118: Shouldn't all of these be scoped to `session`, so they effectively become set once per test suite run?

```suggestion
@pytest.fixture(scope="session")
```

tests/conftest.py
```diff
@@ -109,26 +110,21 @@ def metrics():
...
+def remote_addr_hashed(remote_addr):
+    if remote_addr is None:
+        return None
```
miketheman, line 121: **question:** How could `remote_addr` be `None`? Do we ever override the input to this fixture function, or should this condition be removed?

tests/unit/accounts/test_security_policy.py
```diff
@@ -101,12 +101,12 @@ def test_identity_credentials_fail(self, monkeypatch):
             pretend.stub(
                 matched_route=None,
                 banned=pretend.stub(by_ip=lambda ip_address: False),
-                remote_addr="1.2.3.4",
+                remote_addr="192.0.2.1",
```
miketheman, line 104: These should also probably be replaced by `remote_addr`

miketheman (changes requested):
@Mr-Sunglasses This is coming along nicely, thanks for continuing the effort!

There's a few more cases of non-replaced IP address strings - I haven't exhaustively identified each one, but we're getting closer!

firefly9966 (approved):

aviral-bhardwaj:
why this is still open

JacobCoffee:
Abandoned and way behind. #20265 will take its place i guess

PR #15500, opened 2024-02-27
test(ci): disable randomly plugin to isolate variability

miketheman:


miketheman:
This doesn't appear to have the impact as expected. The first 4 runs completed in time, but the 5th exhibits the "hang" we're seeing.

PR #15503, opened 2024-02-28
Quote-encode backfill URLs

di:
Fixes [WAREHOUSE-PRODUCTION-1QM](https://python-software-foundation.sentry.io/issues/5018377784/).

### DIFF
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -15,6 +15,7 @@ import hashlib
 import logging
 import os
 import tempfile
+import urllib.parse
 
 from collections import namedtuple
 from itertools import product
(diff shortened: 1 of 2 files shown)


warehouse/packaging/tasks.py
```diff
@@ -95,7 +96,7 @@ def metadata_backfill_individual(request, file_id):
         return
 
     base_url = request.registry.settings.get("files.url")
-    file_url = base_url.format(path=file_.path)
+    file_url = urllib.parse.quote(base_url.format(path=file_.path), safe=":/")
```
miketheman, line 99: Probably fine for the backfill, but we've been trying to only use one url parser, and settled on urllib3 - see https://github.com/pypi/warehouse/pull/14497

miketheman (approved):
One non blocking comment

warehouse/packaging/tasks.py
di: AFAICT, I'm not actually seeing a way to do this with `urllib3`. Given that we're going to be deleting this task in ~1 week, I'm just going to merge this as is!

Issue #15498, opened 2024-02-27
GitHub repository in Trusted Publisher should (probably) be case insensitive


woodruffw:
From https://github.com/pypa/gh-action-pypi-publish/issues/217: some users configure their trusted publisher with a different casing than their actual GitHub repository (e.g. the repo might be `Bar` and they configure `bar`).

GitHub appears to treat these names as equivalent (repo names appear to be case insensitive generally), but PyPI performs a case sensitive comparison. As a result, some trusted publishing activities are erroneously rejected, as in the linked issue.

I think the fix here is just to switch to case insensitive comparisons on the `repository` claim, and (maybe?) on the `repository_owner` as well.

miketheman:
> GitHub appears to treat these names as equivalent

Is this common across all of our publishers, or unique to GitHub?

facutuesca:
For `GitHub`, we already do case-insensitive comparisons for the `sub` claim, which includes the repo name:
https://github.com/pypi/warehouse/blob/01be76a12ed14d0919216896505fd7912d26b3b7/warehouse/oidc/models/github.py#L99. 

Extending that to the repo can be done by modifying https://github.com/pypi/warehouse/blob/01be76a12ed14d0919216896505fd7912d26b3b7/warehouse/oidc/models/github.py#L113-L119
so that `repository` is checked by a new `_check_repository()` function that behaves similarly to `_check_job_workflow_ref()`

facutuesca:
> and (maybe?) on the repository_owner as well.

@woodruffw When creating a GitHub trusted publisher, PyPI queries GitHub to see if the owner exists, and stores the name returned by GitHub as the "ground truth":
https://github.com/pypi/warehouse/blob/e50010ac7c54479ce504bb69cb08b04c9ebcda29/warehouse/manage/views/__init__.py#L1273

So for the `repository_owner` claim the normal (case-sensitive) string comparison should be fine, since the ground truth is whatever normalized name GitHub returned during that first check.

woodruffw:
#15501 should fix this.

PR #15512, opened 2024-02-29
oidc/gitlab: make project path comparison case insensitive

facutuesca:
Following the discussion at https://github.com/pypi/warehouse/issues/15498, and the corresponding GitHub fix in https://github.com/pypi/warehouse/pull/15501, this PR fixes the corresponding issue in GitLab trusted publishers.

The GitLab `project_path` claim corresponds to the string `namespace/project` (or `org/repo`, in GH terms). This path is case insensitive, so we change the claim comparison accordingly.

/cc @di @woodruffw


woodruffw (approved):
Thanks @facutuesca!

PR #15514, opened 2024-02-29
Add custom JSON serializer for psycopg

di:
Fixes [WAREHOUSE-PRODUCTION-1PG](https://python-software-foundation.sentry.io/issues/4951160120/), where the underlying issue is that we're trying to include a `Caveat` in the `additional` field (JSONB) of a `UserEvent`, and it can't be serialized to JSON.

I tried a few things here, including:

* manually updating all calls to `record_event` to manually serialize `Caveat`s;
* setting/enforcing a custom JSON-serializable type on `record_event`, to ensure we don't backslide;
* various subclassings of `Caveat` to trick `psycopg` into serializing it automatically (was maybe possible with psycopg2, doesn't seem to be possible with psycopg3).

...but the cleanest and most durable thing seems to be just letting us continue passing things like `Caveat`s as part of other objects that can already be serialized, by setting a custom serializer.

One thing I'm unsure about: would it be faster/better to first try to universally `json.dumps(obj)` and catch the `TypeError` and _then_ do the custom serialization, or is this fine? Does it even matter?


di:
I guess the other thing we could do here is make a custom serializer that's less generic, and only knows how to serialize Caveats + everything else json.dumps can serialize, and not every dict-like thing.

That would also resolve my other concern with this, which is that we could accidentally serialize a field of a dict-like thing that we didn't mean to, and leak data as a result.

dstufft:
Yea, being explicit about what we want to serialize makes sense to me.

I guess one thing to note is that we could use a Pyramid renderer to do the serialization. The main thing that gets us is the Pyramid JSON renderer has functionality to allow you to define how you serialize non primitive objects without having to write a giant if else statement. 

Objects can add a ``__json__`` method, or we can register specific serialization functions for specific objects using ``pyramid.renderers.JSON.add_adapter``.

I don't think that's a requirement for this PR or anything, just a thought that popped into my head. The downside is it wouldn't let us differentiate between stuff we want serialize for the database and stuff we want to serialize for public consumption... but I'm not sure that that is something we care about?

dstufft:
Docs: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#serializing-custom-objects

warehouse/db.py
```diff
@@ -187,3 +189,23 @@ def includeme(config):
...
+            elif hasattr(obj, "__dict__"):
+                # If the object has a __dict__ attribute, serialize its __dict__
+                return serialize(obj.__dict__)
+            else:
+                # For other types, use default serialization
+                return obj
+
+        return json.dumps(serialize(obj))
+
+    psycopg.types.json.set_json_dumps(serialize_as_json)
```
miketheman, line 211: Another thought came up!

Since we're using sqlalchemy, should we be overriding the psycopg library itself, or setting s custom serializer during engine instantiation instead?
Refs: 
https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.JSON
https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.json_serializer

di:
> Docs: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/renderers.html#serializing-custom-objects

I did this, it's a little weird to initialize but works great.

> This serialization format loses the "type" of the originating dataclass. We can usually infer it from the values (like normalized_names pretty likely came from a ProjectName object, but that's not strictly true. It'd be great to preserve that link if we can somehow.

I took a look at this but I'm not really sure how to preserve the type without adding another serialization format on top of JSON?

> Did you look at [dataclasses-json](https://pypi.org/project/dataclasses-json/) as a potential solution? I haven't used it in anger in a while, but last time I did for API purposes, it did a decent job (and yes, it also loses the dataclass type)

I didn't.

> There's also [pydantic's recommended solution](https://docs.pydantic.dev/latest/concepts/dataclasses/#json-dumping) - did you look at that, and did it not work out? (and yes, it also loses the dataclass type)

I'm not sure why, but our `Caveat` subclasses don't have the `model_dump_json` attribute/function.

> Considering that in the specific case we have today - dump whatever caveats we have to an additional jsonb field for recording purposes - do you see a more generalized need for caveat serialization and deseralization to json?

Probably not deserialization, although I don't see why that wouldn't be possible with what we've currently got here.

miketheman:
> > This serialization format loses the "type" of the originating dataclass. We can usually infer it from the values (like normalized_names pretty likely came from a ProjectName object, but that's not strictly true. It'd be great to preserve that link if we can somehow.
> 
> I took a look at this but I'm not really sure how to preserve the type without adding another serialization format on top of JSON?

I think this might be what Donald was alluding to? We have a [`serialize()`](https://github.com/pypi/warehouse/blob/23cd71b664e77541def2eb5ac914cd0d48adc1ee/warehouse/macaroons/caveats/_core.py#L124-L127) function today that governs the format, and includes the "type" as an integer (the tag), so that satisfies "what caveat type is this?" question.

If we can use that function to create the thing we stuff into the database, then at least we're using the same format.

> > Did you look at [dataclasses-json](https://pypi.org/project/dataclasses-json/) as a potential solution? I haven't used it in anger in a while, but last time I did for API purposes, it did a decent job (and yes, it also loses the dataclass type)
> 
> I didn't.

No problemo - I don't know if iot would have helped with pg-specifics anyhow.

> > There's also [pydantic's recommended solution](https://docs.pydantic.dev/latest/concepts/dataclasses/#json-dumping) - did you look at that, and did it not work out? (and yes, it also loses the dataclass type)
> 
> I'm not sure why, but our `Caveat` subclasses don't have the `model_dump_json` attribute/function.
> 

That's because we're using dataclasses, and there's a guide for how we'd have to use the RootModel with the type as a prefix to using `model_dumps_json`, something like:

```python
RootModel[RequestUser](RequestUser(user_id='42')).model_dump_json()
```

which isn't super intuitive to me, so am happy there's other ideas here.

---

My biggest hesitation is around `psycopg.types.json.set_json_dumps(serialize_as_json)` - for some odd reason it feels too low level to me. There's something about this being the new default json renderer for the library that feels a bit heavy-handed, so if there's a better way to guard the serialization to only apply to objects that we've set custom `__json__` properties for, I think it'd make me feel safer.

di:
> My biggest hesitation is around `psycopg.types.json.set_json_dumps(serialize_as_json)` - for some odd reason it feels too low level to me. There's something about this being the new default json renderer for the library that feels a bit heavy-handed, so if there's a better way to guard the serialization to only apply to objects that we've set custom `__json__` properties for, I think it'd make me feel safer.

The issue is that the renderer/serializer doesn't get an object with custom `__json__`, it gets a `dict` or `list` and the object is buried inside that datastructure -- so we can't just put an if/else in our `serialize_as_json` function. The psycopg3 driver doesn't let extend the default serializer to handle custom cases (this was the thing I mentioned that was maybe possible with psycopg2, but doesn't seem to be possible with psycopg3).

Is the concern that the Pyramid JSON renderer is somehow wrong or would produce invalid JSON?

miketheman:
> The psycopg3 driver doesn't let extend the default serializer to handle custom cases (this was the thing I mentioned that was maybe possible with psycopg2, but doesn't seem to be possible with psycopg3).

I must have missed that. I guess if this is how psycopg3 expects us to handle custom serialization, then so be it!

> Is the concern that the Pyramid JSON renderer is somehow wrong or would produce invalid JSON?

We already override the Pyramid JSON renderer once, here:
https://github.com/pypi/warehouse/blob/52fdd6728a70736816c09a55af12be34201b5ef0/warehouse/config.py#L615-L623

Having this be a slightly different json serializer probably won't matter in the long run, since we're storing in the DB as jsonb (already compact, and sorting shouldn't matter), but I wanted to make sure it's noted.

Thanks for changing to the `serialize()` - if you're happy with the json serializer override, then let's go.

miketheman (approved):

PR #15526, opened 2024-03-03
Remove metadata backfill task

### LINKED ISSUE #8254: Expose the METADATA file of wheels in the simple API

di:
Removes the task added in #15368 to resolve #8254.


miketheman (approved):

AsayDeOnly:
Share link

AsayDeOnly:
Add a comment

PR #15544, opened 2024-03-06
Migrate pip-tools settings to shared config

### LINKED ISSUE #15406: Migrate pip-tools settings to shared config
We have multiple places where we use configuration flags to `pip-compile` like in `Makefile` and `bin/deps`.

Since pip-tools now supports configuration via `pyproject.toml`, move these configuration settings into a shared location and simplify the repetitions.

Note: Only move settings that are common across **all** usages - if there are task-specific settings, leave them associated with their task.

Refer to: https://github.com/jazzband/pip-tools#configuration

---

mbyrnepr2:
Specify a global configuration for the `pip-compile` option `--allow-unsafe` in the `pyproject.toml` file.

Closes #15406.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,4 +102,6 @@ filterwarnings = [
 ]
 
 [tool.pip-tools.compile]
+# TODO: This can be removed once it becomes the default.
+# See: https://github.com/jazzband/pip-tools/issues/989
 allow-unsafe = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,3 +100,6 @@ filterwarnings = [
     'ignore::warehouse.utils.exceptions.InsecureOIDCPublisherWarning',
     'ignore::warehouse.packaging.services.InsecureStorageWarning',
 ]
+
+[tool.pip-tools.compile]
+allow-unsafe = true
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ translations: .state/docker-build-base
 	docker compose run --rm base bin/translations
 
 requirements/%.txt: requirements/%.in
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -100,3 +100,6 @@ filterwarnings = [
     'ignore::warehouse.utils.exceptions.InsecureOIDCPublisherWarning',
     'ignore::warehouse.packaging.services.InsecureStorageWarning',
 ]
+
+[tool.pip-tools.compile]
```
woodruffw, line 104: Looks like `--allow-unsafe` will become the default in `pip-tools` 8.x (https://github.com/jazzband/pip-tools/issues/989), so maybe we can add a TODO here?

```suggestion
[tool.pip-tools.compile]
# TODO: This can be removed once it becomes the default.
# See: https://github.com/jazzband/pip-tools/issues/989
```

woodruffw (approved):
LGTM, thanks @mbyrnepr2! One nonblocking comment.

di (approved):

pyproject.toml
mbyrnepr2: That makes a lot of sense, I had the same thought but forgot to put the comment there. Thanks for the review!

miketheman:
Looks like this may have broken dependabot updates.

[They are on 7.3.0](033b4fd0b0000bcaf40481a2e299/python/helpers/requirements.txt#L2), which produces this error:

```
updater | 2024/03/12 16:56:54 ERROR <job_799098147> Usage: pip-compile [OPTIONS] [SRC_FILES]...
updater | Try 'pip-compile -h' for help.
updater | 
updater | Error: No such config key 'compile'.
```

There's a PR to update to 7.4.1, which hasn't been merged yet. https://github.com/dependabot/dependabot-core/pull/9256

PR #15545, opened 2024-03-06
fast(er) fail b2 interactions

ewdurbin:
currently our worker queue is spinning on b2 interactions that exponentially backoff for up to ~1300 seconds

lets just bail on these tasks for now and revert this when b2 is feelign better


di (approved):

ewdurbin:
Notes:

I can't find a way to tune this backoff when calling the API: https://github.com/Backblaze/b2-sdk-python/blob/86f28dc3f567e5a87c3a4b6b4e91d02a99c7805b/b2sdk/b2http.py#L571

It ultimately leads to 20 tries over an exponential backoff resulting in nearly 20 minute hangs for any b2 related task.

PR #15549, opened 2024-03-11
move parts of our db initialization into db service

ewdurbin:
Move as much of our database initialization into the db container itself to improve ergonomics for both `rstuf` and `warehouse`.

Running `make initdb` can now work without `make serve` first!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,14 +97,14 @@ translations: .state/docker-build-base
 requirements/%.txt: requirements/%.in
 	docker compose run --rm base bin/pip-compile --generate-hashes --output-file=$@ $<
 
+resetdb: .state/docker-build-base
+	docker compose up -d db
+	docker compose exec --user postgres db /docker-entrypoint-initdb.d/init-dbs.sh
+	$(MAKE) initdb
+
 initdb: .state/docker-build-base
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname ='warehouse';"
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS warehouse"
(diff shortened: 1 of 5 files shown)


ewdurbin:
@di try `make purge` first.

ewdurbin:
@di looks like you have an orphaned `db` from a `run` command somehow, pull 4fe3e33e2dc59bd6a4db68928a90c9289d0f9a1f and purge again.

Makefile
```diff
@@ -97,17 +99,28 @@ translations: .state/docker-build-base
...
-initdb: .state/docker-build-base
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname ='warehouse';"
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS warehouse"
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "CREATE DATABASE warehouse ENCODING 'UTF8'"
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "DROP DATABASE IF EXISTS rstuf"
-	docker compose run --rm web psql -h db -d postgres -U postgres -c "CREATE DATABASE rstuf ENCODING 'UTF8'"
-	docker compose run --rm web bash -c "xz -d -f -k dev/$(DB).sql.xz --stdout | psql -h db -d warehouse -U postgres -v ON_ERROR_STOP=1 -1 -f -"
-	docker compose run --rm web psql -h db -d warehouse -U postgres -c "UPDATE users SET name='Ee Durbin' WHERE username='ewdurbin'"
-	$(MAKE) runmigrations
+resetdb: .state/docker-build-base
```
di, line 102: My only thought is that it might be nice to pause the web/worker containers while this is running, because otherwise it throws up a bunch of errors because the DB is not fully initialized, but not a blocker!

di (approved):
Works great!

Makefile
ewdurbin: easy enough. ca071c9cb9fc38b8b992a380db2c046ea9e48db9

PR #15552, opened 2024-03-12
Add method to find a Macaroon from a request

facutuesca:
While working on https://github.com/pypi/warehouse/pull/15453 (warn users if they use an API token to upload to a project that has Trusted Publishing configured), I needed to get a Macaroon from a request sent to `forklift.legacy.file_upload()`. 

This Macaroon is needed to check if we have already warned a user about this token in particular (in which case we don't send the warning email).

The problem is that `file_upload()` only takes a request, and currently the only way of extracting the Macaroon from the request is to use an internal utility function `_extract_http_macaroon` located in `macaroons.security_policy`, doing something like:
```python
raw_macaroon = _extract_http_macaroon(request)
if raw_macaroon is not None:
    macaroon_service = request.find_service(IMacaroonService, context=None)
    try:
        macaroon: Macaroon = macaroon_service.find_from_raw(raw_macaroon)
```

This PR introduces a `macaroon_service.find_from_request()` method that does the above (it wraps the use of `_extract_http_macaroon` and `find_from_raw`). This also simplifies the `MacaroonSecurityPolicy`'s `identity()` and `permits()` methods, since both of them extracted the Macaroon from a request using the above procedure. Now they simply call `macaroon_service.find_from_request()`.

As part of this, the utility functions `_extract_http_macaroon` and `_extract_basic_macaroon` were moved from `macaroons.security_policy` to `macaroon.services` (since they are now only used by `services.find_from_request())`.

There is another subtlety to take into account: before, calling `_extract_http_macaroon` increased the metrics for the auth method used in the request. Since `_extract_http_macaroon` was called by both `MacaroonSecurityPolicy.identity()` and `MacaroonSecurityPolicy.permits()`, this means the metrics were duplicated:
https://github.com/pypi/warehouse/blob/dc1d9543ac4db87f1d0ba4d6e627f92eb0b1fde2/warehouse/macaroons/security_policy.py#L71-L74

In order to fix this, `find_from_request` takes a second parameter `increase_metrics: bool`, that specifies whether the authentication metrics should be increased or not. Then, `MacaroonSecurityPolicy` is modified so that only `MacaroonSecurityPolicy.permits()` pass `increase_metrics=True`, fixing the issue of duplicated metrics.

### DIFF
--- a/warehouse/macaroons/services.py
+++ b/warehouse/macaroons/services.py
@@ -9,9 +9,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
+import base64
 import datetime
 import uuid
+
+from typing import TYPE_CHECKING
 
 import pymacaroons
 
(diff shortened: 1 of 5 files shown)


facutuesca:
cc @di @woodruffw

warehouse/macaroons/services.py
```diff
@@ -9,20 +9,83 @@
...
+        metrics.increment(
+            "warehouse.macaroon.auth_method", tags=[f"method:{auth_method}"]
+        )
+
+    if auth_method == "basic":
+        return _extract_basic_macaroon(auth)
+    elif auth_method in ["token", "bearer"]:
+        return auth
+
+    return None
```
woodruffw, line 88: This LGTM behaviorally, but I think the `increase_metrics` flag is a bit of a wart 🙂 

Rather than internalizing metrics here, maybe we can increment this counter into an inline operation in the `identity(...)` method? That way we explicitly only ever call it once per authentication, without having to rely on a flag to do it.

facutuesca: The problem is that `identity()` would need to know how to extract the authentication method from the request (in order to know which metric to increase). So the question is, do we want to extract that logic from `_extract_http_macaroon()`, and make it available to `identity()` ?

woodruffw: Ah, right. Gross 🙂 

Hmm, in that case I suppose it's a bit cleaner to keep it behind that flag. I'll think about this some more.

di:
I'm thinking instead of this, we might want to return something other than a `User` here: https://github.com/pypi/warehouse/blob/dc1d9543ac4db87f1d0ba4d6e627f92eb0b1fde2/warehouse/macaroons/security_policy.py#L122

like a `MacaroonContext`, which has both the macaroon and user, similar to what we do for `OIDCContext`. That way, we could just do something like `request.identity.macaroon` when we want to get the macaroon associated with the request. I think that might be less invasive. @woodruffw, thoughts?

woodruffw:
> I think that might be less invasive. @woodruffw, thoughts?

I like that idea a lot! We might want to call it `UserTokenContext` or something like that for disambiguation, since technically an `OIDCContext` is also a "macaroon-bearing" context. But otherwise, I think that's much cleaner :slightly_smiling_face:

di:
Makes sense to me!

facutuesca:
@di @woodruffw Yeah that is definitely cleaner. I'm closing this PR, the new implementation is here: https://github.com/pypi/warehouse/pull/15581

PR #15553, opened 2024-03-12
feat(deps): include pyramid_openapi3

miketheman:
Introduce `pyramid_openapi3` as a way to:

- create API-specific definitions (`openapi.yaml`) that performs request and response validation, reducing the need to handle validations in-view
- provide an API documentation and try-out site (aka `explorer`)
- lay the foundation for more API endpoint development


docker-compose.yml
```diff
@@ -102,7 +102,7 @@ services:
 
   web:
     image: warehouse:docker-compose
-    command: gunicorn --reload -b 0.0.0.0:8000 --access-logfile - --error-logfile - warehouse.wsgi:application
+    command: gunicorn --reload --reload-extra-file=warehouse/api/openapi.yaml -b 0.0.0.0:8000 --access-logfile - --error-logfile - warehouse.wsgi:application
```
miketheman, line 105: This was a cool hack, and I've found that I probably want to use `hupper` instead, since the entire gunicorn process exits on a failure to start up with a broken openapi example.

warehouse/api/echo.py
```diff
@@ -79,34 +79,15 @@ def api_echo(request: Request):
     permission=Permissions.APIObservationsAdd,
     require_methods=["POST"],
 )
-def api_projects_observations(
-    project: Project, request: Request
-) -> HTTPAccepted | HTTPBadRequest:
+def api_projects_observations(project: Project, request: Request) -> dict:
     data = request.json_body
 
-    # TODO: Are there better mechanisms for validating the payload?
```
miketheman, line 87: These removals and their associated test case removals are now externalized to the "contract" of the OpenAPI document.

warehouse/api/openapi.yaml
```diff
@@ -0,0 +1,112 @@
+openapi: 3.0.0
```
miketheman, line 1: Supports OpenAPI v3.0.0 - not 3.1.0 yet, since the underlying library hasn't been updated yet.
Since it's brand new, we can proceed if we like it, and we're not bound to any specific schema yet, and I don't think we're using too many things that would change with the new version.

warehouse/api/openapi.yaml
```diff
@@ -0,0 +1,112 @@
...
+info:
+  title: Warehouse Management API
+  description: "**BETA** API for managing warehouse interactions"
+  version: 0.0.0
+servers:
+  - url: http://localhost
+    description: Local development server
```
miketheman, line 8: TODO: Is it possible to dynamically create these servers based on pypi/testpypi/localhost?

miketheman (commented):
Try it out by checking out the code, running it, and visiting `http://localhost/api/explorer`

docker-compose.yml
miketheman: hupper isn't the right answer here. I don't know what is, since "gunicorn failing to boot the worker, so master process shuts down" seems like the general right answer.

warehouse/api/openapi.yaml
miketheman: It is! Removing them does the trick.

warehouse/csp.py
```diff
@@ -48,8 +48,13 @@ def content_security_policy_tween(request):
         # We don't want to apply our Content Security Policy to the debug
         # toolbar, that's not part of our application and it doesn't work with
         # our restrictive CSP.
+        # We also want to exclude `/api/explorer` from the CSP, as it uses
+        # `unsafe-eval` and `unsafe-inline` for the Swagger UI.
         policy = _serialize(policy).format(request=request)
-        if not request.path.startswith("/_debug_toolbar/") and policy:
+        if (
+            not request.path.startswith("/_debug_toolbar/")
+            and not request.path.startswith("/api/explorer")
```
miketheman, line 56: I'm not certain if this exposes us to any specific vulnerability, but I've subscribed to https://github.com/swagger-api/swagger-ui/issues/7540 to watch to see if/when the specifics get resolved.

docker-compose.yml
miketheman: Adding `restart: on-failure` to my `web` worker in my docker compose override is what works best. if it crashes, it'll restart. yes, that will be an infinite loopp on my machine, but that works well enough for my local dev purposes.

warehouse/csp.py
ewdurbin: Is the intention that we would have https://pypi.org/api/explorer? Does it have to be hosted on pypi.org domain or could the swagger docs be generated and served from docs.pypi.org?

di: I think I agree with @ewdurbin -- this should probably live on a separate subdomain at least.

di (approved):

PR #15583, opened 2024-03-13
Add pytest-timeout

di:
Adds https://pypi.org/project/pytest-timeout/ to help debug hangs in CI.


di:
Was able to get the entire session to time out here: https://github.com/pypi/warehouse/actions/runs/8269077469/job/22623768007?pr=15583

Interestingly it shows that the `setup` time for all the longest test durations come from the same file (the file where the hang occurred).

PR #15584, opened 2024-03-13
Remove unused db field for ActiveState Trusted Publisher

th3coop:
This field never ended up being used so I'm removing it to declutter the warehouse code and DB a bit.


di (approved):

th3coop:
Thanks @di!

PR #15585, opened 2024-03-13
lets just try dockah

ewdurbin:
everything else involved (dev, prod) is docker, why not ci?

yeee, fucking, haw.

closes #15583.


.github/workflows/ci.yml
```diff
@@ -9,60 +9,102 @@ on:
...
+          cache-source: root-npm
+          cache-target: /root/.npm
+          skip-extraction: ${{ steps.cache-root-npm.outputs.cache-hit }}
+      - name: Build Docker Image
         run: |
-          pip install -U setuptools wheel pip
-          pip install -r requirements.txt --no-deps
-          pip install -r requirements/dev.txt
-          pip check
+            echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.repository_owner }} --password-stdin
```
di, line 82: Using stdin here to avoid writing the password to the log? Wouldn't GHA mask the token anyways?

warehouse/forklift/legacy.py
```diff
@@ -440,7 +440,7 @@ def process_formdata(self, valuelist):
 
 # TODO: Eventually this whole validation thing should move to the packaging
 #       library and we should just call that. However until PEP 426 is done
-#       that library won't have an API for this.
+#       that library won't have an API for this. ref: #14718
```
di, line 443: Unrelated change here.

.github/workflows/ci.yml
ewdurbin: This is just following guidance from GHA docs since we are using our own buildx to access gha cache type

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic

warehouse/forklift/legacy.py
di:
```suggestion
#       that library won't have an API for this.
```

ewdurbin: 8f9faa6bc

.github/workflows/ci.yml
di: Good enough for me!

PR #15588, opened 2024-03-13
Revert "Return `Macaroon` alongside `User` in `MacaroonSecurityPolicy…

di:
….identity` (#15581)"

This reverts commit 862a451bab1f0ccb61ce576fd0026f8132d06054.

I believe this is the cause of https://github.com/pypi/warehouse/issues/15586 and possibly https://github.com/pypi/support/issues/3740.

cc @facutuesca @woodruffw

PR #15581, opened 2024-03-13
Return `Macaroon` alongside `User` in `MacaroonSecurityPolicy.identity`

facutuesca:
Following the discussion here: https://github.com/pypi/warehouse/pull/15552#issuecomment-1992728849, this PR changes `MacaroonSecurityPolicy.identity` so that instead of returning a `User` in the case of API token-authenticated requests, it returns a `UserTokenContext` object, which contains both the `User` and the `Macaroon` associated to the request.

Issue #15586, opened 2024-03-13
"403 Invalid API Token: caveat must be an array" when using hatch publish


mcarans:
**Describe the bug**
I receive the following error when using hatch publish for a package that has been published fine many times before:
```
Error uploading to repository: https://upload.pypi.org/legacy/ - Client error '403 Invalid API Token: caveat must be an array' for url 'https://upload.pypi.org/legacy/'
```

I have not changed anything to do with the publishing (HATCH_INDEX_USER and HATCH_INDEX_AUTH are the same and I haven't touched pyproject.toml.

**Expected behavior**
Package is published

**To Reproduce**
Publish a package using hatch in a GitHub Actions workflow

**My Platform**
I'm running the publish workflow in GitHub Actions. You can see it here if it's any help:
https://github.com/OCHA-DAP/hdx-python-utilities/blob/main/.github/workflows/publish.yaml

Kelsey-Ethyca:
I am getting this same error using twine

di:
Thanks, I'm looking into this. Please avoid adding "+1" or "me too" comments and 👍 the top-level comment if you're affected instead.

di:
This should now be resolved.

PR #15590, opened 2024-03-14
Re-add `UserTokenContext`, with instance checks

woodruffw:
Reverts #15588, reviving https://github.com/pypi/warehouse/pull/15581.

~~WIP while I add both coverage and determine a backstop test approach here.~~


warehouse/accounts/__init__.py
```diff
@@ -54,23 +55,29 @@ def _user(request):
...
         return None
 
-    if not isinstance(request.identity, User):
+    if isinstance(request.identity, UserTokenContext):
+        # A UserTokenContext signals a user-created API token;
+        # take the underlying user.
+        return request.identity.user
+    elif isinstance(request.identity, User):
+        return request.identity
+    else:
```
woodruffw, line 64: NB: This will address https://github.com/pypi/warehouse/pull/15588#issuecomment-1996019812, by ensuring that `request.user` dispatches appropriately for `UserTokenContext`.

warehouse/macaroons/caveats/__init__.py
```diff
@@ -104,7 +104,7 @@ class RequestUser(Caveat):
     user_id: StrictStr
 
     def verify(self, request: Request, context: Any, permission: str) -> Result:
-        if not isinstance(request.identity, User):
+        if not isinstance(request.identity, UserTokenContext):
```
woodruffw, line 107: NB: This will address https://github.com/pypi/warehouse/pull/15588#issuecomment-1996036127

warehouse/macaroons/caveats/__init__.py
```diff
@@ -104,7 +104,7 @@ class RequestUser(Caveat):
...
             return Failure("token with user restriction without a user")
 
         if str(request.identity.id) != self.user_id:
```
woodruffw, line 110: NB: I've added `UserTokenContext.id` so that this line continues to work. However, it could just as easily be this instead:

```suggestion
        if str(request.identity.user.id) != self.user_id:
```

...and that might be more idiomatic than giving the `UserTokenContext` a DB ID attribute.

woodruffw: This looks like the only appearance of this pattern, so I'm going to remove the `UserTokenContext.id` property and go this route.

warehouse/accounts/__init__.py
di: This is a little gross, and makes me think `request.identity` should maybe never be a `User` and always be a `Context` that might have a user. I think this is OK for now though.

woodruffw: Yeah, agreed -- @facutuesca and I can look into a follow-on for that.

di:
There was another failure case in https://github.com/pypi/warehouse/issues/15586 which I couldn't root-cause: "caveat must be an array", which change covers that here?

warehouse/utils/security_policy.py
```diff
@@ -84,6 +85,8 @@ def authenticated_userid(self, request):
             #       more correct pattern before fixing this.
             if isinstance(ident, User):
                 return str(ident.id)
+            elif isinstance(ident, UserTokenContext):
+                return str(ident.user.id)
```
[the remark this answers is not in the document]
woodruffw: Makes sense. Want me to roll that into this changeset, or keep it for the next per https://github.com/pypi/warehouse/pull/15590#discussion_r1524976938?

woodruffw:
> There was another failure case in #15586 which I couldn't root-cause: "caveat must be an array", which change covers that here?

Hmm, that happens in `caveats._core.deserialize_obj`. Looking now...

warehouse/utils/security_policy.py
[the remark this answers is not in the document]
di: Let's keep it for a future refactor.

woodruffw:
Oh, I think that error happens as a variant of https://github.com/pypi/warehouse/pull/15588#issuecomment-1996019812 when the API token is in the "legacy" format: we call `_adapt_v1`, which then contains a `request.user` check which would fail because `request.user == None` before this fix.

Relevant bit:

```python

    # Our V1 token didn't have a way to specify that a token should be
    # restricted to a specific user, just that it was scoped to "the user",
    # which the user was whoever the token was linked to in the database.
    # Our new tokens strengthens that to validate that the linked user
    # matches who it is expected to be, but since we don't have that
    # data for V1 tokens, we'll just use the current user.
    if permissions == "user":
        request = get_current_request()

        # If we don't have a current request, then we can't validate this
        # token.
        if request is None:
            return None

        # If we don't have a user associated with this request, then we
        # can't validate this token.
        if request.user is None:
            return None

        return [3, str(request.user.id)]
```

So we'd return `None`, from `_adapt_v1`, which would then cause the "must be an array" error.

woodruffw:
I've added some tests for the behavior here, but I'm not immediately sure how to "backstop" this -- maybe some additional `test_legacy` tests that ensure uploads succeed when the "user" identity is of both sorts?

woodruffw:
This should be good to go now: I've added a layer of configurations on `test_successful_upload` that ensures `identity=UserTokenContext` works as expected.

di (approved):
@woodruffw Can you rebase & get tests passing?

woodruffw:
> @woodruffw Can you rebase & get tests passing?

Should be good now!

PR #15595, opened 2024-03-14
Revert "lets just try dockah"

ewdurbin:
Reverts pypi/warehouse#15585, as promising as its CI runs look there is an issue handling forks that I do not have time to debug.


di (approved):
womp

ewdurbin:
The summary here is that the permissions for `package` are automatically limited to `read` for pull requests from forks.

See "Maximum access for pull requests from public forked repositories" https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

PR #15596, opened 2024-03-14
Handle forks with GHCR

di:
Maybe?


ewdurbin:
Not possible with GHA tokens:

See "Maximum access for pull requests from public forked repositories" https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Consider #15603 instead!

Issue #7955, opened 2020-05-15
Content originating from JavaScript not marked for translation


nlhkabu:
We have some content that is generated from JS.  It is not currently marked for translation.

- Copy tooltip
- Time that has been humanized, e.g. "2 hours ago"
- Password strength error messages
- Password matching messages

There are probably more.  We need to go through the JS source to identify these.

cofiem:
The copy tooltip, `timeago`, and password match strings are addressed by merging #15612 The password strength phrases are tricky to translate due to being mixed with `zxcvbn` output.

Issue #7455, opened 2020-02-27
Missing translations


NathanBnm:
Some content is still not localizable. This list compiles the missing content I noticed untranslatable. Everything is open to discussion :smile:

- Package classifiers (open discussion here #6882)
- "Locale updated" banner when switching language
- Status message "All Systems Operational" in the footer
- Size units (kB, GB etc.)
- "Our sponsors" page
- "Terms of use" page

di:
I fixed the "locale updated" flash message in https://github.com/pypa/warehouse/pull/7452.

The status message is loaded on the frontend from StatusPage so will probably not be localizable.

Some of these are due to https://github.com/pypa/warehouse/issues/6825 as well.

mondeja:
Status message "All Systems Operational" in the footer is translatable.

di:
> Status message "All Systems Operational" in the footer is translatable.

It's not, these strings come from an API call to an external service that happens with client-side JavaScript:

https://github.com/pypa/warehouse/blob/0ad848d266619c7961bcc0cc1cdad45fc9495d12/warehouse/templates/base.html#L310-L312

https://github.com/pypa/warehouse/blob/0ad848d266619c7961bcc0cc1cdad45fc9495d12/warehouse/static/js/warehouse/utils/statuspage.js#L14-L26

di:
@NathanBnm That page is mostly marked for translation, although there are a few strings that aren't: https://github.com/pypa/warehouse/blob/27b11434358a8ff98db29a8c701a5b7c1289f207/warehouse/templates/pages/sponsors.html

Unfortunately we can't currently translate the sponsor descriptions.

di:
Additionally plaintext emails and email subjects are not being translated because `.txt` files are not being searched for translations:

https://github.com/pypa/warehouse/blob/525002445b2f97024e229651259eb9409c9b582d/babel.cfg#L1-L2

Issue #6673, opened 2019-09-20
Localise JS strings


nlhkabu:
Now that we have l10n support, we need to review our JS files and mark any strings for translation.

woodruffw:
I believe this is out of scope for the OTF work -- we scoped for localizing views and templates, localizing JS will require is to construct a separate (probably client-side) localization system and ensure that the two don't fight.

IMPO, the right approach here is to audit the JS for English strings and remove them entirely (either by moving functionality to the backend or by returning error codes/hitting a renderer somehow to signal which string to display). That would avoid the need for two separate localization systems, and probably wouldn't involve any substantial changes (from what I've seen, there are relatively few English strings in the JS).

miketheman:
The only strings in JS files I was able to identify as of hash 17f76cf36 live in these locations:

- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/clipboard_controller.js#L31
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/password_match_controller.js#L31
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/password_match_controller.js#L35
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/password_strength_gauge_controller.js#L26
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/password_strength_gauge_controller.js#L34
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/controllers/password_breach_controller.js#L47
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/utils/timeago.js#L31
- https://github.com/pypi/warehouse/blob/17f76cf365adc1807a84e6ca481467e73dbe87c8/warehouse/static/js/warehouse/utils/timeago.js#L35-L41

Most are trivial, and could potentially be localized from existing resources, leveraging https://www.i18next.com/ or similar.

[...]

PR #15612, opened 2024-03-17
Allow translating from javascript files

cofiem:
This PR enables translating strings in javascript.

It demonstrates the js translation in use by applying it to the `timeago.js` js util.

Edit: added overview of the two mutually-exclusive approaches.

# Approach: webpack plugin the generates js message bundles

This is the current state of this PR.

I attempted an approach similar to https://www.npmjs.com/package/@zainulbr/i18n-webpack-plugin
i.e. embed the translations into separate js bundles, one for each language.

Unfortunately, the existing webpack plugins were not suitable (issues with webpack 5 and/or did not work for the warehouse setup).
However, I did manage to create a webpack plugin that does the job (in webpack.plugin.localize.js).
The plugin checks the plural forms and embeds them into the generated js bundles.

So, the development process with these changes:

- Use gettext or ngettext as appropriate in js files
- Run make translations: pybabel extracts messages from js files, then a python script generates messages.json files for only the KNOWN_LOCALES and only includes translations that are needed in js
- Run make static_pipeline: the webpack plugin embeds the translated data from the messages.json files
- At runtime, the matching js bundle for the current locale is set in base.html, and gettext.js is initialised with the translation data. It can determine plurals and do string templating.

These are the benefits of this approach:

- Uses the existing translation and js tooling - make, pybabel, locale files, webpack
- Enables plurals and template placeholders
- Limits or does not use the hard to cache routes nor additional http requests
- Limits the number of unmaintained packages required, and works with webpack 5
- Straightforward to maintain and use as part of contributions

# Approach: request from client to server to get translation for a given message code

A js util `fetch-gettext.js` that requests the translation code returns a promise, which contains the translation text from the server, using the locale set on the server-side.

[view the changes for this approach](https://github.com/pypi/warehouse/compare/main...cofiem:warehouse:0d94a8625ce11c95d4516a93f24cca85baf22c6d#files_bucket).

Changes:
- enables extracting strings from javascript using pybabel, 
- adds a route that renders the appropriate translation using the existing machinery, and 
- adds js util for obtaining the translated string.

Considerations:
- may make many http requests for translations

# Notes

- Pybabel seems to require particular js function names and the config to change them does not work, similar to [this pybabel issue](https://github.com/python-babel/babel/issues/439). This is ok, I've used the names expected by pybabel.

Relates to issues:
- https://github.com/pypi/warehouse/pull/15087
- https://github.com/pypi/warehouse/issues/6673
- https://github.com/pypi/warehouse/issues/7955
- https://github.com/pypi/warehouse/issues/7455


di:
Very cool! I like this approach a lot, I think this will definitely work.

One thing I'm wondering: since we probably won't have a _ton_ of JS translations, would it make sense to load them all for a given locale in a bundle via a single request, as opposed to making `N` different requests?

cofiem:
> One thing I'm wondering: since we probably won't have a _ton_ of JS translations, would it make sense to load them all for a given locale in a bundle via a single request, as opposed to making `N` different requests?

I'm not sure I understand this.

I found a way to use the server-side rendering of translations because:
- I can't see that there's a way to get just the js translation strings to make a bundle of translation strings
- each translation can have up to 6 variants, which are chosen based on the [`num` parameter](https://github.com/pypi/warehouse/pull/15612/files#diff-9fa8f7ec0b2cfbcd2b19ff834ee11d061da3cab92ab65905f08c63bd10b94ef1R21)
- some translations have placeholders in the string, and I didn't want to have to find or make a way to fill the placeholders in js

If you're concerned about the server load or frequency of requests, the browser cache seems to work well, and there's likely some additional caching to be done in the server-side view that could reduce the need to execute the [`localiser.pluralize` or `localizer.translate`](https://github.com/pypi/warehouse/pull/15612/files#diff-a2bee5a31fcca1bfbf9dafa20e74f47dc0296a4953fb19cb4ca774dc003551bcR320).

Please let me know if I've misunderstood.

cofiem:
I managed to get the [plurals extracted](https://github.com/pypi/warehouse/pull/15612/files#diff-7cc1dd5466d4caf26bedf81acb95231b510147da1a3ed4ff5a0ae3f8fa3eeb3aR712).

It looks like python-babel expects [particular function names](https://github.com/pypi/warehouse/pull/15612/files#diff-9fa8f7ec0b2cfbcd2b19ff834ee11d061da3cab92ab65905f08c63bd10b94ef1R33).

TODO:
- I'll include some screenshots of the js translations in action.
- I may also have time to go through and make the rest of the text that's been identified in the other issues as translatable.
- fix the couple of failing tests

Question:
- Is there an existing way to cache the output of a view that depends on the `request.params`? That would work well for caching the translation view on the server side.
Maybe [`origin_cache`](https://github.com/pypi/warehouse/blob/main/warehouse/cache/origin/__init__.py#L58)?

di:
> If you're concerned about the server load or frequency of requests, the browser cache seems to work well, and there's likely some additional caching to be done in the server-side view that could reduce the need to execute the [localiser.pluralize or localizer.translate](https://github.com/pypi/warehouse/pull/15612/files#diff-a2bee5a31fcca1bfbf9dafa20e74f47dc0296a4953fb19cb4ca774dc003551bcR320).

Yes, I was thinking about balancing the number of requests with the size of the requests. It's probably an over-optimization though, our CDN will cache these responses by default so it's not the load on our backend that I was concerned with, just the number of requests the user's browser would have to make after the page has loaded. I think it's fine to move forward as-is, just wanted to think through it a bit.

> Is there an existing way to cache the output of a view that depends on the request.params? That would work well for caching the translation view on the server side.

By default, we cache all requests, but strip all URL parameters when caching a given view (with some exceptions). We can add an exception for this view. This happens here:

https://github.com/pypi/infra/blob/9568346c6781a6d9e754e9edc3883dd34febd97b/terraform/warehouse/vcl/main.vcl#L103-L126

cofiem:
The only remaining question is what @di was asking - is the number of requests made by the browser a problem?

Some example screenshots of the js translations in use.

Screenshot 2024-03-24 181338
Screenshot 2024-03-24 181556
Screenshot 2024-03-24 181952
Screenshot 2024-03-24 182303

cofiem:
Test errors:

1. 
> \>               msg += " " + " ".join(results["feedback"]["suggestions"])
> E               TypeError: unsupported operand type(s) for +=: 'LazyString' and 'str'

This one could be fixed by adding `str(msg)`, but there might be a better way?

```diff
- msg += " " + " ".join(results["feedback"]["suggestions"])
+ msg = str(msg) + " " + " ".join(results["feedback"]["suggestions"])
```

2. 
> \>       return request.localizer.translate(_translation_factory(message, **kwargs))
> E       AttributeError: 'NoneType' object has no attribute 'localizer'

I'm not sure how to fix this error. Other form tests seem to be able to set `request=pretend.stub()`, but the Validator directly uses `from warehouse.i18n import localize as _`, which errors at `request.localizer`.

warehouse/forms.py
```diff
@@ -68,7 +68,7 @@ def __call__(self, form, field):
             msg = (
                 results["feedback"]["warning"]
                 if results["feedback"]["warning"]
-                else "Password is too easily guessed."
+                else _("Password is too easily guessed.")
```
di, line 71: I actually think we shouldn't translate this string, because it will be mixed with other non-localized strings from zxcvbn, which would be confusing.
```suggestion
                # Note: we can't localize this string because it will be mixed
                # with other non-localizable strings from zxcvbn
                else "Password is too easily guessed."
```

miketheman:
Not to completely derail this direction, but had you looked at https://www.npmjs.com/package/@zainulbr/i18n-webpack-plugin or such at all? We currently use Webpack to generate bundles, so it'd be cool if we could have those localized and served once, instead of on demand for each page.

cofiem:
@miketheman commit https://github.com/pypi/warehouse/pull/15612/commits/8fb15ddf5964ede68e1a6f6afee1a2c996c5bee4 is a draft of an alternate approach that generates a `messages.json` file as part of `make translations`.
It uses `gettext.js` to show the translations.

It's a draft (I couldn't get gettext.js to import / require properly).

Comparison:

server-side approach:
pros:
- no changes to existing translation generation
- already implemented

cons:
- may make many http requests for translations

client-side approach:
pros:
- generates js translation file as part of translation generation
- loads all translations needed in js as part of page https requests
- no additional requests

cons:
- additional js dependency
- includes all translations in page, even though only one will ever be needed
- requires changing the translation generation

cofiem:
I had a look at that webpack plugin, and some other approaches.
They all seemed to assume that the translations were only done in js, and so included all translations.
Since only a very few translations are needed in js, gettext.js seemed like the most appropriate approach.

miketheman:
@cofiem Taking another look here at the alternative approach from commit https://github.com/pypi/warehouse/commit/8fb15ddf5964ede68e1a6f6afee1a2c996c5bee4 

1. It sounds like `messages.json` would contain all translated strings for all languages, is that correct?

2. If so, can we fingerprint the file (again webpack might help here, it does this for other assets already) so we don't have worry about cDN/cache busting?

3. On failure to load either the js lib or messages.json (degrades gracefully), would any strings end up on the page as a fallback?

If the answer to all of these "yes", it seems like this approach would be better for now.

cofiem:
Alright, I'll try to explain what I've done.

I attempted an approach similar to https://www.npmjs.com/package/@zainulbr/i18n-webpack-plugin
i.e. embed the translations into separate js bundles, one for each language.
Unfortunately, the existing webpack plugins were not suitable (issues with webpack 5 and/or did not work for the warehouse setup).
However, I did manage to create a webpack plugin that does the job (in `webpack.plugin.localize.js`).

I've just encountered a problem with `gettext.js`: it does [a kind of js eval](https://github.com/guillaumepotier/gettext.js/blob/v2.0.2/lib/gettext.js#L102), which is not allowed by the CSP: ` Uncaught EvalError: call to Function() blocked by CSP`.
I think this is solvable - I'll see how else the plural forms could be used.

So, the development process with these changes:

1. Use gettext or ngettext as appropriate in js files
2. Run make translations: pybabel extracts messages from js files, then a python script generates messages.json files for only the KNOWN_LOCALES and only includes translations that are needed in js
3. Run make static_pipeline: the webpack plugin embeds the translated data from the messages.json files
4. At runtime, the matching js bundle for the current locale is set in `base.html`, and gettext.js is initialised with the translation data. It can determine plurals and do string templating.

These are the benefits of this approach:

- Uses the existing translation and js tooling - make, pybabel, locale files, webpack
- Enables plurals and template placeholders
- Limits or does not use the hard to cache routes nor additional http requests 
- Limits the number of unmaintained packages required, and works with webpack 5
- Straightforward to maintain and use as part of contributions

firefly9966 (approved):

cofiem:
Hi @di @miketheman , when you're able, please review this PR (and the associated PR #15087 ).
I am willing to resolve the conflicts on both PRs, and continue working to get them over the line.

di (approved):
This looks just about perfect to me. I'll leave a bit more time for @miketheman to review if he's able, otherwise I'll merge!

bin/translations
```diff
@@ -11,3 +11,4 @@ export LANG="${ENCODING:-en_US.UTF-8}"
 set -x
 
 make -C warehouse/locale/ translations
+python bin/translations_json.py
```
di, line 14: I _think_ this should probably happen as part of our static build instead: https://github.com/pypi/warehouse/blob/1185eef19866ff464a0db09aed53525bd2b7d46f/Dockerfile#L21-L33

My thinking is that the `.json` files will only get updated when `make translations` is run currently, but that's not something that Weblate runs when it gets a new translation. So we'll have an issue where a PR like https://github.com/pypi/warehouse/pull/16426 lands with new translations, but all the `messages.json` files will be stale, even though the 'Translations' check will pass. We'll have to manually run `make translations` once in a while to bring all the `messages.json` files up to date in the source repo, which is annoying.

I think if we generate these files when we build the rest of our static assets, it shouldn't be necessary to include them in the source repository either, as they'll just be an artifact of the build step before the webpack bundles get built.

di (commented):
Took one last look and I have one outstanding issue here before we can merge.

bin/translations
cofiem: I've updated `webpack.plugin.localize.js` so that the translation data needed in the js files is generated in the webpack plugin, instead of in the separate `bin/translations_json.py` in `make translations`. This should mean that the `messages.json` files are not needed and the Weblate PRs will generate one js bundle per language.

webpack.config.js
```diff
@@ -16,25 +16,25 @@
...
+const {WebpackLocalisationPlugin, allLocaleData} = require("./webpack.plugin.localize.js");
 
 /* Shared Plugins */
 
 const sharedCompressionPlugins = [
   new CompressionPlugin({
     filename: "[path][base].gz",
     algorithm: "gzip",
-    compressionOptions: { level: 9, memLevel: 9 },
+    compressionOptions: {level: 9, memLevel: 9},
```
miketheman, line 37: thought: does our linter not check and enforce these kinds of changes?

Makefile
```diff
@@ -104,7 +104,7 @@ licenses: .state/docker-build-base
 deps: .state/docker-build-base
 	docker compose run --rm base bin/deps
 
-translations: .state/docker-build-base
+translations: .state/docker-build-base .state/docker-build-static
```
miketheman, line 107: **question:** now that the build is part of webpack, is this dependency necessary? From [the inline doc](https://github.com/pypi/warehouse/pull/15612/files?w=1#diff-55479889313695828cc288bac91197837669dcc5d650b665c09bdb8b4dce211cR20) it seems like `make static` should depend on translations instead?‏

tests/frontend/messages_access_test.js
```diff
@@ -0,0 +1,109 @@
...
+      "My default message.": "My translated message.",
+      "My %2 message with placeholders %1.": "My translated %1 message with placeholders %2",
+      "My message with plurals": ["My translated message 1.", "My translated messages 2."],
+      "My message with plurals %1 again": ["My translated message 1 %1.", "My translated message 2 %1"],
+    };
+    const pluralForms = function (n) {
+      let nplurals, plural;
+      nplurals = 2; plural = n > 1;
+      return {total: nplurals, index: ((nplurals > 1 && plural === true) ? 1 : (plural ? plural : 0))};
+    };
```
miketheman, line 80: question: this looks a lot like a parametrized test, is that the case here? If so, would it make more sense to use [jest's `.each`](https://jestjs.io/docs/api#each) to express the individual tests to reduce the amount of logic baked in to the test case?

warehouse/static/js/warehouse/utils/messages-access.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param singular {string} The default string for the singular translation.
+ * @param extras {string} Additional values to put in placeholders.
+ * @returns {string} The translated text.
+ */
+export function gettext(singular, ...extras) {
+  return ngettext(singular, null, 1, ...extras);
+}
+
+export function ngettextCustom(singular, plural, num, extras, data, pluralForms) {
+  // this function allows for testing
```
miketheman, line 76: question: is this comment still accurate, if `ngettext` is basically a wrapper for this, with some defaults?

warehouse/static/js/warehouse/utils/messages-access.js
```diff
@@ -0,0 +1,109 @@
...
+ * limitations under the License.
+ */
+
+// the value for 'messagesAccessLocaleData' is set by webpack.plugin.localize.js
+// default is en
+var messagesAccessLocaleData = {"": {"language": "en", "plural-forms": "nplurals = 2; plural = (n != 1)"}};
+
+// the value for 'messagesAccessPluralFormFunction' is set by webpack.plugin.localize.js
+// default is en
+var messagesAccessPluralFormFunction = function (n) {
```
miketheman, line 20: question: should these be `var` or `const/let`? (again, I wonder what's happening with the linter...)

webpack.config.js
```diff
@@ -61,68 +61,72 @@ const sharedCSSPlugins = [
...
-];
+    }
+    // if the filename matches .css or .css.map, add a prefix of css/
+    if (file.name.match(/\.css(\.map)?$/)) {
+      file.name = `css/${file.name}`; // eslint-disable-line no-param-reassign
+    }
+    return file;
+  };
+// Refs: https://github.com/shellscape/webpack-manifest-plugin/issues/229#issuecomment-737617994
+// publicPath: "",
```
miketheman, line 83: question: why is this previously-set value commented out? If it's no longer valid, let's remove it, along with the comment above as to why it's there, or move it to where it's now being set, or extract it to a constant that can be easier commented and reused.

bin/translations
di: I tested this and it works great. Thank you for addressing it!

Makefile
di: Agreed, I don't think this is necessary.

webpack.plugin.localize.js
```diff
@@ -0,0 +1,154 @@
...
+          parser.state.current.addDependency(dep);
+          return true;
+
+        } else if (statement.type === "VariableDeclaration" &&
+          statement.declarations.length === 1 &&
+          statement.declarations[0].id.name === "messagesAccessPluralFormFunction") {
+          const initData = statement.declarations[0].init;
+          const pluralForms = self.localeData[""]["plural-forms"];
+          const denied = new RegExp(
+            "[~`^$_\\[\\]{}\\\\'\"\\.#@\\f\\n\\r\\t\\v\\u00a0\\u1680\\u2000-\\u200a\\u2028\\u2029\\u202f\\u205f\\u3000\\ufeff]+");
```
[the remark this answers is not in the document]
cofiem: addressed in 8e2e3f56848086c55c90190bc63353546c0e9e03

warehouse/admin/static/js/warehouse.js
```diff
@@ -112,8 +113,10 @@ document.querySelectorAll(".copy-text").forEach(function (element) {
       $(element).tooltip("hide")
         .attr("data-original-title", "Click to copy!");
     }, 1000);
+
+    const copiedText = gettext("Copied");
     $(element).tooltip("hide")
-      .attr("data-original-title", "Copied!")
+      .attr("data-original-title", copiedText)
```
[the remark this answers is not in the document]
cofiem: addressed in 0f7196d844c25d6807889891a88183b8bef6eec9 and 010fdd26fe0d522b337b731bfcac895da052befd

webpack.config.js
cofiem: (might be?) addressed in 7d1e8cbeb8b1123ddc241a852b5235bc4956039c

warehouse/static/js/warehouse/utils/messages-access.js
cofiem: addressed in d32b4472ebf326109af61e0cce90e2f7015589c0

warehouse/static/js/warehouse/utils/messages-access.js
cofiem: addressed in d32b4472ebf326109af61e0cce90e2f7015589c0

webpack.config.js
cofiem: addressed in 0f13ec1e6da34b26e9bc7a97c9598aff2c9adae3

webpack.plugin.localize.js
```diff
@@ -0,0 +1,154 @@
...
+      for (const key in translations) {
+        if (key === "") {
+          continue;
+        }
+        const value = translations[key];
+        const refs = value.comments.reference.split("\n");
+        if (refs.every(refLine => !refLine.includes(".js:"))) {
+          continue;
+        }
+        result[value.msgid] = value.msgstr;
```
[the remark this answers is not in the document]
cofiem: (partly?) addressed in fd0dd802d7523f5d1054925138d55d55afca035e

Makefile
```diff
@@ -81,7 +81,7 @@ tests: .state/docker-build-base
 static_tests: .state/docker-build-static
 	docker compose run --rm static bin/static_tests $(T) $(TESTARGS)
 
-static_pipeline: .state/docker-build-static
+static_pipeline: translations .state/docker-build-static
```
di, line 84:
```suggestion
static_pipeline: .state/docker-build-static
```

I think we can assume `translations` is always up to date when doing the static build.

webpack.plugin.localize.js
```diff
@@ -0,0 +1,166 @@
...
+        if (key === "") {
+          continue;
+        }
+        const value = translations[key];
+        const refs = value.comments.reference.split("\n");
+        if (refs.every(refLine => !refLine.includes(".js:"))) {
+          continue;
+        }
+        // The 'msgid' and 'msgstr' must be strings.
+        result[value.msgid.toString()] = value.msgstr.toString();
```
di, line 105: I think we need to take this a step further, because these values will also be embedded directly into the UI, sometimes as attributes on DOM elements. I think we would be vulnerable to XSS and script injection that could embedded in a translation string. Most places we use `textContent` which protects us, but we don't have any guarantees that all translated strings in JS are injected via `textContent`. I think this might need to be something like:

```suggestion
        result[value.msgid] = value.msgstr
            .replace(/&/g, "&amp;")
            .replace(/</g, "&lt;")
            .replace(/>/g, "&gt;")
            .replace(/"/g, "&quot;")
            .replace(/'/g, "&#39;");
```
I think I'm missing what `toString` would be protecting us from here, can you explain?

cofiem: Ah, I was assuming that `.textContent` was always used. In that case, your suggestion makes sense.

di (approved):
LGTM

di:
Looks like this is causing build failures: https://github.com/pypi/warehouse/actions/runs/12811292443/job/35720280486?pr=17373

di:
I think this should resolve it: https://github.com/pypi/warehouse/pull/17424

PR #15632, opened 2024-03-19
Remove depot/setup-action with.oidc

jacobwgillespie:
This removes the `with.oidc` option from the `depot/setup-action` - the Depot CLI should now natively handle this for public PRs ✨

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,6 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up Depot CLI
         uses: depot/setup-action@v1
-        with:
-          oidc: true
       - name: Build image
         id: build
         uses: depot/build-push-action@v1


miketheman:
@jacobwgillespie Thanks!

Seeing these in this run's log:

```
Unable to exchange GitHub OIDC token for temporary Depot token: Error: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
Attempting to acquire open-source pull request OIDC token
Waiting for OIDC auth challenge 03d3a118-00f0-4d55-85fa-1766f0a63861
Waiting for OIDC auth challenge 03d3a118-00f0-4d55-85fa-1766f0a63861
Waiting for OIDC auth challenge 03d3a118-00f0-4d55-85fa-1766f0a63861
...
```

It still succeeds, but is this still trying to do something and times out?

jacobwgillespie:
Hey yeah those logs are somewhat expected (I need to silence the initial error) - the way OIDC works for external PR workflows is that a challenge token is printed to the logs then the OIDC issuer connects to the workflow log stream and verifies the token value is being printed.

The initial error can be ignored, I need to fix the action so it doesn't print it

miketheman:
Thanks @jacobwgillespie ! I'm noticing it kinda spin on that step, and take ~3-4 minutes to continue - which drives up CI times. Is that also something that will be addressed soon?

miketheman:
Looks like this removal had other negative impact on other steps? https://github.com/pypi/warehouse/actions/runs/8346865483/job/22845432075

di:
Looks like we're trying to pass the output of `depot pull-token` to the next job: https://github.com/pypi/warehouse/blob/4b94bfe2d96df8e48436557ca0faf895dbdf629e/.github/workflows/ci.yml#L37-L39

But it's empty now, causing this to fail: https://github.com/pypi/warehouse/blob/4b94bfe2d96df8e48436557ca0faf895dbdf629e/.github/workflows/ci.yml#L63-L66

di:
Maybe only sometimes? Seems to have worked fine for #15631.

PR #15641, opened 2024-03-20
Warn users when API token is used in Trusted Publishing project (take 2)

facutuesca:
If a user uploads files using an API token to a project that has a Trusted Publisher configured, send an email to all the project's users notifying them of the API token use, suggesting they remove it and use Trusted Publishing instead.

This warning is only sent once per (project, API token) combination.

This is a re-implementation of https://github.com/pypi/warehouse/pull/15453, taking into account the following requirements:
1. Send warnings for API tokens both project-scoped and account-level scoped.
2. Send only one email per `(API token, Project)` combination
3. Send the email to all of the project's maintainers
4. The email should be immediately actionable, including a link to the URL where the token owner can delete the token.

The actual logic being added is small, most of the line changes in the PR are tests and the new email body text being added.

cc @di @woodruffw

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -46,12 +46,14 @@ from warehouse.admin.flags import AdminFlagValue
 from warehouse.authnz import Permissions
 from warehouse.classifiers.models import Classifier
 from warehouse.email import (
+    send_api_token_used_in_trusted_publisher_project_email,
     send_gpg_signature_uploaded_email,
     send_two_factor_not_yet_enabled_email,
 )
 from warehouse.events.tags import EventTag
 from warehouse.forklift import metadata
 from warehouse.forklift.forms import UploadForm, _filetype_extension_mapping
+from warehouse.macaroons.models import Macaroon
 from warehouse.metrics import IMetricsService
 from warehouse.packaging.interfaces import IFileStorage, IProjectService
 from warehouse.packaging.models import (
@@ -606,6 +608,24 @@ def file_upload(request):
                 else allowed.msg
             )
         raise _exc_with_message(HTTPForbidden, msg)
+
+    # If this is a user identity (i.e: API token) but there exists
+    # a trusted publisher for this project, send an email warning that an
+    # API token was used to upload a project where Trusted Publishing is configured.
+    # Only do this once per (API token, project) combination.
+    if request.user and project.oidc_publishers:
+        macaroon: Macaroon = request.identity.macaroon
+        # If we haven't warned about use of this particular API token and project
+        # combination, send the warning email
+        if macaroon not in project.macaroons_tp_warning:
+            send_api_token_used_in_trusted_publisher_project_email(
+                request,
+                project.users,
+                project_name=project.name,
+                token_owner_username=request.user.username,
+                token_name=macaroon.description,
+            )
+            project.macaroons_tp_warning.append(macaroon)
 
     # Update name if it differs but is still equivalent. We don't need to check if
     # they are equivalent when normalized because that's already been done when we
--- a/warehouse/migrations/versions/56b3ef8e8af3_add_project_macaroon_warning_table.py
+++ b/warehouse/migrations/versions/56b3ef8e8af3_add_project_macaroon_warning_table.py
@@ -1,0 +1,51 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+add_project_macaroon_warning_table
+
+Revision ID: 56b3ef8e8af3
+Revises: 93a1ca43e356
+Create Date: 2024-03-11 19:41:22.997939
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "56b3ef8e8af3"
+down_revision = "93a1ca43e356"
+
+
+def upgrade():
+    op.create_table(
+        "project_macaroon_warning_association",
+        sa.Column("macaroon_id", sa.UUID(), nullable=False),
+        sa.Column("project_id", sa.UUID(), nullable=False),
+        sa.Column(
+            "id", sa.UUID(), server_default=sa.text("gen_random_uuid()"), nullable=False
+        ),
+        sa.ForeignKeyConstraint(
+            ["macaroon_id"],
+            ["macaroons.id"],
+            onupdate="CASCADE",
+            ondelete="CASCADE",
+        ),
+        sa.ForeignKeyConstraint(
+            ["project_id"],
+            ["projects.id"],
+        ),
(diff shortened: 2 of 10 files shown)


warehouse/templates/email/api-token-used-in-trusted-publisher-project/subject.txt
```diff
@@ -0,0 +1,21 @@
...
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/subject.txt" %}
+
+{% block subject %}
+{% trans project_name=project_name %}
+User API token used in project {{ project_name }} with Trusted Publishing enabled
```
woodruffw, line 19: Extreme nitpick, but we might want to tweak this a bit: this subject could be interpreted as "business as usual," when what we're trying to signal to the user is an unusual case. Maybe this?

```suggestion
Manually configured API token used in project {{ project_name }} despite Trusted Publishing enabled
```

That's a mouthful, though 😅 

(No strong opinion here.)

facutuesca: Yeah I agree that we need the subject to convey that there was "maybe unexpected/undesirable" behavior. Looking at the other email subjects in the repo, none have anything explicit like `WARNING: such and such happened`.
I do like your use of `despite`, since it shows that what was observed goes against (usual) expectations.

To reduce verbosity, I think we can get away we removing `Manually configured` since, from the POV of the user, all tokens are manually configured (the short-term tokens minted by the Trusted Publisher process are invisible to the user). So, something like:

`API token used to upload to project {{ project_name }} despite Trusted Publishing enabled`

?

woodruffw: SGTM!

facutuesca: fixed!

warehouse/templates/email/api-token-used-in-trusted-publisher-project/subject.txt
```diff
@@ -0,0 +1,21 @@
...
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/subject.txt" %}
+
+{% block subject %}
+{% trans project_name=project_name %}
+API token used to upload to project {{ project_name }} despite Trusted Publishing enabled
+{% endtrans %}
+{% endblock %}
```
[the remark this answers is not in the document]
facutuesca: Fixed! I also opened a PR with a possible test case for the issue here: https://github.com/pypi/warehouse/pull/15651

warehouse/forklift/legacy.py
```diff
@@ -607,6 +609,24 @@ def file_upload(request):
...
 
+    # If this is a user identity (i.e: API token) but there exists
+    # a trusted publisher for this project, send an email warning that an
+    # API token was used to upload a project where Trusted Publishing is configured.
+    # Only do this once per (API token, project) combination.
+    if request.user and project.oidc_publishers:
+        macaroon: Macaroon = request.identity.macaroon
+        # If we haven't warned about use of this particular API token and project
+        # combination, send the warning email
+        if macaroon not in project.macaroons_tp_warning:
```
di, line 620: I think the `macaroons_tp_warning` field is probably unnecessary on the `Project`, because we can just do something like this here:

```python
    warning_exists = request.db.query(
        exists().where(
            ProjectMacaroonWarningAssociation.macaroon_id == macaroon.id,
            ProjectMacaroonWarningAssociation.project_id == project.id,
        )
    ).scalar()
    
    if not warning_exists:
        ...
```
This is also a simpler/faster query, because we don't need to fetch all the possible warnings from the database for a project and iterate over them in Python.

warehouse/templates/email/api-token-used-in-trusted-publisher-project/subject.txt
```diff
@@ -0,0 +1,17 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/subject.txt" %}
+
+{% block subject %}{% trans project_name=project_name %}API token used to upload to project {{ project_name }} despite Trusted Publishing enabled{% endtrans %}{% endblock %}
```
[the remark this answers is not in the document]
di: Makes sense. I think I'm fine either way, but if we were able to come up with something a little more succinct I'd prefer that.

warehouse/forklift/legacy.py
di: > Doesn't that happen in the backend anyway?

If by "backend" you mean "in the database", I don't think so.

> Would iterating over it fetch all the warnings from the database?

I think so. I believe this is where it happens: https://github.com/sqlalchemy/sqlalchemy/blob/da639af16f77118bc17bbf5cf78fe41dd1818168/lib/sqlalchemy/ext/associationproxy.py#L1484-L1489

You mentioned something like "don't we need this relationship to keep track of it", and I don't think we strictly need it, we can just add rows to the association table directly, and take advantage of the key constraints there.

warehouse/forklift/legacy.py
```diff
@@ -607,6 +609,24 @@ def file_upload(request):
...
+            send_api_token_used_in_trusted_publisher_project_email(
+                request,
+                project.users,
+                project_name=project.name,
+                token_owner_username=request.user.username,
+                token_name=macaroon.description,
+            )
+            project.macaroons_tp_warning.append(macaroon)
```
di, line 628: We'd need to do this too:
```suggestion
            request.db.add(
                ProjectMacaroonWarningAssociation(
                    macaroon_id=macaroon.id,
                    project_id=project.id,
                )
            )
```

facutuesca: done!

warehouse/forklift/legacy.py
facutuesca: Got it, I made the change to use the table directly

di (approved):
LGTM

warehouse/migrations/versions/56b3ef8e8af3_add_project_macaroon_warning_table.py
```diff
@@ -0,0 +1,51 @@
...
+Revises: 34c3175f4bea
+Create Date: 2024-03-11 19:41:22.997939
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "56b3ef8e8af3"
+down_revision = "34c3175f4bea"
```
di, line 25:
```suggestion
down_revision = "8673550a67a3"
```

warehouse/migrations/versions/56b3ef8e8af3_add_project_macaroon_warning_table.py
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+add_project_macaroon_warning_table
+
+Revision ID: 56b3ef8e8af3
+Revises: 34c3175f4bea
```
di, line 16:
```suggestion
Revises: 8673550a67a3
```

Issue #13216, opened 2023-03-16
Tests or linting should test email formatting


di:
In #13215, it was determined that some emails were failing to send due to being mis-formatted. While #13214 will catch these in production, our tests and/or linting should catch these before merge.

miketheman:
Can you share any details on how the infinite retry was observed? That might help determine the testing/linting strategy.

di:
Yep, this: https://github.com/pypi/warehouse/blob/5046d0b747566dbbb9504e35c36359688b2d89a5/warehouse/email/__init__.py#L66-L74
As far as I'm aware, that will keep retrying indefinitely for most exceptions.

miketheman:
I spent some time on this today, trying to load and render the `subject.txt` templates, but it turns out that's not as simple as it looked at first glance.
The jinja `Template.render()` function needs a valid context that includes the variables **for each template**, so it's not readily generalizable as far as I can tell.

We might want to pursue a different approach, like a static linter rule.

PR #15651, opened 2024-03-21
Add test to check email subject templates do NOT contain newlines

facutuesca:
Fixes (partly) https://github.com/pypi/warehouse/issues/13216, related to https://github.com/pypi/warehouse/pull/13215.

This is a pretty blunt approach. We iterate over all email subject templates, use a regex to find the whole `subject` block, and assert that it does not contain newlines.

It does seem to do the job: it caught an existing subject template with newlines, which is now fixed in the second commit.

If using regexes is too blunt, then we can use this PR to discuss how we can test for this. I opened a PR that would've been affected by this issue, and I never would've realized if not for @woodruffw's comment [here](https://github.com/pypi/warehouse/pull/15641#discussion_r1534185953).

cc @di @miketheman


woodruffw (approved):
Thanks @facutuesca!

woodruffw:
(Needs update/rebase, but otherwise LGTM!)

PR #15652, opened 2024-03-21
Disable capture_message for tags until we cover more

di:
This is too noisy as-is.


miketheman (approved):

alex:
@di can we get an issue going with all the existing ones, I can gradually burn them down. (Do I have sentry access? I think I used to...)

di:
@alex Working on it now. The main problem is that there's so much garbage from years of not validating this.

alex:
👍 :-(

PR #15653, opened 2024-03-21
Cover some more human-friendly interpreter tags

di:


di:
cc @alex I think this is the best we can do here

alex (approved):
lol. great stuff.

miketheman (approved):
Good progress!

I still can't shake the feeling that this probably should be part of `packaging`'s responsibility - maybe this code can be removed if/when `packaging` adopts it for wider use.

alex:
I love when someone else writes code instead of me :-) A meta-point is that these tags aren't _really_ designed for parsing in this way. The way pip decides which are viable is that it produces a list of every tag that might be valid for a given interpreter.

di:
I'm not sure this belongs in `packaging`, since that library is mostly an implementation of various standards, and this transformation is definitely not standardized.

PR #15687, opened 2024-03-29
feat(admin): Add a Malware-specific UI

miketheman:
See individual commit messages for specifics.

Note: there's still plenty to do  - such as figuring out the removal and reporting steps, as well as whether the Observations have been actioned, but this should be good enough to get in prod to look at them for now.

### DIFF
--- a/warehouse/csp.py
+++ b/warehouse/csp.py
@@ -45,6 +45,12 @@ def content_security_policy_tween_factory(handler, registry):
             policy["sandbox"] = ["allow-top-navigation"]
             policy["default-src"] = [NONE]
 
+        # Specific enables for Admin UI
+        if request.path.startswith("/admin/"):
+            policy["frame-src"] = ["https://inspector.pypi.io"]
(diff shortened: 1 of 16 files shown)


ewdurbin (approved):

warehouse/csp.py
```diff
@@ -163,9 +163,10 @@ def includeme(config):
                 "font-src": [SELF, "fonts.gstatic.com"],
                 "form-action": [SELF, "https://checkout.stripe.com"],
                 "frame-ancestors": [NONE],
-                "frame-src": [NONE],
+                "frame-src": ["https://inspector.pypi.io"],
                 "img-src": [
                     SELF,
+                    "data:",  # Admin UI/Bootstrap 4 uses inline SVGs for icons
```
di, line 169: Should we limit these to admin-only somehow?

miketheman: That's a good idea - I have a change that does that, similar to our CSP carveout for `/simple/`.

PR #15688, opened 2024-03-29
refactor: apply 2FA to any non-exempt routes 

miketheman:
New actions, such as reporting malware via web UI, now require 2FA.

Contains other commits, making this change easier to implement. Please review individually.

### DIFF
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -205,39 +205,34 @@ def _check_for_mfa(request, context) -> WarehouseDenied | None:
     # at this point, and we only a User in these policies.
     assert isinstance(request.identity, User)
 
-    # If we're in the manage namespace or file uploads, we'll check if the user
-    # has 2FA enabled, and if they don't we'll deny them.
+    if not request.identity.has_two_factor:
+        # Return a different message for upload endpoint first.
+        if request.matched_route.name == "forklift.legacy.file_upload":
+            return WarehouseDenied(
+                "You must enable two factor authentication to upload",
+                reason="upload_2fa_required",
+            )
 
-    # Management routes that don't require 2FA, mostly to set up 2FA.
-    _exempt_routes = [
-        "manage.account.recovery-codes",
-        "manage.account.totp-provision",
-        "manage.account.two-factor",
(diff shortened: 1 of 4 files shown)


miketheman:
Note to self: rebase and resolve conflicts with security policy once #15692 is out.

warehouse/accounts/security_policy.py
```diff
@@ -205,39 +205,30 @@ def _check_for_mfa(request, context) -> WarehouseDenied | None:
...
-            "You must enable two factor authentication to upload",
-            reason="upload_2fa_required",
-        )
+        ):
+            return WarehouseDenied(
+                "You must enable two factor authentication.",
+                reason="manage_2fa_required",
+            )
 
     return None
```
di, line 234: I think while we're here, I'd prefer if this function failed closed rather than failed open. GitHub won't let me make the actual suggestions here, but roughly:

```python
if request.identity.has_two_factor:
    return None
elif request.matched_route.name in _exempt_routes:
    return None
elif request.matched_route.name == "forklift.legacy.file_upload":
    return WarehouseDenied(
        "You must enable two factor authentication to upload",
        reason="upload_2fa_required",
    )
else:
    return WarehouseDenied(
        "You must enable two factor authentication.",
        reason="manage_2fa_required",
    )
```

I think this is a bit more readable too. The `request.matched_route.name.startswith(route) for route in _exempt_routes` line specifically also seems unnecessary, since `_exempt_routes` is a list of route names and not a list of route prefixes.

miketheman: `_exempt_routes` does contain route prefixs. e.g. `manage.account.webauthn-provision` covers routes:

- "manage.account.webauthn-provision"
- "manage.account.webauthn-provision.options"
- "manage.account.webauthn-provision.validate"
- "manage.account.webauthn-provision.delete"

Similarly `"manage.account.recovery-codes"` covers sub-routes `.generate`, `.regenerate`, and `.burn.

I'll try to refactor a bit as suggested, but will preserve `startswith`

miketheman: See eabc4efa4b2324f991bbb41b11116591b45b3fed

Issue #15274, opened 2024-01-25
Replace deprecated `eslint` style/formatting directives with ESlint Stylistic


miketheman:
Our current setup runs `eslint` on any javascript files, finding linter errors and addressing stylistic concerns.

eslint has decided to deprecate formatting rules - see https://github.com/eslint/eslint/issues/17522 for more context.

The existing formatters have been extracted, and will be updated outside of eslint core.
See more details at https://eslint.style/ and specifically https://eslint.style/packages/js (we don't have any typescript or jsx in this project, so we don't need the meta-plugin)

To accomplish this, review stylistic, learn about the migration path, and apply necessary changes - `package.json` is a good place to start.
We probably need to update some javascript packages - we don't have dependabot do that for us regularly yet.

The end goal is to have `npm run lint` successfully and minimize the resulting code churn, as we have some of our custom styles set in `package.json`

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypi/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

If you are working on this issue and have questions, feel free to ask them here, in the [`#pypa-dev` chat channel on Libera](https://web.libera.chat/#pypa-dev), the [PyPA Discord](https://discord.gg/pypa) or on the [Discourse](https://discuss.python.org/c/packaging/14).

jzohrab:
Hey there, wondering if I can make a small dent in this -- at least my questions may help someone else finish this if I can't.

From the [getting started](https://warehouse.pypa.io/development/getting-started.html) there was no obvious way to do `npm run lint` in docker, so I'm assuming that `make lint` is the right thing to do.

Currently the lint runs successfully:

<detail>
<summary>Lint output</summary>

```
$ make lint
# Build our base container for this project.
... snip ...
...
touch .state/docker-build-base
docker compose run --rm base bin/lint
WARN[0000] Found orphan containers ([warehouse-vault-1]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up. 
[+] Creating 1/0
 ✔ Volume "warehouse_rstuf-metadata"  Created                                                                                     0.0s 
[...]

PR #15689, opened 2024-03-30
Issue 15274: Use eslint stylistic

jzohrab:
Draft PR, initial attempt to address #15274 

There are a few things I'm not clear about with this PR, but perhaps it will help start correct work for the open issue:

* `make lint` fails, but `npm run lint` passes (details for this are below)
* committed package-lock: Docker didn't like it when the package-lock didn't match the package.json, so I committed it.  Not sure what the right thing to do here was.
* I'm new to stylistic, and so while the changes appear to be correct, I'm not completely confident.

Failing `make lint` details: `npm run lint` (outside of Docker) passes, but `make lint` fails after building the image:

<details>
<summary>make lint failure</summary>

```
Oops! Something went wrong! :(
ESLint: 8.47.0
ESLint couldn't find the plugin "@stylistic/eslint-plugin".
(The package "@stylistic/eslint-plugin" was not found when loaded as a Node module from the directory "/opt/warehouse/src".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:
    npm install @stylistic/eslint-plugin@latest --save-dev
The plugin "@stylistic/eslint-plugin" was referenced from the config file in "package.json".
```

</details>

---

**UPDATE:** Branch updated, rebased to single commit, and re-pushed.  Thanks @miketheman for the feedback.

Both `make lint` and `npm run lint` pass now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "@babel/core": "^7.21.3",
         "@babel/eslint-parser": "^7.21.3",
         "@babel/preset-env": "^7.21.4",
+        "@stylistic/eslint-plugin-js": "^1.7.0",
         "@testing-library/dom": "^9.2.0",
         "@testing-library/jest-dom": "^5.16.5",
         "babel-jest": "^29.5.0",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "src",
+  "name": "warehouse",
```
miketheman, line 2: This change happens when running `npm install` outside the container environment.

I tend to use something like `docker compose run --rm static npm ...` for commands to run inside the `static` container.
This also helps avoid any platform-specific dependencies that might be installed as a result - keeping it consistent.

I don't think this is noted anywhere in the dev docs, so apologies! It's been a while since we revamped https://github.com/pypi/warehouse/blob/main/docs/dev/development/frontend.rst - if you wanted to take a stab at that after this, that could be awesome!

miketheman (changes requested):
Excellent start, thanks for taking this on!

I left a couple of notes inline, let me know if they don't make sense.

The `make lint` failure you're seeing appears to be due to not having the container be rebuilt after a change to `package*.json` I think this line needs to change:

https://github.com/pypi/warehouse/blob/d1a074a100022b576fc4f8c202128ced2c2b80f9/Makefile#L77

to:

```makefile
lint: .state/docker-build-base .state/docker-build-static
```

so that the make dependency will rebuild it for you. If you try that out and it works, feel free to add that as a commit to fix that for others!

jzohrab:
Thanks @miketheman for the comments -- your comments are longer than the sum of my changes :-)  Such is the world of guiding open-source PRs, I appreciate your time.

I'll address the points above and will re-push.

jzohrab:
@miketheman - should I commit the package-lock changes, or should I exclude the package-lock from the PR?  I don't want to introduce other changes.  I can at least get rid of that "name - warehouse" thing.

jzohrab:
> so that the make dependency will rebuild it for you. If you try that out and it works, feel free to add that as a commit to fix that for others!

Yep that was the fix.  Adding to the PR. 👍

package.json
```diff
@@ -37,6 +37,7 @@
     "@babel/core": "^7.21.3",
     "@babel/eslint-parser": "^7.21.3",
     "@babel/preset-env": "^7.21.4",
+    "@stylistic/eslint-plugin": "^1.7.0",
```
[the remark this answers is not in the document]
jzohrab: Fixed in the rebased branch.

package-lock.json
jzohrab: I removed this change in the rebased branch.

miketheman:
@jzohrab To get this mergable, feel free to mark as "ready for review" and update with latest from `main`. We squash commits on merge, so if it's easier to click a button in the UI, go for it.

jzohrab:
Merged main as at `4caaf3527`

miketheman (approved):
Excellent, thanks! Will merge in the morning.

Issue #15634, opened 2024-03-19
Cannot test database-created objects in the context of a webtest


miketheman:
Here's an example test case:

```python
from tests.common.db.accounts import UserFactory

def test_user_profile(db_session, webtest):
    user = UserFactory.create()
    assert user.username
    # vist user's page
    resp = webtest.get(f"/user/{user.username}/")
    assert resp.status_code == 200
```

This fails with a 404 Not Found. I want to get a 200.

- We need to include [our `db_session` fixture](https://github.com/pypi/warehouse/blob/569324e23e2032718fbf3d968137eb22389cfee1/tests/conftest.py#L343-L366) to allow us to use the `UserFactory`
- We [create our own `webtest` fixture](https://github.com/pypi/warehouse/blob/569324e23e2032718fbf3d968137eb22389cfee1/tests/conftest.py#L633-L645) for testing
- We use `pyramid_tm` to handle transactions

I tried following the guidance here https://docs.pylonsproject.org/projects/pyramid_tm/en/latest/#testing
but I'm overlooking something with how we create sessions/transactions, and can't seem to figure out how to share the session from `db_session` into the `webtest` test app, since `pyramid_tm` will start a fresh transaction for the web request.

There might be something to restructure between our `database`, `app_config` fixtures to allow this - I'm a little lost, so I thought I'd write it down for anyone else to poke at.

miketheman:
Oh, and the motivation behind this is that this kind of setup would allow us to write more functional (integration) tests that create a database state, and then perform actions via the web/api interface (a la browsing, logging in, uploading things, etc) with fewer mocks.

jzohrab:
Hi @miketheman - I opened a draft PR that doesn't fix anything, but maybe helps others reason with the code a bit.

I've at least confirmed that the db contains an object :-)  In my projects I sometimes do db-level queries to assert table state.  I also do cleanup prior to running tests, usually doing a simple and stupid `for t in tables: delete_all_data_in(t)`.  It's heavy-handed but it works.

The `db_session` fixture creates a session, `webtest` creates a tm, and pyramid creates a tm ... lots of potential sessions/tms.

miketheman:
Thanks @jzohrab ! I haven't had a chance to pursue that branch, but maybe @di is interested, as he's tinkered with this problem as well.

miketheman:
I took another look today, and am still scratching my head.
Thanks for the draft branch, @jzohrab sadly it didn't make it much clearer.

[...]

PR #15690, opened 2024-03-31
Investigating issue 15634 - creating db object for web test -- test still fails

jzohrab:
**Draft PR only for #15634 investigation**

Hi @miketheman , this is not a fix.  I haven't spent enough time to fully understand the code or what's going wrong, but perhaps these scribblings will give you some ideas about how to use the right sessions.


miketheman:
Implemented in #16031

PR #15692, opened 2024-03-31
require a verified email address for any account action

ewdurbin:


warehouse/manage/views/__init__.py
```diff
@@ -139,6 +140,79 @@
...
+                self.request.session.flash(
+                    (
+                        "Too many incomplete attempts to verify email address(es) for "
+                        f"{self.request.user.username}. Complete a pending "
+                        "verification or wait before attempting again."
+                    ),
+                    queue="error",
+                )
+
+        return HTTPSeeOther(self.request.route_path("manage.account"))
```
ewdurbin, line 195: Can we branch here? Redirect to `manage.account` if a user has a verified email and `manage.unverified-account` otherwise? The issue is it flashes the `You must verify your **primary** email address before you can perform this action.` when re-verifying from the unmanaged view.

warehouse/templates/manage/account.html
```diff
@@ -260,6 +260,18 @@ <h2 class="sub-title">{% trans %}Account details{% endtrans %}</h2>
       </p>
     </div>
 
+    {% if not user.has_primary_verified_email %}
```
ewdurbin, line 263: We probably can rollback all these template changes now?

warehouse/accounts/security_policy.py
```diff
@@ -184,8 +184,7 @@ def _permits_for_user_policy(acl, request, context, permission):
     if (
         isinstance(res, Allowed)
         and not request.identity.has_primary_verified_email
-        and request.matched_route.name.startswith("manage")
-        and request.matched_route.name != "manage.account"
+        and not request.matched_route.name.startswith("manage.unverified-account")
```
miketheman, line 187: Allow the user to click the verification link

```suggestion
        and not request.matched_route.name == "accounts.verify-email"
        and not request.matched_route.name.startswith("manage.unverified-account")
```

warehouse/manage/views/__init__.py
```diff
@@ -149,13 +226,8 @@
...
     require_reauth=True,
 )
-class ManageAccountViews:
-    def __init__(self, request):
-        self.request = request
-        self.user_service = request.find_service(IUserService, context=None)
-        self.breach_service = request.find_service(
-            IPasswordBreachedService, context=None
-        )
+@lift()
```
miketheman, line 229: self-note: going to have to read more about this pattern later on

miketheman (changes requested):

warehouse/templates/manage/unverified-account.html
```diff
@@ -0,0 +1,489 @@
...
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "manage_base.html" %}
+
+{% set user = request.user %}
+{% set title = gettext("Activate your account") %}
+
+{% set active_tab = 'account' %}
+
```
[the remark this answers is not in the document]
di: This is being used in the base template.

warehouse/manage/views/__init__.py
```diff
@@ -139,6 +140,82 @@
...
+                    Email.user_id == self.request.user.id,
+                )
+                .one()
+            )
+        except NoResultFound:
+            self.request.session.flash("Email address not found", queue="error")
+            return self.default_response
+
+        if email.verified:
+            self.request.session.flash("Email is already verified", queue="error")
```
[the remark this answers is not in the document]
di: I think it's still an error -- the user has followed an old link for some reason.

warehouse/accounts/security_policy.py
di: Fixed in 5085f78, 168233f and e23d0cc.

warehouse/accounts/security_policy.py
```diff
@@ -214,6 +214,8 @@ def _check_for_mfa(request, context) -> WarehouseDenied | None:
         "manage.account.totp-provision",
         "manage.account.two-factor",
         "manage.account.webauthn-provision",
+        "manage.unverified-account",
+        "manage.verify-email",
```
miketheman, line 218: Remove this extra line
```suggestion
```

di: There's a typo here that is fixed in c225b42 , but I think without this, a user that has no 2FA configured will not be able to follow the link to verify their email address.

warehouse/accounts/security_policy.py
```diff
@@ -214,6 +214,8 @@ def _check_for_mfa(request, context) -> WarehouseDenied | None:
         "manage.account.totp-provision",
         "manage.account.two-factor",
         "manage.account.webauthn-provision",
+        "manage.unverified-account",
+        "account.verify-email",
```
miketheman, line 218: It's `accounts.verify-email`, and it shouldn't be necessary in this scope, since `_check_for_mfa` only happens once the user has a verified email.

Now that I think of it, neither of these should be necessary here

miketheman (approved):
Approving to move forward, since the extra lines in `check_for_mfa` should have no use and can be removed later if needed.

warehouse/accounts/security_policy.py
di: Aha, good point, I missed that that was guarded.

miketheman: Odd that when removing them I get an infinite redirect, so leaving them in place sohuld be fine for now.

PR #15701, opened 2024-04-01
feat: add captcha to login page

miketheman:
Mostly duplicates already-existing logic for `RegistrationForm`.
If we want to add the logic to a third place/form, then we should
refactor to a mixin at that time.

The hardest work here was updates to the test code, to enable a
"smarter" CSP service, since CSP is not implemented as an
interface/service.

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -19,6 +19,7 @@ from email.headerregistry import Address
 import disposable_email_domains
 import humanize
 import markupsafe
+import sentry_sdk
 import wtforms
 import wtforms.fields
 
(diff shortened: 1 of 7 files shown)


miketheman:
Before/after:

<p float="left">
 Screenshot 2024-04-01 at 19 02 15
 Screenshot 2024-04-01 at 19 02 30
</p>

warehouse/accounts/forms.py
```diff
@@ -373,10 +373,24 @@ def validate_g_recaptcha_response(self, field):
...
+    g_recaptcha_response = wtforms.StringField()
+
+    def validate_g_recaptcha_response(self, field):
+        # do required data validation here due to enabled flag being required
+        if self.captcha_service.enabled and not field.data:
+            raise wtforms.validators.ValidationError("Recaptcha error.")
+        try:
+            self.captcha_service.verify_response(field.data)
+        except recaptcha.RecaptchaError:
+            # TODO: log error
```
ewdurbin, line 391: Should we pump this to sentry?

ewdurbin (approved):

warehouse/accounts/forms.py
miketheman: Added in 445aa4ece7a8bc0dc0dc585dec1c9d381cb15126

PR #15709, opened 2024-04-02
Revert "feat: add captcha to login page"

ewdurbin:
Reverts pypi/warehouse#15701


miketheman (approved):
We're getting "Recaptcha error" on login, and need to figure out what went wrong before trying again.

Issue #15431, opened 2024-02-20
Implement PEP 708


di:
[PEP 708](https://peps.python.org/pep-0708/) ("Extending the Repository API to Mitigate Dependency Confusion Attacks") is provisionally accepted, and finalization depends on:

> An implementation of the PEP in PyPI (Warehouse) including any necessary UI elements to allow project owners to set the tracking data.

di:
Done in #15716.

PR #15716, opened 2024-04-03
Implement Alternate Repository Location for PEP 708

cofiem:
Adds per-project [Alternate Repository Location as per PEP 708](https://peps.python.org/pep-0708/#alternate-locations-metadata).

- Implements the UI and database changes.
- Includes changes to the html and json APIs.
- Includes tests.
- The alternate repository location information is on the project settings page.
- There is a form to add, and a list to show and delete existing entries.

Notes:
- The 'tracks' metadata does not make sense for PyPI, as PyPI does not extend other repositories.

Relates to #15431 

Previews:

HTML: `curl http://localhost/simple/[project_name]/`
```html
<!DOCTYPE html>
<html>
  <head>
    <meta name="pypi:repository-version" content="1.1">
    <meta name="pypi:alternate-locations" content="https://example.org">
    <title>Links for [project_name]</title>
  </head>
...
```

JSON: `curl -H "Accept: application/vnd.pypi.simple.v1+json" http://localhost/simple/[project_name]/`
```json
{"alternate-locations":["https://example.org"], ... }
```

image


ewdurbin:
Hi @cofiem and thank you for your contribution! I'm going to ask @dstufft to take a look as PEP author.

ewdurbin (commented):
This is a really strong and comprehensive start, thank you! I left some notes after an initial review.

warehouse/templates/api/simple/detail.html
```diff
@@ -15,6 +15,10 @@
 <html>
   <head>
     <meta name="pypi:repository-version" content="{{ meta['api-version'] }}">
+{#    <meta name="pypi:alternate-locations" content="{{ request.route_path('api.simple.detail', name=name|canonicalize_name) }}">#}
```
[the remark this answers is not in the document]
cofiem: In the [PEP](https://peps.python.org/pep-0708/#alternate-locations-metadata), I wasn't sure if the part about "... clients MUST implicitly assume that the url the response was fetched from was included in the list..." meant that the url of PyPI should be in the list presented to clients or not. If the url the client used, and the url rendered by PyPI differ, that might be a problem? Likely a question for @dstufft .

cofiem:
Thanks very much for the initial review, @ewdurbin! I'll have a look at the cache keys for the model.

tests/unit/api/test_simple.py
```diff
@@ -18,10 +18,11 @@
 from pyramid.testing import DummyRequest
 
 from warehouse.api import simple
-from warehouse.packaging.utils import API_VERSION
+from warehouse.packaging.utils import API_VERSION, _valid_simple_detail_context
```
cofiem, line 21: I haven't changed the `API_VERSION` to `1.2`, as I wasn't sure if or how PyPI presents different versions of APIs?

warehouse/packaging/models.py
```diff
@@ -943,3 +948,34 @@ class ProjectMacaroonWarningAssociation(db.Model):
         ForeignKey("projects.id", onupdate="CASCADE", ondelete="CASCADE"),
         primary_key=True,
     )
+
+
+class AlternateRepository(db.Model):
```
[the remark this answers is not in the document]
cofiem: Added in commit https://github.com/pypi/warehouse/pull/15716/commits/52542f9f790c79d69b5810745e914a145696f6b9

firefly9966 (approved):

tests/unit/api/test_simple.py
dstufft: There's only ever one major version of the API active at once, and the only one we serve is the latest version. It should be fine to update the `API_VERSION` to 1.2, it'll just change the string in the response body.

warehouse/templates/api/simple/detail.html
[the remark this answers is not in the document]
dstufft: The intention is that `http://pypi.org/simple/$project/` will be included.

dstufft:
Thanks for picking this up by the way! Really jazzed to see it getting worked on.

ewdurbin:
Thanks for reviewing @dstufft! @cofiem, I believe the comments from Donald should address the outstanding open questions.

Once this is updated I think it is ready for final review and merge!

tests/unit/api/test_simple.py
cofiem: addressed in 6d4a48472bda2b2b240ff8e336086379f2088a7f

warehouse/templates/api/simple/detail.html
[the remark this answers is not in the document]
cofiem: addressed in 231645590f4a105981165d0d55edf7d5ea748ed3

ewdurbin:
@cofiem thank you! I pushed [a commit](https://github.com/pypi/warehouse/pull/15716/commits/64fa33b25442760f880888530b1705c6c0594c24) to fix tests and include the self-reference in JSON API as well.

@dstufft I think this is ready to go, but I am curious if we should _always_ send the self-reference in alternate-repositories value... or only when an alternate has been specified.

ewdurbin (approved):

cofiem:
@ewdurbin That commit to fix tests and make the JSON & HTML apis match makes sense!

warehouse/templates/api/simple/detail.html
```diff
@@ -15,6 +15,9 @@
 <html>
   <head>
     <meta name="pypi:repository-version" content="{{ meta['api-version'] }}">
+    {% for alt_repo in alternate_locations %}
+    <meta name="pypi:alternate-locations" content="{{ alt_repo }}">
```
ewdurbin, line 19: Note: Text of PEP says "... in HTML as a meta element named pypi-alternate-locations..." while the example shows `pypi:alternate-locations`.

@dstufft I'm assuming that consistency with `pypi:repository-version` is intended, so the PEP should likely be updated for consistency.

ewdurbin:
I'm going to hold off for a couple days, but if I don't hear from @dstufft on the two remaining open questions, I will remove the self-reference in alternate-repositories values and merge. The PEP reads as though they can be implied "When using alternate locations, clients MUST implicitly assume that the url the response was fetched from was included in the list."

ewdurbin:
Added a small callout block that we should remove when PEP 708 provisional acceptance becomes acceptance acceptance.
Screenshot 2024-09-19 at 9 46 34 AM

cofiem:
Excellent, thanks @ewdurbin!

tests/unit/api/test_simple.py
```diff
@@ -439,6 +460,15 @@ def test_with_files_with_version_multi_digit(
...
             assert db_request.override_renderer == renderer_override
 
+
+def _update_context(context, content_type, renderer_override):
+    if renderer_override != "json" or content_type in [
+        simple.MIME_TEXT_HTML,
+        simple.MIME_PYPI_SIMPLE_V1_HTML,
+    ]:
+        return _valid_simple_detail_context(context)
+    return context
```
woodruffw, line 470: I might be missing something, but I think this helper is currently causing the `test_with_files_quarantined_omitted_from_index` test to be skipped (since it's an instance method, but it's currently indented into the helper.) I'll send a PR to fix it.

woodruffw: Opened https://github.com/pypi/warehouse/pull/16777 with the fix.

warehouse/manage/views/__init__.py
```diff
@@ -1158,8 +1164,178 @@ def manage_project_settings(self):
...
+
+        # add the alternate repository location entry
+        alt_repo = AlternateRepository(
+            project=self.project,
+            name=form.display_name.data,
+            url=form.link_url.data,
+            description=form.description.data,
+        )
+        self.request.db.add(alt_repo)
+        self.request.db.add(
```
ewdurbin, line 1200: Ope, totally missed this in review, but this is pretty incorrect and the cause of https://github.com/pypi/warehouse/issues/12933#issuecomment-2378120802

PR #15731, opened 2024-04-05
Email block list: Handle subdomains and case sensitivity

di:
Follow-on to #15672.

### DIFF
--- a/tests/unit/accounts/test_forms.py
+++ b/tests/unit/accounts/test_forms.py
@@ -608,7 +608,9 @@ class TestRegistrationForm:
             "different email."
         )
 
-    @pytest.mark.parametrize("email", ["foo@wutang.net", "foo@clan.wutang.net"])
+    @pytest.mark.parametrize(
+        "email", ["foo@wutang.net", "foo@clan.wutang.net", "foo@wUtAnG.net"]
+    )
     def test_prohibited_email_error(self, db_request, email):
         domain = ProhibitedEmailDomain(domain="wutang.net")
         db_request.db.add(domain)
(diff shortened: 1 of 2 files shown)


warehouse/accounts/forms.py
```diff
@@ -286,7 +286,7 @@ def validate_email(self, field):
             )
 
         # Check if the domain is valid
-        domain = field.data.split("@")[-1]
+        domain = ".".join(field.data.split("@")[-1].split(".")[-2:]).lower()
```
miketheman, line 289: I think this is fine, but I guess I don't understand why we aren't re-using the `Address` we create on L282 above. eg:

```python
address = Address(addr_spec="foo@clan.wutang.net")
assert address.domain == "clan.wutang.net"

domain = ".".join(address.domain.split(".")[-2:]).lower()
assert domain == 'wutang.net'
```

It doesn't save a ton, but tis line conditional is getting more complex to read, and we're already doing the work to parse it.

Dealer's choice!

tests/unit/accounts/test_forms.py
```diff
@@ -608,13 +608,16 @@ def test_disposable_email_error(self, pyramid_request):
             "different email."
         )
 
-    def test_prohibited_email_error(self, db_request):
+    @pytest.mark.parametrize(
+        "email", ["foo@wutang.net", "foo@clan.wutang.net", "foo@wUtAnG.net"]
```
miketheman, line 612: While we're at it, how about a second-level subdomain?

miketheman (approved):
Good as is, couple of minor potential improvements.

PR #15743, opened 2024-04-09
fix: allow projects with publishers to be removed

### LINKED ISSUE #15729: Removing a project with OIDC configured fails
When removing a project via admin (such as malware), currently the removal fails, presumably due to not having an on delete cascade for the foreign key relationship.

```python
(psycopg.errors.ForeignKeyViolation) update or delete on table "projects" violates foreign key constraint "oidc_publisher_project_association_project_id_fkey" on table "oidc_publisher_project_association"
DETAIL:  Key (id)=(REDACTED) is still referenced from table "oidc_publisher_project_association".
[SQL: DELETE FROM projects WHERE projects.id = %(id)s::UUID]
[parameters: {'id': UUID('REDACTED')}]
```

We probably want to delete a publisher on project removal, since it's a one-to-one relationship, is that correct @woodruffw ?

We probably also want Admin UI functionality to remove an OIDC publisher from a project.

miketheman:
Resolves #15729

### DIFF
--- a/warehouse/migrations/versions/34c3175f4bea_remove_oidc_association_on_project_.py
+++ b/warehouse/migrations/versions/34c3175f4bea_remove_oidc_association_on_project_.py
@@ -1,0 +1,51 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Remove OIDC Association on Project Delete
+
+Revision ID: 34c3175f4bea
+Revises: 1fdecaf73541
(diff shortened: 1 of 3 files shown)


woodruffw (approved):
LGTM! IMO we could also delete the publisher itself _if_ the deleted project is the last one associated with it, but that's minor.

(We could also just do that periodically by deleting any "orphaned" providers, i.e. those with no projects.)

warehouse/migrations/versions/34c3175f4bea_remove_oidc_association_on_project_.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def upgrade():
+    op.drop_constraint(
+        "oidc_publisher_project_association_project_id_fkey",
+        "oidc_publisher_project_association",
+        type_="foreignkey",
+    )
+    op.create_foreign_key(
+        None,
```
di, line 33: Just curious, did you change the constraint name to be `None` or was this auto-generated? I'm just surprised it's different than when it was first added.

miketheman: It's autogenerated, and I confirmed that the constraint name remains `oidc_publisher_project_association_project_id_fkey` after the migration completes.

di: Weird! 🤷

di (approved):

PR #15746, opened 2024-04-10
docs(dev): add copy button for code blocks

miketheman:
To make it easier to copy code from the docs and paste into a terminal, add `sphinx-copybutton` which Furo theme already supports. Refs: https://pypi.org/project/sphinx-copybutton/

Remove leading `$` signs for commands, since they should not be copied.


docs/dev/development/development-database.rst
```diff
@@ -8,25 +8,25 @@ This dump is created from the database behind test.pypi.org, not production.
...
+    dropdb warehouse_dev_dump
 
 Now commit the result::
 
-    $ git checkout -b update_dev_db_dump
-    $ git add dev/example.sql.xz
-    $ git commit -m "Update development database dump"
+    git checkout -b update_dev_db_dump
+    git add dev/example.sql.xz
+    git commit -m "Update development database dump"
```
di, line 32: These multi-line command blocks don't really copy correctly, we might want to break them into separate blocks instead.

miketheman: Can you explain what you mean? I tested the copy button, and my cliboard contents have:

```
git checkout -b update_dev_db_dump
git add dev/example.sql.xz
git commit -m "Update development database dump"
```

di: Since copy/pasting this into a terminal would result in the first two lines immediately being executed due to the newlines, I would expect this to either be:

```
git checkout -b update_dev_db_dump
```
```
git add dev/example.sql.xz
```
```
git commit -m "Update development database dump"
```

or:

```
git checkout -b update_dev_db_dump; \
git add dev/example.sql.xz; \
git commit -m "Update development database dump"
```

di: (fwiw, I think I prefer the former)

miketheman: I've set the continuation on these multi-line commands. FWIW, my terminal application (iTerm2) warns me whenever I paste in multi-line command input before it inputs them. I realize not everyone uses my terminal nor my settings.

di (approved):

Issue #15748, opened 2024-04-10
Replace `User``request.identity` usages with `UserContext`


woodruffw:
Following https://github.com/pypi/warehouse/pull/15581 and https://github.com/pypi/warehouse/pull/15590: we now have three types that a `request.identity` can be:

* `User` when the identity is a user backed by a login session
* `UserTokenContext` when the identity is a user backed by an API token (i.e. macaroon)
* `PublisherTokenContext` when the identity is an `OIDCPublisher` backed by an API token

Of these, `User` and `UserTokenContext` are confusable and prone to error. We should probably collapse them into a single `UserContext` type of the following shape:

```py
@dataclass
class UserContext:
    user: User
    """
    The associated user.
    """

    macaroon: Macaroon | None
    """
    The Macaroon associated to the API token used to authenticate, if token authentication was used.
    """

    def __principals__(self) -> list[str]:
        return self.user.__principals__()
```

...with that, we'll be able to remove `UserTokenContext` and simplify `request.identity` back down to just two types.

woodruffw:
CC @facutuesca

facutuesca:
PR open here: https://github.com/pypi/warehouse/pull/15757

PR #15757, opened 2024-04-11
Combine `User` and `UserTokenContext` for user-backed identities in requests

facutuesca:
Fixes https://github.com/pypi/warehouse/issues/15748.

We have three types that a `request.identity` can be:

- `User`: when the identity is a user backed by a login session
- `UserTokenContext`: when the identity is a user backed by an API token (i.e. macaroon)
- `PublisherTokenContext`: when the identity is an `OIDCPublisher` backed by an API token

Of these, `User` and `UserTokenContext` are confusable and prone to error.

This PR collapses them into a single `UserContext` type.

cc @woodruffw @di


warehouse/macaroons/caveats/__init__.py
```diff
@@ -104,7 +104,7 @@ class RequestUser(Caveat):
     user_id: StrictStr
 
     def verify(self, request: Request, context: Any, permission: str) -> Result:
-        if not isinstance(request.identity, UserTokenContext):
+        if not isinstance(request.identity, UserContext):
             return Failure("token with user restriction without a user")
```
facutuesca, line 109: Should there be a new extra check here?:
```python
        if request.identity.macaroon is None:
            return Failure("token with user restriction without a macaroon")
```
or will that never happen?

woodruffw: I _think_ this can't happen, since the only place we end up checking the caveats is in `MacaroonSecurityPolicy.permits`, where `request.identity` will invariably either be an `PublisherTokenContext` or a `UserContext` that always has its macaroon set.

That being said, I see no harm in it as a sanity check either 🙂.

facutuesca: added!

woodruffw (approved):
LGTM, thanks @facutuesca!

warehouse/accounts/security_policy.py
```diff
@@ -102,7 +102,7 @@ def identity(self, request):
             return None
 
         # Sessions can only authenticate users, not any other type of identity.
-        return user
+        return UserContext(user, None)
```
di, line 105:
```suggestion
        return UserContext(user=user, macaroon=None)
```
Just to add a bit of clarity.

di (approved):

PR #15782, opened 2024-04-15
fix: use instantiated MutableDict to track state

miketheman:
Since we are updating the object's dictionary **contents**, and not replacing the entirety of the column, we need to tell SQLAlchemy that it should track that state, and persist it when complete.

Refs: https://docs.sqlalchemy.org/en/20/orm/extensions/mutable.html

### DIFF
--- a/warehouse/email/ses/models.py
+++ b/warehouse/email/ses/models.py
@@ -285,5 +285,5 @@ class Event(db.Model):
     )
 
     data: Mapped[dict] = mapped_column(
-        MutableDict.as_mutable(JSONB), server_default=sql.text("'{}'")  # type: ignore[arg-type] # noqa: E501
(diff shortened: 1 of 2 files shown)


warehouse/email/ses/models.py
```diff
@@ -285,5 +285,5 @@ class Event(db.Model):
     )
 
     data: Mapped[dict] = mapped_column(
-        MutableDict.as_mutable(JSONB), server_default=sql.text("'{}'")  # type: ignore[arg-type] # noqa: E501
+        MutableDict.as_mutable(JSONB()), server_default=sql.text("'{}'")
```
miketheman, line 288: I also fixed this type ignore here while I was at it.

ewdurbin (approved):

PR #15784, opened 2024-04-15
Fix stateful test failure

di:
This fixes a test failure with `--randomly-seed=999435887`, as discovered here: https://github.com/pypi/warehouse/actions/runs/8695732355/job/23847539984?pr=15737


miketheman (approved):
Nice!

PR #15795, opened 2024-04-16
Validate release version in filename

### LINKED ISSUE #15749: Post release source dist gets grouped with stable release wheel
**Describe the bug**

The docutils 0.21 has two files (https://inspector.pypi.io/project/docutils/0.21/, https://pypi.org/project/docutils/0.21/#files):

* docutils-0.21.post1.tar.gz
* docutils-0.21-py3-none-any.whl

The source distributions is part of the 0.21 release, while it's actually of version 0.21.post1. Pip agrees that this version is 0.21.post1:

```
$  virtualenv -p 3.12 --clear -q .venv
$  . .venv/bin/activate
(.venv) $  pip install --no-binary docutils docutils==0.21
ERROR: Could not find a version that satisfies the requirement docutils==0.21 (from versions: 0.3, 0.3.5, 0.3.7, 0.3.9, 0.4, 0.5, 0.6, 0.7, 0.8, 0.8.1, 0.9, 0.9.1, 0.10, 0.11, 0.12, 0.13.1, 0.14rc1, 0.14rc2, 0.14, 0.15, 0.15.1.post1, 0.15.2, 0.16b0.dev0, 0.16rc1, 0.16, 0.17b1, 0.17, 0.17.1b1.dev0, 0.17.1, 0.18b1, 0.18, 0.18.1b0, 0.18.1, 0.19b1, 0.19, 0.20rc1, 0.20, 0.20.1, 0.21rc1, 0.21.post1)
ERROR: No matching distribution found for docutils==0.21
(.venv) $ pip install docutils==0.21
Collecting docutils==0.21
  Downloading docutils-0.21-py3-none-any.whl.metadata (2.7 kB)
Downloading docutils-0.21-py3-none-any.whl (587 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 587.4/587.4 kB 4.6 MB/s eta 0:00:00
Installing collected packages: docutils
Successfully installed docutils-0.21
```

Under the right circumstances, this can break uv too.

**Expected behavior**

Either `docutils-0.21.post1.tar.gz` has its own version 0.21.post1, or source dists with this naming scheme are rejected by pypi.

**To Reproduce**
See description

**My Platform**
Ubuntu 22.04, Python 3.12

**Additional context**
n/a

di:
Fixes #15749 by ensuring that the version in the filename matches what's provided to us as metadata.


tests/unit/forklift/test_legacy.py
```diff
@@ -2000,11 +2000,18 @@ def test_upload_fails_with_diff_filename_same_blake2(
             ("no-way-{version}.tar.gz", "sdist", "no"),
             ("no_way-{version}-py3-none-any.whl", "bdist_wheel", "no"),
             # multiple delimiters
-            ("foo__bar-{version}-py3-none-any.whl", "bdist_wheel", "foo-.bar"),
+            ("foobar-{version}-py3-none-any.whl", "bdist_wheel", "foo-.bar"),
```
di, line 2003: This was an invalid wheel filename to begin with, the key here is that the project name doesn't match and the suggestion needs to be normalized.

tests/unit/forklift/test_legacy.py
```diff
@@ -2020,6 +2027,7 @@ def test_upload_fails_with_wrong_filename(
             IFileStorage: storage_service,
             IMetricsService: metrics,
         }.get(svc)
+        monkeypatch.setattr(legacy, "_is_valid_dist_file", lambda *a, **kw: True)
```
di, line 2030: Filename validation is now happening after dist file validation, so we need to monkeypatch this.

warehouse/forklift/legacy.py
```diff
@@ -693,7 +693,7 @@ def file_upload(request):
         release = (
             request.db.query(Release)
             .filter(
-                (Release.project == project) & (Release.version == form.version.data)
+                (Release.project == project) & (Release.version == str(meta.version))
```
di, line 696: Not strictly necessary but this moves us closer to using the metadata as the source of truth everywhere.

tests/unit/forklift/test_legacy.py
miketheman: The comment above this example is for multiple delimiters, as denoted by `__` - is that covered by another test case somewhere?

warehouse/forklift/legacy.py
miketheman: If I'm reading this correctly, then we no longer use `form.version.data` anywhere else - we're doing some validation on submission, but effectively ignoring the results. Is that intended?

miketheman (commented):
Couple of questions inline - didn't deeply review the logic changes.

tests/unit/forklift/test_legacy.py
di: The multiple delimiters are in `foo-.bar` and are still covered by this. The resulting filename doesn't really matter as long as it doesn't match.

warehouse/forklift/legacy.py
di: Ah, yes, I meant to remove that from the form. All the validation it does is already handled by `packaging` as well. Done in b3b7024.

miketheman: It looks like we still use the form's `version` field, but only because we deconstructed it into `meta` earlier?
https://github.com/pypi/warehouse/blob/f4d18dba7c397a7ceadf5f78db32528596e26ec1/warehouse/forklift/legacy.py#L499-L503

We no longer validate the form's `version`, but we use the version here for lookup and in other comparisons, not from the filename.

miketheman (commented):
This sure is confusing 😆 

The open question about metadata is one that I'd like to know the intent, so we know if/what we're using from the client POST vs the actual filenames/package metadata.

If the intent is to continue to use the POST form metadata for now, then ✅ but I'm not clear how this moves us a step towards not using the form data.

warehouse/forklift/legacy.py
di: > It looks like we still use the form's `version` field, but only because we deconstructed it into `meta` earlier?

Correct. We're still using the version in the POST body as the source of truth, until we resolve the TODO listed here and get all metadata from within the artifact.

> We no longer validate the form's version, but we use the version here for lookup and in other comparisons, not from the filename.

The version is still being validated when we're passing it to `metadata.parse`. We never used the version from the filename for any validation, the key thing this PR adds is validating that the POST data and the filename versions match.

tests/unit/forklift/test_legacy.py
```diff
@@ -480,24 +480,6 @@ def test_fails_invalid_version(self, pyramid_config, pyramid_request, version):
...
-            ),
-            (
-                {"metadata_version": "1.2", "name": "example", "version": "dog"},
-                "'dog' is an invalid value for Version. "
-                "Error: Start and end with a letter or numeral "
-                "containing only ASCII numeric and '.', '_' and '-'. "
-                "See "
-                "https://packaging.python.org/specifications/core-metadata"
-                " for more information.",
-            ),
```
di: I realized that these were still valid and added them back in a5779ab08fe0e9a175c73fb74fd2fa8bc1901276.

PR #15800, opened 2024-04-16
feat: preserve related_name on Observations

miketheman:
If we want to allow ourselves to preserve some details about a given
Observation after the related model has been removed, copy some of the
data on to the Observation record.

Since we can infer the model type from the table name, we don't need to
store that.

Contains a data migration to update existing records and enforce the
requirement at the database level.

### DIFF
--- a/warehouse/observations/models.py
+++ b/warehouse/observations/models.py
@@ -249,7 +249,7 @@ class HasObservations:
             observer=actor.observer,
             payload=payload,
             related=self,
-            related_name=self.name,  # type: ignore[attr-defined]
+            related_name=repr(self),
             summary=summary,
         )
 
--- a/warehouse/observations/models.py
+++ b/warehouse/observations/models.py
@@ -18,7 +18,7 @@ import typing
 
 from uuid import UUID
 
-from sqlalchemy import ForeignKey, sql
+from sqlalchemy import ForeignKey, String, sql
 from sqlalchemy.dialects.postgresql import JSONB, UUID as PG_UUID
 from sqlalchemy.ext.associationproxy import AssociationProxy, association_proxy
 from sqlalchemy.ext.declarative import AbstractConcreteBase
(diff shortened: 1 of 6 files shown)


warehouse/packaging/models.py
```diff
@@ -622,6 +622,16 @@ def __table_args__(cls):  # noqa
...
 
+    @property
+    def name(self) -> str:
+        """
+        Release doesn't have a name column, inherit from its Project.
+
+        NOTE: `name` is used in `warehouse.packaging.tasks.sync_bigquery_release_files`
+        If changed here and not there, may have implications on downstream data.
+        """
+        return self.project.name
```
ewdurbin, line 633: should this (or the property suggested above) perhaps be `f”{self.project.name}@{self.version}`? or similar to not discard too much fidelity?

warehouse/observations/models.py
```diff
@@ -244,6 +249,7 @@ def record_observation(
             observer=actor.observer,
             payload=payload,
             related=self,
```
ewdurbin, line 251: feels bad to be implicitly binding to `name`, should we perhaps use an explicit `related_name` property for models we may denormalize as such?

miketheman: I tried to use the pre-exisitng `__repr__()` call that these models implement, but it didn't exactly look right.

The resulting strings look like: `"Project(name='example')"` for a Project, and `"Release(project=Project(name='example'), version='0.0.2')"`.

These are semi-seralized, as in we could feasibly turn them back into structured data, but they likely won't hold up to efficient SQL queries - but maybe they don't have to?

miketheman: I've updated the scope to use `repr()` for these, and dropped the bind to `name`.

warehouse/packaging/models.py
miketheman: Dropped this scope, and using `repr()` instead.

ewdurbin (approved):
probably fine as is, but i think the raised comment regarding query efficiency if we ever _did_ want to do something like 'show all past observations related to the project name "widget"' would require a refactor.

what's included via repr should be enough for a refactor down the line so that's probably a "when we cross the bridge" concern.

i can't come up with a simple solution. repr is a fine compromise for now but something explicit for the purpose such as a dedicated explicit `related_name` property seems less likely to accidentally drift (not that we change our reprs.... uh ever?).

Issue #15836, opened 2024-04-23
Trusted Publisher support for GitLab subgroups


mwoodbri:
**What's the problem this feature will solve?**
I'm adding a trusted publisher to an existing PyPI project. The project is in a subgroup i.e. following [the GitLab documentation](https://docs.gitlab.com/ee/user/namespace/#determine-which-type-of-namespace-youre-viewing) it would be in the `alex-team/marketing` namespace.

If I try to use this I see the following error (though obviously I'm using our own namespace and project name - it's a private project):

![Screenshot 2024-04-23 at 10 43 05](https://github.com/pypi/warehouse/assets/1724545/a20f3198-5af1-4a41-9126-8fbd1c25033a)

I've also tried specifying namespace as the equivalent of `alex-team`, with project name as `marketing/myproject`.

**Describe the solution you'd like**
I could well be misinterpreting the instructions and doing it wrong, in which case any guidance would be gratefully received. The help label says "group/subgroup" which suggests it's supported but I'm not sure.

facutuesca:
@mwoodbri Thanks for the report! This is definitely a bug on our side, your first attempt using `alex-team/marketing` as the namespace should have worked. I'm creating a fix for it now.

PR #15839, opened 2024-04-23
Fix GitLab Trusted Publisher not accepting valid namespaces

facutuesca:
Fixes https://github.com/pypi/warehouse/issues/15836.

GitLab namespaces can have forward slash ("/") characters, since a repo inside a subgroup will have a namespace where the subgroups are delimited using slashes (e.g: `gitlab.com/group/subgroup/repo` has a namespace of `group/subgroup`).

I double checked by generating an OIDC token from a GitLab CI workflow in a repo inside a subgroup, the namespace claim is of the form `group/subgroup`.

cc @di @woodruffw

### DIFF
--- a/tests/unit/oidc/forms/test_gitlab.py
+++ b/tests/unit/oidc/forms/test_gitlab.py
@@ -48,14 +48,22 @@ class TestPendingGitLabPublisherForm:
 
 
 class TestGitLabPublisherForm:
-    def test_validate(self):
-        data = MultiDict(
+    @pytest.mark.parametrize(
+        "data",
+        [
             {
                 "namespace": "some-owner",
                 "project": "some-repo",
                 "workflow_filepath": "subfolder/some-workflow.yml",
-            }
-        )
+            },
+            {
+                "namespace": "some-group/some-subgroup",
+                "project": "some-repo",
+                "workflow_filepath": "subfolder/some-workflow.yml",
+            },
+        ],
+    )
+    def test_validate(self, data):
         form = gitlab.GitLabPublisherForm(MultiDict(data))
 
         # We're testing only the basic validation here.
@@ -73,6 +81,11 @@ class TestGitLabPublisherForm:
             },
             {
                 "namespace": "invalid_parethen(sis",
+                "project": "some",
+                "workflow_filepath": "some",
+            },
+            {
+                "namespace": "/start_with_slash",
                 "project": "some",
                 "workflow_filepath": "some",
             },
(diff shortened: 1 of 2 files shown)


warehouse/oidc/forms/gitlab.py
```diff
@@ -20,7 +20,7 @@
 
 # https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-project-and-group-names
 _VALID_GITLAB_PROJECT = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]*$")
-_VALID_GITLAB_NAMESPACE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-_.]*$")
+_VALID_GITLAB_NAMESPACE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9-_./]*$")
```
di, line 23: This permits trailing slashes `foobar/` and multiple repeated slashes `foo//bar`, both of which I think should be invalid.

facutuesca: That is also the case also for the other special symbols ([src](https://docs.gitlab.com/ee/user/reserved_names.html#limitations-on-usernames-project-and-group-names)):
> Usernames, project or group slugs:
> - Cannot start or end with a special character.
> - Must not contain consecutive special characters.

And there are also other limitations we don't check (like the project ending with `.git` or `.atom`. At the time, the reasoning was that we wanted some validation for the user input that discarded simple errors but didn't need to be exhaustive about rejecting invalid input (since we're not hitting the GitLab API to validate them).

Should we have more complex regexes and/or validation functions to account for all of the possible invalid inputs?

di: Let's merge this as-is and create a new issue to make the regexes a little more accurate.

tests/unit/oidc/forms/test_gitlab.py
```diff
@@ -76,6 +84,11 @@ def test_validate(self):
                 "project": "some",
                 "workflow_filepath": "some",
             },
+            {
+                "namespace": "/start_with_slash",
```
di, line 88: Is this a valid namespace? Or is this an edge case we shouldn't actually permit?

woodruffw: From GitLab's docs, I believe this is invalid:

> Usernames, project and group names must start and end with a letter (a-zA-Z) or a digit (0-9)

facutuesca: it's invalid (it's part of the test that checks invalid namespaces)

woodruffw: Whoops, missed that!

woodruffw (approved):
Thanks @facutuesca!

tests/unit/oidc/forms/test_gitlab.py
di: Sorry, also missed that!

di (approved):

warehouse/oidc/forms/gitlab.py
facutuesca: issue created here: https://github.com/pypi/warehouse/issues/15852

PR #15840, opened 2024-04-23
docs: self-managed GitLab instances are not supported (for Trusted Publishing)

facutuesca:
Motivated by https://github.com/di/id/issues/216, the docs don't make explicit that Trusted Publishing with GitLab only works when using the official `gitlab.com` instance, and doesn't work with self-managed instances.

This changes the docs to clarify that. It also adds a section in the troubleshooting doc to identify the cause of the error generated by using a self-managed instance.

cc @di @woodruffw

### DIFF
--- a/docs/user/trusted-publishers/adding-a-publisher.md
+++ b/docs/user/trusted-publishers/adding-a-publisher.md
@@ -106,6 +106,11 @@ each.
 
 === "GitLab CI/CD"
 
+    !!! note
+
+        Currently, only projects hosted on <https://gitlab.com> are supported. Self-managed
+        instances are not supported.
+
     For GitLab CI/CD, you **must** provide the repository's namespace, the
     repository's name, and the filepath of the GitLab CI/CD workflow that's
     authorized to upload to PyPI. In addition, you may **optionally**
--- a/docs/user/trusted-publishers/creating-a-project-through-oidc.md
+++ b/docs/user/trusted-publishers/creating-a-project-through-oidc.md
@@ -75,6 +75,11 @@ provide the name of the PyPI project that will be created.
 
 === "GitLab CI/CD"
 
+    !!! note
+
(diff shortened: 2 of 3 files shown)


docs/user/trusted-publishers/adding-a-publisher.md
```diff
@@ -106,6 +106,11 @@ each.
 
 === "GitLab CI/CD"
 
+    !!! note
+
+        Currently, only projects hosted on https://gitlab.com are supported. Self-managed
```
woodruffw, line 111: Nitpick for linkification (similar for other additions below):

```suggestion
        Currently, only projects hosted on <https://gitlab.com> are supported. Self-managed
```

docs/user/trusted-publishers/using-a-publisher.md
```diff
@@ -379,6 +379,11 @@ below describe the setup process for each supported trusted publisher.
 
 === "GitLab CI/CD"
 
+    !!! note
+
+        Currently, only projects hosted on https://gitlab.com are supported. Self-managed
+        instances are not supported.
+
```
woodruffw, line 386: No strong opinion, but IMO we could drop this one to reduce duplication: the note is already showing on both of the creation flows, so we don't need it on the use flow 🙂

docs/user/trusted-publishers/adding-a-publisher.md
facutuesca: fixed

docs/user/trusted-publishers/using-a-publisher.md
facutuesca: done, removed

woodruffw (approved):
LGTM, thank you @facutuesca! 

(CI failures look sporadic)

di (approved):

PR #15854, opened 2024-04-24
Revert "hide RTD flyout (#15580)"

### LINKED ISSUE #15851: Docs: disable Read the Docs flyout _without_ using CSS
**Describe the bug**

I found that you are hiding the Read the Docs flyout shown at the bottom right by using CSS in this line: https://github.com/pypi/warehouse/blob/main/docs/user/stylesheets/extra.css#L13 and there is no need to do it via CSS. You can simple disable it from the Read the Docs dashboard with a checkbox.

**Expected behavior**

The flyout can be disabled from the dashboard _without_ requiring any coding. Just:

1. Go to the https://beta.readthedocs.org/dashboard/docspypiorg/edit/
2. Click on "Addons (Beta)" from the left bar
3. Uncheck "Flyout enabled"
4. Done 👍🏼 

**To Reproduce**

1. Go to https://docs.pypi.org/
2. Inspect the DOM and see that `readthedocs-flyout` element has the property `display: none;`

**Additional context**

Note that Read the Docs is free for Open Source projects. The flyout, besides providing extra features to the readers, was our[^1] way to tell users the site was hosted by Read the Docs and promote our platform more broadly. Since we added the ability to hide the flyout on projects, we may eventually ask users to show some support here communicating the hosting provider in some way. Similar to what CPython's theme is doing by showing "Hosted on Read the Docs" at the bottom right (see https://github.com/python/python-docs-theme/pull/165). This is _not required_, but it would be pretty much appreciated 🙏🏼 

[^1]: I work at Read the Docs 😃

di:
Fixes #15851.

This reverts commit 56f0f6c719809564de44dbca18dff967d69930ac.

### DIFF
--- a/docs/blog/stylesheets/extra.css
+++ b/docs/blog/stylesheets/extra.css
@@ -12,5 +12,3 @@
   white-space: nowrap;
   display: block;
 }
-
-readthedocs-flyout { display: none; }
--- a/docs/blog/stylesheets/extra.css
+++ b/docs/blog/stylesheets/extra.css
@@ -1,3 +1,15 @@
+:root {
+  /* Reduce Read the Docs' flyout font a little bit */
+  --readthedocs-flyout-font-size: 0.7rem;
+
+  /* Reduce Read the Docs' notification font a little bit */
+  --readthedocs-notification-font-size: 0.8rem;
+
+  /* This customization is not yet perfect because we can't change the `line-height` yet. */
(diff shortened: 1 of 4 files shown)


di:
Without the additional styling:
image

With the additional styling:
image

ewdurbin:
I think I’d rather go full hidden and credit for hosting in the footers explicitly.

Issue #8635, opened 2020-09-30
Validate the contents of identity centric metadata


dstufft:
Currently if I'm looking at a project on PyPI, it can be difficult to determine if it's "real" or not. I can look and see the user names that are publishing the project as well as certain key pieces of metadata such as the project home page, the source repository, etc.

Unfortunately, there's no way to verify that a project that has say.. ``https://github.com/pypa/pip`` in it's home page, is actually the real pip, and isn't a fake imposter pip. The same could go for other URLs, or email addresses etc. Thus it would be useful if there was some way to actually prove ownership of those URLs/emails, and either differentiate them in the UI somehow, or hide them completely unless they've been proven to be owned by one of the publishing users.

---

Metadata to verify:
- [x] Project URLs related to Trusted Publishers used - #16205, #16472
- [x] GitHub statistics when Trusted Publisher is used - #16532
- [x] Project maintainers - #15737
- [x] Project organization - #15737
- [x] Emails in metadata that are verified user emails for any collaborator on the project - #16631
- [ ] ACME-style verification of arbitrary URLs - #16597

jamadden:
I recall issues like this coming up at least once, if not a few times, over in pypi-support. Someone would fork a repository, change the name in `setup.py`, and then upload it to PyPI, with or without any other changes. All of the documentation and other links would remain pointed to the originals. This was confusing for users and frustrating to owners of the original package. 

So I'm 👍 for some sort of blue verified checkmark or something from that perspective.

With my publisher hat on, though, I would hope this would be completely automated and I wouldn't have to do anything special to earn that blue checkmark.

ewjoachim:
One idea: we could add a blue checkmark for all links in the sidebar that contain a link back to the project's pypi page or `pip install <project-name>`. This would for us to load those links, but it would be zero-effort for most packages.

That being said, it wouldn't help if they point to forked versions, but in that case, the github star count might be a tell.

calebbrown:
👍 

Any progress on this issue?

I've been looking at malware from PyPI and it is common for the `author_email` to be "spoofed" (either pointing to nowhere, or using somebody else's email address).

Some related context is this HN discussion: https://news.ycombinator.com/item?id=33438678 Many commenters are asking about providing this sort of information.

I see some considerations that need discussion:

- how do you support small personal projects (one or two maintainers) *and* large projects (big team, company) at the same time?
[...]

PR #15862, opened 2024-04-25
Verified project links for project details.

javanlacerda:
Contribute towards closing #8635

It creates the logic for verifying `Source` and `Homepage` as `Project links`. 
It also add them to verified section on project details page.

Example of the package details screen with the new feature:
image

cc @di


warehouse/packaging/models.py
```diff
@@ -399,6 +400,22 @@ def latest_version(self):
             .first()
         )
 
+    def is_verified_url(self, url: str) -> bool:
```
di, line 403:
```suggestion
    def is_verified_url(self, url):
```

di (approved):

PR #15872, opened 2024-04-26
Add `base_publisher_url` to file/release events.

di:
This fixes an issue with #15862, where the `publisher_url` for _most_ events contained a commit SHA, and therefore had additional parts that made it hard to compare with project URLs.

The workaround is to introduce a `base_publisher_url` additional field that represents the unique "root" of the publisher for all publishers (i.e. the github repo) which we can use to compare to project URLs to verify them.

This means that these links will become verified only once a new event is published with this additional field, which should slowly happen over time.

To speed this up, we _could_ do a one-off data migration to add `base_publisher_url` to existing events, but I think this should be fine as-is.

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -763,7 +763,8 @@ class TestRelease:
         [
             ("xpto.com", "https://pub/url/", False),  # Totally different
             ("https://pub/", "https://pub/url/", False),  # Missing parts
-            ("https://pub/url/", "https://pub/url/", True),  # Exactly the same
+            ("https://pub/url/", "https://pub/url/", True),  # Exactly the same with /
+            ("https://pub/url", "https://pub/url", True),  # The same without /
             ("https://pub/url/blah.md", "https://pub/url/", True),  # Additonal parts
             ("https://pub/url", "https://pub/url/", True),  # Missing trailing slash
             ("https://pub/url/", "https://pub/url", True),  # Extratrailing slash
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -759,7 +759,7 @@ class TestRelease:
         assert release.trusted_published
 
     @pytest.mark.parametrize(
-        "url, publisher_url, expected",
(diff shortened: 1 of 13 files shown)


tests/unit/packaging/test_models.py
```diff
@@ -780,12 +781,12 @@ def test_is_url_verified(self, db_session, url, publisher_url, expected):
...
             source=release_file,
             tag="fake:event",
-            additional={"publisher_url": publisher_url},
+            additional={"base_publisher_url": base_publisher_url},
         )
         DBFileEventFactory.create(
             source=release_file,
             tag="fake:event",
-            additional={"publisher_url": publisher_url},
+            additional={"base_publisher_url": base_publisher_url},
```
javanlacerda, line 789: Why do we need this second event?

javanlacerda (approved):

tests/unit/packaging/test_models.py
di: To catch the bug that this commit fixed: https://github.com/pypi/warehouse/pull/15869/commits/7f93c4226600368f8e38e8e303e23425943e1872. It was failing because it doesn't work when there are multiple events found.

javanlacerda: I see!! Awesome! Thanks!

di:
Out of date and mostly resolved by https://github.com/pypi/warehouse/pull/16205 (@facutuesca, you might want to see if there's anything else we can use from this).

facutuesca:
> Out of date and mostly resolved by #16205 (@facutuesca, you might want to see if there's anything else we can use from this).

I see that here the extra field is added to the file upload events.

While the additional field being added is the same (`base_publisher_url` here vs `publisher_base_url` in https://github.com/pypi/warehouse/pull/16205), the difference is that this PR stored that field in the events, and then those events were used to verify URLs on-demand.

Since we are now verifying URLs at the time of the file upload (and storing the `verified: bool` in the DB), I don't think we need to store the field in the events, WDYT @di ?

di:
Agreed!

PR #15890, opened 2024-04-30
fix(blog): remove duplicate RSS feed and add JSON feed

Guts:
According to https://guts.github.io/mkdocs-rss-plugin/integrations/#automatically-set-with-material-theme, the RSS feed is automatically referenced by the Material theme's blog plugin. So, the actual manual integration make it duplicated:

image

> Firefox, extension [Want my RSS](https://github.com/Reeywhaar/want-my-rss)

Otherwise, since its version 1.12, the plugin proposes the JSON feed in addition to the RSS. Not so well known as its XML alternative, it's handled by numerous feed readers and JSON is generally lived as more readable.

Disclaimer: I'm the maintainer of the RSS plugin.

Related previous PR: https://github.com/pypi/warehouse/pull/14105 by @miketheman


miketheman:
Thanks for this - we also set a link to the feed in the footer here:
https://github.com/pypi/warehouse/blob/8c51e0a2d3a54c6d99b73587ee95c23756728e70/docs/mkdocs-blog.yml#L81-L82

Would that need to change as well?

Guts:
> Thanks for this - we also set a link to the feed in the footer here:
> https://github.com/pypi/warehouse/blob/8c51e0a2d3a54c6d99b73587ee95c23756728e70/docs/mkdocs-blog.yml#L81-L82
> 
> Would that need to change as well?

No, the social hyperlinks in the footer are completely manual and do not interfere with the HTML templates.

miketheman (approved):
Looks good, passed my evaluation - now shows non-duplicate feeds.

PR #15896, opened 2024-05-01
Stop sending emails/warnings on PGP signature uploads

### LINKED ISSUE #13733: Stop sending emails (and maybe warnings) on PGP signature uploads
In https://github.com/pypi/warehouse/pull/13707 we removed the support for uploading signatures to PyPI, but we added warnings both in email and in twine if someone attempts to upload a signature to let them know that support is gone.

We should remove that at some point in the future. Probably in roughly a year (so May 2024? or so).

di:
Fixes #13733.


woodruffw (approved):
🎉

dstufft (approved):

PR #15920, opened 2024-05-07
Use grouped version updates for Dependabot

### LINKED ISSUE #15909: Use grouped version updates for Dependabot

There are many pull requests for python dependency updates created by dependabot and then they are closed by `github/ composite-prs`, see #15907, and there are also a lot of notifications generated for watchers of the project.

image

It may be best to use [grouped version updates for Dependabot](https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/)

> As mentioned above is this README, a core reason why this Action exists is to "combine multiple Dependabot PRs into one". Work for this Action was completed before the GitHub Blog Post was published and the Dependabot Grouped Version Updates feature was released. 

And it also be mentioned in the [README page of combine-prs](https://github.com/github/combine-prs?tab=readme-ov-file#notice-) as above. 

For example if use `groups`, it will be work like this: https://github.com/jenkinsci/kubernetes-operator/pull/1004

I see that `groups` seems to be used in dependabot.yml, but not for against all python dependencies, not sure why? and it doesn't seem to work. Maybe I'm mistaken.

https://github.com/pypi/warehouse/blob/fd493259d840828d4f07bd6d7027980cdf767451/.github/dependabot.yml#L21

shenxianpeng:
Closes #15909 

I created the [test-groups-updates repo for testing](https://github.com/shenxianpeng/test-groups-updates). Based on my test, set `patterns` to `- "*"` can create a separate pull request for the remaining python packages.


miketheman (approved):
Thanks! Let's see how this goes.

miketheman:
I tried triggering a single update to rebase, but that doesn't appear to have the desired effect. We can wait a full day cycle for it to trigger on its own - no rush.

miketheman:
I've reopened the original issue, to continue conversation there, since we can't reopen a Pull Request.

Issue #15871, opened 2024-04-26
[Roadmap] Implement PEP 740


woodruffw:
[PEP 740](https://peps.python.org/pep-0740/) is in a final but not yet approved state. This issue is intended to lay out the dependencies/subcomponents of its implementing, including things that can be done in a preliminary manner.

## Index side

- [ ] Support the `attestations` form field in the current legacy upload endpoint
- [ ] Parse each received attestation from `attestations` and verify each against the uploading trusted publisher; fail the file upload if any attestations are invalid
  * This should be done in as generic a manner as possible, to ensure that future extension to email and key identities is possible.
- [ ] Persist verified attestations and expose them via the PEP 503 and Simple JSON API indices (via provenance objects, as specified in PEP 740)

## Uploader/publish side

- [ ] Support attestation generation in `gh-action-pypi-publish`
  * Concretely, this means `gh-action-pypi-publish` should use `sigstore-python` to sign the attestation payload defined in PEP 740
  * The resulting bundle should be transformed into a PEP 740 attestation object
- [ ] Support attestation uploading in `twine`: https://github.com/pypa/twine/issues/1094
- [ ] Once `twine` supports uploading, `gh-action-pypi-publish` should use that support to actually upload the attestation objects it generates above

CC @di @webknjaz @facutuesca for visibility

webknjaz:
I've gone ahead and filed a tracking issue in pip-tools as well: https://github.com/jazzband/pip-tools/issues/2080.

webknjaz:
> * Concretely, this means `gh-action-pypi-publish` should use `sigstore-python` to sign the attestation payload defined in PEP 740

With the exciting changes suggested in https://github.com/pypa/gh-action-pypi-publish/pull/230, it looks like we'll be able to leverage bits of composite actions and call the action even (if it makes sense to do so, as an alternative to using a Python lib within the container like the rest of stuff).

webknjaz:
> * Concretely, this means `gh-action-pypi-publish` should use `sigstore-python` to sign the attestation payload defined in PEP 740

Should this be done by the uploader, though? My initial feeling was that this should be done right after the job building the artifacts. Though, perhaps not by pypa/build (because of the privilege elevation concerns I often complain about). Are we going to treat the artifacts that the uploader gets as coming from trusted sources? Should we be emphasizing the security considerations and teaching the users more about the least privilege principle?

woodruffw:
> Should this be done by the uploader, though?

[...]

PR #15952, opened 2024-05-13
Add support for uploading attestations in legacy API

facutuesca:
## Description
Add support for uploading [PEP 740](https://peps.python.org/pep-0740/) attestations along with distribution files. Part of https://github.com/pypi/warehouse/issues/15871.
For now only GHA-signed attestations are supported, but the implementation allows easily adding other publishers by implementing the `OIDCPublisherMixin.publisher_verification_policy()` method in the corresponding subclass.

## Implementation
The core logic is in `legacy.py`: we check an upload request to see if it includes any attestations. If it does, we:
1. Check if the session is authenticated using Trusted Publishing (and GHA). If not, fail.
2. Parse the attestations using [`pypi-attestations`](https://github.com/trailofbits/pypi-attestations).
3. Get the verification policy corresponding to the current Trusted Publisher (for now, always GitHub) .
4. Verify the attestations using `sigstore` with the above verification policy, against the uploaded distribution file.

For now we only verify the attestations. Storing them will be implemented in a later PR. 

The GHA verification policy (from step 3) is defined in `GitHubPublisherMixin.publisher_verification_policy()`, and it checks the certificate in the attestation against the following claims:
- OIDCBuildConfigURI (e.g: `https://github.com/org/repo/.github/workflows/workflow.yml@....`)
- OIDCSourceRepositoryDigest (the commit SHA corresponding to the version of the repo used)

See [here](https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#mapping-oidc-token-claims-to-fulcio-oids) for the definition of each claim.

cc @woodruffw @di


warehouse/oidc/models/github.py
```diff
@@ -235,6 +241,25 @@ def publisher_url(self, claims=None):
...
+        sha = claims.get("sha") if claims else None
+        ref = claims.get("ref") if claims else None
+        if not (ref or sha):
+            raise InvalidPublisherError("The ref and sha claims are empty")
+
+        expected_build_configs = [
+            OIDCBuildConfigURI(f"https://github.com/{self.job_workflow_ref}@{claim}")
+            for claim in [ref, sha]
+            if claim is not None
+        ]
```
woodruffw, line 254: Nitpick: I might be mistaken, but in practice this should only be `ref`, not `sha`. With that being said, we can probably use `OIDCSourceRepositoryDigest` to enforce that the `sha` also matches 🙂

warehouse/oidc/models/github.py
```diff
@@ -235,6 +241,25 @@ def publisher_url(self, claims=None):
             return f"{base}/commit/{sha}"
         return base
 
+    def publisher_verification_policy(self, claims):
```
woodruffw, line 244: Nitpick: docstring explaining the exact inner policy being constructed would be good!

warehouse/oidc/models/github.py
facutuesca: The trusted publisher claim verification attempts both though: https://github.com/pypi/warehouse/blob/784f446e1ebedfaceda2404efcf76ea29f9d80e2/warehouse/oidc/models/github.py#L56-L62

facutuesca: Thinking about this further, the current claims (plus the Digest one you suggested above) are:
1. OIDCBuildConfigURI (e.g: https://github.com/org/repo/.github/workflows/workflow.yml@$REF/$SHA)
2. OIDCSourceRepositoryURI (e.g: https://github.com/org/repo/)
3. OIDCSourceRepositoryDigest (the commit SHA)

I think we could remove the repo URI (2), since it's included in the build config URI (1). WDYT?

woodruffw: Makes sense to me!

facutuesca: done!

warehouse/oidc/models/github.py
facutuesca: fixed

warehouse/forklift/legacy.py
```diff
@@ -1065,6 +1073,47 @@ def file_upload(request):
...
+                raise _exc_with_message(
+                    HTTPBadRequest,
+                    f"Could not verify the uploaded artifact using the included "
+                    f"attestation: {e}",
+                )
+            except Exception as e:
+                raise _exc_with_message(
+                    HTTPBadRequest,
+                    f"Unknown error while trying to verify included attestations: {e}",
+                )
```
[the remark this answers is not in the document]
facutuesca: Got it, I added metrics for `warehouse.upload.attestations.ok` (successful upload), `warehouse.upload.attestations.malformed` (error decoding/parsing the attestation) and `warehouse.upload.attestations.failed_verify` (error verifying the attestation).

I also added a Sentry message for unknown/unexpected errors

warehouse/oidc/models/github.py
woodruffw: Just tying things together: https://github.com/pypi/warehouse/pull/14335 is where checking both the `ref` and the `sha` was introduced.

facutuesca: Here's a PR that adds a comment explaining why we check for both: https://github.com/pypi/warehouse/pull/15967

requirements/dev.txt
```diff
@@ -3,4 +3,6 @@ hupper>=1.9
 pip-tools>=1.0
 pyramid_debugtoolbar>=2.5
 pip-api
-repository-service-tuf
```
kairoaraujo, line 6: It is now solved with the new releases of RSTUF that support tuf 4.0.0 :)

facutuesca: The new release has another conflict with `warehouse`. See: https://github.com/pypi/warehouse/pull/15958#issuecomment-2131200368

warehouse/forklift/legacy.py
```diff
@@ -1065,6 +1069,56 @@ def file_upload(request):
...
+                attestations = TypeAdapter(list[Attestation]).validate_json(
+                    request.POST["attestations"]
+                )
+                verification_policy = publisher.publisher_verification_policy(
+                    request.oidc_claims
+                )
+                for attestation_model in attestations:
+                    # For now, attestations are not stored, just verified
+                    attestation_model.verify(
+                        Verifier.production(),
```
woodruffw, line 1092: Not an issue with this PR per se, but the presence of this parameter is making me think that the current `verify(...)` API is a little wonky: maybe rather than taking a `Verifier` it should just instantiate one directly, since in practice we expect this to always be `Verifier.production()`?

facutuesca: We do use `Verifier.staging()` in the `pypi-attestation-models` unit tests. If that's not strictly needed, we can remove it. Otherwise, we could move the parameter to the end of the parameter list and give it a default value of `Verifier.production()`

woodruffw: > We do use `Verifier.staging()` in the `pypi-attestation-models` unit tests. If that's not strictly needed, we can remove it. Otherwise, we could move the `verifier` parameter to the end of the parameter list and give it a default value of `Verifier.production()`

Ah yeah, hmm. This is probably more trouble than it's currently worth 🙂

woodruffw (approved):
LGTM overall, nice work @facutuesca!

Note for other reviewers: this PR intentionally discards the attestations after verifying them, both for an initial testing period and to keep the changeset small. We'll follow up with another PR to persist the attestations 🙂

woodruffw (approved):
LGTM, thanks @facutuesca!

(Needs another dependency deconflict.)

facutuesca:
The `CI / Dependencies` failure is also happening in other PRs: https://github.com/pypi/warehouse/actions/runs/9677538567/job/26699500157?pr=16172

warehouse/forklift/legacy.py
```diff
@@ -1065,6 +1069,80 @@ def file_upload(request):
...
+                        "failed_unsupported_predicate_type"
+                    )
+                    raise _exc_with_message(
+                        HTTPBadRequest,
+                        f"Attestation with unsupported predicate type: "
+                        f"{predicate_type}",
+                    )
+
+                # Log successful attestation upload
+                metrics.increment("warehouse.upload.attestations.ok")
```
woodruffw, line 1144: Logic LGTM here, but we might want to factor this out into its own helper (since `file_upload` is already quite chunky) 🙂 

Maybe something like `process_attestations`?

facutuesca: sounds good, fixed!

warehouse/forklift/legacy.py
```diff
@@ -362,6 +366,92 @@ def _is_duplicate_file(db_session, filename, hashes):
     return None
 
 
+def _process_attestations(request, metrics, artifact_path: Path):
```
woodruffw, line 369: Nitpick: rather than pass in `metrics`, get `metrics` via `request.find_service(...)`

(I'll make this change; just flagging.)

warehouse/forklift/legacy.py
```diff
@@ -362,6 +366,92 @@ def _is_duplicate_file(db_session, filename, hashes):
...
+    artifact. Attestations are only allowed when uploading via a Trusted
+    Publisher, because a Trusted Publisher provides the identity that will be
+    used to verify the attestations.
+    Currently, only GitHub Actions Trusted Publishers are supported, and
+    attestations are discarded after verification.
+    """
+
+    # Check that if the file was uploaded with attestations, verification
+    # passes
+    if "attestations" in request.POST:
```
woodruffw, line 383: Nitpick: I think this check should be outside of this helper function, to avoid having the entire body be in a nested block 🙂

warehouse/forklift/legacy.py
```diff
@@ -362,6 +366,92 @@ def _is_duplicate_file(db_session, filename, hashes):
...
+                sentry_sdk.capture_message(
+                    f"Unexpected error while verifying attestation: {e}"
+                )
+                raise _exc_with_message(
+                    HTTPBadRequest,
+                    f"Unknown error while trying to verify included "
+                    f"attestations: {e}",
+                )
+
+            if predicate_type != "https://docs.pypi.org/attestations/publish/v1":
```
woodruffw, line 441: Nitpick: should probably constant-ify this predicate type somewhere, or pull it directly from `pypi-attestations` if possible.

woodruffw (commented):
Some nitpicks that I'll address today!

warehouse/forklift/legacy.py
woodruffw: https://github.com/trailofbits/pypi-attestations/pull/29 will add an appropriate constant.

di (approved):
LGTM

PR #15960, opened 2024-05-15
fix: add hostname to URLs generated in email

miketheman:
Since this email is triggered during upload, the resulting domain sent
to users is broken, as it's using `uploads.pypi.org` which this domain
doesn't support.


miketheman:
Keep in mind: We're in code freeze for now, so unless this is urgent, we may not merge it until we're out of code freeze.

woodruffw (approved):
LGTM behaviorally -- thanks for catching this @miketheman!

(IMO this is not extremely urgent, so it can wait until after the freeze.)

ewdurbin (approved):

PR #15967, opened 2024-05-16
Add comments explaining GitHub's `job_workflow_ref` claim behavior

facutuesca:
The `job_workflow_ref` claim included in GitHub's OIDC tokens is usually something like:
```json
{
  "job_workflow_ref": "org/repository/.github/workflows/release.yml@refs/heads/main"
}
```

Where the last component (`refs/heads/main`) is the same as the `ref` claim, usually the name of a branch or tag.

However, the logic we use to check the claims assumes that last component could also be a commit SHA (the same as the `sha` claim):

https://github.com/pypi/warehouse/blob/15a9a6c39b1bc40f3dbc67b74b86dc0a104b7d44/warehouse/oidc/models/github.py#L55-L60

but there was no explanation of when a SHA would be used instead of the normal `ref`. The [original PR](https://github.com/pypi/warehouse/pull/14335) that introduces this logic mentions that both `ref` and `sha` terminations had been observed, but it was unclear why/when this happened.

This PR adds a comment explaining one of the cases where the `ref` claim will be empty, and the `job_workflow_ref` claim will use the SHA instead.

cc @woodruffw @di

### DIFF
--- a/warehouse/oidc/models/github.py
+++ b/warehouse/oidc/models/github.py
@@ -39,7 +39,7 @@ def _check_repository(ground_truth, signed_claim, all_signed_claims):
 def _check_job_workflow_ref(ground_truth, signed_claim, all_signed_claims):
     # We expect a string formatted as follows:
     #   OWNER/REPO/.github/workflows/WORKFLOW.yml@REF
-    # where REF is the value of the `ref` claim.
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Thanks @facutuesca, great clarification!

ewdurbin (approved):

Issue #15980, opened 2024-05-20
When developing on Windows, a `git clone` operation may automatically convert file line endings to CRLF, leading to a Sad


miketheman:
**Describe the bug**
Our docs tell users to `git clone` the repo, and if performed in the outer context (Windows CLI), this can lead to git changing file endings and making the shells scripts non-executable by our docker containers.

**Expected behavior**
`make` commands on Windows complete successfully.

**Proposed solution**

Create a `.gitattributes` file in the root of the repo, setting all files to be `lf` (or `\n` - whichever is the correct config), so anyone cloning the code on Windows will not run into this as well.

Refs:
- https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#:~:text=core.-,autocrlf,-If%20you%E2%80%99re%20programming
- https://git-scm.com/docs/gitattributes

Credit to @ashtonwooster who uncovered this problem!

PR #15982, opened 2024-05-20
Fix: Included .gitattributes file to fix line endings on Windows devices

AshtonWooster:
Addresses: #15980 

Added a .gitattributes file in the root directory, this will set all line endings in text files to LF, such that when cloning the repo through WSL or another Windows tool, it will update the line endings to LF, fixing the incompatibility. 

I tested by removing my cache, resetting my core.autocrlf to unset, and then doing a make purge. I then cloned the repo again and rebuilt it. The line endings were correctly set to LF, and the docker containers were able to successfully build. 

PS: This is my first pull request, so I'm not sure how to format these, please let me know what I can improve!


miketheman (approved):
Excellent work, thanks!

We squash commits on merge, so don't worry too much about intermediate commits, rather the Pull Request title and description are more important.

We'll likely merge this soon, but not while we're out at PyCon or in transit.

miketheman:
Thanks @AshtonWooster , it's merged now for any other Windows-based contributors in the future 🎉

PR #16027, opened 2024-05-30
Move maintaing the RowCount into an async task

dstufft:
Using a database trigger to maintain the `row_counts` table meant that every `INSERT` or `DELETE` to one of these tables ended up locking a table specific row on the `row_counts` table, which ended up interacting poorly with the lock added in https://github.com/pypi/warehouse/pull/13936.

We don't actually need these row counts to be completely accurate or updated immediately, so we instead remove the triggers to maintain the counts on this table and add an async task that will update the counts every 5 minutes.

Another option here would be to use one of the methods of estimating the count: https://wiki.postgresql.org/wiki/Count_estimate, where PostgreSQL will maintain the count for us (but we have less control over when it happens).

### DIFF
--- a/tests/unit/packaging/test_tasks.py
+++ b/tests/unit/packaging/test_tasks.py
@@ -36,6 +36,7 @@ from warehouse.packaging.tasks import (
     update_release_description,
 )
 from warehouse.utils import readme
+from warehouse.utils.row_counter import compute_row_counts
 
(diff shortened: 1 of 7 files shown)


dstufft:
Note: The only places I could find that we're using this are the counts on the homepage and the project stats we emit into datadog.

ewdurbin (approved):

PR #16031, opened 2024-05-31
Tell warehouse.db to reuse db_session pytest fixture

benjaoming:
Hi all :wave: 

Tested locally and seems like the test case provided by @jzohrab (https://github.com/pypi/warehouse/pull/15690) now works. I couldn't have grasped this without the introductury PR, so great idea to post that.

Fixes https://github.com/pypi/warehouse/issues/15634 (or maybe doesn't)

p.s. I'm very new to Warehouse and may not understand common practices, please be gentle :heart: 

Possible actions:

* [x] Investigate if the explicit transaction manager is necessary - or maybe can be moved into the `db_session` fixture
* [x] ~Comment back in `sqlalchemy_session_persistence = 'commit'`~ Nope, that's a required addition
* [x] Remove SQL debug statements
* [x] ~Fix some issues with stubs in `tests/unit/test_db.py`~ no need in new approach


benjaoming:
Will have a look at fixing the stubbed out unit tests in `tests/unit/test_db.py` :+1:

miketheman:
@benjaoming Thanks for taking a stab at this! Looks like you're coming along well. Please run a `make tests` locally as well, since that ought to point out any missing coverage when you're done.

Since this is your first contribution, one of the repo admins will have to approve a test run each time, so let us know when you're seeing the finish line 😉

warehouse/db.py
```diff
@@ -134,6 +134,12 @@ def _configure_alembic(config):
 
 
 def _create_session(request):
+
+    # Use pre-configured db session (usually for integration test transactions)
+    db_session = request.environ.get("warehouse.db_session")
```
benjaoming, line 139: I don't have a terrible lot of confidence in Pyramid stuff, would it be more safe to guard this with `if hasattr(request, "environ")` ?

benjaoming:
> Since this is your first contribution, one of the repo admins will have to approve a test run each time, so let us know when you're seeing the finish line 😉

It's coming along well :+1: I've run `make tests` and it worked.

Happy to address all feedback!

tests/conftest.py
```diff
@@ -622,18 +624,39 @@ def xmlrpc(self, path, method, *args):
...
+
+@pytest.fixture
+def webtest(app_config, tm, db_session):
     # We want to disable anything that relies on TLS here.
     app_config.add_settings(enforce_https=False)
 
-    try:
-        yield _TestApp(app_config.make_wsgi_app())
-    finally:
-        app_config.registry["sqlalchemy.engine"].dispose()
```
miketheman, line 636: Do you know why we not need to `dispose()` any more?

warehouse/db.py
miketheman: I think it's pretty safe to assume that `request` should respond to `.environ` as it's part of the object's construction. Ref: https://docs.pylonsproject.org/projects/pyramid/en/2.0-branch/api/request.html#pyramid.request.Request
and the underlying object's construction should even fail if a dict-like environ isn't passed:
https://github.com/Pylons/webob/blob/0a69ff129df2da9bba347981c4ef632a8b7c2ddb/src/webob/request.py#L85-L87

That said, I wish there was another way to accomplish this, since reaching into the environment feels a bit weird, especially since we're stuffing it into the environment to begin with.

Since this happens on every new real request, I'd prefer to continue to explore if there's a way we can control the new session behavior, instead of having what we know is a call that will fail in real-world scenarios.

I was re-reading some sqlalchemy docs, and maybe we should control during testing when `autobegin` is called?
https://docs.sqlalchemy.org/en/20/orm/session_transaction.html#setting-transaction-isolation-levels-dbapi-autocommit

tests/conftest.py
```diff
@@ -622,18 +624,39 @@ def xmlrpc(self, path, method, *args):
...
+    # Register the transaction manager with the db_session
+    zope.sqlalchemy.register(db_session, transaction_manager=tm)
+
+    # Create WSGI app with current test settings
+    # (this could be a fixture)
+    app = app_config.make_wsgi_app()
+
+    # Register the app with the external test environment, telling
+    # warehouse.db to use this db_session and use the Transaction manager.
```
miketheman, line 650: Did you mean?
```suggestion
    # request.db to use this db_session and use the Transaction manager.
```

tests/conftest.py
```diff
@@ -622,18 +624,39 @@ def xmlrpc(self, path, method, *args):
...
+def webtest(app_config, tm, db_session):
     # We want to disable anything that relies on TLS here.
     app_config.add_settings(enforce_https=False)
 
-    try:
-        yield _TestApp(app_config.make_wsgi_app())
-    finally:
-        app_config.registry["sqlalchemy.engine"].dispose()
+    # Register the transaction manager with the db_session
+    zope.sqlalchemy.register(db_session, transaction_manager=tm)
```
miketheman, line 643: I _think_ this can be removed safely, since the tm is being set explicitly later on in `testapp`

miketheman (commented):
Some comments inline, but this is shaping up nicely!

warehouse/db.py
benjaoming: > Since this happens on every new real request, I'd prefer to continue to explore if there's a way we can control the new session behavior, instead of having what we know is a call that will fail in real-world scenarios.

I agree!

I'll try to write my thoughts... no 'splaining intended, it's just to share my trail:

In pytest, there are fixtures which are session-, module- and function-scoped. The app_config is session-wide and I suppose[1] that all the `includeme(config)`s are called in the session-wide scope. Which of course avoids doing the same work for every test case :+1: 

But _also_ hence the following mock does NOT work... `warehouse.db.includeme` was called in the session-wide scope and the function pointer to `_create_session` has already been added with `config.add_request_method`, so it's too late to mock it out.

```
@pytest.fixture
def webtest(app_config, tm, db_session):
    # We want to disable anything that relies on TLS here.
    app_config.add_settings(enforce_https=False)

    # Register the transaction manager with the db_session
    zope.sqlalchemy.register(db_session, transaction_manager=tm)

    with mock.patch("warehouse.db._create_session", lambda _: db_session):
        app = app_config.make_wsgi_app()

        # Register the app with the external test environment, telling
        # warehouse.db to use this db_session and use the Transaction manager.
        testapp = _TestApp(
            app,
            extra_environ={
                "tm.active": True,  # disable pyramid_tm
                "tm.manager": tm,  # pass in our own tm for the app to use
            },
        )
        yield testapp
```

(I tried the above and it didn't work - I'm also not sure if `app_config.add_settings(enforce_https=False)` makes a difference)

The previous approach to "side-load" the request environment (`extra_environ`) may be preferred in the example [2] because the environment needs to be read after configuring the app in the pytest session-scope, so it's relatively uncomplicated (but as you said, adds a failed call for every request :no_good: ).

From this ramble (and correct me if I misunderstood something), I would conclude that conftest.py should have a different session-scoped `app_config` fixture? One that injects a configuration that is read by `warehouse.db.includeme`?

I've written and tested an alternative POC which also gets the work done:

https://github.com/benjaoming/warehouse/compare/issue_15634_test_user_profile-investigation...benjaoming:warehouse:issue_15634_test_user_profile-investigation-new-app_config?expand=1

Could that be a way forward?

[1] Disclaimer: I'm new to the Pyramid stack
[2] https://docs.pylonsproject.org/projects/pyramid/en/latest/tutorials/wiki2/tests.html#modifying-the-fixtures

tests/conftest.py
benjaoming: In the `db_session` fixture, the call to `.dispose()` is already issued. I think the responsibility should be with that fixture.. not super sure. But it should only need to happen once.

tests/conftest.py
benjaoming: I didn't find out what magic eventually registers it, but seems likely that it should happen since we define `tm.manager`. I've removed it and it continued to work... :crossed_fingers:

benjaoming:
Thanks for looking @miketheman! I've addressed the feedback in this PR branch. But I added a new approach in another branch: https://github.com/benjaoming/warehouse/compare/issue_15634_test_user_profile-investigation...benjaoming:warehouse:issue_15634_test_user_profile-investigation-new-app_config?expand=1

LMK if you'd prefer to merge that in here for the time being. Could be seen as the "best apprach for now".

I'm converting the PR to a draft again, since it needs some more discussion :+1:

tests/conftest.py
miketheman: Thanks for clarifying!

benjaoming:
I'm pulling in the other approach since it's better than what's in this PR currently :hourglass_flowing_sand:

tests/conftest.py
```diff
@@ -340,6 +344,24 @@ def app_config(database):
     return cfg
 
 
+@pytest.fixture(scope="session")
+def app_config(database):
+
+    return get_app_config(
+        database,
+    )
```
miketheman, line 352: style: Odd that the linter didn't try to inline this to a single line 🤔

warehouse/db.py
```diff
@@ -134,6 +134,7 @@ def _configure_alembic(config):
 
 
 def _create_session(request):
+
```
miketheman, line 137: nit: extra line
```suggestion
```

warehouse/db.py
```diff
@@ -187,8 +188,11 @@ def includeme(config):
         pool_timeout=20,
     )
 
-    # Register our request.db property
-    config.add_request_method(_create_session, name="db", reify=True)
+    # Possibly override how to fetch new db sessions from config.settings
```
miketheman, line 191:
```suggestion
    # Possibly override how to fetch new db sessions from config.settings
    #  Useful in test fixtures
```

tests/conftest.py
```diff
@@ -622,18 +644,34 @@ def xmlrpc(self, path, method, *args):
...
-    #       changes. This probably involves flushing the database or something
-    #       between test cases to wipe any committed changes.
+def tm():
+    tm = transaction.TransactionManager(explicit=True)
+    tm.begin()
+    tm.doom()
 
-    # We want to disable anything that relies on TLS here.
-    app_config.add_settings(enforce_https=False)
+    yield tm
```
miketheman, line 652: question: do we still need the external `tm` fixture, or can it be inlined to the `webtest` fixture? You seemed to consider it in your top-level checklist.

tests/conftest.py
```diff
@@ -340,6 +344,24 @@ def app_config(database):
...
+@pytest.fixture(scope="session")
+def app_config(database):
+
+    return get_app_config(
+        database,
+    )
+
+
+@pytest.fixture(scope="session")
+def app_config_dbsession_from_env(database):
```
miketheman, line 356: Seeing this split out is helpful - since it allows me to ask the following question:

Historically, in the context of test scenarios we've been passing a session-scoped `app_config` fixture that has a `database` fixture passed in.

Is there any reason for us to maintain these two fixtures, or could we collapse these down to `app_config` only?

miketheman (commented):
@benjaoming thanks for continuing to work on this!

I've looked at the most recent changes, and they are shaping up nicely.
Specifically there's now a single call at the top of the pyramid application startup that would change behaviors, which is better than on every worker startup.

I appreciate that this is an early foray into Pyramid - it can take a bit to grok the behaviors, but I think this is getting closer to something very useful.

I think it's fine to skip exploring `autobegin` behaviors if this approach works.
Let me know if the comments aren't clear!

tests/conftest.py
benjaoming: My eyes didn't catch this, neither... I guess I _blindly_ rely on auto-linting :innocent:

benjaoming:
Thanks for the review :pray: Working out some updates now and will request another review when ready :hourglass_flowing_sand:

tests/conftest.py
benjaoming: Fixed!

And totally agree, this tiny fixture is never reused, so feels like some premature optimization. The pattern was from the Pyramid docs, but it might have been in the docs to make the example code easier to read.

Regarding the top-level checklist, I had been wondering if the TransactionManager would be auto-created in another layer, but I tried it... and the tests failed, so I assume we have to manage the manager :)

tests/conftest.py
benjaoming: The `database` is current a session-wide fixture like `app_config`. I suppose that since `app_config` depends on the other fixture, then `database` is created firstly.

However, trying to include the code from `database` in `app_config` gave birth to this error:

```
_________________________________________________ ERROR at teardown of test_user_profile _________________________________________________

    @request.addfinalizer
    def drop_database():
>       janitor.drop()

tests/conftest.py:336: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../lib/python3.11/site-packages/pytest_postgresql/janitor.py:90: in drop
    self._dont_datallowconn(cur, db_to_drop)
../lib/python3.11/site-packages/pytest_postgresql/janitor.py:98: in _dont_datallowconn
    cur.execute(f'ALTER DATABASE "{dbname}" with allow_connections false;')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <psycopg.Cursor [closed] [BAD] at 0x7ce3261f5c70>, query = 'ALTER DATABASE "tests" with allow_connections false;', params = None

    def execute(
        self,
        query: Query,
        params: Optional[Params] = None,
        *,
        prepare: Optional[bool] = None,
        binary: Optional[bool] = None,
    ) -> Self:
        """
        Execute a query or command to the database.
        """
        try:
            with self._conn.lock:
                self._conn.wait(
                    self._execute_gen(query, params, prepare=prepare, binary=binary)
                )
        except e._NO_TRACEBACK as ex:
>           raise ex.with_traceback(None)
E           psycopg.errors.InvalidCatalogName: database "tests" does not exist

../lib/python3.11/site-packages/psycopg/cursor.py:732: InvalidCatalogName
```

I'm not sure, but I think what's happening is that since `request.addfinalizer` adds something to a finalization stack, then _maybe_ some garbage collection has destroyed something (i.e. the database) already at the point where `janitor.drop()` is supposed to be called.

I tried out some different sequences, but in the end I think it's maybe easier for now to keep the old fixtures structure.

benjaoming:
In an attempt to reorganize fixtures, I noted that this line overwrites `config` from the global scope:

https://github.com/pypi/warehouse/blob/eeb119f1bea381a6d5747efa7e8d9f47b341213d/tests/conftest.py#L253

Is there an issue with linting not picking this up?

tests/conftest.py
```diff
@@ -622,18 +642,31 @@ def xmlrpc(self, path, method, *args):
 
 
 @pytest.fixture
-def webtest(app_config):
-    # TODO: Ensure that we have per test isolation of the database level
-    #       changes. This probably involves flushing the database or something
-    #       between test cases to wipe any committed changes.
-
+def webtest(app_config_dbsession_from_env, db_session):
```
benjaoming, line 645: There's an issue left here:

`db_session` already relies on `app_config`, so we need to ensure that pytest only needs to resolve a single fixture, not both `app_config_dbsession_from_env` and `app_config`.

tests/conftest.py
benjaoming: So actually this error was very helpful to understand that a bit more refactoring is needed to arrive at the right place :+1: 

Will try to figure out a minimal change :hourglass_flowing_sand:

benjaoming: > Is there any reason for us to maintain these two fixtures, or could we collapse these down to app_config only?

So, getting back to this question, I can think of 2 reasons to keep them separate:

1) To keep McCabe complexity low :innocent: Not sure if you have mccabe rules on or off in flake8, but very long routines get hard to read and maintain. There's a clear responsibility defined for each of these two fixtures, so seems easy to maintain separate fixtures.

2) Actually pytest can help here: It knows how to reuse the same instance of a fixture, even if the path is different (see diagram below). So while the `app_config` fixture may have a variant, the `database` fixture remains the same.

I'm exploring this alternative approach: https://github.com/benjaoming/warehouse/compare/issue_15634_test_user_profile-investigation...benjaoming:warehouse:issue_15634_test_user_profile-investigation-refactor-dbsession?expand=1

1) It avoids pulling in `db_session` in `webtest` but instead creates the session, using its own variant of `app_config` (`app_config_dbsession_from_env`).

2) ...which effectively avoids having 2 app configs instantiated when only 1 is needed in the session.

A quick drawing (trying out SmartDraw, hence a bit messy)

image

Seeing this does make me want to add some documentation w ascii/mermaid diagrams to visualize how fixtures depend on each other.

There's also something about the names of fixtures and helper functions that could be more intuitive? `webtest` is actually Warehouse's own "test client" - that is, if we look at Django or FastAPI, the name of the fixture would be better as `client` or simply `app` (Pyramid terminology).

<details><summary>Similar conftest.py for FastAPI</summary>
<p>

```
import pytest

from fastapi.testclient import TestClient
from main import create_app


@pytest.fixture
def app():
    return create_app()


@pytest.fixture
def client(app):
    return TestClient(app)
```

</p>
</details>

tests/conftest.py
benjaoming: The above issue is solved in https://github.com/benjaoming/warehouse/compare/issue_15634_test_user_profile-investigation...benjaoming:warehouse:issue_15634_test_user_profile-investigation-refactor-dbsession?expand=1

@miketheman if you're happy to review with the above approach, I'll pull it in and make it ready :blush: 

Otherwise, please feel free to suggest an alternative. I'm not saying that this is the best minimal+correct approach... but more that I didn't have any other sudden ideas while working on it.

di: @benjaoming That looks good to me!

benjaoming: Great, I've pushed the approach to this PR branch :+1:

di (approved):
This looks great to me, thank you! Will wait for @miketheman to approve before merging.

miketheman (approved):
@benjaoming Thanks so much for providing all the details and context, it's been super helpful to see how it gets peeled apart as it were.

I'm 👍 here as is, and if you wanted to take another pass once merged at addressing some of the refactors/todo items to make the fixtures a little "cleaner" and less circular, I'd love to review and merge that as well.

In the meantime, we can start to benefit from a working `webtest` fixture! 🚀

benjaoming:
@miketheman indeed, it's much better to get started with some test cases before spending additional time (prematurely) designing fixtures... those can always be improved when someone gets the inspiration to do so :+1: 

was fun to work on! thanks for commenting and guiding :pray: 

and once again, I have to say that it was a really smart move by @jzohrab to leave a PR to help someone getting started with this.. that concept should have a name, if it doesn't already.

jzohrab:
> that concept should have a name, if it doesn't already.

We could call it a PFTMR - "Please Fix This Monstrosity Request" :-) Thanks @benjaoming for taking this over the finish line, I didn't have the mental bandwidth to take it on, and it looks like it was a lot of work.  Critical work too, so great stuff.

PR #16032, opened 2024-05-31
Move request sanitization and upload disallowed out of file_upload

dstufft:
This is extracting a (small) piece of https://github.com/pypi/warehouse/pull/14716 out into it's own PR, with the overarching goal of making the `file_upload` endpoint easier to read and reason about what's happening. 

There's no real functionality change here (except we move request sanitizing to happen prior to the read only checks, which was done primarily because while the _current_ sanitizers and the _current_ checks are wholly independent of each other, that might not always be the case, so we sanitize first.

Beyond that, all this means is that when reading the `file_upload` method, it's easier to gloss over these checks, as they're just decorators instead of being part of the method body.

This also means that the tests that test this functionality no longer need to do heavy weight things like setup databases and make DB queries, making our tests just a little bit faster.


warehouse/forklift/decorators.py
```diff
@@ -0,0 +1,127 @@
...
+                    request.POST[key] = value.replace("\x00", "\\x00")
+
+        # Check if any fields were supplied as a tuple and have become a
+        # FieldStorage. The 'content' field _should_ be a FieldStorage, however,
+        # and we don't care about the legacy gpg_signature field.
+        # ref: https://github.com/pypi/warehouse/issues/2185
+        # ref: https://github.com/pypi/warehouse/issues/2491
+        for field in set(request.POST) - {"content", "gpg_signature"}:
+            values = request.POST.getall(field)
+            if any(isinstance(value, cgi.FieldStorage) for value in values):
```
miketheman, line 54: Thinking out loud - `cgi` is deprecated and going to be removed in 3.13 - are we still getting things coming in as FieldStorage instances? Might be good to insert a sentry message to try and track down if this is still happening. Not a merge blocker, but would be happy to re-review something like that.

miketheman (approved):
Nice extraction, it indeed does make the forklift a little easier to read.

I didn't think we used the `decorator` kwarg for `view_config` for anything other than HTTP header stuff (cache control, vary, etc) - but I don't see why it can't be as flexible for this kind of purpose either - it was a little surprising since it's the first time I'm really thinking about it.

Is there an advantage to having the implementation be a decorator pattern vs an inline function call? Does Pyramid handle this differently?

In any case, nothing blocking this from going forward, just my own curiosities.

firefly9966 (approved):

warehouse/forklift/decorators.py
dstufft: In the interim, someone (I'm guessing you!) added this already, so this is taken care of now.

dstufft:
I removed the approval on this since it got approved ~2 years ago and then it got lost in the shuffle, and it didn't feel right to just assume that approval still stood.

warehouse/forklift/legacy.py
```diff
@@ -530,6 +515,7 @@ def _commonpath(values):
     require_methods=["POST"],
     has_translations=True,
     permit_duplicate_post_keys=True,
+    decorator=[sanitize, ensure_uploads_allowed],
```
ewdurbin, line 518: do we have control over the order of execution of these decorators? are they stable?

in other words, will this check `ensure_uploads_allowed` first?

ewdurbin (approved):
Overall this is progress in the right direction! Although I didn't audit how many (if any) other instances of it there are, I'm a little hesitant about what a decorator approach looks like for other strictly non-packaging concerns we have baked into the legacy upload endpoint.

warehouse/forklift/legacy.py
dstufft: The order is as if you applied the decorators in this order to the function itself. First one wrapping the second one. So this is equivalent to 

```
@sanitize
@ensure_uploads_allowed
def foo():
    …
```

The benefit of decorators kwarg over decorating. “Naturally” is:

- The kwarg gets a normalized view function signature, so you don’t have to adapt to the different signatures that pyramid supports.
- The decorator applies when running through the pyramid framework, but not when you’re just calling the view function as if it were a normal function. This isn’t super meaningful for sanitize, but it means that tests don’t need to setup admin flags just to test file uploads.

dstufft: FTR, I don’t think decorators are a great pattern for the “meat and potatoes” of what a view does. But I think they’re great for the sort of ancillary checks that are still important but aren’t part of the core logic of what the view does. 

These two were obvious to me. Checking an admin flag isn’t really pertinent to understanding how a file uploads, it’s similiar to how pyramid has extracted out permissions where most views don’t need to manually check them, because that’s boilerplate noise most of the time.

PR #16047, opened 2024-06-04
admin: malware refactor and small features

miketheman:
A few pieces of work I wanted to get out of the way before other code changes take advantage of these.

### DIFF
--- a/warehouse/utils/project.py
+++ b/warehouse/utils/project.py
@@ -13,9 +13,10 @@
 import re
 
 from pyramid.httpexceptions import HTTPSeeOther
+from sqlalchemy.sql import func
 
 from warehouse.packaging.interfaces import IDocsStorage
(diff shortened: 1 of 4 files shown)


ewdurbin (approved):

warehouse/utils/project.py
```diff
@@ -65,6 +66,32 @@ def confirm_project(
         )
 
 
+def prohibit_and_remove_project(
+    project: Project | str, request, comment: str, flash: bool = True
+):
+    """
+    View helper to prohibit and remove a project.
+    """
+    # TODO: See if we can constrain `project` to be a `Project` only.
```
miketheman, line 75: I intend to come back to this, and probably take a `str` only, since we don't actively use the `Project` object other than the name. It's resolved to a `Project` later on before calling `remove_project()`.

Issue #16010, opened 2024-05-24
Refactor Help Scout code to a Service


miketheman:
We currently have two general use cases for Help Scout:

1. An inbound API for users of Help Scout to fetch metadata about the current user/conversation
2. An outbound API integration where a Task will authenticate and call the HS API to post stuff

The first use case is fine - nothing to change. The second use case is where we should refactor.

The initial implementation was naive, since it assumes that we would only ever post to Help Scout.
See here:

https://github.com/pypi/warehouse/blob/15a9a6c39b1bc40f3dbc67b74b86dc0a104b7d44/warehouse/observations/tasks.py#L90-L143
https://github.com/pypi/warehouse/blob/15a9a6c39b1bc40f3dbc67b74b86dc0a104b7d44/warehouse/observations/tasks.py#L149-L165

In order to use Help Scout in other contexts, such as Admin UI "actions taken" back to the same conversation, we need a better interface (a la service), which we can authenticate a client, and use the client as many times as we need in other places in the code.

PR #16061, opened 2024-06-06
feat(admin): views for per-project malware reports with verdicts

miketheman:

### DIFF
--- a/tests/unit/admin/test_routes.py
+++ b/tests/unit/admin/test_routes.py
@@ -262,6 +262,13 @@ def test_includeme():
             domain=warehouse,
         ),
         pretend.call(
+            "admin.malware_reports.project.list",
+            "/admin/projects/{project_name}/malware_reports/",
+            factory="warehouse.packaging.models:ProjectFactory",
+            traverse="/{project_name}",
+            domain=warehouse,
+        ),
+        pretend.call(
             "admin.malware_reports.detail",
             "/admin/malware_reports/{observation_id}/",
             domain=warehouse,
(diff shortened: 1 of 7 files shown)


ewdurbin (commented):
Overall this looks great! I really like the simple actions surfaced at the bottom of the aggregated view for a project.

**EDIT: Below is incorrect, I was looking at the `is_dependency_confusion` tab. Confused indeed.**

<details>
  <summary>**EDIT: Below is incorrect, I was looking at the `is_dependency_confusion` tab. Confused indeed.**</summary>
The list view seems to have a grouping bug, after generating sample observations locally I noticed that while two projects had 3 reports each, one was being misattributed to the wrong project:

Screenshot 2024-06-07 at 10 07 28 AM
Screenshot 2024-06-07 at 10 07 34 AM


Screenshot 2024-06-07 at 10 09 06 AM
Screenshot 2024-06-07 at 10 09 13 AM

</details>

ewdurbin (commented):
This is great, next steps are to interact with the HS ticket or will we still need to respond to each individually?

miketheman:
> This is great, next steps are to interact with the HS ticket or will we still need to respond to each individually?

I captured the need to refactor HS to a service here so we can use it in these views: https://github.com/pypi/warehouse/issues/16010

I'm thinking of deferring that effort for now, and come back to it once we're further along on quarantine - unless you think it's more important for now?

ewdurbin (approved):
got it. just worried about keeping things "in sync" and the potential for missed responses to reporters, but since you're the primary one working malware reports I trust you on the ergonomics.

PR #16062, opened 2024-06-06
add clickpy as option for analytics

gingerwizard:
We are the maintainers of ClickHouse, the open source database, index the data for pypi daily and provide a free app and clickhouse service to query at clickpy.clickhouse.com.

We'd like to add this as a link. From what i can tell we also need to provide translations. Do we need to provide this for all languages?


di:
Hey @gingerwizard, thanks for the PR. We added this to https://pypi.org/stats/ in #16198 instead (might take a bit for that to fall out of cache and get updated).

PR #16089, opened 2024-06-11
run CI on depot-ubuntu-22.04-arm runners

ewdurbin:
Our production deployment is ARM64, so this is a nice harmonizer!


miketheman (approved):
💪💪💪💪

PR #16179, opened 2024-06-27
feat(admin): Project Quarantine

miketheman:
This set of changes introduces a concept of Project Lifecycle Status, and specifically adding and removing projects from a quarantine status.

There are no end-user facing changes yet, and those will only be introduced after this PR is merged/deployed (so there's no race condition with ORM changes and columns not existing in time...)

The intent is to have a status that can convey that we are not certain if a project is good or bad yet - but it's concerning enough that we want to make in unavailable to installers or mirrors. Those changes will follow later.

This is mostly to get a feel for how it will work on the Project's model and how things enter and exit quarantine.

<div>
    <a href="https://www.loom.com/share/a472c06ab76542fca1ecaaef2a419f3d">
      <p>Admin Interface for Quarantine 👩‍💻 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/a472c06ab76542fca1ecaaef2a419f3d">
      
    </a>
  </div>


miketheman:
<div>
    <a href="https://www.loom.com/share/a472c06ab76542fca1ecaaef2a419f3d">
      <p>Admin Interface for Quarantine 👩‍💻 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/a472c06ab76542fca1ecaaef2a419f3d">
      
    </a>
  </div>

warehouse/admin/routes.py
```diff
@@ -180,6 +180,13 @@ def includeme(config):
         traverse="/{project_name}/{version}",
         domain=warehouse,
     )
+    config.add_route(
```
ewdurbin, line 183: Would a matching admin button make sense to have generally? I know _most_ of this kind of activity is due to malware reports, but there could be other valid reasons to quarantine something without an existing observation.

warehouse/utils/project.py
```diff
@@ -92,6 +98,64 @@ def prohibit_and_remove_project(
...
+    project.lifecycle_status_note = f"Quarantined by {request.user.username}."
+
+    project.record_event(
+        tag=EventTag.Project.ProjectQuarantineEnter,
+        request=request,
+        additional={"submitted_by": request.user.username},
+    )
+
+    request.db.add(
+        JournalEntry(
```
ewdurbin, line 115: Thinking about this, I'm not sure we want to emit these journal entries until we are making other user facing changes.

I'm unsure how much we want to broadcast at all. I'm thinking about the scenario where a false positive is quarantined and then released, and the repercussions that could have.

For now... I think its fine but it raises some questions for me.

warehouse/migrations/versions/14ad61e054cf_add_project_lifecycle_status.py
```diff
@@ -0,0 +1,81 @@
...
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "14ad61e054cf"
+down_revision = "b14df478c48f"
+
+
+def upgrade():
```
ewdurbin, line 29: Seems like this should be fast since we're adding nullable columns only, but taking a moment to pause and think twice is always good.

warehouse/migrations/versions/c978a4eaa0f6_add_observation_related_name.py
```diff
@@ -60,17 +60,20 @@ def upgrade():
         ),
     )
 
+    # NOTE: Commenting out the following to avoid failing other migrations during tests.
```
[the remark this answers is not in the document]
miketheman: I don't think we've used ORM models in migrations very often - and specifically haven't used an ORM model that has later had a column changed.
I'm thinking that future data migrations ought to avoid using the live models.
This is also noted in https://stackoverflow.com/a/36087108/244037

```python
# Copy the model definitions into the migration script if
# you want the migration script to be robust against later
# changes to the models. ...
```

warehouse/admin/routes.py
miketheman: Certainly can do! I'm thinking at the "Delete project" card at the bottom of the details - unless you think it ought to be somewhere else?

warehouse/migrations/versions/14ad61e054cf_add_project_lifecycle_status.py
miketheman: Since this touches the often-used Project table, I wanted to be a bit more safe than having to scramble.

Do we have either CLI-execution metrics/timings, or specifically timing for migrations emitted somewhere, or is that something we should add?

warehouse/utils/project.py
miketheman: I can definitely remove these until we have user-facing behaviors (which should be the next PR in this effort). I don't imagine we'll be using the feature in production until it actively performs the actions described.

>  I'm thinking about the scenario where a false positive is quarantined and then released, and the repercussions that could have.

Definitely a concern. From poking at `bandersnatch`, it appears that this will have the intended impact - the code tolerates a 404 after having synced the project once. If the user has "cleanup" mode, it'll be removed from the mirror, which is also probably fine - since that has the intended affect as well.
And if the project appears after exiting quarantine, the entry updating the changelog will cause the mirror to check in again. If it wasn't cleaned up, no harm, no foul. If it was cleaned up, it'll re-sync.

@dstufft had thoughts on omission leading to mirror inconsistencies - maybe you can detail those here?

warehouse/admin/routes.py
miketheman: Added in 4e76f598a93d62105347a4081bd1cdd14f5dfacf

warehouse/utils/project.py
ewdurbin: Discussion on the implications of publishing "quarantined" and "un quarantined" events led me to think its better t one explicit.

Sounds like local testing shows that bandersnatch (at least latest) handles these events well, so I think I'm resolved here.

ewdurbin (approved):

warehouse/packaging/models.py
```diff
@@ -162,6 +162,11 @@ def __contains__(self, project):
             return True
 
 
+class LifecycleStatus(enum.StrEnum):
+    QuarantineEnter = "quarantine-enter"
+    QuarantineExit = "quarantine-exit"
```
di, line 167: Should this include a FSM like we use for `EmailStatus`? https://github.com/pypi/warehouse/blob/15de75987757412a0cb38249ea60697938406c50/warehouse/email/ses/models.py#L40

It might not seem necessary now but I could see it becoming useful if/when we start adding additional statuses.

warehouse/utils/project.py
dstufft: If we're hiding the package from the `/simple/` index, then that _is_ a user facing change yea?

The issue becomes a sequence of events like this:

1. User uploads a package, triggering a journal entry.
2. File gets hidden from `/simple/` without a journal entry.
3. A mirror asks PyPI for changes since some serial, which includes (1), and gets the package.
4. Mirror syncs the package, missing the hidden file (because it's hidden), and marks itself as having synced past (1).
5. File gets unhidden, but no journal entry so mirrors don't sync the Package.

Or alternatively:

1. User uploads a package, triggering a journal entry.
2. Mirror syncs past (1).
3. File gets hidden on PyPI without a journal entry.
4. Mirror continues to serve file, causing different behavior between PyPI and the mirror.
   - This could be combined with the above, causing some mirrors to have it and some not, but all of them think they're at the latest version.


One of the implicit promises we've made is that any meaningful changes to the `/simple/` pages will come with an accompanying entry in the journals table, as that's how the mirrors know that the contents of those pages have changed and they need to resync them.

warehouse/packaging/models.py
miketheman: That's an interesting idea - thanks for suggesting! I'm less familiar with automat or the specifics of our implementation.

I'll defer this for now, considering the unknown effort to implement this, and make note for further lifecycle work that we might entertain.
cc @woodruffw as an interested party

warehouse/packaging/models.py
```diff
@@ -183,6 +188,16 @@ class Project(SitemapMixin, HasEvents, HasObservations, db.Model):
...
+    lifecycle_status: Mapped[LifecycleStatus | None] = mapped_column(
+        comment="Lifecycle status can change project visibility and access"
+    )
+    lifecycle_status_changed: Mapped[datetime_now | None] = mapped_column(
+        onupdate=func.now(),
+        comment="When the lifecycle status was last changed",
+    )
+    lifecycle_status_note: Mapped[str | None] = mapped_column(
+        comment="Note about the lifecycle status"
+    )
```
woodruffw, line 200: Question: You know way more about schemas and effective layouts than I do, so possibly a dumb question: could this be a `Mapped[Lifecycle]` instead, to avoid the `lifecycle_{attribute}` pattern? Or would that necessitate a `relationship` and therefore be slower 🙂

warehouse/packaging/models.py
woodruffw, line 167: General question: this and the `lifecycle_status` field imply a state transition graph like the following:

```
None -> (QuarantineEnter <-> QuarantineExit)
```

...which makes sense to me, but might make adding other `LifecycleStatus` states a bit verbose (since each may need to have separate `Enter` and `Exit` states, and define whether a transition from one "color" exit to another "color" enter is valid).

Instead, would it make sense to have `LifecycleStatus.Nominal` (or similar) as a general "exit" state (and default state as well)? That would allow for recording of transitions without requiring each lifecycle state to have its own "exit" marker.

warehouse/utils/project.py
```diff
@@ -92,6 +98,64 @@ def prohibit_and_remove_project(
...
+    Quarantine a project. Reversible action.
+    """
+    project.lifecycle_status = LifecycleStatus.QuarantineEnter
+    project.lifecycle_status_note = f"Quarantined by {request.user.username}."
+
+    project.record_event(
+        tag=EventTag.Project.ProjectQuarantineEnter,
+        request=request,
+        additional={"submitted_by": request.user.username},
+    )
```
woodruffw, line 112: Nitpick: this tag might need another branch in the project security events HTML, to pretty-print it rather than dump it as a raw event 🙂

warehouse/packaging/models.py
di: See also https://github.com/pypi/warehouse/pull/16179#discussion_r1667152383 😉

warehouse/utils/project.py
miketheman: But yes, absolutely - as part of the user-facing work that should follow this PR.

warehouse/packaging/models.py
miketheman: (I upgraded your state diagram 😉 )

Similar to the other comments di linked to, I think that these states should emerge as we discover them, and decide if the FSM complexity is worth it.

warehouse/packaging/models.py
woodruffw: Makes sense, thanks for explaining!

warehouse/packaging/models.py
woodruffw: > (I upgraded your state diagram 😉 )

Nice, I always forget GH MD can do Mermaid.

I totally missed @di's original comment, sorry about that 😅 -- I'll fold my thoughts in there!

woodruffw: (Following up from https://github.com/pypi/warehouse/pull/16179#discussion_r1668897815)

I agree with @di about there being value in having this be modeled as a FSM (although it doesn't necessarily have to use `automat`, since it can be an anydirectional machine as long as there's a single "nominal" state. But deferring on that also seems fine, given your point about data migration + unknown effort; just something to consider in the longer run.

(As part of that though, maybe it makes sense to omit `QuarantineExit` and use only `None` as the terminal state? That would make future changes easier I think. But maybe that makes it harder to track the semantics of "leaving quarantine"...)

PR #16197, opened 2024-07-01
oidc/services: fix mischaracterized error

woodruffw:
This fixes a confusing error render, which previously assumed that all non-`PyJWTErrors` were really leaky exceptions from `pyjwt`'s APIs. This is true for the later-down callsite, but not for this one (where the leaky errors have been patched upstream, and any remaining errors are from sporadic Redis or JWK Set retrieval failures).

See: https://python-software-foundation.sentry.io/share/issue/c7d9e2b44bcb4e1da1cb722ac4c13256/

### DIFF
--- a/warehouse/oidc/services.py
+++ b/warehouse/oidc/services.py
@@ -223,19 +223,25 @@ class OIDCPublisherService:
         try:
             key = self._get_key_for_token(unverified_token)
         except Exception as e:
-            # The user might feed us an entirely nonsense JWT, e.g. one
-            # with missing components.
-            self.metrics.increment(
-                "warehouse.oidc.verify_jwt_signature.malformed_jwt",
-                tags=[f"publisher:{self.publisher}"],
-            )
-
-            if not isinstance(e, jwt.PyJWTError):
+            if isinstance(e, jwt.PyJWTError):
+                # The user might feed us an entirely nonsense JWT, e.g. one
+                # with missing components.
+                self.metrics.increment(
+                    "warehouse.oidc.verify_jwt_signature.malformed_jwt",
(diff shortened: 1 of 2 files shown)


woodruffw:
RTD failure looks unrelated:

```
fatal: unable to access 'https://github.com/pypi/warehouse.git/': GnuTLS recv error (-110): The TLS connection was non-properly terminated.
```

warehouse/oidc/services.py
```diff
@@ -223,19 +223,25 @@ def verify_jwt_signature(self, unverified_token: str) -> SignedClaims | None:
...
+                )
                 with sentry_sdk.push_scope() as scope:
                     scope.fingerprint = e
-                    # Similar to below: Other exceptions indicate an abstraction
-                    # leak, so we log them for upstream reporting.
-                    sentry_sdk.capture_message(f"JWT backend raised generic error: {e}")
+                    sentry_sdk.capture_message(
+                        "verify_jwt_signature failed to obtain matching key for JWT: "
+                        f"{type(e).__name__}: {e}"
+                    )
```
di, line 244: I think it might be OK to just let these raise? We're not really getting exceptions from pyjwt that aren't an instance of PyJWTError anymore AFAICT.

woodruffw: Makes sense, I'll remove all the handling besides the JWT case.

woodruffw: https://github.com/pypi/warehouse/pull/16197/commits/7a93f0f77dc8fed7561e5a33057053bcb1e93abe

di (approved):

PR #16206, opened 2024-07-02
Parallelize the unit tests

### LINKED ISSUE #16178: Parallelize the test suite
From conversation wit @miketheman: the tests *do* currently parallelize cleanly, but coverage is broken in the process. My SWAG explanation is that coverage's intermediate files may be generated with the same seed in each process, resulting in clobbered outputs.

WIP branch from above: https://github.com/pypi/warehouse/compare/main...miketheman:miketheman/pytest-xdist?expand=1

I've asked @DarkaMaul to take a look at this once his queue is empty 🙂

woodruffw:
Opening this o/b/o @DarkaMaul from @miketheman's original work.

On my local machine, this results in a ~2 speedup in testing: from roughly 2 minutes for the entire suite to 1. More speedups *may* be possible, e.g. in stand-up and coverage merging.

Closes #16178.


sitecustomize.py
woodruffw, line 1: NB: This file could probably live somewhere other than the repo root, as long as it gets copied into the right place at container time.

woodruffw:
Looks like the speedup isn't as significant on CI, which is interesting.

miketheman (approved):
image

Very cool, thanks for figuring out all the pieces to this puzzle.

On my 12-core machine, drops the tests runtime from ~83 to ~41 seconds - that's about 50% speed up 🐎

di:
Is there maybe a way to locally disable this? For whatever reason, starting up the xdist workers on my machine made the test suite run take _longer_ than it did before this change. I'm also missing the output which shows which test files failed as the test suite runs, as they're no longer displayed.

woodruffw:
> Is there maybe a way to locally disable this?

This should disable parallelization (by forcing everything onto one worker): `make tests TESTARGS='-n 1'`. On my local machine that spawns only one worker, at least.

I'll look into improving the failure render as well.

di:
FWIW, looks like `-n 0` fully disables pytest-xdist. Using `-n 1` for me was still just as slow.

PR #16245, opened 2024-07-09
requirements: add pytest-sugar

woodruffw:
This is an attempted follow-on to https://github.com/pypi/warehouse/pull/16206 and https://github.com/pypi/warehouse/pull/16206#issuecomment-2218093684 in particular: when the tests are invoked with `-v`, this results in a slightly nicer output as well as a colorized testcase summary.

Unfortunately, by default this still doesn't allow test filenames to be shown; that's a known limitation of `pytest-xdist`, even when run with `-n 1`: https://github.com/pytest-dev/pytest-xdist/issues/450

CC @miketheman in particular for opinions: this makes the output _very slightly_ nicer, in exchange for yet more dev-time deps. May not be worth it, curious what you think 🙂 

Ref for `pytest-sugar`: https://pypi.org/project/pytest-sugar/

### DIFF
--- a/requirements/tests.in
+++ b/requirements/tests.in
@@ -5,6 +5,7 @@ pretend
 pytest>=3.0.0
 pytest-icdiff
 pytest-postgresql>=3.1.3,<7.0.0
+pytest-pretty
 pytest-randomly
 pytest-socket
 pytest-xdist
--- a/requirements/tests.in
+++ b/requirements/tests.in
@@ -5,9 +5,9 @@ pretend
 pytest>=3.0.0
 pytest-icdiff
 pytest-postgresql>=3.1.3,<7.0.0
-pytest-pretty
 pytest-randomly
 pytest-socket
+pytest-sugar
 pytest-xdist
 pytz
 responses>=0.5.1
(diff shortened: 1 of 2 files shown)


JacobCoffee (approved):
+1 to anything that makes pytest more readable. 

There is also https://pypi.org/project/pytest-sugar/ that we use in https://github.com/litestar-org/pytest-databases/,

woodruffw:
TIL about `pytest-sugar`! I'll give that a try as well.

woodruffw:
Looks like `sugar` and `xdist` fight for the output a bit, but the results are still decently readable (and include the filename, unlike `xdist` + `pretty`). I'll induce a test failure and take a screencap to see if the result is to @di's liking 🙂

woodruffw:
The render looks good to me, and appears to still parallelize well!

Screenshot 2024-07-09 at 1 50 43 PM

woodruffw:
NB that the percentages are slightly wrong/OOO on `sugar`, but IMO this is a pretty minor defect.

miketheman (approved):
Passing even more makes it super expressive: `make tests TESTARGS="-vvv"`  and emits the failure content as it's running, making debugging while tests are running possible, so that's nice.

Combined with `-x` to exit on first failure, and `--ff` to run the failed tests first, we should have enough to combine for specific executions

di (approved):
This is great, thanks!

PR #16247, opened 2024-07-09
Admin updates for support

ewdurbin:
While reviewing some support tickets with @Thespi-Brain and @di, a few things were noticed that could be improved:

- Surface Organization membership in user detail
- Surface Organization invitations in user detail
- Surface Project invitations in user detail
- Expose all available User.events in admin (not just most recent)
- When resetting a password via admin, set a unique status (and improve the email! we were sending the "compromised" one!)

### DIFF
--- a/warehouse/templates/email/password-reset-by-admin/body.html
+++ b/warehouse/templates/email/password-reset-by-admin/body.html
@@ -15,23 +15,23 @@
 
 
 {% block content %}
-<h3>{% trans %}What?{% endtrans %}</h3>
+<h3>What?</h3>
 <p>
-  {% trans %}PyPI administrators have have initiated a password reset for your account.
-  You will no longer be able to log in or upload to PyPI with your existing password.{% endtrans %}
+  PyPI administrators have have initiated a password reset for your account.
+  You will no longer be able to log in to PyPI with your existing password.
 </p>
 
-<h3>{% trans %}What should I do?{% endtrans %}</h3>
+<h3>What should I do?</h3>
 <p>
-  {% trans href=request.route_url('accounts.request-password-reset', _host=request.registry.settings.get('warehouse.domain')) %}
-  To regain access to your account, <a href="{{ href }}">reset your password</a> on PyPI.
-  {% endtrans %}
+  To regain access to your account,
+  <a href="{{ request.route_url('accounts.request-password-reset', _host=request.registry.settings.get('warehouse.domain')) }}">reset your password</a>
+  on PyPI.
 </p>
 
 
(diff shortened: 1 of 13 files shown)


ewdurbin:
Email unverified:

Screenshot 2024-07-09 at 5 09 45 PM

Project invitations:

Note: the intended role is encoded in the token, so it's hard to unwrap in the template. We really only need to know that someone _has_ been invited so I think this is fine?

Screenshot 2024-07-09 at 5 10 02 PM

Organization membership and invitations:

Note: the intended role is encoded in the token, so it's hard to unwrap in the template. We really only need to know that someone _has_ been invited so I think this is fine?
Screenshot 2024-07-09 at 5 10 08 PM

warehouse/templates/email/password-reset-by-admin/body.html
```diff
@@ -0,0 +1,37 @@
...
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+
+{% block content %}
+<h3>{% trans %}What?{% endtrans %}</h3>
```
di, line 18: I realize this got copy/pasted from elsewhere but we probably shouldn't be translating these emails -- they will be localized to the sender (`request.user`, i.e. us), and not the recipient.

di (approved):
One comment otherwise LGTM!

warehouse/templates/email/password-reset-by-admin/body.html
ewdurbin: ok! I'll untranslated this one.

ewdurbin: hm, end result either way (given that my understanding is _most_ admins are operating in en-us) is an email in US English.

is leaving the translatable strings there better than removing them, in the event we some day offered users to set a language preference in the db... and applied it to emails?

di: Probably an OK assumption to make right now, but might produce confusion if we ever have an admin/moderator that sets their locale to something other than English... maybe better to avoid it? I've been removing these where possible and tracking them in https://github.com/pypi/warehouse/issues/7455

ewdurbin: https://github.com/pypi/warehouse/pull/16247/commits/0b9a63a992a832a4fa868571067858ca89398497

Issue #16194, opened 2024-07-01
Trusted publishing: prevent OIDC credential re-use


woodruffw:
Pointed out by @sethmlarson and @di: an OIDC credential can currently be re-used to issue multiple temporary API tokens, for as long as its expiry window remains open. This results in two potential weaknesses:

1. Disclosure: a user may (incorrectly) believe that the OIDC credential is "spent" after being exchanged, and disclose or log it somewhere public.
2. Exfiltration: an attacker who isn't able to access the temporary API token or OIDC issuance directly _may_ be able to access the extant OIDC credential, and issue a new temporary API token from it.

(2) is unlikely on GitHub, since any context in which the attacker has access to the OIDC credential is also likely one in which they can mint a fresh credential. But this may be still relevant for other supported OIDC IdPs.

Both of these can be addressed by enforcing one-time use of the OIDC credential within Warehouse:

1. Require that all supported OIDC IdPs include the `jti` claim in their JWTs.
2. Enforce per-provider uniqueness of `jti`s: if the same JTI is seen again, the JWT it belongs to should be rejected and token exchange should fail.

For (1), both GitLab and GitHub support `jti`. Google Cloud supports `nonce`, which is subtly different (`jti` is server controlled, `nonce` is user-controlled). I'm not 100% sure what ActiveState since their docs don't mention any claims explicitly 🙂 

For (2), we probably need either a single new table with a unique constraint on `(provider-name, jti)` or a new table per-provider with a unique constraint on `jti`. With both, the table will probably have unbounded growth *unless* we also periodically delete it (on the same cadence as our self-expiring API tokens, maybe?)

di:
> For (2), we probably need either a single new table with a unique constraint on (provider-name, jti) or a new table per-provider with a unique constraint on jti. With both, the table will probably have unbounded growth unless we also periodically delete it (on the same cadence as our self-expiring API tokens, maybe?)

If we make this `(provider-name, jti, exp)`, we could just periodically prune all tokens past their expiration date without worrying about provider-specific token lifetime lengths.

woodruffw:
> Google Cloud supports nonce, which is subtly different (jti is server controlled, nonce is user-controlled)

Just double-checked, and Google Cloud does **not** support either `jti` *or* `nonce` in the OIDC credential retrieval flow that we use.

@th3coop could you provide an example of the claim set that ActiveState offers in its JWTs? I'm curious if your claims include a `jti` we could use 🙂

[...]

PR #16254, opened 2024-07-10
Trusted publishing: prevent OIDC credential re-use

DarkaMaul:
Fixes https://github.com/pypi/warehouse/issues/16194

This PR uses a Redis cache to keep JWT token identifiers in memory until their expiration.
It then prevents reusing an already used JWT to mint a new OIDC token.


tests/unit/oidc/models/test_github.py
```diff
@@ -51,6 +51,7 @@ def test_github_publisher_all_known_claims(self):
             # required unverifiable claims
             "ref",
             "sha",
+            "jti",
```
woodruffw, line 54: I _think_ this should be in `__preverified_claims__` and not marked as a "required unverifiable claim."

warehouse/oidc/models/github.py
```diff
@@ -132,7 +132,7 @@ class GitHubPublisherMixin:
         "job_workflow_ref": _check_job_workflow_ref,
     }
 
-    __required_unverifiable_claims__: set[str] = {"ref", "sha"}
+    __required_unverifiable_claims__: set[str] = {"ref", "sha", "jti"}
```
woodruffw, line 135: Per note above: I think this should be in `OIDCPublisherMixin.__preverified_claims__` instead 🙂

warehouse/oidc/views.py
```diff
@@ -238,6 +238,16 @@ def mint_token(
         publisher = oidc_service.find_publisher(claims, pending=False)
         # NOTE: assert to persuade mypy of the correct type here.
         assert isinstance(publisher, OIDCPublisher)
+    except ReusedTokenError:
+        return _invalid(
+            errors=[
+                {
+                    "code": "invalid-reuse-token",
+                    "description": "valid token, but already used",
```
woodruffw, line 246: Random thought: we might want to make this error message match the general "token invalid" message, to avoid giving an attacker an oracle for whether the token has been reused. I'm not sure that matters in practice though (since the reused token is useless), and may just end up confusing users.

CC @di for thoughts 🙂

tests/unit/oidc/models/test_github.py
DarkaMaul: Moved there.

warehouse/oidc/services.py
```diff
@@ -219,6 +220,23 @@ def _get_key_for_token(self, token):
...
+
+    def store_jwt_identifier(self, jti: str, expiration: int) -> None:
+        """
+        Store the JTI with its expiration date if the key does not exist.
+        """
+        with redis.StrictRedis.from_url(self.cache_url) as r:
+            # Defensive: to prevent races, we expire the JTI slightly after
+            # the token expiration date. Thus, the lock will not be
+            # released before the token invalidation.
+            r.set(jti, exat=expiration + 1, value="placeholder", nx=True)
```
DarkaMaul, line 238: I've added a tiny offset here to make sure we don't encounter a situation where redis expires the key before the token invalidation (and thus, allowing to reuse it). WDYT?

(Just wanted to pinpoint it as the change is rather small and easy to miss)

woodruffw: Makes sense to me. Local jitter can be more than 1s, though -- we could do 5s, since that's the standard-ish "leeway" that we use for TOTP expiries as well.

DarkaMaul: Updated in 1d8156737366208cdf399

warehouse/oidc/models/github.py
DarkaMaul: Updated in [5a65892](https://github.com/pypi/warehouse/pull/16254/commits/5a658928ff16e73a8e7ff5df2fe74422164e772d)

warehouse/oidc/models/github.py
```diff
@@ -138,10 +138,18 @@ class GitHubPublisherMixin:
...
     }
 
+    __preverified_claims__ = {
+        "iss",
+        "iat",
+        "nbf",
+        "exp",
+        "aud",
+        "jti",
+    }
```
woodruffw, line 148: Can we put `jti` in `OIDCPublisherMixin` instead of duping the whole set here? I believe those claims should be strictly optional, so putting it in the mixin instead would make the diff a bit smaller.

warehouse/oidc/services.py
```diff
@@ -219,6 +220,23 @@ def _get_key_for_token(self, token):
...
+
+    def store_jwt_identifier(self, jti: str, expiration: int) -> None:
+        """
+        Store the JTI with its expiration date if the key does not exist.
+        """
+        with redis.StrictRedis.from_url(self.cache_url) as r:
+            # Defensive: to prevent races, we expire the JTI slightly after
+            # the token expiration date. Thus, the lock will not be
+            # released before the token invalidation.
+            r.set(jti, exat=expiration + 5, value="placeholder", nx=True)
```
woodruffw, line 238: I _think_ we need to do some namespacing/key distinction here: right now two different IdPs could in theory (not likely in practice) send the same `jti`, which would result in a false positive. 

Similar to how the JWKS set is stored under `/warehouse/oidc/jwks/{self.publisher}`, I think we probably want to do `/warehouse/oidc/{self.publisher}/{jti}` here 🙁

warehouse/oidc/services.py
```diff
@@ -287,11 +305,30 @@ def find_publisher(
             publisher = find_publisher_by_issuer(
                 self.db, self.issuer_url, signed_claims, pending=pending
             )
+
+            jwt_token_identifier: str | None = signed_claims.get("jti", None)
```
woodruffw, line 309: Nitpick: explicit `None` default not needed:

```suggestion
            jwt_token_identifier: str | None = signed_claims.get("jti")
```

warehouse/oidc/services.py
woodruffw: (Separately from `jti` confusion between IdPs, a compromised IdP could also use the non-namespaced implementation to set an arbitrary Redis key to an integer they control, which could be a security issue e.g. for ratelimiting and session tracking state.)

warehouse/oidc/models/github.py
woodruffw: Another idea here, just to minimize duplication: I think this could be:

```
__preverified_claims__ = OIDCPublisherMixin.__preverified_claims__ | {"jti"}
```

warehouse/oidc/services.py
```diff
@@ -219,6 +220,28 @@ def _get_key_for_token(self, token):
...
+        Store the JTI with its expiration date if the key does not exist.
+        """
+        with redis.StrictRedis.from_url(self.cache_url) as r:
+            # Defensive: to prevent races, we expire the JTI slightly after
+            # the token expiration date. Thus, the lock will not be
+            # released before the token invalidation.
+            r.set(
+                f"/warehouse/oidc/{self.publisher}/{jti}",
+                exat=expiration + 5,
+                value="placeholder",
```
woodruffw, line 241: Nitpick: sentinel value can be an empty string or int instead to lower the memory cost per token here:

```suggestion
                value="",
```

or:

```suggestion
                value=1,
```

warehouse/oidc/services.py
```diff
@@ -287,11 +310,30 @@ def find_publisher(
...
             self.metrics.increment(
                 "warehouse.oidc.find_publisher.ok",
                 tags=metrics_tags,
             )
+
+            if pending is False and jwt_token_identifier:
+                # Of note, exp is coming from a trusted source here,
+                # so we don't validate it
+                expiration = typing.cast(int, signed_claims.get("exp"))
+                self.store_jwt_identifier(jwt_token_identifier, expiration)
```
woodruffw, line 335: Can we merge these two conditional blocks? I think the `store_jwt_identifier` could happen as a fall-through if `token_identifier_exists` returns `False`.

DarkaMaul: I may have misunderstood your suggestion but the reasoning here was :

- Check if the identifier exists and bail out if it does
- Verify the claims (and return early if they are not verified)
- Finally, mark the JTI as already used

We only store JTI associated to valid claims to prevent flooding the database with signed but invalid claims.

Merging the two blocks would require to follow one of this two sequences : 

- Validating the uniqueness of the JTI, storing the JTI, validating the claims ( and maybe remove it if the claims are invalids)
- Validating the claims, validating the uniqueness of the JTI, and storing it.

If believe the current solution is more elegant, but introduces a slight duplication of conditions. However, I'll be happy to change it if that makes more sense.

woodruffw: Oh, you're right -- I missed that `verify_claims` is called between the two blocks. This LGTM then 🙂

warehouse/oidc/views.py
woodruffw, line 246: A valid but used token is invalid 🙂 

```suggestion
                    "description": "invalid token: already used",
```

woodruffw (commented):
One small suggestion, but otherwise this LGTM, thanks @DarkaMaul!

woodruffw (approved):

woodruffw (approved):
LGTM, thank you @DarkaMaul!

warehouse/oidc/services.py
```diff
@@ -287,11 +310,30 @@ def find_publisher(
...
+            jwt_token_identifier: str | None = signed_claims.get("jti")
+            # jti is in the __preverified_claims__ set, so if it was present,
+            # it was already checked
+            if pending is False and jwt_token_identifier:
+                if self.token_identifier_exists(jwt_token_identifier):
+                    self.metrics.increment(
+                        "warehouse.oidc.reused_token",
+                        tags=metrics_tags,
+                    )
+                    raise ReusedTokenError("JWT Token already used to mint a token.")
```
di, line 323: This error message isn't used anywhere as far as I can tell. Would it be better to pass it through when we catch this exception, or should this just be a `raise ReusedTokenError`?

warehouse/oidc/services.py
di, line 335: I could be convinced otherwise, but I would expect us to actually store a JWT identifier as "used" only after it's been used -- specifically, when we've finished minting a token. As is, this invalidates tokens every time we see them, not necessarily every time we use them.

warehouse/oidc/services.py
di, line 323: I was kind of expecting this logic to live within a verifier for the `jti` claim, and for claim verification to just fail if the identifier had been seen before. Is there a reason we're doing this here instead? It feels a bit out of place to be part of the publisher lookup.

warehouse/oidc/views.py
di: I don't think it matters, an invalid token and a reused token are equally useless. I'm more concerned about telling users who are legitimately trying to reuse a token that token reuse is not permitted.

warehouse/oidc/services.py
woodruffw: Yeah, I think we can remove this message outright.

di: (Also, per my other comment, this error message isn't strictly true in every case, as the token might not have been minted.)

woodruffw: @DarkaMaul can offer his rationale as well, but my thinking has two parts:

1. Consistency/single sourcing: not every supported OIDC IdP uses the `jti` claim, so keeping the check at the service/lookup level avoids us needing to add the check explicitly to each publisher provider. We could probably handle that equivalently with a mixin to minimize duplication, however.
2. Claim verification is currently independent of the current `request` object, so we'd have to do a larger refactor to plumb `request` (or the appropriate service) into `check_jti` or similar to handle this on the per-claim level.

warehouse/oidc/services.py
woodruffw: Hmm, good point -- yeah, I think we should perform `store_jwt_identifier` at the point of minting, rather than here.

warehouse/oidc/services.py
di: The first one seems to be a case for adding per-provider `jti` claim verification instead :)

Not sure I follow the second one, what is the `request` object needed for? The configuration for the redis connection?

woodruffw: > The first one seems to be a case for adding per-provider `jti` claim verification instead :)

True!

> Not sure I follow the second one, what is the request object needed for? The configuration for the redis connection?

Yep -- I think we need it in some form, either to create an appropriate `IOIDCPublisherService` _or_ to directly connect to Redis (in which case we need the `request` to access the appropriate Redis URL via `request.registry.settings["oidc.jwk_cache_url"]`.

Now that I think about it, I guess we don't actually need that though -- we can pass `self` (i.e. the service) as a new parameter to `publisher.verify_claims`, and claim verification can use that service to check the JTI.

warehouse/oidc/services.py
DarkaMaul: 👍

warehouse/oidc/services.py
DarkaMaul: Implemented in [b405fae](https://github.com/pypi/warehouse/pull/16254/commits/b405faed27746d8e6de41b7e8651a266e0023ab2)

warehouse/oidc/models/gitlab.py
```diff
@@ -265,6 +278,10 @@ class PendingGitLabPublisher(GitLabPublisherMixin, PendingOIDCPublisher):
         ),
     )
 
+    __preverified_claims__ = OIDCPublisherMixin.__preverified_claims__ | {
+        "jti",
+    }
```
di, line 283: Same question here about this being preverified.

warehouse/oidc/models/github.py
```diff
@@ -322,6 +335,10 @@ class PendingGitHubPublisher(GitHubPublisherMixin, PendingOIDCPublisher):
         UUID(as_uuid=True), ForeignKey(PendingOIDCPublisher.id), primary_key=True
     )
 
+    __preverified_claims__ = OIDCPublisherMixin.__preverified_claims__ | {
+        "jti",
+    }
```
di, line 340: Why is `jti` considered preverified for pending publishers?

warehouse/oidc/models/gitlab.py
```diff
@@ -247,6 +253,13 @@ class GitLabPublisher(GitLabPublisherMixin, OIDCPublisher):
         ),
     )
 
+    __required_verifiable_claims__ = (
+        GitLabPublisherMixin.__required_verifiable_claims__
+        | {
+            "jti": check_existing_jti,
+        }
+    )
```
di, line 261: Same question here about putting this in the mixin.

warehouse/oidc/services.py
```diff
@@ -219,6 +233,28 @@ def _get_key_for_token(self, token):
         unverified_header = jwt.get_unverified_header(token)
         return self._get_key(unverified_header["kid"])
 
+    def token_identifier_exists(self, jti: str) -> bool:
```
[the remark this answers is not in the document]
DarkaMaul: Indeed, that's more consistent.

woodruffw:
(Premature close caused by GitHub being too clever with pattern matching on `fix #XXX`)

di (approved):

PR #16256, opened 2024-07-10
Dev sanity

ewdurbin:
A few things to improve local dev.

- Run more than one gunicorn worker to prevent locking up the server
- Upgrade all user passwords to argon2 so logins work on first shot
- Add a known TOTP secret `IU7UP3EMIPI7EBPQUUSEHEJUFNBIWOYG` for select users

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,7 @@ resetdb: .state/docker-build-base
 	docker compose pause web worker
 	docker compose up -d db
 	docker compose exec --user postgres db /docker-entrypoint-initdb.d/init-dbs.sh
+	docker compose exec --user postgres db psql -U postgres warehouse -f /post-migrations.sql
 	rm -f .state/db-populated .state/db-migrated
 	$(MAKE) initdb
 	docker compose unpause web worker
@@ -114,6 +115,7 @@ resetdb: .state/docker-build-base
 .state/db-populated: .state/db-migrated
 	docker compose run --rm web python -m warehouse sponsors populate-db
 	docker compose run --rm web python -m warehouse classifiers sync
+	docker compose exec --user postgres db psql -U postgres warehouse -f /post-migrations.sql
 	mkdir -p .state && touch .state/db-populated
 
 .state/db-migrated: .state/docker-build-base
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -110,7 +110,7 @@ services:
 
   web:
     image: warehouse:docker-compose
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -114,6 +115,7 @@ resetdb: .state/docker-build-base
 .state/db-populated: .state/db-migrated
 	docker compose run --rm web python -m warehouse sponsors populate-db
 	docker compose run --rm web python -m warehouse classifiers sync
+	docker compose exec --user postgres db psql -U postgres warehouse -f /post-migrations.sql
```
miketheman, line 118: Should this be here as well as earlier? I feel like it should exist here once only - and probably remove this line.
```suggestion
```

miketheman (commented):
All looking nice - one question

Makefile
ewdurbin: the earlier clause is for `make resetdb`, which needs to duplicate the call (at least how we're currently setup)

miketheman: okay. maybe when we tackle db seeds we can rethink a bunch of this.

miketheman (approved):

Makefile
ewdurbin: on closer inspection, you're right. the duplicate in `resetdb` can be removed!

miketheman: https://github.com/pypi/warehouse/pull/16257

docker-compose.yml
```diff
@@ -110,7 +111,7 @@ services:
 
   web:
     image: warehouse:docker-compose
-    command: gunicorn --reload --reload-extra-file=warehouse/api/openapi.yaml -b 0.0.0.0:8000 --access-logfile - --error-logfile - warehouse.wsgi:application
+    command: gunicorn --reload --reload-extra-file=warehouse/api/openapi.yaml -w 4 -b 0.0.0.0:8000 --access-logfile - --error-logfile - warehouse.wsgi:application
```
miketheman, line 114: This change disabled the debug toolbar, as shown:

```
/opt/warehouse/lib/python3.12/site-packages/pyramid_debugtoolbar/toolbar.py:244: UserWarning: pyramid_debugtoolbar has detected that the application is being served by a forking / multiprocess web server. The toolbar relies on global state to work and is not compatible with this environment. The toolbar will be disabled.
```

What was the behavior you observed that led to adding this?

ewdurbin: oops!

I experienced lockups pretty regularly when I had two browsers open with different profiles attempting to interact with the web server.

PR #16260, opened 2024-07-11
Trusted publishing: surface pending publisher name collisions

### LINKED ISSUE #16226: Trusted publishing: pending publisher should warn about ultranormalized name collision

I created a pending publisher when I named my package:

image

But when I tried to use it I [got an error](https://github.com/twm/coveragepth/actions/runs/9824421246/job/27123375336#step:4:12):

> Error: Trusted publishing exchange failure: 
> Token request failed: the server refused the request for the following reasons:
> 
> * `invalid-payload`: The name 'coveragepth' is too similar to an existing project. See https://pypi.org/help/#project-name for more information.

It's probably colliding with [coverage_pth](https://pypi.org/project/coverage_pth/), the abandoned package I'm trying to replace.

The PyPI UI should warn that the pending publisher will fail to create the package due to a name conflict. I'd expect this check to happen on page load, since the UI warns that races are possible:

image

Alternatively, it could be treated as a form validation error like a closer match does:

image

**Additional context**

Related issues:

- #11139
- #2082

twm:
Provide up-front validation of the project name within a pending publisher by:

- Extracting the validation logic from `ProjectService.create_project` to a new `ProjectService.check_project_name` routine
- Update `ProjectService.create_project` to call `check_project_name`
- Thread `ProjectService` into the forms in place of `ProjectFactory`.

Fixes #16226.


di:
Hey @twm, FYI our general policy is to not review draft PRs unless explicitly asked -- let us know if you have any questions here, otherwise we will wait for this to become ready to review.

twm:
Hi @di, thanks for approving the build! Some connectivity issues are making it difficult to get things working locally so it's very helpful. I'll remove the draft status when I'm ready for review!

twm (commented):
Hi @di, this is ready for review now!

firefly9966 (approved):

warehouse/packaging/services.py
```diff
@@ -442,63 +443,70 @@ def _hit_ratelimits(self, request, creator):
         self.ratelimiters["project.create.user"].hit(creator.id)
         self.ratelimiters["project.create.ip"].hit(request.remote_addr)
 
+    def check_project_name(
+        self, name: str, request: Request
```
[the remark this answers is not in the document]
di: They should be the same, you can eliminate the `request` parameter here.

warehouse/packaging/services.py
```diff
@@ -596,6 +595,7 @@ def create_project(
             )
             .all()
         )
+        # TODO: Also dispose of pending publishers with ultranormalized collisions
```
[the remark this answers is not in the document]
di: Yes, let's do that here. We could probably reuse the same email template as the stale publishers (maybe tweaking it to say 'with the same or similar project name').

warehouse/migrations/versions/f7720656a33c_index_ultranormalized_pending_project_.py
```diff
@@ -0,0 +1,58 @@
...
+#
+#       To whatever values are reasonable for this migration as part of your
+#       migration.
+
+
+def upgrade():
+    op.execute(
+        """
+        CREATE INDEX pending_project_name_ultranormalized
+          ON pending_oidc_publishers (ultranormalize_name(project_name));
```
twm, line 52: I opted against `CREATE INDEX CONCURRENTLY` here because [the original migration didn't use it](https://github.com/pypi/warehouse/blob/8a5ad08a1bf65b769efaa97ad0f49b7739a20857/warehouse/migrations/versions/d18d443f89f0_ultranormalize_name_function.py#L45-L50) and I expect that there are orders of magnitude fewer `pending_oidc_publishers` rows than `projects` rows.

twm:
Hi @di, this is ready again. I'd appreciate a review within a few days if you can, as I keep having to rebase through conflicts.

warehouse/packaging/services.py
```diff
@@ -583,15 +579,19 @@ def create_project(
...
             request.db.query(PendingOIDCPublisher)
             .filter(
-                func.normalize_pep426_name(PendingOIDCPublisher.project_name)
-                == func.normalize_pep426_name(project.name),
+                or_(
+                    func.normalize_pep426_name(PendingOIDCPublisher.project_name)
+                    == func.normalize_pep426_name(project.name),
+                    func.ultranormalize_name(PendingOIDCPublisher.project_name)
+                    == func.ultranormalize_name(project.name),
+                ),
```
di, line 594: I think the 'ultranormalized' set will include all of the 'normalized' matches as well? This `or_` condition might not be necessary.

di (commented):
Just one last comment, otherwise LGTM!

warehouse/packaging/services.py
twm: I thought it safer to match on both since there is no formal relationship between the functions that guarantees that to be the case, but I will drop it per your preference.

twm:
Hi @di, sorry for losing track of this. I've addressed your comment and merged forward.

warehouse/packaging/services.py
```diff
@@ -583,15 +579,17 @@ def create_project(
...
         # also sending emails to their owners.
         stale_pending_publishers = (
             request.db.query(PendingOIDCPublisher)
             .filter(
-                func.normalize_pep426_name(PendingOIDCPublisher.project_name)
-                == func.normalize_pep426_name(project.name),
+                or_(
+                    func.ultranormalize_name(PendingOIDCPublisher.project_name)
+                    == func.ultranormalize_name(project.name),
+                ),
```
di, line 592: I think the `or_` is no longer necessary? This can just become:
```suggestion
                func.ultranormalize_name(PendingOIDCPublisher.project_name)
                == func.ultranormalize_name(project.name),
```

di (approved):

di:
Thanks for your patience here @twm!

PR #16264, opened 2024-07-11
feat: quarantine UI updates

miketheman:
Adds quarantine messages in:

- public-facing project page
- logged-in project management page

Adds a help page section on quarantine.

Note: This is presentation-only, the functional changes to the APIs will be in a separate PR.

PR #16265, opened 2024-07-11
feat: implement quarantined projects restrictions

miketheman:
Disallow `pip install`-style commands and file downloads when in quarantine.

Disallow user-made modifications to project when in quarantine.

Follows #16264

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -300,6 +300,10 @@ class TestProject:
 
             acls.extend(acl)
 
+        _perms_read_and_upload = [
+            Permissions.ProjectsRead,
+            Permissions.ProjectsUpload,
+        ]
         assert acls == [
             (
                 Allow,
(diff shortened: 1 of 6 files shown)


tests/unit/packaging/test_models.py
```diff
@@ -266,6 +266,99 @@ def test_acl(self, db_session):
...
+                "group:observers",
+                Permissions.APIObservationsAdd,
+            ),
+            (
+                Allow,
+                Authenticated,
+                Permissions.SubmitMalwareObservation,
+            ),
+        ] + sorted(
+            [(Allow, f"oidc:{publisher.id}", [Permissions.ProjectsUpload])],
```
ewdurbin, line 344: I'm reading this as allowing uploads via Trusted Publisher can still occur for quarantined projects, while uploads via API token would 403. Is that correct?

Is there a reason we would continue to allow quarantined projects to upload one way but not another?

Should we disallow Trusted Publisher uploads, or continue to allow API Token uploads to maintain consistency?

ewdurbin (approved):
Really just one question here.

tests/unit/packaging/test_models.py
ewdurbin: Thinking a moment more, I think we probably want to continue to allow uploads either way. In the event of a false-positive quarantine, breaking uploads is generally more disruptive to projects.

miketheman: > I'm reading this as allowing uploads via Trusted Publisher can still occur for quarantined projects, while uploads via API token would 403. Is that correct?

That is correct, since the change is for the user token ACL to disallow modifications.

> Thinking a moment more, I think we probably want to continue to allow uploads either way. In the event of a false-positive quarantine, breaking uploads is generally more disruptive to projects.

I agree that disallowing uploads is more disruptive, but if the project is using Trusted Publishing, then it's not prevented from uploads. I don't think I've seen any malware **yet** come through a TP upload.
I considered disallowing TP as well, and decided that if they went to the trouble of creating, linking a TP, then at least we'd have a backlink to a more concrete location from whence the code came from.

If you feel strongly that a user-scoped token should continue to allow uploads to a quarantine-flagged project (and not other projects), then I'll modify to allow that, but I wanted to provide the "press pause on this project until it's been reviewed" semantics.
The risk of allowing further uploads is that we have more junk in storages vs not.

firefly9966 (approved):

tests/unit/packaging/test_models.py
ewdurbin: I am still a slight lean towards "allow all uploads to continue", but I am a strong lean towards "the constraint should be consistent between projects".

miketheman: In that light, I agree for consistency, so if allowing TP, I'll update to allow user-based uploads as well.
We can shift that policy later if we find that the support burden rises if folks are uploading new releases and reach out to determine why they are not allowed to install them.

PR #16266, opened 2024-07-11
(admin): Account recovery tooling

ewdurbin:


TODO
miketheman, line 1: Was this file intentionally added?

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,42 @@ if (malwareReportsTable.length) {
   new $.fn.dataTable.Buttons(table, {buttons: ["copy", "csv", "colvis"]});
   table.buttons().container().appendTo($(".col-md-6:eq(0)", table.table().container()));
 }
+
+// Link Checking
+var links = document.querySelectorAll("a");
+links.forEach(function(link){
+  if (link.dataset.checkLinkUrl !== undefined) {
```
miketheman, line 192: Optimize: I _think_ you could replace this selector with `"a[data-check-link-url]"` and thus not iterate over every link on the page.

Since admin uses jQuery, we could use it with `var links = $('a[data-check-link-url]')` and then `link.each(function...` - but I don't mind native JS.

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,42 @@ if (malwareReportsTable.length) {
...
+    fetch(reportLine.url, {
+      method: "GET",
+      mode: "cors",
+    })
+      .then(function(response) {
+        var responseText = "";
+        response.text().then((text) => {
+          responseText = text;
+          console.log(response.status, responseText);
+          if (response.status == 400 && responseText == "Unsupported content-type returned\n") {
```
miketheman, line 203: Is 400 the only response status we're looking to handle? What about other non-400s, like a 429 if we get rate limited for some reason? I think 429 is likely to hit the `else` block and not end up in `.catch`

Would we prefer to verify the response earlier with `if (!response.ok) { ...`

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,42 @@ if (malwareReportsTable.length) {
...
+          } else {
+            reportLine.status = 0;
+            reportLine.element.firstChild.classList.remove("fa-question");
+            reportLine.element.firstChild.classList.add("fa-times");
+            reportLine.element.firstChild.classList.add("text-red");
+          }
+          console.log(reportLine);
+        });
+      })
+      .catch(function(error){
```
miketheman, line 217: Dunno why linter doesn't catch this?
```suggestion
      .catch(function(error) {
```

tests/unit/accounts/test_services.py
```diff
@@ -379,6 +379,22 @@ def test_add_email_rate_limited(self, user_service, metrics, remote_addr):
             )
         ]
 
+    def test_add_email_bypass_ratelimit(self, user_service, metrics, remote_addr):
```
miketheman, line 382: question: should this test assert that email is added successfully as well, not only that the limiter wasn't called?

warehouse/email/services.py
```diff
@@ -32,10 +32,17 @@ def _format_sender(sitename, sender):
 
 
 class EmailMessage:
-    def __init__(self, subject: str, body_text: str, body_html: str | None = None):
+    def __init__(
+        self,
+        subject: str,
+        body_text: str,
+        body_html: str | None = None,
+        sender: str | None = None,
```
miketheman, line 40: 💯 this is nice

warehouse/admin/templates/admin/users/account_recovery/initiate.html
```diff
@@ -0,0 +1,108 @@
...
+      <label class="form-check-label" for="{{ project_name }}"><a href="{{ request.route_url("admin.project.detail", project_name=project_name) }}">{{ project_name }}</a></label>
+      <ul>
+      {% for kind, repo_url in repo_urls %}
+      <li>{{ kind }}: <a href="{{ repo_url }}" data-check-link-url="{{ request.camo_url(repo_url) }}"><i class="fa fa-question"></i> {{ repo_url }}</a></li>
+      {% endfor %}
+      </ul>
+    </div>
+    {% endif %}
+    {% endfor %}
+    {% else %}
```
miketheman, line 87: nit: add indentations for nesting visibility

I have a local branch for a better template linter/formatter, but that's a non-trivial lift to incorporate, so I haven't proposed that yet.

miketheman (approved):
I haven't gone through with a fine tooth comb, considering many of the admin strings are internal only, so if we need to tweak, we can tweak as needed.

I've dropped some things inline that should be considered prior to merging - none are blockers, approving now, and feel free to fix&merge

warehouse/admin/static/js/warehouse.js
ewdurbin: I'll stick with plain js, but good note.

ewdurbin: eb28ba47c

warehouse/admin/static/js/warehouse.js
ewdurbin: this is _pretty_ rudimentary and relies on Camo's response of 400 "Unsupported content-type returned" when it successfully loads a page and doesn't find an image. camo won't rate limit us, and i'm not sure that other statuses like 429 would bubble up.

for now, I'd like to just proceed with this and see how it behaves in practice.

warehouse/admin/static/js/warehouse.js
ewdurbin: 41e60672e

tests/unit/accounts/test_services.py
ewdurbin: 67e26a350ebbbe270899b6ee19a1aaa7b0ec7227

warehouse/admin/templates/admin/users/account_recovery/initiate.html
ewdurbin: 89f7dc18f1d4f6ba3ebecd0e24b410ac2ce27583

warehouse/accounts/models.py
```diff
@@ -293,6 +297,16 @@ def __acl__(self):
                     Permissions.AdminDashboardSidebarRead,
                 ),
             ),
+            (
+                Allow,
+                "group:support",
+                (
+                    Permissions.AdminUsersRead,
+                    Permissions.AdminUsersEmailWrite,
+                    Permissions.AdminUsersAccountRecoveryWrite,
```
miketheman, line 308: These two new permissions need to be added to `group:admins` as well, so that admins can also perform these actions.

warehouse/admin/templates/admin/users/detail.html
```diff
@@ -356,18 +512,16 @@ <h3 class="card-title">Emails</h3>
...
               <div class="form-group">
                 <div class="col-sm-offset-10 col-sm-2">
-                  <button type="submit" class="btn btn-danger" title="{{ 'Submitting requires superuser privileges' if not perms_admin_users_write }}" {{ "disabled" if not perms_admin_users_write }}>Submit</button>
+                  <button type="submit" action="email_form" class="btn btn-danger" title="{{ 'Submitting requires superuser privileges' if not perms_admin_users_email_write }}" {{ "disabled" if not perms_admin_users_email_write }}>Submit</button>
                 </div>
               </div>
             </div>
           </form>
         </div>
-      </div> <!-- .card -->
```
miketheman, line 370: One of these div changes made the Name field appear disjoint from the rest of the form

Screenshot 2024-07-15 at 12 11 04

Not super important, but wanted to note the visual oddity.

ewdurbin: That's not a regression FWIW, it was like that before!

warehouse/accounts/models.py
ewdurbin: Resolved in bc32652e4af681a0f10332ae962f688337398fdb

warehouse/admin/templates/admin/users/detail.html
miketheman: The change makes the "Submit" button part of the Card footer, which isolates the "Name", but it's no biggie - it all still works just fine. I'll poke at it some other time.

ewdurbin: that's also not introduced by these changes, this card has been goofed for some time:

Screenshot 2024-07-15 at 1 19 43 PM

miketheman (approved):

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,40 @@ if (malwareReportsTable.length) {
...
+  fetch(reportLine.url, {
+    method: "GET",
+    mode: "cors",
+  })
+    .then(function(response) {
+      var responseText = "";
+      response.text().then((text) => {
+        responseText = text;
+        console.log(response.status, responseText);
+        if (response.status == 400 && responseText == "Unsupported content-type returned\n") {
```
miketheman, line 202: Prevent JS coercion

```suggestion
        if (response.status === 400 && responseText === "Unsupported content-type returned\n") {
```

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,40 @@ if (malwareReportsTable.length) {
   new $.fn.dataTable.Buttons(table, {buttons: ["copy", "csv", "colvis"]});
   table.buttons().container().appendTo($(".col-md-6:eq(0)", table.table().container()));
 }
+
+// Link Checking
+var links = document.querySelectorAll("a[data-check-link-url]");
```
miketheman, line 190:
```suggestion
const links = document.querySelectorAll("a[data-check-link-url]");
```

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,40 @@ if (malwareReportsTable.length) {
...
+links.forEach(function(link){
+  var reportLine = {bareUrl: link.href, url: link.dataset.checkLinkUrl, status:0, element : link};
+  fetch(reportLine.url, {
+    method: "GET",
+    mode: "cors",
+  })
+    .then(function(response) {
+      var responseText = "";
```
miketheman, line 198:
```suggestion
      let responseText = "";
```

warehouse/admin/static/js/warehouse.js
```diff
@@ -185,3 +185,40 @@ if (malwareReportsTable.length) {
   new $.fn.dataTable.Buttons(table, {buttons: ["copy", "csv", "colvis"]});
   table.buttons().container().appendTo($(".col-md-6:eq(0)", table.table().container()));
 }
+
+// Link Checking
+var links = document.querySelectorAll("a[data-check-link-url]");
+links.forEach(function(link){
+  var reportLine = {bareUrl: link.href, url: link.dataset.checkLinkUrl, status:0, element : link};
```
miketheman, line 192:
```suggestion
  let reportLine = {bareUrl: link.href, url: link.dataset.checkLinkUrl, status:0, element : link};
```

miketheman (approved):
Minor JS nits, still unclear why linter isn't catching these

warehouse/admin/static/js/warehouse.js
ewdurbin: f59ecf07b

warehouse/admin/static/js/warehouse.js
ewdurbin: f59ecf07b

warehouse/admin/static/js/warehouse.js
ewdurbin: f59ecf07b

warehouse/admin/static/js/warehouse.js
ewdurbin: f59ecf07b

PR #16267, opened 2024-07-11
Simplify link underlining

### LINKED ISSUE #16191: Links underlined rather aggressively on Chromium on Android
**Describe the bug**
Links are underlined rather aggressively and different from Firefox on Android or Firefox/Chromium on desktop.

**Expected behavior**
I expect links to be underlined in the same way as on other browsers/platforms.

**To Reproduce**
Visit pypi.org on Chromium on Android

**My Platform**
Vanadium 126.0.6478.122 on GrapheneOS

vanadium

di:
We were using a background image (!) with a gradient (!!) to underline links instead of just using regular text decoration. This should make things slightly simpler and look slightly nicer, and also I'm pretty sure this fixes #16191 as well (which is how I got sniped into changing this). Click to embiggen:

| Before | After |
|--------|--------|
| image | image |

Note that on platforms that support it, descenders will now clear the underline as well:

| Before | After |
|--------|--------|
| image | image |

This PR also adds a slight color change on hover for white links that matches what we do for other links.

### DIFF
--- a/warehouse/static/sass/base/_typography.scss
+++ b/warehouse/static/sass/base/_typography.scss
@@ -107,10 +107,6 @@ strong {
 }
 
 a {
-  text-decoration: none;
-  background-position: 0 1.15em;
-  background-repeat: repeat-x;
-  background-size: 100% 1.5px;
   @include primary-underlined-link;
 
   // Add 'external link' icon to most external links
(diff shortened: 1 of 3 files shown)


ewdurbin (approved):
oh my god

Issue #16268, opened 2024-07-12
"Last released" date in org view incosistent with actual last release date of package


edgarrmondragon:
**Describe the bug**

If you go to https://pypi.org/org/aio-libs/ and look for the `aiohttp` package, you'll notice it says it was last release on Oct 9, 2019:

<img width="826" alt="Screenshot 2024-07-11 at 10 43 40 p m" src="https://github.com/user-attachments/assets/6c4ece98-dfe7-42b6-abab-7124baec7315">

However, clicking on that to see the detailed package page on https://pypi.org/project/aiohttp/, the actual last release date of Apr 16, 2024 is displayed:

<img width="306" alt="Screenshot 2024-07-11 at 10 46 01 p m" src="https://github.com/user-attachments/assets/8a9fb51b-9833-452c-af39-1aa810545e9c">

This `aiohttp` example stands out because it has a very old pre-release with a _semantically_ greater version than the much recent latest stable release.

**Expected behavior**

The dates should be the same.

**To Reproduce**

Steps described above.

**My Platform**

**Additional context**

I can take a stab at submitting a PR if this is indeed a bug.

PR #16269, opened 2024-07-12
Use latest version instead of first in list to display org project

edgarrmondragon:
Fixes https://github.com/pypi/warehouse/issues/16268.

### DIFF
--- a/warehouse/templates/organizations/profile.html
+++ b/warehouse/templates/organizations/profile.html
@@ -63,9 +63,9 @@
           <a class="package-snippet" href="{{ request.route_path('packaging.project', name=project.normalized_name) }}">
             <h3 class="package-snippet__title">{{ project.name }}</h3>
             <p class="package-snippet__meta">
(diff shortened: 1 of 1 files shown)


miketheman:
@edgarrmondragon Please consider that as a Draft PR, admins are less likely to review - since it's unclear if this is ready.

di (commented):
This LGTM but I see a few other places where we're using `releases[0]` that could be updated as well.

edgarrmondragon:
> This LGTM but I see a few other places where we're using `releases[0]` that could be updated as well.

I noticed but I hesitated because of the apparent different intents of the releases list in different contexts. For example, in my private account view I'd still want to view a yanked release as the latest for a project.

Happy to make those changes of course, just wondering if they make sense.

PS: the `_pypi_ordering` attribute seems to affect the ordering of releases rather arbitrarily, and it's outside of users' control.

di (approved):

PR #16273, opened 2024-07-12
don't query for rendered release_descriptions in JSON api

ewdurbin:
JSON api only sends raw and content_type... so lets not query for rendered


ewdurbin:
Before:

Screenshot 2024-07-12 at 12 16 44 PM

After:

Screenshot 2024-07-12 at 12 16 59 PM

di (approved):
Good enough for now to stop the bleeding.

PR #16274, opened 2024-07-12
Remove the fallback to rendering descriptions inline

di:
This resolves a TODO now that we don't have any un-rendered descriptions in the database.


firefly9966 (approved):

Carreau (approved):

miketheman (approved):
Works as advertised

PR #16278, opened 2024-07-15
Disallow trailing new line in `PROJECT_NAME_RE`

twm:
I noticed this while working on #16260: the regex used to validate project names in some context permits trailing newlines in the project name. This is because in Python regexes `$` matches the end of the string *or* a trailing newline. `\Z` is how you match the end of the string.

I do not _think_ that this has any security implications because, though the [equivalent Postgres constraint](https://github.com/pypi/warehouse/blob/2876c73261e51330faf17e9079211aeb3c3cf012/warehouse/migrations/versions/283c68f2ab2_initial_migration.py#L82-L85) uses `$`, in Postgres `$` [matches only the end of the string](https://www.postgresql.org/docs/16/functions-matching.html#POSIX-SYNTAX-DETAILS). The constraint has been present since [time immemorial](https://github.com/pypi/warehouse/commit/82965a56fc67d0c75fce32a2320d2fdb5899f51c).

### DIFF
--- a/tests/unit/utils/test_project.py
+++ b/tests/unit/utils/test_project.py
@@ -26,6 +26,7 @@ from warehouse.packaging.models import (
     Role,
 )
 from warehouse.utils.project import (
+    PROJECT_NAME_RE,
     clear_project_quarantine,
     confirm_project,
     destroy_docs,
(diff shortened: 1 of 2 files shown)


firefly9966 (approved):

miketheman (approved):
Seems reasonable

PR #16284, opened 2024-07-15
misc: fix some sentry captures

woodruffw:
These were using `fingerprint` incorrectly;
I've removed their custom scopes and made them
use the top-level APIs instead, since the scope
isn't adding much more.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -425,7 +425,7 @@ def _process_attestations(request, artifact_path: Path):
             )
         except Exception as e:
             with sentry_sdk.push_scope() as scope:
-                scope.fingerprint = e
+                scope.fingerprint = [e]
                 sentry_sdk.capture_message(
                     f"Unexpected error while verifying attestation: {e}"
                 )
--- a/warehouse/oidc/services.py
+++ b/warehouse/oidc/services.py
@@ -266,7 +266,7 @@ class OIDCPublisherService:
             )
             if not isinstance(e, jwt.PyJWTError):
                 with sentry_sdk.push_scope() as scope:
-                    scope.fingerprint = e
+                    scope.fingerprint = [e]
                     # We expect pyjwt to only raise subclasses of PyJWTError, but
                     # we can't enforce this. Other exceptions indicate an abstraction
                     # leak, so we log them for upstream reporting.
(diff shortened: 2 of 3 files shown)


miketheman (approved):
I'm fine with these, however I wonder if we'll see dual-reporting for the same issue, and thus not need a `capture_message()` call alongside `capture_exception()`

woodruffw:
> I'm fine with these, however I wonder if we'll see dual-reporting for the same issue, and thus not need a `capture_message()` call alongside `capture_exception()`

Yeah, I was wondering about that -- if this is too noisy in practice, I can drop it to just the `capture_exception` 🙂

di:
If `capture_exception` is properly discerning/grouping between different exceptions, I don't see a reason to use `capture_message`.

woodruffw:
> If `capture_exception` is properly discerning/grouping between different exceptions, I don't see a reason to use `capture_message`.

Cool, removing then!

di:
Although... IIRC the reason we were using `capture_message` is because this wasn't happening? How is `capture_exception` here any different than just letting the exception raise?

woodruffw:
Ah, that rings a bell -- yeah, I think the issue here was that some of these exceptions were `ValueError` and similar, which ended up being hard to distinguish. 

I'll push up what I have here, but another option would be to re-introduce `fingerprint` with correct usage, e.g.:

```python
scope.fingerprint = ["some-distinguishing-key", exc]
```

di:
What is the problem with the current usage of `fingerprint`?

woodruffw:
> What is the problem with the current usage of `fingerprint`?

Apparently `fingerprint` can't be a scalar, so `fingerprint = e` wasn't working when `e` was just an exception:

image

(@miketheman observed this -- maybe it's a recent change in the Sentry API?)

di:
Can't we just do `scope.fingerprint = [e]`?

woodruffw:
> Can't we just do `scope.fingerprint = [e]`?

Yep, that's a lot simpler. Sorry, my brain is thoroughly fried -- I've been at a Homebrew hackathon for the last 72 hours 🙂 

I'll rework this PR...

di (approved):

woodruffw:
Gentle ping for merge 🙂

PR #16290, opened 2024-07-17
Support support grab bag

ewdurbin:
Grab bag of fixes from shoulder surfing @Thespi-Brain performing PyPI support work.

- Helpscout Sidebar: Fix user matching when users include `+` in their email address
- Improve copy button functionality in admin. Use a tooltip instead of an element to keep UI from jumping around
- Add a copy-to-clipboard button for user recovery tokens when an account recovery has been initiated
- Fix permissions on admin route for resetting user passwords to allow User.is_support=True to be able to use it
- Update text in initiate account recovery template
- Filter Observations for delivery to Helpscout


package.json
```diff
@@ -28,7 +28,8 @@
     "debounce": "^1.2.1",
     "jquery": "^3.7.0",
     "normalize.css": "^8.0.1",
-    "stimulus-autocomplete": "^3.1.0"
+    "stimulus-autocomplete": "^3.1.0",
+    "tippy.js": "^6.3.7"
```
miketheman, line 32: We get `popper` (an older version) already with admintlte/bootstrap 4, and need to activate it before using.

https://adminlte.io/docs/3.2/dependencies.html
https://getbootstrap.com/docs/4.1/components/tooltips/
https://popper.js.org/docs/v1/

Did you try that before using `tippy`? I don't think we have an existing example to point at, but wanted to call it out before we add another dependency

tests/unit/admin/views/test_helpscout.py
```diff
@@ -55,10 +57,20 @@ def test_valid_auth_no_payload(self, db_request):
...
+        ],
+    )
+    def test_valid_auth_no_such_email(self, db_request, search_email):
+        EmailFactory.create(email="wutang@defjam.com")
+
         db_request.registry.settings["admin.helpscout.app_secret"] = "s3cr3t"
-        db_request.body = b'{"customer": {"email": "wutang@loudrecords.com"}}'
-        db_request.json_body = {"customer": {"email": "wutang@loudrecords.com"}}
+        db_request.json_body = {"customer": {"email": search_email}}
+        db_request.body = json.dumps(db_request.json_body).encode()
```
miketheman, line 73: nice way of doing that

warehouse/admin/static/js/warehouse.js
```diff
@@ -103,13 +106,17 @@ document.querySelectorAll(".btn-group[data-input][data-state]").forEach(function
 
 // Copy handler for copying text, e.g.
 //   - prohibited project names confirmation page
+//   - user account recovieres
```
miketheman, line 109: typo: 
```suggestion
//   - user account recoveries
```

warehouse/observations/tasks.py
```diff
@@ -65,6 +65,13 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
     # Fetch the Observation from the database
     model = request.db.get(Observation, model_id)
 
+    # Check to see if this ObservationKind should be sent
+    if model.kind not in [
+        ObservationKind.IsDependencyConfusion.value[0],
+        ObservationKind.IsMalware.value[0],
```
miketheman, line 71: Optimize: I think you can use the `OBSERVATION_KIND_MAP` here instead of value access, similar to
https://github.com/pypi/warehouse/blob/c41aa5b16e8511eac7090858c9b3f29af0a2ac6a/warehouse/observations/tasks.py#L90

miketheman (approved):
Nice stuff. No blockers, couple of questions inline

warehouse/observations/tasks.py
ewdurbin: ah, thanks definitely felt like I was holding it wrong.

package.json
ewdurbin: neat, a little less ergonomic but its there. 95efc0dbdd9f167abcce85a7e00e644368d4d46a

warehouse/observations/tasks.py
ewdurbin: 281781fa9

PR #16291, opened 2024-07-17
Update `pypi-attestations` to `0.0.9`

facutuesca:
The new attestation verification API now takes a `Distribution` object, containing the filename and the SHA256 digest of the artifact:
```python
class Distribution(BaseModel):
    name: str
    digest: str
```

as opposed to before, where it would take the artifact's `Path`, read the file and compute the digest.

This new API means that we can pass the already computed SHA256 digest directly (instead of having `pypi-attestations` re-compute it), and it also avoids `pypi-attestations` having to do I/O.

cc @di @woodruffw @DarkaMaul

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -62,7 +62,7 @@ rfc3986
 sentry-sdk
 setuptools
 sigstore~=3.0.0
-pypi-attestations==0.0.8
+pypi-attestations==0.0.9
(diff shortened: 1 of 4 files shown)


di (approved):

woodruffw (approved):
Thanks @facutuesca!

PR #16322, opened 2024-07-22
tests/functional: assert R_OK/W_OK for XDG dirs

woodruffw:
This follows https://github.com/pypi/warehouse/pull/16304 -- this test passes for local and CI runs because they don't use the `nobody` user, which then fails at runtime (in deployment) to access these dirs.

https://github.com/cabotage/cabotage-app/pull/85 fixes the above by moving the XDG dir creation to within Cabotage's wrapper; this updates the backstop test to be permission/access based, rather than existence based.

### DIFF
--- a/tests/functional/test_environment.py
+++ b/tests/functional/test_environment.py
@@ -10,6 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import platformdirs
 
 
@@ -20,3 +21,6 @@ def test_xdg_environment():
 
     assert user_data == "/tmp/share"
     assert user_cache == "/tmp/cache"
+
+    assert os.access(user_data, os.R_OK | os.W_OK)
+    assert os.access(user_cache, os.R_OK | os.W_OK)
--- a/tests/functional/test_environment.py
+++ b/tests/functional/test_environment.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import os
+
 import platformdirs
 
 
--- a/tests/functional/test_environment.py
+++ b/tests/functional/test_environment.py
@@ -19,8 +19,8 @@ def test_xdg_environment():
     # NOTE: We don't check the exact paths here, because they vary between
     # environments: they're in `$HOME` in local and CI builds, and in `/tmp`
     # when deployed.
-    user_data = platformdirs.user_data_dir()
-    user_cache = platformdirs.user_cache_dir()
+    user_data = platformdirs.user_data_dir(ensure_exists=True)
+    user_cache = platformdirs.user_cache_dir(ensure_exists=True)
 
     assert os.access(user_data, os.R_OK | os.W_OK)
     assert os.access(user_cache, os.R_OK | os.W_OK)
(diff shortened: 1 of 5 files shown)


tests/functional/test_environment.py
```diff
@@ -10,13 +10,17 @@
...
 
 def test_xdg_environment():
     # Backstop checks for Warehouse's (Dockerfile-configured) environment.
-    user_data = platformdirs.user_data_dir()
-    user_cache = platformdirs.user_cache_dir()
+    # NOTE: We don't check the exact paths here, because they vary between
+    # environments: they're in `$HOME` in local and CI builds, and in `/tmp`
+    # when deployed.
+    user_data = platformdirs.user_data_dir(ensure_exists=True)
+    user_cache = platformdirs.user_cache_dir(ensure_exists=True)
```
woodruffw, line 23: NB: With `ensure_exists` this is now partially a no-op test, at least in local/CI testing (since these dirs don't exist). However for Cabotage/deployment it'll end up testing something useful, per https://github.com/cabotage/cabotage-app/pull/85

ewdurbin: just for clarity, we do not execute tests in the cabotage environment. should this be a runtime check?

woodruffw: Ah, didn't realize that -- yeah, in that case it probably makes more sense as a runtime check. I'll look at putting it somewhere early in `python -m warehouse`'s initialization 🙂

woodruffw: I've put this in `warehouse/config.py`, as part of the early initialization that happens there. But I'm not 100% sure if that's the best place 🙂

ewdurbin: In smoke testing this, it looks like this doesn't correctly detect that things are working.

I rolled out a new release of test.pypi.org using this wrapper and you can see results in https://github.com/cabotage/cabotage-app/pull/85#issuecomment-2243361188

`platformdirs.user_data_dir(ensure_exists=True)` succeeds, but `platformdirs.user_data_dir('app', ensure_exists=True)` fails, indicating that this wouldn't actually ensure success for xdg/platformdirs use later on.

ewdurbin: appears this is due to the existing /tmp/{share, cache} files created by the Dockerfile on main :)

ewdurbin: validated this locally!

I still think it may be wise to change these to `platformdirs.user_cache_dir(secrets.token_urlsafe(), ensure_exists=True)` and similar to ensure not only that the dirs exist but that they're writable.

woodruffw: > appears this is due to the existing /tmp/{share, cache} files created by the Dockerfile on main :)

Whoops 🤦 -- I _think_ merging this first should fix that then, since we won't be doing `mkdir` in the Dockerfile anymore.

woodruffw: > I still think it may be wise to change these to `platformdirs.user_cache_dir(secrets.token_urlsafe(), ensure_exists=True)` and similar to ensure not only that the dirs exist but that they're writable.

Done!

ewdurbin (approved):

ewdurbin:
Note: this successfully deployed, so appears to be working end-to-end with the changes from cabotage-app.

woodruffw:
> Note: this successfully deployed, so appears to be working end-to-end with the changes from cabotage-app.

Awesome. I'll attempt to kick the tires again 🙂

PR #16323, opened 2024-07-22
Makefile: optimize subset test runs

woodruffw:
This is a small local DX improvement; please close if you think it's too hacky or undesirable 🙂 

TL;DR: When the user explicitly passes a subset of tests via `T=` and _doesn't_ pass `TESTARGS=`, this disables `pytest-xdist` by setting `TESTARGS=-n 0`. This results in significant speedups in practice, since worker spawning in `pytest-xdist` dominates execution time when there's only a small number of tests.

From local testing: running a single (trivial) test took ~16s before this, and takes ~180ms afterwards, for an approx. 100x speedup.


Makefile
```diff
@@ -7,6 +7,17 @@ ifeq ($(WAREHOUSE_IPYTHON_SHELL), 1)
     IPYTHON = yes
 endif
 
+# Optimization: if the user explicitly passes tests via `T`,
+# disable xdist (since the overhead of spawning workers is typically
+# higher than running a small handful of specific tests).
+# Only do this when the user doesn't set any explicit `TESTARGS` to avoid
+# confusion.
+ifneq ($(T),)
+ifeq ($(TESTARGS),)
```
woodruffw, line 16: NB: There are no boolean operators in Make (standard or GNU), so this kind of ugly nesting is the only way to express "T is not empty and TESTARGS is empty".

firefly9966 (approved):

di (approved):

miketheman (approved):
Looks good to me, and works as advertised.

Should there be a callout to this somewhere here?
https://github.com/pypi/warehouse/blob/4a63ffc0c46db3592f9a034d6201754ccc108b45/docs/dev/development/getting-started.rst?plain=1#L684-L697

Also - in a perfect world, I'd want any `T=` scope to disable coverage reporting/failures since it's almost guaranteed to fail and adds test time overhead, but that's a different problem.

woodruffw:
> Should there be a callout to this somewhere here?

Added!

miketheman (approved):

woodruffw:
(Fixed the lint.)

miketheman (approved):

PR #16339, opened 2024-07-25
Add CORS headers to HTTP exceptions

### LINKED ISSUE #14229: Next.js fetch package CORS error
**Describe the bug**

Calling `GET https://pypi.org/pypi/asdfasdfasdfasdfasfas/json`
results in 
```
https://pypi.org/pypi/asdfasdfasdfasdfasfas/json
CORS Missing Allow Origin

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://pypi.org/pypi/asdfasdfasdfasdfasfas/json. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.
```

**Expected behavior**

I expect to get a 404, but not a CORS error.

**To Reproduce**

1. Create a nextjs project from this template https://vercel.com/new/clone?s=https%3A%2F%2Fgithub.com%2Fvercel%2Fvercel%2Ftree%2Fmain%2Fexamples%2Fnextjs&template=nextjs&id=67753070&b=main&showOptionalTeamCreation=false
2. Add and run this code:
```
      let resp = await fetch(`https://pypi.org/pypi/asdfasdfasdfasdfasfas/json`, {
        headers: {
          Host: "pypi.org",
          Accept: "application/json",
        },
      });
```

**My Platform**

**Additional context**

I have no issues when pypi returns 200 for a found package, it's only when it returns 404 for a non-existing package. Also I have no issues hitting pypi from the command line, only with nextjs/fetch.

ryanking13:
Resolve #14229

This adds CORS headers to the response when warehouse returns404, 410, 50x responses.

I added the same CORS headers to simple APIs in https://github.com/pypi/warehouse/pull/13222, but they were not applied on an error case.

__Context__

In [pyodide](https://github.com/pyodide/pyodide), we are implementing a fallback option for package index, so that if a package is not available in one package index, then search it in another package index. However, without the CORS headers, it is impossible to distinguish the 404 Not Found error.

cc: @Carreau


tests/unit/legacy/api/test_json.py
```diff
@@ -30,14 +31,9 @@
...
-    assert headers["Access-Control-Allow-Headers"] == (
-        "Content-Type, If-Match, If-Modified-Since, If-None-Match, "
-        "If-Unmodified-Since"
-    )
-    assert headers["Access-Control-Allow-Methods"] == "GET"
-    assert headers["Access-Control-Max-Age"] == "86400"
-    assert headers["Access-Control-Expose-Headers"] == "X-PyPI-Last-Serial"
+    for k, v in _CORS_HEADERS.items():
+        assert k in headers
+        assert headers[k] == v
```
Carreau, line 36: I would guess the reason for hardcoding is that it is easy to modify `warehouse.utils.cors:_CORS_HEADERS` and have the test still passing.

I'm happy with this change, but maybe the core devs will prefer to make sure the test comparison does not rely on data in the main codebase.

warehouse/views.py
```diff
@@ -94,12 +95,17 @@ def httpexception_view(exc, request):
     try:
         # Lightweight version of 404 page for `/simple/`
         if isinstance(exc, HTTPNotFound) and request.path.startswith("/simple/"):
-            response = HTTPNotFound(body="404 Not Found", content_type="text/plain")
+            response = HTTPNotFound(
+                body="404 Not Found",
+                content_type="text/plain",
+                headers=_CORS_HEADERS,
+            )
```
Carreau, line 102: 👍

Carreau (approved):
Thanks for looking into this, that is great !

I haven't tested locally yet though.

ryanking13:
@di Can I ask for your review when you have time?

warehouse/views.py
```diff
@@ -118,6 +124,7 @@ def httpexception_view(exc, request):
     response.headers.extend(
         (k, v) for k, v in exc.headers.items() if k not in response.headers
     )
+    response.headers.extend(_CORS_HEADERS)
```
miketheman, line 127: **question:** With this addition, `_CORS_HEADERS` should be applied to any `response` object in this function. ‏Are the other additions of `headers=_CORS_HEADERS` even necessary?

ryanking13: Indeed. Thanks for the catch. I removed redundant `headers=_CORS_HEADERS`.

miketheman (approved):

Carreau:
Thanks !

PR #16364, opened 2024-07-30
Add metrics for GH Trusted Publishers with reusable workflows

facutuesca:
While looking into adding support for reusable workflows in GitHub Trusted Publishers (https://github.com/pypi/warehouse/issues/11096), I found that we (accidentally) already support a subset of reusable workflow scenarios.
If the reusable workflow is located in the same repository as the top-level workflow, then the OIDC claims will look something like this:
```json
{
    "job_workflow_ref": "org/repo/.github/workflows/reusable.yml@sha",
    "workflow_ref": "org/repo/.github/workflows/top-level.yml@sha",
}
```

That is, the `job_workflow_ref` claim will refer to the reusable workflow file, and `workflow_ref` to the top level workflow file, both in the same repository.

Since we check the Trusted Publisher's workflow filename against the `job_workflow_ref` claim, if the user set up the Trusted Publisher with `reusable.yml` in the workflow filename field, then verification will pass and the Trusted Publisher will work. This is the only scenario where reusable workflows currently work.

The problem is that we should be checking against `workflow_ref` instead. The other fields we check against (repo owner and repo name) refer to the repository that contains the top-level workflow, so the filename field should also refer to the top-level workflow.

One way of fixing this is to change the checks from using `job_workflow_ref` to using `workflow_ref`. This is fine for non-reusable workflows, since there those claims are equal (they contain the same path). However, there might be users who depend on the incorrect behavior mentioned above. They might have configured their Trusted Publisher to trust `reusable.yml`, and changing the check to `workflow_ref` (which contains `top-level.yml`) would break their Publisher.

In order to understand how many users depend on this behavior, this PR adds a counter for every time a user mints an API token using a GitHub Trusted Publisher with a reusable workflow. This way we can understand the impact of changing the behavior, and if it makes sense to do so.

cc @woodruffw @di

### DIFF
--- a/warehouse/oidc/views.py
+++ b/warehouse/oidc/views.py
@@ -30,7 +30,7 @@ from warehouse.macaroons.services import DatabaseMacaroonService
 from warehouse.metrics.interfaces import IMetricsService
 from warehouse.oidc.errors import InvalidPublisherError
 from warehouse.oidc.interfaces import IOIDCPublisherService
-from warehouse.oidc.models import OIDCPublisher, PendingOIDCPublisher
+from warehouse.oidc.models import GitHubPublisher, OIDCPublisher, PendingOIDCPublisher
 from warehouse.oidc.services import OIDCPublisherService
 from warehouse.oidc.utils import OIDC_ISSUER_ADMIN_FLAGS, OIDC_ISSUER_SERVICE_NAMES
 from warehouse.packaging.interfaces import IProjectService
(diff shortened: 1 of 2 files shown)


warehouse/oidc/views.py
```diff
@@ -284,4 +284,17 @@ def mint_token(
                 "publisher_url": publisher.publisher_url(),
             },
         )
+
+    if isinstance(publisher, GitHubPublisher) and claims:
```
woodruffw, line 288: Might be worth adding a `NOTE` or similar comment here explaining that this is for temporary metrics collection 🙂

woodruffw (approved):
This approach makes sense to me, thanks @facutuesca! Left one nitpick about adding a comment explaining the metric being added (and that + why it's temporary)

warehouse/oidc/views.py
facutuesca: Good point, fixed!

woodruffw: Perfect!

woodruffw (approved):

di (approved):

PR #16365, opened 2024-07-30
chore: enforce 2fa for legacy uploaders

miketheman:
Disallow uploads if the token from a user account does not have 2FA.

- Add `UserFactory` trait to simplify email creation in test cases
- Update test coverage for `repeat_window` to exercise logic independent
  of email-specific test cases
- Remove email template


miketheman:
Here's a chart showing the dropoff over time.

Screenshot 2024-07-30 at 16 12 15

warehouse/forklift/legacy.py
```diff
@@ -519,6 +516,19 @@ def file_upload(request):
                     project_help=request.help_url(_anchor="verified-email"),
                 ),
             ) from None
+        # Ensure user has enabled 2FA before they can upload a file.
+        if not request.user.has_two_factor:
```
woodruffw, line 520: I don't think this can happen in this case because of our restrictions on Trusted Publishing, but just calling it out: `request.user` can be `None` if the uploader is a Trusted Publisher instead of a user-controlled API token.

woodruffw (approved):
Exciting!

warehouse/forklift/legacy.py
miketheman: Thanks! I think this is already guarded by `if request.user:` above, isn't it? https://github.com/pypi/warehouse/pull/16365/files#diff-bffd49e4f943217b54b4cfcf5835737bec7a145360bd1dc2ec1541d2dcafad16L501-L503

woodruffw: Ah yep, indeed -- I missed that in the diff context 😅

dstufft (approved):

PR #16376, opened 2024-08-01
Ignore disconnection errors

di:
We get disconnection exceptions when the client disconnects quite frequently (https://python-software-foundation.sentry.io/issues/3104967674/).

There's not much we can do about this (the exception doesn't seem to be configurable/preventable) and we can't prevent clients from disconnecting, so just ignore this error.

Also removes an outdated comment (the `ignore_errors` option was never added to the latest Sentry SDK).


dstufft (approved):

warehouse/sentry.py
```diff
@@ -52,6 +48,8 @@ def _sentry(request):
     "gunicorn.http.errors.InvalidProxyLine",
     "gunicorn.http.errors.ForbiddenProxyRequest",
     "gunicorn.http.errors.InvalidSchemeHeaders",
+    # Webob raises this when the client disconnects, which we can't do anything about
+    "webob.request.DisconnectionError",
```
dstufft, line 52: It might make sense to import this instead of using a string? I'm fine either way, but the above comment indicates the `gunicorn` errors were only string-ified so we didn't import gunicorn.

miketheman (approved):
LGTM - same comment as dstufft

PR #16377, opened 2024-08-01
fix: retry hcaptcha on timeouts

miketheman:
Both `ConnectTimeout` and `ReadTimeout` are retryable errors (and are subclasses of `requests.Timeout`).

Raising a `RetryableException` should trigger a retry up to the max retry attempts (3, already pre-configured).

Fixes WAREHOUSE-PRODUCTION-1YY


di (approved):
Nice!

PR #16384, opened 2024-08-02
Disable fsync in test postgres

twm:
This shaves ~15 seconds (6%) from the runtime of the test suite on my machine, which has a slowish SATA SSD (~1:30 → ~1:15).

Extracted from #16295.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,7 @@ jobs:
           - 5432:5432
         env:
           POSTGRES_HOST_AUTH_METHOD: trust  # never do this in production!
+          POSTGRES_INITDB_ARGS: '--no-sync --set fsync=off --set full_page_writes=off'
         # Set health checks to wait until postgres has started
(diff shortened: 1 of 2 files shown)


miketheman (approved):
Tested on macOS - no material improvement observed, but no harm either.

Should the same be applied for CI, where the runtime is also Linux?
https://github.com/pypi/warehouse/blob/a1f916bc65c57de45cb1dd72f31d0508ac7afed9/.github/workflows/ci.yml#L70-L75

twm:
> Should the same be applied for CI, where the runtime is also Linux?

Absolutely &mdash; I didn't realize that was a separate config. I'll make that change now.

twm:
It doesn't make an obviously big difference in CI &mdash; [3m 8s](https://github.com/pypi/warehouse/actions/runs/10221275243/job/28283571801) to [3m 4s](https://github.com/pypi/warehouse/actions/runs/10221739393/job/28285050824) in the runs here &mdash; but doesn't seem to harm anything, either.

miketheman:
Excellent, thanks. Will merge on Monday.

PR #16385, opened 2024-08-02
test: use double size runners

miketheman:
Since the cost is linear to CPU (2x), jobs should cost the same but finish faster.

Refs: https://depot.dev/docs/github-actions/runner-types#arm-runners

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   build:
     if: github.repository == 'pypi/warehouse'
-    runs-on: depot-ubuntu-22.04-arm
+    runs-on: depot-ubuntu-22.04-arm-4
     outputs:
       buildId: ${{ steps.build.outputs.build-id}}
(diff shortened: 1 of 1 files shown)


miketheman:
Using arm-4 on a cold cache:

Screenshot 2024-08-02 at 17 10 00

miketheman:
Using arm-8 with cold cache:

image

miketheman:
Conclusion:

- build doesn't appear to benefit from a larger runner, so let's not do that 💸 
- Tests does speed up as there are more CPUs to schedule against, but the improvement isn't as linear as I'd hoped.

Tests on `-arm` (2 CPU) generally take ~4 minutes. With 4 CPUs = 3m 9s, and 8 CPUs = 2m 45s

4x the cost might not be worth the less-than-2x-speedup.

I'm happy I checked - because [GH Actions increased their defaults to 4 CPU and 16GB RAM](https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/), but don't have ARM runners yet.

PR #16388, opened 2024-08-03
Fixed #16190 -- Added missing JSON API reference information on metad…

### LINKED ISSUE #16190: JSON API Reference is missing information on metadata 2.2 and 2.3 fields
**Describe the bug**

This JSON API reference: https://github.com/pypi/warehouse/blob/main/docs/dev/api-reference/json.rst

Is missing documentation on the "dynamic" and "provides_extra" fields, added in versions 2.2 and 2.3 of the core metadata specification (https://packaging.python.org/en/latest/specifications/core-metadata/).

The example output shown in the reference is missing them entirely, yet they can currently be seen if you query the JSON API.

**Expected behavior**
API reference should include these fields

samjove:
…ata 2.2 and 2.3 fields

Fixes #16190.


miketheman (dismissed):
Thanks for starting on this!

Did you see this comment? https://github.com/pypi/warehouse/issues/16190#issuecomment-2211368970

Please provide some example of what these might be, other than a `null`

samjove:
Thanks for the feedback, I've made some changes to it. Please review when you get a chance

di (approved):

PR #16399, opened 2024-08-05
Support pre-filling Trusted Publisher form via URL params

facutuesca:
## What is this?
This PR adds support for pre-filling the Trusted Publisher form via URL parameters. That is, opening the following URL (while logged in):

```https://pypi.org/manage/project/my_project/settings/publishing/?provider=github&owner=my_user&repository=my_project&workflow_filename=my_workflow.yml```

will open the page to create a new Trusted Publisher for `my_project` with the fields pre-filled:

image

This fixes https://github.com/pypi/warehouse/issues/13661 (see the last comment)

## Design & implementation considerations
- This is only implemented for the Project view, since the usecase is generating magic links to create a Trusted Publisher for an existing project. The concrete usecase is for the GH action `gh-action-pypi-publish` to generate a link and show it to the user whenever it's used to publish using an API token, encouraging the user to create a Trusted Publisher for that (existing) project by clicking on the magic link.
- The only required parameter to enable pre-filling is `provider`. Any provider can be specified (`github`, `gitlab`, `google` and `activestate`) and they are case-insensitive.
- The name of the parameters (`owner`, `email`, etc) are the same as the name of the form fields. This allows us to keep the implementation simple by using the URL parameters to fill the form directly without an intermediary mapping between URL param names and form field names.
- The URL parameters do not need to have all of the form's fields. Missing fields will just be left empty.
- URL parameters that do not match any form fields will be ignored.
- If an unknown provider is specified, nothing is done (all fields will be empty, like the default view).
- The tab for the pre-filled provider will be selected on page load.
- A warning will be shown when using this feature, asking the user to double-check the values since they have been pre-filled (see the screenshot above).

cc @woodruffw @webknjaz  @di

### DIFF
--- a/warehouse/templates/manage/project/publishing.html
+++ b/warehouse/templates/manage/project/publishing.html
@@ -28,6 +28,7 @@
 {% endmacro %}
 
 {% macro github_form(request, github_publisher_form) %}
+    {{ prefilled_warning() if prefilled_provider == "github" }}
     <p>
       {% trans href="https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect" %}
       Read more about GitHub Actions' OpenID Connect support <a href="{{ href }}">here</a>.
@@ -125,7 +126,6 @@
           {{ field_errors(github_publisher_form.environment) }}
         </div>
       </div>
-      {{ prefilled_warning() if prefilled_provider == "github" }}
       <div>
         <input type="submit" value="{% trans %}Add{% endtrans %}" class="button button--primary">
       </div>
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: warehouse/templates/manage/project/publishing.html [hunk 1]
```diff
@@ -118,6 +125,7 @@
           {{ field_errors(github_publisher_form.environment) }}
         </div>
       </div>
+      {{ prefilled_warning() if prefilled_provider == "github" }}
```
woodruffw, line 128: Nitpick: IMO putting this at the top of the form instead of the bottom might catch the user's attention slightly better/flow a little better 🙂

woodruffw: (Same for the others)

facutuesca: Fixed! I also updated the screenshot

di (approved):
LGTM. This is blocked on https://github.com/pypi/infra/pull/169.

PR #16413, opened 2024-08-06
feat: append task name to SQS Message

miketheman:
Add metadata to the message of what the task name generated the message.


di (approved):
Looks correct to me based on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs/message/message_attributes.html#message-attributes

PR #16429, opened 2024-08-07
chore: add tracing to worker

miketheman:
Instead of previous attempts where we applied tracing to all workers, create an individual service that we can control how many workers run to build trust tracing.


miketheman:
Per https://formulae.brew.sh/formula/oxker

ewdurbin (approved):

PR #16450, opened 2024-08-09
fix: use output from packaging.metadata parsed

miketheman:
Now that we use `packaging.metadata` to pre-parse the inbound metadata, we no longer have to split a string to keywords, as that's being done for us already.

Fixes WAREHOUSE-PRODUCTION-21A
Fixes WAREHOUSE-PRODUCTION-21B

### DIFF
--- a/warehouse/utils/release.py
+++ b/warehouse/utils/release.py
@@ -13,7 +13,7 @@
 
 def strip_keywords(keywords: list[str]) -> list[str]:
     """
-    Strip whitespace, remove empties.
+    Remove empties.
     Useful to cleanse user input prior to storing in Release.keywords_array.
     """
-    return [stripped for keyword in keywords if (stripped := keyword.strip())]
+    return [keyword for keyword in keywords if keyword]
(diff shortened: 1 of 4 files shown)


warehouse/utils/release.py
```diff
@@ -11,13 +11,9 @@
...
-    Split keywords on commas and strip whitespace, remove empties.
+    Strip whitespace, remove empties.
     Useful to cleanse user input prior to storing in Release.keywords_array.
     """
-    return [
-        stripped
-        for keyword in keyword_input.split(",")
-        if (stripped := keyword.strip())
-    ]
+    return [stripped for keyword in keywords if (stripped := keyword.strip())]
```
di, line 19: Looks like these are already stripped: https://github.com/pypa/packaging/blob/2c885fe91a54559e2382902dce28428ad2887be5/src/packaging/metadata.py#L167-L169, do we just need to check for empty strings?

di (approved):

warehouse/utils/release.py
miketheman: Done!

PR #16461, opened 2024-08-12
Remove sentry message on password token use after login

di:
We're already returning an error here, so there's no real use in continuing to emit Sentry messages when this happens.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -17,7 +17,6 @@ import uuid
 
 import humanize
 import pytz
-import sentry_sdk
 
 from first import first
 from pyramid.httpexceptions import (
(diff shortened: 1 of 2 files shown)


miketheman (approved):
Generally fine. This originated back in 2018 from here:
https://github.com/pypi/warehouse/pull/2764/files#diff-4d7e23e1c19fbedae06550a287ef846d8f8efe27ee3c54f1a7ec3f4f8b9c5af6R224-R228

The implementation of the TODO was last year in [`dabd388` (#14563)](https://github.com/pypi/warehouse/pull/14563/commits/dabd388ab2babb099a3de3d53f008f81696da0eb)

Would we prefer a metric tracking the occurrence so as to determine if something else is causing an increase in these error conditions? Or does another path emit this to a metric/trace somewhere?

di:
I think we decided it's happening frequently enough that it's not a very meaningful signal, and since we're already preventing the reset token from being used after a login, there isn't really much else to do with this information.

PR #16462, opened 2024-08-12
GitHub, GitLab: improve claim matching during lookup

woodruffw:
This fixes a constraint violation that users were previously able to induce by registering two similar-looking-but-not-exact Trusted Publishers. For example:

```
repo: foo/bar
workflow: publish-pypi.yml
```

and:

```
repo: foo/bar
workflow: publish_pypi.yml
```

Before this change, PyPI's publisher lookup would treat the `_` in `publish_pypi.yml` as a match expression, resulting in both publishers being matched when only exactly one publisher was expected to match. As a result, a user could configure two publishers in such a way that minting would *always* fail.

~~To fix this, we replace our use of a raw `like` with a more constrained `startswith` *and* escape `_` and `%` when they appear in the workflow identifier being matched on. We would ideally do this using SQLAlchemy's `autoescape`, but `autoescape` only supports string clauses and not column clauses. Instead, we build up a somewhat verbose chain of escape replacements manually.~~

Update: this PR now fixes the above by removing the SQL-level string matching entirely, and instead using regular expressions to extract the relevant claim component. This is both faster and much easier to test than the previous approach, as well as less brittle.

Fixes https://python-software-foundation.sentry.io/issues/5695739612

### DIFF
--- a/tests/unit/oidc/models/test_github.py
+++ b/tests/unit/oidc/models/test_github.py
@@ -26,6 +26,7 @@ from warehouse.oidc.models import _core, github
         # Well-formed workflow refs, including exceedingly obnoxious ones
         # with `@` or extra suffixes or `git` refs that look like workflows.
         ("foo/bar/.github/workflows/basic.yml@refs/heads/main", "basic.yml"),
+        ("foo/bar/.github/workflows/basic.yaml@refs/heads/main", "basic.yaml"),
         ("foo/bar/.github/workflows/has-dash.yml@refs/heads/main", "has-dash.yml"),
         (
             "foo/bar/.github/workflows/has--dashes.yml@refs/heads/main",
@@ -69,7 +70,7 @@ from warehouse.oidc.models import _core, github
             "basic.yml",
         ),
         (
-            "foo/bar/.github/workflows/basic.yml@refs/heads/misleading@branch@causestwomatches.yml",
+            "foo/bar/.github/workflows/basic.yml@refs/heads/bad@branch@twomatches.yml",
             "basic.yml",
         ),
         ("foo/bar/.github/workflows/foo.yml.yml@refs/heads/main", "foo.yml.yml"),
@@ -78,9 +79,11 @@ from warehouse.oidc.models import _core, github
             "foo.yml.foo.yml",
         ),
         # Malformed workflow refs.
+        ("foo/bar/.github/workflows/basic.wrongsuffix@refs/heads/main", None),
         ("foo/bar/.github/workflows/@refs/heads/main", None),
         ("foo/bar/.github/workflows/nosuffix@refs/heads/main", None),
         ("foo/bar/.github/workflows/.yml@refs/heads/main", None),
+        ("foo/bar/.github/workflows/.yaml@refs/heads/main", None),
         ("foo/bar/.github/workflows/main.yml", None),
     ],
 )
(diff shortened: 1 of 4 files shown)


warehouse/oidc/models/github.py
```diff
@@ -189,7 +189,26 @@ def __lookup_all__(klass, signed_claims: SignedClaims) -> Query | None:
...
+                    func.replace(
+                        func.replace(
+                            func.replace(klass.workflow_filename, "\\", "\\\\"),
+                            "%",
+                            "\\%",
+                        ),
+                        "_",
+                        "\\_",
+                    ),
+                    escape="\\",
```
woodruffw, line 210: NB: I could abstract this out into a helper, since it ends up being repeated in 4 different places.

warehouse/oidc/models/github.py
```diff
@@ -189,7 +189,26 @@ def __lookup_all__(klass, signed_claims: SignedClaims) -> Query | None:
                 environment=environment.lower(),
             )
             .filter(
-                literal(workflow_ref).like(func.concat(klass.workflow_filename, "%"))
+                literal(workflow_ref).startswith(
```
facutuesca, line 192: Since we're switching the use of `like(concat(name, %))` with `.startswith(name)`, we could go one step further and fix the issue we discussed where `filename.yml.whatever.yml` will match the TP for `filename.yml` (due to the `%` at the end).

 We could do `.startswith(name + '@')`, which would fix the issue for all cases except for the case where another `@` is used, like `filename.yml@whatever.yml`.

woodruffw: Yeah, I was thinking a bit about that -- I think we can address both of those cases at the same time by imitating what we do for the `_check_job_workflow_ref` claim validator, i.e. *reconstruct* the parts of the `workflow_ref` rather than trying to disambiguate them. The main downside to that is that it means we'll have come full circle, and will essentially be doing claim validation twice (once at the DB level, once at the `verify_claims` level).

(The other alternative here would be to have GitHub fix this at the claims level, e.g. by adding a new `workflow_filename` or similar claim. I've pinged some folks to see if that's something they can do.)

woodruffw: (And NB: the other benefit to a new `workflow_filename` claim for this is that we'll no longer need to do a string search across publishers -- that search isn't a huge performance hit at the moment since it only happens after JWT signature verification, but removing it would be nice!)

warehouse/oidc/models/gitlab.py
```diff
@@ -177,7 +177,20 @@ def __lookup_all__(klass, signed_claims: SignedClaims) -> Query | None:
...
+                        func.replace(
+                            func.replace(klass.workflow_filepath, "\\", "\\\\"),
+                            "%",
+                            "\\%",
+                        ),
+                        "_",
+                        "\\_",
+                    ),
+                    escape="\\",
+                )
```
di, line 193: Can we do the same here?

tests/unit/oidc/models/test_github.py
```diff
@@ -20,19 +20,67 @@
...
 
 
-def test_lookup_strategies():
-    assert (
-        len(github.GitHubPublisher.__lookup_strategies__)
-        == len(github.PendingGitHubPublisher.__lookup_strategies__)
-        == 2
-    )
-
-
```
di, line 35: Intentional deletion?

woodruffw: Yep, this was duplicated with the class-scoped test immediately below 🙂:

https://github.com/pypi/warehouse/pull/16462/files#diff-cc2dfefbb24e9b4bb64ca653fe64240ed824fda8f40e7ef0f233c55e94ad138fR85

warehouse/oidc/models/gitlab.py
woodruffw: Yep! The regexp is slightly different thanks to GitLab's support for path-qualified workflows, but I'll customize it.

di (approved):

PR #16464, opened 2024-08-12
fix(admin): disable account recovery if user frozen

miketheman:
Also fixes setting the `disabled` property as a class on `a` which is different than a property on a `button`.


miketheman:
https://github.com/user-attachments/assets/9302e42a-b460-4ebd-b190-0712a5569118

di (approved):

ewdurbin (changes requested):
Can we disable admin password reset button and wipe factors as well?

di (changes requested):
Just seeing if branch protection works with requested changes

di (approved):

PR #16472, opened 2024-08-13
Move verified Release URLs to the Verified section

facutuesca:
This moves a release's project URLs that have been verified via Trusted Publishing to the "Verified details" section of the release page:
image

The backend side of this (marking URLs as verified or not depending on the Trusted Publisher) was merged [here](https://github.com/pypi/warehouse/pull/16205). This fixes https://github.com/pypi/warehouse/issues/8635.

@woodruffw @di

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -652,6 +652,47 @@ class TestRelease:
         # TODO: It'd be nice to test for the actual ordering here.
         assert dict(release.urls) == dict(expected)
 
+    @pytest.mark.parametrize(
+        "release_urls",
+        [
+            [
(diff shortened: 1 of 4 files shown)


di (approved):

woodruffw:
Nice work @facutuesca!

di:
Here's an example of it in action: https://pypi.org/project/scikit-build-core/

Thinking now we should probably move the links up above the list of maintainers, which is where they used to be. @facutuesca can you fix that?

facutuesca:
> Here's an example of it in action: https://pypi.org/project/scikit-build-core/
> 
> Thinking now we should probably move the links up above the list of maintainers, which is where they used to be. @facutuesca can you fix that?

@di on it

PR #16483, opened 2024-08-14
feat(dev): alembic check

miketheman:
Implements wrapper for `check` command.

Refs: https://alembic.sqlalchemy.org/en/latest/autogenerate.html#running-alembic-check-to-test-for-new-upgrade-operations
Refs: https://alembic.sqlalchemy.org/en/latest/api/commands.html#alembic.command.check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,3 +93,41 @@ jobs:
           key: ${{ runner.os }}-mypy-${{ env.pythonLocation }}-${{ hashFiles('requirements.txt', 'requirements/*.txt') }}
       - name: Run ${{ matrix.name }}
         run: ${{ matrix.command }}
+
+  check_db:
+    name: Check Database Consistency
+    needs: build
+    runs-on: depot-ubuntu-22.04-arm
+    continue-on-error: true
+    container:
+      image: registry.depot.dev/rltf7cln5v:${{ needs.build.outputs.buildId }}
+      credentials:
+        username: x-token
+        password: ${{ needs.build.outputs.token }}
+    services:
+      postgres:
+        image: postgres:16.1
+        ports:
+        - 5432:5432
+        env:
+          POSTGRES_DB: warehouse
+          POSTGRES_HOST_AUTH_METHOD: trust  # never do this in production!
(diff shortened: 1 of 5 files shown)


miketheman:
Note: this check should **not be marked as required**, since there are cases where we want to allow for a code-side removal (which would generate a migration) and merge that PR, and ship the database migration after the code no longer relies on it.

miketheman:
You can see a failing run here, where I introduce a code change that has no associated migration:
https://github.com/pypi/warehouse/actions/runs/10394102434/job/28783297165?pr=16483

.github/workflows/ci.yml
```diff
@@ -93,3 +93,41 @@ jobs:
...
+        uses: falti/dotenv-action@v1
+        with:
+          path: dev/environment
+          export-variables: true
+          keys-case: upper
+      - name: Check Database
+        run: bin/db-check
+        env:
+          # override the hostname set in `dev/environment`
+          DATABASE_URL: 'postgresql+psycopg://postgres@postgres/warehouse'
```
di, line 133:
```suggestion
      - name: Check for 'partial-migration' label
        id: check_label
        run: |
          if [[ "$(jq '.pull_request.labels[].name' <<<"$GITHUB_EVENT_PATH")" == *'"partial-migration"'* ]]; then
            echo "Label 'partial-migration' found, skipping database consistency check."
            exit 0
          fi
        continue-on-error: true

      - name: Check Database
        if: steps.check_label.outcome == 'success'
        run: bin/db-check
        env:
          DATABASE_URL: 'postgresql+psycopg://postgres@postgres/warehouse'
```

di, line 133:
```suggestion
      - name: Check Database
        if: "! contains(github.event.pull_request.labels.*.name, 'partial-migration')"
        run: bin/db-check
        env:
          # override the hostname set in `dev/environment`
          DATABASE_URL: 'postgresql+psycopg://postgres@postgres/warehouse'
```
Might be possible to just do this instead, I'm not sure what the overall outcome of the job will be if the last check is skipped like this though.

di (approved):

Issue #16474, opened 2024-08-13
Project metadata: direct PyPI URL is not considered verified


woodruffw:
This is a tiny thing I noticed while testing out @facutuesca's changes: if a project lists its own PyPI URL via `project.urls`, we don't currently consider it verified.

For example, `abi3info`: https://pypi.org/project/abi3info/2024.8.13/

<img width="332" alt="Screenshot 2024-08-13 at 3 12 07 PM" src="https://github.com/user-attachments/assets/e41252d1-0928-4d3b-a505-398e79de5e20">

In this case the `Homepage` is marked as unverified, despite being `https://pypi.org/project/abi3info/`. 

Here's the metadata itself, as it appears in `abi3info`'s `pyproject.toml`:

```toml
[project.urls]
Homepage = "https://pypi.org/project/abi3info/"
Issues = "https://github.com/woodruffw/abi3info/issues"
Source = "https://github.com/woodruffw/abi3info"
Documentation = "https://woodruffw.github.io/abi3info/"
```

## Expected behavior

I think we should consider these "trivially" verified, i.e. if project `$foo` has a URL that equals exactly `https://pypi.org/project/$foo/` (or `https://pypi.org/p/$foo/`), then we should consider that URL verified.

(There might be a minor hiccup around normalizing `$foo` for comparison, since I believe the `/p/` and `/project/` routes are tolerant of non-normalized project names).

CC @facutuesca @di for thoughts 🙂 

xref https://github.com/pypi/warehouse/pull/16205

di:
Makes sense to me! This should probably support a couple different equivalent URLs:

* https://pypi.org/project/abi3info/
* https://pypi.org/p/abi3info/
* https://pypi.python.org/project/abi3info/
* https://pypi.python.org/p/abi3info/
* https://python.org/pypi/abi3info/

Including with/without trailing slashes too.

(It should also be trivial to verify the `github.io` URL as well I think!)

facutuesca:
Also makes sense to me, looking into the implementation now

facutuesca:
https://github.com/pypi/warehouse/pull/16485 implements the verification for links to PyPI's project page. I'll do the `github.io` URL in a separate PR

PR #16485, opened 2024-08-14
Verify URLs that link to the project page on PyPI

facutuesca:
Fixes https://github.com/pypi/warehouse/issues/16474

Verifies URLs that link to the project's page on PyPI. URLs are normalized using `rfc3986`, and trailing slashes (e.g: `https://pypi.org/p/my_project/`) can be present or not.

image

(Homepage links to `https://pypi.org/project/$PROJECT_NAME`)

cc @di @woodruffw

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -520,18 +520,19 @@ def _verify_url_with_trusted_publisher(url: str, publisher_url: str) -> bool:
 def _verify_url(
     url: str, publisher_url: str | None, project_name: str, project_normalized_name: str
 ) -> bool:
-    is_verified_pypi = _verify_url_pypi(
+    if _verify_url_pypi(
         url=url,
         project_name=project_name,
         project_normalized_name=project_normalized_name,
-    )
+    ):
+        return True
+        
+    if not publisher_url:
+        return False
 
-    if publisher_url:
-        return is_verified_pypi or _verify_url_with_trusted_publisher(
-            url=url, publisher_url=publisher_url
-        )
-    else:
-        return is_verified_pypi
+    return _verify_url_with_trusted_publisher(
+        url=url, publisher_url=publisher_url
+    )
 
 
 def _sort_releases(request: Request, project: Project):
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -481,7 +481,7 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str) 
     )
 
 
-def _verify_url_with_trusted_publisher(url: str, publisher_url: str | None) -> bool:
+def _verify_url_with_trusted_publisher(url: str, publisher_url: str) -> bool:
     """
     Verify a given URL against a Trusted Publisher URL
 
@@ -497,9 +497,6 @@ def _verify_url_with_trusted_publisher(url: str, publisher_url: str | None) -> b
     the authority includes the host, and in practice neither URL should have user
     nor port information.
     """
-    if not publisher_url:
-        return False
-
     publisher_uri = rfc3986.api.uri_reference(publisher_url).normalize()
     user_uri = rfc3986.api.uri_reference(url).normalize()
     if publisher_uri.path is None:
@@ -518,6 +515,23 @@ def _verify_url_with_trusted_publisher(url: str, publisher_url: str | None) -> b
         and publisher_uri.authority == user_uri.authority
         and is_subpath
     )
+
+
+def _verify_url(
+    url: str, publisher_url: str | None, project_name: str, project_normalized_name: str
+) -> bool:
+    is_verified_pypi = _verify_url_pypi(
+        url=url,
+        project_name=project_name,
+        project_normalized_name=project_normalized_name,
+    )
+
+    if publisher_url:
+        return is_verified_pypi or _verify_url_with_trusted_publisher(
+            url=url, publisher_url=publisher_url
+        )
+    else:
+        return is_verified_pypi
 
 
 def _sort_releases(request: Request, project: Project):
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -866,7 +890,15 @@ def file_upload(request):
...
-                "verified": _verify_url(url=url, publisher_url=publisher_base_url),
+                "verified": _verify_url_with_trusted_publisher(
+                    url=url,
+                    publisher_url=publisher_base_url,
+                )
+                or _verify_url_pypi(
+                    url=url,
+                    project_name=project.name,
+                    project_normalized_name=project.normalized_name,
+                ),
```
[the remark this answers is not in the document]
facutuesca: I added the helper function. As for refactoring to a separate module/file, I agree, particularly since we'll be adding even more verification options in the future

warehouse/forklift/legacy.py
```diff
@@ -496,6 +517,23 @@ def _verify_url(url: str, publisher_url: str | None) -> bool:
...
+        project_name=project_name,
+        project_normalized_name=project_normalized_name,
+    )
+
+    if publisher_url:
+        return is_verified_pypi or _verify_url_with_trusted_publisher(
+            url=url, publisher_url=publisher_url
+        )
+    else:
+        return is_verified_pypi
```
woodruffw, line 534: Short-circuit, maybe?

(Not sure if this is actually better, but it avoids a few unnecessary fall-throughs 🙂)

```suggestion
    if _verify_url_pypi(
        url=url,
        project_name=project_name,
        project_normalized_name=project_normalized_name,
    ):
        return True
        
    if not publisher_url:
        return False

    return _verify_url_with_trusted_publisher(
        url=url, publisher_url=publisher_url
    )
```

woodruffw (approved):
LGTM, thanks @facutuesca!

warehouse/forklift/legacy.py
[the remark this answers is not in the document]
facutuesca: @woodruffw while adding support for the `github.io` URLs, I realised that it might make sense to move the current `_verify_url_with_trusted_publisher` to the `OIDCPublisherMixin` class as a method (since we're verifying the URL against the publisher's base URL, which lives in that class)

Then, for provider-specific verification rules we want to add (like GitHub's `user.github.io/repo`) , we can override that method in the GitHub subclass to add that logic. WDYT?

For illustration, I mean something like this:
```python
class GitHubPublisher(GitHubPublisherMixin, OIDCPublisher)
...
    def verify_url(self, url: str):
        # the parent's `verify_url` contains the logic that is currently in `_verify_url_with_trusted_publisher()`
        if super().verify_url(url):  
            return True

        docs_url = f"https://{self.repository_owner}.github.io/{self.repository_name}"
        return url == docs_url or url.startswith(f"{docs_url}/")
```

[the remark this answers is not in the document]
woodruffw: > Then, for provider-specific verification rules we want to add (like GitHub's `user.github.io/repo`) , we can override that method in the GitHub subclass to add that logic. WDYT?

That seems reasonable to me -- the only hesitation I have is that it makes these OIDC mixins bigger and bigger, but in terms of grouping related components together it makes sense 🙂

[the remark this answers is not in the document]
di: Given that we might be verifying URLs from a variety of source truths (which may or may not be trusted publishers) I think it's OK as-is for now.

warehouse/forklift/legacy.py
```diff
@@ -866,7 +903,12 @@ def file_upload(request):
         else {
             name: {
                 "url": url,
-                "verified": _verify_url(url=url, publisher_url=publisher_base_url),
+                "verified": _verify_url(
+                    url=url,
+                    publisher_url=publisher_base_url,
+                    project_name=project.name,
+                    project_normalized_name=project.normalized_name,
+                ),
```
di, line 911: I think I would prefer something like this instead unless there's a strong reason not to:

```suggestion
                "verified": any(
                    check(
                        url=url,
                        publisher_url=publisher_base_url,
                        project_name=project.name,
                        project_normalized_name=project.normalized_name,
                    )
                    for check in [_verify_url_pypi, _verify_url_with_trusted_publisher]
                ),
```

facutuesca: It would mean we need to pass the same parameters to both `_verify_*` functions, so some of them will be unused in each.

di: I think that's OK? I'm concerned that `_verify_url` has logic that short-circuits verification outside of the individual checks, and thinking it'll be more straightforward to just write additional checks and add them to this list than determine where they should fall in the logic for that function.

facutuesca: I see your point. Could you check https://github.com/trail-of-forks/warehouse/pull/3013 ? It's my idea for moving the TP-specific verification to the OIDCPublisher class, since we'll need to specialize them depending on the specific TP provider. 

If we go ahead with that, at least one check will be a method of that class, and the proposed change here would have to be:
```python
any(
    check(...)
    for check in [_verify_url_pypi, publisher.verify_url] if publisher else [_verify_url_pypi])
```
(or similar), and we would still need to keep the same common list of parameters for all verify functions.

If that's fine I can accept the suggestion. My only concern is that as verification logic grows, having all current and future functions take the same list of parameters might not be great.

di: > since we'll need to specialize them depending on the specific TP provider.

That's a good point. I think I'd be in favor of that refactor once we introduce more verifications. Then this could become something like:

```suggestion
                "verified": any(
                    check(url)
                    for check in [
                        partial(
                            _verify_url_pypi(
                                project_name=project_name,
                                project_normalized_name=project_normalized_name,
                            )
                        ), 
                        partial(publisher.verify_url) if publisher else lambda _: False
                ),
```

I think I'm over-optimizing a bit here though, going to approve/merge this as-is for now and we can revisit when things change.

di (approved):

Issue #14812, opened 2023-10-26
Upload blocked by a pre-existing file that does not exist.


m3vaz:
**Describe the bug**

While uploading wheels for a new version of 11.8.3, wheel upload was blocked for python 3.9 linux-x64 reporting
``` 400 File already exists. See https://pypi.org/help/#file-name-reuse for more information. ```

The security log confirms that the version was created today and the linux-x64 file has not been uploaded today. The rest of the wheel uploads were completed successfully. 

**Expected behavior**

Wheel should upload.

**To Reproduce**

Uploading the wheel for cuda-python 11.8.3 for python 3.9 for x86_64 i.e. `cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` 

**My Platform**

Upload was attempted using twine with the following command:
``` python -m twine upload .\cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl --verbose ```

**Additional context**

**Comment by jakirkham** [comment]
@di is there an easy way to clear the error status on PyPI?

As far as we can tell PyPI shows no record of this file being uploaded, but the CLI still gives us this error. Unclear why that is happening

**Comment by di** [comment]
Can you provide the SHA-256 hash of the file you're trying to upload? E.g.:

```
$ sha256sum cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

**Comment by m3vaz** [comment]
Hi @di I missed your response, sorry about that:
```
$ sha256sum cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
04465e1179213e14a316f0eedbb50dc416f701f3f135d4ba16ce4b2892ec4286  cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

**Comment by leofang** [comment]
Friendly nudge @di 😅

**Comment by di** [comment]
Oddly I see no record of this file already existing:

```
warehouse=> select id from release_files where filename = 'cuda_python-11.8.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl';
 id
----
(0 rows)
```

```
warehouse=> select id from release_files where sha256_digest = '04465e1179213e14a316f0eedbb50dc416f701f3f135d4ba16ce4b2892ec4286';
 id
----
[...]

PR #16488, opened 2024-08-14
Include filename and hash in duplicate file error message

di:
This PR adds the filename provided and the hash of the file contents provided on upload to the error message generated when a duplicate file is found.

This should help with debugging issues like #14812, and also give the user more information about which file was duplicated in the event of multiple file uploads.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1096,7 +1096,9 @@ def file_upload(request):
                 # --skip-existing functionality in twine
                 # ref: https://github.com/pypi/warehouse/issues/3482
                 # ref: https://github.com/pypa/twine/issues/332
-                "File already exists. See "
+                "File already exists "
+                + f"({filename!r}, with blake2_256 hash {file_hashes['blake2_256']!r})."
+                + " See "
                 + request.help_url(_anchor="file-name-reuse")
                 + " for more information.",
             )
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -1096,7 +1096,9 @@ def file_upload(request):
                 # --skip-existing functionality in twine
                 # ref: https://github.com/pypi/warehouse/issues/3482
                 # ref: https://github.com/pypa/twine/issues/332
-                "File already exists. See "
+                "File already exists "
+                + f"({filename!r}, with blake2_256 hash {file_hashes['blake2_256']!r})."
```
miketheman, line 1100: The only concern I have here is whether `'blake2_256'` is guaranteed here, or should we do a `.get()` instead, so we don't raise a `KeyError`:

```suggestion
                + f"({filename!r}, with blake2_256 hash {file_hashes.get('blake2_256')!r})."
```

If the key is guaranteed to be in the dictionary, then this is fine as is!

miketheman (approved):

warehouse/forklift/legacy.py
di: It'll be there as long as it's here, and I don't expect this to go anywhere anytime soon: https://github.com/pypi/warehouse/blob/b4b04243e1e77329edca590c27207aeff06a2ce2/warehouse/forklift/legacy.py#L1075

Issue #16487, opened 2024-08-14
Trusted Publishing: tweak reusable workflow metric to include project name


woodruffw:
Following https://github.com/pypi/warehouse/pull/16364: we now know that only a very small number of projects are actively triggering this metric, but we don't know which. We should add a field/qualifier to the metric showing which project (or publisher URL would be fine) triggered it.

PR #16497, opened 2024-08-16
Add publisher_url to the github_reusable_workflow metric.

DarkaMaul:
Closes https://github.com/pypi/warehouse/issues/16487


woodruffw (approved):
Thanks @DarkaMaul!

di:
Taking a step back: our initial goal was to determine if reusable workflows were being used (they are) and now we need to determine which projects are using reusable workflows, correct?

I don't think tags are the right thing to use here. I think we could either add a field to the `Publisher` model that captures whether it has ever used reusable workflows (might be helpful to maintain backwards compatibility if we break this in the future?) or adding some additional information to the event that gets created when the publisher is used.

Maybe both?

woodruffw:
> Taking a step back: our initial goal was to determine if reusable workflows were being used (they are) and now we need to determine which projects are using reusable workflows, correct?

Yep, correct!

And hmm, that makes sense to me -- I *think* we might want to do this just at the `Event` layer, since we're doing this in preparation for a series of changes/migration against the `Publisher` model, and the magnitude of reusable workflows observed will probably influence how we approach designing those migrations 🙂

di:
That works for me, let's do that instead.

DarkaMaul:
I've updated the PR to generate an Event instead.

I created a new Event Type because assigning it to either Project or File seemed off. I think this is the most forward-thinking approach, but I'm happy to change it to another EventType if needed.

di:
We have an existing `FileEvent` that included details about the publisher specific to the file, I think that's the right place for this: https://github.com/pypi/warehouse/blob/bfe6dc7e5fa642cc63981511972e05b2d69404bd/warehouse/forklift/legacy.py#L989-L1004

di (approved):
Thanks!

PR #16499, opened 2024-08-16
Verify github.io URLs with Trusted Publishing

facutuesca:
As discussed on https://github.com/pypi/warehouse/pull/16485#discussion_r1718417410, this moves the logic for verifying URLs against a Trusted Publisher to `OIDCPublisherMixin.verify_url()`. It also specializes that function on `GithubPublisher` to also allow `github.io` documentation URLs.

Here's a screenshot where the documentation link (a link to `github.io`) is verified:

image

Finishes fixing https://github.com/pypi/warehouse/issues/16474

cc @woodruffw @di


tests/unit/oidc/models/test_github.py
```diff
@@ -656,6 +656,26 @@ def test_github_publisher_duplicates_cant_be_created(self, db_request):
...
             db_request.db.commit()
 
+    @pytest.mark.parametrize(
+        ("url", "expected"),
+        [
+            ("https://repository_owner.github.io/repository_name/", True),
+            ("https://repository_owner.github.io/repository_name", True),
+            ("https://repository_owner.github.io/repository_name/subpage", True),
+            ("https://repository_owner.github.io/repository_name/../malicious", False),
+            ("https://repository_owner.github.io/", False),
```
woodruffw, line 666: Might want a testcase for an unrelated subpage as well, e.g.:

```suggestion
            ("https://repository_owner.github.io/unrelated_name/", False),
```

warehouse/forklift/legacy.py
```diff
@@ -481,44 +482,11 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
-        and is_subpath
-    )
-
-
 def _verify_url(
-    url: str, publisher_url: str | None, project_name: str, project_normalized_name: str
+    url: str,
+    publisher: OIDCPublisher | None,
+    project_name: str,
+    project_normalized_name: str,
```
woodruffw, line 489: I might be missing something, but why do we pass both the project name and its normalized version here? Could we instead produce the normalized name within the function body, to save ourselves a parameter here?

(Apologies in advance if there's a good reason to do this that I've missed!)

facutuesca: I couldn't find the logic for normalizing the name, other than a reference to a SQL function called `normalize_pep426_name`.

woodruffw: Ah, right 🤦. I suppose we could pass the entire `Project` model in, but that's a lot heavier. This seems fine to me, then.

di (approved):

PR #16515, opened 2024-08-19
Improve Pending Trusted Publishers UX when project already exists

facutuesca:
This PR changes the error message displayed when a user tries to create a Pending Trusted Publisher for a project that already exists. The new error message includes a link to the regular Trusted Publisher form for that project, with URL parameters that will pre-fill the form with the values the user already input in the Pending form:

The link will be something like:
```
https://pypi.org/manage/project/ruff/settings/publishing/?project_name=ruff&owner=my_user&repository=my_project&workflow_filename=release.yml&provider=github
```

The main logic change is in `forms/_core.py`. The validator for the `project_name` field (`validate_project_name()`) is changed so that the link to the pre-filled Trusted Publisher form is included in the error message when the project already exists.

Creating the link requires us to have access to the `request.route_url` method, which we pass the same way the `project_factory` parameter is currently passed to the validator (through the form constructor)

See the comment in https://github.com/pypi/warehouse/issues/14232#issuecomment-2276538291 for more context on why we use this approach rather than implicitly creating a Trusted Publisher for the existing project.

Fixes https://github.com/pypi/warehouse/issues/14232

cc @woodruffw

### DIFF
--- a/warehouse/oidc/forms/_core.py
+++ b/warehouse/oidc/forms/_core.py
@@ -9,7 +9,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import markupsafe
 import wtforms
 
 from warehouse import forms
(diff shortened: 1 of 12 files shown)


di (approved):

warehouse/oidc/forms/_core.py
```diff
@@ -31,13 +31,29 @@ def validate_project_name(self, field):
...
+            # We mark the error message as safe, so that the HTML hyperlink is
+            # not escaped by Jinja
             raise wtforms.validators.ValidationError(
-                _(
-                    "This project already exists, create an ordinary Trusted "
-                    "Publisher instead"
+                markupsafe.Markup(
+                    f"This project already exists, use the project's publishing"
+                    f" settings <a href='{url}'>here</a> to create a Trusted"
+                    f" Publisher for it."
```
twm, line 48: Shouldn't this be translated?

woodruffw: Yeah -- I think we need to either push this down a layer, or make sure the snippet generated here passes through `_(...)` correctly.

facutuesca: I'm looking into it now

facutuesca: Fixed here: https://github.com/pypi/warehouse/pull/16526

PR #16516, opened 2024-08-19
feat(admin): set primary email after recovery complete

### LINKED ISSUE #16401: "Start Account Recovery" Process Doesn't Update to Alternate Email Address
Currently, when a user with existing projects requests an email update/change, as part of the account verification process, the "start account recovery" function doesn't automatically update to the user's new/preferred email address after branch verification has been completed. This feature will remove the need for PyPI Support and/or Admin to manually update a user's email address.

An ideal solution would be: when the account verification process is initiated, if an input for an alternate email exists, the "Complete Recovery" function should also update a user's primary email address to the alternate email address, in addition to its current functionality of resetting the 2FA and recovery codes. 

**Additional context**

miketheman:
If the Account Recovery process has an **alternate** email that is to be used during the process, 'activate' it at the end once completed.

Resolves #16401

### DIFF
--- a/tests/unit/admin/views/test_users.py
+++ b/tests/unit/admin/views/test_users.py
@@ -1414,6 +1414,59 @@ class TestUserRecoverAccountComplete:
             pretend.call(username=user.username)
         ]
 
+    def test_user_recover_account_complete_with_override_email_sets_as_primary(
+        self, db_request, monkeypatch
+    ):
+        user = UserFactory.create(with_verified_primary_email=True)
+        existing_primary_email = user.primary_email
+
+        assert len(user.emails) == 1
+
+        # Create preconditions from `views.user_recover_account_initiate`
+        override_to_email = EmailFactory.create(
+            user=user, primary=False, verified=False
+        )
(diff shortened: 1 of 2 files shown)


ewdurbin:
I think this fits the need. One thought that I hadn't really considered is if we should skip password/2FA reset in these conditions. My understanding is that _generally_ this state occurs when someone has their password/2FA but has lost access to the original primary email address.

@Thespi-Brain, do you have any input here?

Thespi-Brain:
@ewdurbin I think we should still keep the password/2FA reset because there have been cases where 2FA was not set up in addition to the user not having access to their original primary email address. It's also nice to have a "clean" reset for due diligence and security.

ewdurbin (approved):
@Thespi-Brain I am assuming that for situations where users still control their password/2FA, we will update their email address without resetting password/2FA. If so this is good to go.

Thespi-Brain:
@ewdurbin Yes, we only don't go through the account recovery process for an email update when the user either (a) doesn't have 2FA/recovery codes or (b) when a user doesn't have any projects

Issue #16475, opened 2024-08-13
Gitlab's repository URL didn't marked as verified after trusted publishing


nE0sIghT:
**Describe the bug**
I just tested URL verify (merged in the https://github.com/pypi/warehouse/pull/16205) for Gitlab and found that repository git URL didn't became verified: https://pypi.org/project/apt-mirror/9a1/

**Expected behavior**
Gitlab URLs like `https://gitlab.com/apt-mirror2/apt-mirror2.git` are verified after trusted publishing.

woodruffw:
What happens if you change your GitLab URL to remove the `.git` prefix? As-is, I suspect it's not being marked as verified because it's failing this check:

https://github.com/pypi/warehouse/blob/293a2304379326abf2bdc90073f944ff9d8119b2/warehouse/forklift/legacy.py#L485-L488

(Tweaking this check is tricky, but not necessarily impossible -- we'd need to special case suffixes like `.git` when checking subpaths. But I think the majority of people use "bare" URLs for GitHub and GitLab, in practice.)

di:
We should be safe to strip `.git` from the end of GitHub URLs if present:

<img width="708" alt="image" src="https://github.com/user-attachments/assets/873ad71d-adf8-4bd7-9453-6a8f70e10f6c">

Add I'm assuming the same is true with GitLab as well.

nE0sIghT:
> What happens if you change your GitLab URL to remove the .git suffix?

Didn't tested it yet, but I believe it will be fine

> I think the majority of people use "bare" URLs for GitHub and GitLab, in practice

Just a note here that the repository URL example in the ["Writing your pyproject.toml"](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) PyPA article have `.git` suffix:

![image](https://github.com/user-attachments/assets/60f0c4b6-47cf-4ccb-8035-d63f0bc31b5a)
 
(it's a place where I referenced while writing pyproject.toml for my initial pypi publishing)

woodruffw:
Gotcha! TIL that's recommended in the PyPA guide 🙂 

(I was off in my estimation too -- it should be relatively easy for us to specialize the suffix check for GitHub and GitLab.)

facutuesca:
> We should be safe to strip `.git` from the end of GitHub URLs if present:
> 
[...]

PR #16525, opened 2024-08-20
Verify URLs ending with `.git` for GitHub and GitLab

facutuesca:
Fixes https://github.com/pypi/warehouse/issues/16475

cc @woodruffw


woodruffw (approved):
LGTM, thanks @facutuesca!

(I just checked, and GitHub allows all kinds of funkiness in their redirects, e.g. `foo/bar.git.git` will redirect to `foo/bar`. But I suspect that trying to chase all of these down has limited value/would drive us nuts 🙂)

di (approved):

PR #16528, opened 2024-08-20
Fix warning in tests for URL verification

facutuesca:
Fixes the warnings mentioned here: https://github.com/pypi/warehouse/pull/16205#issuecomment-2299327343.

Now we get the Release object from the DB, and we access the verified/unverified URLs using the method `Release.urls_by_verify_status()`, instead of manually querying the database for the URLs.

cc @woodruffw


miketheman (approved):
Thanks!

PR #16530, opened 2024-08-20
Adopt a linter for pytest style & fix linting issues

### LINKED ISSUE #16517: Adopt a linter for pytest style
See https://pypi.org/project/flake8-pytest-style/

A sample run on `main` produces ~450 violations:

```shellsession
$ python -m flake8 --select=P --statistics .
...
9     PT004 fixture 'enable_organizations' does not return anything, add leading underscore
1     PT005 fixture '_request' returns a value, remove leading underscore
117   PT006 wrong name(s) type in @pytest.mark.parametrize, expected tuple
7     PT007 wrong values type in @pytest.mark.parametrize, expected list of tuples
271   PT009 use a regular assert instead of unittest-style 'assert_'
4     PT011 pytest.raises(ValueError) is too broad, set the match parameter or use a more specific exception
4     PT012 pytest.raises() block should contain a single simple statement
9     PT014 found duplicate test cases (1, 3) in @pytest.mark.parametrize
2     PT015 assertion always fails, replace with pytest.fail()
1     PT018 assertion should be broken down into multiple parts
20    PT019 fixture _purge_key_calls without value is injected as parameter, use @pytest.mark.usefixtures instead
1     PT022 no teardown in fixture mockredis, use return instead of yield
```

We could adopt the plugin and not enforce it yet, and burn down the violations gradually, ro try and do it in larger chunks, and selectively disable rules we're not ready to adopt yet.

di:
Fixes #16517.

The only warning being ignored is [`PT009`](https://github.com/m-burst/flake8-pytest-style/blob/v2.0.0/docs/rules/PT009.md) due to all the renderer assertions in files like https://github.com/pypi/warehouse/blob/main/tests/unit/email/test_init.py that come from Pyramid itself.


miketheman (commented):
This is super excellent and amazing, nice work!

One minor improvement could be to only ignore PT009 on the test files that use it, via config like so:

```ini
per-file-ignores =
    tests/unit/email/test_init.py: PT009
```

globs should work there as well.

miketheman (approved):

PR #16531, opened 2024-08-20
Fix missing unverified URLs

facutuesca:
Currently, verified and unverified URLs are shown by getting them from a `Release` object using the `Release.urls_by_verify_status()` method. However, this method only takes into account URLs in `Release._project_urls`, ignoring the fields `Release.home_page` and `Release.download_url`, which means the URLs in those two fields are not displayed (unless the URLs are also present in `_project_urls`)

Since those two fields are currently just `str` and do not store any verification information, we should consider them as unverified and return them when the user calls `Release.urls_by_verify_status(False)`, unless they happen to also be present in `_project_urls` as verified URLs.

This PR fixes `Release.urls_by_verify_status` so that it also includes `Release.home_page` and `Release.download_url` in the list of unverified URLs, iff those URLs are not verified in `Release._project_urls`. 

cc @woodruffw


warehouse/packaging/models.py
```diff
@@ -690,12 +690,31 @@ def urls(self):
...
+            # The Homepage and Download URLs in the release metadata are currently not
+            # verified, so we return them if the user requests non-verified URLs *and*
+            # they are not verified in `project_urls`.
+            matching_urls.update(
+                {
+                    url
+                    for url in (self.home_page, self.download_url)
+                    if url is not None and url not in verified_urls
+                }
+            )
```
woodruffw, line 716: Just summarizing to make sure I follow:

* If the context requests verified URLs `urls_by_verify_status(True)`, we return only URLs from `_project_urls` that are marked as verified.
* If the context requests unverified URLs `urls_by_verify_status(False)`, we return unverified URLs from `_project_urls`, *merged with* the free-standing `Release` URLs (`home_page` and `download_url`) *if* those URLs aren't in the verified set.

Is that correct? If so, this LGTM!

warehouse/packaging/models.py
```diff
@@ -690,12 +690,31 @@ def urls(self):
         return _urls
 
     def urls_by_verify_status(self, verified: bool):
```
woodruffw, line 692: Tiny nitpick: perhaps not worth addressing in this PR, but I think `verified: bool` could become a require kwarg to make method calls here slightly comprehensible:

```suggestion
    def urls_by_verify_status(self, *, verified: bool):
```

warehouse/packaging/models.py
facutuesca: Yes, that is correct!

di (approved):

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -88,7 +88,7 @@ <h6>{% trans %}GitHub Statistics{% endtrans %}</h6>
     </ul>
   </div>
   {% endif %}
-
+
```
di, line 91:
```suggestion

```

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -107,13 +107,13 @@ <h6>{% trans %}Maintainers{% endtrans %}</h6>
   {% endif %}
 </div>
 
-<div class="sidebar-section unverified">
+<div class="sidebar-section verified">
```
di, line 110:
```suggestion
<div class="sidebar-section unverified">
```

PR #16546, opened 2024-08-21
Reapply "Store attestations for PEP740 (#16302)" (#16545)

woodruffw:
WIP, needs full backstop tests.

This reverts commit da7e1ed.


warehouse/migrations/versions/4037669366ca_recreate_attestations_table.py
woodruffw, line 1: NB: I made a whole new migration rather than futzing with the history on the old ones. But I'm not sure if that's the preferred approach 🙂

tests/conftest.py
```diff
@@ -315,6 +315,7 @@ def get_app_config(database, nondefaults=None):
         "opensearch.url": "https://localhost/warehouse",
         "files.backend": "warehouse.packaging.services.LocalFileStorage",
         "archive_files.backend": "warehouse.packaging.services.LocalArchiveFileStorage",
+        "archive_files.path": "/var/opt/warehouse/packages-archive/",
```
woodruffw, line 318: Flagging: I cargo-culted this from `dev/environment`. These changes need this setting as well, since the `IntegrityService` constructs an `IFileStorage` with `name="archive"` under the hood, which translates to `LocalArchiveFileStorage` in the local environment. That in turn requires `archive_files.path`, per:

```py
@implementer(IFileStorage)
class LocalFileStorage(GenericLocalBlobStorage):
    @classmethod
    def create_service(cls, context, request):
        return cls(request.registry.settings["files.path"])
```

tests/unit/api/test_simple.py
```diff
@@ -87,6 +88,16 @@ def test_selects(self, header, expected):
...
 class TestSimpleIndex:
+
+    @pytest.fixture
+    def db_request(self, db_request):
+        """Override db_request to add the Release Verification service"""
+        db_request.find_service = lambda svc, name=None, context=None: {
+            IIntegrityService: pretend.stub(),
+        }.get(svc)
+
+        return db_request
```
woodruffw, line 99: Flagging: I think we want to drop this stub and below. This is probably what masked the underlying lack of a configured service.

woodruffw: Removed in https://github.com/pypi/warehouse/pull/16546/commits/b151b821271882715432ec2922b230db55fa542b.

tests/unit/attestations/test_services.py
```diff
@@ -0,0 +1,421 @@
...
+        with tempfile.NamedTemporaryFile() as f:
+            integrity_service = IntegrityService(
+                storage=pretend.stub(get=pretend.call_recorder(lambda p: f)),
+                metrics=pretend.stub(),
+            )
+
+            assert (
+                integrity_service.get_provenance_digest(file)
+                == hashlib.file_digest(f, "sha256").hexdigest()
+            )
```
woodruffw, line 379: Flagging: I think we need to not mock at this layer to gain more confidence here -- `db_request` should now have a proper `IntegrityService` injected into it that we can use to create, persist, and digest an actual attestation object.

tests/conftest.py
```diff
@@ -533,6 +536,16 @@ def activestate_oidc_service(db_session):
     )
 
 
+@pytest.fixture
+def local_file_storage(tmp_path):
+    return packaging_services.GenericLocalBlobStorage(tmp_path)
```
woodruffw, line 541: Flagging: No idea if this is a good idea or not 🙂. Some other options:

1. Avoid a fixture here, inline the `GenericLocalBlobStorage` instance into the `integrity_service` fixture
2. Avoid an actual service entirely and prefer a stub/mock of an `IFileService`.

woodruffw:
Listing some action items that need to be accomplished before undrafting:

- [ ] The current `TestSimpleDetail` tests should be extended to be include a case where `provenance` is actually present, not just `None`
- [ ] https://github.com/pypi/warehouse/pull/16546#pullrequestreview-2252143243
- [ ] More generally, go through `TestAttestationsService` (which should be renamed `TestIntegrityService`) with a fine-toothed comb and ensure we aren't mocking overly agressively

tests/conftest.py
woodruffw: Removed this entirely, now that we have a dev-level `NullIntegrityService`.

tests/unit/attestations/test_services.py
woodruffw: Done -- I've added a non-default `storage_service` fixture that can be injected for these kinds of tests/

tests/conftest.py
woodruffw: No longer relevant.

tests/conftest.py
```diff
@@ -110,6 +112,15 @@ def metrics():
     )
 
 
+@pytest.fixture
+def storage_service(tmp_path):
+    """
+    A good-enough local file storage service.
+    """
+
+    return packaging_services.LocalArchiveFileStorage(tmp_path)
```
woodruffw, line 121: NB: I added this fixture, but I didn't include it in the default `pyramid_services` fixture -- I don't think there's any harm in including it there, although relatively few tests should actually need it.

warehouse/config.py
```diff
@@ -454,12 +460,14 @@ def configure(settings=None):
...
-        default="100 per hour",
-    ),
+    (
+        maybe_set(
+            settings,
+            "warehouse.account.accounts_search_ratelimit_string",
+            "ACCOUNTS_SEARCH_RATELIMIT_STRING",
+            default="100 per hour",
+        ),
+    )
```
DarkaMaul, line 470:
```suggestion
    maybe_set(
        settings,
        "warehouse.account.accounts_search_ratelimit_string",
        "ACCOUNTS_SEARCH_RATELIMIT_STRING",
        default="100 per hour",
    )
```

tests/unit/packaging/test_utils.py
```diff
@@ -27,11 +28,37 @@ def test_simple_detail_empty_string(db_request):
...
+    db_request.find_service = pretend.call_recorder(
+        lambda svc, name=None, context=None: {
+            IIntegrityService: pretend.stub(
+                get_provenance_digest=pretend.call_recorder(lambda f: hash_digest),
+            ),
+        }.get(svc)
+    )
+
+    expected_content = _simple_detail(project, db_request)
+    assert expected_content["files"][0]["provenance"] == hash_digest
```
DarkaMaul, line 59: Suggestion: I think it would make sense to use the integrity service here instead of mocking it. 

```python
def test_simple_detail_with_provenance(db_request, integrity_service):
    release = ReleaseFactory.create(project=project, version="1.0")
    file = FileFactory.create(release=release)
    file.attestations.append(AttestationFactory.create(file=file))

    db_request.route_url = lambda *a, **kw: "the-url"
    db_request.find_service = pretend.call_recorder(
        lambda svc, name=None, context=None: {
            IIntegrityService: integrity_service,
        }.get(svc)
    )

    expected_content = _simple_detail(project, db_request)
    assert isinstance(expected_content["files"][0]["provenance"], str)
```

tests/conftest.py
DarkaMaul: I've added it to the service registered in the `pyramid_services` so we can use it in several tests without redefining the `find_services` method.

tests/unit/attestations/test_services.py
```diff
@@ -80,37 +79,18 @@ class TestIntegrityService:
...
-    def test_persist_attestations(self, db_request, monkeypatch):
-        @pretend.call_recorder
-        def storage_service_store(path: str, file_path, *_args, **_kwargs):
-            expected = VALID_ATTESTATION.model_dump_json().encode("utf-8")
-            with open(file_path, "rb") as fp:
-                assert fp.read() == expected
-
-            assert path.endswith(".attestation")
+    def test_create_service(self, db_request):
+        service = IntegrityService.create_service(None, db_request)
```
DarkaMaul, line 83: FLAG: I've changed this test to reuse the `db_request` fixture.

The newly created service gets its two dependent services directly from the fixture. 
I wanted to test that they were correctly created but the `service.storage.base.exists` seems a bit off.

tests/conftest.py
```diff
@@ -191,7 +204,9 @@ def pyramid_services(
     services.register_service(
         activestate_oidc_service, IOIDCPublisherService, None, name="activestate"
     )
+    services.register_service(integrity_service, IIntegrityService, None, name="")
     services.register_service(macaroon_service, IMacaroonService, None, name="")
+    services.register_service(storage_service, IFileStorage, None, "archive")
```
woodruffw, line 209: For consistency:

```suggestion
    services.register_service(storage_service, IFileStorage, None, name="archive")
```

tests/unit/attestations/test_services.py
woodruffw: > I wanted to test that they were correctly created but the service.storage.base.exists seems a bit off.

Yeah, let's not assert that here -- that's an implementation detail from an unrelated interface/service.

tests/unit/packaging/test_utils.py
woodruffw: Makes sense to me, go ahead and make this change!

warehouse/migrations/versions/4037669366ca_recreate_attestations_table.py
di: That's fine!

di:
Oh interesting, the TUF failure seems to have been an intermittent thing, this now works. Will have to look into why the attestation hash digest is different.

di:
Here's what I was getting, for context:

```
============================================= short test summary info =============================================
FAILED tests/functional/api/test_simple.py::test_simple_attestations_from_upload - webtest.app.AppError: Bad response: 400 Unknown error while trying to verify included attestations: Failed to refresh TUF metadata (not 200)
400 Unknown error while trying to verify included attestations: Failed to refresh TUF metadata

The server could not comply with the request since it is either malformed or otherwise incorrect.


Unknown error while trying to verify included attestations: Failed to refresh TUF metadata

Results (16.29s):
       1 failed
         - tests/functional/api/test_simple.py:48 test_simple_attestations_from_upload
make: *** [Makefile:79: tests] Error 1
```

warehouse/packaging/utils.py
```diff
@@ -97,6 +100,7 @@ def _simple_detail(project, request):
                     if file.metadata_file_sha256_digest
                     else False
                 ),
+                "provenance": integrity_service.get_provenance_digest(file),
```
di, line 103:
```suggestion
                "provenance_digest": integrity_service.get_provenance_digest(file),
```
Or something similar

DarkaMaul: This key name is defined by the [PEP 740](https://peps.python.org/pep-0740/#json-based-simple-api).
We could slightly tweak the PEP if we want to update the name here, but we should keep both synced.

warehouse/attestations/services.py
```diff
@@ -0,0 +1,323 @@
...
+        if not file.attestations:
+            return None
+
+        # For the null service, our "provenance digest" is just the digest
+        # of the release file's name merged with the number of attestations.
+        # We do this because there's no verification involved; we just need
+        # a unique value to preserve invariants.
+        return hashlib.sha256(
+            f"{file.filename}:{len(file.attestations)}".encode()
+        ).hexdigest()
```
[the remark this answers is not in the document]
DarkaMaul: I may be wrong but I think you are trying to compare the attestation file's hash against the provenance file's hash.

DarkaMaul:
> I added a functional test for this in [17e9d45](https://github.com/pypi/warehouse/commit/17e9d45f9c66f4781cf5dc29c6fdc854f9719437) but it's currently failing trying to refresh the TUF metadata. @woodruffw @DarkaMaul thoughts on what we should do about TUF here?

I've also started writing some tests for this at https://github.com/trail-of-forks/warehouse/pull/3063, and my solution there was to mock the verification.

However, I did not assert that the hashes were correct (and they are probably not).

warehouse/packaging/utils.py
di: This is just the variable name that's passed to the Jinja template for the /simple page, we can make it whatever we want.

warehouse/templates/api/simple/detail.html
```diff
@@ -20,7 +20,7 @@
   <body>
     <h1>Links for {{ name }}</h1>
     {% for file in files -%}
-    <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}" {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}" {% endif %}{% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}" {% endif %}{% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}>{{ file.filename }}</a><br />
+    <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}" {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}" {% endif %}{% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}" {% endif %}{% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}{% if file.get('provenance') %}data-provenance="{{ file['provenance'] }}" {% endif %}>{{ file.filename }}</a><br />
```
di, line 23:
```suggestion
    <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}" {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}" {% endif %}{% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}" {% endif %}{% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}{% if file.get('provenance_digest') %}data-provenance="{{ file['provenance_digest'] }}" {% endif %}>{{ file.filename }}</a><br />
```

tests/functional/api/test_simple.py
```diff
@@ -31,3 +45,76 @@ def test_simple_api_detail(webtest):
...
+            "attestations": f"[{attestation}]",
+        },
+        upload_files=[("content", "sampleproject-3.0.0.tar.gz", content)],
+        status=HTTPStatus.OK,
+    )
+
+    response = webtest.get("/simple/sampleproject/", status=HTTPStatus.OK)
+    link = response.html.find("a", text="sampleproject-3.0.0.tar.gz")
+    assert "data-provenance" in link.attrs
+    assert link.get("data-provenance") == expected_hash
```
DarkaMaul, line 120: This test fails here because the hash we get here is 

```python
hashlib.sha256(
    b"sampleproject-3.0.0.tar.gz:2"   # notice the 2 here
).hexdigest()
```

However, the `len(attestations)` should be 1.

The problem lies in [generate_provenance](https://github.com/pypi/warehouse/blob/e72b746d034b32f19c20eeefb11d5edf606d1011/warehouse/attestations/services.py#L141-L149) of `NullService` that creates the `DatabaseAttestation` and adds it to the `file.attestations` (like in [_persist_attestations](https://github.com/pypi/warehouse/blob/e72b746d034b32f19c20eeefb11d5edf606d1011/warehouse/attestations/services.py#L277-L287) )

Because the relationship is noted as [`back_populates`](https://docs.sqlalchemy.org/en/20/orm/relationship_api.html#sqlalchemy.orm.relationship.params.back_populates), sqlalchemy automatically adds the newly created `DatabaseAttestation` to the `file` instance. 

The statement below is thus redundant and creates add a second time the attestation to the file instance.

```python
file.attestations.append(database_attestation)
```

However, if we look at [`test_persist_attestations_succeeds`](https://github.com/pypi/warehouse/blob/226204147a7c743b447a8b6014f955266b26824f/tests/unit/attestations/test_services.py#L86), the test passes with the following statements : 

```python
        assert len(attestations_db) == 1
        assert len(file.attestations) == 1
```

I've observed the state in a debugger using sqlalchemy tools and they looked appropriate. 
```
from sqlachemy import inspect
state = inspect(database_attestation)
```

Any idea here of the reason for this behavior ?

warehouse/packaging/utils.py
DarkaMaul: I must be missing something here but I think the `_simple_detail` function is called by [`simple_detail`](https://github.com/pypi/warehouse/blob/4271f2e9e7d9ec28aa5a5a531d3b004bb5f1375f/warehouse/api/simple.py#L112-L113) and used when returning the `JSON` response for the `simple` API.

I have tried with :
```
requests.get("http://localhost/simple/requests", headers={"Accept": "application/vnd.pypi.simple.v1+json"}).json()
```

And I effectively get the new key in the answer : 

```
   'provenance_digest': None
```

di: Ah, you're right, forgot it was being re-used there. Nevermind!

tests/functional/api/test_simple.py
di: The issue is that the `Attestation` objects weren't actually being added to the session, the test was emitting a warning like:

```
tests/unit/attestations/test_services.py::TestIntegrityService::test_generate_provenance_succeeds[GitHubPublisherFactory]
  /opt/warehouse/src/warehouse/attestations/models.py:53: SAWarning: Object of type <Attestation> not in session, add operation along 'File.attestations' will not proceed (This warning originated from the Session 'autoflush' process, which was invoked automatically in response to a user-initiated operation.)
    f"{Path(self.file.path).name}.attestation",
```

I think f7e277e and fa1bd58 should resolve this.

warehouse/attestations/services.py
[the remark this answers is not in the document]
di: Good catch!

di (approved):
LGTM, will wait to merge until next week though.

tests/functional/api/test_simple.py
DarkaMaul: Thanks for sorting this out - I missed the warning in the logs.

ewdurbin (approved):

PR #16556, opened 2024-08-22
Add some releases to the Simple API functional test

di:
A follow-on to #16547: this PR adds releases to the project used in the functional test for the Simple API.

It also restructures the functional test directory structure to mirror the source directory structure.


tests/functional/api/test_simple.py
```diff
@@ -24,6 +24,8 @@ def test_simple_api_html(webtest):
 
 def test_simple_api_detail(webtest):
     project = ProjectFactory.create()
+    ReleaseFactory.create(version="1.0.0", project=project)
+    ReleaseFactory.create(version="2.0.0", project=project)
```
miketheman, line 28: totally fine, could also be:
```suggestion
    ReleaseFactory.create_batch(2, project=project)
```

tests/functional/forklift/test_legacy.py
```diff
@@ -13,6 +13,24 @@
...
 
+def test_incorrect_post_redirect(webtest):
+    """
+    Per issue #8104, we should issue an HTTP-308 for a POST
+    in /legacy and point the user to the correct endpoint,
+    /legacy/
+
+    See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+    """
+    resp = webtest.post("/legacy", status=308)
```
miketheman, line 24: improvement: I've been trying to use more descriptive statuses from `http` since I can't ever recall the specific codes.

needs `from http import HTTPStatus`

```suggestion
    resp = webtest.post("/legacy", status=HTTPStatus.PERMANENT_REDIRECT)
```

miketheman (approved):
Nice! I have a couple of potential improvements, neither are blockers.

tests/functional/api/test_simple.py
di: Does this ensure that the versions are unique? It's unclear from https://factoryboy.readthedocs.io/en/stable/reference.html#factory.create_batch

miketheman: We control the `version` with a sequence already to ensure uniqe: https://github.com/pypi/warehouse/blob/ca2781e6f4b947a9739fe05ac33315fcbac3075d/tests/common/db/packaging.py#L93

tests/functional/forklift/test_legacy.py
di: OK, I've updated this everywhere.

PR #16562, opened 2024-08-23
refactor: replace integers with durations in seconds

miketheman:
Refactor usages of integers to constants calculated from `timedelta` for clarity.

---

Example of some caught:

```python
$ python -m flake8 warehouse --show-source --tee | tee | head -n5
warehouse/accounts/views.py:163:22: WH003 Prefer `datetime.timedelta([seconds|hours|days]=...).total_seconds()` over integer multiplication to create durations. See https://docs.python.org/3/library/datetime.html#datetime.timedelta
        origin_cache(1 * 24 * 60 * 60, stale_if_error=1 * 24 * 60 * 60)  # 1 day each.
                     ^
warehouse/accounts/views.py:163:55: WH003 Prefer `datetime.timedelta([seconds|hours|days]=...).total_seconds()` over integer multiplication to create durations. See https://docs.python.org/3/library/datetime.html#datetime.timedelta
        origin_cache(1 * 24 * 60 * 60, stale_if_error=1 * 24 * 60 * 60)  # 1 day each.
```

We could also probably create a new [constant](https://github.com/pypi/warehouse/blob/6e9123d44518eaf13a0c03245d9b9e91ffcd98ae/warehouse/constants.py#L1-L0) for things like `DAY_IN_SECONDS` and reuse liberally.

### DIFF
--- a/warehouse/search/tasks.py
+++ b/warehouse/search/tasks.py
@@ -121,7 +121,7 @@ def reindex(self, request):
     try:
         with SearchLock(
             r,
-            timeout=datetime.timedelta(minutes=3).total_seconds(),
+            timeout=datetime.timedelta(minutes=30).total_seconds(),
             blocking_timeout=30,
         ):
             p = parse_url(request.registry.settings["opensearch.url"])
--- a/warehouse/search/tasks.py
+++ b/warehouse/search/tasks.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import binascii
+import datetime
 import os
 import urllib.parse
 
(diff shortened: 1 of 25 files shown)


miketheman:
The checker isn't perfect, for sure. But I think all of the cases caught are indeed time duration calculations.

warehouse/search/tasks.py
```diff
@@ -118,7 +119,11 @@ def reindex(self, request):
     """
     r = redis.StrictRedis.from_url(request.registry.settings["celery.scheduler_url"])
     try:
-        with SearchLock(r, timeout=30 * 60, blocking_timeout=30):
+        with SearchLock(
+            r,
+            timeout=datetime.timedelta(minutes=3).total_seconds(),
```
ewdurbin, line 124: typo: this moved timeout from 30 min to 3 minutes.

ewdurbin (approved):

Issue #16540, opened 2024-08-21
Verify `Home-page` and `Download-URL` URLs


facutuesca:
**What's the problem this feature will solve?**
Currently, only URLs specified via [`Project-URL`](https://packaging.python.org/en/latest/specifications/core-metadata/#project-url-multiple-use) are verified (as part of the work for https://github.com/pypi/warehouse/issues/8635). 

However, the [`Home-page`](https://packaging.python.org/en/latest/specifications/core-metadata/#home-page) and [`Download-URL`](https://packaging.python.org/en/latest/specifications/core-metadata/#download-url) URLs are not verified. These are currently stored in the backend inside the `Release` object as [plain](https://github.com/pypi/warehouse/blob/c1dffa93e5f68b0ade87aaeacae8857442bb6130/warehouse/packaging/models.py#L561) [strings](https://github.com/pypi/warehouse/blob/c1dffa93e5f68b0ade87aaeacae8857442bb6130/warehouse/packaging/models.py#L573), as opposed to the URLs from `Project-URL` which have their own type ([`ReleaseURL`](https://github.com/pypi/warehouse/blob/c1dffa93e5f68b0ade87aaeacae8857442bb6130/warehouse/packaging/models.py#L485)) and are stored in their own table along with their verification status.

Since these two URLs are not verified, they are always returned as unverified (via https://github.com/pypi/warehouse/pull/16531).

**Describe the solution you'd like**
We can change the backend so that these two URLs are verified during upload, and their verification status is stored, similar to what we do for URLs in `Project-URL` fields.

/cc @di @woodruffw

miketheman:
> Since these two URLs are not verified, they are always returned as unverified (via #16531).

They are also no longer displayed at all.

facutuesca:
> They are also no longer displayed at all.

@miketheman That's what https://github.com/pypi/warehouse/pull/16531 fixes

miketheman:
> > They are also no longer displayed at all.
> 
> @miketheman That's what #16531 fixes

@facutuesca thanks, I confirmed the fix is there after a page cache re-render.

PR #16568, opened 2024-08-26
Verify `Home-Page` and `Download-URL` metadata URLs

facutuesca:
This PR adds verification for the `Home-Page` and `Download-URL` fields in the Release metadata.

Currently, a `Release` has:

```python
class Release:
    home_page: Mapped[str | None]
    download_url: Mapped[str | None]

    _project_urls: Mapped[list[ReleaseURL]] = orm.relationship(
        collection_class=attribute_keyed_dict("name"),
        cascade="all, delete-orphan",
        order_by=lambda: ReleaseURL.name.asc(),
        passive_deletes=True,
    )
```

So the `_project_urls` field contains all of the `Project-URL`s for that release, from a table called `ReleaseURL`. That table also contains a `verified: bool` column, which stores the verification status for those URLs.

However, the `home_page` and `download_url` fields are stored separately as `str`. To make things more complicated, `_project_urls` can also contain URLs for the Homepage and Download URL, and they can differ from the ones above.

We could try to move `home_page` and `download_url` into the `ReleaseURL` table. However, that would mean every time we interact with `_project_urls` we would need to add logic to handle the special case of those two URLs. Furthermore, it would make the `_project_urls` data not match the metadata sent by the user using the `Project-URL` key.

The solution proposed by this PR is to add two new boolean fields to `Release`:
```python
class Release:
    home_page: Mapped[str | None]
    home_page_verified: Mapped[bool_false]

    download_url: Mapped[str | None]
    download_url_verified: Mapped[bool_false]
```

This way we can store the verified status right next to the URL, and it makes the verification logic straightforward. 

Fixes https://github.com/pypi/warehouse/issues/16540

cc @woodruffw


woodruffw:
Tying things together: I've opened https://discuss.python.org/t/core-metadata-should-home-page-and-download-url-be-deprecated/62037 to discuss whether `Home-page` and `Download-URL` can be deprecated outright. That'll need to go through a PEP, though, so I _think_ it probably makes sense to add verification handling for these URLs for the time being.

warehouse/forklift/legacy.py
```diff
@@ -1384,6 +1410,15 @@ def file_upload(request):
...
 
+        if release.home_page and not release.home_page_verified and home_page_verified:
+            release.home_page_verified = True
+        if (
+            release.download_url
+            and not release.download_url_verified
+            and download_url_verified
+        ):
+            release.download_url_verified = True
+
```
di, line 1421: I don't think there's a case where `release.home_page` would be falsey but `home_page_verified` would be true, so this can just be:

```suggestion
        if home_page_verified and not release.home_page_verified:
            release.home_page_verified = True
        if download_url_verified and not release.download_url_verified:
            release.download_url_verified = True

```

di (commented):
LGTM aside from one comment!

di (approved):

PR #16571, opened 2024-08-26
feat: create helpdesk service interface

miketheman:
As well as implement services for HTTP and Console


warehouse/helpdesk/interfaces.py
```diff
@@ -0,0 +1,16 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from zope.interface import Interface
+
+
+class IHelpDeskService(Interface): ...
```
miketheman, line 16: Note to self: probably need to actually write the interface functions here, but nothing in linter-land seems to care? 🤔

ewdurbin (commented):
Overall, I like this!

Thoughts:

We currently only interact with one mailbox (security), does it make sense to refactor the configuration to use `maybe_set_compound` so that we could have

```
HELPDESK_SECURITY_BACKEND="warehouse.helpdesk.services.HelpscoutService mailbox_id=111111"
HELPDESK_SUPPORT_BACKEND="warehouse.helpdesk.services.HelpscoutService mailbox_id=222222"
```

So that we can have multiple contexts to load the service to initiate conversations in either inbox.

warehouse/helpdesk/interfaces.py
miketheman: Yep, that was it. I've added in 28462ff9de631a94f3a99c1bdd1eef64a87f34d2

miketheman:
> Don't forget to set `HELPDESK_BACKEND` in cabotage before deploy!

Thanks for the reminder, done!

PR #16575, opened 2024-08-27
Remove Depot registry pull token auth

jacobwgillespie:
We have implemented automatic authentication for the Depot registry inside Depot Actions runners - this means that because the `test` and `check_db` jobs run on Depot runners, they now automatically have permission to pull the build container from the Depot registry without needing to create a new `depot pull-token` as part of the workflow.

This simplifies the workflow and also fixes an issue where if those jobs were retried greater than 60 minutes after the pull token was created, they would fail to pull the build container with the stale credentials.

Eventually this new automatic auth feature will be enabled via a toggle in the Depot UI, for now we've manually enabled it for your `pypi` org on the backend.

cc @ewdurbin @miketheman

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     runs-on: depot-ubuntu-22.04-arm
     outputs:
       buildId: ${{ steps.build.outputs.build-id}}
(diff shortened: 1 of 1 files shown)


ewdurbin (approved):

ewdurbin:
Thank you @jacobwgillespie!

PR #16585, opened 2024-08-28
Add verification date to Verified Details section

facutuesca:
Once URLs are marked as verified, PyPI does not repeat the verification afterwards (for that specific package version. New versions do trigger a new verification). This means that the "URL verified" status should be interpreted in that context: there is no guarantee that a URL verified 5 years ago is still in control of the package maintainer.

In order to make this a bit more explicit, this PR adds a date to the tooltip of the verified checkmark:
image

This date is the release creation date. While this date will be correct for the big majority of packages (since URLs are verified at the time of release creation), there are some cases where the uploader will upload files to an existing release at a later date, triggering verification for URLs that are not yet verified.

So while some URLs might have been verified afterwards, using the release creation date should still be fine, since what we want to avoid is giving the users the impression that a URL was verified more recently than it was, and the release creation date is the furthest point in the past where verification could have happened.

cc @woodruffw


warehouse/templates/includes/packaging/project-data.html
```diff
@@ -18,7 +18,7 @@ <h3 class="sidebar-section__title">{% trans %}Verified details{% endtrans %}&nbs
           <small><i>{% trans %}(What is this?){% endtrans %}</i></small>
       </a>
   </h3>
-  <small><i>These details have been verified by PyPI</i></small>
+  <small><i>These details have been verified by PyPI on {{ release.created.date() }}</i></small>
```
woodruffw, line 21: Language nitpick:

```suggestion
  <small><i>These details were verified by PyPI on {{ release.created.date() }}</i></small>
```

Thinking out loud, though: maybe we should put the verification date behind a tooltip, rather than rendering it in the text here? My thinking is that rendering it adds to the visual noise in this section, but I'm curious what @miketheman and @di think.

facutuesca: > have been -> were
fixed!

> text section is missing a {% trans %} group

fixed, and also fixed for the Unverified Details section

di: I agree, this is getting a little busy (especially with the addition of the 'what is this?' bit).

What about a tooltip on the checkmark that says "Verified on {date}"? They would all be the same, but it would make it possible to have re-verifications at different times (or asynchronously after a release is made) in the future.

facutuesca: Fixed, I added the date to the existing tooltip of the checkmark, and updated the screenshot and PR description.

miketheman (approved):

Issue #16136, opened 2024-06-20
Check Email MX domains against prohibited email list


miketheman:
A user account may use an email domain that is handled by an email service, "whitelabelling" the new domain, as a means to evade detection.

Validation check [`wtforms.validators.Email()`](https://wtforms.readthedocs.io/en/3.0.x/validators/#wtforms.validators.Email) accepts a `check_deliverability` boolean that performs DNS lookups to make this determination via the `email_validator` package.

I'm not sure how to collect the response from a `wtforms.validator` built-in like `Email()`, but we could leverage the perform our own checks, like so:

```pycon
>>> from email_validator import validate_email
>>> res = validate_email("foo@gmail.com", check_deliverability=True)
>>> print(res.mx)
[(5, 'gmail-smtp-in.l.google.com'), (10, 'alt1.gmail-smtp-in.l.google.com'), (20, 'alt2.gmail-smtp-in.l.google.com'), (30, 'alt3.gmail-smtp-in.l.google.com'), (40, 'alt4.gmail-smtp-in.l.google.com')]
```

We could feasibly iterate through each mx result and compare against the blocked list and reject new registrations thusly.
Probably around here:
https://github.com/pypi/warehouse/blob/7eb6030d073ef31968a278005803eb88d18081fa/warehouse/accounts/forms.py#L287-L301

miketheman:
Another example `spacehotline.com` and `in.mail.tm`

PR #16596, opened 2024-08-29
feat: check MX records against prohibited domains

miketheman:
Using the `email-validator` package, which is already part of `wtforms`'s dependencies, check if the domain is actually deliverable prior to accepting the address.

Deliverability check produces the MX records in the response.
Use them to check against the Prohibited Domains lists.

Resolves: #16136


ewdurbin (commented):
I'm worried that this is going to have unintended side effects in combination with #16586.

For instance:

```
$ dig microsoft.com MX +short
10 microsoft-com.mail.protection.outlook.com.
```

And we have [`outlook.com`on our prohibited domains list](https://blog.pypi.org/posts/2024-06-16-prohibiting-msn-emails/), which I think means we will bar @microsoft.com addresses from registering?

ewdurbin:
Yeah, it seems applying:

```diff
diff --git a/tests/unit/manage/test_forms.py b/tests/unit/manage/test_forms.py
index 6250fbb6e..22bfd934d 100644
--- a/tests/unit/manage/test_forms.py
+++ b/tests/unit/manage/test_forms.py
@@ -247,7 +247,7 @@ class TestAddEmailForm:
 
         Simulate that `mail.net` handles emails sent to `wutang.net`.
         """
-        mock_deliverability_info = {"mx": [(10, "mail.net")]}
+        mock_deliverability_info = {"mx": [(10, "microsoft-com.mail.protection.outlook.com.")]}
 
         def mock_function(*args, **kwargs):
             return mock_deliverability_info
@@ -257,12 +257,12 @@ class TestAddEmailForm:
             mock_function,
         )
 
-        prohibited_mx_domain = ProhibitedEmailDomain(domain="mail.net")
+        prohibited_mx_domain = ProhibitedEmailDomain(domain="outlook.com")
         db_request.db.add(prohibited_mx_domain)
 
         form = forms.AddEmailForm(
             request=db_request,
-            formdata=MultiDict({"email": "foo@wutang.net"}),
+            formdata=MultiDict({"email": "foo@microsoft.com"}),
             user_service=pretend.stub(find_userid_by_email=lambda _: None),
             user_id=pretend.stub(),
         )
```

still passes test (aka blocks the address).

miketheman:
Thanks for poking at it!
I think this is a use case that needs some more thought, so I'll convert to Draft for now.

warehouse/accounts/forms.py
```diff
@@ -279,20 +277,27 @@ def __init__(self, *args, request, **kwargs):
...
+                for _prio, mx_host in resp.mx
+            }
+        all_domains = mx_domains.union({domain})
 
         if (
-            domain in disposable_email_domains.blocklist
+            any(d in disposable_email_domains.blocklist for d in all_domains)
             or self.request.db.query(
-                exists().where(ProhibitedEmailDomain.domain == domain)
+                exists().where(ProhibitedEmailDomain.domain.in_(all_domains))
```
miketheman, line 300: From conversation, @di suggests a separate ProhibitedMxDomains table and use that explicitly, or extend the existing domain with an `mx = true` column, so we could decide how widely we want to prohibit a given domain name.

warehouse/accounts/forms.py
```diff
@@ -562,11 +574,11 @@ def validate_username_or_email(self, field):
         if "@" in field.data:
             # Additional checks for the validity of the address
             try:
-                Address(addr_spec=field.data)
-            except (IndexError, ValueError, HeaderParseError):
+                email_validator.validate_email(field.data, check_deliverability=True)
```
ewdurbin, line 577: This doesn't seem like the right place to do this. Wouldn't this block someone from resetting their password if their domain or MX domain is blocked?

ewdurbin: Oh, or is this just additional validation of the email along the way. :)

warehouse/accounts/forms.py
```diff
@@ -562,11 +574,11 @@ def validate_username_or_email(self, field):
...
+            except email_validator.EmailNotValidError as e:
                 raise wtforms.validators.ValidationError(
                     message=INVALID_PASSWORD_MESSAGE
```
ewdurbin, line 580: While we're here, is this the right message?

ewdurbin (approved):

warehouse/accounts/forms.py
miketheman: This is changing the validation to be uniform with registration, as well as prevent someone from setting an seemingly valid-appearing email address that has no MX records, meaning they effectively cut off that email address from use, since it can never be verified.

warehouse/accounts/forms.py
miketheman: Heh, nope. Thanks for pointing that out, will add a commit to use a better message.

miketheman: Added in b1d6704aeaeb754246fd3ce3de9f3848a8ee4b40

PR #16597, opened 2024-08-29
Add logic to verify URLs using HTML meta tag

facutuesca:
Part of https://github.com/pypi/warehouse/issues/8635, this PR adds a function to verify anbritrary URLs by parsing their HTML and looking for a specific `meta` tag.

Concretely, a webpage with a `meta` tag in its header element like the following:
```html
<meta content="package1 package2" namespace="pypi.org" rel="me" />
```
would pass validation for the `project1` and `project2` PyPI projects.

**This PR only adds the function and its tests. The function is not used anywhere yet.**

This implementation takes into account the discussion in the issue linked above, starting with this comment: https://github.com/pypi/warehouse/issues/8635#issuecomment-2292013010.

Concretely:
- URLs must use `https://`
- The hostname must be a regular name (i.e.: `domain.tld`), it cannot be an P address (e.g: `https://100.100.100.100`)
- If a port is present, it must be 443 (we could also remove this, and require that no port is present)
- Before getting the HTML, we resolve the URL to an IP address, and check that it's a global IP and not a private or shared IP
- We limit the amount of content we download to 1024 bytes (this number was an arbitrary choice, it's open to changes)
- HTML is parsed using `lxml`, which recovers from partial HTML, meaning only reading the first N bytes should be fine as long as it contains the tag we are looking for

I'm opening this PR with only the verification logic since it's the part that requires the most review and discussion. Once it's done we can see how to integrate it with the current upload flow (probably as an asynchronous task).

cc @woodruffw @ewjoachim

### DIFF
--- a/warehouse/packaging/metadata_verification.py
+++ b/warehouse/packaging/metadata_verification.py
@@ -10,7 +10,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ipaddress
+import socket
+
+import requests
 import rfc3986
+
+from lxml import etree, html
+from rfc3986 import validators as rfc3986_validators
+from rfc3986.exceptions import ValidationError as rfc3986ValidationError
+from rfc3986.misc import IPv4_MATCHER, IPv6_MATCHER
 
 from warehouse.oidc.models import OIDCPublisher
 
@@ -39,6 +48,86 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str) 
         user_uri == rfc3986.api.uri_reference(candidate_url).normalize()
         for candidate_url in candidate_urls
     )
+
+
+def _get_url_content(url: rfc3986.api.URIReference, max_length_bytes: int) -> str:
+    """
+    Get at least `max_length_bytes` of the contents of a URL
+
+    Raises `requests.exception.RequestException` on any errors while trying to
+    access the URL.
+    """
+    headers = {"User-Agent": "warehouse/1.0 (PyPI URL verifier, https://pypi.org)"}
+    r = requests.get(
+        url.unsplit(),
+        allow_redirects=False,
+        timeout=5,
+        stream=True,
+        headers=headers,
+    )
+    r.raise_for_status()
+
+    content = next(r.iter_content(max_length_bytes))
+    return content
+
+
+def _verify_url_meta_tag(
+    url: str, project_name: str, project_normalized_name: str
+) -> bool:
+    user_uri = rfc3986.api.uri_reference(url).normalize()
+
+    # Require the presence of "https" and the host name. If a port
+    # is present, it must be 443.
+    validator = (
+        rfc3986_validators.Validator()
+        .require_presence_of("scheme", "host")
+        .allow_schemes("https")
+        .allow_ports("443")
+    )
+
+    try:
+        validator.validate(user_uri)
+    except rfc3986ValidationError:
+        return False
+
+    # Only allow registered names for the host (e.g.: google.com, api.github.com),
+    # don't allow IP addresses.
+    if IPv4_MATCHER.fullmatch(user_uri.host) or IPv6_MATCHER.fullmatch(user_uri.host):
+        return False
+
+    # The domain name should not resolve to a private or shared IP address
+    try:
+        address_tuples = socket.getaddrinfo(user_uri.host, user_uri.port)
+        for family, _, _, _, sockaddr in address_tuples:
+            ip_address: ipaddress.IPv4Address | ipaddress.IPv6Address | None = None
+            if family == socket.AF_INET:
+                ip_address = ipaddress.IPv4Address(sockaddr[0])
+            elif family == socket.AF_INET6:
+                ip_address = ipaddress.IPv6Address(sockaddr[0])
+            if ip_address is None or not ip_address.is_global:
+                return False
+    except (socket.gaierror, ipaddress.AddressValueError):
+        return False
+
+    # We get the first 1024 bytes
+    try:
+        content = _get_url_content(user_uri, max_length_bytes=1024)
+    except requests.exceptions.RequestException:
+        return False
+
+    try:
+        html_root = html.document_fromstring(content)
+    except (StopIteration, etree.ParserError):
+        return False
+
+    meta_tag = html_root.xpath("//head/meta[@namespace='pypi.org' and @rel='me']")
+
+    content = meta_tag[0].get("content") if len(meta_tag) > 0 else None
+    if content is None:
+        return False
+
+    packages = content.split()
+    return project_name in packages or project_normalized_name in packages
 
 
 def verify_url(
--- a/requirements/lint.in
+++ b/requirements/lint.in
@@ -16,6 +16,7 @@ types-certifi
 types-first
 types-html5lib
 types-itsdangerous
+types-lxml
 types-passlib
 types-python-slugify
 types-pytz
--- a/requirements/lint.txt
+++ b/requirements/lint.txt
@@ -155,6 +155,10 @@ cryptography==42.0.8 \
     # via
     #   types-pyopenssl
     #   types-redis
+cssselect==1.2.0 \
+    --hash=sha256:666b19839cfaddb9ce9d36bfe4c969132c647b92fc9088c4e23f786b30f1b3dc \
+    --hash=sha256:da1885f0c10b60c03ed5eccbb6b68d6eff248d91976fcde348f395d54c9fd35e
+    # via types-lxml
 curlylint==0.13.1 \
     --hash=sha256:008b9d160f3920404ac12efb05c0a39e209cb972f9aafd956b79c5f4e2162752 \
     --hash=sha256:9546ea82cdfc9292fd6fe49dca28587164bd315782a209c0a46e013d7f38d2fa
@@ -430,6 +434,10 @@ types-babel==2.11.0.15 \
     --hash=sha256:282c184c8c9d81e8269212c1b8fa0d39ee88fb8bc43be47980412781c9c85f7e \
     --hash=sha256:d0579f2e8adeaef3fbe2eb63e5a2ecf01767fc018e5f3f36a3c9d8b723bd62c7
     # via -r requirements/lint.in
+types-beautifulsoup4==4.12.0.20240511 \
+    --hash=sha256:004f6096fdd83b19cdbf6cb10e4eae57b10205eccc365d0a69d77da836012e28 \
+    --hash=sha256:7ceda66a93ba28d759d5046d7fec9f4cad2f563a77b3a789efc90bcadafeefd1
+    # via types-lxml
 types-boto3==1.0.2 \
     --hash=sha256:15f3ffad0314e40a0708fec25f94891414f93260202422bf8b19b6913853c983 \
     --hash=sha256:a6a88e94d59d887839863a64095493956efc148e747206880a7eb47d90ae8398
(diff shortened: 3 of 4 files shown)


woodruffw:
> * If a port is present, it must be 443 (we could also remove this, and require that no port is present)

I'm +1 on removing this outright -- I think the volume of legitimate users who actually need to explicitly list a port is probably vanishingly small 🙂

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +50,86 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+    r = requests.get(
+        url.unsplit(),
+        allow_redirects=False,
+        timeout=5,
+        stream=True,
+        headers=headers,
+    )
+    r.raise_for_status()
+
+    content = next(r.iter_content(max_length_bytes))
```
woodruffw, line 70: I _think_ `r.raw.read(max_length_bytes)` will be slightly faster + more idiomatic here, since we're already setting `stream=True` 🙂

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +50,86 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+    return content
```
woodruffw, line 71: Thinking out loud: should we explicitly `r.close()` before existing this method? As-is, I _think_ this will leave the connection open and dangling until the server hangs up, meaning that we could end up slowly exhausting the number of available outbound sockets.

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +50,86 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+            if ip_address is None or not ip_address.is_global:
+                return False
+    except (socket.gaierror, ipaddress.AddressValueError):
+        return False
+
+    # We get the first 1024 bytes
+    try:
+        content = _get_url_content(user_uri, max_length_bytes=1024)
+    except requests.exceptions.RequestException:
+        return False
```
woodruffw, line 116: Flagging: this has an unfortunate TOC/TOU weakness, where `getaddrinfo` might return a public IP, and the subsequent resolution in `requests` might return a private/internal one. In other words, an attacker could race the check here to get it to pass.

I _think_ the correct way to do this is to either inspect the `socket` object underneath the `requests` response, _or_ to use the resolved IP directly and employ the `Host` header + SNI to resolve the correct domain from the server. But both of these are annoying to do 🙂 

For the latter, `HostHeaderSSLAdapter` from `requests_toolbelt` is possibly the most painless approach: https://toolbelt.readthedocs.io/en/latest/adapters.html#requests_toolbelt.adapters.host_header_ssl.HostHeaderSSLAdapter

(This may also be possible more easily via `urllib3` instead -- maybe @sethmlarson knows?)

sethmlarson: This *should* be easier with urllib3, you can send your own `Host` and SNI values more directly.

sethmlarson: Just confirmed with Wireshark that this works for urllib3:

```python
http = urllib3.HTTPSConnectionPool(
    host="93.184.215.14",
    port=443,
    headers={"Host": "example.com"},
    server_hostname="example.com",
    assert_hostname="example.com",
)
resp = http.request("GET", "/")
```

Sends SNI of `example.com`, asserts `example.com` on the cert, sends `example.com` in the `Host` header, but doesn't do any DNS resolution. I should probably add a super-strict integration test for this construction to the urllib3 test suite, the individual parts are tested but having it all tested together is quite nice to double-check.

woodruffw: Amazing, thanks @sethmlarson!

ewjoachim:
(I just opened a random blog (typed "some blog" in google, got into a list of best blog per category, in "education", first link was https://blog.ed.ted.com/, and its complete `<head>` tag it about 10k bytes. I think 100k bytes is probably much safer than 1024)

warehouse/packaging/metadata_verification.py
facutuesca: Good catch! This also applies to the new `urllib3` implementation, so we now call 
```python
    r.drain_conn()
    r.release_conn()
```
before returning

warehouse/packaging/metadata_verification.py
facutuesca: Thanks! The implementation now uses `urllib3`

facutuesca:
> (I just opened a random blog (typed "some blog" in google, got into a list of best blog per category, in "education", first link was https://blog.ed.ted.com/, and its complete `<head>` tag it about 10k bytes. I think 100k bytes is probably much safer than 1024)

Changed to 100000 bytes

warehouse/packaging/metadata_verification.py
woodruffw: Hmm, is there a way we can avoid `drain_conn`? My understanding is that `r.drain_conn(...)` will block until the entire response is read (and dropped), meaning that this will end up effectively reading the whole response instead of truncating after the first X bytes.

Per: https://urllib3.readthedocs.io/en/stable/reference/urllib3.response.html#urllib3.response.HTTPResponse.drain_conn

(If I got that right, then I _think_ we can avoid this by removing the call and doing nothing else, since we're already using a separate connection pool per request and there's no reason to release back to a pool that we don't reuse 🙂)

warehouse/packaging/metadata_verification.py
facutuesca: Fixed!

warehouse/packaging/metadata_verification.py
facutuesca: I think that makes sense, but shouldn't we close the connection? The [docs](https://urllib3.readthedocs.io/en/stable/advanced-usage.html#streaming-and-i-o) mention that if you don't care about returning the connection to the pool, you can call `close()` to close the connection:

> You can call the [close()](https://urllib3.readthedocs.io/en/stable/reference/urllib3.response.html#urllib3.response.HTTPResponse.close) to close the connection, but this call doesn’t return the connection to the pool, throws away the unread data on the wire, and leaves the connection in an undefined protocol state. This is desirable if you prefer not reading data from the socket to re-using the HTTP connection.

sethmlarson: Yeah I recommend calling `HTTPResponse.close()` instead of `drain_conn()` if you don't plan on reusing the connection, that will close out the socket.

facutuesca: fixed!

flying-sheep:
what’s the state of this? it’s super unfortunate that everything else being verified effectively hides our readthedocs link a little.

facutuesca:
@flying-sheep I believe there are currently no plans for merging this. There is an issue for improving the visibility of unverified links here: https://github.com/pypi/warehouse/issues/18199

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+        )
+    except GetContentError:
+        return False
+
+    try:
+        html_root = html.document_fromstring(content)
+    except (StopIteration, etree.ParserError):
+        return False
+
+    meta_tag = html_root.xpath("//head/meta[@namespace='pypi.org' and @rel='me']")
```
di, line 152:
```suggestion
    meta_tag = html_root.xpath("//head/meta[@name='pypi-project' and @rel='me']")
```
`namespace` doesn't seem to be a valid attribute per https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta.

Additionally, making this something that looks kind of like a URL potentially opens this up to unnecessary user confusion (e.g. putting `https://pypi.org` or `pypi.org/` or something else).

Finally, making the name just `pypi.org` might prevent us from using meta tags for other types of validation. It might not be necessary, but we could easily future-proof this a bit by making the name specific to project verification

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+        content = _get_url_content(
+            resolved_ip=resolved_ip,
+            url=user_uri,
+            max_length_bytes=_MAX_RESPONSE_LENGTH_BYTES,
+        )
+    except GetContentError:
+        return False
+
+    try:
+        html_root = html.document_fromstring(content)
```
di, line 148: I'm a little wary of invoking an entire HTML parser just for this -- given that we don't directly use `lxml` anywhere else, and that this can be pointed at any arbitrary URL.

I think we could probably be way less permissive in the 'valid' HTML that we accept and handle this with a regex instead.

If that isn't sufficient, I think I'd prefer using [`html.parser`](https://docs.python.org/3/library/html.parser.html) instead of `lxml` here.

If we _are_ going to use `lxml` for this, it needs to be added as a dependency in `requirements/main.in` as it's just a transitive dependency right now.

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+    Raises `GetContentError` on any errors while trying to
+    access the URL.
+    """
+    try:
+        http = urllib3.HTTPSConnectionPool(
+            host=resolved_ip,
+            port=443,
+            headers={
+                "Host": url.host,
+                "User-Agent": "warehouse/1.0 (PyPI URL verifier, https://pypi.org)",
```
di, line 74: Eventually making this a link to docs about this feature would be nice.

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+    # When using `preload_content=False`, the HTTP connection must be manually closed
+    r.close()
+
+    return content
+
+
+def _verify_url_meta_tag(
+    url: str, project_name: str, project_normalized_name: str
+) -> bool:
+    user_uri = rfc3986.api.uri_reference(url).normalize()
```
sethmlarson, line 96: To avoid URL parser differentials we should parse using `urllib3.util.parse_url()`. We can cross-check the result of the `scheme` and `host` and `port`.

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+    )
+
+    try:
+        validator.validate(user_uri)
+    except rfc3986ValidationError:
+        return False
+
+    # Only allow registered names for the host (e.g.: google.com, api.github.com),
+    # don't allow IP addresses.
+    if IPv4_MATCHER.fullmatch(user_uri.host) or IPv6_MATCHER.fullmatch(user_uri.host):
```
sethmlarson, line 115: Instead of only blocklisting IP addresses maybe we also allowlist common domain name structures? (eg `search(r"\.[a-z][^.]+$", user_uri.host)`). TBH, I would even welcome having an allowlist of gTLDs just to completely avoid the issue of platform-specific `socket.getaddrinfo()` behavior around encoding IP addresses.

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+
+    content = r.read(max_length_bytes)
+
+    # When using `preload_content=False`, the HTTP connection must be manually closed
+    r.close()
+
+    return content
+
+
+def _verify_url_meta_tag(
```
sethmlarson, line 93: Does this function need to handle I18N via IDNA? Or is that already applied to the URL at this point?

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+                ip_address = ipaddress.IPv4Address(sockaddr[0])
+            elif family == socket.AF_INET6:
+                ip_address = ipaddress.IPv6Address(sockaddr[0])
+            if ip_address is None or not ip_address.is_global:
+                return False
+    except (socket.gaierror, ipaddress.AddressValueError):
+        return False
+
+    # Use the first IP address returned by `getaddrinfo`
+    _, _, _, _, resolved_sockaddr = address_tuples[0]
```
sethmlarson, line 135: nit: You can unpack all in one here: `_, _, _, _, (resolved_ip, _) = address_tuples[0]`

warehouse/packaging/metadata_verification.py
```diff
@@ -41,6 +56,109 @@ def _verify_url_pypi(url: str, project_name: str, project_normalized_name: str)
...
+            server_hostname=url.host,
+            assert_hostname=url.host,
+            retries=False,
+            timeout=5,
+        )
+        r = http.request("GET", url.path if url.path else "/", preload_content=False)
+    except Exception:
+        raise GetContentError()
+
+    content = r.read(max_length_bytes)
```
sethmlarson, line 85: Exceptions can get raised here and `.close()`, too. Especially if content isn't preloaded.

sethmlarson (commented):
Some thoughts on URLs and HTTP

di:
One last thought: we should probably think about how this function should surface the reason _why_ validation failed. I think it's not going to be very helpful (for users or admins) if validation just fails and we can't explain why. I'm not sure this needs to be shared with the user but at least having a way to run the function and get something back other than `False` is going to be important here.

warehouse/packaging/metadata_verification.py
woodruffw: Ah yeah, good catch -- technically this could be `data-namespace` to be compliant, but I don't think there's any good reason to not use `name` 🙂 

(FWICT `rel="me"` is also not documented on `<meta>`, but seems to be widely used with `<link>`. So maybe we should use `<link>` for this? OTOH there's no actual "link" per se.)

di: Yeah, I'm also not really sure `rel=me` is necessary here, all we really need is the key/value pair that `name/content` already provides.

PR #16598, opened 2024-08-29
Add a functional test for file uploads

di:
This PR adds a functional test which performs a `POST` with a file upload.


tests/common/db/macaroons.py
```diff
@@ -0,0 +1,24 @@
...
+# limitations under the License.
+
+import factory
+
+from warehouse.macaroons.models import Macaroon
+
+from .base import WarehouseFactory
+
+
+class MacaroonFactory(WarehouseFactory):
```
miketheman, line 20: ✨ Yay! A new 🏭 Factory!

tests/conftest.py
```diff
@@ -333,6 +333,7 @@ def get_app_config(database, nondefaults=None):
         "sessions.url": "redis://localhost:0/",
         "statuspage.url": "https://2p66nmmycsj3.statuspage.io",
         "warehouse.xmlrpc.cache.url": "redis://localhost:0/",
+        "archive_files.path": "/tmp",
```
miketheman, line 336: nit: move this up, right beneath `"archive_files.backend"`

tests/functional/_fixtures/sampleproject-3.0.0.tar.gz
miketheman, line 1: It'd be nice for us to have something somewhere on what's in this, and how to make a new one

tests/functional/forklift/test_legacy.py
```diff
@@ -48,3 +56,61 @@ def test_remove_doc_upload(webtest):
...
+
+
+def test_file_upload(webtest):
+    user = UserFactory.create(
+        password=(  # 'password'
+            "$argon2id$v=19$m=1024,t=6,p=6$EiLE2Nsbo9S6N+acs/beGw$ccyZDCZstr1/+Y/1s3BVZ"
+            "HOJaqfBroT0JCieHug281c"
+        )
+    )
+    EmailFactory.create(user=user, verified=True)
```
miketheman, line 68: We have a UserFactory trait now, would that accomplish your needs?
```suggestion
    user = UserFactory.create(with_verified_primary_email=True)
```

What purpose does setting a hashed password value here? Could we supply a string instead, or does some other mechanic prevent this?

miketheman (commented):
Yay, functional testing at play!

Left a couple of inline notes.

The [test now appears on the "slow" list](https://github.com/pypi/warehouse/actions/runs/10621734746/job/29444571167#step:5:215) - this test takes ~1.2 seconds - I wonder what the slow part is?

tests/functional/forklift/test_legacy.py
di: I was having an issue where the factory was not actually making this a valid 'enabled' password, and the user account was not usable as a result. Maybe this should be a default value? I think we can definitely use the trait though.

tests/functional/_fixtures/sampleproject-3.0.0.tar.gz
di: This is just from https://pypi.org/project/sampleproject/3.0.0/#files. Where would you expect to find those details? Would inline with the test be OK?

tests/functional/forklift/test_legacy.py
di: Yeah this is why: https://github.com/pypi/warehouse/blob/ccff64920db7965078cf1fdb50f028e640328887/tests/common/db/accounts.py#L39

Causes this to fail: https://github.com/pypi/warehouse/blob/ccff64920db7965078cf1fdb50f028e640328887/warehouse/accounts/services.py#L342

tests/functional/_fixtures/sampleproject-3.0.0.tar.gz
miketheman: Ah excellent - that's cool. A link in the test case docstring would be where I'd look for it, or in a README file in the _fixtures subdir next to this file

di: I added a `README.md` to the fixtures directory.

tests/functional/forklift/test_legacy.py
miketheman: Yeah, the default password is `"!"` which is unix-y for disabled. We could probably change our default and see what happens!

di:
> The test now appears on the "slow" list - this test takes ~1.2 seconds - I wonder what the slow part is?

Probably the fact that this is invoking the entire upload view, which is pretty heavy, plus doing a file upload.

miketheman (approved):
Nice stuff! Feel free to explore the password issue now or later, but this is good to go

miketheman (approved):

PR #16622, opened 2024-09-03
services: don't send a Path where a str is expected

woodruffw:
This removes a use of a `Path` object, which was entirely superfluous since we perform no actual path operations on it.

Fixes https://python-software-foundation.sentry.io/share/issue/ba5ffbe9b9b34c6c841c80aa45e48b17/

### DIFF
--- a/warehouse/attestations/services.py
+++ b/warehouse/attestations/services.py
@@ -14,8 +14,6 @@ import tempfile
 import typing
 import warnings
 
-from pathlib import Path
-
 import sentry_sdk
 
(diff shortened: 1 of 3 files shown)


woodruffw:
Added type hints in https://github.com/pypi/warehouse/pull/16622/commits/385505e8ec3c25da93857bd2ea35d31fbf4275f6 and confirmed that they would have caught the mistyped variable here.

ewdurbin (approved):

PR #16623, opened 2024-09-03
Revert PEP 740 persistence

woodruffw:
Reverts:

* 16b49a1
* d1c1161


woodruffw:
Some TODOs:

* Triage the performance hit that @ewdurbin saw on the last 2 deployments
* Replay #16622 when un-reverting this
* Add an admin flag so that we don't have to keep reverting here
* Address https://python-software-foundation.sentry.io/share/issue/d6047f0cad724577842674c483df0f9e/ (caused by interface mismatch between `botocore`'s `StreamingBody` and `hashlib.file_digest`)

woodruffw:
NB: If possible, I suggest merging without a squash here to make un-reverting easier (since the un-revert can avoid having to rebuild the migration history in https://github.com/pypi/warehouse/pull/16623/commits/07ad5d3c15414397e433c36de6b387251cb47aee).

woodruffw:
> Would it be terrible to _leave_ the migration in place and apply additional migrations over it moving forward? We disable everything but squash merges as policy on the repo.

Hmm, I think that'll be okay: if my mental model is right that means the `attestations` table will be in place and the un-revert here will simply not include any migrations, right?

If so, I'll amend that last migration commit and the rest should be good to go 🙂

ewdurbin:
Not 100% sure honestly. I am wondering if alembic would detect the missing model and belch during the check_db task.

woodruffw:
> I am wondering if alembic would detect the missing model and belch during the check_db task.

I'll amend locally and see what happens!

woodruffw:
Given the above, I think the easiest/least painful thing to here is probably to squash merge, and then I'll just rebuild the migration again 🙂

ewdurbin (approved):

PR #16624, opened 2024-09-03
PEP 740 persistence, take 3

woodruffw:
WIP.

Action plan:

- [ ] Make a separate PR for the models + migration, to eliminate the zippered reverts that we've had to do in the past.
- [ ] Make a separate PR with an appropriate admin flag for disabling provenance persistence.
- [ ] Investigate the other action items in #16623 

This reverts commit b6cf775fe6e48c5cdc6d0ef145972da57a1156a3.


woodruffw:
I discussed the architectural challenges here a bit with @ewdurbin, and we came to the conclusion that the current scale/scope and expected volume doesn't warrant the complexity that's coming from the attestations feature needing to interact with object storage.

Given that files can currently only have a single attestation, our current conclusion is that the DB should contain a `Provenance` model, which in turn bears the individual attestation objects/bundles as members of a JSONB list.

In other words, the current data model of `File -> [Attestation]` will become `File -> Provenance?`, where `Provenance` tracks an interior set of attestation bundles as JSONB.

ewdurbin:
Hm, I think I came to a slightly different mental model than you.

I imagined immutable `Attestation` objects remaining as a concrete-db-backed model and `Provenance` being rendered from all attestations related to a file by joining them up then generating a hash. Bonus: this retains the ability for additional attestations to be added down the line.

woodruffw:
> I imagined immutable `Attestation` objects remaining as a concrete-db-backed model and `Provenance` being rendered from all attestations related to a file by joining them up then generating a hash. Bonus: this retains the ability for additional attestations to be added down the line.

I messed around with this a bit, but it has the "lifecycle" problem that our earlier design (which avoided persisting provenance objects to artifact storage at all) had: the provenance object needs access to the uploading `request` object to persist the appropriate Trusted Publisher metadata + claims. We could generate it outside of that `request` context, but doing so requires another mechanism to hold onto that `request` state. That becomes especially annoying with the OIDC claims, since those are a free-structured dict that aren't part of the TP state itself 😅 

I'm halfway through rebasing this PR on top of #16625, which will hopefully make it a _lot_ simpler and demonstrate how I think (I hope) we can still do mutability while sticking to only a single concrete DB-backed `Provenance` model.

tests/functional/api/test_simple.py
```diff
@@ -31,3 +42,88 @@ def test_simple_api_detail(webtest):
...
+            "attestations": f"[{attestation}]",
+        },
+        upload_files=[("content", "sampleproject-3.0.0.tar.gz", content)],
+        status=HTTPStatus.OK,
+    )
+
+    assert len(project.releases) == 1
+    assert project.releases[0].files.count() == 1
+    assert project.releases[0].files[0].provenance is not None
+
```
DarkaMaul, line 108:
```suggestion
    # While we needed to be authenticated to upload a project, this is no longer
    # required to view it.
    webtest.authorization = None
```

tests/unit/packaging/test_utils.py
```diff
@@ -32,6 +34,24 @@ def test_simple_detail_empty_string(db_request):
...
+    project = ProjectFactory.create()
+    release = ReleaseFactory.create(project=project, version="1.0")
+    file = FileFactory.create(release=release)
+
+    db_request.route_url = lambda *a, **kw: "the-url"
+    db_request.oidc_publisher = GitHubPublisherFactory.create()
+
+    provenance = integrity_service.build_provenance(
+        db_request, file, [dummy_attestation]
+    )
```
DarkaMaul, line 49:
```suggestion

    # Provenance objects are generated at upload time, so we
    # are mocking it here.
    provenance = integrity_service.build_provenance(
        pretend.stub(
            oidc_publisher=GitHubPublisherFactory.create()
        ), file, [dummy_attestation]
    )
```

I think we should not reuse the `db_request` object in the `integrity_service` because this is supposed to be during upload time.

woodruffw: True, good call!

warehouse/migrations/versions/1b9ae6ec6ec0_add_provenance_table.py
```diff
@@ -0,0 +1,61 @@
...
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    op.execute(
+        """
+        INSERT INTO admin_flags(id, description, enabled, notify)
+        VALUES (
+            'disable-pep740-support',
+            'Disable PEP 740 support.',
+            FALSE,
```
DarkaMaul, line 52: FLAG: By default, the flag is disabled meaning the feature is enabled.

warehouse/attestations/services.py
```diff
@@ -0,0 +1,249 @@
...
+        self.db = session
+
+    @classmethod
+    def create_service(cls, _context, request):
+        return cls(session=request.db)
+
+    def parse_attestations(
+        self, request: Request, _distribution: Distribution
+    ) -> list[Attestation]:
+        publisher: OIDCPublisher | None = request.oidc_publisher
```
DarkaMaul, line 160: FLAG: This test is performed in the IntegrityService but was not done here and adding it highlighted some issues.

woodruffw:
Superseded by #16801.

PR #16625, opened 2024-09-03
warehouse: PEP 740 models

woodruffw:
This is a breakout from #16624,
to reduce the complexity/headache of zippered reverts.


DarkaMaul:
I added a simple test and Factory in [34ec661](https://github.com/pypi/warehouse/pull/16625/commits/34ec6611f5d716095dd4ea05a835f6e122ff5b4e) that can easily be reverted if needed.

However, just thinking out loud here, I think we should also have a `version` column in the `Provenance` that contains the Provenance object version.

The current design forces us to keep a forward-compatible version of the Provenance object to use the opaque JSON object with a `model_validate_json` call. 
Having a separate field allows us to deviate from the original design and handle the deserialization logic without resorting to try/catch.

woodruffw:
> However, just thinking out loud here, I think we should also have a `version` column in the `Provenance` that contains the Provenance object version.
> 
> The current design forces us to keep a forward-compatible version of the Provenance object to use the opaque JSON object with a `model_validate_json` call.

I might be under-thinking it, but I don't think this will be an issue -- if we need to roll the provenance version forwards, we can always do something like `TypeAdapter(Provenance | ProvenanceV2).validate_json(...)` to avoid needing to explicitly test each version.

(But also, if we need to bump the version, we can always add a `version` column with `default=1`. So I don't think we're excessively constraining ourselves at the moment.)

tests/common/db/attestations.py
```diff
@@ -0,0 +1,30 @@
...
+class ProvenanceFactory(WarehouseFactory):
+    class Meta:
+        model = Provenance
+
+    file = factory.SubFactory(FileFactory)
+    # TODO(DM) Generate a better provenance mock
+    provenance = {}
+    provenance_digest = factory.LazyAttribute(
+        lambda o: hashlib.sha256(o.file.filename.encode("utf8")).hexdigest()
+    )
```
woodruffw, line 30: Modeling these fields accurately requires access to the services that we're defining in #16624, so I'm inclined to punt on adding the factory in this PR 🙂 

I'm going to save a patch for this and apply it on top of #16624, however.

warehouse/packaging/models.py
```diff
@@ -838,6 +839,11 @@ def __table_args__(cls):  # noqa
         nullable=True,
         comment="If True, the metadata for the file cannot be backfilled.",
     )
+    provenance: Mapped[Provenance] = orm.relationship(
```
di, line 844:
```suggestion

    # PEP 740
    provenance: Mapped[Provenance] = orm.relationship(
```

PR #16631, opened 2024-09-04
Verify emails in release metadata using PyPI user information

facutuesca:
Part of https://github.com/pypi/warehouse/issues/8635, this PR starts verifying the emails included in a release's metadata (`author_email` and `maintainer_email`) by comparing that email against the public and verified emails of the PyPI project's owners and maintainers.

If the emails match, we include the corresponding field in the "Verified details" section, under the Meta heading.

image

cc @woodruffw


ewdurbin:
I'm not sure that we want to do this. Since PyPI account email addresses are private, and release metadata is public, this would act as an oracle to determine user email addresses?

We may want to limit this to the case where the user has marked an email address as [public](https://github.com/pypi/warehouse/blob/7c4f2ff79d4f45b10a05f6d6852a6248934baad5/warehouse/accounts/models.py#L187-L191).

woodruffw:
> I'm not sure that we want to do this. Since PyPI account email addresses are private, and release metadata is public, this would act as an oracle to determine user email addresses?

Yeah, agreed -- I think a user leaving their email private is probably a strong signal that they don't want it positively identified by PyPI in a public fashion, even if it's disclosed as part of the metadata anyways.

facutuesca:
> We may want to limit this to the case where the user has marked an email address as [public](https://github.com/pypi/warehouse/blob/7c4f2ff79d4f45b10a05f6d6852a6248934baad5/warehouse/accounts/models.py#L187-L191).

Makes sense! I changed the code so that now we verify against all emails of an owner/maintainer that are public and verified.

warehouse/migrations/versions/0b74ed7d4880_add_verified_emails_columns_to_release.py
```diff
@@ -0,0 +1,51 @@
...
+Revises: dcf1e3986782
+Create Date: 2024-09-04 14:04:12.622697
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "0b74ed7d4880"
+down_revision = "dcf1e3986782"
```
di, line 25:
```suggestion
down_revision = "2e049cda494f"
```

warehouse/migrations/versions/0b74ed7d4880_add_verified_emails_columns_to_release.py
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add verified emails columns to Release
+
+Revision ID: 0b74ed7d4880
+Revises: dcf1e3986782
```
di, line 16:
```suggestion
Revises: 2e049cda494f
```

di (approved):

PR #16633, opened 2024-09-05
Remove dep on types-boto3

DarkaMaul:
This PR removes the dependency on `types-boto3`.

Indeed, the last version of `types-boto3` is in April 2022 : https://pypi.org/project/types-boto3/
And it is simply an alias for `boto3-stubs` which is regularly updated.

However, since there is no new version, running the command below does not update the dependency
```
pip-compile --allow-unsafe --generate-hashes --output-file=requirements/lint.txt --upgrade-package types-boto3 requirements/lint.in
```

And since `boto*` dependencies are excluded from dependabot, I also performed the update to move to 1.35 (from 1.34).

/cc @woodruffw


miketheman (approved):
Thanks!

PR #16641, opened 2024-09-05
Document upcoming XMLRPC deprecations

ewdurbin:
The actual deprecation will occur at Fastly edge initially allowing us to do this in a "rolling brownout" to notify users, so no code changes.


ewdurbin:
Ref: #16642

docs/dev/api-reference/xml-rpc.rst
```diff
@@ -20,30 +20,29 @@ Example usage (Python 3)::
...
+   '0.618': 14863648,
+  ...
 
 .. _changes-to-legacy-api:
 
 Changes to XMLRPC API
 ---------------------
 
+- ``list_packages``, ``package_releases``, ``release_urls``, and ``release_data``
+  permanently deprecated and disabled. See `Deprecated Methods`_ for alternatives.
```
miketheman, line 44: Yay! Should the existing changes entires remain for `package_releases` and `release_data` a few lines below this, or should they get strikethrough/removed?

miketheman (approved):
One question, but otherwise 🎉

docs/dev/api-reference/xml-rpc.rst
ewdurbin: Good catch! 044226077

ewdurbin:
Added updates to the Simple API docs in 72d02eadc, since we are pointing people there as replacements.

docs/dev/api-reference/legacy.rst
```diff
@@ -4,6 +4,11 @@ Legacy API
 The "Legacy API" provides feature parity with `pypi-legacy`_, hence the term
 "legacy".
 
+.. note::
+  This API is available as both HTML and JSON as described in `PEP 691`_.
+
+  It is reccommended that new intergrations use the JSON version.
```
mauritsvanrees, line 10: Two types, should be:

```
It is recommended that new integrations use the JSON version.
```

I can make a PR if wanted.  Dinner first though. :-)

ewdurbin: https://github.com/pypi/warehouse/pull/16644

PR #16652, opened 2024-09-06
remove documented but non-existent XMLRPC methods

ewdurbin:


miketheman (approved):
Never implemented, begone!

PR #16664, opened 2024-09-09
chore: remove obsolete csp header

miketheman:
Introduced in #1105.

> It is now obsolete, because all mixed content is now blocked if it can’t be autoupgraded.

Refs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content
Refs: https://www.w3.org/TR/mixed-content/#strict-checking


ewdurbin (approved):

ewdurbin:
Seems fine given docs, but is there harm in leaving it? I'm curious how this impacts older browsers that may have been enforcing the directive.

miketheman:
The harm I can currently conceive is that we emit a small amount of extra bytes in every header, and browsers like Firefox emit a warning.

The standard appears to try to have browsers upgrade insecure by default, and block anything not upgradable.
Since we already implement [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security), browsers should either load resources over HTTPS or block them automatically.

PR #16684, opened 2024-09-11
PEP 740: add IntegrityService and interface

woodruffw:
Another breakout from https://github.com/pypi/warehouse/pull/16624.

~Draft/WIP since it's blocked on https://github.com/pypi/warehouse/pull/16683.~


woodruffw:
This is good for a look! It ended up being larger than I was hoping due to the tests, but the core functionality here (the new services + interfaces) are only ~400 LoC.

I can try to split it apart futher (e.g. by starting with just the `NullIntegrityService`), LMKWYT 🙂

dev/environment
```diff
@@ -46,6 +46,7 @@ BREACHED_EMAILS=warehouse.accounts.NullEmailBreachedService
 BREACHED_PASSWORDS=warehouse.accounts.NullPasswordBreachedService
 
 OIDC_BACKEND=warehouse.oidc.services.NullOIDCPublisherService
+ATTESTATIONS_BACKEND=warehouse.attestations.services.NullIntegrityService
```
di, line 49:
```suggestion
INTEGRITY_BACKEND=warehouse.attestations.services.NullIntegrityService
```

warehouse/attestations/__init__.py
```diff
@@ -9,3 +9,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from warehouse.attestations.interfaces import IIntegrityService
+
+
+def includeme(config):
+    integrity_service_class = config.maybe_dotted(
+        config.registry.settings["attestations.backend"]
```
di, line 18:
```suggestion
        config.registry.settings["integrity.backend"]
```
A minor nit, but I think this was a holdover from before the rename.

warehouse/attestations/services.py
```diff
@@ -0,0 +1,249 @@
...
+        Process any attestations included in a file upload request
+
+        Attestations, if present, will be parsed and verified against the uploaded
+        artifact. Attestations are only allowed when uploading via a Trusted
+        Publisher, because a Trusted Publisher provides the identity that will be
+        used to verify the attestations.
+        Only GitHub Actions Trusted Publishers are supported.
+        """
+        publisher: OIDCPublisher | None = request.oidc_publisher
+        if not publisher or not publisher.publisher_name == "GitHub":
```
di, line 200: We should avoid special-casing Github here to make this easier to add new publishers in the future, something like this would be nice:

```suggestion
        if not publisher or not publisher.supports_attestations
```

di (approved):
lgtm with some nits

warehouse/attestations/services.py
woodruffw: Good catch! Adding.

woodruffw: Done! I've added a blanket-`False` `OIDCPublisher.supports_attestations` that can be overridden by each publisher type.

di (approved):

PR #16697, opened 2024-09-12
Configuration of discrete Redis databases

ewdurbin:
Towards #13702. Allow for configuration of different redis databases for a given usecase... move our least sensitive use cases to their own DBs.


warehouse/config.py
```diff
@@ -304,15 +318,15 @@ def configure(settings=None):
...
+    maybe_set_redis(settings, "celery.result_url", "REDIS_URL", db=12)
+    maybe_set_redis(settings, "celery.scheduler_url", "REDIS_URL", db=0)
+    maybe_set_redis(settings, "oidc.jwk_cache_url", "REDIS_URL", db=1)
     maybe_set(settings, "database.url", "DATABASE_URL")
     maybe_set(settings, "opensearch.url", "OPENSEARCH_URL")
     maybe_set(settings, "sentry.dsn", "SENTRY_DSN")
     maybe_set(settings, "sentry.transport", "SENTRY_TRANSPORT")
-    maybe_set(settings, "sessions.url", "REDIS_URL")
-    maybe_set(settings, "ratelimit.url", "REDIS_URL")
+    maybe_set_redis(settings, "sessions.url", "REDIS_URL", db=2)
```
ewdurbin, line 328: This will sign everyone out, once.

warehouse/config.py
```diff
@@ -304,15 +318,15 @@ def configure(settings=None):
...
+    maybe_set_redis(settings, "ratelimit.url", "REDIS_URL", db=3)
```
ewdurbin, line 329: This will reset all rate limiter states, once.

warehouse/config.py
```diff
@@ -337,7 +351,7 @@ def configure(settings=None):
         coercer=asbool,
         default=True,
     )
-    maybe_set(settings, "warehouse.xmlrpc.cache.url", "REDIS_URL")
+    maybe_set_redis(settings, "warehouse.xmlrpc.cache.url", "REDIS_URL", db=4)
```
ewdurbin, line 354: This will reset the state of our xmlrpc cache, once.

warehouse/config.py
```diff
@@ -304,15 +318,15 @@ def configure(settings=None):
     )
     maybe_set(settings, "warehouse.downloads_table", "WAREHOUSE_DOWNLOADS_TABLE")
     maybe_set(settings, "celery.broker_url", "BROKER_URL")
-    maybe_set(settings, "celery.result_url", "REDIS_URL")
-    maybe_set(settings, "celery.scheduler_url", "REDIS_URL")
-    maybe_set(settings, "oidc.jwk_cache_url", "REDIS_URL")
+    maybe_set_redis(settings, "celery.result_url", "REDIS_URL", db=12)
```
ewdurbin, line 321: We don't do anything of interest with celery results, so this should not have any impact.

warehouse/config.py
```diff
@@ -304,15 +318,15 @@ def configure(settings=None):
...
+    maybe_set_redis(settings, "celery.scheduler_url", "REDIS_URL", db=0)
```
ewdurbin, line 322: This is a no-op, but makes the db index for the scheduler explicitly 0.

warehouse/config.py
```diff
@@ -304,15 +318,15 @@ def configure(settings=None):
...
+    maybe_set_redis(settings, "oidc.jwk_cache_url", "REDIS_URL", db=1)
```
ewdurbin, line 323: This will reset our cache state for JWKs in Trusted Publishers, once.

miketheman: I _think_ this is the code that will auto-fill the cache on the next invocation, so probably pretty safe.
https://github.com/pypi/warehouse/blob/33049d18e3c37d7b09644b11b0f1e65582deed56/warehouse/oidc/services.py#L215-L217

warehouse/config.py
miketheman: Should we create a maintenance page to inform folks, or Just Do It?

tests/unit/test_config.py
```diff
@@ -141,6 +141,109 @@ def test_maybe_set_compound(monkeypatch, environ, base, name, envvar, expected):
...
+            "REDIS_URL",
+            str,
+            None,
+            9,
+            {"test.foo": "rediss://foo:bar@example.com:6379/9?fizz=buzz&wu=tang"},
+        ),
+    ],
+)
+def test_maybe_set_redis(
+    monkeypatch, environ, name, envvar, coercer, default, db, expected
```
miketheman, line 236: optimization: If `name` can never be `None`, simplify the test cases by one.
`name` is always `"test.foo"`, pass it explicitly in to `config.maybe_set_redis()`
Same with passing the same `envvar` in each test case.

Not a sticking point by any means.

miketheman (approved):
Good stuff here. I dropped in some comments, questions et al, nothing that should block this from going out.

warehouse/config.py
ewdurbin: aye it will.

tests/unit/test_config.py
ewdurbin: aacb95a4a

warehouse/config.py
```diff
@@ -238,6 +239,19 @@ def maybe_set_compound(settings, base, name, envvar):
             settings[".".join([base, key])] = value
 
 
+def maybe_set_redis(settings, name, envvar, coercer=None, default=None, db=None):
+    if envvar in os.environ:
+        value = os.environ[envvar]
+        if coercer is not None:
+            value = coercer(value)
+        parsed_url = urlparse(value)  # noqa: WH001
```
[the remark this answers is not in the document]
ewdurbin: 36179efa0

warehouse/config.py
ewdurbin: in my head, I was going to put "caches" in the 1-9 and celery junk in 10-12. I'll add a note.

ewdurbin: 22f0ba2d0

warehouse/config.py
ewdurbin: Yeah, plan to address that in the SQS->Redis part of the work. but good note!

warehouse/config.py
ewdurbin: Just Do It. 🤷🏼

PR #16698, opened 2024-09-12
oidc: add missing claims check in publisher lookup

facutuesca:
When minting a token using Trusted Publishing:
1. We try to match the OIDC claims to an existing publisher
2. Once we match it, we check that all the required claims for that publisher are present

These two steps match these two lines in the codebase: [link](https://github.com/pypi/warehouse/blob/af9e88a9fb48b6cc7260ed66675a8026748527b7/warehouse/oidc/services.py#L323-L327)

However, during the first step (finding a publisher), we read some of the claims in the token. Since the claims we read are supposed to be always present, we assumed that they existed even though the actual check for their existence happened afterwards, in step 2.

This means that if the OIDC provider generates a token without the expected claims (due to a bug, or an unexpected change), we fail with an unhandled exception. This recently happened with a GitLab OIDC token missing the `ci_config_ref_uri` claim, which according to their [docs](https://docs.gitlab.com/ee/ci/secrets/id_token_authentication.html#token-payload:~:text=in%20GitLab%2016.0.-,ci_config_ref_uri,-Always) should always be present.

This PR moves the check for the required claims to before publisher lookup.

cc @woodruffw


di:
Later in the flow we verify that all required claims are present before we verify the claims themselves: https://github.com/pypi/warehouse/blob/e900d74d9b6ccd96973c1beeafdd976bf650b877/warehouse/oidc/models/_core.py#L230-L247

Is there any reason not to just move this check higher? Why are we only checking for the subset that is required for lookup? Seems like a) this sort of duplicates the set of required claims, and b) makes the second check somewhat redundant.

facutuesca:
> Is there any reason not to just move this check higher? 

I think that makes sense. We can move the first part of `verify_claims()` which checks for missing and unaccounted claims ([link](https://github.com/pypi/warehouse/blob/e900d74d9b6ccd96973c1beeafdd976bf650b877/warehouse/oidc/models/_core.py#L211-L247)), to before we do the publisher lookup here: [link](https://github.com/pypi/warehouse/blob/af9e88a9fb48b6cc7260ed66675a8026748527b7/warehouse/oidc/utils.py#L92)

Since the sets of claims related to a publisher are class variables, we can run the checks before we find the specific publisher instance. 

The only downside I can think of is that we'll be separating the claims verification logic to two different places: one checking for existence and/or extra claims, and the other verifying the claims themselves. But that should be fine I think.

WDYT? (cc @woodruffw )

di:
I think that would be a little more maintainable: required claims would just live in one place, and we'd always check for all of them.

I think it should be OK to separate "claim presence verification" from "claim verification".

woodruffw:
> I think it should be OK to separate "claim presence verification" from "claim verification".

Yep, strong agree!

facutuesca:
@di @woodruffw Done. Now the check to see if all required claims exist (and unexpected claims are present) is done before publisher lookup. I also made it a class method rather than a free function.

I also adapted the tests to account for the fact that the existence check moved from `OIDCPublisher.verify_claims()` to `OIDCPublisher.check_claims_existence()`.

di (approved):
Much cleaner, thanks!

PR #16701, opened 2024-09-12
refactor: extract validator from inline

miketheman:
This allows us to prevent null bytes prior to checking the DB records.

Follows similar refactor in #16599


di (approved):
Maybe we should add a comment somewhere explaining why we're doing it this way for historical context? Or add a test that would prevent a regression.

miketheman:
Closing, as this is no longer necessary now that the custom WarehouseForm subclass has been removed.

PR #16703, opened 2024-09-12
Refactor `_is_valid_dist_file`

di:
Originally planned to just fix this to catch a `BadZipFile` exception that we were missing (https://python-software-foundation.sentry.io/share/issue/12f7133d26bc4d7f8ff634338270bd22/), but took the opportunity to remove some duplication as well while I was at it.


warehouse/forklift/legacy.py
```diff
@@ -260,85 +260,66 @@ def _is_valid_dist_file(filename, filetype):
...
-            decompressed_size > COMPRESSION_RATIO_MIN_SIZE
-            and decompressed_size / compressed_size > COMPRESSION_RATIO_THRESHOLD
-        ):
-            sentry_sdk.capture_message(
-                f"File {filename} ({filetype}) exceeds compression ratio "
-                f"of {COMPRESSION_RATIO_THRESHOLD} "
-                f"({decompressed_size}/{compressed_size})"
-            )
-            return False
+    if filename.endswith((".zip", ".whl")):
```
miketheman, line 263: This change replaces the original `is_zipfile()` check prior to running other operations with a string comparison.
Would it make sense to restore it to bail early, is is that check "heavy" compared to the try/except we have here?

```suggestion
    if filename.endswith((".zip", ".whl")) and zipfile.is_zipfile(filename):
```

warehouse/forklift/legacy.py
```diff
@@ -260,85 +260,66 @@ def _is_valid_dist_file(filename, filetype):
...
-        # PKG-INFO file.
-        try:
-            with zipfile.ZipFile(filename, "r") as zfp:
-                for zipname in zfp.namelist():
-                    parts = os.path.split(zipname)
-                    if len(parts) == 2 and parts[1] == "PKG-INFO":
-                        # We need the no branch below to work around a bug in
-                        # coverage.py where it's detecting a missed branch
-                        # where there isn't one.
-                        break
```
miketheman, line 321: This comment is no longer, has the bug in coverage.py been fixed and we can remove a "no branch" (not sure what its referring to, is the comment referring to something that was there before?)

miketheman (approved):
Generally good, had a couple of questions inline.

I was also looking at the final rendered function and wondered if there's a bit more optimization to be had there - specifically in `if filename.endswith((".zip", ".whl")):` branch of checks.
The process we have today is:

- check decompression
- check for compression types
- check for filenames

This hasn't changed with this refactor, it's only put them closer together, which is great!

Would it be faster to change the order so that we can exit earlier in the case of invalid filenames or compression types, so we don't have to decompress the entire package first? Something like:

- check for filenames
- check for compression types
- check decompression

In any case, this is an improvement on the previous, so yay!

warehouse/forklift/legacy.py
di: I have no idea what this comment is talking about, coverage seems fine with this section.

warehouse/forklift/legacy.py
di: There is a subtlety here: if only one of these conditionals is `False`, the section will be skipped. We need to check the value of `is_zipfile` separately and return if it is `False`: 3e3f0b2

warehouse/forklift/legacy.py
```diff
@@ -260,85 +260,70 @@ def _is_valid_dist_file(filename, filetype):
...
+                        break
+                    if filename.endswith(".whl") and tail == "WHEEL":
+                        break
+                else:
                     return False
 
-    if filename.endswith(".tar.gz"):
-        # TODO: Ideally Ensure the compression ratio is not absurd
-        # (decompression bomb), like we do for wheel/zip above.
+        except zipfile.BadZipFile:  # pragma: no cover
```
di, line 303: Flagging: I tried a few things to craft a zip file that passed `is_zipfile` but would raise this exception and couldn't. Keeping this here as a safeguard but skipping coverage checks for this line.

miketheman: Instead of the case of `is_zipfile()` check's `return False`, could you `raise zipfile.BadZipFile` instead, and have the `except` catch it?

di: Why? I don't really think it's a great idea to manually raise non-generic exceptions like that, what would that get us?

miketheman (approved):
Let's give it a shot!

PR #16709, opened 2024-09-13
packaging: add initial hints to storage services

woodruffw:
This is a purely hygenic/QC breakout from #16624 -- it adds some initial type hints to the `packaging` storage interfaces and services, to (hopefully) make it harder for people to pass `Path` and other non-`str` types into these APIs (which only take strings).


warehouse/packaging/interfaces.py
```diff
@@ -26,25 +26,25 @@ def create_service(context, request):
         created for, passing a name for settings.
         """
 
-    def get(path):
+    def get(path: str):
```
miketheman, line 29: While we're here, would it be beneficial to also implement return types?

miketheman (approved):

warehouse/packaging/interfaces.py
woodruffw: Probably yes, although I think there's significant variance between the individual implementations of this interface in terms of the actual "shape" of the return types here 😅 -- they're all vaguely file-like, but beyond that it's a bit mushy.

I could spend some time trying to sort that out, either with this PR or with a follow-up, whatever you'd prefer.

miketheman: Going to not merge today, so if you wanted to poke at this some more I can take a look again on Monday - just hit that "Re-request review" 🔄 button when ready.

di (approved):

di (approved):

PR #16724, opened 2024-09-16
test(functional): log in, change password

miketheman:


miketheman:
There's a few TODOs to add IDs to forms to make this test less brittle, and I'm still not convinced that this is a Good Idea, but with this test, functional test coverage rises to 52% (from 49%).

tests/functional/manage/test_views.py
```diff
@@ -48,6 +53,69 @@ def test_save_account(self, pyramid_services, user_service, db_request):
         assert user.name == "new name"
         assert user.public_email is None
 
+    def test_changing_password_succeeds(self, webtest, remote_addr, socket_enabled):
+        """If the user changes their password, their session should be invalidated."""
```
miketheman, line 57:
```suggestion
        """A user can log in, and change their password."""
```

tests/functional/manage/test_views.py
```diff
@@ -48,6 +53,69 @@ def test_save_account(self, pyramid_services, user_service, db_request):
...
+        totp_known_secret = b"\x453f47ec8c43d1f205f0a5244391342b428b3b06"
+        # create a User
+        user = UserFactory.create(
+            with_verified_primary_email=True,
+            password=(  # 'password'
+                "$argon2id$v=19$m=1024,"
+                "t=6,p=6$EiLE2Nsbo9S6N+acs/beGw$ccyZDCZstr1/+Y/1s3BVZ"
+                "HOJaqfBroT0JCieHug281c"
+            ),
```
miketheman, line 66: TODO: I think the UserFactory should be able to take in a string input and generate the appropriate hashed value

di: Something like `unhashed_password="password"` would be nice, but I think we could also just make this the default value.

miketheman: Added in 817819cad0fb405e25b86748e9e124ad2aa9a18a

tests/functional/manage/test_views.py
```diff
@@ -48,6 +53,62 @@ def test_save_account(self, pyramid_services, user_service, db_request):
...
+        # create a User
+        user = UserFactory.create(
+            with_verified_primary_email=True, clear_pwd="password"
+        )
+
+        # visit login page
+        login_page = webtest.get("/account/login/", status=HTTPStatus.OK)
+
+        # Fill & submit the login form
+        login_form = login_page.forms[2]  # TODO: form should have an ID, doesn't yet
```
miketheman, line 67: Add IDs to the existing forms and replace these with ID lookups

di (approved):

tests/functional/manage/test_views.py
```diff
@@ -48,6 +53,62 @@ def test_save_account(self, pyramid_services, user_service, db_request):
         assert user.name == "new name"
         assert user.public_email is None
 
+    def test_changing_password_succeeds(self, webtest, remote_addr, socket_enabled):
```
di, line 56: Is `remote_addr` needed here?

tests/functional/manage/test_views.py
di: https://github.com/pypi/warehouse/pull/16732/commits/17fd99028ee584718e3041d5faf86e1395054dfb

PR #16739, opened 2024-09-18
feat: capture metrics during email validation

miketheman:
Includes changes to enable `request.metrics.increment(...)` and coverage output.


di (approved):
I like it!

PR #16744, opened 2024-09-18
Migrate to redis as Celery broker.

ewdurbin:
Temporarily hard-codes a BROKER_URL in a separate worker process.

If not present, REDIS_URL is used for broker.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -5,3 +5,4 @@ web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicor
 worker: bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
 worker-beat: bin/start-worker celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
 worker-traced: env DD_SERVICE=warehouse-worker bin/start-worker ddtrace-run celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
+worker-drain-sqs: env BROKER_URL=sqs:///?region=us-east-2&queue_name_prefix=pypi-worker bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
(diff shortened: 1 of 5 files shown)


Procfile
```diff
@@ -5,3 +5,4 @@ web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicor
 worker: bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
 worker-beat: bin/start-worker celery -A warehouse beat -S redbeat.RedBeatScheduler -l info
 worker-traced: env DD_SERVICE=warehouse-worker bin/start-worker ddtrace-run celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
+worker-drain-sqs: env BROKER_URL=sqs:///?region=us-east-2&queue_name_prefix=pypi-worker bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
```
miketheman, line 8: question: will this value work without an actual SQS hostname?

warehouse/tasks.py
```diff
@@ -195,7 +195,10 @@ def includeme(config):
 
     broker_transport_options = {}
 
-    broker_url = s["celery.broker_url"]
+    broker_url = s.get("celery.broker_url")
+    if broker_url is None:
+        broker_url = s["celery.broker_redis_url"]
```
miketheman, line 200: I'm guessing that this will also be paired with removing `BROKER_URL` envvar from the runtime as part of deployment?

miketheman (approved):
Generally looks fine. See questions inline.

Procfile
ewdurbin: yep, that's the value we currently have set in Cabotage.

warehouse/tasks.py
ewdurbin: correct, it would need to be.

Issue #16743, opened 2024-09-18
Clickable elements in Project sidebar are easy to "mis click"


ewdurbin:
**Describe the bug**

Project URLs in the sidebar on pages like https://pypi.org/project/django-polymorphic/ are difficult to click due to overlapping hit boxes.

For instance, when cursoring up through the list attempting to click "Homepage" here, I accidentally hit "Download":

<img width="218" alt="Screenshot 2024-09-18 at 2 26 36 PM" src="https://github.com/user-attachments/assets/ef346040-85c8-4078-a1b5-11aa831e7fc1">

Additionally, _non clickable_ things like "License" and tags act as clickable when scrolling over them:

<img width="309" alt="Screenshot 2024-09-18 at 2 32 08 PM" src="https://github.com/user-attachments/assets/ba1117c5-26d3-4399-b430-77f36c699722">

**Expected behavior**

Links should be easy to click, things that are not links should not behave as links.

**To Reproduce**

Navigate to a project page and trial/inspect the elements to see how cursor behaves and which link is active.

**My Platform**

Noticed on Safari 18.0 on MacOS 15.0.

**Additional context**

May be related to #15737 or #16532.

di:
Looks like the "clickable"-ness is due to the `span` having the `vertical-tabs__tab` and `vertical-tabs__tab--condensed` classes, and the parent `ul` having the `vertical-tabs__list` class. Removing these will reintroduce the bullets as well, so we would need to suppress them to maintain the same view.

The spacing seems to be due to https://github.com/pypi/warehouse/blob/f3dcfcc44e2a23ef6498b5246ff47a5183f19afd/warehouse/static/sass/blocks/_verified.scss#L28

which is overridding: https://github.com/pypi/warehouse/blob/f3dcfcc44e2a23ef6498b5246ff47a5183f19afd/warehouse/static/sass/blocks/_vertical-tabs.scss#L107

However this was done so the checkboxes display inline with the link, so an alternative method to do that will need to be found.

di:
Also, it was noted that the checkmarks may be redundant given they are already within a "verified" section, so perhaps it makes the most sense to remove them.

ewdurbin:
[...]

PR #16745, opened 2024-09-18
address #16743

ewdurbin:
Two distinct commits for the issues noted in #16743.

1) explicitly removes extraneous padding on `a` elements in vertical-tabs elements
2) removes cursor/hover behavior for vertical-tabs__tab class, which is predominately used on a elements which have the desired behavior directly.

Screenshot 2024-09-18 at 3 23 52 PM

### DIFF
--- a/warehouse/static/sass/blocks/_verified.scss
+++ b/warehouse/static/sass/blocks/_verified.scss
@@ -26,6 +26,7 @@
   li {
     a {
         display: inline;
+        padding: 0;
     }
   }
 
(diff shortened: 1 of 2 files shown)


di:
> removes cursor/hover behavior for vertical-tabs__tab class, which is predominately used on a elements which have the desired behavior directly.

I think this probably breaks the effect on mobile where the entire element is clickable, not just the link?

ewdurbin:
> I think this probably breaks the effect on mobile where the entire element is clickable, not just the link?

Can you say where? I'm testing it with vertical tabs on my phone and not seeing a difference.

ewdurbin:
superseded by #16746

PR #16747, opened 2024-09-18
feat(admin): add prohibited email domains

### LINKED ISSUE #16135: Add Prohibited Domains list to Admin UI
Now that we not only rely on the disposable-email-domains, and have our own domain blocklist (see #15672) we should be able to display these and manage them in the Admin UI.

We likely need:

- new routes to display paginated entries, update/remove entries
- html templates
- a search function - or at least a DataTables setting

miketheman:
Resolves #16135


warehouse/admin/views/prohibited_email_domains.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@view_config(
+    route_name="admin.prohibited_email_domains.add",
+    permission=Permissions.AdminProhibitedEmailDomainsWrite,
+    request_method="POST",
+    uses_session=True,
+    require_methods=False,
+)
+def add_prohibited_email_domain(request):
```
di, line 65: Do we care if the domain has a protocol or not? Looks like this will accept either:

image

miketheman: Good call-out! We should remove some of the stuff we don't care about after it's been posted, so we only store parts of the input we care about and would compare against during validation.

warehouse/admin/views/prohibited_email_domains.py
```diff
@@ -0,0 +1,131 @@
...
+    # validate that the domain is valid
+    if not extract(email_domain).registered_domain:
+        request.session.flash(f"Invalid domain name '{email_domain}'", queue="error")
+        raise HTTPSeeOther(request.route_path("admin.prohibited_email_domains.list"))
+    # make sure we don't have a duplicate entry
+    if (
+        request.db.query(func.count(ProhibitedEmailDomain.id))
+        .filter(ProhibitedEmailDomain.domain == email_domain)
+        .scalar()
+        > 0
```
di, line 79: My understanding is something like this is generally preferred for testing presence of a row:

```suggestion
        request.db.session.query(
            exists().where(ProhibitedEmailDomain.domain == email_domain)
        ).scalar()
```
Is there any reason to prefer one over the other?

miketheman: Changed in 5e519f66b9802f0ef9e1f9183dcd44a9e890d2a5 - syntax is a little more explicit.

The real answer is that exists subquery is often faster than a count (aggregate) query - since exists stops at the first result found, and count wants to count all.

For small tables like this one, it's unlikely to ever grow to the scale of being a material issue, but thanks for calling it out!

warehouse/admin/views/prohibited_email_domains.py
miketheman: Updated in e60459158541b6538785fcc597a4f0106a493b45

di (approved):
Nice!

PR #16757, opened 2024-09-19
requirements: bump pypi-attestations to 0.0.12

woodruffw:
This bumps `pypi-attestations` to 0.0.12 to fix a bug we discovered in the previous series (https://github.com/trailofbits/pypi-attestations/pull/48). The TL;DR is that Pydantic's default Base64 encode/decode behavior isn't what we expected (it injects newlines into the encoded value), resulting in the generation of attestations that aren't consistent with PEP740. 

This could have been much worse had we begun to persist and serve attestations, but that fortunately hasn't happened yet 🙂.

See https://github.com/pypa/gh-action-pypi-publish/pull/262 for additional context.

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -65,7 +65,7 @@ rfc8785
 sentry-sdk
 setuptools
 sigstore~=3.2.0
-pypi-attestations==0.0.11
+pypi-attestations==0.0.12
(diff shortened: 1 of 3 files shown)


di (approved):

webknjaz:
xref https://github.com/pypa/gh-action-pypi-publish/issues/263

PR #16758, opened 2024-09-19
Move view deriver lower

di:
Left out of #16755.


di:
Put this in #16759 instead.

PR #16796, opened 2024-09-26
conftest: put transaction manager in its own fixture

woodruffw:
This moves the `tm` local from the `webtest` fixture into its own fixture.

This should have no impact on its own, but it will hopefully make it easier to tease apart whatever session/TM mismatch is causing CI failures like https://github.com/pypi/warehouse/actions/runs/11057423074/job/30721387457?pr=16778.

See: https://github.com/pypi/warehouse/issues/15634#issuecomment-2377980914


di (approved):

di:
@woodruffw Looks like this breaks some tests (probably because it's session-scoped).

tests/conftest.py
```diff
@@ -701,8 +701,21 @@ def xmlrpc(self, path, method, *args):
         return xmlrpc.client.loads(resp.body)
 
 
+@pytest.fixture(scope="session")
```
woodruffw, line 704:
```suggestion
@pytest.fixture
```

woodruffw:
> @woodruffw Looks like this breaks some tests (probably because it's session-scoped).

Whoops, fixed!

miketheman (approved):
I'm 👍 generally, but I'm curious about the timing of this - since the `tm.begin()/tm.doom()` calls now happen _after_ ` get_db_session_for_app_config()` is called - does that change the behavior at all?

woodruffw:
> I'm 👍 generally, but I'm curious about the timing of this - since the `tm.begin()/tm.doom()` calls now happen _after_ ` get_db_session_for_app_config()` is called - does that change the behavior at all?

I might not be following, but I think the order is correct here -- if my understanding of `pytest` fixture scoping is correct, the ordering here should be:

```
database (session-scoped)
  app_config_dbsession_from_end (session-scoped)
    tm (function-scoped)
      get_db_session_for_app_config -> _db-_session (context manager within function scope)
        actual test code
```

But it's possible I'm missing where the scope has changed 🙂

miketheman:
Maybe I didn't phrase the question well?
It's specifically that currently, `tm.begin()/tm.doom()` are explicitly called prior to calling  `get_db_session_for_app_config` and creating the TestApp, and with this change, the call to begin/doom happens later.

I don't _think_ it will matter, so let's ship this and find out.

woodruffw:
> It's specifically that currently, `tm.begin()/tm.doom()` are explicitly called prior to calling `get_db_session_for_app_config` and creating the TestApp, and with this change, the call to begin/doom happens later.

Ah, I see what you mesn! Yeah, I _think_ that won't matter, but that is indeed a slight difference.

PR #16812, opened 2024-09-30
implement some basic guards on our /search endpoint

ewdurbin:
- Move OpenSearch client timeout from 2000ms to 500ms
- Add a simple rate limiter to the endpoint by client ip

### DIFF
--- a/warehouse/views.py
+++ b/warehouse/views.py
@@ -26,6 +26,7 @@ from pyramid.httpexceptions import (
     HTTPRequestEntityTooLarge,
     HTTPSeeOther,
     HTTPServiceUnavailable,
+    HTTPTooManyRequests,
     exception_response,
 )
 from pyramid.i18n import make_localizer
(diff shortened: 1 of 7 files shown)


warehouse/views.py
```diff
@@ -322,8 +324,23 @@ def list_classifiers(request):
     has_translations=True,
 )
 def search(request):
+    ratelimiter = request.find_service(IRateLimiter, name="search", context=None)
     metrics = request.find_service(IMetricsService, context=None)
 
+    ratelimiter.hit(request.remote_addr)
+    if not ratelimiter.test(request.remote_addr):
+        metrics.increment("warehouse.search.ratelimiter.exceeded", tags=[])
```
miketheman, line 332: nit: do we need to expressly add tags as an empty list (here and in .hit below), or were we expecting some tags?

miketheman (approved):

warehouse/views.py
ewdurbin: that's good old fashioned copy-pasta, removed in a4b751cc1

PR #16813, opened 2024-09-30
packaging: deletion window machinery

woodruffw:
This adds `in_deletion_window` to the File, Release, and Project models. The three models currently compose in deletion semantics as follows:

1. A file is deletable iff it was uploaded within the last 7 days
2. A release is deletable iff all of its files are deletable
3. A project is deletable iff all of its releases are deletable

This results in no special casing: the deletability of releases and projects is entirely "driven" by the deletability of files, as files are the primary way in which users drive the creation of releases and projects.

This PR does **not** connect the deletion semantics to actual UI or view changes; I'll do that in a follow-up, to keep the patches small.

### DIFF
--- a/tests/unit/packaging/test_models.py
+++ b/tests/unit/packaging/test_models.py
@@ -11,7 +11,9 @@
 # limitations under the License.
 
 from collections import OrderedDict
+from datetime import datetime, timedelta
 
+import freezegun
 import pretend
 import pytest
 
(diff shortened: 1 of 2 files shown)


facutuesca:
Added some changes to reflect the [current PEP 763 draft](https://peps.python.org/pep-0763/). Specifically:
- Deletion window changed from <= 7 days to <3 days
- Pre-releases are always deletable

Relevant [section](https://peps.python.org/pep-0763/#deletion-eligibility-rules) from the PEP:
> -  A file is deletable if and only if it was uploaded to PyPI less than 72 hours from the current time, or if it has a [pre-release specifier](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers).
> - A release is deletable if and only if all of its contained files are deletable.
> - A project is deletable if and only if all of its releases are deletable.

PR #16814, opened 2024-09-30
feat: follow mx records back to pointers

miketheman:
In the event that the MX domain returned is not on the list,
but the IP address of the MX server is the same as another one on our
list, resolve those records back to their registered domains to compare.


warehouse/accounts/forms.py
```diff
@@ -301,6 +303,20 @@ def validate_email(self, field):
                 for _prio, mx_host in resp.mx
             }
 
+        # Resolve the returned MX domain's IP address to a PTR record, to a domain
+        all_mx_domains = set()
+        for mx_domain in mx_domains:
+            with contextlib.suppress(dns.resolver.NoAnswer, dns.resolver.NXDOMAIN):
```
ewdurbin, line 309: Will this cover slow/broken responses? Do we need to wrap this in some kind of timeout?

ewdurbin (approved):

warehouse/accounts/forms.py
miketheman: If I'm reading the source of dnspython correctly, when it's initialized it sets a default timeout of 2 seconds. https://github.com/rthalley/dnspython/blob/889385e71021f53e1dc4f117d92aed66bb90d087/dns/resolver.py#L958

PR #16817, opened 2024-10-01
Validate Alternate Repository Location URLS (PEP 708)

ewdurbin:
User behavior for this configuration has shown that most URLs supplied are not simple indexes at all, predominately they have been source repository links.

This is a bare-bones approach to validating that the URL is actually a simple index.

### DIFF
--- a/warehouse/manage/forms.py
+++ b/warehouse/manage/forms.py
@@ -14,6 +14,8 @@ import json
 
 import wtforms
 
+from urllib3 import PoolManager, Timeout
+
 import warehouse.utils.otp as otp
 import warehouse.utils.webauthn as webauthn
 
@@ -25,6 +27,7 @@ from warehouse.accounts.forms import (
     TOTPValueMixin,
     WebAuthnCredentialMixin,
 )
+from warehouse.constants import MIME_PYPI_SIMPLE_V1_ALL
 from warehouse.i18n import localize as _
 from warehouse.organizations.models import (
     OrganizationRoleType,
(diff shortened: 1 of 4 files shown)


warehouse/manage/forms.py
```diff
@@ -14,6 +14,8 @@
 
 import wtforms
 
+from urllib3 import PoolManager, Timeout
```
miketheman, line 17: **question:** ‏we use `requests` in other places in the code ([example](https://github.com/pypi/warehouse/blob/d502d5e233b7e51fad7bb161dc01d99455fec5c6/warehouse/oidc/forms/github.py#L71)), and this appears to be the first time we're using urllib3 directly. Is that to get both connect and read timeouts? If so, then we can pass those as a tuple to requests like so:

```
response = requests.head(field.data, timeout=(1.0, 1.0), headers=...)
```

But that apparently has the same affect as only specifying it once, since th value will be applied to both, so we could satisfy with

```
response = requests.head(field.data, timeout=1.0, headers=...)
```
https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

I'm also wondering why we aren't using `request.http.head(...)` from https://github.com/pypi/warehouse/blob/1fbb4ac752e68b5840b9e09b68e44a165569bfa6/warehouse/http.py#L48-L53 more - is it that the forms don't have the `request` object available to them?

warehouse/manage/forms.py
```diff
@@ -713,6 +716,23 @@ class CreateTeamForm(SaveTeamForm):
...
+        response = http.request(
+            "HEAD", field.data, headers={"Accept": ", ".join(MIME_PYPI_SIMPLE_V1_ALL)}
+        )
+        if response.headers.get("Content-Type") not in MIME_PYPI_SIMPLE_V1_ALL:
+            raise wtforms.validators.ValidationError(
+                _("Unable to parse simple index at given url")
+            )
+    except Exception as exc:
+        raise wtforms.validators.ValidationError(
+            _(f"Unable to parse simple index at given url: {exc}")
```
miketheman, line 732:
```suggestion
            _(f"The URL provided does not respond as a Simple repository: {exc}")
```

And maybe even tack on a URL? https://packaging.python.org/specifications/simple-repository-api/

warehouse/manage/views/__init__.py
```diff
@@ -1129,7 +1129,7 @@ def __init__(self, project, request):
         self.add_alternate_repository_form_class = AddAlternateRepositoryForm
 
     @view_config(request_method="GET")
-    def manage_project_settings(self):
+    def manage_project_settings(self, add_alternate_repository_form=None):
```
miketheman, line 1132: **question:** is this because we're using two methods for GET vs POST, and how to make sure that the form values aren't reset on validation failure? I don't know if I've seen this pattern before, and wanted to understand it more.‏

ewdurbin:
PEP 708 was rejected

PR #16877, opened 2024-10-12
allow uploads of wheels targeting macosx 15

LeoHsiao1:
Hi!
I can upload wheel packages for macosx_14_0 but not for macosx_15_0:
```sh
$ twine upload -u __token__ dist/*
Uploading pyexiv2-2.15.2-cp310-none-macosx_14_0_arm64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.4/1.4 MB • 00:02 • 716.4 kB/s
Uploading pyexiv2-2.15.2-cp310-none-macosx_15_0_arm64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.4/1.4 MB • 00:01 • 762.3 kB/s
WARNING  Error during upload. Retry with the --verbose option for more details.
         Binary wheel 'pyexiv2-2.15.2-cp310-none-macosx_15_0_arm64.whl' has an unsupported platform tag 'macosx_15_0_arm64'.
```
I spent half an hour checking to see if I'd made a mistake with a letter.
It turns out that pypi.org has a whitelist of MacOS version numbers.
https://github.com/pypi/warehouse/blob/10570d141ceb7377b374b425f80a4bf284df126c/warehouse/forklift/legacy.py#L145-L151
So I added the number 15.


LeoHsiao1:
Even with this one change, there may be other code that prevents users on MacOS15 computers from using pypi.org. I'll test it.

miketheman:
Generally find, but the test case here needs to be updated to macOS 16, since we don't support that yet.
https://github.com/pypi/warehouse/blob/cafec725bf70a9907055123906bee50d82f0a6c2/tests/unit/forklift/test_legacy.py#L3107

LeoHsiao1:
I got it. I mimicked this previous submit:
https://github.com/pypi/warehouse/commit/4ab1ee18d56f68011e8545e6794fbe9ceb50f6a8#diff-16f9b38ea1f1c507586a6c911d1fe76d163560c90da80caa112474bffe02db6aR3143

PR #16891, opened 2024-10-15
Update to Ubuntu 24.04 for readthedocs

radarhere:
https://github.com/pypi/warehouse/blob/10570d141ceb7377b374b425f80a4bf284df126c/.readthedocs.yml#L2-L9

https://docs.readthedocs.io/en/stable/config-file/v2.html mentions that 'ubuntu-24.04' can now be used. However, perhaps better than that is 'ubuntu-lts-latest', to automatically stay on the latest available OS.


miketheman (changes requested):
Thanks for the suggestion.

We'd prefer to have the underlying version "known" and using the lts version means that it could swap out underneath us.

miketheman (approved):
Thanks!

PR #16899, opened 2024-10-16
Ignore case when verifying GitLab/GitHub URLs

### LINKED ISSUE #16666: Verified GitHub links should probably ignore case for orgs
**Describe the bug**

For project, [meltanolabs-tap-postgres](https://pypi.org/project/meltanolabs-tap-postgres/), the homepage link points to https://github.com/meltanolabs/tap-postgres (org in lower case), but the trusted publisher was created with `MeltanoLabs/tap-postgres` (org in upper case).

The result is that the links to the GitHub repo are not "verified".

**Expected behavior**

Links to the GitHub repo are verified when using trusted publishers even when the GitHub org doesn't match the case.

**To Reproduce**

NA, see description of the problem above.

**My Platform**

**Additional context**

The easy workaround on our end is simply to use consistent casing throughout, but I though I'd still log in case y'all think this is a bug.

DarkaMaul:
This PR fixes the bug reported in #16666 ~by adding a new field in `OIDCPublisherMixin` regarding the case-sensitivity of URL verification~.

Both GitLab and GitHub works with case-insensitive slugs but ActiveState does not [1].

[1]: https://platform.activestate.com/DarkaMaul-org/Python-3.12.6-Linux-Ubuntu

/cc @facutuesca @woodruffw


warehouse/oidc/models/github.py
```diff
@@ -356,13 +359,18 @@ def verify_url(self, url: str):
         The suffix `.git` in repo URLs is ignored, since `github.com/org/repo.git`
         always redirects to `github.com/org/repo`. This does not apply to subpaths,
         like `github.com/org/repo.git/issues`, which do not redirect to the correct URL.
+
+        GitHub uses case-insensitive owner/repo slugs - so we perform a case-insensitive
+        comparison.
         """
+        url = url.lower()
```
DarkaMaul, line 366: FLAG: This is a (small) over-approximation because `https://github.com/pypi/warehouse.GIT` would also be verified, while the URL cannot be resolved on GitHub.

warehouse/oidc/models/_core.py
```diff
@@ -173,6 +173,12 @@ class OIDCPublisherMixin:
     # even have an "equivalent" column.
     __lookup_strategies__: list = []
 
+    # URL Verification Case Sensitivity
+    # Determines whether the user URLs should be verified in a case-sensitive
+    # manner. By default, the URLs are treated as case-sensitive. This behavior
+    # can be overridden in individual Publisher definitions.
+    url_verification_is_case_sensitive: bool = True
```
woodruffw, line 180: I'm _okay_ with this approach, although I wish we didn't have to keep adding class-level state to these notionally OIDC-only models 😅 

As one suggestion: maybe we could dunder it like the other class variables, e.g.:

```suggestion
    __url_verification_is_case_sensitive__: bool = True
```

warehouse/oidc/models/github.py
facutuesca: I don't think we need to call `url.lower()` here, since `super().verify_url(url)` will also call `lower()` on the URL. Removing that call would also mean that we won't verify invalid URLs ending in `.GIT`.

I think adding the `lower()` call below, in 
```python
        user_uri = rfc3986.api.uri_reference(url.lower()).normalize()
```

should work

warehouse/oidc/models/_core.py
facutuesca: Maybe a bit of a nitpick, but I wonder if the name `url_verification_is_case_sensitive` isn't a bit too broad: The case-sensitive part is only about the `org/repo` part of the URL. Maybe we could call it something like `repository_is_case_sensitive`? (since `self.repository == 'owner/repo'`)

woodruffw: I think that's also unfortunately confusing, since not all TP providers have a notion of "repository."

Thinking about it some more, I think perhaps we shouldn't try to build an abstraction here: the GitHub and GitLab publishers should just do case normalization internally.

warehouse/oidc/models/github.py
DarkaMaul: Fixed in the new version

warehouse/oidc/models/_core.py
DarkaMaul: > Thinking about it some more, I think perhaps we shouldn't try to build an abstraction here: the GitHub and GitLab publishers should just do case normalization internally.

I think this makes sense here - I've updated the code to handle the logic (albeit) similar in the respective publishers.

warehouse/oidc/models/gitlab.py
```diff
@@ -290,6 +290,12 @@ def verify_url(self, url: str):
         `gitlab.com/org/repo`. This does not apply to subpaths like
         `gitlab.com/org/repo.git/issues`, which do not redirect to the correct URL.
         """
+        normalized_url_prefixes = (self.publisher_base_url.lower(),)
+        for prefix in normalized_url_prefixes:
+            if url.lower().startswith(prefix):
+                url = prefix + url[len(prefix) :]
+                break
+
```
facutuesca, line 298:
```suggestion
        lowercase_base_url = self.publisher_base_url.lower()
        if url.lower().startswith(lowercase_base_url):
            url = lowercase_base_url + url[len(lowercase_base_url) :]

```

warehouse/oidc/models/gitlab.py
```diff
@@ -290,6 +290,10 @@ def verify_url(self, url: str):
         `gitlab.com/org/repo`. This does not apply to subpaths like
         `gitlab.com/org/repo.git/issues`, which do not redirect to the correct URL.
         """
```
facutuesca, line 292:
```suggestion
        `gitlab.com/org/repo.git/issues`, which do not redirect to the correct URL.
        
        GitLab uses case-insensitive owner/repo slugs - so we perform a case-insensitive
        comparison.
        """
```

di (approved):

miketheman:
I think a missed condition of this fix in the test cases is where the **configured** GitHub `publisher_base_url` is also case-sensitive, since that happens too, and the URLs are not verified as a result - I couldn't find that expressed as a test case, this diff kinda exposes it:

```diff
diff --git a/tests/unit/oidc/models/test_github.py b/tests/unit/oidc/models/test_github.py
index fa08b4da7..a2c66a509 100644
--- a/tests/unit/oidc/models/test_github.py
+++ b/tests/unit/oidc/models/test_github.py
@@ -624,6 +624,20 @@ class TestGitHubPublisher:
         with pytest.raises(sqlalchemy.exc.IntegrityError):
             db_request.db.commit()

+    @pytest.mark.parametrize(
+        "repository_name",
+        [
+            "repository_name",
+            "Repository_Name",  # currently fails
+        ],
+    )
+    @pytest.mark.parametrize(
+        "repository_owner",
+        [
+            "repository_owner",
+            "Repository_Owner",  # currently fails
+        ],
+    )
     @pytest.mark.parametrize(
         ("url", "expected"),
         [
@@ -640,10 +654,10 @@ class TestGitHubPublisher:
             ("https://repository_owner.github.io/RePoSiToRy_NaMe/subpage", True),
         ],
     )
-    def test_github_publisher_verify_url(self, url, expected):
+    def test_github_publisher_verify_url(self, url, expected, repository_name, repository_owner):
         publisher = github.GitHubPublisher(
-            repository_name="repository_name",
-            repository_owner="repository_owner",
+            repository_name=repository_name,
+            repository_owner=repository_owner,
             repository_owner_id="666",
             workflow_filename="workflow_filename.yml",
             environment="",
```

Is this check expressed somewhere else? From reading the code, the [parent class' verification won't pass](https://github.com/pypi/warehouse/blob/b837b1d4999ab195bbec7740c2fab809df496338/warehouse/oidc/models/github.py#L340), resulting in running https://github.com/pypi/warehouse/blob/b837b1d4999ab195bbec7740c2fab809df496338/warehouse/oidc/models/github.py#L343C16-L343C41 which also won't pass.

Issue #16753, opened 2024-09-19
Verify GitLab pages URLs


wangenau:
**What's the problem this feature will solve?**
After playing around with the new verified sections through trusted publishing, I noticed that GitLab pages will not be marked as verified.
According to the documentation, this is the case for GitHub, but not GitLab: https://docs.pypi.org/project_metadata/#verified-details
Since documentations are often uploaded to GitLab pages, this will create an unpleasant split between verified details (i.e., the repository URL) and unverified URLs (i.e., the documentation URL).

**Describe the solution you'd like**
Similar to GitHub, the GitLab pages should get verified as well. Similar to the documentation, these would need to be added:
* `https://pypa.gitlab.io/pip`
* `https://pypa.gitlab.io/pip/*` (all subpaths)

PR #16918, opened 2024-10-18
Verify GitLab URLs

DarkaMaul:
Fixes https://github.com/pypi/warehouse/issues/16753

This implements a verification for GitLab page URLs.

/cc @facutuesca @woodruffw


warehouse/oidc/models/_core.py
```diff
@@ -102,6 +102,39 @@ def check_existing_jti(
     return True
 
 
+def verify_url_from_reference(*, reference_url: str, url: str) -> bool:
```
woodruffw, line 105: Nitpick: I think ideally this would live somewhere else, since it doesn't have much to do with the core OIDC models. But I don't have any good ideas about where it should live...

woodruffw (approved):

warehouse/oidc/models/_core.py
di: Agreed. At the very least, let's move it to `warehouse/oidc/utils.py`.

DarkaMaul: Moving it to `utils` led to a circular import but I created a new `urls.py` file for the function.

di:
Sorry @DarkaMaul, we have merge conflicts here now after merging #16899.

DarkaMaul:
> Sorry @DarkaMaul, we have merge conflicts here now after merging #16899.

We should be good now

di:
@DarkaMaul There are linting failures now FYI!

di (approved):

PR #16921, opened 2024-10-18
test/oidc: rename TestPublisher, mark as abstract

woodruffw:
Locally, this appears to suppress the warnings that @miketheman noticed in CI. Let's see what happens in CI now 🙂


woodruffw:
Confirmed that the OIDC warnings are no longer present on CI as well.

miketheman (approved):
Excellent, thanks! Will likely merge next week.

PR #16922, opened 2024-10-18
refactor: replace html5lib with lxml

miketheman:
The only place we use `html5lib` is in `camoify()`, and `lxml` has built-in support for HTML5 now.

Refs: https://lxml.de/html5parser.html

### DIFF
--- a/requirements/lint.in
+++ b/requirements/lint.in
@@ -14,7 +14,6 @@ types-babel
 boto3-stubs
 types-certifi
 types-first
-types-html5lib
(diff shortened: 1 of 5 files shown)


miketheman:
LOL whoops. Turns out that lxml.html.html5parser actually uses html5lib - that wasn't apparent from the docs! Oh well 😉

miketheman:
We _could_ feasibly not use the html5parser, and use the regular html parser...

PR #16933, opened 2024-10-21
feat: add tags to conversations on removal

miketheman:
Create `add_tag` to set a tag to a Help Scout conversation.

We can use the added tag to trigger automated workflows.

### DIFF
--- a/warehouse/helpdesk/services.py
+++ b/warehouse/helpdesk/services.py
@@ -50,6 +50,10 @@ class ConsoleHelpDeskService:
         print(dedent(pprint.pformat(request_json)))
         return "localhost"
 
+    def add_tag(self, *, conversation_url: str, tag: str) -> None:
+        print(f"Adding tag '{tag}' to conversation '{conversation_url}'")
+        return
+
 
 @implementer(IHelpDeskService)
 class HelpScoutService:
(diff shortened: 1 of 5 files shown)


ewdurbin (approved):

warehouse/helpdesk/services.py
```diff
@@ -120,3 +124,32 @@ def create_conversation(self, *, request_json: dict) -> str:
...
+        tags.append(tag)
+
+        resp = self.http.put(
+            f"{conversation_url}/tags",
+            headers={"Authorization": f"Bearer {self.bearer_token}"},
+            json={"tags": tags},
+            timeout=10,
+        )
+        resp.raise_for_status()
+        return
```
di, line 155: Driveby comment, but based on my read of https://api.helpdesk.com/docs#operation/ticketUpdateAddTag it looks like this could just be:

```suggestion
        resp = self.http.put(
            f"{conversation_url}/tags",
            headers={"Authorization": f"Bearer {self.bearer_token}"},
            json={"tags": [tag]},
            timeout=10,
        )
        resp.raise_for_status()
        return
````

I suspect the API doesn't drop tags that already exist but aren't included in the request.

miketheman: @di Thanks, but that's the API docs for Help Desk, not Help Scout.

PR #16934, opened 2024-10-21
attestations: provenance added metric

woodruffw:
Adds `warehouse.attestations.build_provenance.ok`
to signal when a provenance object is persisted to DB.

### DIFF
--- a/warehouse/attestations/services.py
+++ b/warehouse/attestations/services.py
@@ -242,4 +242,6 @@ class IntegrityService:
     def build_provenance(
         self, request: Request, file: File, attestations: list[Attestation]
     ) -> DatabaseProvenance:
-        return _build_provenance(request, file, attestations)
+        provenance = _build_provenance(request, file, attestations)
+        self.metrics.increment("warehouse.attestations.build_provenance.ok")
+        return provenance
--- a/tests/unit/attestations/test_services.py
+++ b/tests/unit/attestations/test_services.py
@@ -261,10 +261,10 @@ class TestIntegrityService:
         assert attestations == [dummy_attestation]
 
     def test_build_provenance_fails_unsupported_publisher(
-        self, db_request, dummy_attestation
+        self, metrics, db_request, dummy_attestation
     ):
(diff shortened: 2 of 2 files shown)


warehouse/attestations/services.py
```diff
@@ -242,4 +242,6 @@ def parse_attestations(
     def build_provenance(
         self, request: Request, file: File, attestations: list[Attestation]
     ) -> DatabaseProvenance:
-        return _build_provenance(request, file, attestations)
+        provenance = _build_provenance(request, file, attestations)
+        self.metrics.increment("warehouse.attestations.build_provenance.ok")
```
miketheman, line 246: Would we also want to monitor how many attestations are created? A gauge might work better here https://docs.datadoghq.com/metrics/types/?tab=gauge#definition

```suggestion
        self.metrics.gauge("warehouse.attestations.build_provenance.ok", len(attestations))
```

Depending on what it is we actually want to chart out.

woodruffw: I _think_ a gauge won't reveal anything super relevant here, since the attestation set is fixed at `1` for now (and when we allow more, we don't expect more than a small handful per file).

ewdurbin: Is this metric hit when attestations are requested? If so I read this as surfacing how popular the feature is, so yeah an increment makes sense?

If so it will still be unreliable since CDN should cache the request.

If we want a closer look at the volume of requests for provenance/attestations we would need to to add it to our edge metrics with linehaul, which has the added benefit of showing us aggregate data on user agent and geographic region.

woodruffw: It's hit when attestations are uploaded (i.e. in the HTTP POST), so I think CDN/caching won't be an issue!

(The goal with this metric is to measure adoption of attestations relative to trusted publishing, so it's focused entirely on the upload side and not the request side.)

Joj501:
Approve

ewdurbin (approved):

PR #16935, opened 2024-10-21
oidc: move reusable_worfklow_used field to the correct event

woodruffw:
This _should_ fix the `reusable_workflow_used` data quality issues we're seeing, which stem from us originally trying to get it at `file_upload` time, where the request has only a limited view into the original minted token's OIDC claims.

We instead now collect this field during `mint_token`, where the full set of OIDC claims is available to us.

h/t @miketheman for pointing this out, cc @facutuesca

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -56,7 +56,6 @@ from warehouse.forklift import metadata
 from warehouse.forklift.forms import UploadForm, _filetype_extension_mapping
 from warehouse.macaroons.models import Macaroon
 from warehouse.metrics import IMetricsService
-from warehouse.oidc.views import is_from_reusable_workflow
 from warehouse.packaging.interfaces import IFileStorage, IProjectService
 from warehouse.packaging.metadata_verification import verify_email, verify_url
 from warehouse.packaging.models import (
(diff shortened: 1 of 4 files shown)


warehouse/forklift/legacy.py
```diff
@@ -897,13 +897,13 @@ def file_upload(request):
...
-                    if request.oidc_publisher
-                    else False
-                ),
+                # NOTE(ww, 2024-10-21): This unfortunately typo'd event field
+                # was always incorrect, since the `file_upload` endpoint isn't
+                # itself aware of the kind of workflow that minted its API
+                # token. The correctly named datapoint is now under the
+                # `account:short_lived_api_token:added` event which, despite
+                # its name, is actually a project event.
+                "reusable_worfklow_used": False,
```
woodruffw, line 906: NB: I _think_ we could also just remove this outright: there should be no dependencies on this event field at all, since we added it solely to collect figures on reusable workflow use.

miketheman: I'd prefer removal if we considered it transient only, less context to carry forward.

miketheman (commented):
Generaly fine, and I agree on the removal front.

warehouse/forklift/legacy.py
woodruffw: Makes sense, removing!

facutuesca:
also LGTM

miketheman (approved):

PR #16949, opened 2024-10-22
Implement PEP 639, Metadata 2.4

### LINKED ISSUE #16620: Implement PEP 639
Standardized licensing metadata will in time unify the currently largely chaotic landscape.

See: https://peps.python.org/pep-0639
PEP 639 has been provisionally accepted with one of the conditions being implementation in PyPI.

ewdurbin:
Closes #16620

~~Depends on merge/release of https://github.com/pypa/packaging/pull/828~~
~~TODO: Move to release of packaging with license-expression parser~~


warehouse/forklift/legacy.py
```diff
@@ -1097,6 +1121,30 @@ def file_upload(request):
...
+                """
+                license_files = copy.deepcopy(meta.license_files)
+                with tarfile.open(temporary_filename, "r:gz") as tar:
+                    # Already validated as a tarfile by _is_valid_dist_file above
+                    member = tar.next()
+                    while member:
+                        if member.name in license_files:
+                            license_files.remove(member.name)
+                        member = tar.next()
+                if license_files != []:
```
miketheman, line 1141: I'm wondering if [`getmember()`](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.getmember) might work here? Something like:

```python
with tarfile.open(temporary_filename, "r:gz") as tar:
    for license_file_path in license_files:
        try:
            tar.getmember(license_file_path)
        except KeyError:
             ...
```

This operation also has to scan the entire archive, so I don't know if this is particularly beneficial or not, similar to how you've implemented it in zipfile.

miketheman (commented):
Looking cool, and coming along nicely! I commented a bit inside.

One thing I'm curious about - at what point should we be looking to consolidate any file operations where we pop open the file and read stuff from it, so we only have to open/iterate once?

warehouse/packaging/models.py
```diff
@@ -587,6 +589,14 @@ def __table_args__(cls):  # noqa
     home_page: Mapped[str | None]
     home_page_verified: Mapped[bool_false]
     license: Mapped[str | None]
+    license_expression: Mapped[str | None]
+    license_files: Mapped[list[str] | None] = mapped_column(
+        ARRAY(String),
+        comment=(
+            "Array of license files. Null indicates no License-File(s) were supplied "
```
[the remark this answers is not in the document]
ewdurbin: 01e71e471

warehouse/forklift/legacy.py
ewdurbin: ecda1a978502c028ec24a90fcb051b85c7222822

ewdurbin:
@befeleme would you be able to take a look at this to review the implementation details?

warehouse/templates/includes/packaging/project-data.html
```diff
@@ -199,7 +199,20 @@ <h6>{% trans %}Project links{% endtrans %}</h6>
 <div class="sidebar-section unverified">
   <h6>{% trans %}Meta{% endtrans %}</h6>
   <ul>
-  {% if license %}
+  {% if release.license_expression %}
+    <li>
+      <span>
+        <strong>{% trans %}License Expression:{% endtrans %}</strong> {{ release.license_expression }}
```
dnicolodi, line 205: I would use just "License" for the user visible string here.

befeleme:
> @befeleme would you be able to take a look at this to review the implementation details?

Yes, hopefully tomorrow.

befeleme:
After reading the code I've got just one question: I can't locate the validation of the `License-Expression` per https://peps.python.org/pep-0639/#add-license-expression-field ("[Python Package Index (PyPI) MUST validate that they contain a valid, case-normalized license expression with valid identifiers and MUST reject uploads that do not"). Or did I miss it?

ewdurbin:
> After reading the code I've got just one question: I can't locate the validation of the `License-Expression` per https://peps.python.org/pep-0639/#add-license-expression-field ("[Python Package Index (PyPI) MUST validate that they contain a valid, case-normalized license expression with valid identifiers and MUST reject uploads that do not"). Or did I miss it?

That's implemented upstream in pypa/packaging, see https://github.com/pypa/packaging/pull/828

befeleme:
Ah, looking at the whole `warehouse/forklift/metadata.py` I see `packaging` validations are applied automatically and no additional changes regarding that are needed here. Makes sense.

warehouse/migrations/versions/6ee23f5a6c1b_add_pep_639_columns.py
```diff
@@ -0,0 +1,124 @@
...
+
+from alembic import op
+from alembic_postgresql_enum import ColumnType, TableReference
+from sqlalchemy.dialects import postgresql
+
+revision = "6ee23f5a6c1b"
+down_revision = "2af8015830dd"
+
+
+def upgrade():
```
ewdurbin, line 30: I think this needs a closer look to try to avoid contention when rolled out.

warehouse/templates/includes/packaging/project-data.html
ewdurbin: I'm waiting for a more thorough review from @pypi/warehouse-committers before I address suggestions/questions.

miketheman (approved):
Yay, we're almost there! I didn't re-review a lot of what I'd already covered, so I focused on the migration and testing, and looks like everything generally conforms to the PEP.

I'm thinking that tracking usage/progress of this can be deferred, but that might be a nice metric to emit via DB query later - I can imagine wanting to see new projects using the syntax, older projects converting to the new syntax, etc - but definitely beyond scope of this changeset.

Approving to you can decide when you want to run the migration, but please include some timeouts based on some guesswork from TestPyPI

warehouse/forklift/legacy.py
```diff
@@ -1092,6 +1115,27 @@ def file_upload(request):
                     f"{str(version)!r}.",
                 )
 
+            filename = os.path.basename(temporary_filename)
+
+            if meta.license_files and packaging.version.Version(
+                meta.metadata_version
+            ) >= packaging.version.Version("2.4"):
```
di, line 1122: Seems like the check for `metadata_version` is redundant here (and elsewhere), I don't think `meta.license_files` will exist if the version is < 2.4:

```
>>> Metadata.from_raw({"metadata_version": "2.3", "license_files": ["something"], "version": "6.6.6", "name": "throwaway"})
  + Exception Group Traceback (most recent call last):
  |   File "<stdin>", line 1, in <module>
  |   File "/home/di/.pyenv/versions/3.12.7/lib/python3.12/site-packages/packaging/metadata.py", line 752, in from_raw
  |     raise ExceptionGroup("invalid metadata", exceptions)
  | ExceptionGroup: invalid metadata (1 sub-exception)
  +-+---------------- 1 ----------------
    | packaging.metadata.InvalidMetadata: license-file introduced in metadata version 2.4, not 2.3
    +------------------------------------

>>> Metadata.from_raw({"metadata_version": "2.3", "version": "6.6.6", "name": "throwaway"}).license_files is None
True
```

warehouse/templates/includes/packaging/project-data.html
di: Agreed, we should distinguish between the old "License" and the new "License-Expression" here.

warehouse/forklift/metadata.py
```diff
@@ -233,6 +233,19 @@ def _validate_metadata(metadata: Metadata, *, backfill: bool = False):
...
+        errors.append(
+            InvalidMetadata(
+                "license",
+                (
+                    "License is deprecated when License-Expression is present. "
+                    "Only License-Expression should be present."
+                ),
+            )
+        )
+
```
di, line 248: Seems like this is a check that should be happening in pypa/packaging instead, but isn't?

```
>>> Metadata.from_raw({"metadata_version": "2.4", "license": "MIT", "license_expression": "MIT", "version": "6.6.6", "name": "throwaway"})
<packaging.metadata.Metadata object at 0x7f44b9647680>
```

ewdurbin: That validation isn't backed into packaging as the [PEP](https://peps.python.org/pep-0639/#deprecate-license-field) says "tools MAY issue a warning" and "PyPI MUST reject"

warehouse/forklift/legacy.py
```diff
@@ -1041,6 +1043,27 @@ def file_upload(request):
...
+                with zipfile.ZipFile(temporary_filename) as zfp:
+                    for license_file in meta.license_files:
+                        try:
+                            zfp.read(license_file)
+                        except KeyError:
+                            raise _exc_with_message(
+                                HTTPBadRequest,
+                                f"License-File {license_file} does not exist in "
+                                f"distribution file {filename}",
+                            )
```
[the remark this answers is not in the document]
ewdurbin: It does. a8dd46958

warehouse/migrations/versions/6ee23f5a6c1b_add_pep_639_columns.py
```diff
@@ -0,0 +1,124 @@
...
+        sa.Column(
+            "license_files",
+            postgresql.ARRAY(sa.String()),
+            nullable=True,
+            comment=(
+                "Array of license filenames. "
+                "Null indicates no License-File(s) were supplied by the uploader."
+            ),
+        ),
+    )
```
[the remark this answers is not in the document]
ewdurbin: FWIW I think that the exclusive lock point is ultimately our "problem", since we perform reads with `READ COMMITTED` in a transaction.

warehouse/migrations/versions/6ee23f5a6c1b_add_pep_639_columns.py
```diff
@@ -0,0 +1,124 @@
...
+    op.sync_enum_values(
+        "public",
+        "release_dynamic_fields",
```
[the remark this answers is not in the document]
ewdurbin: 74b7b8cde

warehouse/forklift/metadata.py
ewdurbin: I'm not sure if I agree or disagree here, mainly because I am not certain where the responsibility belongs.

This specific check does ensure that PyPI's implementation meets the specification that

> For all newly-uploaded [distribution archive](https://packaging.python.org/en/latest/glossary/#term-Distribution-Archive)s that include a License-Expression field, the [Python Package Index (PyPI)](https://pypi.org/) MUST reject any that specify both License and License-Expression fields.

No matter the underlying validation, so I'm prone to remain explicit and keep it.

warehouse/forklift/legacy.py
ewdurbin: 8c2859ae686a674eab04b543899984395be4abed

ewdurbin:
Re @miketheman:

> I'm thinking that tracking usage/progress of this can be deferred, but that might be a nice metric to emit via DB query later - I can imagine wanting to see new projects using the syntax, older projects converting to the new syntax, etc - but definitely beyond scope of this changeset.

That is all technically captured in the immutable release metadata, so aside from watching a pot boil in real-time in datadog I agree adoption metrics are well beyond scope.

warehouse/forklift/metadata.py
di: I think we can leave this for now and figure out later if it should belong upstream!

di (approved):

ewdurbin:
Migrations are going out for this _ahead_ of merge during this window to avoid unnecessary disruption: http://status.python.org/incidents/q5q5347n3s2b

ofek:
Thanks everyone!

PR #16956, opened 2024-10-23
Guard our BigQuery release metadata sync against upcoming schema changes

ewdurbin:
PEP 639 implementation will need the following columns added to the big query schema:

license_expression, STRING, NULLABLE
license_files, STRING, REPEATED

We'll want that done in advance of merging an implementation, but need to have the Big Query public dataset team do that for us... so to avoid the need for coordination, lets guard ourselves against the schema changing underneath us...

### DIFF
--- a/tests/unit/packaging/test_tasks.py
+++ b/tests/unit/packaging/test_tasks.py
@@ -419,6 +419,8 @@ bq_schema = [
     SchemaField("maintainer", "STRING", "NULLABLE"),
     SchemaField("maintainer_email", "STRING", "NULLABLE"),
     SchemaField("license", "STRING", "NULLABLE"),
+    SchemaField("license_expression", "STRING", "NULLABLE"),
+    SchemaField("license_files", "STRING", "REPEATED"),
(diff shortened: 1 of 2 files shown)


tests/unit/packaging/test_tasks.py
```diff
@@ -624,6 +626,8 @@ def find_service(name=None):
                         "maintainer_email": form_factory["maintainer_email"].data
                         or None,
                         "license": form_factory["license"].data or None,
+                        "license_expression": None,
+                        "license_files": [],
```
miketheman, line 630: Should these have new entries added in `input_parameters` to feed the `form_factory` and follow the `... or None` pattern? (I haven't read this much yet, so am not certain of the test usage, but it seems like the other fields follow that pattern, so I noted it.)

miketheman (approved):
One question, take it or leave it

tests/unit/packaging/test_tasks.py
ewdurbin: For now no, the test ensure that the correct result comes out for BigQuery when no keys exist in the dist_metadata that comes in.

Issue #6688, opened 2019-09-22
Add links to corresponding resources into the email notification when receiving access to packages


webknjaz:
**What's the problem this feature will solve?**

I received this email recently:
```console
You have been added as {{role}} to the PyPI project {{project_name}} by {{other_user}}.

—
You are receiving this because you have been added by {{other_user}} to a project on PyPI.
```

But there's not a single link in it. So if I were to open a role explanation on the help page, the project page or the user page, I'd have to type/search/copy things, which is not a good UX.

**Describe the solution you'd like**

I just want to have clickable things in that email.

**Additional context**

Corresponding templates:
* https://github.com/pypa/warehouse/blob/97c4a5124e86119eabe57c7b896a34d2732dc1d0/warehouse/templates/email/added-as-collaborator/body.html
* https://github.com/pypa/warehouse/blob/97c4a5124e86119eabe57c7b896a34d2732dc1d0/warehouse/templates/email/added-as-collaborator/body.txt

miketheman:
cc @aram-cinnamon we chatted about this - let me know if you need more pointers!

CoderLogy:
i can do it using aherf like this 
`You have been added as <a href="https://pypi.org/help/#{{role}}">{{role}}</a> to the PyPI project <a href="https://pypi.org/project/{{project_name}}/">{{project_name}}</a> by <a href="https://pypi.org/user/{{other_user}}/">{{other_user}}</a>`

miketheman:
@CoderLogy We do not assign issues to folks outside the organization, but if you'd like to put up a PR to resolve, please do so!
Read the contributing details in https://warehouse.pypa.io/development/index.html
and specifically the notes about Emails: https://warehouse.pypa.io/development/email.html
and Translations: https://warehouse.pypa.io/development/submitting-patches.html#translations

CoderLogy:
> @CoderLogy We do not assign issues to folks outside the organization, but if you'd like to put up a PR to resolve, please do so!
> 
> Read the contributing details in https://warehouse.pypa.io/development/index.html
> 
> and specifically the notes about Emails: https://warehouse.pypa.io/development/email.html
> 
> and Translations: https://warehouse.pypa.io/development/submitting-patches.html#translations

How can I join this organization.

webknjaz:
@CoderLogy you don't have to. You can go ahead and open a PR right away if you're interested in working on this.

CoderLogy:
[...]

PR #16975, opened 2024-10-26
Issue 6688: Added project, role and initiator user URLs to email

VineetHBhat:
fixes: #6688
NOTE:
I'm working on a machine with 4GB RAM, and the `make serve` step failed due to timeout. However, since the patch only involved minor additions to Jinja2 templates, I proceeded with the submission despite not fully testing it. I understand this isn't ideal, but the fix appeared straightforward.


VineetHBhat:
I have executed `make reformat` and `make lint`. No changes were made/suggested by these.

webknjaz:
Looks like most of this was implemented by #19430 (short of linking the roles FAQ).

mashhoorahdal:
> Looks like most of this was implemented by #19430 (short of linking the roles FAQ).

Added roles faq url (https://github.com/pypi/warehouse/pull/19458)

PR #16996, opened 2024-10-29
chore(ci): address overly broad permissions scopes

miketheman:
- inform checkout to not persist credentials
- move permissions from workflow to job scope

Thanks, Dr. Zizmor!
Refs: https://github.com/woodruffw/zizmor

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,8 @@ on:
   merge_group:
     types: [checks_requested]
   workflow_dispatch:  # generally only for the "combine-prs" workflow
+permissions:
+  contents: read
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
@@ -18,7 +20,6 @@ jobs:
       buildId: ${{ steps.build.outputs.build-id}}
     permissions:
       id-token: write
-      contents: read
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -19,9 +16,14 @@ jobs:
     runs-on: depot-ubuntu-22.04-arm
     outputs:
       buildId: ${{ steps.build.outputs.build-id}}
+    permissions:
+      id-token: write
+      contents: read
```
woodruffw, line 21: Nitpick: `zizmor` doesn't currently suggest this but I'd recommend having a comment next to `id-token: write`, so that future reviewers can figure out why the job needs an OIDC credential (since they're pretty powerful).

OTOH this already didn't have one so I think it's OK to punt on this for a follow-up 🙂

woodruffw (approved):
LGTM! 🌈

ewdurbin (commented):
looks like excellent best practices work. assuming this was based on the output of a tool, can the tool be included in our linting to make sure we don’t regress and that new workflows and permissions are similarly well scoped?

woodruffw:
> looks like excellent best practices work. assuming this was based on the output of a tool, can the tool be included in our linting to make sure we don’t regress and that new workflows and permissions are similarly well scoped?

Yep! The easiest thing to do might be to copy this workflow, which uses the current SARIF support to integrate `zizmor`'s results directly into GitHub's security tool/reporting infrastructure: https://github.com/woodruffw/zizmor/blob/main/.github/workflows/zizmor.yml

(The other option would be to just run `zizmor` in CI, and manually triage the results it produces.)

woodruffw:
For ref, this is what the SARIF integration looks like (it's currently barebones, but will get better over time):

Screenshot 2024-10-30 at 5 34 50 PM

miketheman:
> can the tool be included in our linting to make sure we don’t regress and that new workflows and permissions are similarly well scoped?

Done in [311777b](https://github.com/pypi/warehouse/pull/16996/commits/311777b8930b9ac86790e85f4f3a18c127bb3092)

ewdurbin (approved):

PR #16998, opened 2024-10-29
refactor: drop custom warehouse.forms.Form

miketheman:
Split by commits to make it easier to review.

`wtforms` gained form-level errors in 3.0 circa 2020.
Refs: https://wtforms.readthedocs.io/en/3.0.x/forms/#wtforms.form.Form.form_errors

Most changes are drop-in replacements, only `UploadForm` used full-form validation, so should be read a bit closer.

### DIFF
--- a/tests/unit/forklift/test_forms.py
+++ b/tests/unit/forklift/test_forms.py
@@ -48,7 +48,7 @@ class TestUploadForm:
     )
     def test_full_validate_valid(self, data):
         form = UploadForm(MultiDict(data))
-        form.full_validate()
+        assert form.full_validate() is None
 
     @pytest.mark.parametrize(
         "data",
--- a/tests/unit/forklift/test_forms.py
+++ b/tests/unit/forklift/test_forms.py
@@ -10,45 +10,54 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import hashlib
 
 import pytest
 
 from webob.multidict import MultiDict
-from wtforms.validators import ValidationError
 
 from warehouse.forklift.forms import UploadForm
 
 
 class TestUploadForm:
+    _sha256 = hashlib.sha256().hexdigest()
+    _blake2b = hashlib.blake2b(digest_size=32).hexdigest()
+
     @pytest.mark.parametrize(
         "data",
         [
             # Test for singular supported digests
             {"filetype": "sdist", "md5_digest": "bad"},
             {"filetype": "bdist_wheel", "pyversion": "3.4", "md5_digest": "bad"},
-            {"filetype": "sdist", "sha256_digest": "bad"},
-            {"filetype": "bdist_wheel", "pyversion": "3.4", "sha256_digest": "bad"},
-            {"filetype": "sdist", "blake2_256_digest": "bad"},
-            {"filetype": "bdist_wheel", "pyversion": "3.4", "blake2_256_digest": "bad"},
+            {"filetype": "sdist", "sha256_digest": _sha256},
+            {"filetype": "bdist_wheel", "pyversion": "3.4", "sha256_digest": _sha256},
+            {"filetype": "sdist", "blake2_256_digest": _blake2b},
+            {
+                "filetype": "bdist_wheel",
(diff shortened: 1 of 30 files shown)


tests/unit/forklift/test_legacy.py
```diff
@@ -591,8 +591,10 @@ def test_fails_invalid_version(self, pyramid_config, pyramid_request, version):
                     "metadata_version": "1.2",
                     "name": "example",
                     "version": "1.0",
-                    "filetype": "bdist_wat",
+                    "filetype": "bdist_wheel",
+                    "content": "fake binary content",
                 },
+                "Invalid value for pyversion. "
```
miketheman, line 597: These additions to the output are due to now using wtform's validator response

tests/unit/manage/test_forms.py
```diff
@@ -899,14 +899,12 @@ def test_validate_name_with_fewer_than_max_applications(self, db_session):
...
-                    "Organization requests (3)."
-                )
-            ]
-        }
+        assert form.form_errors == [
+            (
+                "You have already submitted the maximum number of "
+                "Organization requests (3)."
+            )
+        ]
```
miketheman, line 907: This is the crux of the main API change - we could have kept iterating on `form.errors` but the key in their implementation is an empty string `""` instead of `"__all__"`, which we could customize the form class to override, but this felt cleaner.

tests/unit/forklift/test_forms.py
```diff
@@ -10,45 +10,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO:
+#  `tests.unit.forklift.test_legacy.TestFileUpload.test_fails_invalid_post_data`
+#  already covers the full form with specific cases.
+#  Should we remove these test cases? We will remain at 100% coverage.
```
miketheman, line 16: Answer and resolve this prior to merging.

ewdurbin: Seems like more test coverage on a critical part of the service is a good thing? I'd vote to keep the form level tests unless there is a good reason to remove them.

tests/unit/forklift/test_forms.py
```diff
@@ -10,45 +10,59 @@
...
+                "blake2_256_digest": _blake2b,
             },
         ],
     )
     def test_full_validate_valid(self, data):
+        # `name` is required for any submission
+        data["name"] = "fake-package"
         form = UploadForm(MultiDict(data))
-        form.full_validate()
+        assert form.validate(), form.errors
```
ewdurbin, line 65: I don't understand the assert here. does this mean `form.validate()` and `form.errors` both are truthy? I'd assume on a valid form that form.errors would be false.

warehouse/forklift/legacy.py
```diff
@@ -1363,8 +1361,8 @@ def file_upload(request):
     # Dispatch our task to sync this to cache as soon as possible
     request.task(sync_file_to_cache).delay(file_.id)
 
-    # Return any warnings that we've accumulated as the response body.
-    return HTTPOk(body="\n".join(warnings))
+    # Return successful response
+    return HTTPOk()
```
ewdurbin, line 1365: Seems like a valid cleanup since nothing seems to append to `warnings` anymore. I'm not sure how this was used in the past, but feels like removing it will lose the (seemingly) undocumented method of notifying clients.

ewdurbin (approved):
Overall seems great! A couple of non-blocking comments/questions.

tests/unit/forklift/test_forms.py
miketheman: This is leveraging `assert`s built-in behavior to allow customization of the assertion error output.

Paraphrasing:

> If form.validate() is not truthy, show an assertion error with the form.errors so we can more easily debug why it didn't validate

https://docs.python.org/3/reference/simple_stmts.html#assert

warehouse/forklift/legacy.py
miketheman: Good point, will restore with some inline docs

Issue #16844, opened 2024-10-07
Allow project owners to set status markers ("deprecated", "archived", etc) to their PyPI projects


facutuesca:
## **What's the problem this feature will solve?**

Currently, PyPI does not have a standard way of indicating that a project has been deprecated, archived, etc by the maintainers.
This information is (sometimes) included in the project's description, but there's no standard way of setting (or getting) it, other than manually checking.

If owners can mark a project as deprecated, archived, etc. on PyPI and this information is exposed through the web UI and the index APIs, downstream consumers can monitor this information to make better decisions about their supply chain.

## **Describe the solution you'd like**
Recently `warehouse` added support for marking a project as quarantined: https://github.com/pypi/warehouse/pull/16179.
This involved adding a new field to the Project model:
```python
class LifecycleStatus(enum.StrEnum):
    QuarantineEnter = "quarantine-enter"
    QuarantineExit = "quarantine-exit"
```
We could model the new project statuses as lifecycle statuses. For example:
```python
class LifecycleStatus(enum.StrEnum):
    QuarantineEnter = "quarantine-enter"
    QuarantineExit = "quarantine-exit"
    Deprecated = "deprecated"
    Archived = "archived"
    Finished = "finished"
```

The project owner should only be able to set the new statuses (not the quarantine related ones), and only if the project is not quarantined. Something like:
```mermaid
stateDiagram-v2
    direction LR
    [*] --> UserStates
    state UserStates {
        None
        Archived
        Deprecated
        Finished
    }
    UserStates --> UserStates
    UserStates --> QuarantineEnter : 
    QuarantineEnter --> QuarantineExit
    QuarantineExit --> QuarantineEnter
    QuarantineExit --> UserStates
```

The UI for setting the status can be a simple drop down:
<img width="300" alt="image" src="https://github.com/user-attachments/assets/ce81922f-9526-41d0-ac76-a8373bde1e21"> <img width="300" alt="image" src="https://github.com/user-attachments/assets/1d9d3a70-9f11-4dd9-8bff-394481dc5029">

And the UI for the project's main page can be similar to the quarantined one:
<img width="1109" alt="image" src="https://github.com/user-attachments/assets/e7668b45-2e57-47c7-a884-1f344f7adfb2">

## **Additional context**
[...]

PR #17005, opened 2024-10-30
Allow marking projects as "archived"

facutuesca:
## What is this?
Following the discussion [here](https://github.com/pypi/warehouse/issues/16844), this PR shows what the implementation could look like for adding "status markers" to projects.

**Specifically, this PR allows changing a project's status to "Archived"**.

## Details
The implementation is straightforward, since a `Project.lifecycle_status` field already exists (to mark a project as quarantined). This change adds a new "archived" value for that field, and adds the UI so that users can archive and unarchive their projects.

Even though the issue mentioned above proposes more statuses ("deprecated", "finished", etc), those can be ambiguous and overlapping, and would require further discussion to define clearly. "Archived", however, is already widely used on [GitHub](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) repos and its meaning should be easy for users to understand:
> You can archive a repository to make it read-only for all users and indicate that it's no longer actively maintained. You can also unarchive repositories that have been archived

## TODO
- [x] Actually restrict file uploads when project is Archived
- [x] Add admin UI to archive/unarchive projects

### Project settings

image

image

image

### Project landing page
image

cc @woodruffw @miketheman

### DIFF
--- a/warehouse/admin/templates/admin/projects/detail.html
+++ b/warehouse/admin/templates/admin/projects/detail.html
@@ -96,6 +96,27 @@
   </div>
   <!-- /.modal -->
 
+  {% elif project.lifecycle_status == 'archived' %}
+    <div class="card card-info">
+      <div class="card-header">
+        <h3 class="card-title">
+          <i class="icon fas fa-info-circle"></i> Archived
+        </h3>
+        <div class="card-tools">
+          <span class="time">
+            <i class="fas fa-clock"></i> <time datetime="{{ project.lifecycle_status_changed }}" title="{{ project.lifecycle_status_changed }}">{{ project.lifecycle_status_changed }}</time>
+          </span>
+        </div>
+      </div>
+      <div class="card-body">
+        <p>
+          This project is archived.
+          It should not allow any new uploads unless it's unarchived.
+        </p>
+      </div>
+    </div>
+    <!-- .card -->
+
   {% endif %}
 
 <div class="card card-primary" id="attributes">
@@ -599,13 +620,13 @@
       <div class="float-right">
         <button type="submit" class="btn btn-primary" title="{{ 'Unarchiving requires superuser privileges' if not request.has_permission(Permissions.AdminProjectsWrite) }}" {{ "disabled" if not request.has_permission(Permissions.AdminProjectsWrite) or not can_be_unarchived }}>Unarchive</button>
       </div>
-    </form> 
+    </form>
     <form method="POST" action="{{ request.route_path('admin.project.archive', project_name=project.name) }}">
       <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
       <div class="float-right">
         <button type="submit" class="btn btn-primary" title="{{ 'Archiving requires superuser privileges' if not request.has_permission(Permissions.AdminProjectsWrite) }}" {{ "disabled" if not request.has_permission(Permissions.AdminProjectsWrite) or not can_be_archived}}>Archive</button>
       </div>
-    </form> 
+    </form>
   </div>
 </div>
 
--- a/warehouse/admin/templates/admin/projects/detail.html
+++ b/warehouse/admin/templates/admin/projects/detail.html
@@ -582,6 +582,33 @@
   </div>
 </form> <!-- .card #prohibitedProjectName -->
 
+<div class="card card-warning card-outline collapsed-card" id="archivedProjectName">
+  <div class="card-header">
+    <h3 class="card-title">Archive project</h3>
+    <div class="card-tools">
+      <button type="button" class="btn btn-tool" data-card-widget="collapse"><i class="fas fa-plus"></i>
+      </button>
+    </div>
+  </div>
+
+  <div class="card-footer">
+    {% set can_be_archived = not project.lifecycle_status or project.lifecycle_status == "quarantine-exit" %}
+    {% set can_be_unarchived = project.lifecycle_status == "archived" %}
+    <form method="POST" action="{{ request.route_path('admin.project.unarchive', project_name=project.name) }}">
+      <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+      <div class="float-right">
+        <button type="submit" class="btn btn-primary" title="{{ 'Unarchiving requires superuser privileges' if not request.has_permission(Permissions.AdminProjectsWrite) }}" {{ "disabled" if not request.has_permission(Permissions.AdminProjectsWrite) or not can_be_unarchived }}>Unarchive</button>
+      </div>
+    </form> 
+    <form method="POST" action="{{ request.route_path('admin.project.archive', project_name=project.name) }}">
+      <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+      <div class="float-right">
+        <button type="submit" class="btn btn-primary" title="{{ 'Archiving requires superuser privileges' if not request.has_permission(Permissions.AdminProjectsWrite) }}" {{ "disabled" if not request.has_permission(Permissions.AdminProjectsWrite) or not can_be_archived}}>Archive</button>
+      </div>
+    </form> 
+  </div>
+</div>
+
 <!-- Deletion form -->
 {% include 'delete.html' %}
 
--- a/warehouse/migrations/versions/12a43f12cc18_add_new_lifecycle_statuses.py
+++ b/warehouse/migrations/versions/12a43f12cc18_add_new_lifecycle_statuses.py
@@ -1,0 +1,55 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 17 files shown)


miketheman (commented):
Generally, looking good, thanks for scoping it down to a single lifecycle status.

I think something good to add to the owner-facing management page would be a suggestion to publish a new release with a readme update or something to point folks where to go next.
Prior ideas have been to allow user-submitted text in the form, but I like that less, since then we have to figure out where to store that.

I think a warning modal popup confirmation might also be in order when archiving would be a good idea to prevent misclicks.

This also doesn't appear to block users from making further releases to an archived project - is that intentionally excluded? In my head, once Archived, it's in read-only mode - similar to GitHub's Archived.

miketheman:
I also just realized - this probably needs some added notation in the Admin UI/ability to change there as well. I didn't see anything in scope here yet.

facutuesca:
@miketheman 
> I think something good to add to the owner-facing management page would be a suggestion to publish a new release with a readme update or something to point folks where to go next.
Prior ideas have been to allow user-submitted text in the form, but I like that less, since then we have to figure out where to store that.
> 
> I think a warning modal popup confirmation might also be in order when archiving would be a good idea to prevent misclicks.> 

I changed the UI to add the suggestion and the modal popup. Since these are specific to archiving a project, I changed the UI so that instead of a dropdown with the backend statuses, the user sees two buttons to archive or unarchive their project (see the updated screenshots)

> This also doesn't appear to block users from making further releases to an archived project - is that intentionally excluded? In my head, once Archived, it's in read-only mode - similar to GitHub's Archived.

I'll add the restrictions once the rest is ready. I was thinking to only disallow uploads for now, WDYT?

> this probably needs some added notation in the Admin UI/ability to change there as well. I didn't see anything in scope here yet.

What should the UI for this look?

facutuesca:
Pushed a commit to disallow uploads on archived projects

facutuesca:
Added admin UI to archive/unarchive projects:


image

warehouse/utils/project.py
```diff
@@ -189,3 +189,42 @@ def destroy_docs(project, request, flash=True):
...
+            additional={
+                "submitted_by": request.user.username,
+            },
+        )
+        request.session.flash("Project archived", queue="success")
+    else:
+        request.session.flash(
+            f"Cannot archive project with status {project.lifecycle_status}",
+            queue="error",
+        )
```
miketheman, line 212: style: Consider early-exit inversion of the logic, like so (pseudocode):

```python
if not <conditions that must be met>:
    flash "cannot archive", error
    return

do the work
flash "done", success
```

thus communicate early on the exit clause, continue normally otherwise.
Removes one layer of indentation for "the work", and makes it simpler to create another if/else when necessary with fewer logical jumps to make.

Similar applies to `unarchive_project`

warehouse/packaging/models.py
```diff
@@ -330,7 +331,11 @@ def __acl__(self):
         # The project has zero or more OIDC publishers registered to it,
         # each of which serves as an identity with the ability to upload releases.
         for publisher in self.oidc_publishers:
-            acls.append((Allow, f"oidc:{publisher.id}", [Permissions.ProjectsUpload]))
+            if self.lifecycle_status != LifecycleStatus.Archived:
+                # Only allow uploads in non-archived projects
+                acls.append(
+                    (Allow, f"oidc:{publisher.id}", [Permissions.ProjectsUpload])
+                )
```
miketheman, line 338: optimize: Adding the `if` will now happen on each iteration of the loop per `publisher`, but the project's lifecycle happens once, probably can shortcut the loop instead by placing the `if` outside.
Most projects only have a single publisher, but just in case, save the iteration if unnecessary.

warehouse/packaging/models.py
```diff
@@ -364,27 +369,27 @@ def __acl__(self):
...
+                ]
             else:
-                acls.append((Allow, f"user:{user_id}", [Permissions.ProjectsUpload]))
+                current_permissions = [Permissions.ProjectsUpload]
+
+            if self.lifecycle_status == LifecycleStatus.Archived:
+                # Disallow upload permissions for archived projects
+                current_permissions = [
+                    p for p in current_permissions if p != Permissions.ProjectsUpload
+                ]
```
miketheman, line 389: optimize: could this be instead?
```suggestion
                current_permissions.remove(Permissions.ProjectsUpload)
```
I _think_ each branch of `current_permissions` ends up with `ProjectsUpload`, and this might be faster than a list comprehension?

miketheman (changes requested):
Coming along nicely, thanks for pushing forward on this. I had some notes inline, mostly structural in nature, but some questions as well - nothing too "heavy".

Let me know if anything isn't clear!

warehouse/packaging/models.py
facutuesca: makes sense!

warehouse/utils/project.py
facutuesca: fixed!

warehouse/packaging/models.py
facutuesca: good catch, fixed!

miketheman (changes requested):
Hey @facutuesca ! Nice progress here. I tested everything locally.

I left some more notes inline, let me know if they dpn't make sense!

warehouse/migrations/versions/12a43f12cc18_add_new_lifecycle_statuses.py
```diff
@@ -0,0 +1,55 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add new lifecycle statuses
+
+Revision ID: 12a43f12cc18
+Revises: ed4cc2ef6b0f
```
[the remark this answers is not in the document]
facutuesca: Fixed

warehouse/templates/manage/project/settings.html
```diff
@@ -40,6 +40,38 @@ <h2>{% trans %}Project settings{% endtrans %}</h2>
     <a href="{{ help_url }}"> (request an increase) </a>{% endtrans %}</li>
   </ul>
 
+  <h2>{% trans %}Archive project{% endtrans %}</h2>
```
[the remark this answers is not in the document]
facutuesca: good point, fixed:

Screenshot 2025-01-16 at 16 35 58

warehouse/templates/manage/project/settings.html
```diff
@@ -40,6 +40,38 @@ <h2>{% trans %}Project settings{% endtrans %}</h2>
...
+  {% set can_be_archived = not project.lifecycle_status or project.lifecycle_status == "quarantine-exit" %}
+  {% set can_be_unarchived = project.lifecycle_status == "archived" %}
+  <div class="callout-block callout-block--danger" data-controller="archive-confirm">
+  <p>
+    {% trans %}
+    Archiving a project will block any new uploads. Before doing so, we
```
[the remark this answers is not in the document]
facutuesca: fixed

warehouse/admin/templates/admin/projects/detail.html
```diff
@@ -582,6 +582,33 @@ <h3 class="card-title">Prohibit Project Name</h3>
   </div>
 </form> <!-- .card #prohibitedProjectName -->
 
+<div class="card card-warning card-outline collapsed-card" id="archivedProjectName">
```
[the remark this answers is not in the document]
facutuesca: oh yeah that's a good idea, fixed:

Screenshot 2025-01-16 at 16 53 00

warehouse/migrations/versions/12a43f12cc18_add_new_lifecycle_statuses.py
```diff
@@ -0,0 +1,55 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Add new lifecycle statuses
+
+Revision ID: 12a43f12cc18
+Revises: e24aa37164e72
```
miketheman, line 16:
```suggestion
Revises: 24aa37164e72
```

warehouse/templates/manage/project/settings.html
```diff
@@ -351,6 +351,43 @@ <h3>
 
   <hr>
 
+  <h2>{% trans %}Archive project{% endtrans %}</h2>
+  {% set can_be_archived = not project.lifecycle_status or project.lifecycle_status == "quarantine-exit" %}
+  {% set can_be_unarchived = project.lifecycle_status == "archived" %}
+  <div class="callout-block callout-block--danger" data-controller="archive-confirm">
+  <p>
+    {% trans readme_description_href='https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/'%}
```
miketheman, line 359: Drop locale and version to let the platform auto-route
```suggestion
    {% trans readme_description_href='https://packaging.python.org/guides/making-a-pypi-friendly-readme/'%}
```

miketheman (approved):

Issue #16491, opened 2024-08-15
Add a per-artifact view


di:
**What's the problem this feature will solve?**
Currently it's hard to add additional information about individual artifacts without overloading the "Download files" tab on the project page. For example, file hashes need to be displayed in a modal.

**Describe the solution you'd like**
Instead of listing all artifacts on the "Files" page of a project (e.g. https://pypi.org/project/pip/#files), link to a new view for individual files, which contains additional details about the file:

* Hash digests
* File size
* Human-readable wheel/platform tags
* When the artifact was published
* How the artifact was published (by a maintainer? trusted publisher?)
* Any PEP 740 provenance for the file

**Additional context**
This also allows us to deep-link to a context page for an individual artifact, and potentially gives us a place to include a per-artifact file browser (e.g. by embedding http://inspector.pypi.io/ on this page)

#15087 is potentially related as well.

di:
@woodruffw Yep!

di:
Fixed in [#17052](https://github.com/pypi/warehouse/pull/17052).

PR #17052, opened 2024-11-07
File details view, including attestations

woodruffw:
See #16491.


requirements/main.txt
```diff
@@ -1779,9 +1779,9 @@ pyparsing==3.2.0 \
     --hash=sha256:93d9577b88da0bbea8cc8334ee8b918ed014968fd2ec383e868fb8afb1ccef84 \
     --hash=sha256:cbf74e27246d595d9a74b186b810f6fbb86726dbf3b9532efb343f6d7294fe9c
     # via linehaul
-pypi-attestations==0.0.15 \
-    --hash=sha256:79d974fc162676383c531eb04c2197876dc09c1983ba915ebd09c7c18d626aed \
-    --hash=sha256:892de82329ccbc88f0866aa3cbf7b2a071a6ff22bcb03b7fd44bd002709ad481
+pypi-attestations==0.0.16 \
+    --hash=sha256:1d816719b5067ef49ac47c7ae229bd03752c57f957daf95b19b2ba19a66220e4 \
+    --hash=sha256:cbd2b946fe160793606dee4516ef58ac5959456e69630a7f03e7de73aa7f2737
```
woodruffw, line 1784: NB: Bumped this to expose `attestation.statement`, which in turn allows us to poke through the in-toto statement without having to decode JSON within the Jinja2 templates.

See https://github.com/trailofbits/pypi-attestations/commit/79a57b00213bddd40d589675fa275d5fa70468be for the changes there.

warehouse/static/sass/blocks/_file_details.scss
woodruffw, line 1: NB: This entire file can probably be deleted; this was my attempt to replicate the current project view layout with flex layouts instead, since the current project views assume a bunch of tab styling that doesn't mesh well with this new view.

warehouse/templates/packaging/file-details.html
```diff
@@ -0,0 +1,143 @@
...
+              <ul>
+                {% for attestation in bundle.attestations %}
+                <li>
+                  {% set statement = attestation.statement %}
+                  Statement type: {{ statement._type }}
+                  <ul>
+                    <li>Predicate type: {{ statement.predicateType }}</li>
+                    <li>Subject name: {{ statement.subject[0].name }}</li>
+                    <li>Subject digest: {{ statement.subject[0].digest.sha256 }}</li>
+                    <li>Transparency log index: {{ attestation.verification_material.transparency_entries[0].logIndex }} </li>
```
woodruffw, line 125: NB: This could link to the right entry via <https://search.sigstore.dev/>

di:
Heads up, going to be pushing some changes here shortly & then we can revisit next steps.

di:
I reworked this to load all the file details onto the project details page and use anchors instead of having a separate route. This means that we won't have NxM new URLs to put into cache, but still allows us to deep-link to project details.

requirements/main.txt
di: Bumped in https://github.com/pypi/warehouse/pull/17054.

warehouse/templates/includes/file-details.html
```diff
@@ -0,0 +1,134 @@
...
+  </p>
+</div>
+
+{% if provenance %}
+<div>
+  <p>
+    The following attestations were made for <code>{{ file.filename }}</code>:
+  </p>
+
+  {% for bundle in provenance.attestation_bundles %}
```
woodruffw, line 100: FWIW this doesn't quite work: `provenance` is now unbound, and `file.provenance` is now the raw JSON column rather than a `Provenance` model. I'll look at adjusting for this.

woodruffw: Done with 71a4036561fe9c9e965643f05f2672c26c9cb14d: I've added `Provenance.as_model` to return the structured `pypi_attestations.Provenance` representation. I'm not sure if this is the best approach, though.

di (approved):

PR #17065, opened 2024-11-12
legacy: ensure invalid attestation error is never empty

woodruffw:
This adds a prefix to ensure that we never emit a fully empty `HTTP 400`. I'll do a follow-up for moving the validation further up as well.

See: https://github.com/pypa/gh-action-pypi-publish/issues/303

CC @di


di (approved):

woodruffw:
> @woodruffw Looks like there's a test failure here.

Whoops, missed a space. Fixed!

PR #17067, opened 2024-11-12
legacy: split attestation handling phases

woodruffw:
~~(WIP; I need to add some coverage.)~~

Only persist attestations once the file is created.


warehouse/forklift/legacy.py
```diff
@@ -1229,19 +1250,9 @@ def file_upload(request):
...
-        if "attestations" in request.POST and not request.flags.enabled(
-            AdminFlagValue.DISABLE_PEP740
-        ):
-            integrity_service: IIntegrityService = request.find_service(
-                IIntegrityService, context=None
-            )
-
+        # If we have attestations from above, persist them.
+        if attestations:
             try:
```
di, line 1255: I think we want to remove the try/except here as well? I don't think that `build_provenance` will fail with an `AttestationUploadError`.

woodruffw: > I think we want to remove the try/except here as well? I don't think that `build_provenance` will fail with an `AttestationUploadError`.

Yeah, in practice it can't fail -- there're one `AttestationUploadError` in `build_provenance`, but it can only occur if there's a discrepancy between `OIDCPublisher.supports_attestations` and `__publisher_from_oidc_publisher`.

I'll look at refactoring this more generally to eliminate that illegal state -- there's no reason for a discrepancy between the two.

di:
@woodruffw FYI, this still needs some work.

di (approved):

PR #17091, opened 2024-11-15
user-docs: add section on trustworthiness

woodruffw:
This adds another sub-section to the attestation security docs, explaining the concepts of authenticity/trustworthiness and why attestations can only provide the former while *enabling* the latter.


woodruffw:
NB: This *might* belong somewhere else (maybe a separate FAQ/terms page)? Open to ideas 🙂

di (approved):

PR #17094, opened 2024-11-18
lint: modernize usage of babel

miketheman:
eslint now provides its own `Espree` parser, so we don't need to include Babel for that purpose.

Since we only need babel for jest, update references accordingly.


ewdurbin (approved):

ewdurbin:
Smol question. Are the obvious formatting changes related or just an aside of editing those files?

PR #17107, opened 2024-11-18
Be slightly more strict in checking for distribution file contents

ewdurbin:
This improves our validation of distribution file contents by being more precise in checking for sdist's to have _one_ top level directory, this allows us to be more precise in checking for PKG-INFO and License-File files that should exist in a given sdist.

Also improves error messages for users if an invalid distribution is attempted to be uploaded.


ewdurbin:
In an ideal world, the top-level directory name inside an sdist would have consistent normalization to enable us to skip the heuristic that's applied here with `os.path.commonprefix`, but the upshot of that is that it does start to enforce the "one top level directory" spec for sdists.

di (approved):

PR #17108, opened 2024-11-18
attestations: remove double states, simplify tests

woodruffw:
This removes some usage of stubs in favor of real models (via factories) where possible, and eliminates some potential sources of double-state/divergence in the original services.

In particular:

* `OIDCPublisherMixin.supports_attestations` is now `attestation_identity`, and returns a `Publisher | None` that can be used directly for verification
* `OIDCPublisherMixin.publisher_verification_policy` is removed, since the `Publisher` now encodes the verification policy
* `_publisher_from_oidc_publisher` is removed, since `attestation_identity` serves the same purpose

### DIFF
--- a/warehouse/attestations/services.py
+++ b/warehouse/attestations/services.py
@@ -23,10 +23,7 @@ from pypi_attestations import (
     AttestationBundle,
     AttestationType,
     Distribution,
-    GitHubPublisher,
-    GitLabPublisher,
     Provenance,
-    Publisher,
     VerificationError,
 )
 from pyramid.request import Request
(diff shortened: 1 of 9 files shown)


woodruffw:
NB: This also removes our top-level dependency on `sigstore`, since all ops go through `pypi_attestations`. I can remove that in this PR or a follow-on.

di:
> NB: This also removes our top-level dependency on sigstore, since all ops go through pypi_attestations (which still uses sigstore itself, of course). I can remove that in this PR or a follow-on.

Let's go ahead and do it here, thanks!

warehouse/attestations/services.py
```diff
@@ -36,53 +33,17 @@
...
-        publisher: Publisher = _publisher_from_oidc_publisher(request.oidc_publisher)
-    except AttestationUploadError as exc:
-        sentry_sdk.capture_message(
-            f"Unsupported OIDCPublisher found {request.oidc_publisher.publisher_name}"
-        )
-        raise exc
-
     attestation_bundle = AttestationBundle(
-        publisher=publisher,
+        publisher=request.oidc_publisher.attestation_identity,
```
di, line 46: Bit of a nit: This function is now simple enough that I think it can be eliminated and we can call this inline instead.

di (approved):
LGTM aside from 1 nit.

warehouse/attestations/services.py
woodruffw: Yeah, good call -- doing!

PR #17109, opened 2024-11-18
feat(admin): freeze on quarantine, add user observation on removal

miketheman:
When a project is quarantined, also freeze the associated user accounts
to prevent them from creating further projects/releases.

Help us see why a user ended up in a frozen state for anyone frozen
during malware verdicts.

### DIFF
--- a/warehouse/admin/templates/admin/users/detail.html
+++ b/warehouse/admin/templates/admin/users/detail.html
@@ -883,10 +883,42 @@
         </div>
       </div> <!-- .card -->
 
+    {% if user.observations %}
+      <div class="card card-warning card-outline">
+        <div class="card-header with-border">
+          <h3 class="card-title">Observations about <code>{{ user.username }}</code></h3>
+        </div>
+        <div class="card-body">
+          <table id="user_observations" class="table table-striped table-hover">
+            <thead>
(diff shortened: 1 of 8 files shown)


warehouse/admin/templates/admin/users/detail.html
```diff
@@ -883,10 +883,42 @@ <h3 class="card-title">Account activity</h3>
...
+          </table>
+        </div>
+      </div> <!-- .card -->
+    {% endif %}
+
     {% if user.observer.observations %}
       <div class="card">
         <div class="card-header with-border">
-          <h3 class="card-title">Observations</h3>
+          <h3 class="card-title">User-submitted Observations</h3>
```
ewdurbin, line 921: Would "Admin-submitted" be more accurate?

ewdurbin (approved):

warehouse/admin/templates/admin/users/detail.html
miketheman: It would, I like that better!

miketheman, line 921:
```suggestion
          <h3 class="card-title">Admin-submitted Observations</h3>
```

Issue #17111, opened 2024-11-18
Cannot upload whl to testpypi - 400 Invalid distribution file


msquaredds:
**Describe the bug**
I cannot upload my whl file to test PyPI. I have a project that I've been uploading to regularly and today, when trying to upload my whl file, I get the following error:

```
INFO     Response from https://test.pypi.org/legacy/:
         400 Bad Request
INFO     <html>
          <head>
           <title>400 Invalid distribution file. WHEEL not found at streamlitauth-0.7.18.dist-info/WHEEL</title>
          </head>
          <body>
           <h1>400 Invalid distribution file. WHEEL not found at streamlitauth-0.7.18.dist-info/WHEEL</h1>
           The server could not comply with the request since it is either malformed or otherwise incorrect.<br/><br/>
         Invalid distribution file. WHEEL not found at streamlitauth-0.7.18.dist-info/WHEEL
          </body>
         </html>
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
         Bad Request
```

When I look in the whl file, I see the WHEEL file and it looks the same as it has in the past
This is the top level whl:
![Image](https://github.com/user-attachments/assets/75f63b34-b4d6-4af3-ae9d-9d600a275783)
Then the is the StreamlitAuth-0.7.18.dist-info folder:
![Image](https://github.com/user-attachments/assets/edb45855-8177-41fc-bf69-d02f1cdd65d4)
And finally the WHEEL file:
![Image](https://github.com/user-attachments/assets/9878ee67-5b13-4ad0-b3aa-c814511ed9b8)

**Expected behavior**
I would expect the upload to succeed.

**To Reproduce**
I run this in cmd:
`python -m twine upload --repository testpypi dist/StreamlitAuth-0.7.18-py3-none-any.whl --verbose`
Followed by __token__ and my password

**My Platform**
I'm using cmd in Windows 10 Home

**Additional context**
I can attach the actual files if necessary

mrmundt:
+1 - we are having the same issue on Pyomo. We are trying to upload our wheels using `twine upload *.whl` and we get the error:
```
Uploading distributions to https://upload.pypi.org/legacy/
Enter your API token: 
Uploading Pyomo-6.8.2-cp310-cp310-macosx_10_9_x86_64.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 5.6/5.6 MB • 00:05 • 1.1 MB/s
[...]

PR #17112, opened 2024-11-18
relax wheel filename restriction from #17107

ewdurbin:
the restrictions applied in #17107 introduced a bit _too_ strict handling of wheels than we're ready for. using the parse_wheel_filename normalized name/version, which is a new and more strict constraint than intended


di (approved):

ewdurbin:
Closes #17111

PR #17124, opened 2024-11-19
add a helper to project admin include for ultranormalization collisions

### LINKED ISSUE #17113: No Project Info Button Exists on PyPI Project Page Banner for Similarly Named Projects
Currently, there is no info button for a project/package name that is similar to another existing package name in the "Admin tools for project" banner. If a project name is prohibited, there is an info button that indicates that it is "Prohibited", or if a project already exists, an info button such as "View project in admin" exists. Having a similar solution for project names that have not been allowed due existing similarly named projects/packages would be a useful feature as it would let admin/support see and confirm that a project name has been denied due to its similarity with an existing project name. 

An info button indicating if a project name is already similar to an existing package name. It can simply say "Similar Name." Clicking on it can provide additional info on the backend of what project name it is specifically similar to for additional information. 

**Additional context**

ewdurbin:
resolves #17113

Example:
Screenshot 2024-11-19 at 4 54 23 PM

### DIFF
--- a/warehouse/admin/views/includes.py
+++ b/warehouse/admin/views/includes.py
@@ -11,7 +11,7 @@
 # limitations under the License.
 
 from pyramid.view import view_config
-from sqlalchemy import func
+from sqlalchemy import func, select
 
 from warehouse.accounts.models import User
 from warehouse.packaging.models import ProhibitedProjectName, Project
@@ -34,8 +34,8 @@ def administer_project_include(request):
         .filter(ProhibitedProjectName.name == func.normalize_pep426_name(project_name))
         .one_or_none()
     )
-    collisions = (
-        request.db.query(Project)
+    collisions = request.db.scalars(
+        select(Project.name)
         .filter(
             func.ultranormalize_name(Project.name)
             == func.ultranormalize_name(project_name)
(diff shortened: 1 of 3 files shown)


warehouse/admin/views/includes.py
```diff
@@ -34,8 +34,25 @@ def administer_project_include(request):
...
+        .all()
+    )
 
-    return {"project": project, "project_name": project_name, "prohibited": prohibited}
+    return {
+        "project": project,
+        "project_name": project_name,
+        "prohibited": prohibited,
+        "collisions": [p.name for p in collisions],
+    }
```
miketheman, line 55: Optimize: Since we're not using any part of the results other than the colliding name we can optimize and only return the `name` part from the DB, saving data load/roundtrip, serialization work, etc.
This is more aligned with modern SQLAlchemy syntax, where we construct a statement and pass it to the session.

Needs an update to import `select` as well

```python
collisions = request.db.scalars(
    select(Project.name)
    .filter(
        func.ultranormalize_name(Project.name)
        == func.ultranormalize_name(project_name)
    )
    .filter(
        func.normalize_pep426_name(Project.name)
        != func.normalize_pep426_name(project_name)
    )
).all()
```

And then no need to listcomp `"collisions"` later, since it'll be constructed as a list of strings.

miketheman (approved):
Nice stuff! I have sone suggestion for optimizing the query - it's not going to be materially impactful due to the low usage/latency of this view, but I'd be remiss if I didn't note it.

warehouse/admin/views/includes.py
ewdurbin: 8796408c1630ced73fcf30dd9499bbdddb917e06

Issue #17001, opened 2024-10-30
PEP 740: Post-deployment tasks


woodruffw:
The main roadmap for PEP 740 is in https://github.com/pypi/warehouse/issues/15871; this tracks related items that aren't blockers, but need to be thought about more and/or addressed in the medium-to-long term.

- [ ] Support for other attestation types. Right now the "PyPI Publish" and SLSA types are supported, although the SLSA type is currently gated off since nothing is uploading it yet. The latter should be enabled.
- [ ] Evaluate additional forms of domain separation in the attestation body. Right now there's a scenario in which a single workflow identity can sign publish attestations for both PyPI and TestPyPI, and those two attestations are indistinguishable (since a publish attestation doesn't currently encode its intended "recipient" index). This leaves open a subset of index poisoning attacks, where an attacker manages to obtain attestations intended for TestPyPI and gets PyPI to serve them instead. @alex and I originally discussed this ~2 years ago and the original approach within Fulcio itself didn't pan out, so it needs to be tackled within the attestation body itself.
    * The solution to this is to encode a domain separator in the attestation payload, which could be as simple as the current OIDC audiences for PyPI and TestPyPI (`pypi` and `testpypi` respectively). We have a strong source of timeliness in the form of signed time, so we could add these and then require them in new attestations after a period of adoption. 
- [ ] Enable attestation support for other Trusted Publisher providers, e.g. GCB and ActiveState. This should be straightforward, just needs tests.

webknjaz:
> * Support for other attestation types. Right now the "PyPI Publish" and SLSA types are supported, although the SLSA type is currently gated off since nothing is uploading it yet. The latter should be enabled.

Is it possible to do this in a job following pypi-upload?

woodruffw:
> > * Support for other attestation types. Right now the "PyPI Publish" and SLSA types are supported, although the SLSA type is currently gated off since nothing is uploading it yet. The latter should be enabled.
> 
> Is it possible to do this in a job following pypi-upload?

Not at the moment -- in principle we could add a separate upload endpoint/codepath for uploading attestations to an already-uploaded release, but that doesn't exist yet. For the time being, the assumption is that one or more attestations get uploaded with the release itself, so a user who wants to upload a SLSA attestation should put it in their `dist/` alongside their publish attestation.

In other words, a `dist/` structure like:

```
dist/
  foo-1.2.3.tar.gz
  foo-1.2.3.tar.gz.publish.attestation
[...]

PR #17125, opened 2024-11-19
Add support PEP-740 attestations for GitLab CI/CD

facutuesca:
Adds support for uploading PEP-740 attestations generated from a GitLab CI/CD workflow.
The actual verification is done inside `pypi-attestations`, similar to GitHub.

Tested locally with an attestation signed from GitLab CI/CD:
image

Rekor entry: https://search.sigstore.dev/?logIndex=149980684

Part of https://github.com/pypi/warehouse/issues/17001

cc @woodruffw @di

### DIFF
--- a/warehouse/attestations/services.py
+++ b/warehouse/attestations/services.py
@@ -151,7 +151,6 @@ class IntegrityService:
         artifact. Attestations are only allowed when uploading via a Trusted
         Publisher, because a Trusted Publisher provides the identity that will be
         used to verify the attestations.
-        Only GitHub Actions Trusted Publishers are supported.
         """
 
         attestations = _extract_attestations_from_request(request)
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -64,7 +64,7 @@ redis>=2.8.0,<6.0.0
 rfc3986
 sentry-sdk
 setuptools
-pypi-attestations==0.0.16
+pypi-attestations==0.0.17
 sqlalchemy[asyncio]>=2.0,<3.0
 stdlib-list
 stripe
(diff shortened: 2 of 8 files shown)


facutuesca:
TODO: Once merged, I'll open a PR adding documentation on how to generate the attestations from a GitLab workflow

warehouse/attestations/services.py
```diff
@@ -151,7 +151,7 @@ def parse_attestations(
         artifact. Attestations are only allowed when uploading via a Trusted
         Publisher, because a Trusted Publisher provides the identity that will be
         used to verify the attestations.
-        Only GitHub Actions Trusted Publishers are supported.
+        Only GitHub Actions and GitLab CI/CD Trusted Publishers are currently supported.
```
woodruffw, line 154: FWIW, I think we can just remove this line -- this is documented elsewhere and is encoded in `attestation_identity` being present or now anyways.

facutuesca: fixed!

woodruffw (approved):
LGTM, great work @facutuesca!

woodruffw:
Flagging one thing (could be done now or in a follow-up): the `publisher` macro also needs to be updated:

https://github.com/pypi/warehouse/blob/dde8b3d51ac46840f42c8bca0f2a67b875aa5a65/warehouse/templates/includes/file-details.html#L15-L24

facutuesca:
> Flagging one thing (could be done now or in a follow-up): the `publisher` macro also needs to be updated:
> 
Good catch, fixed!

facutuesca:
Opened PR for the docs here: https://github.com/pypi/warehouse/pull/17133

di (approved):

PR #17134, opened 2024-11-20
Allow multiple attestations per distribution

facutuesca:
Now that [SLSA provenances are supported](https://github.com/pypi/warehouse/pull/17121) in PEP-740 attestations, this PR removes the limit of 1 attestation per file.

This is bumped to 2 attestations per file, one per predicate type supported in PEP-740. This should allow users to upload both SLSA provenance and PyPI publish attestations for each distribution file.

This is part of https://github.com/pypi/warehouse/issues/17001.

cc @woodruffw @di

### DIFF
--- a/warehouse/attestations/services.py
+++ b/warehouse/attestations/services.py
@@ -159,11 +159,11 @@ class IntegrityService:
         # Sanity-checked above.
         expected_identity = request.oidc_publisher.attestation_identity
 
-        seen_predicate_types: set[str] = set()
+        seen_predicate_types: set[AttestationType] = set()
 
         for attestation_model in attestations:
             try:
-                predicate_type, _ = attestation_model.verify(
+                predicate_type_str, _ = attestation_model.verify(
                     expected_identity,
                     distribution,
                 )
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
LGTM, thanks @facutuesca!

warehouse/attestations/services.py
```diff
@@ -190,6 +193,17 @@ def parse_attestations(
                     f"Attestation with unsupported predicate type: {predicate_type}",
                 )
 
+            if predicate_type in seen_predicate_types:
+                self.metrics.increment(
+                    "warehouse.upload.attestations.failed_duplicate_predicate_type"
+                )
+                raise AttestationUploadError(
+                    f"Multiple attestations for the same file with the same predicate "
+                    f"type: {predicate_type}",
```
di, line 202:
```suggestion
                    f"Multiple attestations for the same file with the same predicate "
                    f"type: {predicate_type!r} is not supported",
```

di (commented):
As an aside, would be nice if https://docs.pypi.org/attestations/#supported-attestations went into a bit more detail about this behavior. Can be in this PR or another.

warehouse/attestations/services.py
facutuesca: Here the error is not that `predicate_type` is not supported, but rather that multiple attestations with the same predicate type were uploaded.

di: Yes, I think that's still conveyed by the error message?

facutuesca:
> As an aside, would be nice if https://docs.pypi.org/attestations/#supported-attestations went into a bit more detail about this behavior. Can be in this PR or another.

Pushed a commit with the doc change

warehouse/attestations/services.py
facutuesca: But the suggestion says to add "{predicate_type} is not supported"

di: I think you're mis-reading the diff here.

facutuesca: Got it. I reworded it a bit to:
```python
f"Multiple attestations for the same file with the same "
f"predicate type ({predicate_type.value}) are not supported",
```

Also check the last commit for a minor change on how we handle the enum value for `predicate_type`

di (approved):

PR #17135, opened 2024-11-20
feat(admin): update all open malware reports on removal

miketheman:
When viewing a single malware report for a project, removing it will now update all other open malware observations to ensure we don't "miss" open Help Scout conversations.


warehouse/admin/views/malware_reports.py
```diff
@@ -306,20 +306,28 @@ def remove_malware_for_project(request):
...
-    # Each Action is a dictionary entry keyed by timestamp
-    observation.actions[int(now.timestamp())] = {
-        "actor": request.user.username,
-        "action": "remove_malware",
-        "created_at": str(now),
+    # Obtain all related, un-actioned Malware Observations to update.
+    project_malware_observations = {
+        obs
+        for obs in project.observations
+        if obs.kind == "is_malware" and not obs.actions
```
ewdurbin, line 313: this assumes any action has occurred. considering the situation where an initial report is responded to by quarantining quickly for later analysis, then on further inspection it's noted that the project is malware and should be removed, the quarantine action would interfere with the bulk action, right? that's probably OK?

miketheman: 5e58959c0

ewdurbin (approved):

PR #17147, opened 2024-11-21
fix: prevent html tags from input

miketheman:


warehouse/forms.py
```diff
@@ -10,13 +10,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
+import typing as t
```
di, line 15: Why the alias?

miketheman: It's a common way to shorten the overall line length. If this is the sticking point, I'll revert it, but I've got a long branch on type hints that makes heavy use of it.

miketheman: I may also change my stance later, after reviewing current practices a bit more, but that'll delay getting this out.

warehouse/forms.py
```diff
@@ -75,6 +84,21 @@ def __call__(self, form, field):
...
+    Validate the field to ensure that it does not contain any HTML tags.
+    """
+
+    def __init__(self, message: str | None = None):
+        if message is None:
+            message = "HTML tags are not allowed"
+        self.message = message
+
+    def __call__(self, form: BaseForm, field: Field):
+        if escape(field.data) != field.data:
```
di, line 98: This is going to catch a lot of things that aren't HTML, basically any use of `<`, `>`, `'`, `"`, or `&` will effectively be prohibited from being included. I think that might be too broad?

miketheman: I've updated to use nh3 on only HTML strings.

warehouse/forms.py
miketheman: Removed alias

pyproject.toml
```diff
@@ -25,7 +25,7 @@ parallel = true
 exclude_lines = [
   "pragma: no cover",
   "class \\w+\\(Interface\\):",
-  "if (typing\\.)?TYPE_CHECKING:",
+  "if (typing\\.|t\\.)?TYPE_CHECKING:",
```
di, line 28:
```suggestion
  "if (typing\\.)?TYPE_CHECKING:",
```

warehouse/forms.py
```diff
@@ -75,6 +83,24 @@ def __call__(self, form, field):
...
+    def __init__(self, message: str | None = None):
+        if message is None:
+            message = "HTML tags are not allowed"
+        self.message = message
+
+    def __call__(self, form: BaseForm, field: Field):
+        # Override the default nh3.ALLOWED_TAGS to be an empty set
+        allowed_tags: set[str] = set()
+
+        if is_html(field.data) and field.data != clean(field.data, tags=allowed_tags):
```
di, line 100: Won't the second comparison always fail if `is_html` is true?

di: Seems like this could just be
```suggestion
        if is_html(field.data):
```

miketheman: That's correct. I'll remove the second condition, and this can be how we'd implement if if we _did_ want to allow any tags in the future.

PR #17154, opened 2024-11-22
Fix URL verification for GitHub/GitLab 

DarkaMaul:
Fixes https://github.com/pypi/warehouse/pull/16899#issuecomment-2494054726

(Thanks @miketheman )


miketheman (approved):
Thanks for the speedy fix!

Issue #16541, opened 2024-08-21
Move API documentation from Warehouse developer docs to PyPI user docs


di:
Documentation for our APIs and datasets currently lives in our developer-facing docs: https://warehouse.pypa.io/api-reference/index.html

However, these are really end-user docs, and should live at https://docs.pypi.org/ instead.

woodruffw:
Adding a roadmap:

- [ ] Index
    - [ ] Feeds
    - [ ] JSON API
    - [ ] Legacy API
    - [ ] Stats API
    - [ ] XML-RPC (N/A?)
    - [ ] Integration guide (N/A?)
    - [ ] BigQuery Datasets
    - [ ] Integrity API (i.e. PEP 740, not currently documented)

Once each of these is moved/added, it might make sense to add redirects from the old dev-docs URLs to the new user-docs ones.

di:
> XML-RPC (N/A?)

I think it would be wise to not bother porting these docs.

> Integration guide (N/A?)

I think this can probably get included in the top-level page at https://docs.pypi.org/api/

woodruffw:
All planned migrations are complete, so closing!

di:
Should we add redirects from the old URLs?

woodruffw:
> Should we add redirects from the old URLs?

Yep! I was trying to figure out a good way to do that in Sphinx but couldn't find one (there's a lot of ways to redirect to a _different_ Sphinx doc, but not an arbitrary URL). I'll do some more looking here.

woodruffw:
https://documatt.com/sphinx-reredirects/usage.html might do the trick; I'll play around with that.

woodruffw:
https://github.com/pypi/warehouse/pull/17213 has the redirects!

PR #17162, opened 2024-11-25
docs: move BigQuery to user docs

woodruffw:
Per #16541.

### DIFF
--- a/docs/user/api/bigquery.md
+++ b/docs/user/api/bigquery.md
@@ -1,0 +1,32 @@
+# BigQuery Datasets
+
+We use BigQuery to serve our public datasets. PyPI offers two tables whose
+data is sourced from projects on PyPI. The tables and its pertaining data are licensed
+under the [Creative Commons License].
+
+## Download Statistics Table
+
+*Table name*: `bigquery-public-data.pypi.file_downloads`
+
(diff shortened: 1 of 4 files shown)


docs/user/api/bigquery.md
```diff
@@ -0,0 +1,32 @@
+# BigQuery Datasets
+
+We use BigQuery to serve our public datasets. PyPI offers two tables whose
```
woodruffw, line 3: Thinking out loud: it might be nice/useful to link to @sethmlarson's [pypi-data](https://github.com/sethmlarson/pypi-data) as a local/cheaper alternative to BQ for some use cases. 

LMK if you'd like that; I've left it out so far since the current changes are mostly a straight conversion 🙂

di: I like the project but I'm concerned about two things here: 1) there are other similar projects, and we'd have to justify which ones we include in the docs & don't include 2) up-to-date-edness, e.g. looks like that project is missing the new `License-File` and `License-Expression` fields.

woodruffw:
> Might be surprising to find this under "API Reference", should we rename that to be "APIs & Datasets" or something similar?

SGTM!

di (approved):

PR #17171, opened 2024-11-25
docs: migrate RSS Feed docs to user docs

woodruffw:
This moves the RSS docs and removes an XML-RPC note under them.

See https://github.com/pypi/warehouse/issues/16541.

### DIFF
--- a/docs/dev/api-reference/feeds.rst
+++ b/docs/dev/api-reference/feeds.rst
@@ -1,69 +1,9 @@
 Feeds
 =====
 
-PyPI offers three RSS feeds, the `Newest Packages Feed`_, the `Latest Updates
-Feed`_, and the `Project Releases Feed`_. You can also call its APIs to get
-more details on project activity.
+.. important::
 
+  This API documentation has been migrated to a new page in
+  the `user documentation <https://docs.pypi.org/>`_:
 
-Newest Packages Feed
---------------------
-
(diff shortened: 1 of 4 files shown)


docs/dev/api-reference/feeds.rst
```diff
@@ -1,69 +1,9 @@
...
-  >>> recentchanges = client.changelog_since_serial(serial)
-  >>> for entry in recentchanges:
-  ...     print(entry)
-  ['openllm', '0.4.33.dev3', 1701280908, 'new release', 4601225]
-  ['openllm', '0.4.33.dev3', 1701280908, 'add py3 file openllm-0.4.33.dev3-py3-none-any.whl', 4601226]
-
-You could also request ``GET /simple/``, and record the ``ETag``, and
-then periodically do a conditional HTTP GET to ``/simple/`` with that
-ETag included. A 200 OK response indicates something has been added or
-removed; if you get a 304 Not Modified, then nothing has changed.
```
di, line 69: Can we stick this in the XML-RPC docs somewhere instead? It does seem useful even if XML-RPC on a whole is not.

woodruffw: Yep, doing now!

woodruffw: Done -- I've moved this section into `xml-rpc.rst`.

di (approved):

PR #17174, opened 2024-11-25
Remove 2.0 from list of supported metadata versions

dnicolodi:
The metadata version is first validated by `packaging` and then by the additional validation code in warehouse.forklift.metadata. Therefore, only metadata versions supported by `packaging` can be supported.

Metadata version 2.0 has never officially been codified. `packaging` does not accept 2.0 as a valid metadata version. To avoid confusion, remove 2.0 from the list of supported metadata versions.

While at it, describe this fact in a comment, and add an assert.


warehouse/forklift/metadata.py
```diff
@@ -32,8 +33,8 @@
 
 from warehouse.utils import http
 
-SUPPORTED_METADATA_VERSIONS = {"1.0", "1.1", "1.2", "2.0", "2.1", "2.2", "2.3", "2.4"}
-
+SUPPORTED_METADATA_VERSIONS = {"1.0", "1.1", "1.2", "2.1", "2.2", "2.3", "2.4"}
+assert all(ver in _VALID_METADATA_VERSIONS for ver in SUPPORTED_METADATA_VERSIONS)
```
di, line 37:
```suggestion
assert SUPPORTED_METADATA_VERSIONS.issubset(_VALID_METADATA_VERSIONS)
```

dnicolodi: It's much nicer this way 🙂 Thanks!

warehouse/forklift/metadata.py
```diff
@@ -32,8 +33,8 @@
 
 from warehouse.utils import http
 
-SUPPORTED_METADATA_VERSIONS = {"1.0", "1.1", "1.2", "2.0", "2.1", "2.2", "2.3", "2.4"}
-
+SUPPORTED_METADATA_VERSIONS = {"1.0", "1.1", "1.2", "2.1", "2.2", "2.3", "2.4"}
+assert SUPPORTED_METADATA_VERSIONS.issubset(_VALID_METADATA_VERSIONS)
```
di, line 37:
```suggestion
```

On second thought, I think we're OK here without the assertion. We generally don't want assertions to raise at runtime because this would cause everything importing this file to fail (e.g. all uploads) rather than just cause the upload with a bad version to fail (which will happen anyways).

warehouse/forklift/metadata.py
```diff
@@ -20,6 +20,7 @@
     _LIST_FIELDS,
     _RAW_TO_EMAIL_MAPPING,
     _STRING_FIELDS,
+    _VALID_METADATA_VERSIONS,
```
di, line 23:
```suggestion
```

di (approved):

PR #17204, opened 2024-12-02
Drop version from search results

### LINKED ISSUE #17179: PyPI caching packages too long (updated release still not showing up after 12 hours after upload)
I have upload my package `pp.server` in version 3.4.0 on 2024-11-26 at around 18:00 UTC and uploaded version 3.4.1 on 2024-11-27 at around  06:00 UTC.

When I look and search at PyPI, the old version 3.3.0.2 is still shown as most current version.

Image
Image

Is this an issue of over-caching or missing cache invalidation?

di:
This is often out of date due to slow re-indexing, which is confusing to users, and not particularly helpful in the search results, so just remove it.

Closes #17179.


miketheman (approved):
LGTM, since we leave the underlying model alone, and this is only for search results.

alexanderilyin:
Why re-indexing is slow?

PR #17213, opened 2024-12-03
docs: redirect all old API docs to new equivalents

woodruffw:
This uses `[sphinx-reredirects`](https://documatt.com/sphinx-reredirects/index.html) to redirect most of the `/api-reference/*` hierarchy to the new user docs.

The exception is `/api-reference/xml-rpc`, which was intentionally not migrated and is still present at its original URL. To ensure that URL is still discoverable, it's now direct linked at the top-level `index.rst` in the dev-docs, where `/api-reference/index.rst` was previously linked.


docs/dev/api-reference/index.rst
```diff
@@ -1,77 +0,0 @@
-API reference
-=============
-
-.. note::
-
-  API documentation is being moved to PyPI's
-  `user documentation site <https://docs.pypi.org/api/>`_.
-
-  Please see `issue #16541 <https://github.com/pypi/warehouse/issues/16541>`_
-  for more information and status updates.
```
di, line 10: I'm thinking we should keep this file for now and add a short note that the docs have moved. (This also gives usa. better place to put the XML-RPC docs rather than at top level).

woodruffw: Sounds good, I'll bring this page back.

di: (Just the page, not all the content, just to be clear)

docs/dev/api-reference/index.rst
```diff
@@ -0,0 +1,15 @@
...
+=============
+
+.. note::
+
+  All API documentation except for :doc:`xml-rpc` has been moved to the
+  `user documentation site <https://docs.pypi.org/api/>`_.
+
+Available APIs & Datasets
+-------------------------
+
```
di, line 11:
```suggestion
```

docs/dev/api-reference/index.rst
woodruffw: Done! I've put the index back in as a stub documenting the migration of everything except XML-RPC.

di (approved):

Issue #17217, opened 2024-12-04
API docs: document the token disclosure reporting integration


woodruffw:
Filing this as a tracking issue: the current token disclosure integration (which GitHub uses to report compromised PyPI API tokens back to PyPI for revocation) is generic and could be used by additional secret-reporting services.

However, that integration interface isn't documented, meaning that other potential integrators don't know where to start. The API docs could include docs on it!

CC @di

di:
It's not quite generic, it's pretty GitHub-specific as it relies on GitHub-specific headers and has GitHub's key URL baked in: https://github.com/pypi/warehouse/blob/c9231a69413f218adcc8c2236b78059f524eadb6/warehouse/integrations/github/views.py

We should decide if we should offer a generic API in this style, and have GitHub be an outlier, or whether we want to have a different endpoint per integrator.

woodruffw:
Gotcha, I didn't realize it was GitHub specific! I can document it as-is as a starting point, to inform a decision about whether each integrator should have its own endpoint or not.

di:
I think we can make this a generic-ish endpoint w/ a select list of integrators (some mapping from valid integrator headers to key verification URL) and ask any future potential integrators to align w/ GitHub's pattern here. That would mean adding a new integrator would only be a few line change, not an entirely new view/endpoint.

di:
Updates to make the API generic are in https://github.com/pypi/warehouse/pull/17218.

I think we can include this at https://docs.pypi.org/api/ with the caveat that it's only available to integrators on a case-by-case basis.

di:
PR with docs is here: https://github.com/pypi/warehouse/pull/17236

PR #17218, opened 2024-12-04
Generic token disclosure

di:
This makes our token disclosure endpoint generic, in preparation for user-facing docs & additional integrators in #17217.


warehouse/routes.py
```diff
@@ -538,14 +538,18 @@ def includeme(config):
...
 
+    # Integration URLs
+    config.add_route(
+        "integrations.secrets.disclose-token",
+        "/_/secrets/disclose-token",
+        domain=warehouse,
+    )
     config.add_route(
-        "integrations.github.disclose-token",
+        "integrations.github.disclose-token",  # For backwards compatiblity
```
ewdurbin, line 549: does this _need_ a different name/view_config?

ewdurbin (approved):

warehouse/routes.py
di: Yes, I don't think we can mount the same view twice at two different routes.

PR #17232, opened 2024-12-05
Don't install deploy dependencies for tests

DarkaMaul:
I believe `ddtrace` and `gunicorn` are not required when running tests.

Importing `ddtrace` is somewhat costly (tests in a container with warehouse test dependencies installed):

-  Without ddtrace
```
> time pytest --help
real	0m2.521s
user	0m2.217s
sys	0m0.255s
```

- With ddtrace
```
> time pytest --help
real	0m3.094s
user	0m2.713s
sys	0m0.346s
```

This yielded to a modest improvement on my computer (time to run the test suite - cache deleted between runs):
- With the improvement: 19.20s
- Baseline: 19.88s

/cc @miketheman


DarkaMaul:
However, I believe the image in CI is built there :

https://github.com/pypi/warehouse/blob/c9231a69413f218adcc8c2236b78059f524eadb6/.github/workflows/ci.yml#L30-L38

But I don't know if it is safe to remove dependencies from there too or if this image is also used in other places.

ewdurbin:
Can the ddtrace plugin not be disabled in pytest?

DarkaMaul:
> Can the ddtrace plugin not be disabled in pytest?
It can but I _think_ it will still be loaded.

Let me check this and how it impacts the performance in the CI.

DarkaMaul:
There are two sets of complementary changes. The table below summarizes them, with the first line `build` representing the image [build](https://github.com/pypi/warehouse/blob/main/.github/workflows/ci.yml#L16) and [test](https://github.com/pypi/warehouse/blob/main/.github/workflows/ci.yml#L39) the testing job.

| | Baseline | Plugin disabled | Plugin not-installed |
|-|------|-----|----|
| `build` | / | 3m52s | 3m30s |
| `test` | 1m32s | 1m24s | 1m26s|

- The `build` changes (when the CI image is constructed) are not releveant because the image is cached by `depot` and then layers are re-used. We only benefit from not installing a package (and its dependencies) that is not used in the image.
- Disabling the plugin (either because it is not installed or because it is explicitly disabled) yields some nice improvements (about 8%)

Given the previous results - I think only disabling `trace` plugins in the tests is enough.

Dockerfile
```diff
@@ -177,8 +181,8 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     set -x \
     && pip --disable-pip-version-check \
             install --no-deps \
-                    -r /tmp/requirements/deploy.txt \
                     -r /tmp/requirements/main.txt \
+    								$(if [ "$TESTS" = "no" ]; then echo '-r /tmp/requirements/deploy.txt'; fi) \
```
miketheman, line 185: nitpick: mixing spaces with tabs on this line. Don't fix if removing ;)

pyproject.toml
```diff
@@ -85,6 +85,10 @@ addopts = [
   "--allow-hosts=localhost,::1,stripe",
   "--durations=20",
   "--numprocesses=auto",
+  # Disable ddtrace for tests
+  "-p no:ddtrace",
+  "-p no:ddtrace.pytest_bdd",
+  "-p no:ddtrace.pytest_benchmark",
```
miketheman, line 91: > Given the previous results - I think only disabling trace plugins in the tests is enough.

Then should this be the only change necessary, and restore the other changes proposed?

miketheman (changes requested):
I'm fine with disabling the plugin for now - as we don't yet [integrate deeply with Datadog via ddtrace to advertise test timings](https://docs.datadoghq.com/tests/setup/python/?tab=ciproviderwithautoinstrumentationsupport) yet, but may wish to in the future.

Let's keep the installation around, since that simplifies the overall stack and removing it doesn't have a material impact on overall CI timings.

Testing only the `-p` additions to `pyproject.toml` drops about 50% test time on my machine as well, confirming the other timings you've shared.

PR #17160, opened 2024-11-25
refactor: Remove Deprecated Bourbon/Neat from CSS stack

miketheman:
Bourbon 4.2.5 was released on Sep 17, 2015.
Neat 1.7.2 was released on Mar 13, 2015.

Warehouse vendored them on Nov 18, 2015 from a design repo, using `gulp` to handle the build process, replaced since by webpack.

Bourbon and Neat are now both deprecated libraries, with the final Neat release of 4.0.0 on Jul 10, 2019, and the final Bourbon release of 7.3.0 on Jan 23, 2023.

The authors have also helpfully provided a [Bourbon replacement guide](https://thoughtbot.com/blog/you-might-not-need-bourbon) to native CSS features.

## Steps

Here's a list of tasks to get us there:

- [x] replace `neat` with native CSS features
- [x] replace vendored `bourbon` 4.2.5 with package, reintroducing warnings from #12742
- [x] update to most recent `bourbon` 4.2.x version
- [x] update to most recent `bourbon` 4.3.x version, fix issues found
- [x] update to most recent bourbon 5.x version
- [x] update to most recent bourbon 6.x version
- [x] update to most recent and final bourbon 7.x version
- [x] Replace bourbon functions with native CSS features
    - [x] `all-text-inputs`
    - [x] `margin`
    - [x] `clearfix` - vendored, and deferred for now
    - [x] `border-width`
    - [x] `ellipsis`
- [x] Remove bourbon from the stack
- [ ] Celebrate! :tada:

## Bundle Size

Before: 252k uncompressed, 30k gzip, 24k brotli
After: 195k uncompressed, 29k gzip, 23k brotli

The CSS bundle size drops a little, largely due to removing the vendored files with custom license headers - since the rest of bourbon/neat were only used during compile-time, the resulting compiled bundles have the same functionality as before.

## Related links

- warehouse issue about [vendored imports](https://github.com/pypi/warehouse/issues/2144)
- [Bourbon Migrating from v4 to v5](https://www.bourbon.io/docs/migrating-from-v4-to-v5/)

PR #17237, opened 2024-12-06
fix: create a parent for column grids

### LINKED ISSUE #17233: Undesirable formatting on recovery code generation page
https://pypi.org/manage/account/recovery-codes/generate

Image

-----

https://pypi.org/manage/account/recovery-codes/burn

Image

miketheman:
`display: grid` needs to be at a parent level, and `col-half` didn't have a container per se, as the previous implementation was a computed column width, not part of a CSS Grid.

Fixes #17233
Refs: #17160

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -3885,7 +3885,7 @@ msgstr ""
 msgid "%(remaining)s unused"
 msgstr ""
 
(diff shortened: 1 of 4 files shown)


ewdurbin (approved):

ofek:
Thanks!

PR #17242, opened 2024-12-07
Add summaries for the security models of attestations and trusted publishers

mac-chaffee:
Hello! 

I recently wanted to learn more about trusted publishing and attestations but I struggled initially to understand the big-picture of the security goals and non-goals, and I had a lot of misconceptions initially.

So I added some summaries with my goal being to surface the answers to the questions I had initially, while trying to keep it simple. But that does mean I possibly over-simplified or glossed over important details, so feedback is welcome!

### DIFF
--- a/docs/user/attestations/security-model.md
+++ b/docs/user/attestations/security-model.md
@@ -4,6 +4,22 @@ title: Security Model and Considerations
 
 # Security model and considerations
 
+Since attestations are a framework for asserting certain facts about a package,
+the security model of attestations depends on which types of facts are being
+asserted.
+
+The most basic type of attestation (the type enabled by default in
+[`pypa/gh-action-pypi-publish`][gh-action-pypi-publish]) asserts that the
+package was built by an authorized build process such as a particular CI
+provider in a particular code repository. This is primarily designed to protect
+against modification of a package *after* it was built such as while it is
+stored in a package index.
+
+More advanced types of attestations can assert more things about the package
+such as whether it was tampered with *before* it was built, but these
+attestations require specialized build processes and are rarer as a result. See
+"[What about reproducible builds?]" in the SLSA FAQ.
+
 ## General considerations
 
 PyPI's support for attestations is built on top of a combination
@@ -14,27 +30,27 @@ See below for security considerations for each.
 
 !!! note
 
-    TL;DR: An attestation will tell you **who** produced a
-    PyPI package, but not **whether** you should trust them.
+    TL;DR: An attestation will tell you **where** a PyPI package came from, but
+    not **whether** you should trust it.
 
 Like with all signing schemes, it's tempting to treat the *presence* of
 an attestation as proof of a package's *trustworthiness*.
 
-However, this is **not** what an attestation (or any kind of signature)
-conveys. At its core, a valid signature for some identity conveys a
-proof of **authenticity**: if the signature is verifiable against a key,
-then we know that the identity that controls that key produced the signature
-and is saying *something* about the signed-over data.
+However, this is **not** what an attestation (or any kind of signature) conveys.
+At its core, a valid signature for an identity on a package conveys a proof of
+access to that identity while the package was built.
 
-In other words: a valid signature asserts **authenticity**, but it does **not**
-tell the verifying party (e.g., a user installing packages from PyPI) **whether**
-they should trust the identity that holds the key.
+In other words: a valid signature does **not** tell the verifying party (e.g., a
+user installing packages from PyPI) **whether** they should trust the identity
+that holds the key or whether they should trust that malicious/vulnerable code
+was added before or during the build.
 
 As a concrete example: PyPI serves `sampleproject-4.0.0.tar.gz`, which is
 [attested] by [pypa/sampleproject] on GitHub. This is a proof that
-`sampleproject-4.0.0.tar.gz` is authentic for that identity, but it does **not**
-tell the user **whether** they should trust [pypa/sampleproject]. To determine
-trust, the user *must* make a trust decision about [pypa/sampleproject].
+`sampleproject-4.0.0.tar.gz` came from that identity unmodified, but it does
+**not** tell the user **whether** they should trust [pypa/sampleproject]. To
+determine trust, the user *must* make a trust decision about
+[pypa/sampleproject].
 
 This trust decision can have a time dimension: a user might decide to trust
 [pypa/sampleproject] because it was the first identity seen for the
@@ -115,6 +131,10 @@ verifiable*.
 PyPI's attestations feature makes full use of these trust-reduction techniques:
 attestations are not considered verified unless they include an inclusion proof
 from Rekor, as well as an inclusion proof from Fulcio's CT log.
+
+[gh-action-pypi-publish]: https://github.com/pypa/gh-action-pypi-publish
+
+[What about reproducible builds?]: https://slsa.dev/spec/v1.0/faq#q-what-about-reproducible-builds
 
 [Trusted Publishing]: /trusted-publishers/
 
--- a/docs/user/trusted-publishers/security-model.md
+++ b/docs/user/trusted-publishers/security-model.md
@@ -4,21 +4,43 @@ title: Security Model and Considerations
 
 # Security model and considerations
 
+Trusted publishing is primarily designed to be a more secure alternative than
+the long-lived API tokens that have traditionally been used for publishing to
+PyPI.
+
+In recent years, theft of credentials such as API tokens has [played a major
+role in cyber attacks]. The reason for this is the unfortunate reality that
+managing credentials can be unexpectedly difficult. Trusted publishing addresses
+this risk by using short-lived tokens instead of long-lived tokens that can
+easily get misplaced, leaked in logs, stolen by malware, or any number of other
+ways.
+
+However, it is important to still be aware of the kinds of risks that
+trusted publishing does not cover. You should think of trusted publishing as one
+tool in the toolbelt for securing packages.
+
 ## General considerations
 
-While more secure than passwords and long-lived API tokens, OIDC publishing
-is not a panacea. In particular:
+* Trusted publishing asserts that a package was published to PyPI by someone
+  or something that had access to an API token that was created within 15
+  minutes of a valid OIDC authentication flow. It does not assert anything about
+  the safety of the code or the trustworthiness of the authors.
+ 
+* Trusted publishing does not address whether the package has been modified
+  before or after it was built. [Attestations] can address those risks.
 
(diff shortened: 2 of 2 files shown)


woodruffw:
Hey @mac-chaffee! Sorry for the delay here -- I had this flagged for myself, but it looks like I lost it in the stacks. I'll do a review of these changes today!

docs/user/attestations/security-model.md
```diff
@@ -4,6 +4,22 @@ title: Security Model and Considerations
...
+Since attestations are a framework for asserting certain facts about a package,
+the security model of attestations depends on which types of facts are being
+asserted.
+
+The most basic type of attestation (the type enabled by default in
+[`pypa/gh-action-pypi-publish`][gh-action-pypi-publish]) asserts that the
+package was built by an authorized build process such as a particular CI
+provider in a particular code repository. This is primarily designed to protect
+against modification of a package *after* it was built such as while it is
+stored in a package index.
```
woodruffw, line 16: Nitpick: the default attestation asserts that the project was *published* by a particular identity, not necessarily built by it. This distinction is subtle but important: someone _could_ upload their own built to a GitHub Actions workflow and then publish it from there, which would be considered "valid" from a publish attestation perspective 🙂 

I've also expanded on the purpose of this type of attestation: providing integrity/authenticity is one of the purposes, but another equally important one is allowing users to verifiably observe changes in a project's publish provenance!

```suggestion
The most basic type of attestation (the type enabled by default in
[`pypa/gh-action-pypi-publish`][gh-action-pypi-publish]) asserts that the
project was published by an authorized publisher such as a particular CI
provider in a particular code repository.

This type of attestation has two main purposes:

- It protects against modification of a project *after* it was built,
  such as while it is stored in a package index or mirror.
- It allows verifying parties to observe *changes* to the project's Trusted Publisher.
  For example, a verifying party might observe that a project's new releases
  are coming from a different Trusted Publisher, indicating that control
  of the project may have changed hands maliciously.
```

(Some more details here as well: <https://docs.pypi.org/attestations/publish/v1/>)

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
 
 # Security model and considerations
 
+Trusted publishing is primarily designed to be a more secure alternative than
```
woodruffw, line 7: Nitpick: we generally capitalize it as a term of art:

```suggestion
Trusted Publishing is primarily designed to be a more secure alternative to
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
+the long-lived API tokens that have traditionally been used for publishing to
+PyPI.
+
+In recent years, theft of credentials such as API tokens has [played a major
+role in cyber attacks]. The reason for this is the unfortunate reality that
+managing credentials can be unexpectedly difficult. Trusted publishing addresses
```
woodruffw, line 13:
```suggestion
managing credentials can be unexpectedly difficult. Trusted Publishing addresses
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
+this risk by using short-lived tokens instead of long-lived tokens that can
+easily get misplaced, leaked in logs, stolen by malware, or any number of other
+ways.
+
+However, it is important to still be aware of the kinds of risks that
+trusted publishing does not cover. You should think of trusted publishing as one
```
woodruffw, line 19:
```suggestion
Trusted Publishing does not cover. You should think of Trusted Publishing as one
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
+tool in the toolbelt for securing packages.
+
 ## General considerations
 
-While more secure than passwords and long-lived API tokens, OIDC publishing
-is not a panacea. In particular:
+* Trusted publishing asserts that a package was published to PyPI by someone
+  or something that had access to an API token that was created within 15
+  minutes of a valid OIDC authentication flow. It does not assert anything about
+  the safety of the code or the trustworthiness of the authors.
```
woodruffw, line 27: The wording here is a little confusing IMO. Suggestion:

```suggestion
* Trusted Publishing uses short-lived API tokens that expire 
  no more than 15 minutes after the OIDC flow that authorizes them.
  Just like normal API authentication, Trusted Publishing
  does not assert the safety of the code or the trustworthiness
  of its authors.
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
+ 
+* Trusted publishing does not address whether the package has been modified
```
woodruffw, line 29:
```suggestion
* Trusted Publishing does not address whether the package has been modified
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
+  successfully intercepts one can use it to generate API tokens until it
+  expires.
 
-* Configuring a trusted publisher means establishing trust in a particular piece
-  of external state (such as a GitHub Actions workflow); that state **must not**
-  be controllable by untrusted parties.
+* Configuring a trusted publisher means trusting in the auth system of a
+  particular OIDC provider (such as GitHub Actions). Trusted publishing relies
+  on that auth system not being contolled by attackers and not containing
+  vulnerabilities.
```
woodruffw, line 43:
```suggestion
* Configuring a Trusted Publisher means trusting an identity provider (IdP),
  such as GitHub Actions. Trusted Publishing relies on the integrity of that
  IdP and the actors that are authorized to use it. In practice, this means
  that users of Trusted Publishing must protect and secure the CI/CD workflows
  that they register as Trusted Publishers, as weaknesses in those workflows
  can be equivalent to credential compromise.
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,21 +4,43 @@ title: Security Model and Considerations
...
 
 In summary: treat your trusted publishers *as if* they were API tokens. If you
```
woodruffw, line 45: (This one is my fault from an earlier version of the doc!)

```suggestion
In summary: treat your Trusted Publishers *as if* they are API tokens. If you
```

docs/user/trusted-publishers/security-model.md
```diff
@@ -258,6 +280,10 @@ own security model and considerations.
       access the OIDC token to a bare minimum. This prevents both accidental
       and malicious disclosure.
 
+[played a major role in cyber attacks]: https://verizon.com/dbir
```
woodruffw, line 283: This HTTPS link doesn't connect successfully for me, but the HTTP version redirects to <https://www.verizon.com/business/resources/reports/dbir/>. Perhaps we should update it to that?

woodruffw: Looks like the actual report is [here](https://www.verizon.com/business/resources/reports/2024-dbir-data-breach-investigations-report.pdf), but it's emailwalled. I think we should probably find an alternative source/cite, since PyPI's docs shouldn't link people to walled sources 🙂 

Here are some candidates:

* <https://therecord.media/cisa-cyberattacks-using-valid-credentials>
* <https://www.cisa.gov/sites/default/files/2023-07/FY22-RVA-Analysis%20-%20Final_508c.pdf>

mac-chaffee: The email-walled version is unfortunately the one the authors ask that we link to. But that first link looks good, will replace it.

woodruffw (approved):
Thanks @mac-chaffee! These improvements look great to me.

mac-chaffee:
Thanks for the suggestions! I rebased and edited the author to remove the unverified commit in case that causes issues

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,23 +4,48 @@ title: Security Model and Considerations
...
+PyPI.
 
-* Short-lived API tokens are still sensitive material, and should not be
-  disclosed (ideally not at all, but certainly not before they expire).
+In recent years, theft of credentials such as API tokens has [played a major
+role in cyber attacks]. The reason for this is the unfortunate reality that
+managing credentials can be unexpectedly difficult. Trusted Publishing addresses
+this risk by using short-lived tokens instead of long-lived tokens that can
+easily get misplaced, leaked in logs, stolen by malware, or any number of other
+ways.
```
di, line 16: I think the key distinction we should make here is that long-lived tokens need to be stored and do not quickly expire. That somewhat reduces the likelihood that they will "misplaced, leaked in logs, stolen by malware" (although it's still possible) but _also_ reduces the impact if they are compromised.

di (commented):
Looks great, thanks for improving these! Just one note from me.

docs/user/trusted-publishers/security-model.md
mac-chaffee: Good point, let me know what you think of the new version I just pushed 👍

docs/user/trusted-publishers/security-model.md
```diff
@@ -4,23 +4,49 @@ title: Security Model and Considerations
...
+the long-lived API tokens that have traditionally been used for publishing to
+PyPI.
+
+In recent years, theft of credentials such as API tokens has [played a major
+role in cyber attacks]. The reason for this is the unfortunate reality that
+managing credentials can be complicated and risky. Trusted Publishing reduces
+this risk by using short-lived tokens instead of long-lived tokens. Short-lived
+tokens are less likely to be misplaced, leaked in logs, or stolen by malware.
+If short-lived tokens are leaked, they only give attackers a narrow time window
+to exploit the leaked token, which minimizes the potential damage.
```
di, line 17: I think this is still missing the distinction that when used with Trusted Publishing short-lived tokens are ephemeral and don't need to be stored.

mac-chaffee: Oh I thought you wanted emphasis on the fact that short-lived tokens only reduce but don't eliminate the risk. How about this?

> Short-lived tokens are less likely to be misplaced, leaked in logs, or stolen by malware since they don't have to be stored.

di: Perfect!

di, line 17:
```suggestion
tokens are less likely to be misplaced, leaked in logs, or stolen by malware
since they don't have to be stored. Additionally, if short-lived tokens are
leaked, they only give attackers a narrow time window to exploit the leaked
token, which minimizes the potential damage.
```

di (approved):

Issue #6091, opened 2019-06-25
Soft Deletes Meta Issue


ewdurbin:
Currently, all deletes are good and proper Deletions.

This imposes quite a bit of risk for moderator activities, causes confusion and frustration for maintainers, and blocks us from implementing some important features down the line.

This issue is a "Meta Issue" for tracking progress on soft deletion implementation for models for which it makes sense.

Currently:

- [ ] #4440 Projects/Releases/Files

benjaoming:
Is this issue okay to use for feedback/input? Otherwise, I'll happily delete and post elsewhere.

Here is a scenario that might cause frustrations in the future, as the pattern might propagate: Bots (like dependabot) + Pipenv.lock + package deletion = :confounded:

[This is already happening](https://github.com/learningequality/studio/issues/1780).

In this scenario, bots create updates (PRs) for Pipenv.lock that are hard for humans to read, they quickly and automatically pass the build, and once they are merged, they're broken -- because the package take a while for a human to delete, and PRs take a while for humans to merge.

There are some ingredients in the deletion that can perhaps be put to use?

* Statistics = 0-10 downloads: If (almost) no one ever used the release or project, then fine just delete it. The threshold should of course be defined and adapted over time.
* <30 minutes ago: Soft deletion is possible but the number of downloads are visible to the one deleting the release with a big warning.
* >30 minutes: You cannot delete the package. It's considered "in the wild" and only a malicious package will be marked as deleted.
* Marked as deleted: Future pip should be able to raise an exception "This release was deleted by the owner or package index administrators".

di:
Hi @benjaoming, thanks for the feedback. I think that this issue would be resolved by completing https://github.com/pypa/warehouse/issues/5837 and implementing "yanking" as an alternative to performing a deletion (regardless of whether it's a hard delete or a soft delete), as you would still be able to pin to a yanked package without issue.

ewdurbin:
https://github.com/flipbit03/sqlalchemy-easy-softdelete seems to be a potential option for implementing soft-deletes.

miketheman:
Adding related conversation: https://discuss.python.org/t/stop-allowing-deleting-things-from-pypi/17227
and https://github.com/pypi/warehouse/issues/11841

woodruffw:
[...]

PR #17257, opened 2024-12-09
Add published field to Release

DarkaMaul:
This PR introduces the following changes :

- a new boolean `published` field to the Release model 
- ~a migration to update every existing releases and set their published value to be the creation time~

Part of https://github.com/pypi/warehouse/issues/17230

## Some notes:

~I dislike how I changed the default `Release` creation call to force the publish time to be set to a timestamp, but I did not find a better solution. This created a new quirk: every time someone wants to create a new Release, they must make sure the `published` field is set.~
 
~The problem boils down to not setting a value for `publish` at the release creation, which would either semantically mean the time is not set (the solution I chose) or use a default value (e.g., the current timestamp). When choosing the latter case, we would need a placeholder value on the Python side (e.g., `published=False`) and to convert this placeholder to an explicit `null` value ( using [forcing-null-on-a-column-with-a-default](https://docs.sqlalchemy.org/en/13/orm/persistence_techniques.html#forcing-null-on-a-column-with-a-default) )~

The new implementation uses a boolean field.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -11,7 +11,7 @@ jobs:
   zizmor:
     # Advanced Security is not enabled on private repositories
     if: github.repository == 'pypi/warehouse'
-    name: Zizmor latest via Cargo
+    name: Zizmor
     runs-on: ubuntu-24.04
     permissions:
       security-events: write
@@ -23,12 +23,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           persist-credentials: false
-      - name: Setup Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-      - name: Get zizmor
-        run: cargo install zizmor
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: '.python-version'
       - name: Run zizmor
-        run: zizmor --format sarif . > results.sarif
+        run: pipx run zizmor --format sarif . > results.sarif
       - name: Upload SARIF file
         uses: github/codeql-action/upload-sarif@v3
         with:


di:
Should this follow the same pattern we have for yanking, where there are yanking events with timestamps, and we have a `yanked` field on the model that is a boolean instead?

DarkaMaul:
> Should this follow the same pattern we have for yanking, where there are yanking events with timestamps, and we have a `yanked` field on the model that is a boolean instead?

I like the idea as it would also not modify how new Releases are created.

woodruffw (approved):
LGTM, nice work @DarkaMaul!

Given that `published` is currently invariant as `True`, this shouldn't cause any observable index changes. However, for the follow-up where we make it controllable, I think we should add some integration tests to make sure the index responses are what we expect.

warehouse/legacy/api/json.py
```diff
@@ -62,7 +62,10 @@ def _json_data(request, project, release, *, all_releases):
             )
         )
         .outerjoin(File)
-        .filter(Release.project == project)
+        .filter(
+            Release.project == project,
+            Release.published.is_(True),
+        )
```
di, line 68: I am very wary of this pattern, where the default gives us all releases and we have to manually filter those that are published. I think it's inevitable that we will miss this filter somewhere and this will result in the wrong releases being returned.

I think instead we need something that behaves like https://github.com/flipbit03/sqlalchemy-easy-softdelete, which makes all queries contain this filter by default, and requires manually querying for the _unpublished_ releases instead.

(There is some additional discussion about this in https://github.com/pypi/warehouse/issues/6091 because we have the same fundamental issue with any mechanics for soft deletes as well.)

DarkaMaul: Great call! I've updated the PR to use an SQLAlchemy event listener and intercept SELECT statements to rewrite them with an exclusion clause.

warehouse/packaging/models.py
```diff
@@ -1041,6 +1047,21 @@ def ensure_monotonic_journals(config, session, flush_context, instances):
             return
 
 
+@db.listens_for(db.Session, "do_orm_execute")
+def filter_staged_release(config, state: ORMExecuteState):
+    if (
+        state.is_select
+        and not state.is_column_load
+        and not state.is_relationship_load
+        and not state.statement.get_execution_options().get("include_staged", False)
```
DarkaMaul, line 1056: FLAG: This is added to allow querying for the `published=False` query but is not used in this PR.

warehouse/migrations/versions/bd2bf218e63f_add_published_field.py
```diff
@@ -0,0 +1,38 @@
...
+    op.add_column(
+        "releases",
+        sa.Column(
+            "published", sa.Boolean(), server_default=sa.text("true"), nullable=False
+        ),
+    )
+
+
+def downgrade():
+    op.drop_column("releases", "published")
```
di, line 38: Let's put this migration into a separate PR. We might want to increase the timeouts around this as well... I foresee issues adding this new column to the table (there are 6.5M rows currently).

DarkaMaul: 👍 I moved this part to https://github.com/pypi/warehouse/pull/17472 .

DarkaMaul:
As this PR is now based on top of https://github.com/pypi/warehouse/pull/17472, I've converted it as a Draft until the other one is merged.

woodruffw (approved):

di:
I'm holding off on merging this until we think through https://github.com/pypi/warehouse/issues/17230#issuecomment-2643839592 a little bit more (specifically, the last bit about project/release/file lifecycles needs a bit more thought here I think)

PR #17259, opened 2024-12-09
Expose file events in JSON API

di:
This PR exposes a subset of the fields on every `File.Event` in `file.events`, specifically:

* `File.Event.tag` (all events)
* `File.Event.time` (all events)
* `File.Event.additional["submitted_by"]` (`file:add` events)
* `File.Event.additional["publisher_url"]` (`file:add` events)
* `File.Event.additional["uploaded_via_trusted_publisher"]` (`file:add` events)

Currently in practice this just exposes the `file:add` event because this is the only `File.Event` that will be available when a file is present, but the implementation will at a minimum expose `File.Event.tag` and `file.Event.time` for any future events that might exist.

### DIFF
--- a/tests/unit/legacy/api/test_json.py
+++ b/tests/unit/legacy/api/test_json.py
@@ -291,6 +291,7 @@ class TestJSONProject:
                         "requires_python": None,
                         "yanked": False,
                         "yanked_reason": None,
+                        "events": [],
                     }
                 ],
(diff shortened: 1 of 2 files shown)


di:
Closing this for now, I'm not sure there's a good way to shoehorn this into the existing API in a performant way.

PR #17262, opened 2024-12-10
Celery tuning

ewdurbin:
Primarily focused on being able to explicitly set `--concurrency` in prod.

Unfortunately Celery doesn't respect an environment variable for this, so allow for setting additional Celery flags via the environment.

Also sets the dev worker to a concurrency of 1 to fix a locking issue with running Beat in the same process leading to many tasks being enqueued instead of one.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -2,7 +2,7 @@ release: bin/release
 web: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-api: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-uploads.conf.py warehouse.wsgi:application
-worker: bin/start-worker celery -A warehouse worker $EXTRA_CELERY_ARGS -Q default -l info --max-tasks-per-child 1024
+worker: bin/start-worker celery -A warehouse worker --concurrency=${CELERY_CONCURRENCY:-1} -Q default -l info --max-tasks-per-child 1024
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -2,7 +2,7 @@ release: bin/release
 web: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-api: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-prod.conf.py warehouse.wsgi:application
 web-uploads: bin/start-web ddtrace-run python -m gunicorn.app.wsgiapp -c gunicorn-uploads.conf.py warehouse.wsgi:application
-worker: bin/start-worker celery -A warehouse worker -Q default -l info --max-tasks-per-child 32
+worker: bin/start-worker celery -A warehouse worker $EXTRA_CELERY_ARGS -Q default -l info --max-tasks-per-child 1024
```
di, line 5: Should this be something like
```suggestion
worker: bin/start-worker celery -A warehouse worker --concurrency $CELERY_CONCURRENCY -Q default -l info --max-tasks-per-child 1024
```
instead?

miketheman (approved):

di (approved):
Seems fine regardless of the comment

Procfile
ewdurbin: Downside here is we _must_ set a value of $CELERY_CONCURRENCY, but its an option.

ewdurbin: If no value is set, we'll get:

```
worker-1        | Error: Invalid value for '-c' / '--concurrency': '--beat' is not a valid integer.
```

ewdurbin: lemme see if bash expansion works in procfiles lol.

ewdurbin: It does. reworking!

di (approved):

PR #17269, opened 2024-12-11
perf: speed up user profile page

miketheman:
For users with a large amount of projects, the template iteration through
`projects` to fetch their latest `release` details turns into an n+1
query, and can time out on larger user profile pages.

Refactor the query to perform most of the logic to fetch only the data
necessary to render the template correctly, removing the need to load
full entities.


miketheman:
From a local dev, largest user is OCA (similar to prod), with only 193 projects, the results are demonstrative.

Before:
Screenshot 2024-12-11 at 16 06 51

After:
Screenshot 2024-12-11 at 16 20 25

ewdurbin (approved):

PR #970, opened 2016-02-14
Design cleanup

nlhkabu:
Minor cleanups:

- Focus search bar on desktop
- Use svg rather than png for logo and block images
- Add favicon

PR #17277, opened 2024-12-12
feat: add a favicon route on the root

miketheman:
Many browsers may not respect the inline HTML directive on where to find a favicon file, so add a route to serve it directly and not respond with a 404 Not Found.

Refs: #970


miketheman:
This might not be the best way to do it, but I couldn't find a clean way to use Pyramid's built-in static handling for a single file.
Since we use Whitenoise to parse our manifest and provide paths to the compiled files, and `whitenoise_add_files` directive works on directories, this is what I came up with.

If there's a better way, I'm all for it!

di:
Seems like Pyramid's recommendation is to do pretty much exactly this: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/assets.html#registering-a-view-callable-to-serve-a-static-asset

di (approved):

Issue #17241, opened 2024-12-06
Warning or error for Trusted Publisher users when GitHub Environment claim is included but not checked


sethmlarson:
**What's the problem this feature will solve?**

Creating a Trusted Publisher for GitHub optionally allows adding a "GitHub Environment"
to be checked during the OIDC->API token exchange. In the documentation it's recommended to use
a GitHub Environment, but users may not know they need to update the Trusted Publisher in addition to the workflow definition to get the benefits of a GitHub Environment requirement.

**Describe the solution you'd like**

Instead of silently passing when a GitHub Environment claim is included during the OIDC->API token exchange, either send a warning as an email or error out with a descriptive error message, such as:

`GitHub Environment in use without being required on Trusted Publisher: <link to docs how to remediate>`

**Additional context**

This was recently relevant to the Ultralytics supply-chain attack.

cc @woodruffw

woodruffw:
Thanks @sethmlarson! Fully agreed on sending a warning email in this case, and this cleanly falls into the engineering scope my team already has for lifecycle improvements (funded by A/O).

facutuesca:
I'm looking into this now

facutuesca:
Do we want to warn every time this happens, or only once?

sethmlarson:
Yeah once per release makes sense to me! Users can make the emails "go away" by either amending their Trusted Publisher or adding the GitHub Environment so there's no risk of unresolvable annoying emails.

facutuesca:
> Yeah once per release makes sense to me! Users can make the emails "go away" by either amending their Trusted Publisher or adding the GitHub Environment so there's no risk of unresolvable annoying emails.

What is the difference between those 2 (amending the TP or adding the environment) ? I thought there was only one possible solution: Removing the TP that has no environment and adding a new one that does.

di:
> Remove the environment from GHA entirely (not ideal, but they could do this)

Definitely don't want people to do this, so let's be sure that it's easier to add an environment to a trusted publisher. Maybe we should support doing that directly, rather than having to recreate it? Maybe via a magic link we send them in the email?

woodruffw:
Yeah, great point -- I think a magic link similar to the one we have for one-shot configuration would work well here!

webknjaz:
[...]

PR #17281, opened 2024-12-14
Warn users when GitHub/GitLab environments are not checked during Trusted Publishing

facutuesca:
## What is this?
This PR adds two things:
1. Allow users to constrain the environment of an existing GitHub/GitLab Trusted Publisher.
_This means changing a publisher that allows any environment so that it only accepts one._
2. Send project owners an email when a Trusted Publisher that allows any environment is used within a specific GitHub/GitLab environment
    _The email suggests constraining the Trusted Publisher so that it only accepts the environment which was just used. The email also contains a link to the PyPI page to constrain the environment to the one used, so that project owners can just follow this link and confirm the change._

This fixes https://github.com/pypi/warehouse/issues/17241

### UI/UX
- The UI to constrain the environment can only be accessed via a URL like the following:
 ```
https://pypi.org/manage/project/$MY_PROJECT/settings/publishing/?constrained_publisher_id=$PUBLISHER_ID&constrained_environment=$NEW_ENVIRONMENT_NAME#constrain_environment-modal
```
- That URL is the one included in the email, and it shows the following confirmation modal:

image

- Constraining the environment will only work if all of the following are true:
  - The logged-in user is an owner (has write permissions) for `$MY_PROJECT`
  - The Trusted Publisher with ID `$PUBLISHER_ID` is a valid GitHub/GitLab publisher with no configured environment
  - The Trusted Publisher is associated to `$MY_PROJECT`
- Internally, constraining the environment first creates a new TP which is a copy of the original one except for the new environment. Then, it removes the association between the original TP and the project. If this project was the only one associated to the original TP, the TP is deleted.

### Email
- The email is sent only when all of the following are true:
  - An OIDC token is exchanged for a PyPI token using a GitHub/GitLab Trusted Publisher
  - The Trusted Publisher is configured to accept any environment
  - The OIDC token contains a non-empty GitHub/GitLab environment claim
  - The Trusted Publisher is associated to only one project**

 ** This is done in order to avoid confusion and edge cases: if a single Trusted Publisher is configured for multiple projects, to send an email during the token exchange we would need to send it to the owners of all the projects, which might be confusing if the token exchange was for the upload of a single project's release.

### Email
image

### Confirmation modal

image

### Action successful
image

cc @woodruffw @sethmlarson @di

### CODE REVIEW

File: warehouse/manage/views/__init__.py [hunk 1]
```diff
@@ -1435,6 +1436,117 @@ def manage_project_oidc_publishers_prefill(self):
...
+                "submitted_by": self.request.user.username,
+            },
+        )
+
+        # Then, we remove the old trusted publisher from the project
+        # and, if there are no projects left associated with the publisher,
+        # we delete it entirely.
+        self.project.oidc_publishers.remove(publisher)
+        if len(publisher.projects) == 0:
+            self.request.db.delete(publisher)
```
woodruffw, line 1526: I might be missing something, but can we simplify this by modifying the publisher rather than creating a new one and deleting the old one?

facutuesca: It wouldn't cover the case where the publisher is associated to more than one project: since the magic link is associated to a single project, it should only constraint the publisher for that project. And modifying a publisher would modify it for all associated projects.

woodruffw: Ah right, completely forgot about that case! It might be good to leave a comment to that effect (the current comment explains it indirectly, but an explicit one might stop an over-eager refactor 🙂)

warehouse/templates/email/environment-ignored-in-trusted-publisher/body.html
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-#}
+{% extends "email/_base/body.html" %}
+
+{% set domain = request.registry.settings.get('warehouse.domain') %}
+
+{% block content %}
+  <p>
+    {% trans publisher_name=publisher.publisher_name, project_name=project_name, environment_name=environment_name,
```
di, line 20: These emails shouldn't be marked as translated, we don't know the locale of the user that we're sending them to.

warehouse/manage/views/__init__.py
```diff
@@ -1435,6 +1436,117 @@ def manage_project_oidc_publishers_prefill(self):
 
         return self.manage_project_oidc_publishers()
 
+    @view_config(
+        request_method="GET", request_param=ConstrainEnvirormentForm.__params__
+    )
+    def manage_project_oidc_publisher_constrain_environment(self):
```
di, line 1442: I don't think we should have a view that can be triggered from a GET request that modifies the publisher (or anything, really). Even though this is a relatively minor change, it could be used semi-maliciously but also breaks a general pattern we have of links being non-mutating.

Ideally this would go from email link -> `GET` interstitial that confirms the changes -> `POST` that applies the changes (the last two can be the same view, varied by the request method).

warehouse/templates/email/environment-ignored-in-trusted-publisher/body.html
facutuesca: fixed!

warehouse/manage/views/__init__.py
facutuesca: done! I added a modal for confirming the changes, which then applies them POST

image

(see the updated PR description, and the last commit)

warehouse/templates/manage/project/publishing.html
```diff
@@ -438,4 +438,25 @@ <h2 class="no-bottom-padding">{% trans %}Add a new publisher{% endtrans %}</h2>
...
+{% set confirm_button_label = gettext("Constrain environment") %}
+{% set slug = confirm_name.lower().replace(' ', '-') + '-modal' %}
+{% set extra_fields %}
+  <input name="constrained_publisher_id" type="hidden" value="{{ request.params.constrained_publisher_id }}">
+  <input name="constrained_environment_name" type="hidden" value="{{ constrained_environment_name }}">
+{% endset %}
+{% set extra_description %}
+<p>{% trans %}Restrict Trusted Publisher environment to: {% endtrans %}<strong>{{ constrained_environment_name }}</strong></p>
+  <ul class="no-bullets no-left-margin">
+    <li><input class="inline-checkbox" type="checkbox" data-action="input->confirm#check" data-confirm-target="checkbox"> {% trans %}I understand that this Trusted Publisher will only allow uploads frow CI/CD workflows running in {{ constrained_environment_name }}{% endtrans %}</li>
```
di, line 455:
```suggestion
    <li><input class="inline-checkbox" type="checkbox" data-action="input->confirm#check" data-confirm-target="checkbox"> {% trans %}I understand that this Trusted Publisher will only allow uploads frow CI/CD workflows running in the '{{ constrained_environment_name }}' environment.{% endtrans %}</li>
```

di (commented):
LGTM otherwise

di:
Looking good. My only additional thought is that this doesn't really handle the edge case where a user might be publishing from multiple environments: we would guide them to constrain it to one, and their publishing from the other(s) would then start failing.

I think that might be fine, but we might want to add a note to the modal advising the user to add additional publishers if they plan to use multiple environments.

warehouse/templates/manage/project/publishing.html
```diff
@@ -438,4 +438,25 @@ <h2 class="no-bottom-padding">{% trans %}Add a new publisher{% endtrans %}</h2>
...
+{% set constrained_environment_name = request.params.constrained_environment_name %}
+{% set confirm_name = "constrain_environment" %}
+{% set confirm_button_label = gettext("Constrain environment") %}
+{% set slug = confirm_name.lower().replace(' ', '-') + '-modal' %}
+{% set extra_fields %}
+  <input name="constrained_publisher_id" type="hidden" value="{{ request.params.constrained_publisher_id }}">
+  <input name="constrained_environment_name" type="hidden" value="{{ constrained_environment_name }}">
+{% endset %}
+{% set extra_description %}
+<p>{% trans %}Restrict Trusted Publisher environment to: {% endtrans %}<strong>{{ constrained_environment_name }}</strong></p>
```
di, line 453:
```suggestion
<p>{% trans %}Restrict Trusted Publisher environment to:{% endtrans %} <strong>{{ constrained_environment_name }}</strong></p>
```

I think this will fix the missing space here:

image

woodruffw:
> My only additional thought is that this doesn't really handle the edge case where a user might be publishing from multiple environments: we would guide them to constrain it to one, and their publishing from the other(s) would then start failing.

Yeah, good point! I'll add some language to the modal explaining this.

woodruffw:
Done:

Screenshot 2025-01-13 at 1 52 45 PM

di (approved):

Issue #2211, opened 2017-07-19
Project images


mahmoud:
Right now the project pages look pretty good, but are kind of same-y:

<img width="49%" src="https://user-images.githubusercontent.com/130193/28394080-70947586-6c9f-11e7-9234-c61d5b18ddf6.png"> <img width="49%" src="https://user-images.githubusercontent.com/130193/28394081-709584da-6c9f-11e7-82fc-d28d6842fc9d.png">

Theoretically these could be dressed up by ReST-savvy users. Realistically, I would love to see project images. I think this is especially important for the innumerable tile-based integrations out there, like this one on twitter:

<img width="70%" src="https://user-images.githubusercontent.com/130193/28394249-6b4a5a40-6ca0-11e7-9466-1e4f26bd2a40.png">

That gray square seems so drab. (Even if we don't do project images, it'd be a quick fix to add the logo into the card.)

Maybe a URL in an argument to `setup()` (or in a toml file)? Maybe PyPI can help out with hosting? Maybe a data URL with mimetype? I think there are a lot of options, but having something graphical will really help PyPI keep visual pace in a media-rich world. :)

dstufft:
I don't have a problem with this, though it will likely need @nlhkabu to take a look at how to integrate it into the design sanely. I guess one question would be whether it should be a PyPI specific enhancement that is managed *on* PyPI, or whether it should be part of the packages themselves.

If the former then we need to add it to the (yet undone) project management pages it it likely makes the most sense to simply have it be a form that people can upload images for. If the latter then it'll need to go through disutitls-sig to figure out how to add it to the packaging metadata.

brainwane:
Thanks for your suggestion, @mahmoud!

As I think you know, but for the benefit of other folks finding this thread: the folks working on Warehouse have gotten [funding to concentrate on improving and deploying Warehouse](https://pyfound.blogspot.com/2017/11/the-psf-awarded-moss-grant-pypi.html), and have kicked off work towards [our development roadmap](https://wiki.python.org/psf/WarehouseRoadmap) -- the most urgent task is to improve Warehouse to the point where we can redirect pypi.python.org to pypi.org so the site is more sustainable and reliable. Since project-level images/avatars aren't something that the legacy site has, I've moved it to a future milestone.

But I agree that it would be great to give projects a way to display their logo/screenshot-type images someplace in the page and in, e.g., the Twitter card for the PyPI project URL.

Thanks for the idea, and sorry again for the wait.

[...]

PR #17282, opened 2024-12-14
Removes large pypi image in Slack link unfurling

rmasters:
One of my pet peeves is that sharing a pypi link in Slack embeds a tall image - this displays at most two pypi links on my 13" screen until the expansion is deleted by the user.

CleanShot 2024-12-14 at 11 28 32@2x

This PR prevents inclusion of the `og:image` meta tag when the user-agent represents the Slack link unfurler. It keeps the existing images for Twitter and other social network embeds.

Slack | Twitter | Facebook
--- | --- | ---
Slack preview, without an image | Twitter preview, showing an image | Facebook preview, with an image

This PR also makes the following changes to the Twitter embed:
* Declares the card as a summary card, which was preventing some preview tooling from working. I believe this was previously falling back to a summary card by default.
* Adds [@pypi](https://x.com/pypi) as the site.

It's worth noting that if #2211 is implemented, this should be reverted, ideally with an image height constraint.

---

Despite it's simplicity, I'm not sure if adding to the project/release view context is the best approach for this. Would you prefer this to be a service? 

I'm considering some further expansion to this - adding fields like the github repo link, downloads count and latest release date to the metadata (as seen in [this example](https://api.slack.com/reference/messaging/link-unfurling#classic_unfurl), using the `twitter:label1`/`twitter:data1` meta properties).


warehouse/packaging/views.py
```diff
@@ -288,6 +289,7 @@ def release_detail(release, request):
         "license": license,
         # Additional function to format the attestations
         "PEP740AttestationViewer": PEP740AttestationViewer,
+        "show_share_image": should_show_share_image(request),
```
di, line 292: I think it probably makes more sense to pass the user agent here...
```suggestion
        "user_agent": request.user_agent,
```

warehouse/templates/base.html
```diff
@@ -113,9 +113,13 @@
     <meta property="og:url" content="{% if request.matched_route %}{{ request.current_route_url() }}{% else %}{{ request.url }}{% endif %}">
     <meta property="og:site_name" content="PyPI">
     <meta property="og:type" content="website">
+    {% if show_share_image %}
```
di, line 116: ...and have this become a [Jinja filter](https://github.com/pypi/warehouse/blob/9a270a2d01cba250795daf1e05133fbbafeeb7b0/warehouse/filters.py) instead
```suggestion
    {% if user_agent | show_share_image %}
```
That way, we can just write additional filters if we need to in the future, rather than adding to the return value for `release_detail`.

di (commented):
This looks great! I agree these large images are annoying in Slack.

Just have one refactoring suggestion but I think the overall approach is great.

warehouse/packaging/views.py
rmasters: Thanks - I used the request that's already in the template (lmk if that's not always present). I also added some handling for `Request.user_agent is None` in case the user-agent isn't supplied.

rmasters:
Thanks for the review @di! That's made it a bit cleaner.

```
# Slackbot
$ http localhost/project/mkdocs-ponylang/ 'User-Agent: Slackbot-LinkExpanding 1.0 (+https://api.slack.com/robots)' | grep og:image

# User-agent header not sent
$ http localhost/project/mkdocs-ponylang/ 'User-Agent:' | grep og:image
    <meta property="og:image" content="http://localhost/static/images/twitter.abaf4b19.jpg">
```

di:
Thinking about this a bit more: I don't think we can actually vary the response here based on the user agent, unfortunately.

These pages will almost always be served from our CDN cache, which doesn't currently vary the cache based on the user agent, so we would end up with a mix of pages with/without this image based on who hit the stale cache first and which request made it to the backend. While we could _technically_ vary on the user agent, I don't think we want to do this in practice because it would make our cache size grow to be extremely large.

I think we'll need to take a different approach here: either find a way to make the image more reasonable for Slack, or                                                                                                 a way to make Slack prefer a different image without having to vary the response, or something else.

woodruffw:
It's not perfect, but I think there are a few other OpenGraph tricks that could be used here:

* `og:image:width` and `og:image:height` could be used to constrain the size to something reasonable (64x64?), although I'm not sure if Slack respects these tags.
* Putting more metadata into the tags might cause Slack to reflow the unfurl into something more reasonable. In particular, the `twitter:label*` and `twitter:data*` tags can be used to embed additional metadata, e.g.:

```html
<meta name="twitter:label1" content="Latest version" />
<meta name="twitter:data1" content="$LATEST_VERSION" />
<meta name="twitter:label2" content="Uploaded on" />
<meta name="twitter:data2" content="$UPLOAD_DATE" />
```

(I learned about these tags from this post: <https://whitep4nth3r.com/blog/level-up-your-link-previews-in-slack/> -- they're technically not OGP tags but appear to be widely supported, including by Slack.)

di:
Closing due to inactivity.

PR #17284, opened 2024-12-14
Titlize language names

eggplants:
Aligned the case styles of the language names displayed on the language switch buttons.

|||
|-|-|
|[Before](https://pypi.org/)|localhost_ (2)|
|[After](http://localhost:80)|localhost_ (1)|

---

```py
langs = ["English", "español", "français", "日本語", "português (Brasil)", "українська", "Ελληνικά", "Deutsch", "中文 (简体)", "中文 (繁體)", "русский", "עברית ", "Esperanto"]
for lang in langs: print(lang, "=>", lang.title())
```

```
English => English
español => Español
français => Français
日本語 => 日本語
português (Brasil) => Português (Brasil)
українська => Українська
Ελληνικά => Ελληνικά
Deutsch => Deutsch
中文 (简体) => 中文 (简体)
中文 (繁體) => 中文 (繁體)
русский => Русский
עברית  => עברית
Esperanto => Esperanto
```

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -365,7 +365,7 @@
               {% endif %}
               name="locale_id" value="{{ identifier }}" type="submit"
             >
-              {{ locale.display_name }}
+              {{ locale.display_name.title() }}
             </button>
           </li>
           {% endfor %}


miketheman (commented):
Thanks for the contribution. I think this enforces an English-language construct onto how languages represent themselves.

The [`babel` library](https://babel.pocoo.org/) is responsible for holding the rules for each language, and offers methods to obtain the languages's name in a localized rule for a given locale, so I think this is currently operating correctly, and shouldn't be changed.

Here's an example for French:

```python
>>> from babel import Locale
>>> Locale.parse("fr").display_name
'français'
>>> Locale.parse("fr").get_display_name("en")
'French'
>>> Locale.parse("fr").get_display_name("fr")
'français'
>>> Locale.parse("fr").get_display_name("de")
'Französisch'
```

We can see that the default `display_name` is localized to the language of origin (in this case `fr`), instead of the currently-viewed language.
I believe this to be the correct behavior, so will close this request.

If you have supporting information as to why this behavior should be changed and applicable to all the languages, even when the locales have changed, supply that here, and we can reopen the request.

eggplants:
> If you have supporting information as to why this behavior should be changed and applicable to all the languages, even when the locales have changed, supply that here, and we can reopen the request.

I am not a native speaker of each language, so I don't know if it is necessary to write in different case, but I have confirmed from the language selectors of supposedly well-known websites in each language that it is OK to start with all uppercase letters.

## `español`

https://administracion.gob.es/pag_Home/index.html
Screenshot_20241217-013030_Chrome

## `français`

https://www.france.fr/fr/
Screenshot_20241217-012454_Chrome

## `português`

https://www.parlamento.pt/sites/en/

Screenshot_20241217-013525_Chrome

## `українська`

https://www.kmu.gov.ua/en

Screenshot_20241217-014136_Chrome

## `русский`

http://en.kremlin.ru/
Screenshot_20241217-014332_Chrome

eggplants:
@miketheman I don't know why the Babel library changes the initial capitalization depending on the locale, but I can see from the aforementioned use cases for each language site that it would look more natural if the language names were unified by starting with all uppercase letters.

di:
I agree that we should defer to whatever Babel does here. Looks like this is the upstream issue, which was closed as "wontfix": https://github.com/python-babel/babel/issues/544

eggplants:
Wouldn't it be acceptable if there was a motive to unify it as a “title case?”

Anyway, if you do not want to add extra processing to the page that is displayed each time it is accessed, or if it is easier to leave it to the library's recommended method, there is no need to make any changes.

Thank you for reviewing.

PR #17296, opened 2024-12-17
fix: retry observation reporting task

miketheman:
After seeing the occasional error raised from the upstream service, retry a few times (defaults to 3) with exponential backoff.

Refs: https://docs.celeryq.dev/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions

### DIFF
--- a/warehouse/observations/tasks.py
+++ b/warehouse/observations/tasks.py
@@ -17,6 +17,8 @@ import typing
 from base64 import b64encode
 from textwrap import dedent
 
+from requests.exceptions import RequestException
+
 from warehouse import db, tasks
 from warehouse.helpdesk.interfaces import IHelpDeskService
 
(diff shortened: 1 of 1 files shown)


warehouse/observations/tasks.py
```diff
@@ -51,7 +51,13 @@ def execute_observation_report(config: Configurator, session: SA_Session):
         config.task(report_observation_to_helpscout).delay(obj.id)
 
 
-@tasks.task(bind=True, ignore_result=True, acks_late=True)
+@tasks.task(
+    bind=True,
+    ignore_result=True,
+    acks_late=True,
+    autoretry_for=(Exception,),
```
ewdurbin, line 58: not strictly opposed to a catch all here, but is it worth being more restrictive?

ewdurbin (approved):

warehouse/observations/tasks.py
miketheman: Good catch - I'll narrow it to only the ones we've seen so far - from remote services via `requests`.

PR #17338, opened 2024-12-30
fix: use correct host in URL construction

miketheman:
From #17324 

This catches us often in tasks, defaulting to `localhost`.

### DIFF
--- a/warehouse/observations/tasks.py
+++ b/warehouse/observations/tasks.py
@@ -78,6 +78,8 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
     #  Maybe need a mapping of ObservationType and the name we want to use.
     target_name = model.related.name
 
+    warehouse_domain = request.registry.settings.get("warehouse.domain")
+
     # Add new Conversation to HelpScout for tracking purposes
     convo_text = dedent(
         f"""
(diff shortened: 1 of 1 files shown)


miketheman:
I did some code spelunking to see if this was relatively easy to make "standard", and here's the path:

- construct a request-like object when called via [`get_request`](https://github.com/pypi/warehouse/blob/85bb371709d0b7e28c1580c3f7041a0252fbbf7e/warehouse/tasks.py#L92-L104
) via `pyramid.scripting.prepare(registry=registry)`
- as we don't pass a `request` (how could we? it doesn't exist yet!), only a `registry` to [prepare()](https://github.com/Pylons/pyramid/blob/eb0e01a0fd7aff84c345b1c43bd4fa9d2fe04441/src/pyramid/scripting.py#L93)
- it in turn calls [`_make_request`](https://github.com/Pylons/pyramid/blob/eb0e01a0fd7aff84c345b1c43bd4fa9d2fe04441/src/pyramid/scripting.py#L130-L154) with a path of `/` and `registry` 
- inside [`webob.request.BaseRequest.blank`](https://docs.pylonsproject.org/projects/webob/en/stable/api/request.html#webob.request.BaseRequest.blank), where a call to `environ_from_url(path='/')` will return `http://localhost:80/` from [this logic](https://github.com/Pylons/webob/blob/39d5af3c797e7b867f152c2e8c979de42d029403/src/webob/request.py#L1521-L1523)
It doesn't look like there's an opportunity to inject a domain setting from the registry along that path.

I experimented with crafting a dummy request object by pulling the domain from the settings, but since we don't have a full URL handy, it doesn't work well.

Going to proceed with these changes for now.

Issue #17370, opened 2025-01-08
Unable to enable two factor authentication


sgoadhouse:
**Describe the bug**

I have not logged in for over a year. When I logged in today, I see a banner message saying that I must enable two-factor authentication. I click on the link and it takes me to the page to start the process but the two buttons for the two methods are grayed out. If I hover over the buttons, I get the circle with a line through it symbol which is typically for an action not allowed. I tried clicking one and I am taken to my profile settings page and not to any page that allows me to enable 2FA. How can I enable 2FA?

**Expected behavior**

After I click the link to enable two-factor authentication, I should be sent to a page with working buttons that allow me to enable 2FA.

**To Reproduce**

Click the Enable Two-Factor Authentication link after I log into my account.

**My Platform**

I have tried it on macOS in Safari and Chrome. I have tried it on my iPhone in Safari. I have also tried it in Windows 11 in Chrome. In all cases, I get the same result.

**Additional context**

Here is a screen shot of what I see showing the buttons grayed out.
![Screenshot of what I see in each of the browsers that I have tried it on](https://github.com/user-attachments/assets/078564c5-18ce-461f-b50a-f3ed23fc44a6)

di:
Hi, you must generate recovery codes before enabling 2FA, you should click the "Generate recovery codes" button.

sgoadhouse:
Doh! I started on test.pypi.com and got myself confused. It works now.

PR #17371, opened 2025-01-08
Remove potentially confusing callout when setting up 2FA

di:
This removes an unnecessary and duplicate warning callout that can be distracting from what we actually want the user to do when setting up 2FA.

| Before | After |
|--------|--------|
| Screenshot 2025-01-08 at 12 13 01 PM | Screenshot 2025-01-08 at 12 15 45 PM | 

The callout is only shown when the user is actually able to begin setting up 2FA:

Screenshot 2025-01-08 at 12 21 17 PM

Helps to alleviate confusion raised in #17370.

### DIFF
--- a/warehouse/templates/manage/manage_base.html
+++ b/warehouse/templates/manage/manage_base.html
@@ -177,7 +177,7 @@
     {% endif %}
 
     {% set disabled = not user.has_burned_recovery_codes or not user.has_primary_verified_email %}
-    {% if not user.has_two_factor %}
+    {% if not disabled and not user.has_two_factor %}
      <div class="callout-block">
         <p>
         {% trans %}


miketheman (commented):
I haven't tested this yet, but with this change, what happens if a user *has* 2fa enabled, but hasn't generated any recovery codes yet if they enrolled in 2FA before burning a recovery code was required?

di:
WIth recovery codes generated and one burned:
Screenshot 2025-01-08 at 4 28 48 PM

With recovery codes generated and none burned:
Screenshot 2025-01-08 at 4 30 37 PM

WIth no recovery codes:
Screenshot 2025-01-08 at 4 31 04 PM

miketheman (approved):
Excellent, thanks for verifying!

PR #17380, opened 2025-01-09
orgs: stripe data hygiene 

ewdurbin:
Couple fixes for our stripe integration:

- set metadata that allows us to differentiate stripe objects/events relevant to us from other tenants in our account
- ignore events we cannot/shouldnt act on
- allows accounts with lapsed/cancelled subscriptions to create a new subscription


warehouse/manage/views/organizations.py
```diff
@@ -573,7 +574,16 @@ def create_or_manage_subscription(self):
...
+        live_subscriptions = (
+            self.request.db.query(StripeSubscription)
+            .filter(StripeSubscription.organization == self.organization)
+            .filter(
+                StripeSubscription.status.not_in([StripeSubscriptionStatus.Canceled])
+            )
+            .all()
+        )
+
+        if not live_subscriptions:
```
di, line 586: Should we use `organization.active_subscription` here instead, or create a similar property?

ewdurbin: active is not quite "manageable" so I implemented "manageable_subscriptions" in 1e8b28432f343392ba6bba19b25e455533921e3a

di (approved):

PR #17381, opened 2025-01-09
fix: environment name should be pypi

### LINKED ISSUE #17374: doc: update screenshots for trusted publisher (and dev docs sync issue)
Hi friends. This issue is being opened as it came up in **[this pr](https://github.com/pyOpenSci/pyopensci.github.io/pull/538#discussion_r1889505496)** around screenshots i was using in a blog post!

**Describe the bug**
The [images here](https://docs.pypi.org/trusted-publishers/adding-a-publisher/#github-actions) use release as the environment name, but it sounds like PyPA  prefers pypi.

 Also, when I went to recreate it, the actual interface should look like this - there is now a project-name box at the top as well. 

Image
If you want to use these screenshots (or i can submit a pr) i can remove the 2025 from the second one. I had to do that because sampleProject already exists on pypi :)

Image

**********

**Expected behavior**
We want the screenshots to have PyPI as the environment name rather than release.

**To Reproduce**

**My Platform**
**Additional context**

More context & a suggestion if you are open to it. As a new contributor to warehouse 🚀 I'd be happy to submit a pr! But also, please know that your dev instructions are a bit off, and I'd love to update those, too (or at least help). 

* When I follow the new contributor path, I land on [building documentation](https://warehouse.pypa.io/development/getting-started.html#building-documentation); this page says the docs are built with Sphinx and use make to build. But when I dug deeper into the readme for the docs, it is mkdocs and markdown instead! And, you need to have a docker setup (which I learned when following the link to installation instructions). 

My suggestion, if you are open, is rather than having that information twice, have one source of truth and in the development guide, link to the [readme](https://github.com/pypi/warehouse/blob/main/docs/user/README.md), and installations instructions where things will most likely always be up to date. This might avoid things from becoming out of sync. 

Please let me know if you'd be open to a pr to update the screenshots AND if you'd like me to update the dev docs as well to ensure information is correct across sections (unless I'm confused!!). 

cc @webknjaz @woodruffw

lwasser:
Fixes #17374

This pr updates the screenshots for trusted publisher to ensure the environment name is listed as pypi rather than release. 

## Dev instructions were tricky to follow 

Friends, I wasn't able to get docker to run as expected. I first went down a rabbit hole that I realized was building the entire warehouse. I did this because the README suggested that I follow the detailed instructions, but i think that I should have only followed the 

* git/clone part
* docker compose setup part (i already have docker installed but not all users will)

I wonder if it's worth adding a note in the dev docs that says - if you want to only build the docs - go to (and add links to the readme for the user docs). I'm still wrapping my head around your setup, so this could also be my misunderstanding. 

Let me know, however, if this PR seems ok. It's just image updates so in theory i shouldn't need to rebuild the docs locally (but i do want to figure out how to).


lwasser:
ok yay for CI -- it seems like it's rendering ok 🚀 
https://docspypiorg--17381.org.readthedocs.build/trusted-publishers/creating-a-project-through-oidc/

docs/user/assets/trusted-publishing/github/pending-publisher-form-filled.png
di, line 1: There's a subtle difference with this one: the "Environment name" field is focused and so it has a blue outline.

Also, for consistency's sake, these screenshots were previously made by taking a "node screenshot" of the enclosed section:

image
 
Doing this is not strictly necessary but will keep the cropping around the element the same as before.

lwasser: Ahhh ok i'll go back and fix the images!! I didn't know about node screenshot as an option!

lwasser: wow that worked well . the only downside is that i couldn't make the fonts a bit bigger but i suppose that it ok! An update is coming soon. 🚀 thanks @di 

image

lwasser: Dustin - note i'm getting a slightly different graphic size because there's a note about configuring a pending publisher. i could edit it in photoshop OR perhaps i'm pulling from the wrong place. otherwise this is well aligned with the existing graphic taken as a node screenshot.

docs/user/assets/trusted-publishing/github/project-publishing-form.png
lwasser, line 1: These are also slightly different sizes but i did capture as a node screenshot.

docs/user/assets/trusted-publishing/github/pending-publisher-registered.png
lwasser, line 1: i saved this as 8 bit and it's a lot smaller. I could do the same for the others if that is of interest.

di (approved):
Looks great! Not worried about the file sizes.

Issue #17389, opened 2025-01-10
HTTPError: 500 Internal Server Error


AnthonyLim23:
**Describe the bug**

My github action for automatically uploading my project to PyPi used to work (Oct/Nov). However, it now produces the error ` HTTPError: 500 Internal Server Error` https://github.com/ISISMuon/MuonDataLib/actions/runs/12713026079/job/35440721845?pr=37

I updated the upload/download versions as part of a PR, but I cannot see why my workflow is failing. 

**Expected behavior**

I expect the workflow to publish the package on PyPi

**To Reproduce**

https://github.com/ISISMuon/MuonDataLib/blob/42afbc12acf7d953985ee5c288e0b2325201f39e/.github/workflows/merge.yml is the file in the PR https://github.com/ISISMuon/MuonDataLib/pull/37/files#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1
**My Platform**

**Additional context**

di:
Hi, thanks for the issue. Looks like this is ultimately due to a bug in setuptools (https://github.com/pypa/setuptools/issues/4797) but we should be producing a reasonable error message here. We're looking into it, in the meantime you can pin back your setuptools version [here](https://github.com/ISISMuon/MuonDataLib/blob/main/pyproject.toml#L2) to `setuptools<=75.7.0`.

PR #17390, opened 2025-01-10
fix: handle incoming invalid `dynamic` metadata

miketheman:
Not currently validated by `packaging` library.
Provides the user a better error message than a 500 response.

Fixes: #17389
Fixes WAREHOUSE-PRODUCTION-252

### DIFF
--- a/warehouse/forklift/metadata.py
+++ b/warehouse/forklift/metadata.py
@@ -30,6 +30,7 @@ from packaging.requirements import InvalidRequirement, Requirement
 from trove_classifiers import all_classifiers, deprecated_classifiers
 from webob.multidict import MultiDict
 
+from warehouse.packaging.models import DynamicFieldsEnum
 from warehouse.utils import http
 
 SUPPORTED_METADATA_VERSIONS = {"1.0", "1.1", "1.2", "2.1", "2.2", "2.3", "2.4"}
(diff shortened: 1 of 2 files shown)


warehouse/forklift/metadata.py
```diff
@@ -235,6 +236,19 @@ def _validate_metadata(metadata: Metadata, *, backfill: bool = False):
                         )
                     )
 
+    # Validate that any `dynamic` fields passed are in the allowed list
+    # TODO: This probably should be lifted up to packaging.metadata
+    for field in {"dynamic"}:
+        if (value := getattr(metadata, field)) is not None:
+            for key in value:
+                if key not in map(str.lower, DynamicFieldsEnum.enums):
```
ewdurbin, line 244: will `key` already have been normalized to lowercase for comparison?

miketheman: I had to check, but yes - packaging does that for us
https://github.com/pypa/packaging/blob/9375ec2eff48257967c97d331b9a76019e95bdb0/src/packaging/metadata.py#L314-L316

miketheman: I added a test case change to prove the case-insensitive behavior, just in case.

ewdurbin: Thank you!

ewdurbin (approved):

Issue #17392, opened 2025-01-11
Trusted Publisher error message is confusing when project exists but user doesn't have ownership


unrealsolver:
**Describe the bug**

At 'Publishing', PyPI tells that "This project already exists: use the project's publishing settings [here](https://pypi.org/manage/project/pyxlmapper/settings/publishing/?project_name=pyxlmapper&owner=unrealsolver&repository=pyxlmapper&workflow_filename=python-pypi.yml&environment=pypi&provider=github) to create a Trusted Publisher for it." However, it can not find any projects by following that link.

**Expected behavior**

Publishing a new (non-existent) project should not raise an error that project already exists

**To Reproduce**

Actual steps to reproduce are unknown, however I did following:
1. Considering you are logged in and account is ready to publish a package 
2. Go to Publish
3. PyPI Project name: `pyxlmapper`
4. Owner: `unrealsolver`
5. Repository name: `pyxlmapper`
6. Workflow name: `python-pypi.yml`
7. Environment name: `pypi`
8. Press `Add` button
9. Observe an error (see Describe the bug section)

**My Platform**

**Additional context**

TestPyPI is working very well: https://test.pypi.org/project/pyxlmapper/

unrealsolver:
Aslo, attaching the log from the publishing job just in case:
Job: https://github.com/unrealsolver/pyxlmapper/actions/runs/12727660052/job/35477433098
```
Error: Trusted publishing exchange failure: 
Token request failed: the server refused the request for the following reasons:

* `invalid-publisher`: valid token, but no corresponding publisher (All lookup strategies exhausted)

This generally indicates a trusted publisher configuration error, but could
also indicate an internal error on GitHub or PyPI's part.


The claims rendered below are **for debugging purposes only**. You should **not**
use them to configure a trusted publisher unless they already match your expectations.

If a claim is not present in the claim set, then it is rendered as `MISSING`.

* `sub`: `repo:unrealsolver/pyxlmapper:environment:pypi`
* `repository`: `unrealsolver/pyxlmapper`
* `repository_owner`: `unrealsolver`
* `repository_owner_id`: `1101771`
* `workflow_ref`: `unrealsolver/pyxlmapper/.github/workflows/python-pypi.yml@refs/tags/v0.0.2.4`
* `job_workflow_ref`: `unrealsolver/pyxlmapper/.github/workflows/python-pypi.yml@refs/tags/v0.0.2.4`
* `ref`: `refs/tags/v0.0.2.4`
[...]

PR #17405, opened 2025-01-13
Improve error messages when a pending Trusted Publisher's project name already exists

facutuesca:
## The problem
There are some cases where we validate the name of a new project, where the error messages could be better:
- When creating a pending Trusted Publisher with a project name that already exists, the error message includes a link to the project's settings. This link is only useful if the user is the owner of the existing project, otherwise it's confusing (see https://github.com/pypi/warehouse/issues/17392).

## The implementation
This PR removes the `enum` used for describing the invalid project name errors, replacing it with simple classes that allow including some metadata in each error object. This metadata allows us to tell if the current user is the owner of the existing project, which we then use to display better error messages.

```python
# Before
class ProjectNameUnavailableReason(enum.Enum):
    Invalid = "invalid"
    Stdlib = "stdlib"
    AlreadyExists = "already-exists"
    Prohibited = "prohibited"
    TooSimilar = "too-similar"

# After
class ProjectNameUnavailableInvalid:
    pass
class ProjectNameUnavailableStdlib:
    pass
class ProjectNameUnavailableProhibited:
    pass

class ProjectNameUnavailableExisting:
    def __init__(self, existing_project: Project):
        self.existing_project: Project = existing_project
class ProjectNameUnavailableSimilar:
    def __init__(self, similar_project: Project):
        self.similar_project: Project = similar_project

ProjectNameUnavailableError = (
    ProjectNameUnavailableInvalid
    | ProjectNameUnavailableStdlib
    | ProjectNameUnavailableExisting
    | ProjectNameUnavailableProhibited
    | ProjectNameUnavailableSimilar
)
```

#### Project already exists and user is owner
Screenshot 2025-01-13 at 22 39 42

#### Project already exists and user is not owner
Screenshot 2025-01-13 at 22 42 15

---- 

This fixes https://github.com/pypi/warehouse/issues/17392

cc @di @woodruffw

### DIFF
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -19,7 +19,7 @@ from requests import ConnectionError, HTTPError, Timeout
 from webob.multidict import MultiDict
 
 from warehouse.oidc.forms import activestate
-from warehouse.packaging.models import ProjectNameUnavailableExisting
+from warehouse.packaging.interfaces import ProjectNameUnavailableExisting
 
 fake_username = "some-username"
 fake_org_name = "some-org"
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -19,7 +19,7 @@ from requests import ConnectionError, HTTPError, Timeout
 from webob.multidict import MultiDict
 
 from warehouse.oidc.forms import activestate
-from warehouse.packaging.interfaces import ProjectNameUnavailableReason
+from warehouse.packaging.models import ProjectNameUnavailableExisting
 
 fake_username = "some-username"
 fake_org_name = "some-org"
@@ -50,6 +50,7 @@ class TestPendingActiveStatePublisherForm:
             MultiDict(data),
             route_url=route_url,
             check_project_name=check_project_name,
+            user=pretend.stub(),
         )
 
         # Test built-in validations
@@ -61,27 +62,43 @@ class TestPendingActiveStatePublisherForm:
         assert form._route_url == route_url
         assert form.validate()
 
-    def test_validate_project_name_already_in_use(self, pyramid_config):
+    @pytest.mark.parametrize("is_project_owner", [True, False])
+    def test_validate_project_name_already_in_use(
+        self, pyramid_config, is_project_owner
+    ):
         route_url = pretend.call_recorder(lambda *args, **kwargs: "")
+        user = pretend.stub()
+        owners = []
+        if is_project_owner:
+            owners.append(user)
 
         def check_project_name(name):
-            return ProjectNameUnavailableReason.AlreadyExists
+            return ProjectNameUnavailableExisting(
+                existing_project=pretend.stub(owners=owners)
+            )
 
         form = activestate.PendingActiveStatePublisherForm(
             route_url=route_url,
             check_project_name=check_project_name,
+            user=user,
         )
 
         field = pretend.stub(data="some-project")
         with pytest.raises(wtforms.validators.ValidationError):
             form.validate_project_name(field)
-        assert route_url.calls == [
-            pretend.call(
-                "manage.project.settings.publishing",
-                project_name="some-project",
-                _query={"provider": {"activestate"}},
-            )
-        ]
+
+        if is_project_owner:
+            # The project settings URL is only shown in the error message if
+            # the user is the owner of the project
+            assert route_url.calls == [
+                pretend.call(
+                    "manage.project.settings.publishing",
+                    project_name="some-project",
+                    _query={"provider": {"activestate"}},
+                )
+            ]
+        else:
+            assert route_url.calls == []
 
 
 class TestActiveStatePublisherForm:
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -62,15 +62,10 @@ class TestPendingActiveStatePublisherForm:
         assert form._route_url == route_url
         assert form.validate()
 
-    @pytest.mark.parametrize("is_project_owner", [True, False])
-    def test_validate_project_name_already_in_use(
-        self, pyramid_config, is_project_owner
-    ):
+    def test_validate_project_name_already_in_use_owner(self, pyramid_config):
         route_url = pretend.call_recorder(lambda *args, **kwargs: "")
         user = pretend.stub()
-        owners = []
-        if is_project_owner:
-            owners.append(user)
+        owners = [user]
 
         def check_project_name(name):
             return ProjectNameUnavailableExisting(
@@ -87,18 +82,37 @@ class TestPendingActiveStatePublisherForm:
         with pytest.raises(wtforms.validators.ValidationError):
             form.validate_project_name(field)
 
-        if is_project_owner:
-            # The project settings URL is only shown in the error message if
-            # the user is the owner of the project
-            assert route_url.calls == [
-                pretend.call(
-                    "manage.project.settings.publishing",
-                    project_name="some-project",
-                    _query={"provider": {"activestate"}},
-                )
-            ]
-        else:
-            assert route_url.calls == []
+        # The project settings URL is only shown in the error message if
+        # the user is the owner of the project
+        assert route_url.calls == [
+            pretend.call(
+                "manage.project.settings.publishing",
+                project_name="some-project",
+                _query={"provider": {"activestate"}},
+            )
+        ]
+
+    def test_validate_project_name_already_in_use_not_owner(self, pyramid_config):
+        route_url = pretend.call_recorder(lambda *args, **kwargs: "")
+        user = pretend.stub()
+        owners = []
+
+        def check_project_name(name):
+            return ProjectNameUnavailableExisting(
+                existing_project=pretend.stub(owners=owners)
+            )
+
+        form = activestate.PendingActiveStatePublisherForm(
+            route_url=route_url,
+            check_project_name=check_project_name,
+            user=user,
+        )
+
+        field = pretend.stub(data="some-project")
+        with pytest.raises(wtforms.validators.ValidationError):
+            form.validate_project_name(field)
+
+        assert route_url.calls == []
 
 
 class TestActiveStatePublisherForm:
--- a/tests/unit/oidc/forms/test_activestate.py
+++ b/tests/unit/oidc/forms/test_activestate.py
@@ -19,7 +19,7 @@ from requests import ConnectionError, HTTPError, Timeout
 from webob.multidict import MultiDict
 
 from warehouse.oidc.forms import activestate
-from warehouse.packaging.interfaces import ProjectNameUnavailableExisting
+from warehouse.packaging.interfaces import ProjectNameUnavailableExistingError
 
 fake_username = "some-username"
 fake_org_name = "some-org"
@@ -68,7 +68,7 @@ class TestPendingActiveStatePublisherForm:
         owners = [user]
 
         def check_project_name(name):
-            return ProjectNameUnavailableExisting(
+            return ProjectNameUnavailableExistingError(
                 existing_project=pretend.stub(owners=owners)
             )
 
@@ -98,7 +98,7 @@ class TestPendingActiveStatePublisherForm:
         owners = []
 
         def check_project_name(name):
-            return ProjectNameUnavailableExisting(
+            return ProjectNameUnavailableExistingError(
                 existing_project=pretend.stub(owners=owners)
             )
 
--- a/warehouse/oidc/forms/_core.py
+++ b/warehouse/oidc/forms/_core.py
@@ -13,7 +13,13 @@ import markupsafe
 import wtforms
 
 from warehouse.i18n import localize as _
-from warehouse.packaging.interfaces import ProjectNameUnavailableReason
+from warehouse.packaging.models import (
+    ProjectNameUnavailableExisting,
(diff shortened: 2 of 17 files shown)


warehouse/packaging/services.py
```diff
@@ -443,32 +448,36 @@ def _hit_ratelimits(self, request, creator):
...
+            return ProjectNameUnavailableStdlib()
 
-        if self.db.query(
-            exists().where(Project.normalized_name == func.normalize_pep426_name(name))
-        ).scalar():
-            return ProjectNameUnavailableReason.AlreadyExists
+        if (
+            existing_project := self.db.query(Project)
+            .where(Project.normalized_name == func.normalize_pep426_name(name))
+            .scalar()
```
ewdurbin, line 462: Hm, what would happen in the event that two conflicting names exist? We didn't always enforce this and did not retro-actively "clean up" so it could happen?

I think in the pre-existing implementation papers over that because the "exists" clause will end up a boolean through the "scalar" clause, where as after I think you'll get a list of Project objects.

samuelcolvin:
Thanks so much for fixing so quickly ❤️ .

warehouse/packaging/services.py
facutuesca: doesn't `scalar` return the first result?

ewdurbin: I recall it doing so if one and only one exists, otherwise a multiple results exception.

Would you mind adding a test to validate the behavior if two preexisting conflicting names exist?

facutuesca: uh yeah, I was looking at the wrong docs. Good catch! Fixed by replacing `scalar()` with `first()`, and added a test for it.

warehouse/oidc/forms/_core.py
```diff
@@ -31,38 +37,52 @@ def validate_project_name(self, field):
...
                 )
-            case ProjectNameUnavailableReason.TooSimilar:
+            case ProjectNameUnavailableSimilar(similar_project=similar_project):
                 raise wtforms.validators.ValidationError(
-                    _("This project name is too similar to an existing project")
+                    _(
+                        "This project name is too similar to an existing project "
+                        "named '${name}'",
+                        mapping={"name": similar_project.name},
+                    )
```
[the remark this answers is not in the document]
facutuesca: For 1, we could return a list of project names if that was the case.

For 2, I think that only makes sense if the conflicting project is abandoned, in which case yes, the user could try to contact the owner to get them to delete it.
But for active projects, it's useful for the user to know which project is conflicting: if it's an active or well-known package they probably won't want to take over it.

Knowing exactly which project is too similar in name allows the user to make the decision if they should go forward or not with trying to get that name, whereas otherwise they are forced to go through PyPI support.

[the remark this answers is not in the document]
di: This case is when a user explicitly wants a _different_ name than the conflicting name though, so the ideal situation would be that they could go forward with publishing under the original name, and not have to take over or delete a slightly different name in order to use the original name.

Adding multiple conflicting names to the error message would just exacerbate this I think: what would we expect the user to do with that information? The better outcome would be that regardless of what the conflicting names(s) are, they are able to request the original name.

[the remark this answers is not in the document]
woodruffw: > For 1, we could return a list of project names if that was the case.

I think this could work, although in principle the list of conflicting names is unbounded (or could become unbounded with a future restriction on similarity). 

I think @di's point with contact is that maintainers will find it annoying/frustrating to have deletion requests directed at them first, especially since there's typically no (strong) public signal that a project is actually abandoned. There's also a possibility that less-experienced users think that the project *itself* is imposing the restriction, which will lead to reports from those users that the projects can't actually do anything about (e.g. "`pip` is preventing me from uploading `p1p` to PyPI").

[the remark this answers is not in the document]
facutuesca: > so the ideal situation would be that they could go forward with publishing under the original name, and not have to take over or delete

Ah, I wasn't aware that this could be done (allowing a new project that has a too-similar name as an existing one). I see you mentioned at the end of your original comment, I just didn't register it 🤦 

I see your point. I still think knowing the name might be useful in some cases (for example, if I'm trying to create `p1p`, and I see that it conflicts with `pip`, I will just find a different name rather than ask the PyPI admins about it). 

But I'm guessing this is rare, and not worth modifying the error message for.
WDYT about the rest of the PR? Should I remove the changes related to the too-similar error, and leave the rest as-is?. Or should we go with a different implementation, since we're only adding extra information to the "already exists" error?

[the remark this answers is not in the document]
woodruffw: > WDYT about the rest of the PR? Should I remove the changes related to the too-similar error, and leave the rest as-is?. Or should we go with a different implementation, since we're only adding extra information to the "already exists" error?

IMO we can leave it in, and just drop it from the public facing message for now 🙂.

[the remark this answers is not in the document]
di: > I see your point. I still think knowing the name might be useful in some cases (for example, if I'm trying to create p1p, and I see that it conflicts with pip, I will just find a different name rather than ask the PyPI admins about it).

Generally we want the user to be able to use their first preference rather than hunt around for a less ideal name, though!

> IMO we can leave it in, and just drop it from the public facing message for now 🙂.

Seems fine to me.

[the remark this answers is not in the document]
facutuesca: Done, removed the name from the error message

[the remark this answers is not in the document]
miketheman: > E.g. we've reported copyright legal violations and had security concerns and had no response after months.

@samuelcolvin For security-concern requests going without response, can you elaborate where these were reported? I monitor the security inbox closely and investigate violations pretty quickly.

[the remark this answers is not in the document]
samuelcolvin: Hi @miketheman, sorry for the slow response - we had a company offsite which was all consuming.

The case I was thinking of was https://github.com/pypi/support/issues/5187 where we had (I think) a legitimate legal and security concern, and it took months to get a response. My point with regard to this issue was just that historically "just ask admins to take X down" has not been a quick solution.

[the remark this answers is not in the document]
miketheman: Hey @samuelcolvin ! No worries, hope the offsite went well! 
Thanks for sharing the case in question. I hear you on the requests - [PyPI got a support specialist in July](https://pyfound.blogspot.com/2024/07/announcing-our-new-pypi-support.html) who's been cranking through account recovery requests, followed by PEP 541 requests, so there's definitely an uptick in the pace of resolution.
In this particular case, following the instructions in PEP 541 Intellectual Property policy is probably the best move.

warehouse/packaging/models.py
```diff
@@ -1130,3 +1130,34 @@ class AlternateRepository(db.Model):
...
+class ProjectNameUnavailableProhibited:
+    pass
+
+
+class ProjectNameUnavailableSimilar:
+    def __init__(self, similar_project: Project):
+        self.similar_project: Project = similar_project
+
+
+ProjectNameUnavailableError = (
```
miketheman, line 1157: thought: The `*Error` name here made me think we're starting to treat these like exceptions, but they aren't.

Would using explicit exceptions, and adopting a try/except block in the caller be better here?

warehouse/packaging/services.py
```diff
@@ -443,32 +448,36 @@ def _hit_ratelimits(self, request, creator):
...
                 ProhibitedProjectName.name == func.normalize_pep426_name(name)
             )
         ).scalar():
-            return ProjectNameUnavailableReason.Prohibited
+            return ProjectNameUnavailableProhibited()
 
-        if self.db.query(
-            exists().where(
+        if (
+            similar_project := self.db.query(Project)
```
miketheman, line 474: Now that the project name is no longer returned, are these changes necessary for the assignment? The previous call of `exists()...` is less expensive than creating a Project object as a result, especially if we end up throwing it away.

Alternately, if we **do only need the `name`**, this could be rewritten as selecting only the name, something like so?

```python
if (similar_project_name := self.db.scalars(
    select(Project.name)
    .where(func.ultranormalize_name(Project.name) == func.ultranormalize_name(name))
    .first()
):
...
```
as then `similar_project_name` would be a string, not an object, we we don't need the entire thing.

There's also something subtly wrong with selecting the first result and returning the name, since there may be other collisions (unlikely, but not impossible) - have you considered that in this change?

warehouse/packaging/models.py
```diff
@@ -1130,3 +1130,34 @@ class AlternateRepository(db.Model):
     name: Mapped[str]
     url: Mapped[str]
     description: Mapped[str]
+
+
+class ProjectNameUnavailableInvalid:
```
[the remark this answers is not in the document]
facutuesca: no good reason, just a leftover from a previous implementation. Moved back to `interfaces.py`

tests/unit/oidc/forms/test_activestate.py
```diff
@@ -61,27 +62,43 @@ def check_project_name(name):
...
+            # the user is the owner of the project
+            assert route_url.calls == [
+                pretend.call(
+                    "manage.project.settings.publishing",
+                    project_name="some-project",
+                    _query={"provider": {"activestate"}},
+                )
+            ]
+        else:
+            assert route_url.calls == []
```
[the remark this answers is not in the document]
facutuesca: makes sense, fixed!

warehouse/packaging/models.py
facutuesca: Yeah, thinking about it, using exceptions here seems like the pythonic way of doing it. Fixed

warehouse/packaging/services.py
facutuesca: > especially if we end up throwing it away.

I think the idea is that the name will be used once we figure out a way of displaying it to the user in a non-confusing way.

> Alternately, if we do only need the name, this could be rewritten as selecting only the name, something like so?

Yeah, this makes sense. The "existing project" error needed to store the project since we also want the information about the project's owners, but this "similar project" error doesn't need that. I changed it to only store the project name, and modified the query like you suggested


> There's also something subtly wrong with selecting the first result and returning the name, since there may be other collisions (unlikely, but not impossible) - have you considered that in this change?

Yeah, my thinking was that for the purposes of this error, the name of a single "too similar" conflicting project would be enough information for the user. But if that's not the case, we could change this to store a list of package names instead.

warehouse/packaging/interfaces.py
```diff
@@ -96,3 +89,41 @@ def create_project(name, creator, request, *, creator_is_owner=True):
...
         """
+
+
+class ProjectNameUnavailableError(Exception):
+    """Base exception for project name unavailability errors."""
+
+    pass
+
+
+class ProjectNameUnavailableInvalid(ProjectNameUnavailableError):
```
miketheman, line 100: I don't know how strict we are with PEP 8, as none of our linters yelled about this, but I had learned that [exception classes should end in `*Error`](https://peps.python.org/pep-0008/#exception-names).
We do that in other places like [these exceptions](https://github.com/pypi/warehouse/blob/0e7e85f11c84c52d037915b2880cd442ca7006ed/warehouse/captcha/hcaptcha.py#L26-L43).

I don't think it's a huge deal, but would be curious what others thought about that.

warehouse/packaging/interfaces.py
```diff
@@ -14,6 +14,7 @@
 
 from zope.interface import Interface
 
+from warehouse.packaging.models import Project
```
miketheman, line 17: suggestion: If the only reason to import `Project` is for type hints, let's update the imports to the equivalent of:

```python
from __future__ import annotations

import typing

if typing.TYPE_CHECKING:
    from warehouse.packaging.models import Project
```
(run `isort` after adding!)

That leaves a clear signal that this is not a run-time import, and can help avoid import cycles.

If it's needed for some other runtime reason, then leave as is.

warehouse/packaging/services.py
```diff
@@ -443,32 +447,34 @@ def _hit_ratelimits(self, request, creator):
...
+            raise ProjectNameUnavailableStdlib()
 
-        if self.db.query(
-            exists().where(Project.normalized_name == func.normalize_pep426_name(name))
-        ).scalar():
-            return ProjectNameUnavailableReason.AlreadyExists
+        if (
+            existing_project := self.db.query(Project)
+            .where(Project.normalized_name == func.normalize_pep426_name(name))
+            .first()
```
miketheman, line 461: suggestion: While we're here, use similar `db.scalars(select..` pattern that is used later for `similar_project_name`, so we have a consistent query interface. We should be moving to the `select()` style pattern everywhere eventually.

miketheman (changes requested):
Coming along nicely, I think we're almost there! A few last bits to consider, and then I think we're ready to go. Let me know when it's ready for re-review by clicking the 🔄 button on my username.

warehouse/packaging/interfaces.py
facutuesca: Fixed

warehouse/packaging/interfaces.py
facutuesca: Fixed

warehouse/packaging/services.py
facutuesca: Fixed

miketheman (approved):
Well done! :shipit:

ewdurbin:
https://python-software-foundation.sentry.io/share/issue/27a8e5055a0c4c54a2fcceb2c3479241/

PR #17409, opened 2025-01-14
Update DynamicFieldsEnum to include all deprecated fields

di:
Alternative to https://github.com/pypi/warehouse/pull/17390 given https://github.com/pypa/setuptools/issues/4797#issuecomment-2589514950.

### DIFF
--- a/tests/unit/forklift/test_metadata.py
+++ b/tests/unit/forklift/test_metadata.py
@@ -283,7 +283,7 @@ class TestValidation:
 
     def test_invalid_dynamic(self):
         data = MultiDict(metadata_version="2.2", name="spam", version="2.0")
-        data.add("dynamic", "Requires")
+        data.add("dynamic", "Invalid")
         with pytest.raises(ExceptionGroup) as excinfo:
             metadata.parse(None, form_data=data)
         _assert_invalid_metadata(excinfo.value, "dynamic")
--- a/tests/unit/forklift/test_metadata.py
+++ b/tests/unit/forklift/test_metadata.py
@@ -14,6 +14,7 @@ import packaging.metadata
 import pytest
 
 from packaging.version import Version
+from sqlalchemy.dialects.postgresql import ENUM
 from webob.multidict import MultiDict
 
 from warehouse.forklift import metadata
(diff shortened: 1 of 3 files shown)


warehouse/forklift/metadata.py
```diff
@@ -236,19 +235,6 @@ def _validate_metadata(metadata: Metadata, *, backfill: bool = False):
...
-        if (value := getattr(metadata, field)) is not None:
-            for key in value:
-                if key not in map(str.lower, DynamicFieldsEnum.enums):
-                    errors.append(
-                        InvalidMetadata(
-                            _RAW_TO_EMAIL_MAPPING.get(field, field),
-                            f"Dynamic field {key!r} is not a valid dynamic field.",
-                        )
-                    )
-
```
miketheman, line 251: Will any other part of metadata validation catch invalid inputs to the `dynamic` field?

ewdurbin: Agree that this check should remain.

ewdurbin (changes requested):
I suggest removing the revert commit and just updating the enum. This will retain validation that Dynamic contains values that we have in the enum, as well as the typo fixes from the previous PR.

warehouse/forklift/metadata.py
di: I think `packaging` should already be handling this? 

```
>>> Metadata.from_raw({"metadata_version": "2.4", "version": "0.0.0", "name": "lol", "dynamic": ["invalid"]})
  + Exception Group Traceback (most recent call last):
  |   File "<stdin>", line 1, in <module>
  |   File "/home/di/.pyenv/versions/3.12.7/lib/python3.12/site-packages/packaging/metadata.py", line 752, in from_raw
  |     raise ExceptionGroup("invalid metadata", exceptions)
  | ExceptionGroup: invalid metadata (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "/home/di/.pyenv/versions/3.12.7/lib/python3.12/site-packages/packaging/metadata.py", line 747, in from_raw
    |     getattr(ins, key)
    |   File "/home/di/.pyenv/versions/3.12.7/lib/python3.12/site-packages/packaging/metadata.py", line 514, in __get__
    |     value = converter(value)
    |             ^^^^^^^^^^^^^^^^
    |   File "/home/di/.pyenv/versions/3.12.7/lib/python3.12/site-packages/packaging/metadata.py", line 607, in _process_dynamic
    |     raise self._invalid_metadata(
    | packaging.metadata.InvalidMetadata: 'invalid' is not a valid dynamic field
    +------------------------------------
 ```

ewdurbin: Yes, but we still need to restrict based on what we have in our Enum. If packaging for some reason adds a new field to the allowed list that we do not have in our Dynamic enum, the error will be opaque again.

di: OK, left the check in place and updated the test accordingly instead.

ewdurbin (approved):

PR #17412, opened 2025-01-15
feat: automatically quarantine project via task

miketheman:
Every time a new observation is added, a background task will evaluate whether it should perform an automated quarantine or not.

If the criteria are satisfied, apply the quarantine logic and notify admins.


warehouse/observations/tasks.py
```diff
@@ -149,3 +154,127 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
...
+
+
+@tasks.task(
+    bind=True,
+    ignore_result=True,
+    acks_late=True,
+    autoretry_for=(RequestException,),
+    retry_backoff=True,
+)
+def auto_quarantine_project(task, request: Request, model_id: UUID) -> None:
```
ewdurbin, line 166: nit: I wonder if this could have a better name. When reading the call in `execute_observation_report` I got confused why it wasn't conditional. I wish I had an obvious answer that wasn't just `maybe_auto_quarantine_project`

warehouse/observations/tasks.py
```diff
@@ -149,3 +154,127 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
...
+    """
+    Automatically quarantine a Project for Admin review.
```
ewdurbin, line 168:
```suggestion
    Conditionally quarantine a Project for Admin review.
```

warehouse/observations/tasks.py
```diff
@@ -149,3 +154,127 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
...
+
+    Conditions must be met:
+
+    - ObservationKind is IsMalware
+    - Observed Project is not already quarantined
+    - Observed Project has at least 2 Observations, at least 1 by `User.is_observer`
+    """
+    # Fetch the Observation from the database, load the related Project
+    model = request.db.get(Observation, model_id)
```
ewdurbin, line 177: would `observation` and `observation_id` maybe be a bit clearer? especially when referenced later in this function.

ewdurbin (approved):

warehouse/observations/tasks.py
di: Agreed. `evaluate_project_quarantine`?

warehouse/utils/project.py
```diff
@@ -102,20 +104,27 @@ def quarantine_project(project: Project, request, flash=True) -> None:
     """
     Quarantine a project. Reversible action.
     """
+    # TODO: This should probably be extracted to somewhere more general for tasks,
+    #  but it got confusing where to add it in the context of this PR.
+    #  Since JournalEntry has FK to `User`, it needs to be a real object.
+    actor = request.user or request.db.scalar(
+        select(User).where(User.username == "admin")
+    )
```
di, line 112: We do have `get_admin_user` [here](https://github.com/pypi/warehouse/blob/c477e43c7b7910acab6f5d39cdd70589e15e85d0/warehouse/accounts/services.py#L138-L141), although you'd need to grab a `IUserService` to use it.

warehouse/observations/tasks.py
```diff
@@ -149,3 +154,127 @@ def report_observation_to_helpscout(task, request: Request, model_id: UUID) -> N
...
+        logger.info("Project is already quarantined. No change needed.")
+        return
+    # Check Observers
+    observer_users = {obs.observer.parent for obs in project.observations}
+    if len(observer_users) < 2:
+        logger.info("Project has fewer than 2 observers. Not quarantining.")
+        return
+    if not any(observer.is_observer for observer in observer_users):
+        logger.info("Project has no `User.is_observer` Observers. Not quarantining.")
+        return
```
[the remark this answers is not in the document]
miketheman: We accept observations from *any* authenticated user, the `is_observer` flag is currently part of the API-driven beta program for trusted folks.

warehouse/observations/tasks.py
miketheman: Good points, thanks - I'll also rename execute_observation_Report since that's no longer what it does.
How about `evaluate_project_for_quarantine()`?

warehouse/utils/project.py
miketheman: Thanks, I used that instead

miketheman:
Example of slack notification with the blocks in this changest
Screenshot 2025-01-15 at 12 19 57

warehouse/observations/tasks.py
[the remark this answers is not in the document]
di: Got it, I think [this comment](https://github.com/pypi/warehouse/blob/7c2bf391305f47ff96f5f1b858ff321b7a209bda/warehouse/accounts/models.py#L100) confused me.

di (approved):

PR #17424, opened 2025-01-16
Fix JS localization: `msgstr` is an array

di:
Fix for #15612, per https://github.com/smhg/gettext-parser?tab=readme-ov-file#translations


miketheman (approved):
Thanks!

Issue #16008, opened 2024-05-24
BigQuery `bigquery-public-data.pypi.distribution_metadata` missing data


mensfeld:
Running this query:

```
SELECT 
  name,
  version,
  summary,
  description
FROM 
  `bigquery-public-data.pypi.distribution_metadata`
WHERE 
  name = 'virtualenv'
ORDER BY 
  version;
```

misses several new versions available here: https://pypi.org/project/virtualenv/#history released in April and May. It's similar for some other packages.

**Describe the bug**

All versions info should be available in BigQuery.

**Expected behavior**

I would expect them (except eventual consistency ofc) to be available in BQ.

**To Reproduce**

Run in BigQuery:

```
SELECT 
  name,
  version,
  summary,
  description
FROM 
  `bigquery-public-data.pypi.distribution_metadata`
WHERE 
  name = 'virtualenv'
ORDER BY 
  version;
```

and see versions are missing.

ewdurbin:
The task that ensures consistency was disabled due to poor performance in... 2021 🙃 

https://github.com/pypi/warehouse/pull/10256

But was never subsequently re-enabled that I can tell, as the contributor never returned to address the issue.

For triage, I have manually run this task, can you confirm if you're seeing consistency?

mensfeld:
@ewdurbin was all the data synced? That is, should all the historical gaps be filled now? 

When I query `virtualenv` I'm still missing `20.25.2+` versions (anything newer). 

Is there any other endpoint to get the recent releases data?

mensfeld:
@ewdurbin I'm still not seeing the newer releases of `virtualenv` in the BigQuery dataset :(

ewdurbin:
Hmmm, unclear what the issue is. @di are you familiar with why the sync wouldn't capture past releases?

di:
That's not the job that inserts new metadata, that job just syncs missing metadata if insertion fails for some reason.

Insertion of new metadata happens on upload: https://github.com/pypi/warehouse/blob/main/warehouse/forklift/legacy.py#L1222-L1223

The timeline here is suspiciously close to when we did some migrations on these schemas, my guess is that the `update_bigquery_release_files‎` is failing and we're unaware.

ewdurbin:
So `sync_bigquery_release_files` is not the bulk equivalent of `update_bigquery_release_files‎`?

di:
[...]

PR #17429, opened 2025-01-16
Mechanics for backfilling missing BigQuery metadata

di:
This PR adds a new temporary model `MissingDatasetFile` (which will be populated out-of-band), and updates the `sync_bigquery_release_files` task to process any unprocessed rows in this table, reading the metadata from the file/release/project and populating the BigQuery dataset.

Towards https://github.com/pypi/warehouse/issues/16008.


tests/unit/packaging/test_init.py
```diff
@@ -169,10 +170,10 @@ def key_factory(keystring, iterate_on=None, if_attr_exists=None):
     ]
 
     if with_bq_sync:
-        # assert (
-        #    pretend.call(crontab(minute=0), sync_bigquery_release_files)
-        #    in config.add_periodic_task.calls
-        # )
+        assert (
+            pretend.call(crontab(minute="*"), sync_bigquery_release_files)
```
ewdurbin, line 174: I'm unsure about running this every minute, primarily since if an individual batch takes longer than 1 minute, we will re-run that batch (since processed will still be False if the task hasn't completed and thus committed).

Thoughts on making processed a nullable, and setting processed to False when a file is added to a batch, then True once processing is complete?

This would allow us to initiate batches by looking for processed = None, and monitor for failures of batches if processed = False on completion of the backfill.

warehouse/packaging/tasks.py
```diff
@@ -357,90 +361,69 @@ def sync_bigquery_release_files(request):
...
+    for missing_file in missing_files:
+        release_file = missing_file.file
+
+        for table_name in table_names:
+            table_schema = bq.get_table(table_name).schema
+
+            # Using the schema to populate the data allows us to automatically
+            # set the values to their respective fields rather than assigning
+            # values individually
+            def populate_data_using_schema(file):
```
ewdurbin, line 379: Just a nit, and I'm sure it's annoying, but I am struggling to feel certain that the way we are populating the object here and the way we are populating it during upload are indeed identical.

In my dream world, the [construction of dist_metadata from forklift/legacy/upload](https://github.com/pypi/warehouse/blob/main/warehouse/forklift/legacy.py#L1456) was a function that took a `File` object and returned the dict to pass to the `update_bigquery_release_files` task.

Maybe a bit overkill but otherwise there's just a lot to consider to ensure that we're populating the table in the same way as the mainline task.

di: Unfortunately (as you've discovered) there is data we send to BigQuery which only comes off the parsed metadata and isn't stored in the database (e.g., the metadata version), so there is going to be some slight differences here.

warehouse/packaging/tasks.py
```diff
@@ -357,90 +361,74 @@ def sync_bigquery_release_files(request):
...
+    missing_files = (
+        request.db.query(MissingDatasetFile)
+        .filter(MissingDatasetFile.processed.is_(None))
+        .limit(10)
+    )
+    for missing_file in missing_files:
+        missing_file.processed = False
+
+    # Commit these changes so another task doesn't pick up these files
+    request.db.commit()
```
di, line 373: I think this is necessary to prevent other tasks from picking up these rows, but I'm not sure we can do this here?

ewdurbin: Hmmm, yeah trying to do this in a task results in:

`AssertionError: Transaction must be committed using the transaction manager`

ewdurbin: You can:

```python
    # Commit these changes so another task doesn't pick up these files
    request.tm.commit()
    request.tm.begin()
```

tests/conftest.py
```diff
@@ -734,7 +735,6 @@ def tm():
     # Create a new transaction manager for dependant test cases
     tm = transaction.TransactionManager(explicit=True)
     tm.begin()
-    tm.doom()
```
di, line 737: Flagging this: I had to remove this otherwise the task would refuse to commit the transaction mid-task. This doesn't seem to affect any other tests and I _think_ the database will be reset anyways, so there's no harm here. This was added in #16031.

ewdurbin (approved):

PR #17442, opened 2025-01-17
run our docker ci on gha runners directly

ewdurbin:


ewdurbin:
Overall, this has proven pretty flaky... seemingly unrelated to anything we're doing in particular :/

Screenshot 2025-01-17 at 4 02 01 PM

Screenshot 2025-01-17 at 2 24 45 PM (1)

Screenshot 2025-01-17 at 2 34 17 PM

ewdurbin:
Screenshot 2025-01-18 at 10 26 07 AM

miketheman:
zizmor errors can probably be resolved easily with https://gha-update.readthedocs.io/

miketheman:
#18103 should unblock this

di (approved):

.github/workflows/ci.yml
```diff
@@ -14,28 +14,57 @@
...
           persist-credentials: false
-      - name: Set up Depot CLI
-        uses: depot/setup-action@b0b1ea4f69e92ebf5dea3f8713a1b0c37b2126a5 # v1.6.0
-      - name: Build image
-        id: build
-        uses: depot/build-push-action@636daae76684e38c301daa0c5eca1c095b24e780 # v1.14.0
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+      - name: Cache
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
```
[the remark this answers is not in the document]
miketheman: @woodruffw What's the right answer here?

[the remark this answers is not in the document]
woodruffw: If I understand correctly this is an intentional cache usage, so _probably_ a `zizmor: ignore` comment (or setting in the config file, whichever you prefer) 🙂 

GitHub doesn't do a good job of presenting the full SARIF here, but this is why it's flagging:

```
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
  --> .github/workflows/ci.yml:57:9
   |
30 |         uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cache enabled by default here
31 |         id: cache
...
56 |       - name: Build and push
57 |         uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1 # v6.16.0
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ runtime artifacts usually published here
   |
   = note: audit confidence → Low
```

i.e. it sees cache usage and then a "publish sink" that the cache might influence, so `zizmor` is warning that a poisoned cache could influence that Docker image publish step. But if this is a case where the cache is wholly trusted (e.g. because the only ways to invoke the workflow are trusted), then it can be safely ignored.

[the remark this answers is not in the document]
miketheman: Thanks! That makes sense. The workflow is able to be triggered by anyone sending a PR - but since the final artifact is a container image with the run_id, and it's not used for anything else but a second CI run, I think an ignore is the right moe.

.github/workflows/ci.yml
```diff
@@ -14,28 +14,57 @@ concurrency:
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1 # v6.16.0
         with:
-          save: true
+          context: .
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
di, line 61: I think we need something like:
```suggestion
          cache-to: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository && '' || 'type=gha,mode=max' }}
```
Otherwise action runs on PRs from untrusted forks can overwrite this cache as well.

(cc @woodruffw, should zizmor flag this?)

.github/workflows/ci.yml
```diff
@@ -14,28 +14,57 @@ concurrency:
...
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3 # zizmor: ignore[cache-poisoning]
+        id: cache
         with:
-          save: true
+          path: |
+            var-cache-apt
+            var-lib-apt
+            root-cache-pip
+            root-npm
+          key: cache-${{ hashFiles('Dockerfile') }}
```
di, line 38:
```suggestion
          key: cache-${{ hashFiles('Dockerfile') }}
          lookup-only: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository }}
```
https://github.com/actions/cache/blob/5a3ec84eff668545956fd18022155c47e93e2684/README.md?plain=1#L78

.github/workflows/ci.yml
woodruffw: Huh yeah, it should -- I'm actually confused as to why it didn't, since I have a coordinate for the `push: true` part of this:

https://github.com/woodruffw/zizmor/blob/9790cec2727f04fee3cbaef1725f910c114e343e/src/audit/cache_poisoning.rs#L183-L186

woodruffw: Oh never mind, I see why it didn't: it considers `docker/build-push-action` a "publish sink" but not a "cache source" at the moment, since I'm only detecting `push: true` and not the `cache-to: ...` field.

This is a shortcoming in the current audit, I'll file an issue 🙂

woodruffw: https://github.com/woodruffw/zizmor/issues/774

di: Awesome, thanks! Just to confirm, this suggestion resolves the issue, correct?

woodruffw: I _think_ so, although I'm murky on whether doing the same for `cache-from` is also needed: that's a sink rather than a source so I _think_ it's okay, but I might be missing something.

di: I think we specifically want `cache-from` to always be permitted here, so PRs can take advantage of our central cache to speed up builds (but can't write to it).

woodruffw: Makes sense!

PR #17512, opened 2025-01-28
Use zizmor from PyPI

DarkaMaul:
This PR switches the installation of zizmor from `cargo install` to `pip install`.

/cc @woodruffw

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -27,10 +27,8 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version-file: '.python-version'
-      - name: Get zizmor
-        run: python -m pip install zizmor
       - name: Run zizmor
-        run: zizmor --format sarif . > results.sarif
+        run: pipx run zizmor --format sarif . > results.sarif
       - name: Upload SARIF file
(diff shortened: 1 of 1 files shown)


miketheman (approved):

.github/workflows/zizmor.yml
```diff
@@ -23,12 +23,12 @@ jobs:
...
         with:
           persist-credentials: false
-      - name: Setup Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-      - name: Get zizmor
-        run: cargo install zizmor
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: '.python-version'
```
di, line 29: This should be the default I think?

https://github.com/actions/setup-python/blob/42375524e23c412d93fb67b49958b491fce71c38/src/setup-python.ts#L36-L51

DarkaMaul: We can update both this one and the other [workflow](
https://github.com/pypi/warehouse/blob/e740cecc0931441d38426f163881d801d9ecefaf/.github/workflows/codeql-analysis.yml#L49)

Issue #2, opened 2013-04-04
Ensure that the database handles the ON DELETE for Emails


dstufft:
Implement ON DELETE at the database level for the ForeignKey from `warehouse.accounts.models.Email` to `warehouse.accounts.models.User`.

PR #17513, opened 2025-01-28
Update workflows

DarkaMaul:
This PR contains two workflow improvements:

1. Removes redundant default setting (per @di's feedback in https://github.com/pypi/warehouse/pull/17512#discussion_r1932412820)
2. Optimizes CI performance by running `zizmor` checks only when `.github/workflows/**` files change

Note: While change #2 reduces CI load, new `zizmor` checks won't run until either:
- Workflow files are modified
- Weekly scheduled run occurs (Monday)

I am happy to split these changes into separate PRs if preferred.

/cc @woodruffw

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -45,8 +45,6 @@ jobs:
     - name: Set up Python
       if: matrix.language == 'python'
       uses: actions/setup-python@v5
-      with:
-        python-version-file: '.python-version'
 
     - name: Install dependencies
       # Needed for pycurl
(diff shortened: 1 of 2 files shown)


di:
> Note: While change https://github.com/pypi/warehouse/issues/2 reduces CI load, new zizmor checks won't run until either

FWIW I don't think CI load has been an issue here. I assume this also mean that we won't run new versions of zizmor until we modify workflows / the cron runs as well, correct? I think I'm OK with keeping the workflow as-is to ensure we're up to date.

woodruffw (approved):
Thanks @DarkaMaul!

PR #17519, opened 2025-01-29
lint: narrow types to ensure correct usage

miketheman:
TL,DR: use [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html) on objects of ambiguous nature.

These commits push us closer to being able to run `mypy` with `--check-untyped-defs`, we're not fully there yet, but let's get these out the door.

See each commit message for specifics on why the particular action was taken.

With these changes in place, we're down to:

> Found 115 errors in 30 files (checked 394 source files)

### DIFF
--- a/warehouse/migrations/env.py
+++ b/warehouse/migrations/env.py
@@ -31,6 +31,8 @@ def run_migrations_offline():
     script output.
     """
     options = context.config.get_section(context.config.config_ini_section)
+    if options is None:
+        raise ValueError("No options found in config")
     url = options.pop("url")
     context.configure(url=url, compare_server_default=True)
 
@@ -46,6 +48,8 @@ def run_migrations_online():
     and associate a connection with the context.
     """
     options = context.config.get_section(context.config.config_ini_section)
+    if options is None:
+        raise ValueError("No options found in config")
     url = options.pop("url")
     connectable = create_engine(url, poolclass=pool.NullPool)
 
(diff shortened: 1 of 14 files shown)


warehouse/migrations/env.py
```diff
@@ -31,6 +31,7 @@ def run_migrations_offline():
     script output.
     """
     options = context.config.get_section(context.config.config_ini_section)
+    assert options is not None  # type narrowing
```
di, line 34: We should probably better handle the case where `options` could be `None` here (raise an exception?) instead of assuming it will always be non-`None`.

warehouse/accounts/models.py
```diff
@@ -237,6 +237,7 @@ def has_primary_verified_email(self):
     @property
     def recent_events(self):
         session = orm.object_session(self)
+        assert session is not None
```
di, line 240: This (and the many other `assert session is not None`s makes me think we probably need a helper function that is responsible for narrowing the type and explicitly handling the case where the session might be `None` rather than assuming it always is and throwing an `AssertionError` if it's not. AFAICT, this probably should just raise our own exception with more details.

miketheman: I'll extract this to a helper and push as a separate PR, so that can be evaluated on its merits alone.

miketheman: https://github.com/pypi/warehouse/pull/17582

di (approved):

PR #17526, opened 2025-01-30
Add an admin UI to burn leaked recovery codes

### LINKED ISSUE #17301: Recovery code reporting & revocation
Currently, we don't support disclosure for our 2FA Recovery codes, nor do we support manual revocation by admins.

- [ ] Admins can manually revoke leaked recovery codes
- [ ] ~Recovery codes have a recognizable pattern~
- [ ] ~Recovery codes can be revoked automatically via our secret-reporting API~
- [ ] ~Our existing secret-scanning partners support automated reporting of recovery codes~

**Additional context**
Currently, recovery codes have the following format:

```python
r"^[0-9a-fA-F]{16}$"
```

This is the result of calling [`secrets.token_hex(8)`](https://docs.python.org/3/library/secrets.html#secrets.token_hex). This is probably too generic for most secret scanning partners as it will likely result in a large number of false positives.

Ideally recovery codes would have a shared prefix like `pypi-` which our API tokens use, but we should be careful not to make these codes easily confusable with API tokens.

di:
Adds an admin UI to burn provided recovery codes (if they are valid). Fixes #17301.

image

### DIFF
--- a/warehouse/admin/views/users.py
+++ b/warehouse/admin/views/users.py
@@ -27,10 +27,10 @@ from sqlalchemy import or_, select
 from sqlalchemy.orm import joinedload
 
 from warehouse.accounts.interfaces import (
-    IEmailBreachedService,
-    IUserService,
     BurnedRecoveryCode,
+    IEmailBreachedService,
     InvalidRecoveryCode,
+    IUserService,
 )
 from warehouse.accounts.models import (
     DisableReason,
--- a/warehouse/admin/views/users.py
+++ b/warehouse/admin/views/users.py
@@ -26,7 +26,12 @@ from pyramid.view import view_config
 from sqlalchemy import or_, select
 from sqlalchemy.orm import joinedload
 
-from warehouse.accounts.interfaces import IEmailBreachedService, IUserService
+from warehouse.accounts.interfaces import (
+    IEmailBreachedService,
+    IUserService,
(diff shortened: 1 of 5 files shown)


ewdurbin (approved):
🔥

ewdurbin:
Only question I have is if we might want to have this be a blind endpoint. If we see something that looks like a recovery code but can't associate it to a user we can't burn it.

ewdurbin:
On second thought, I'm not certain that is possible 🙃

di:
> On second thought, I'm not certain that is possible 🙃

It's not possible!

warehouse/admin/views/users.py
```diff
@@ -593,3 +598,31 @@ def user_recover_account_complete(user: User, request):
...
+    if not codes:
+        request.session.flash("No recovery codes provided", queue="error")
+
+    else:
+        user_service = request.find_service(IUserService, context=None)
+        n_burned = 0
+
+        for code in codes:
+            try:
+                if user_service.check_recovery_code(user.id, code):
```
miketheman, line 622: Will this add more events to a user's security history, one per code burned?

di: No, the events are generated outside of the `check_recovery_code` function when the user actually burns a code.

PR #17556, opened 2025-02-04
Support PEP 738

### LINKED ISSUE #17549: Support PEP 738 (Adding Android as a supported platform)
PEP 738 is accepted, https://peps.python.org/pep-0738/, PyPI needs to be updated to permit the additional wheel tags: https://peps.python.org/pep-0738/#packaging 

The list of supported platform tags is currently here: https://github.com/pypi/warehouse/blob/6792d3862aa740b2249690f72770a2ff9dd809f3/warehouse/forklift/legacy.py#L98-L187

di:
Fixes #17549.

Blocked on https://github.com/pypa/packaging.python.org/issues/1801 since the PEP is under-specified.


di:
Since python/peps#4250 was declined due to the PEP being marked as final, this is blocked until we can get a firmer description of the tags elsewhere.

PR #17559, opened 2025-02-05
Add support for iOS and Android platform tags.

### LINKED ISSUE #17549: Support PEP 738 (Adding Android as a supported platform)

### LINKED ISSUE #17557: Support PEP 730 (Adding iOS as a supported platform)
PEP 730 is accepted, https://peps.python.org/pep-0730/, PyPI needs to be updated to permit the additional wheel tags: https://peps.python.org/pep-0730/#packaging

The list of supported platform tags is currently here: https://github.com/pypi/warehouse/blob/6792d3862aa740b2249690f72770a2ff9dd809f3/warehouse/forklift/legacy.py#L98-L187

freakboy3742:
Adds support for PEP 730 iOS tags, and PEP 738 Android tags.

Fixes #17549
Fixes #17557
Closes #17556


freakboy3742:
FYI @mhsmith

di:
I think the PEP 738 bits here are blocked as noted in https://github.com/pypi/warehouse/pull/17556.

But if you would like to ship the PEP 730 part separately, the specification in the PEP seems sufficient and I'd be ok merging that today.

Although this seems to be missing the minimum iOS version?

freakboy3742:
> I think the PEP 738 bits here are blocked as noted in #17556.

I've submitted pypa/packaging.python.org#1804 to address that requirement. 
 
> But if you would like to ship the PEP 730 part separately, the specification in the PEP seems sufficient and I'd be ok merging that today.
> 
> Although this seems to be missing the minimum iOS version?

As noted in [this comment about Android](https://github.com/python/peps/pull/4250#issuecomment-2634539915), while Python sets a default minimum iOS version (13.0 at present), it's possible to use older versions. iOS 12.0 is known to work with some compile-time warnings. I haven't tested older versions, but I wouldn't be surprised to learn older iOS versions (back to 8.0, when support for dynamic linking was added) could be made to work. 

On that basis, I figured it wasn't worth adding an explicit version check, as there's nothing that *fundamentally* prevents older versions from working if you've gone to the trouble of compiling Python with an earlier minimum iOS version. However, I'm happy to add a minimum version check (12.0 would be a reasonable bound based on the wording of the PEP) if you feel it is required.

di:
> I've submitted https://github.com/pypa/packaging.python.org/pull/1804 to address that requirement.

Thanks! I added a review there.

> On that basis, I figured it wasn't worth adding an explicit version check, as there's nothing that fundamentally prevents older versions from working if you've gone to the trouble of compiling Python with an earlier minimum iOS version. However, I'm happy to add a minimum version check (12.0 would be a reasonable bound based on the wording of the PEP) if you feel it is required.

Got it, I think I misread https://peps.python.org/pep-0730/#packaging which is just providing "12.0" as an example and not an actual minimum. I'm OK with not including a minimum but I think we need to specify what we consider valid values for these fields just a little more (as I noted in my review of the update to the guide).

warehouse/utils/wheel.py
```diff
@@ -40,11 +40,25 @@
         re.compile(r"^macosx_(\d+)_(\d+)_(.*?)$"),
         lambda m: f"macOS {m.group(1)}.{m.group(2)}+ {_normalize_arch(m.group(3))}",
     ),
+    (
+        re.compile(r"^android_(\d+)_(.*?)$"),
+        lambda m: f"Android SDK {m.group(1)}+ {_normalize_arch(m.group(2))}",
```
mhsmith, line 45: As mentioned [here](https://github.com/pypa/packaging.python.org/pull/1804#discussion_r1943019546), I'd prefer to replace "SDK" with "API level".

mhsmith:
> I'm OK with not including a minimum

I agree with not including a minimum, for the reasons given [here](https://github.com/pypa/packaging.python.org/pull/1804#discussion_r1943126942).

mhsmith (approved):

freakboy3742:
@di pypa/packaging.python.org#1804 has now been merged; was that the only remaining blocker on merging this?

di (approved):

PR #17568, opened 2025-02-06
resolve new errors introduced in #17405

ewdurbin:
changing from return value to exception was not caught due to stubs.

it's unclear to me how the changes in #17405 were expected to work, or if they were intended to be surfaced in the UI.

cc @facutuesca


miketheman (approved):

ewdurbin:
I see now, the case/match was updated to try/except in the path that legacy upload takes, but not in the UI form, and was covered up by the stubs in tests.

facutuesca:
Ouch, sorry for the mistake. Thanks for the fix @ewdurbin !

PR #17571, opened 2025-02-06
profile: add archived projects section

### LINKED ISSUE #17550: Label archived projects as such in public user profiles
Project listings on public user profile pages like https://pypi.org/user/jwodder/ currently do not indicate which projects are archived.  It would be helpful if an "Archived" label (similar to that shown on https://pypi.org/manage/projects/) were added.

woodruffw:
This splits the project list in a user's public profile into two groups: "active" and "archived". Both are rendered as before, but archived projects are additionally given their own header (linked to a new help section) and each project is given a little "archived" badge that's identical to the badge elsewhere.

Example render: 

Screenshot 2025-02-06 at 4 12 09 PM

CC @miketheman for feedback 🙂 

Closes #17550.


warehouse/templates/accounts/profile.html
```diff
@@ -66,12 +66,12 @@ <h1 class="author-profile__name">{{ user.name }}</h1>
...
           {% trans count=projects|length %}
-            {{ count }} project
+            {{ count }} active project
           {% pluralize %}
-            {{ count }} projects
+            {{ count }} active projects
           {% endtrans %}
         {% else %}
-          {% trans %}No projects{% endtrans %}
+          {% trans %}No active projects{% endtrans %}
```
di, line 74: Not sure I would go as far as to say these are "active" (maybe the maintainer hasn't discovered the archive feature yet?) but I'm on the fence.

woodruffw: Yeah, good point! I could just drop this and tweak the count so that it includes the archived projects, WDYT?

di (approved):

warehouse/templates/accounts/profile.html
```diff
@@ -75,7 +76,7 @@ <h2>
         {% endif %}
       </h2>
       <div class="package-list">
-        {% if projects %}
+        {% if project_count %}
```
miketheman, line 79: I dug through this, and realized what was bothering me.

After this change, we confirm that there's actual projects, but the way the HTML is structured we end up with two `package-list` entries, and nothing differentiates them. Could we/should we combine the logic so that there's one instance of a `package-list` or at least provide some context to the div class that these are `active` vs `archived`? Or is that overthinking it for classes that don't have any CSS specific to them right now? (Ignore for a second that the CSS for `package-list` was removed a long time ago.)

This is not a blocker, but I wanted to note it in case someone thinks it's important for later.

warehouse/accounts/views.py
```diff
@@ -206,17 +208,28 @@ def profile(user, request):
...
 
-    return {"user": user, "projects": projects}
+        if row.lifecycle_status == LifecycleStatus.Archived:
+            archived_projects.append(project)
+        else:
+            live_projects.append(project)
+
+    return {
+        "user": user,
+        "projects": live_projects,
```
miketheman, line 230: question: considering the only variable use in the template of `projects` is now to combine to project_count and to iterate over it, any reason to not have consistency and rename it `live_projects` to reduce cognitive load?

warehouse/accounts/views.py
```diff
@@ -206,17 +208,28 @@ def profile(user, request):
...
+            "lifecycle_status": row.lifecycle_status,
             "created": row.created,
             "summary": row.summary,
         }
-        for row in request.db.execute(stmt)
-    ]
 
-    return {"user": user, "projects": projects}
+        if row.lifecycle_status == LifecycleStatus.Archived:
+            archived_projects.append(project)
```
miketheman, line 224: This conditional is not expressed by tests just yet. Considering this change is also mucking about in the template-land, maybe a functional test is in order.

Likely location: https://github.com/pypi/warehouse/blob/e8195e40986d63f9b72926a67eb6acfd43b36932/tests/functional/test_user_profile.py

I think you've done some client-side HTML assertions in other functional tests before, let me know if you need any pointers!

miketheman (changes requested):

warehouse/accounts/views.py
woodruffw: Yeah, good call. I didn't rename it because I didn't want to churn a template variable but it's definitely confusing as-is!

warehouse/accounts/views.py
woodruffw: Done! I've added both a functional test and a unit-level one: the functional test confirms that the view doesn't explode when archived projects are present, and the unit test confirms that the `live_projects`/`archived_projects` states are what we expect.

miketheman: Thanks! I appreciate the functional test not-exploding part, but since we have some actual logic encoded in the HTML template like counts and whether or not lists appear, would it make sense to assert those as well? Doesn't need to be exhaustive, but something a bit more useful than "the template doesn't blow up the call".

miketheman (commented):
Last two comments (hopefully)!

warehouse/accounts/views.py
woodruffw: > would it make sense to assert those as well? Doesn't need to be exhaustive, but something a bit more useful than "the template doesn't blow up the call".

I think so, although I don't know how to test those 😅 -- are you thinking we pattern match on the rendered HTML, or something else?

miketheman: Yay, HTML testing time! 
```python
resp = webtest.get(...)

resp.html  # a BeautifulSoup object!!

assert resp.html.h2.text == 


```

woodruffw: Oh right, I forgot there was a BS object in the response. Nice, I'll use that.

woodruffw: Done -- I've added an assert on the `h2`. I could probably add similar ones on the `package-list` divs, although that goes back to your point about providing more context. So I'm slightly inclined to leave those for a follow-up, unless you feel strongly that they should be included here 🙂

miketheman: This is better than not, and good enough to confirm that the math and pluralization works.

miketheman (approved):

miketheman (approved):

miketheman:
Keep in mind: profile pages are cached for a day, so changes may not be immediately visible if the pages are not cached.

PR #17572, opened 2025-02-07
Correct handling of fat3 macOS wheels and eggs.

### LINKED ISSUE #17566: Accepted macOS tags don't match canonical list.
**Describe the bug**

The list of architecture tags accepted for macOS includes:
https://github.com/pypi/warehouse/blob/88bc28b20504a301725a945b9d4b3059dab35276/warehouse/forklift/legacy.py#L135-L147

However, the official list of tags generated by CPython are:
https://github.com/python/cpython/blob/cdcacec79f7a216c3c988baa4dc31ce4e76c97ac/Lib/_osx_support.py#L546-L562

This means warehouse accepts `fat32` (which isn't defined), but doesn't accept `fat3` (which is).

**Expected behavior**

`fat3` wheels should be accepted for upload.

`fat32` wheels should be rejected.

**To Reproduce**

Attempt to upload a `fat3` macOS wheel.

**My Platform**

macOS.

**Additional context**

The code fix is trivial - it's almost certainly a typo misreading `fat3` as `fat32`. However, it's not clear if there's any additional fixes required on the backend - if there are any `fat32` wheels that *have* been successfully uploaded.

freakboy3742:
Fixes #17566.

Warehouse currently accepts wheels of type `fat32`; this tag doesn't exist, and appears to be a typo of `fat3`. There are no examples of `fat32` wheels in the Warehouse database (as Python tooling has never produced them, and pip would never request them). 

It also labels eggs as "Source" (see https://pypi.org/project/pillow/2.5.0/#files for an example).

This PR:
* Corrects the `fat3` wheel tag
* Adds architecture descriptors for the macOS combined architectures other than `universal2`
* Ensures eggs are described as such
* Adds tests for all of the above.

These changes are almost entirely anachronistic... but they are part of the spec, and the discrepancy (discovered during writing pypa/packaging.python.org#1804) was just itching me :-)


di (approved):
Thanks!

PR #17582, opened 2025-02-10
refactor: extract and narrow `orm.object_session()`

miketheman:
Since `object_session(...)` can return `None`, we need to help mypy have confidence that the session is indeed there.

One way to narrow the type is to assert that it's not `None`, and therefore should not alert about `union-attr` problems.

Includes a couple of renames, and variable extraction where relevant.


miketheman:
Extracted from #17519 - focused only on the changes to support `object_session` narrowing.

di (approved):

Issue #6631, opened 2019-09-13
Define translation workflow / policy


nlhkabu:
We need to answer the following questions:

1. What happens with translations when we update the (English) source? 
1. When do we consider a language to be 'supported' and link to it from the PyPI footer?
1. What happens if a language is no longer maintained?

My 2c:

## Updating English source

First, we need to reach out to the community to register maintainers for each language.
When we update the UI copy:

1. The change is merged as normal
2. The UI falls back to the original English
3. We notify translators via [Weblate](https://weblate.org/en-gb/) that a new translation is required

Note: In some cases (e.g. fixing typos, formatting) translators may be able to use the [translation memory](https://docs.weblate.org/en/latest/admin/memory.html) to reduce the maintenance burden

## When do we consider a language 'supported'?

If 90% (???) of strings are translated, we should link to it from the PyPI footer.

## What happens if a language is no longer maintained?

We remove it from the PyPI footer if it falls below 90% (???) coverage

nlhkabu:
Note, once this workflow is decided, we should document it in the Warehouse docs.  I have added a basic translations page in https://github.com/pypa/warehouse/pull/6681.

brainwane:
I'm happy with all these policy decisions @nlhkabu - @ewdurbin you have any objections?

brainwane:
@ewdurbin will "spend some time late this week trying to figure out the best way for us to handle building translation files from source" per conversation in IRC just now. That might include some updates to #6749 to document the process.

webknjaz:
So I've tried out Weblate and there's one thing I see missing which is the ability to create variants of phrases depending on the number/gender for languages that use declension for this.
I suppose that the way to go is to use poEdit app locally instead and then send PRs. But how does that integrate back into the Weblate UI?

*(Ref: https://twitter.com/webKnjaZ/status/1181495636946014208)*

fitojb:
But Weblate is just gettext; in its side it _does_ allow for plural forms according to the language.

webknjaz:
Alright, I've seen just one string with multiple versions depending on the number there. Which means that the string source should be fixed to properly mark such places.

nlhkabu:
@webknjaz - sounds like this justifies it's own ticket. 

Would you be able to open a new issue to explain the problem? Please use the 'translation' issue type.

Thanks

webknjaz:
https://github.com/pypa/warehouse/issues/6798

NathanBnm:
[...]

PR #17602, opened 2025-02-13
Fix `s/frow/from` typo @ `publishing.html` template

webknjaz:


miketheman (approved):

miketheman:
> It'd be a good idea to have automation that would do this on its own. A-la pre-commit.ci. Would you be open to exploring this? I'm pretty confident I know how to implement it.

This is an interesting idea, and I'd very much be open to learning more about this! One of the challenges is how to manage merge conflicts on this kind of thing, so I'm curious how you might approach that as well. https://github.com/pypi/warehouse/issues/6631 has some background on translation workflow ideas

webknjaz:
Forgot to mention that I filed #17733 with some ideas.

Issue #17590, opened 2025-02-11
Squash Alembic migrations to improve test execution time


DarkaMaul:
`warehouse` test suite currently executes 246 Alembic migrations sequentially during database setup, adding significant overhead to test execution times. I investigated squashing these migrations into a single migration that represents the current database state.

Using `alembic`'s automatic migration feature provides a good starting point, but requires manual review to address limitations:

1. SQL functions are not automatically exported
2. Triggers are not included in the export

| Configuration | Full Tests | API Tests |
|--------------|------------|------------|
| With Squash | 19.69s | 3.12s |
| Without Squash | 21.75s | 3.66s |

*Note: Currently experiencing 128 failures out of 4402 tests in the full test suite, while API tests are passing successfully.*

Proposed Implementation:
1. Preserve historical migrations by moving them to `migrations/old/`
2. Replace current migrations with a single consolidated migration in `migrations/versions/`

Primary Consideration:

The main tradeoff is losing the ability to easily roll back to intermediate database states. We could mitigate this by choosing an earlier cutoff date for the consolidated migration if needed.

I'm opening this issue to discuss if we should proceed forward before trying to fix every edge case in the migration.

/cc @miketheman @woodruffw

miketheman:
I am generally 👍 on things that go faster, so yay!

When it comes to alembic and [initial DB state](https://warehouse.pypa.io/application.html#:~:text=Since%20Warehouse%20was,in%20the%20future), that's where it's a little tricky.

For the functions and triggers not represented - is that due to an alembic limitation, or some other reason? Do you have a sense of which those are? I'm curious if they predate `warehouse` or exist for some other reason.

Considering that all migrations live in git history, do we need to preserve them at all? I routinely search git history with `git log -p -S <somestring>` to find the commits with that text, and then can use other techniques to browse from there.

> The main tradeoff is losing the ability to easily roll back to intermediate database states. We could mitigate this by choosing an earlier cutoff date for the consolidated migration if needed.

Possibly, however it's rare that we roll back the db more than a single migration, usually due to some deployment issue or other error.

I think the Most Important Part of this kind of effort is that running the new "base" migration should be a no-op in production, and since the dev db doesn't fully represent the production database, that's a risk we'll have to consider and mitigate..

DarkaMaul:
I finished running the migration today and fixed the tests - everything is green now, but there is a slight coverage difference I still need to investigate.

There are 11 triggers in the database:

- projects_update_sitemap_bucket
[...]

PR #17607, opened 2025-02-14
Squash alembic migrations

DarkaMaul:
Discussion is in #17590 

The idea is to remove all previous alembic migrations and consolidate them into a single file to improve testing time.

Marked as _draft_ until we know if it's a good idea or not.


DarkaMaul:
The PR is now ready for review, but the first question is: do the benefits outweigh the drawbacks?

Benefits:

Timing | CI | Local |
---- | --- | --- |
With PR | 52s ([0])| 19s |
Without |  1m ([1]) | 22s |
Improvement |  13% | 13% |


[0]: https://github.com/pypi/warehouse/actions/runs/14196372469/job/39772662544?pr=17607
[1]: https://github.com/pypi/warehouse/actions/runs/14176665275/job/39713175354

The changes yield a consistent 13% performance improvement in both CI and local test execution times, saving several seconds per test run.

Drawbacks:

- Increased code complexity: The migration process now includes two potential paths - a snapshot-based path for tests and the standard migration path

di:
I think this is probably not worth the added complexity. I appreciate you putting the effort in here, it's hard to know if performance improvements are going to pay off in advance.

PR #17630, opened 2025-02-20
apply same max password length between forms and zxcvbn

ewdurbin:
zxcvbn@4.5.0 implements a max_length parameter with a default of 72.

Move our MAX_PASSWORD_SIZE to a constant and use it for the password forms as well as zxcvbn validator


miketheman (approved):

di (approved):
Needs translations.

PR #17649, opened 2025-02-24
feat: detect typos in project names during upload

miketheman:
Follows approaches described in `typoguard` paper and `typomania` project, with changes for local conditions.


ewdurbin (approved):
LGTM, assuming we can keep track of the performance via ddtrace.

di (approved):

PR #17662, opened 2025-02-26
fix flaky tests for user notification task

ewdurbin:


miketheman (approved):

miketheman:
This failed for me locally with `--randomly-seed=5645308` but wasn't reproducible, and didn't fail a second time around, so something is still a little flaky here.

ewdurbin:
https://github.com/pypi/warehouse/actions/runs/13587988564/job/37987277388?pr=17675#step:5:284

will look closer!

PR #17676, opened 2025-02-28
fix flaky test for user notification task

ewdurbin:
followup to #17662, in batch size test... just test batch size

ordering in the query won't necessarily match ordering of usernames... so just compare call counts. the other tests ensure the right users arenotified, let this one just look at the batch size


tests/unit/accounts/test_tasks.py
```diff
@@ -64,7 +64,7 @@ def test_notify_users_of_tos_update_respects_batch_size(
         "terms.revision": "initial",
         "terms.notification_batch_size": batch_size,
     }
-    users_to_notify = UserFactory.create_batch(20, with_verified_primary_email=True)
+    UserFactory.create_batch(max(1, batch_size * 2), with_verified_primary_email=True)
```
miketheman, line 67: It took me far longer than it should have to realize that the parametrized input was 10 not 20, and I kept trying to figure out why you were multiplying it by two.
Using max() is a nifty way to make the input of 0 work.

PR #17677, opened 2025-02-28
feat: add index to Provenance.file_id

miketheman:
When used as a JOIN condition in `legacy.api.json.release` for the release's files, the `provenance` table is consulted for any records related to the `file_id`.

Currently unindexed, this is a sequential scan of the table of ~164k rows, and ~87% of the total query cost.

Add an index to what is effectively a primary key to prevent further slowdown as the provenance table grows with more projects publishing.

### DIFF
--- a/warehouse/migrations/versions/7f0c9f105f44_create_attestations_table.py
+++ b/warehouse/migrations/versions/7f0c9f105f44_create_attestations_table.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 """
 create Attestations table
+
 Revision ID: 7f0c9f105f44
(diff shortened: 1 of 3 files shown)


warehouse/migrations/versions/7f0c9f105f44_create_attestations_table.py
```diff
@@ -11,6 +11,7 @@
 # limitations under the License.
 """
 create Attestations table
+
```
miketheman, line 14: This was breaking the output so I added it back

ewdurbin (approved):

ewdurbin (approved):

Issue #17702, opened 2025-03-03
Space management


devinrsmith:
We are currently in a situation where we are blocked on our project size limit of 10GB (https://github.com/pypi/support/issues/5628), but we are hoping to proceed ASAP to get new releases out to our users; it's unclear if yanking or deleting releases will free up space from the perspective of PyPi limits. Will yanking releases free up space? Will deleting releases free up space? Although less than ideal, we are willing to delete old releases if that means we can proceed in the meantime without needing to get an increase in our project size limit.

Somewhat related to https://github.com/pypi/warehouse/issues/17650, https://github.com/pypi/warehouse/issues/17520

di:
Hi @devinrsmith, answering your questions here:

> Will yanking releases free up space?

No.

> Will deleting releases free up space?

Yes.

Leaving this issue open as I think we could probably stand to update our docs accordingly.

woodruffw:
> Leaving this issue open as I think we could probably stand to update our docs accordingly.

I can look into doing this 🙂

woodruffw:
Forgot to loop back: #17716 addresses this!

PR #17716, opened 2025-03-05
docs: new and shiny storage limit docs

### LINKED ISSUE #9013: Cover "how do I delete my package/project?" in help docs
From https://fieldguide.opennews.org/en/latest/Chapter08-Handoffs-Sunsets/ I came across https://stackoverflow.com/questions/20403387/how-to-remove-a-package-from-pypi and realized that https://pypi.org/help/ and https://packaging.python.org do not explain how a maintainer or owner ought to delete or remove releases. https://pypi.org/help/#yanked is partially informative here but doesn't cover the general question of whether and how to delete a release or remove an entire project.

This is NOT a good issue for first-timers to take on because

a) we need to decide whether this goes in the PyPI FAQ or in PyPUG
b) I suspect there will be subtle phrasing concerns in order to properly say "you probably should not delete, you should yank instead, but if you need to, here's how".

woodruffw:
This is my attempt to scaffold out a clean segue from the existing `help.html` to the newer `mkdocs`-managed user documentation, using #17702 as a testcase.

Key changes:

* There's a new `request.user_docs_url` request method that makes it easy to route to user documentation pages, including specific anchors.
* I've added `/project-management/storage-limits/` with most of the file/project storage limit content from `help.html`, along with additional context/rendering niceties/screenshots.
* I've added `/project-management/yanking/` with the yanking content from `help.html`, along with a short user tutorial on yanking releases.
* I've removed the original yanking and file/project storage limit sections from `help.html` and changed the endpoints/views that linked it to link to the new docs page instead.

Some open questions:

* Is this a good idea at all? I think the newer user docs are easier to iterate on, but I don't know if it's worth the complexity of the new `user_docs_url` request method 😅 
* Does `/project-management/` make sense as a hierarchy for these docs? Or should this be just a top level `/storage-limits/` or similar, or something different?

See #17702.

Closes #9013.


woodruffw:
For reference, this is what the new storage limits docs look like with this PR:

image

dev/environment
```diff
@@ -5,6 +5,8 @@ WAREHOUSE_ENV=development
 WAREHOUSE_TOKEN=insecuretoken
 WAREHOUSE_IP_SALT="insecure himalayan pink salt"
 
+USERDOCS_DOMAIN="http://localhost:10000"
```
woodruffw, line 8: NB: This ensures that local deployments link to the local user docs build, rather than `docs.pypi.org` (which is otherwise the default).

warehouse/templates/pages/help.html
```diff
@@ -54,7 +54,6 @@
 {% macro installing() %}{% trans %}How do I install a file (package) from PyPI?{% endtrans %}{% endmacro %}
 {% macro publishing() %}{% trans %}How do I package and publish my code for PyPI?{% endtrans %}{% endmacro %}
 {% macro trove_classifier() %}{% trans %}What's a trove classifier?{% endtrans %}{% endmacro %}
-{% macro yanked() %}{% trans %}What's a "yanked" release?{% endtrans %}{% endmacro %}
```
di, line 57: Two concerns here: 
- We're losing the localization (this was raised by @ewdurbin)
- We're also losing links/context from `/help` (let's assume there is at least one external link to this, or that some search engine will deposit users here rather than in the docs).

Given that, I think we should probably leave the FAQ intact, but just have a brief answer that links to the relevant page in the user docs (and likely won't become out of sync).

warehouse/templates/pages/help.html
```diff
@@ -86,8 +85,6 @@
 {% macro collaborator_roles() %}{% trans %}What collaborator roles are available for a project on PyPI?{% endtrans %}{% endmacro %}
 {% macro request_ownership() %}{% trans %}How do I become an owner/maintainer of a project on PyPI?{% endtrans %}{% endmacro %}
 {% macro description_content_type() %}{% trans %}How can I upload a project description in a different format?{% endtrans %}{% endmacro %}
-{% macro file_size_limit() %}{% trans %}How do I get a file size limit exemption or increase for my project?{% endtrans %}{% endmacro %}
-{% macro project_size_limit() %}{% trans %}How do I get a total project size limit exemption or increase for my project?{% endtrans %}{% endmacro %}
```
di, line 90: Given https://github.com/pypi/warehouse/pull/17716/files/dfc8837225f0224d8e94b60e72b951e4075619fc..5d1452e32b44508906281208c9dcabc012c764ae#r1985565493, I would argue that we should add these back in a separate PR.

warehouse/templates/pages/help.html
woodruffw: Makes sense, I can revert this and the other removals + tweak the `help.html` to be a short answer that links to the user docs.

warehouse/templates/pages/help.html
woodruffw: I went ahead and re-added all three of these, with slightly tweaked language/links to the new user docs 🙂

di:
@woodruffw FYI, you need to rebuild translations here.

woodruffw:
> @woodruffw FYI, you need to rebuild translations here.

Done!

di (approved):

PR #17781, opened 2025-03-14
feat: add service for query cache results

miketheman:
Using Redis as a backing store for serialized results of longer, expensive queries that can be computed periodically, and then used in a client-facing operation where the query would take to long inline.


miketheman:
I also briefly looked at https://dogpilecache.sqlalchemy.org/ - that seemed a little heavyweight for this purpose, but might be worth exploring at some point? 🤔

ewdurbin (approved):

PR #17794, opened 2025-03-17
Conditional includes

di:
This PR updates our frontend JS to only fetch authed "includes" if there is an authenticated user ID stored in the cookie.

As a result it splits all includes into two sub-routes: `/_includes/authed` and `/_includes/unauthed/`, with the latter always being included.

Additionally, robots are permitted to make requests to any `/_includes/unauthed/` routes.


warehouse/static/js/warehouse/utils/html-include.js
```diff
@@ -19,6 +19,12 @@ const fetchOptions = {
 };
 
 export default () => {
+  // Check if we have an authenticated session
+  console.log(document.cookie);
```
ewdurbin, line 23:
```suggestion
```

warehouse/routes.py
```diff
@@ -76,65 +76,65 @@ def includeme(config):
...
-        "/_includes/administer-user-include/{user_name}",
+        "/_includes/authed/administer-user-include/{user_name}",
         factory="warehouse.accounts.models:UserFactory",
         traverse="/{user_name}",
         domain=warehouse,
     )
     config.add_route(
         "includes.submit_malware_report",
-        "/_includes/submit-malware-report/{project_name}",
+        "/_includes/authed/submit-malware-report/{project_name}",
```
ewdurbin, line 137: Noticed this one was still firing while unauth'd.

It appears that it is because the template uses `route_url` and not `route_path` so it gets a full URL (https://....).

Can we either update the JS that filters it to parse just the path, or fix that include to use route_path? https://github.com/pypi/warehouse/blob/071ee72a906815df6be0a2076e1f792986701f7c/warehouse/templates/includes/packaging/submit-malware-report.html#L17

warehouse/static/js/warehouse/utils/html-include.js
```diff
@@ -31,7 +36,14 @@ export default () => {
   // data-html-include attribute and replace it's content with that. This uses
   // the new fetch() API which returns a Promise.
   elements.forEach((element) => {
-    let p = fetch(element.getAttribute("data-html-include"), fetchOptions)
+    let url = element.getAttribute("data-html-include");
+    console.log(authed, url);
```
di, line 40:
```suggestion
```

warehouse/routes.py
di: Feels safer to just parse the path, I'll do that.

ewdurbin (approved):

PR #17795, opened 2025-03-17
More canonical URL tags

di:
This PR ensures that we always set a canonical URL tag, which should eliminate 1.11M indexing issues:

image


ewdurbin (approved):

di:
I think this needs a special-case for org pages too, which is being missed by the tests.

di:
This still needs more updates for other pages that need custom canonical URLs.

PR #17797, opened 2025-03-17
feat: compute top dependent corpus into cache

miketheman:
In preparation of being able to use a longer corpus, compute the top 10k dependents and place them in the query results cache.

The results are not yet used, so this is a good way of exercising the task logic without impeding any other processing.

### DIFF
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -396,11 +396,11 @@ def compute_top_dependents_corpus(request: Request) -> dict[str, int]:
     # 3. Converting to lowercase for normalization
     parsed_dependencies_cte = (
         select(
-            func.lower(
+            func.normalize_pep426_name(
                 # TODO: this isn't perfect, but it's a start.
                 #  A better solution would be to use a proper parser, but we'd need
                 #  to teach Postgres how to parse it.
-                func.regexp_replace(Dependency.specifier, "^([A-Za-z0-9_-]+).*", "\\1")
+                func.regexp_replace(Dependency.specifier, "^([A-Za-z0-9_.-]+).*", "\\1")
             ).label("dependent_name")
         )
         .select_from(recent_releases_cte)
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -30,6 +30,7 @@ from warehouse.metrics import IMetricsService
 from warehouse.packaging.interfaces import IFileStorage
 from warehouse.packaging.models import (
     Dependency,
+    DependencyKind,
     Description,
     File,
     Project,
@@ -407,7 +408,7 @@ def compute_top_dependents_corpus(request: Request) -> dict[str, int]:
         .join(Dependency, Dependency.release_id == recent_releases_cte.c.release_id)
         .where(
             recent_releases_cte.c.rn == 1,  # "latest" release per-project
-            Dependency.kind != 8,  # Exclude legacy project_url dependencies
+            Dependency.kind.in_([DependencyKind.requires_dist, DependencyKind.requires]),
         )
         .cte("parsed_dependencies")
     )
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -408,7 +408,9 @@ def compute_top_dependents_corpus(request: Request) -> dict[str, int]:
         .join(Dependency, Dependency.release_id == recent_releases_cte.c.release_id)
         .where(
             recent_releases_cte.c.rn == 1,  # "latest" release per-project
-            Dependency.kind.in_([DependencyKind.requires_dist, DependencyKind.requires]),
+            Dependency.kind.in_(
+                [DependencyKind.requires_dist, DependencyKind.requires]
+            ),
         )
         .cte("parsed_dependencies")
     )
--- a/warehouse/packaging/tasks.py
+++ b/warehouse/packaging/tasks.py
@@ -10,22 +10,36 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import datetime
 import logging
 import tempfile
+import typing
 
 from collections import namedtuple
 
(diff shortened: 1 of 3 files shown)


warehouse/packaging/tasks.py
```diff
@@ -342,3 +356,82 @@ def update_bigquery_release_files(task, request, dist_metadata):
...
+    # Result is Rows, so convert to a dicts of "name: count" pairs
+    results = {row.dependent_name: row.dependent_count for row in results}
+
+    # Store the results in the query results cache
+    cache = request.find_service(IQueryResultsCache)
+    cache_key = "top_dependents_corpus"
+    cache.set(cache_key, results)
+    logger.info("Stored `top_dependents_corpus` in query results cache.")
+
+    return results
```
ewdurbin, line 437: Do we actually want to return this object here? In practice I assume we do not. Assuming it is just for debugging.

In the case that having a function to call for debugging is useful, maybe we should refactor the bulk of this to a utility function and just call that in the task.

ewdurbin (commented):
I have never used Common Table Expressions before, so I'm not 100% confident in their behavior or performance, but this seems ~reasonable.

Do you have any idea what performance for this query will look like against a prod sized db?

warehouse/packaging/tasks.py
```diff
@@ -342,3 +356,82 @@ def update_bigquery_release_files(task, request, dist_metadata):
...
+    # 1. Taking the specifier from release_dependencies
+    # 2. Using regex to extract just the package name portion
+    # 3. Converting to lowercase for normalization
+    parsed_dependencies_cte = (
+        select(
+            func.lower(
+                # TODO: this isn't perfect, but it's a start.
+                #  A better solution would be to use a proper parser, but we'd need
+                #  to teach Postgres how to parse it.
+                func.regexp_replace(Dependency.specifier, "^([A-Za-z0-9_-]+).*", "\\1")
```
di, line 403: I believe this is going to fail to capture dependency specifiers where the project name contains periods, one example from the database:

```
warehouse=> select kind, specifier from release_dependencies where specifier ilike '%\.% (%' limit 1;
 kind |      specifier
------+----------------------
    4 | zope.event (==4.5.0)
(1 row)
```
Is that intentional?

ewdurbin (approved):

miketheman:
> Do you have any idea what performance for this query will look like against a prod sized db?

I ran the query on prod, and generated the explain analyze plan.
With the optimizations in place, this now takes ~40 seconds.

Explain Analyze ([visualized](https://explain.depesz.com/s/DFyR)):

<details>

```sql
                                                                                               QUERY PLAN                                                                                               
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=3056755.71..3056780.71 rows=10000 width=40) (actual time=40867.541..40868.995 rows=10000 loops=1)
   ->  Sort  (cost=3056755.71..3056907.31 rows=60637 width=40) (actual time=40867.540..40868.241 rows=10000 loops=1)
         Sort Key: (count(*)) DESC, (lower(regexp_replace(regexp_replace((release_dependencies.specifier)::text, '^([A-Za-z0-9_.-]+).*'::text, '\1'::text), '(\.|_|-)+'::text, '-'::text, 'ig'::text)))
         Sort Method: top-N heapsort  Memory: 1702kB
         ->  GroupAggregate  (cost=3049729.39..3052423.89 rows=60637 width=40) (actual time=40390.294..40819.848 rows=94575 loops=1)
               Group Key: (lower(regexp_replace(regexp_replace((release_dependencies.specifier)::text, '^([A-Za-z0-9_.-]+).*'::text, '\1'::text), '(\.|_|-)+'::text, '-'::text, 'ig'::text)))
               ->  Sort  (cost=3049729.39..3050273.84 rows=217781 width=32) (actual time=40390.286..40623.530 rows=2070839 loops=1)
                     Sort Key: (lower(regexp_replace(regexp_replace((release_dependencies.specifier)::text, '^([A-Za-z0-9_.-]+).*'::text, '\1'::text), '(\.|_|-)+'::text, '-'::text, 'ig'::text)))
                     Sort Method: external merge  Disk: 35160kB
                     ->  Hash Join  (cost=1760032.37..3025208.86 rows=217781 width=32) (actual time=8346.159..38378.026 rows=2070839 loops=1)
                           Hash Cond: (release_dependencies.release_id = recent_releases.release_id)
                           ->  Seq Scan on release_dependencies  (cost=0.00..1098031.00 rows=43555818 width=43) (actual time=0.678..9281.218 rows=43625972 loops=1)
                                 Filter: (kind = ANY ('{4,1}'::integer[]))
                                 Rows Removed by Filter: 1238996
                           ->  Hash  (cost=1759615.59..1759615.59 rows=33343 width=16) (actual time=8342.303..8342.306 rows=595411 loops=1)
                                 Buckets: 262144 (originally 65536)  Batches: 8 (originally 1)  Memory Usage: 6145kB
                                 ->  Subquery Scan on recent_releases  (cost=1526216.83..1759615.59 rows=33343 width=16) (actual time=6343.257..8209.391 rows=595411 loops=1)
                                       Filter: (recent_releases.rn = 1)
                                       ->  WindowAgg  (cost=1526216.83..1676258.89 rows=6668536 width=77) (actual time=6343.256..8153.097 rows=595411 loops=1)
                                             Run Condition: (row_number() OVER (?) <= 1)
                                             ->  Sort  (cost=1526216.83..1542888.17 rows=6668536 width=37) (actual time=6343.246..7435.649 rows=6666167 loops=1)
                                                   Sort Key: releases.project_id, releases.is_prerelease, releases._pypi_ordering DESC
                                                   Sort Method: external merge  Disk: 326200kB
                                                   ->  Seq Scan on releases  (cost=0.00..405687.66 rows=6668536 width=37) (actual time=0.007..2809.208 rows=6666167 loops=1)
                                                         Filter: (yanked IS FALSE)
                                                         Rows Removed by Filter: 44715
 Planning Time: 0.238 ms
 Execution Time: 40927.338 ms
(28 rows)
```

</details>

PR #17843, opened 2025-03-24
Don't index quarantined/archived projects

di:
This PR:
* ensures quarantined/archived projects are un-indexed from the search index when then are quarantined/archived
* ensures all quarantine/archived projects are not included when the search index is rebuilt


ewdurbin (approved):

warehouse/migrations/versions/d44f083952e4_add_new_lifecyclestatus_for_.py
```diff
@@ -0,0 +1,47 @@
...
+from alembic import op
+from alembic_postgresql_enum import TableReference
+
+revision = "d44f083952e4"
+down_revision = "6de76386dbf4"
+
+
+def upgrade():
+    op.execute(
+        "ALTER TYPE public.lifecyclestatus ADD VALUE IF NOT EXISTS 'archived-noindex'"
```
voidvii77, line 30: Execute

Issue #17860, opened 2025-03-27
Convert from ElastiCache Redis to ElastiCache Valkey


miketheman:
[Valkey](https://valkey.io/) was launched in AWS ElastiCache back in 2024, and claims backwards compatible wire-protocol compliance with open source Redis 7.2, while operating at 20% lower costs (more efficient credits usage for PyPI), and provides a zero-downtime upgrade path.
Read more on the launch blog here: https://aws.amazon.com/blogs/database/get-started-with-amazon-elasticache-for-valkey/ which also includes UI directions on how to perform the upgrade.

The listed differences are here: https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/VersionManagementConsiderations-valkey.html
By my reading, nothing there is concerning.

If any upgrades become blocked, there's this page to help: https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/resolving-blocked-engine-upgrades.html

To make the transition easy, here's some steps I conceive of:

- [ ] update local dev and CI to use valkey
- [ ] update testpypi's elasticache instance to valkey 7.2.x
- [ ] update pypi's elasticache instance to valkey 7.2.x (probably set a scheduled maintenance for this one)
- [ ] celebrate

As a second stage, once this is completed, we could also entertain replacing the client-side library from `redis-py` to either [valkey](https://pypi.org/project/valkey/) or [valkey-glide](https://pypi.org/project/valkey-glide/), but that's not strictly necessary right now.

avifenesh:
Hi :) 
When the migration to glide starts, feel free to ping for best practices, better usage of the client or the server (I do both), and anything else needed. Angry comments are acceptable, but enhancement requests are even more.
And when celebrating, ill also will be happy to take part! :)


And since that might be the best forum to ask -
https://github.com/valkey-io/valkey-glide/discussions/3464
Currently, Glide is supporting async API only, and there's extensive work to add support for sync API which, I believe, will be released soon.
A discussion opened by a community member brought up the idea of supporting anyio to allow structured concurrency.
If you have a take on how popular it is, effort of maintenance, and generally, which API you would use, ill be happy to hear.

PR #17861, opened 2025-03-27
chore(dev): replace redis with valkey

miketheman:
As part of the changeover, replace most of the references to `redis` as a service concept to `cache` since that is what function it performs.

Refs: https://github.com/pypi/warehouse/issues/17860


di:
Is this still worth pursuing given https://antirez.com/news/151?

ewdurbin:
Labeling as blocked, until @JacobCoffee reports back on migration technique.

ewdurbin:
> Is this still worth pursuing given https://antirez.com/news/151?

TBD based on if AWS picks up Redis OSS 8.x series or not. Seems plausible if ValKey continues its own journey that AWS will just keep trucking with that in Elasticache.

PR #17890, opened 2025-04-01
More org review tweaks

ewdurbin:
a lil more quality-of-life for me in da mines

### DIFF
--- a/warehouse/admin/templates/admin/organization_applications/detail.html
+++ b/warehouse/admin/templates/admin/organization_applications/detail.html
@@ -169,7 +169,7 @@
                 <div class="hidden" id="requestMoreInformationEmailTemplate">
                     Hello {{ organization_application.submitted_by.username }}!
                     We'd be happy to approve this organization request,
-                    but ask that you first add an @{% with parsed = organization_application.link_url|urlparse %}{{ parsed.host.lstrip("www.") }}{% endwith %}
+                    but ask that you first add an @{% with parsed = organization_application.link_url|urlparse %}{% if parsed.host %}{{ parsed.host.lstrip("www.") }}{% else %}NONE{% endif %}{% endwith %}
                     email address to your PyPI account,
                     complete the verification,
                     and respond using the linked form.
--- a/warehouse/admin/templates/admin/organization_applications/detail.html
+++ b/warehouse/admin/templates/admin/organization_applications/detail.html
@@ -169,7 +169,7 @@
                 <div class="hidden" id="requestMoreInformationEmailTemplate">
                     Hello {{ organization_application.submitted_by.username }}!
                     We'd be happy to approve this organization request,
-                    but ask that you first add an @{% with parsed = organization_application.link_url|urlparse %}{{ parsed.host }}{% endwith %}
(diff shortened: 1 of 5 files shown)


warehouse/admin/templates/admin/organization_applications/detail.html
```diff
@@ -169,7 +169,7 @@ <h4 class="modal-title" id="deferModalLabel">
                 <div class="hidden" id="requestMoreInformationEmailTemplate">
                     Hello {{ organization_application.submitted_by.username }}!
                     We'd be happy to approve this organization request,
-                    but ask that you first add an @{% with parsed = organization_application.link_url|urlparse %}{{ parsed.host }}{% endwith %}
+                    but ask that you first add an @{% with parsed = organization_application.link_url|urlparse %}{{ parsed.host.lstrip("www.") }}{% endwith %}
```
miketheman, line 172: question: can parsed.host ever not be a string, and would fail an lstrip?

warehouse/admin/templates/admin/organization_applications/detail.html
```diff
@@ -277,10 +277,12 @@ <h4 class="modal-title" id="declineModalLabel">
       </div>
     </div>
 
+    {% set information_requests = organization_application.information_requests %}
+    {% set outstanding_information_requests = organization_application.information_requests|selectattr("additional", "defined")|map(attribute="additional")|selectattr("response", "undefined")|list %}
```
miketheman, line 281: question: does the selectattr... and following commands work if `information_requests` is empty? I ask, as it's used as an `if information_requests` conditional later on. Same question applied for the pattern used in `list.html`

miketheman (approved):
Approving with some inline questions, since if it blows up, it only hurts admin users.

warehouse/admin/templates/admin/organization_applications/detail.html
ewdurbin: hm, worth checking since our validation is a lil naive... https://github.com/pypi/warehouse/blob/5c0cd842ef55f19fbb9db2ff5e799184b7a30bb5/warehouse/manage/forms.py#L606-L621

warehouse/admin/templates/admin/organization_applications/detail.html
ewdurbin: yep works for empty information_requests

warehouse/admin/templates/admin/organization_applications/detail.html
ewdurbin: bypassing HTML 5 validation and submitting `https://` yields 
Screenshot 2025-04-01 at 5 35 50 PM

miketheman: Maybe a min length of `https://` + 1 would do the trick?

ewdurbin: 22df10f7334202ead206c59cbe8e60d72c9c7246

miketheman: yeah, that'll work too 😆

miketheman (approved):

PR #17898, opened 2025-04-02
orgs: add fields to application, add more info to admin view

ewdurbin:


cmaureir (commented):
Read the content a couple of times, and it looks good to me. The code also looks good, but I'd keep someone with more Django experience to check that.

di (approved):
I don't have any Django experience, but this looks good to me 😉

warehouse/manage/forms.py
```diff
@@ -647,6 +649,35 @@ class CreateOrganizationApplicationForm(OrganizationNameMixin, SaveOrganizationF
...
+            wtforms.validators.Length(
+                min=32,
+                message=(
+                    "Tell us a little more about how you plan to use PyPI Organizations"
+                ),
+            ),
+            wtforms.validators.Length(
+                max=1024,
+                message=_(
+                    "We don't need to know quite that much :), "
```
miketheman, line 674: epic

miketheman (approved):
Appears generally correct - I haven't tested it directly.

I _think_ that the default for `col` is "span all 12 columns unless there's another column in this row" - so you might save some bits by replacing `col-12` with `col`, but that's not a sticking point.
`<wistful>`Some day, I'll replace all the hand-built HTML with macros and components, but today is not that day...`</wistful>`

PR #17922, opened 2025-04-08
Drop links to @PythonStatus, prefer status.python.org

di:


ewdurbin (approved):

ewdurbin:
Thanks @di! I think we need to copy/paste the warehouse/static/html/mirror.html file into https://github.com/pypi/infra/blob/main/terraform/warehouse/html/error.html

PR #17931, opened 2025-04-09
feat(admin): display an inspector link if found

miketheman:
If the inbound report contained an inspector link, display it on the
project list for speedy access.


miketheman:
Example:
image

Issue #17925, opened 2025-04-08
Normalize and store email addresses


miketheman:
Email addresses are currently stored in a `varchar(254)` column in the database, with a non-null, and unique constraint.

However, as some email providers allow for extra characters in email addresses, and the column is not `citext` (case insensitive), we can have duplicate values.

Proposal:

- add a new empty `normalized_email` citext column to `Emails` model
- populate the column with normalized values of each email address during email addition
- backfill existing records

This effort could be complemented by also adding a `domain` column to the table, and do the same work as normalization effort, to make the data representation very clear and unambiguous or reliant on string splitting.

This effort should be preceded by some queries on the table to determine how many of these we might expect to see.

miketheman:
Might be interesting to combine logic from https://pypi.org/project/email-normalize/ along with the `email_validator` we have today.
Definitely need a good way to find and solve for existing conflicting patterns.

PR #17946, opened 2025-04-12
Normalized email column in Email

Daksh2000:


Daksh2000:
> Fixes: #17925
> 
> In this PR, I have added 2 columns -> normalized_email and domain (both of which are derived from email itself) So, made 2 helper functions to get the corresponding values of normalized_email and domain
> 
> [ I did not use "normalize" property of "email_validator" library because according to it - after using it -> The local part (before @) remains case-sensitive (which I thought was defeating our purpose, so I went with custom function)
> 
> Also, Added migrations accordingly, and fixed some test cases.

Hi @miketheman , for normalization , 
could we use func.normalize_pep426_name() instead ?

PR #17981, opened 2025-04-17
Migrate entirely to our self-hosted Plausible instance for analytics

ewdurbin:
* Move blog.pypi.org from GA to analytics.python.org
* Add analytics.python.org to docs.pypi.org
* Remove support for Google Analytics


ewdurbin:
Once this is live, we can import the historic google analytics data into analytics.python.org

warehouse/static/js/warehouse/utils/analytics.js
```diff
@@ -1,53 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-/* global dataLayer */
-
-import * as cookie from "cookie";
-
-
-export default () => {
```
ewdurbin, line 19: Just a note that plausible already respects DNT and doesn't store PII, so an equivalent wrapper is not needed.

di (approved):

PR #17999, opened 2025-04-21
base: replace PEP 541 link with user documentation link

### LINKED ISSUE #17959: Add link to user documentation at https://docs.pypi.org/ in footer or other prominent location

I could not find a link to the Warehouse or PyPI user documentation anywhere. There is an FAQ, but it did not answer my question, specifically how to get unverified details to be verified for my package.

Add a link to the user documentation at https://docs.pypi.org/ in the footer or other prominent location.

**Additional context**

I eventually found the user documentation as follows.

1. Scrolled to the footer to look for the right link under Help.
2. [User guide](https://packaging.python.org/): wrong link.
3. [FAQ](https://pypi.org/help/): nope again.
5. [Contribute on GitHub](https://github.com/pypi/warehouse): Aha! Now we're getting somewhere!
6. Searched for the term [Unverified details](https://github.com/search?q=repo%3Apypi%2Fwarehouse%20Unverified%20details&type=code). Bingo!
7. Navigated to https://github.com/pypi/warehouse/blob/main/docs/user/project_metadata.md?plain=1
8. Read the relevant docs. Much rejoicing!
9. Saw a `README.md` in the same directory, and read it to finally find the URL to the PyPI User Documentation as https://docs.pypi.org/
10. Viewed and searched the HTML source of PyPI for `https://docs.pypi.org/` to ensure it wasn't hidden somewhere that I missed. Came up empty.

Thank you for your consideration.

woodruffw:
To keep things balanced between the different columns, I've removed the link to [PEP 541](https://peps.python.org/pep-0541/) and replaced it with a link to the user docs.

This has the downside of removing a link to the index's name retention polcies. However, I think there are two possible resolutions here:

1. Add the retention policies within the user docs (i.e. by copying them from the PEP and adding a "living" banner to the PEP, similar to other packaging PEPs)
2. Add the retention policies to <https://policies.python.org> (under the PyPI group) instead, since they're conceptually related to the other policies (AUP, T&C).

Curious what others think about either of the above 🙂 

Closes #17959.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -954,13 +954,12 @@ msgstr ""
 #: warehouse/templates/404.html:34 warehouse/templates/500.html:28
 #: warehouse/templates/accounts/two-factor.html:55
 #: warehouse/templates/base.html:274 warehouse/templates/base.html:275
-#: warehouse/templates/base.html:276 warehouse/templates/base.html:277
-#: warehouse/templates/base.html:287 warehouse/templates/base.html:288
-#: warehouse/templates/base.html:301 warehouse/templates/base.html:302
-#: warehouse/templates/base.html:304 warehouse/templates/base.html:313
-#: warehouse/templates/base.html:315 warehouse/templates/base.html:316
-#: warehouse/templates/base.html:317 warehouse/templates/base.html:327
-#: warehouse/templates/base.html:340
+#: warehouse/templates/base.html:277 warehouse/templates/base.html:287
+#: warehouse/templates/base.html:288 warehouse/templates/base.html:301
+#: warehouse/templates/base.html:302 warehouse/templates/base.html:304
+#: warehouse/templates/base.html:313 warehouse/templates/base.html:315
+#: warehouse/templates/base.html:316 warehouse/templates/base.html:317
+#: warehouse/templates/base.html:327 warehouse/templates/base.html:340
 #: warehouse/templates/includes/accounts/profile-callout.html:18
 #: warehouse/templates/includes/file-details.html:101
 #: warehouse/templates/index.html:100 warehouse/templates/index.html:104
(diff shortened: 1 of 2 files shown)


miketheman (commented):
I'm generally in favor of moving away from calling things a "PEP 541 request" and codifying it in a Policy, thus allowing further evolution of the policy and departing from the PEP number.

Not to increase scope too heavily, but that effort should be paired with updates to `pypi/support` tracker and other places where the policy is mentioned.

woodruffw:
> I'm generally in favor of moving away from calling things a "PEP 541 request" and codifying it in a Policy, thus allowing further evolution of the policy and departing from the PEP number.

Sounds good! In that case what I'll do is copy the policy bits of PEP 541 into a new PR against `policies.python.org`, which will then be a blocker before landing this 🙂 

> Not to increase scope too heavily, but that effort should be paired with updates to pypi/support tracker and other places where the policy is mentioned.

I can do a survey of these points.

woodruffw:
(Just responding to say I haven't forgotten about this. I'll probably do it at PyCon!)

woodruffw:
Revisiting this -- I think this is the stack of requirements for unblocking this:

- [ ] `policies.python.org` needs a new PyPI-specific "name retention policy" page that reflects the policy specified in PEP 541: https://github.com/psf/policies/pull/41
- [ ] PEP 541 needs to be updated to link to the above with the "historical reference" banner
- [ ] `pypi/support` needs changes to its `pep541-request.yml` issue template to have it reference the above policy instead of referencing a PEP by number
- [ ] This PR needs to be updated to reference `policies.python.org` where appropriate rather than PEP 541

@miketheman do you think I'm missing anything above? 🙂

miketheman:
I think that's correct - although we're probably missing a bunch of references somewhere. Maybe an update to the PEP page that states "this page is a historical reference..." kind of thing?

woodruffw:
xref: I've done the first part of above with https://github.com/psf/policies/pull/41

di:
> policies.python.org needs a new PyPI-specific "name retention policy" page that reflects the policy specified in PEP 541: https://github.com/psf/policies/pull/41

This is done: https://docs.pypi.org/project-management/name-retention/, I think the next step is to update the PEP?

PR #18008, opened 2025-04-22
test: add pytest-mock

miketheman:
See https://pytest-mock.readthedocs.io/

Example of using `mocker` fixture to add a `mocker.spy` for a single function and asserting its calls in the test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -550,8 +550,10 @@ def search_service():
 
 
 @pytest.fixture
-def domain_status_service():
-    return account_services.NullDomainStatusService()
+def domain_status_service(mocker):
+    service = account_services.NullDomainStatusService()
+    mocker.spy(service, "get_domain_status")
+    return service
 
 
 class QueryRecorder:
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -550,8 +550,10 @@ def search_service():
 
 
 @pytest.fixture
-def domain_status_service():
-    return account_services.NullDomainStatusService()
+def domain_status_service(mocker):
+    service = account_services.NullDomainStatusService()
+    mocker.spy(service, "get_domain_status")
```
miketheman, line 555: Note that the fixture sets up the tracking internally, and doesn't change what's returned by calling the fixture, only adds some methods.

If I try to spy on a method that doesn't exist, an `AttributeError` is raised, so that protects us from spying on things that don't exist, whereas a `call_recorder` on a created stub offers no such protection.

miketheman:
Closing, as it helped illustrate the point. The inclusion of the helper is in #18014

PR #18014, opened 2025-04-23
feat: add periodic task to update domain statuses

miketheman:
Introduces a task to periodically update the domain status of email addresses and includes several supporting changes to the codebase. The most significant updates involve adding a periodic task for batch domain status updates, refactoring domain status update logic into a reusable utility function, and optimizing database queries with a new index.

See each commit message for further details.

### DIFF
--- a/warehouse/accounts/tasks.py
+++ b/warehouse/accounts/tasks.py
@@ -10,9 +10,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import datetime, timedelta, timezone
+from __future__ import annotations
 
-from sqlalchemy import func
+import typing
+
+from datetime import UTC, datetime, timedelta, timezone
+
+from sqlalchemy import func, nullsfirst, or_, select
 
 from warehouse import tasks
 from warehouse.accounts.models import (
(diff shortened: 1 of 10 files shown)


warehouse/accounts/tasks.py
```diff
@@ -136,3 +144,32 @@ def compute_user_metrics(request):
             "primary:true",
         ],
     )
+
+
+@tasks.task(ignore_result=True, acks_late=True)
+def batch_update_email_domain_status(request: Request) -> None:
+    """
+    Update the email domain status for any checked over 30 days ago.
```
di, line 152:
```suggestion
    Update the email domain status for any domain last checked over 30 days ago.
```

warehouse/accounts/utils.py
```diff
@@ -44,3 +51,17 @@ class UserContext:
...
     def __principals__(self) -> list[str]:
         return self.user.__principals__()
+
+
+def update_email_domain_status(email: Email, request: Request) -> None:
+    """
+    Update the domain status of the given email address.
+    """
+    domain_status_service = request.find_service(IDomainStatusService)
+    domain_status = domain_status_service.get_domain_status(email.domain)
```
di, line 61: There's a failure case here where this returns `[]`:

https://github.com/pypi/warehouse/blob/a685531499e81433727bb51e7a0ef9db2ffbf0d2/warehouse/accounts/services.py#L1007-L1009

I think we shouldn't consider a domain for which we weren't able to get a status 'checked', and it should probably go back into the pool of unchecked domains to be retried, unless I'm missing something we will just store this empty status, which will remain until 30 days have passed.

miketheman: Thanks - I've updated the logic and included a test case to confirm that we'll re-check these before the threshold is reached, as we won't update the record at all.

di (approved):

PR #18021, opened 2025-04-24
chore: remove alt text from sponsor images

miketheman:
WAVE recommends removing duplicate alt text when the image is accompanies by identical text.


ewdurbin (commented):
why leave the empty tags? is that the guidance?

miketheman:
Yes, the [WAVE tool](https://wave.webaim.org/) states:

> What To Do
> Change either the alternative text or the adjacent text to eliminate the redundancy. In most cases, you can give the image empty/null alternative text (alt="") because the content of the image is already provided in context through text. Linked images may often be combined with the adjacent text into one link, in which case the image may be given null/empty alternative text (alt="").

di (approved):

PR #18034, opened 2025-04-25
feat(admin): icon to find a user account from email

miketheman:
Since these are not strongly-referenced via relationships, add an icon to perform a search for the user.


di (approved):
The reverse would be really nice to have too!

miketheman:
> The reverse would be really nice to have too!

Yeah, I was thinking of where to put that on the already-crowded user details email pane - if you have a good spot for it, happy to do in another PR

PR #18036, opened 2025-04-27
surface conflicting prefixes when reviewing organizations

ewdurbin:
Screenshot 2025-04-29 at 4 56 53 AM

### DIFF
--- a/warehouse/admin/templates/admin/organization_applications/detail.html
+++ b/warehouse/admin/templates/admin/organization_applications/detail.html
@@ -388,10 +388,31 @@
               </label>
               <div class="col-12">
                 <table class="table">
-                  <tr><th>Application</th><th>Submitted</th><th>Requestor</th></tr>
+                  <tr><th>Application</th><th>Status</th><th>Submitted</th><th>Requestor</th></tr>
                   {% for application in conflicting_applications %}
                     <tr>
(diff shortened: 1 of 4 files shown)


warehouse/admin/templates/admin/organization_applications/detail.html
```diff
@@ -388,10 +388,31 @@ <h3 class="card-title">Organization Request{% if information_requests %}{% if ou
...
+              <label class="col-12 control-label">
+                Conflicting Prefixes <i class="fa fa-clipboard-question text-orange"></i>
+              </label>
+              <div class="col-12">
+                <table class="table">
+                  <tr><th>Application</th><th>Status</th><th>Submitted</th><th>Requestor</th></tr>
+                  {% for application in conflicting_namespace_prefixes %}
+                    <tr>
+                      <td><a target="_blank" href="{{ request.route_url('admin.organization_application.detail', organization_application_id=application.id) }}">{{ application.name }}</a></td>
+                      <td>{{ application.status }}</td>
```
di, line 415: I think we could probably just replace the 'conflicting applications' section with this rather than potentially duplicating it here, it should be clear enough which are exact matches & which are prefixes?

di (approved):
lgtm otherwise

warehouse/admin/templates/admin/organization_applications/detail.html
ewdurbin: smart. 71d5f4b2777327b1b6179200c7953b0f4207b61b

PR #18037, opened 2025-04-27
Implement Social Auth

ewdurbin:
The goal here is to provide a way for PyPI users to _associate_ their PyPI User with social applications, **not** log-in with those applications.

These associations would be used for verification and association of identity for purposes like:

- Account Recovery
- Organization Review/Recovery


miketheman:
Superseded by https://github.com/pypi/warehouse/pull/19197

PR #18039, opened 2025-04-28
feat(admin): add modal to delete an email address

miketheman:
In some cases, removing an email address is desirable, such as when the user has already added a correct second address to their account, and editing the existing invalid one would be a duplicate record.

### DIFF
--- a/warehouse/admin/views/users.py
+++ b/warehouse/admin/views/users.py
@@ -10,8 +10,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import datetime
 import shlex
+import typing
 
 from collections import defaultdict
 from secrets import token_urlsafe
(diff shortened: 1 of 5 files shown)


warehouse/admin/views/users.py
```diff
@@ -301,6 +307,35 @@ def user_add_email(user, request):
...
+    route_name="admin.user.delete_email",
+    require_methods=["POST"],
+    permission=Permissions.AdminUsersEmailWrite,
+    uses_session=True,
+    require_csrf=True,
+    context=User,
+)
+def user_email_delete(user: User, request: Request) -> HTTPSeeOther:
+    email = request.db.scalar(
+        select(Email).where(Email.email == request.POST.get("email_address"))
```
di, line 320: This would remove the need to check against `user.emails`:
```suggestion
        select(Email).where(_and(
            Email.email == request.POST.get("email_address"),
            Email.user_id == user.id,
        )
```

miketheman: Great idea, simplifies a bit. Commit incoming!

PR #18055, opened 2025-04-30
feat: unverify expired domains

miketheman:
With domain status checking now in place, add a task to unverify any email address that is in one of the statuses considered "expired".

Add an Observation on the User, as well updates the unverify reason.


ewdurbin:
unverify + freeze?

di:
That would work but would probably be the most support-intensive option.

miketheman:
With #18088 in place, this can now move forward with performing unverification/add observation steps periodically. 

Based on this query, there's under 2k email addresses that should become unverified initally:

```sql
select domain_last_status, count(*) as count
from user_emails 
where domain_last_checked is not null 
and domain_last_status::text[] && ARRAY['undelegated','inactive','expiring','deleting','unknown']
and verified is true
group by domain_last_status
order by count desc;
```


| domain\_last\_status | count |
| :--- | :--- |
| {undelegated,inactive} | 1091 |
| {undelegated,active} | 236 |
| {unknown} | 87 |
| {inactive} | 24 |
| {undelegated} | 14 |
| {marketed,priced,undelegated,active} | 13 |
| {undelegated,inactive,premium} | 8 |
| {undelegated,active,deleting} | 6 |
| {inactive,active} | 4 |
| {undelegated,reserved} | 3 |
| {undelegated,claimed} | 3 |
| {marketed,priced,deleting} | 2 |
| {marketed,undelegated,inactive} | 2 |
| {undelegated,deleting} | 1 |
| {undelegated,claimed,active} | 1 |
| {marketed,priced,undelegated,active,deleting} | 1 |
| {active,deleting} | 1 |

Read more about the statuses here: https://domainr.com/docs/api/v2/status#status-results

warehouse/accounts/__init__.py
```diff
@@ -212,3 +213,6 @@ def includeme(config):
     config.add_periodic_task(
         crontab(minute="0", hour=4), batch_update_email_domain_status
     )
+    config.add_periodic_task(
+        crontab(minute="0", hour="*"), unverify_emails_with_expired_domains
```
ewdurbin: Should we move this to N minutes after the `batch_update_email_domain_status` task runs? that will make "time to remediation" N minutes, won't it?

warehouse/accounts/tasks.py
```diff
@@ -160,3 +162,49 @@ def batch_update_email_domain_status(request: Request) -> None:
...
+                    "unknown",
+                ]
+            ),
+            Email.verified.is_(True),
+        )
+        .order_by(Email.domain_last_checked)
+        .limit(1_000)
+    )
+
+    for email in request.db.scalars(stmt):
```
ewdurbin: Emit a datadog metric. I think it would be good to have quick visual or perhaps an alert on mass unverifications.

warehouse/accounts/__init__.py
miketheman: Updated to run after batch update

warehouse/accounts/tasks.py
miketheman: Added

warehouse/accounts/tasks.py
```diff
@@ -208,3 +210,7 @@ def unverify_emails_with_expired_domains(request: Request) -> None:
         )
 
         request.db.add(email)
+
+    request.metrics.gauge(
```
ewdurbin: I would have expected an increment call, won't a gauge persist until next report? Does it allow us to sum in aggregate over a time period, for like alerting?

miketheman: If I were tracking it in the loop, then I could emit an increment during each one, and then in the web ui I'd look to sum() them. 

Since there's little value to individual items since we're doing it during a batch call, calling gauge says effectively "this is how many there were when this task ran".

See the other existing calls to gauge in this module for existing behaviors.

ewdurbin: I think the difference is all other calls to gauge in this module are tracking a number that represents "number of X right now" while this is "number of X that just happened". increment seems correct, even if it is incrementing by the number that just occurred.

```
request.metrics.increment(
    "warehouse.emails.unverified", value=len(results), tags=["reason:domain_expired"]
)
```

Under the hood the datadog client will issue a `count` rather than an increment.

ewdurbin: and summing gauges is... precarious:
Screenshot 2025-06-10 at 4 41 49 PM

miketheman: yes, we shouldn't be summing gauges

ewdurbin: then a sum of "emails unverified over the last hour" as a driver for an alert is _wrong_, if this is submitted as a gauge.

miketheman: Converted to increment.

ewdurbin (approved):

PR #18057, opened 2025-04-30
chore: remove setuptools/wheel installation in image

miketheman:
As `setuptools` isn't needed during the installation and image building, we can remove the step that automatically upgrades it, along with the TODO about pinning.

The requirements files control the version of `setuptools`, removing one step of uninstalling the updated version during the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,12 +65,9 @@ RUN set -x \
 # our $PATH to refer to it first.
 ENV PATH="/opt/warehouse/bin:${PATH}"
 
-# Next, we want to update pip, setuptools, and wheel inside of this virtual
+# Next, we want to update pip and wheel inside of this virtual
 # environment to ensure that we have the latest versions of them.
-# TODO: We use --require-hashes in our requirements files, but not here, making
-#       the ones in the requirements files kind of a moot point. We should
-#       probably pin these too, and update them as we do anything else.
-RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip setuptools wheel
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -65,12 +65,9 @@ RUN set -x \
...
 ENV PATH="/opt/warehouse/bin:${PATH}"
 
-# Next, we want to update pip, setuptools, and wheel inside of this virtual
+# Next, we want to update pip and wheel inside of this virtual
 # environment to ensure that we have the latest versions of them.
-# TODO: We use --require-hashes in our requirements files, but not here, making
-#       the ones in the requirements files kind of a moot point. We should
-#       probably pin these too, and update them as we do anything else.
-RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip setuptools wheel
+RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip wheel
```
di, line 70: Because we're using `--no-deps`, we probably don't need to upgrade/install `wheel` either as we won't be building anything that requires it at build-time (and also `pip` now has build isolation with PEP 517, so it should be unnecessary as well).

https://github.com/pypi/warehouse/blob/e0f3c86a653db38e23a5d51aba0c6ec52666c0c5/Dockerfile#L87

miketheman: We still build some wheels (`pyscopg-c` et al), but I'm seeing that they don't require `wheel` to be installed for it all to work

di: Ah, was mixing this up with `--only-binary` which... maybe we should be using.

di (approved):

PR #18063, opened 2025-05-01
chore: pin dev-related dependencies

miketheman:
By placing the main ones first, most of the dev deps transitive requirements may be satisfies (like pip-tools -> setuptools) and thus prevent uninstallation/reinstallation later on.


di:
Not sure I follow what the reason for the change is here: are you saying that one set of dependencies requires a different versions of setuptools, which is causing uninstallation/reinstallation when installing the second set? Or what is it that we're trying to avoid here?

Since our dev dependencies are unpinned, I think this would result in us ending up with a different version of a different dependency installed in the image rather than what is in our main.txt file. Maybe the solution here would be to pin `dev.txt` as well?

miketheman:
> are you saying that one set of dependencies requires a different versions of setuptools, which is causing uninstallation/reinstallation when installing the second set?

This. One of the dev deps wants setuptools, so the latest will be installed. main.txt and others pin setuptools to often the non-latest version, so after dev installs it, main uninstalls and then installs the pinned version.

By reversing the order, the dev dep is satisfied and nothing else is downloaded/uninstalled/installed than necessary.

As to pinning, the dev deps don't make it into the production image, is that what you're worried about?

di:
No, I'm worried about developing/testing against a different version of a dependency than we would actually use in production. I think we should probably just pin our development dependencies to keep them consistent and to prevent the flux you're describing.

Dockerfile
```diff
@@ -182,7 +172,15 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     && pip check \
     && find /opt/warehouse -name '*.pyc' -delete
 
+# Install our development dependencies if we're building a development install
```
di, line 175: I think we don't need to change these lines anymore?

miketheman: yep, done

di (approved):

PR #18064, opened 2025-05-01
chore(dev,deps): remove specific ipython transitive deps

miketheman:
When #2702 introduced the list of dependencies, ipython was more in flux with other dependencies than it is today.

Allowing pip to resolve the necessary versions is preferable, especially as ipython changes over time.
For example, `ipython-genutils` and `pickleshare` are no longer needed.


di:
TBH I think we could probably just remove the IPython support, I don't think anyone is actively using it.

miketheman:
I actively use it daily

di:
TIL!

di (approved):

PR #18066, opened 2025-05-01
Switch from `pyqrcode` to `pyqrcode-binary`

di:
Towards https://github.com/pypi/warehouse/pull/18058.

This changes our dependency from https://pypi.org/project/PyQRCode/ (last release in 2016, no wheels available) to https://pypi.org/project/pyqrcode-binary/, which mirrors binary builds for each release of the original dependency.

### DIFF
--- a/requirements/main.in
+++ b/requirements/main.in
@@ -43,7 +43,7 @@ platformdirs>=4.2
 premailer
 psycopg[c]
 pydantic
-pyqrcode
+pyqrcode-binary
 pyramid>=2.0
 pymacaroons
 pyramid_jinja2>=2.5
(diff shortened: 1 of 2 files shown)


di:
Better how? I think the current library is working fine for our needs, it just needs wheels.

miketheman:
Better, as in reducing the amount of compiled binaries necessary to audit in the event of some problem.

di:
To be clear, `pyqrcode-binary` has pure Python wheels, there is no native code or binaries. So the tradeoff here is one wheel based on a dependency we already use vs. one wheel for a new dependency we'd have to adopt.

ewdurbin (approved):

PR #18088, opened 2025-05-05
chore: disallow password resets to unverified email

miketheman:
Only allow account resets to verified email addresses to prevent account
domain resurrection attacks.

If the user account exists for a given email address, check if the
account is verified. If it's not, reject with a flash message.

Does not disclose the name of a given user account, logs an event/warning for inspection.

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -2169,6 +2169,34 @@ class TestRequestPasswordReset:
         assert isinstance(result, HTTPSeeOther)
         assert result.headers["Location"] == "/the-redirect"
 
+    def test_unverified_email_is_rejected(self, db_request, mocker):
+        unverified_email = EmailFactory(verified=False)
+
+        # Prevent form's validation from checking deliverability
+        mock_form_validation = mocker.patch(
+            "warehouse.accounts.forms."
+            "RequestPasswordResetForm.validate_username_or_email",
+            autospec=True,
+            return_value=True,
+        )
+        # Spy on the flash method to check if it was called
+        mock_spy_flash = mocker.spy(db_request.session, "flash")
+
+        db_request.method = "POST"
+        db_request.POST = MultiDict({"username_or_email": unverified_email.email})
+        db_request.route_path = pretend.call_recorder(lambda a: "/the-redirect")
+
+        result = views.request_password_reset(db_request)
+
+        mock_form_validation.assert_called_once()
+        mock_spy_flash.assert_called_once_with(
+            f"Email address '{unverified_email.email}' is not verified. "
+            "Contact PyPI support for assistance.",
+            queue="error",
+        )
+        assert isinstance(result, HTTPSeeOther)
+        assert result.headers["Location"] == "/the-redirect"
+
 
 class TestResetPassword:
     @pytest.mark.parametrize("dates_utc", [True, False])
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -1906,7 +1906,7 @@ class TestRequestPasswordReset:
         stub_user = pretend.stub(
             id=uuid.uuid4(),
             email="foo@example.com",
-            emails=[pretend.stub(email="foo@example.com")],
+            emails=[pretend.stub(email="foo@example.com", verified=True)],
             can_reset_password=True,
             record_event=pretend.call_recorder(lambda *a, **kw: None),
         )
@@ -1977,8 +1977,8 @@ class TestRequestPasswordReset:
             id=uuid.uuid4(),
             email="foo@example.com",
             emails=[
-                pretend.stub(email="foo@example.com"),
-                pretend.stub(email="other@example.com"),
+                pretend.stub(email="foo@example.com", verified=True),
+                pretend.stub(email="other@example.com", verified=True),
             ],
             can_reset_password=True,
             record_event=pretend.call_recorder(lambda *a, **kw: None),
@@ -2055,7 +2055,7 @@ class TestRequestPasswordReset:
         stub_user = pretend.stub(
             id=uuid.uuid4(),
             email="foo@example.com",
-            emails=[pretend.stub(email="foo@example.com")],
+            emails=[pretend.stub(email="foo@example.com", verified=True)],
             can_reset_password=True,
             record_event=pretend.call_recorder(lambda *a, **kw: None),
         )
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -1845,18 +1845,22 @@ class TestRequestPasswordReset:
         ]
 
     def test_request_password_reset(
-        self, monkeypatch, pyramid_request, pyramid_config, user_service, token_service
+        self,
+        monkeypatch,
+        pyramid_request,
+        user_service,
+        token_service,
+        mocker,
     ):
-        stub_user = pretend.stub(
-            id=pretend.stub(),
-            username="username_value",
-            emails=[pretend.stub(email="foo@example.com", verified=True)],
-            can_reset_password=True,
-            record_event=pretend.call_recorder(lambda *a, **kw: None),
+        user = UserFactory.create(with_verified_primary_email=True)
+        mock_record_event = mocker.patch(
+            "warehouse.accounts.models.HasEvents.record_event",
+            autospec=True,
+            return_value=True,
         )
         pyramid_request.method = "POST"
         token_service.dumps = pretend.call_recorder(lambda a: "TOK")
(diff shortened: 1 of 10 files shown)


miketheman:
Screenshot 2025-05-05 at 13 51 57

warehouse/accounts/views.py
```diff
@@ -800,9 +800,38 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
...
+                pred=lambda e: e.email == form.username_or_email.data and e.verified,
             )
+
+            if not verified_email:
+                # If the user is found but the email is not verified,
+                # tell the user to contact support instead
+                if unverified_email := first_true(
+                    user.emails,
+                    pred=lambda e: e.email == form.username_or_email.data,
+                ):
```
di, line 814: This conditional doesn't work when attempting to request password reset by username: `verified_email` becomes `None`, but this query is searching for an email that matches the username, not the email address, so it will never succeed.

I'm surprised the tests & coverage are OK with this? Something is falling between the cracks here.

warehouse/accounts/views.py
```diff
@@ -800,9 +800,38 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
...
+                        email_address=unverified_email.email,
+                    )
+                    request.session.flash(
+                        request._(
+                            "Email address '${email}' is not verified. "
+                            "Contact PyPI support for assistance.",
+                            mapping={"email": unverified_email.email},
+                        ),
+                        queue="error",
+                    )
```
di, line 831: I'm also slightly concerned that this will become an oracle. I think instead, if we get a password reset for an unverified email, we should just send that address an email "Password reset requested but this email address is currently unverified" and just display the same ambiguous message here instead.

miketheman: Can you help me understand the threat vector this kind of oracle would provide?
It'll confirm the existence of an email address that isn't able to be used without an account recovery via support.

I'm generally okay to consider another email template to send to unverified email addresses, but that has the downside of us trying to email places that have already had either a hard bounce, expiration, or some other reason that the address is unverified, and could affect our sender scores.

warehouse/accounts/views.py
miketheman: The comparison isn't changed much from before - before verified, we were looking for the first true user.emails match based on the inbound form, so I'm not sure how it worked before.

miketheman: I hunted it down, and it's super subtle down here:
https://github.com/pypi/warehouse/blob/eb5e8fb62e48763f46868cc0fde6577f71cfb57c/warehouse/email/__init__.py#L96-L99

So if we never pass an email, it'll go to the primary on file.

di: Which might be an unverified email, yeah?

miketheman: Currently, yes. Which is the behavior I'm trying to disallow.

di: OK, but right now because you've set `allow_unverified=True`, when a user requests a password reset by username , and they don't have any verified emails on their account, nothing happens. They get the usual "If you submitted a valid username or email address, an email has been sent to your registered email address" but nothing happens: they have submitted a valid username but we won't send any emails.

warehouse/accounts/views.py
miketheman: We discussed this via chat and confirmed that changing the behavior is desirable.

warehouse/templates/email/password-reset-unverified/subject.txt
```diff
@@ -0,0 +1,17 @@
...
+ # Unless required by applicable law or agreed to in writing, software
+ # distributed under the License is distributed on an "AS IS" BASIS,
+ # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+-#}
+
+{% extends "email/_base/subject.txt" %}
+
+{% block subject %}{% trans %}Password reset request{% endtrans %}{% endblock %}
```
di, line 17:
```suggestion
{% block subject %}{% trans %}Password reset request for unverified email address{% endtrans %}{% endblock %}
```

This should be different than the regular "password reset" email subject, otherwise they might get threaded together.

warehouse/templates/email/password-reset-unverified/body.html
```diff
@@ -0,0 +1,31 @@
...
+{% block content %}
+  <p>
+    {% trans %}Someone, perhaps you, has made a password reset request for a PyPI account associated with this email address.{% endtrans %}
+  </p>
+  <p>
+    {% trans %}However, the email used to make this request is not verified. You must verify your email address before you can reset your password.{% endtrans %}
+  </p>
+  <p>
+    {% trans href=request.route_url('help', _anchor='account-recovery') %}Follow <a href="{{ href }}">account recovery steps</a> for your PyPI account if you are unable to verify your email address.{% endtrans %}
+  </p>
```
di, line 26: If the user has forgotten their password, they probably can't verify their email address, so telling them to verify their email isn't very helpful. This should probably say something like "if there is another email associated with your account, you should attempt a password reset via that email instead, otherwise you will need to follow the account recovery process."

warehouse/accounts/views.py
```diff
@@ -795,14 +796,46 @@ def request_password_reset(request, _form_class=RequestPasswordResetForm):
...
+                        "User requested password reset for unverified email",
+                        username=user.username,
+                        email_address=unverified_email.email,
+                    )
+                    user_service.ratelimiters["password.reset"].hit(user.id)
+
+                    token_service = request.find_service(ITokenService, name="password")
+                    n_hours = token_service.max_age // 60 // 60
+                    return {"n_hours": n_hours}
+
```
di, line 838: This still doesn't work right when a) there's a password reset by username and b) the user has no verified emails, since this conditional only happens when there's a password reset by email. We need to also call `send_password_reset_unverified_email` in this case.

I think we need to do something like this (untested):

```python
if request.method == "POST" and form.validate():
    requested_email = None
    user = user_service.get_user_by_username(form.username_or_email.data)
    
    if user:
        requested_email = user.primary_email
    else: 
        user = user_service.get_user_by_email(form.username_or_email.data)
        if user:
            requested_email = first_true(
                user.emails,
                pred=lambda e: e.email == form_field_input,
            )

    if requested_email and not requested_email.verified:
        # The requested email is not verified. Log the attempt, ping the rate limit,
        # notify to the email as to why no reset, return generic response.
        send_password_reset_unverified_email(
            request, (user, requested_email)
        )
        user.record_event(
            tag=EventTag.Account.PasswordResetAttempt,
            request=request,
        )
        request.log.warning(
            "User requested password reset for unverified email",
            username=user.username,
            email_address=requested_email.email,
        )
        user_service.ratelimiters["password.reset"].hit(user.id)

        token_service = request.find_service(ITokenService, name="password")
        n_hours = token_service.max_age // 60 // 60
        return {"n_hours": n_hours}

Comment (miketheman) [reply warehouse/accounts/views.py]: I think thios should now be covered by [eedf88f](https://github.com/pypi/warehouse/pull/18088/commits/eedf88fb6516eb6b48c084725963e795b6534575) - I had to perform a little more surgery on the tests since they had a lot of stubs, so I replaced some of the stub users with more real objects.

File: tests/unit/accounts/test_views.py [hunk 6]
```diff
@@ -2169,6 +2169,52 @@ def test_redirect_authenticated_user(self):
         assert isinstance(result, HTTPSeeOther)
         assert result.headers["Location"] == "/the-redirect"
 
+    @pytest.mark.parametrize(
+        "user_input",
+        [
+            "email",
+            "username",
+        ],
```
miketheman, line 2177: This was the material change to the unverified tests from the previous version - now this test takes either oif these as an input.

warehouse/templates/email/password-reset-unverified/body.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% extends "email/_base/body.html" %}
+
+{% block content %}
+  <p>
+    {% trans %}Someone, perhaps you, has made a password reset request for a PyPI account associated with this email address.{% endtrans %}
+  </p>
+  <p>
+    {% trans %}However, the email used to make this request is not verified. Your email address must be verified before you can use it to reset your password.{% endtrans %}
+    {% trans %}If you have another verified email address associated with your PyPI account, try that instead.{% endtrans %}
```
di, line 23:
```suggestion
    {% trans %}However, the email used to make this request is not verified. Your email address must be verified before you can use it to reset your password.{% endtrans %}
  </p>
  <p>
    {% trans %}If you have another verified email address associated with your PyPI account, try that instead.{% endtrans %}
```

di (approved):

PR #18101, opened 2025-05-08
allow for overriding cache decorator ttl on response

ewdurbin:
This is primarily to support discriminating between HTML and JSON rendered /simple/ index cache times.

In support of https://github.com/pypa/bandersnatch/pull/1898

### DIFF
--- a/tests/unit/cache/origin/test_fastly.py
+++ b/tests/unit/cache/origin/test_fastly.py
@@ -211,6 +211,33 @@ class TestFastlyCache:
             ),
         }
 
+    def test_override_ttl_on_response(self):
+        request = pretend.stub()
+        response = pretend.stub(headers={}, override_ttl=6969)
+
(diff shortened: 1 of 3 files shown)


cooperlees:
I don't know if this is possible, but could we reduce the TTL for only bandersnatch clients so others get what they get today and only bandersnatch get's lower and maybe lower than 1 hour?

I get this could be a DOS/DDOS enabling feature, but isn't the xmlrpc endpoint today anyways?
- I guess it's more separate/isolated tho

di (approved):

ewdurbin:
> I don't know if this is possible, but could we reduce the TTL for only bandersnatch clients so others get what they get today and only bandersnatch get's lower and maybe lower than 1 hour?

Not necessary since this should only result in one request to the backend every 30 minutes or so. The response will be be cached in our CDN.

PR #18130, opened 2025-05-13
blog: nitpicks

miketheman:
See commits for specifics.

- ~~Set up a privacy mode ~where we download and host all the artifacts "ourselves"~ to annotate links with `rel` attributes~~
- Update tags setting to remove warning
- leverage the related links feature to navigate between related posts

### DIFF
--- a/docs/mkdocs-blog.yml
+++ b/docs/mkdocs-blog.yml
@@ -53,6 +53,7 @@ plugins:
       # consider https://github.com/mkdocs/mkdocs-redirects
       post_url_format: "posts/{file}"
   - material-plausible
+  - privacy
   - rss:
       abstract_chars_count: -1
       categories:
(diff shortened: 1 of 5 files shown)


docs/mkdocs-blog.yml
```diff
@@ -53,6 +53,7 @@ plugins:
       # consider https://github.com/mkdocs/mkdocs-redirects
       post_url_format: "posts/{file}"
   - material-plausible
+  - privacy
```
ewdurbin: will this impact the plausible script also? I'm not 100% sure if we want that to be baked.

miketheman: It does - it downloads and bundles the script into the assets directory.

One approach could be to only protect [external links via config](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.assets), which is a step in that direction.

Sadly the open source version [doesn't yet contain the option config which assets/domains](https://squidfunk.github.io/mkdocs-material/plugins/privacy/#config.assets_exclude), so I'll update the PR to disable asset fetching for now, so at the very least we can protect external links.

miketheman: That didn't work as expected, so I'll remove that commit from scope.

ewdurbin (approved):

PR #18148, opened 2025-05-16
chore: update license headers to  use shorter SPDX header

miketheman:
Instead of the longer license header, replace with a shorter SPDX version.

Refs: https://spdx.dev/learn/handling-license-info/


miketheman:
CodeQL failures are due to touching files, and thus scanning "changed" files, so nothing is actually new here.

ewdurbin:
I'm:

- +1 to the outcome/intent.
- NULL on the actual implications of this from a licensing concern ([this document](https://spdx.dev/learn/handling-license-info/#:~:text=SPDX%20License%20IDs) says it's fine/sufficient but I am not an expert).
- +1 to including handling of `.git-blame-ignore-revs` to keep this from cluttering blame history

miketheman:
Closing now that the deed is done, and the ignore is part of #18219

PR #18156, opened 2025-05-19
lint: add zizmor check

woodruffw:
This adds `zizmor` to `lint.in` and runs it as part of `bin/lint`, i.e. `make lint`.

This in turn _should_ shorten the loop for local testing. One downside to this is that I think it'll effectively run `zizmor` twice in CI -- once in the linting job and once again in the dedicated job. We could maybe disable this by adding a `make local-lint` or similar, or having `bin/lint` skip `zizmor` when `CI=true`.

CC @ewdurbin 🙂


di:
> and once again in the dedicated job

Is there a reason to continue doing this if it's already run in the linting job?

woodruffw:
I believe @ewdurbin wanted it as part of their local development cycle, but it's definitely a bit duplicative with the `zizmor` job in CI.

@ewdurbin do you have a preference here? Another potential option here would be to remove `zizmor.yml` as a separate CI job and coalesce it into the rest of the listing steps, so that `make lint` would do the exact same thing both locally and remotely.

di:
Ah, maybe I'm not being clear, what I'm saying is: is there any advantage to a separate zizmor job in CI if we can run zizmor in linting in CI?

woodruffw:
Whoops, my misunderstanding! Yeah, there's not much advantage -- the main thing that running `zizmor` separately "gets" you is integration with GitHub via SARIF, but I'm not sure how useful you all find that vs. just having it fail in CI with results 🙂

miketheman:
Thanks for explaining the difference.
Going to close this for now, as it's not adding much beyond the CI action.

PR #18163, opened 2025-05-20
docs(dev): add meta docs on writing docs

woodruffw:
This adds some development-level docs on writing docs.

Note: this is an initial set; I'll follow up with more detailed sections on writing new sections, adding files, etc.

CC @Thespi-Brain @ewdurbin

### DIFF
--- a/docs/dev/development/documentation.rst
+++ b/docs/dev/development/documentation.rst
@@ -1,0 +1,85 @@
+#############
+Documentation
+#############
+
+Developer documentation
+=======================
+
+The developer documentation is hosted at `warehouse.pypa.io`_. It's written in
+`reStructuredText`_ and built using `Sphinx`_.
+
+.. _warehouse.pypa.io: https://warehouse.pypa.io
+.. _reStructuredText: https://docutils.sourceforge.io/rst.html
+.. _Sphinx: https://www.sphinx-doc.org/
+
+.. _dev-docs-layout:
+
+Layout
+------
+
+The developer documentation is located in the ``docs/dev`` directory.
+
+.. _dev-docs-local-dev:
+
+Local development
+-----------------
+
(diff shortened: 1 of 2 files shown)


docs/dev/development/documentation.rst
```diff
@@ -0,0 +1,85 @@
+#############
+Documentation
+#############
+
+Developer documentation
+=======================
+
+The developer documentation is hosted at `warehouse.pypa.io`_. It's written in
+`reStructuredText`_ and built using `Sphinx`_.
```
miketheman, line 9: We allow Markdown as well, by including `myst_parser`. See https://github.com/pypi/warehouse/blob/51ffd8ff87d763509486ac3cd03e07cdb92be693/docs/dev/architecture.md for one example.

miketheman (commented):
I am generally in favor of this, thanks for getting it started!

One nit inline.

How does this page relate to other documentation stuff like https://github.com/pypi/warehouse/blob/51ffd8ff87d763509486ac3cd03e07cdb92be693/docs/dev/development/getting-started.rst#building-documentation ? (that section is stale too, since we don't do Python 3.8 any more, and everything is kinda sorta self-contained)

woodruffw:
> How does this page relate to other documentation stuff like [`51ffd8f`/docs/dev/development/getting-started.rst#building-documentation](https://github.com/pypi/warehouse/blob/51ffd8ff87d763509486ac3cd03e07cdb92be693/docs/dev/development/getting-started.rst#building-documentation) ? (that section is stale too, since we don't do Python 3.8 any more, and everything is kinda sorta self-contained)

Huh, I completely missed that section! Unless you feel like it's worth keeping, perhaps we could remove that and break the doc documentation out fully? IMO there will be enough here for it to need its own hierarchy.

miketheman (approved):
Good to go!

PR #18164, opened 2025-05-20
docs: link to PyPI user docs more

woodruffw:
This adds two more prominent user docs links: one at the bottom of the help page, and one in the site-wide header (next to the help link).


warehouse/templates/pages/help.html
```diff
@@ -1031,6 +1031,7 @@ <h3 id="pronunciation">{{ pronunciation() }}</h3>
       <h3>{% trans %}Resources{% endtrans %}</h3>
       <p>{% trans %}Looking for something else? Perhaps these links will help:{% endtrans %}</p>
       <ul>
+        <li><a href="{{ request.user_docs_url('/') }}">{% trans %}PyPI User Documentation{% endtrans %}</a></li>
```
miketheman: This is technically an External link, and should be highlighted as such. Might as well also add the target/rel properties.

woodruffw: Fixed, thanks!

PR #18169, opened 2025-05-21
oidc: Refactor lookup strategies into single functions

facutuesca:
This PR refactors the Trusted Publishing "lookup strategies" pattern into a single `lookup_by_claims()` method for each of the publishers.

## Context
A strategy is a way of, given a set of OIDC claims, query the database for a matching Trusted Publisher. Concretely, a strategy is a function that takes a set of claims and returns a `Query` object. Each publisher has a list of these strategies, ordered from specific to general. When trying to find a Trusted Publisher, the more specific strategies are tried first, and if they fail the more general ones are tried.

For example, for the given set of claims for a GitHub OIDC token:
```json
{
    "repository": "foo/bar",
    "job_workflow_ref": "foo/bar/.github/workflows/release.yml@refs/heads/main",
    "environment": "my_environment",
}
```

first we ran a strategy that tried to find publishers with exactly those values (in particular, `environment==my_environment`). If that strategy failed, we tried the second strategy, where we tried to find publishers with `environment==None` (that is, allowing any environment).
The "specific to general" order in this case meant going from "Publishers that only allow `my_environment` as an environment" to "Publishers that allow any environment".

## New implementation
This PR changes the above approach to a single function per provider called `lookup_by_claims()` which takes a set of claims and returns a Publisher.

The multiple strategies are collapsed into a single query: we query for all publishers where the non-optional fields match the claims. In the example above, this means our single query looks for all publishers that match `repository` and `job_workflow_ref`, ignoring the `environment` value.

We then look at the resulting Python Publisher objects, and select the most specific one. 

## Rationale
The reasons for this change are:
- Removing the multiple lookup strategies, which are unnecessary since a single (general) query is enough
- Preparing for adding support for GitHub reusable workflows (as sketched in https://github.com/pypi/warehouse/issues/11096#issuecomment-2895081700). 

cc @woodruffw @miketheman @di

### DIFF
--- a/warehouse/oidc/models/_core.py
+++ b/warehouse/oidc/models/_core.py
@@ -13,7 +13,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, TypedDict, TypeVar, Unpack
+from typing import TYPE_CHECKING, Any, Self, TypedDict, TypeVar, Unpack
 
 import rfc3986
 import sentry_sdk
@@ -174,20 +174,10 @@ class OIDCPublisherMixin:
     # but there are a few problems: those claim sets don't map to their
     # "equivalent" column (only to an instantiated property), and may not
     # even have an "equivalent" column.
-    __lookup_strategies__: list = []
 
     @classmethod
-    def lookup_by_claims(cls, session, signed_claims: SignedClaims):
-        for lookup in cls.__lookup_strategies__:
-            query = lookup(cls, signed_claims)
-            if not query:
-                # We might not build a query if we know the claim set can't
-                # satisfy it. If that's the case, then we skip.
-                continue
-
-            if publisher := query.with_session(session).one_or_none():
-                return publisher
-        raise InvalidPublisherError("All lookup strategies exhausted")
+    def lookup_by_claims(cls, session, signed_claims: SignedClaims) -> Self:
+        raise NotImplementedError
 
     @classmethod
     def all_known_claims(cls) -> set[str]:
--- a/warehouse/oidc/models/gitlab.py
+++ b/warehouse/oidc/models/gitlab.py
@@ -12,7 +12,7 @@
 
 import re
 
-from typing import Any
+from typing import Any, Self
 
 from pypi_attestations import GitLabPublisher as GitLabIdentity, Publisher
 from sqlalchemy import ForeignKey, String, UniqueConstraint, and_, exists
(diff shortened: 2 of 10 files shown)


warehouse/oidc/models/_core.py
```diff
@@ -174,20 +174,10 @@ class OIDCPublisherMixin:
     # but there are a few problems: those claim sets don't map to their
     # "equivalent" column (only to an instantiated property), and may not
     # even have an "equivalent" column.
```
miketheman: I think this comment is now invalid and probably should be written to express the new design of "implement the checks in a subclass' `lookup_by_claims()`

warehouse/oidc/models/google.py
```diff
@@ -68,20 +69,20 @@ class GooglePublisherMixin:
...
+        query: Query = Query(cls).filter_by(email=signed_claims["email"])
+        publishers = query.with_session(session).all()
 
-    @staticmethod
-    def __lookup_no_sub__(klass, signed_claims: SignedClaims) -> Query | None:
-        return Query(klass).filter_by(email=signed_claims["email"], sub="")
+        if sub := signed_claims.get("sub"):
+            specific_publishers = [p for p in publishers if p.sub == sub]
+            if specific_publishers:
+                return specific_publishers[0]
```
miketheman: Optimize: This might be cleaner with `more_itertools.first_true`, like so:

```suggestion
            if specific_publisher := first_true(publishers, pred=lambda p: p.sub == sub):
                return specific_publisher
```

And in other places that are using this iteration pattern. Definitely worth thinking about other selector patterns from more-itertools, since we have it available for use.

Not a sticking point, but wanted to call out the pattern while I noticed it.

miketheman (commented):
I like the move from using a list-in-subclass to a function-that-contains-the-complexity, as that simplifies the overall design and allows individual publishers to decide their own lookup method.

I'm also a little wary of how well-tested these publishers are. There's [one functional test for provenance](https://github.com/pypi/warehouse/blob/f59df186dc62274b5831a72f639a1e92bbe3f94c/tests/functional/forklift/test_legacy.py#L322-L337) that creates a GitHub publisher, but everything else in theOIDC publishers stack seems heavily unit/mocked tests, so would it make sense to add some more functional tests around these interactions? The amount of mixins and abstractions make me a little cautions, being less familiar with the code behaviors.

warehouse/oidc/models/_core.py
facutuesca: good catch, fixed

warehouse/oidc/models/google.py
facutuesca: fixed

facutuesca:
@miketheman 
> so would it make sense to add some more functional tests around these interactions

I have a question that came up while trying to write the functional tests: I started the tests by POSTing to `/_/oidc/mint-token`, since the token exchange is where the trusted publisher lookup happens. However, since I'm not mocking the `request` object (like the unit tests do), I get a failure here:

https://github.com/pypi/warehouse/blob/79691b416ee1905902f383360986361c3e14494b/warehouse/oidc/services.py#L343-L346

since the `oidc.jwk_cache_url` settings key does not exist:

```
>       cache_url = request.registry.settings["oidc.jwk_cache_url"]
E       KeyError: 'oidc.jwk_cache_url'
```

Do we have a way to set it from the tests, without mocking the whole `request` object?

miketheman:
> Do we have a way to set it from the tests, without mocking the whole `request` object?

I believe the correct resolution for that would be to add an entry to:

https://github.com/pypi/warehouse/blob/f59df186dc62274b5831a72f639a1e92bbe3f94c/tests/conftest.py#L307-L342

Similar to `"celery.scheduler_url": "redis://localhost:0/",` this likely wants a "real" redis instance.

The `get_app_config` fixture is used in the `webtest` fixture creation

facutuesca:
Nice! That worked. I've added two functional tests with a GitHub trusted publisher (while also modifying the test config to use `NullOIDCPublisherService`, since we won't be able to verify JWTs in the tests due to their expiration).

Last question: I tried to add tests for the other providers, but OIDC is disabled for those providers in the admin flags of the test db:
```sh
E           webtest.app.AppError: Bad response: 422 Unprocessable Entity (not 200)
E           b'{"errors":[{"code":"not-enabled","description":"gitlab trusted publishing functionality not enabled"}],"message":"Token request failed"}\n'
```

these are the flags:
https://github.com/pypi/warehouse/blob/79691b416ee1905902f383360986361c3e14494b/warehouse/admin/flags.py#L18-L22



What would be the correct way to enable them?

miketheman:
> Last question: I tried to add tests for the other providers, but OIDC is disabled for those providers in the admin flags of the test db:
> ...
> What would be the correct way to enable them?

I haven't fully tested this, but this should be a good start.

In the context of `webtest`, we can dig through and get the preconfigured database session and use it:

```python
from warehouse.admin.flags import AdminFlag, AdminFlagValue

db_sess = webtest.extra_environ['warehouse.db_session']

flag = db_sess.get(AdminFlag, AdminFlagValue.DISALLOW_GITLAB_OIDC.value)
flag.enabled = False

webtest.post(....)
```

Or something close to it.

facutuesca:
> I haven't fully tested this, but this should be a good start.

Awesome, that worked perfectly. I pushed a commit adding the tests for the missing providers

warehouse/oidc/models/gitlab.py
```diff
@@ -178,47 +179,48 @@ class GitLabPublisherMixin:
...
+    # Get the most specific publisher from a list of publishers,
+    # where publishers constrained with an environment are more
+    # specific than publishers not constrained on environment.
+    @classmethod
+    def _get_publisher_for_environment(
+        cls, publishers: list[Self], environment: str | None
+    ) -> Self | None:
+        if environment:
+            if specific_publisher := first_true(
+                publishers, pred=lambda p: p.environment == environment.lower()
```
di, line 191: We should not be normalizing the environment name here, this introduced this bug: https://github.com/pypi/warehouse/issues/18330. @facutuesca Are you able to make a PR to fix this and add a test as a backstop?

facutuesca: @di on it

facutuesca: PR open here: https://github.com/pypi/warehouse/pull/18338

PR #18228, opened 2025-06-03
allow outsized headers on the upload endpoint

ewdurbin:
because macaroons


JacobCoffee (approved):

woodruffw:
Oof, are there really ~8KB macaroons being shipped over the wire? That seems huge even for our encoding; I wonder if someone stuffed some custom 3p caveats into a macaroon to see what PyPI would do.

PR #18238, opened 2025-06-07
oidc: add expiry key to mint_token response

### LINKED ISSUE #18235: Trusted Publishing: include an `expiration` key in the `mint-token` response?
Filing this for discussion/feedback before I send a small PR 🙂 

This idea came out of a discussion with @sigmavirus24 via the "automatic token refresh" idea in https://github.com/pypa/twine/issues/1246 -- the basic idea there is that `twine` (and other tools) want to automatically refresh their TP-issued API tokens so that long-running multi-file uploads don't fail when the token expires between files.

From this, the observation is that `twine` (and other tools) need a way to know _when_ they should refresh the API token. To do this they need to know when the token is going to expire, so that they can refresh with sufficient leeway.

This is technically possible today: any tool that uses the `mint-token` endpoint to do a TP exchange can decode the resulting macaroon and extract the expiration caveat. However, doing so means employing a macaroon deserializer, which is a nontrivial dependency to add to foundational dependencies like `twine`. Additionally, it requires client tools to track and maintain compatibility with PyPI's caveat encoding, which is _not_ a public API (and has changed in the past).

## Proposed solution

Rather than requiring tools to decode and consume the macaroon's caveat, the `mint-token` endpoint could duplicate the expiry directly in the response.

This should be trivial to do, since the expiry is computed directly in the `mint_token` helper right before being added to its caveat:

https://github.com/pypi/warehouse/blob/bb6a22cea5ffd0c1957b4a3502f4b390db451c6c/warehouse/oidc/views.py#L278

In effect, the success response from `mint-token` would change from:

```json
{
  "success": true,
  "token": "pypi-..."
}
```

to:

```json
{
  "success": true,
  "token": "pypi-...",
  "expires": <unix timestamp>
}
```

...where `<unix timestamp>` is the exact same timestamp as in the macaroon's expiration caveat.

I believe this poses no significant backwards compatibility risk, since it's a new field.

## Alternatives considered

1. Instead of attempting to refresh the token "intelligently," tools could refresh it on *every* file upload. This would work and avoid the immediate expiration problem, but would result in a much larger volume of security event noise (since each TP exchange produces a project-level security event).
1. Tools could absorb a bit of additional complexity and decode the Macaroon themselves, although this would come with compatibility risks around the caveat encoding.

CC @di @ewdurbin @miketheman for thoughts 🙂

woodruffw:
This allows clients that do TP exchanges to keep
track of the short-lived token's expiry without
poking through the macaroon's caveats.

Closes #18235.


sigmavirus24 (approved):

warehouse/oidc/views.py
```diff
@@ -340,7 +340,7 @@ def mint_token(
         metrics = request.find_service(IMetricsService, context=None)
         metrics.increment("warehouse.oidc.mint_token.github_reusable_workflow")
 
-    return {"success": True, "token": serialized}
+    return {"success": True, "token": serialized, "expires": expires_at}
```
sigmavirus24, line 347: Need to update around L42 to include this in the TypedDict definition

woodruffw: Fixed!

warehouse/oidc/views.py
```diff
@@ -39,6 +39,7 @@ class JsonResponse(TypedDict, total=False):
     message: str | None
     errors: list[Error] | None
     token: StrictStr | None
+    expiration: int | None
```
sigmavirus24:
```suggestion
    expires: int | None
```

sigmavirus24 (changes requested):

warehouse/oidc/views.py
woodruffw: :facepalm:

sigmavirus24 (approved):

facutuesca (approved):

warehouse/oidc/views.py
```diff
@@ -39,6 +39,7 @@ class JsonResponse(TypedDict, total=False):
     message: str | None
     errors: list[Error] | None
     token: StrictStr | None
+    expires: int | None
```
miketheman, line 45: question: how can this ever be `None` when an OIDC is minting this? If it's `None` ever in the context of OIDC exchange, then it would not expire, and that's counter to the Trusted Publishing design.

miketheman (changes requested):
Generally okay, one question inline needs to be addressed first.

warehouse/oidc/views.py
facutuesca: when the minting fails, like here:
```
def _invalid(errors: list[Error], request: Request) -> JsonResponse:
    request.response.status = 422
    return {
        "message": "Token request failed",
        "errors": errors,
    }
```

miketheman: An easyish workaround to prevent None would be to return `-1` or such for the invalid response, and thus prevent accidental non-expiring tokens.

facutuesca: This is the response that we send to users after a token is minted. The expiration of the token is set before this, here:
https://github.com/pypi/warehouse/blob/82e9149eed960a59b50bbd267326db6df7c2ec31/warehouse/oidc/views.py#L278

The JSON response only communicates this value to the user. If we incorrectly set `expires` to `None` in the response, the token would still expire at the correct time.

miketheman: Oh, that makes more sense. In that case, a comment here would help prevent future confusion as to how that works.

facutuesca, line 45: @miketheman something like this?

```suggestion
    # This value is used to communicate the minted token's expiration
    # time to the user. It is not used to determine the expiration,
    # changing this field does not change the token's expiration time.
    expires: int | None
```

miketheman: Yes, that ought to do it!

facutuesca: I don't have permissions to push to this branch, I'll leave the suggestion here so that @woodruffw can accept it.

woodruffw: Done, thanks @facutuesca 🙂

sigmavirus24 (approved):

miketheman (approved):

miketheman:
@woodruffw Branch needs an update, and then I can merge.

PR #18276, opened 2025-06-16
chore: ignore Simple API templates from djlint formatting

### LINKED ISSUE #18275: revert HTML page change, broke artifactory
Hello, 

The new release of pypi warehouse changed the HTML output and broke the HTML API.
It's causing our artifactory to fail to fetch packages from pypi, as it can no longer parse the page.

Can you please revert this change?
this file https://github.com/pypi/warehouse/blob/main/warehouse/templates/api/simple/detail.html
PR https://github.com/pypi/warehouse/pull/18266

Explanation, the PR automatically reformatted all the HTML templates with a formatting tool and it changed the output.
The changed output is breaking tools that are no longer able to parse it and it's worse for performance (more whitespaces).

Output now: 
Image

Output before: I can't take a screenshot since it gone, but it should be exactly one package per line and no whitespace.
PR:

Image

Explanation: The HTML page is the pypi API, it is downloaded and processed by numerous python packaging tools to identify available packages.
Historically pypi did not a have a json API, the HTML page is the API.
We've noticed that our Artifactory started failing to mirror/download packages from pypi.org on Friday, we tracked down the issue to this recent PR. (obviously artifactory wrote their parser poorly!).

Second issue: the reformatting added whitespaces and new lines, that makes every page a few percent larger, which makes both warehouse and packaging tools slower (a lot of packages have thousands of files available and the page is very large!). 
It's bad for performance, the whitespace needs to be removed.
The parsing of the /simple HTML page is a noticeable amount of the time taken to install packages (for old pip and tools that haven't been updated to use the json API instead). I bet it's a fair amount of the time taken for warehouse to generate the page too.

Can you revert the change please? Don't need to revert all the PR, just the changes to this file.
(or if you prefer to adjust the template instead, the jinja template needs to have a minus sign `{%-` and `-%}` to strip whitespaces before/after the template.)

miketheman:
Closes #18275
Closes #18284

### DIFF
--- a/warehouse/templates/api/simple/detail.html
+++ b/warehouse/templates/api/simple/detail.html
@@ -12,13 +12,15 @@
     <h1>Links for {{ name }}</h1>
     {% for file in files -%}
       <a href="{{ file.url }}#sha256={{ file.hashes.sha256 }}"
-         {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}"{% endif %}
-         {% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}
-         "
-         {% endif %}
-         {% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}
-         {% if file.get('provenance') %}data-provenance="{{ file['provenance'] }}"{% endif %}>{{ file.filename }}</a>
-      <br />
+        {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}"{% endif %}
+        {% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}
+        "
+      {% endif %}
+      {% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}
+      {# djlint:off #}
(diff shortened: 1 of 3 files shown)


warehouse/templates/api/simple/detail.html
```diff
@@ -12,13 +12,15 @@
...
-         {% endif %}
-         {% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}
-         {% if file.get('provenance') %}data-provenance="{{ file['provenance'] }}"{% endif %}>{{ file.filename }}</a>
-      <br />
+        {% if file.get('requires-python') %}data-requires-python="{{ file['requires-python'] }}"{% endif %}
+        {% if file.yanked %}data-yanked="{% if file.yanked is string %}{{ file.yanked }}{% endif %}
+        "
+      {% endif %}
+      {% if file['core-metadata'] %}data-dist-info-metadata="sha256={{ file['core-metadata']['sha256'] }}" data-core-metadata="sha256={{ file['core-metadata']['sha256'] }}"{% endif %}
+      {# djlint:off #}
```
ewdurbin: thought: can we disable djlint for all templates under `api/simple` somehow and nevert this fully to the template we were serving before? that seems more assured to create the expected result.

ewdurbin (approved):

warehouse/templates/api/simple/detail.html
miketheman: I like this better, but still dislike that folks are relying on the textual structure vs HTML parsing, sigh. Change inbound.

ewdurbin: of course. of course. thank you!

PR #18277, opened 2025-06-16
perf: update template renderer paths to use package qualifiers

brettlangdon:
Avoid unnecessary template lookup failures by specifying full package paths for all template renderers. Previously, for endpoints like /simple/<package>, the template resolver would first attempt to find 'warehouse.api:api/simple/detail.html' which would always fail, then fall back to 'templates/api/simple/detail.html'. By using explicit package qualifiers like 'warehouse:templates/api/simple/detail.html', we skip the initial failed lookup and exception handling.

This provides a small performance optimization by eliminating known failed disk lookups, as well as avoiding exception traceback generation when ddtrace is enabled (for attaching the traceback to the span metadata).

Testing locally, the template load time goes from ~830us to ~80us, and template rendering time goes from ~4ms to ~3.7ms... a small optimization, but an easy win.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -151,7 +151,7 @@ def incomplete_password_resets(exc, request):
 @view_config(
     route_name="accounts.profile",
     context=User,
-    renderer="accounts/profile.html",
+    renderer="warehouse:templates/accounts/profile.html",
     decorator=[
         origin_cache(1 * 24 * 60 * 60, stale_if_error=1 * 24 * 60 * 60)  # 1 day each.
     ],
@@ -228,7 +228,7 @@ def profile(user, request):
 
 @view_config(
     route_name="accounts.search",
-    renderer="api/account_search.html",
+    renderer="warehouse:templates/api/account_search.html",
     uses_session=True,
 )
 def accounts_search(request) -> dict[str, list[User]]:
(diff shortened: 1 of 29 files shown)


brettlangdon:
What should we do about: https://github.com/pypi/warehouse/blob/3a1f8c86e59a4d54bfaf2ce6c42d59d918730600/warehouse/config.py#L728-L733

Is this proposed change even desired for the minimal gain vs having the default search path?

brettlangdon:
An alternative here is to move all the templates into their respective module paths, e.g. `warehouse/templates/api/simple/detail.html` -> `warehouse/api/templates/simple/detail.html` and then update `renderer="templates/simple/detail.html" (since `warehouse.api:` will get prepended to all template lookups by default which don't have `:` in the name)

miketheman (commented):
I generally like this, as it makes it more explicit and clear which template is where.

Sadly, the functional tests (`webtest`-style tests) don't cover every `@view_config` instance to validate that the changes are safe, and we'd only find out at runtime if the template is truly missing. 

Can you conceive of a way to validate that each of these template paths exists in a lint/test phase, or another way to verify that these string changes don't break at runtime?

miketheman:
https://github.com/pypi/warehouse/pull/18366 likely solves the concern, so once that lands, this PR should be checked against that for accuracy.

miketheman:
@brettlangdon looks like the checker I wrote doesn't yet account for the updated paths - would you want to take a stab at adding that?

miketheman:
> @brettlangdon looks like the checker I wrote doesn't yet account for the updated paths - would you want to take a stab at adding that?

I did it :P https://github.com/pypi/warehouse/pull/18388

PR #18321, opened 2025-06-30
fix delete call to opensearch py client

ewdurbin:


warehouse/search/tasks.py
```diff
@@ -196,7 +196,8 @@ def reindex(self, request):
                     actions.append({"remove": {"index": name, "alias": index_base}})
                 actions.append({"add": {"index": new_index_name, "alias": index_base}})
                 client.indices.update_aliases(body={"actions": actions})
-                client.indices.delete(",".join(to_delete))
+                for index_to_delete in to_delete:
+                    client.indices.delete(index=index_to_delete)
```
miketheman, line 200: I think this method still accepts a comma-delimited string, but needed the kwarg instead of the positional arg:

```suggestion
                client.indices.delete(index=",".join(to_delete))
```

This is likely to work as well, but will make multiple calls to the index if there's more than one index to delete at a time. I don't think we test for that though.

Issue #18330, opened 2025-07-01
Trusted publishing from GitLab CI/CD fails with caps in environment name


francispotter:
_Previous title: "Trusted publishing to a project in an orgnization from GitLab CI/CD failed with 'no corresponding publisher' (previously worked)"_

Trusted publishing to a project in an organization from GitLab CI/CD **with capital letters in the environment name** failed with error:

```
* `invalid-publisher`: valid token, but no corresponding publisher     
  (Publisher with matching claims was not found)
```

Note the exact same project on GitLab (same code base, configuration, etc) published successfully to the exact same trusted publisher target on PyPI.org just two days ago.

---

I've seen the behaviour on two projects: [kwark](https://pypi.org/project/kwark/) and [busy](https://pypi.org/project/busy/). I own both projects and recently transferred them to the [steamwiz](https://pypi.org/org/steamwiz/) organization, which I also own.

Here's the trusted publisher setup for `busy`, which previously worked:

<img width="651" alt="Image" src="https://github.com/user-attachments/assets/f8509a21-0b9a-4687-984c-8eaaf1899cc5" />

Here's the job configuration, which previously worked and now failed:

```yaml
pypi-publish-job:
  image: python:3.13
  variables:
    GIT_STRATEGY: none
  id_tokens:
    PYPI_ID_TOKEN:
      aud: pypi
  script:
  - pip install twine
  - twine upload ./dist/*
  environment:
    name: PyPI
    url: https://pypi.org/project/${CI_PROJECT_NAME}

  before_script:  # Removed for clarity
  artifacts:      # Removed for clarity
  extends:        # Removed for clarity
  rules:          # Removed for clarity
  stage:          # Removed for clarity
```

(Note the actual build operation happens in an earlier job with delivers the dist as artifacts for the publishing job.)

The [CI config itself is public](https://gitlab.com/steamwiz/busy/-/blob/main/.gitlab-ci.yml) and the job comes from some nested includes from a [CI config library](https://gitlab.com/procicd/lib/-/blob/base/python/pypi-publish-job.gitlab-ci.yml).

The successful publishing job, with identical configuration to the present one, took place 'June 28, 2025 at 4:36:04 p.m. PDT' (2 days ago). The job log output included:

```
$ twine upload ./dist/*
Uploading distributions to https://upload.pypi.org/legacy/
Uploading busy-7.13.2-py3-none-any.whl
[...]

PR #18338, opened 2025-07-02
oidc: fix bug when matching GitLab environment claims

facutuesca:
Bug reported here: https://github.com/pypi/warehouse/issues/18330

The OIDC publisher lookup refactor in https://github.com/pypi/warehouse/pull/18169 introduced a bug where we normalize the incoming environment claim (making it lowercase) when doing the DB lookup for matching GitLab publishers.

This means that the lookup would incorrectly fail when the environment was capitalized (e.g: `SomeEnvironment`), since it would be lowercased before searching the DB (which contained the correct capitalized version).

Inversely, a lookup would incorrectly succeed when the incoming environment claim was capitalized but the environment configured in the Publisher was not. These are different environments (since GitLab's environments are case-sensitive), but by lowercasing the incoming claim, we would match against the publisher with a different (lowercase) claim.

Luckily this second scenario does not result in successful authentication, since this buggy lookup logic is separate from the "claim checking" logic, which happens after a publisher is found, and [does not contain the bug](https://github.com/pypi/warehouse/blob/256248f6ae18aa5fb45d5001f85af30c675421bf/warehouse/oidc/models/gitlab.py#L89-L105).

This PR removes the incorrect `.lower()` call, and adds tests for the two scenarios mentioned above

This fixes https://github.com/pypi/warehouse/issues/18330

cc @di @miketheman

### DIFF
--- a/tests/unit/oidc/models/test_gitlab.py
+++ b/tests/unit/oidc/models/test_gitlab.py
@@ -71,6 +71,59 @@ class TestGitLabPublisher:
         ):
             gitlab.GitLabPublisher.lookup_by_claims(pretend.stub(), signed_claims)
 
+    @pytest.mark.parametrize("environment", ["SomeEnvironment", "SOME_ENVIRONMENT"])
+    def test_lookup_succeeds_with_non_lowercase_environment(
+        self, db_request, environment
+    ):
+        # Test that we find a matching publisher when the environment claims match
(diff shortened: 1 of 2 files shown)


miketheman (approved):
I confirmed that the added tests expose the problem if the call to `.lower()` remains.

miketheman:
Thanks for the speedy fix, @facutuesca !

PR #18349, opened 2025-07-08
feat: registration email confirmation dialog

miketheman:
During registration, users will be prompted to confirm their email
address visually prior to proceeding.

By presenting the user with the dialog, they have a chance to review the
submitted email address one last time, hopefully preventing them from
getting into a registration "stuck" loop as a result of a typo.


miketheman:
Here's what it looks like:

https://github.com/user-attachments/assets/d0fd96bf-bc02-4fff-8227-8d9a245f0e02

ewdurbin (approved):
I think this is a really straightforward, but seemingly novel, solution. I wonder why it isn't implemented like this other places that's all.

One thought would be to have them type it again, but this is less intrusive so thumbs up from me.

PR #18353, opened 2025-07-09
fix: enable client-side menu post-register

miketheman:
Once a user has completed registration, the cookie used to determine
what to display in the menu dropdown doesn't yet exist, so it's
confusing as to what to do next.

Includes a refactor commit to consolidate cookie setting logic.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -765,9 +765,24 @@ def register(request, _form_class=RegistrationForm):
         send_email_verification_email(request, (user, email))
         email_limiter.hit(user.id)
 
-        return HTTPSeeOther(
+        resp = HTTPSeeOther(
             request.route_path("index"), headers=dict(_login_user(request, user.id))
         )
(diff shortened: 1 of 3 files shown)


warehouse/accounts/views.py
```diff
@@ -344,22 +344,7 @@ def login(request, redirect_field_name=REDIRECT_FIELD_NAME, _form_class=LoginFor
...
-                # IDs here, even though it really shouldn't matter.
-                resp.set_cookie(
-                    USER_ID_INSECURE_COOKIE,
-                    hashlib.blake2b(
-                        str(userid).encode("ascii"), person=b"warehouse.userid"
-                    )
-                    .hexdigest()
-                    .lower(),
-                )
+                _set_userid_cookie(resp, userid)
```
ewdurbin:
```suggestion
                _set_userid_insecure_cookie(resp, userid)
```


Since there are two cookies, I think it's probably best that this one is shown to be insecure in the call

ewdurbin (approved):

PR #18357, opened 2025-07-10
fix: update domain status during verification

### LINKED ISSUE #18356: Domain status should be re-checked during email verification
**Describe the bug**
If an email address domain enters a stage that triggers the "unverification" status, when the end user logs in to re-verify the address, the domain status is **not** checked and updated.
The previous status will be used during the daily check and become unverified, leading to a confusing state by the end user, prompting them to reverify unnecessarily and disrupting their access.

**Expected behavior**
Once a user has verified their email status, run the check and update.

miketheman:
When verifying an email, trigger a status update immediately, replacing existing domain status markers.

Prevent previously-existing domain statuses from impacting newly-verified domains, which we have a high degree of confidence in their status as the user has just received an email to this address and clicked the link.

Includes test case refactors to use `pretend` less and leverage more service fixtures, while preserving assertions.

Fixes #18356

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -58,6 +58,7 @@ from warehouse.organizations import services as organization_services
 from warehouse.organizations.interfaces import IOrganizationService
 from warehouse.packaging import services as packaging_services
 from warehouse.packaging.interfaces import IProjectService
+from warehouse.rate_limiting import DummyRateLimiter, IRateLimiter
 from warehouse.search import services as search_services
 from warehouse.search.interfaces import ISearchService
 from warehouse.subscriptions import services as subscription_services
@@ -164,6 +165,7 @@ def pyramid_services(
     query_results_cache_service,
     search_service,
     domain_status_service,
+    ratelimit_service,
 ):
     services = _Services()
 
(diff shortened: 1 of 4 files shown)


ewdurbin (commented):
Functionally this all makes sense but I'm curious about the outcome here. If a domain had expired, and I am the person who just renewed or hijacked it will this impact my ability to verify the email address?

Edit: for clarity I mean in the case where a domain had expired with an existing email address on the account is this code path part of that same process too?

miketheman:
Thanks for asking, made me think about this for a bit.

> If a domain had expired, and I am the person who just renewed or hijacked it will this impact my ability to verify the email address?

The current behavior: If the domain has entered any of the grace periods for lack of payment, etc that would trigger a status update, the email address gets unverified, and **cannot** be used as a password reset.

That hasn't changed.

The only part that's changed is that the timing of when a domain status is updated for a previously-checked email record. Instead of waiting out the 30 days between checks, if the user verifies now, then the domain status will get checked now as well, instead of becoming unverified daily for the remainder of the time (up to 30 days in the extreme case).

PR #18387, opened 2025-07-18
style: wrap the yanked reason to prevent overflow

### LINKED ISSUE #18186: Yanked info box breaks layout
**Describe the bug**

https://pypi.org/project/aiohttp/#history

Image

Looks like the box does not wrap correctly.

**Expected behavior**

The box should fit inside the version card, with the mesage correctly wrapped.

**To Reproduce**

Visit a history page with a sufficiently long yanked message in a sufficiently narrow browser window.

**My Platform**

The screenshot is from Firefox, but I see a similar layout on Safari too. This is on macOS 15.5.

**Additional context**

miketheman:
Add a top-level directive to override `.callout-block` attribute and
trigger the ellipsis behavior on wrap.

Fixes #18186

### DIFF
--- a/warehouse/static/sass/blocks/_release.scss
+++ b/warehouse/static/sass/blocks/_release.scss
@@ -99,10 +99,14 @@
     padding-bottom: 2px;
   }
 
-  &__yanked-reason p {
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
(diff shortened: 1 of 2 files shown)


miketheman:
Before:
Screenshot 2025-07-18 at 10 26 27

After:

image

PR #18397, opened 2025-07-21
feat(admin): make better use of screen real estate

miketheman:
- Collapse three columns into one
- Display Observer role when applicable
- Add Frozen status

### DIFF
--- a/warehouse/admin/templates/admin/users/list.html
+++ b/warehouse/admin/templates/admin/users/list.html
@@ -40,11 +40,10 @@
         <th>Username</th>
         <th>Name</th>
         <th>Email</th>
-        <th>Admin</th>
-        <th>Moderator</th>
-        <th>PSF Staff</th>
+        <th>PyPI Roles</th>
         <th>Active</th>
         <th>2FA</th>
+        <th>Frozen</th>
       </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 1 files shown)


warehouse/admin/templates/admin/users/list.html
```diff
@@ -56,11 +55,18 @@
...
         <td>{{ user.name }}</td>
         <td>{{ user.email }}</td>
-        <td>{% if user.is_superuser %}<i class="fa fa-check text-green"></i>{% endif %}</td>
-        <td>{% if user.is_moderator %}<i class="fa fa-check text-green"></i>{% endif %}</td>
-        <td>{% if user.is_psf_staff %}<i class="fa fa-check text-green"></i>{% endif %}</td>
+        <td>
+          {% set pypi_roles = [] %}
+          {% if user.is_superuser %}{% set _ = pypi_roles.append("Admin") %}{% endif %}
+          {% if user.is_moderator %}{% set _ = pypi_roles.append("Moderator") %}{% endif %}
+          {% if user.is_psf_staff %}{% set _ = pypi_roles.append("PSF Staff") %}{% endif %}
```
miketheman, line 62: For reviewers - I opted to use `set _ = ` syntax instead of importing another jinja2 extension to use `do` instead. here's what that would look like:

```jinja2
{% if user.is_psf_staff %}{% do pypi_roles.append("PSF Staff") %}{% endif %}
```

If we have more use cases like `set _ =`, we could entertain adding the extension and using that syntax.

di (approved):

PR #18406, opened 2025-07-23
chore: exclude analytics tracking on some routes

miketheman:
Some behaviors like password resets include a token sent to the client's email.
When that token is used, the loaded webpage emits a tracking payload to
our private analytics service and is stripped prior to storage.

By excluding the tracking script from these pages, we prevent emitting
a tracking event on these pages that could contain the query string with the
token. We have other methods for tracking the traffic on these routes
without transmitting the token value.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -957,14 +957,14 @@ msgstr ""
 
 #: warehouse/templates/404.html:30 warehouse/templates/500.html:18
 #: warehouse/templates/accounts/two-factor.html:35
-#: warehouse/templates/base.html:319 warehouse/templates/base.html:325
 #: warehouse/templates/base.html:331 warehouse/templates/base.html:337
-#: warehouse/templates/base.html:353 warehouse/templates/base.html:359
-#: warehouse/templates/base.html:384 warehouse/templates/base.html:390
-#: warehouse/templates/base.html:399 warehouse/templates/base.html:412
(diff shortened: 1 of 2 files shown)


ewdurbin:
An alternative is: https://plausible.io/docs/custom-locations

I'm not particularly convinced we need this in plausible, but it isn't something we should throw away necessarily.

miketheman:
> An alternative is: [plausible.io/docs/custom-locations](https://plausible.io/docs/custom-locations)
> 
> I'm not particularly convinced we need this in plausible, but it isn't something we should throw away necessarily.

I'd also looked at https://plausible.io/docs/stop-tracking-utm-tags - and both of these approaches involve more surgery to the plausible setup than we have today.

Considering we have alternate ways of tracking these pages activity, I'd prefer to ship this and if you find the need that these pages deserve plausible tracking, we can revisit that later on.

ewdurbin:
I'm confused now, are we saying that the tokens associated with these routes are currently visible in our plausible analytics? I am not able to find them for any of the current routes that are added as an exclusion here.

di:
> I'm confused now, are we saying that the tokens associated with these routes are currently visible in our plausible analytics? I am not able to find them for any of the current routes that are added as an exclusion here.

I assume they're being sent to plausible, it's just ignoring URL parameters?

miketheman:
Closing, implemented via #18407

PR #18407, opened 2025-07-23
Sanitize query parameters on select routes when submitting to analytics

ewdurbin:
Supersedes #18406

### DIFF
--- a/warehouse/templates/base.html
+++ b/warehouse/templates/base.html
@@ -168,9 +168,31 @@
     <script async src="{{ request.static_path('warehouse:static/dist/js/warehouse' + ('' if request.localizer.locale_name == 'en' else '.' + request.localizer.locale_name) + '.js') }}">
     </script>
     {% block extra_js %}{% endblock %}
-    <script defer
-            data-domain="pypi.org"
-            src="https://analytics.python.org/js/script.js"></script>
+    {# Exclude all query parameters from analytics payload on pages where they may contain sensitive information.
+     # PSF controls the analytics destination, so it is not a third-party tracker,
+     # but this prevents the parameters from being sent to the analytics service in the payload,
+     # where it would be further filtered before storage. #}
+    {% set analytics_sanatize_query_params_routes = [
+      "accounts.reset-password",
+      "accounts.verify-email",
+      "accounts.verify-organization-role",
+      "accounts.verify-project-role"
+    ] %}
+    {% if request.matched_route and request.matched_route.name in analytics_sanatize_query_params_routes %}
+      <script defer
+              data-domain="pypi.org"
+              src="https://analytics.python.org/js/script.manual.js"></script>
+      <script>
+          window.plausible = window.plausible || function() { (window.plausible.q = window.plausible.q || []).push(arguments) }
+      </script>
+      <script>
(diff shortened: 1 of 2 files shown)


warehouse/templates/base.html
```diff
@@ -168,9 +168,31 @@
...
     </script>
     {% block extra_js %}{% endblock %}
-    <script defer
-            data-domain="pypi.org"
-            src="https://analytics.python.org/js/script.js"></script>
+    {# Exclude all query parameters from analytics payload on pages where they may contain sensitive information.
+     # PSF controls the analytics destination, so it is not a third-party tracker,
+     # but this prevents the parameters from being sent to the analytics service in the payload,
+     # where it would be further filtered before storage. #}
+    {% set analytics_sanatize_query_params_routes = [
```
miketheman:
```suggestion
    {% set analytics_sanitize_query_params_routes = [
```

warehouse/templates/base.html
```diff
@@ -168,9 +168,31 @@
...
+      "accounts.reset-password",
+      "accounts.verify-email",
+      "accounts.verify-organization-role",
+      "accounts.verify-project-role"
+    ] %}
+    {% if request.matched_route and request.matched_route.name in analytics_sanatize_query_params_routes %}
```
miketheman:
```suggestion
    {% if request.matched_route and request.matched_route.name in analytics_sanitize_query_params_routes %}
```

miketheman (commented):
This looks closer to a good solution, but fails CSP with:

> reset-password/:55 Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https://analytics.python.org *.fastly-insights.com *.ethicalads.io 'sha256-U3hKDidudIaxBDEzwGJApJgPEf2mWk6cfMWghrAa6i0=' https://cdn.jsdelivr.net/npm/mathjax@3.2.2/ 'sha256-1CldwzdEg2k1wTmf7s5RWVd7NMXI/7nxxjJM2C4DqII=' 'sha256-0POaN8stWYQxhzjKS+/eOfbbJ/u4YHO5ZagJvLpMypo=' http://localhost:35729/livereload.js". Either the 'unsafe-inline' keyword, a hash ('sha256-cT8Egmxr1rCx5CcX55CqcQBjJWsITO6i1c2beGK0FUo='), or a nonce ('nonce-...') is required to enable inline execution.


Were you able to get it to work?

ewdurbin:
*sigh* no I did not hit that CSP. I just looked to see that the right thing rendered on curl. I'm not sure what the best mechanism is if we want to add a nonce for each request or if we can extract this to a file somehow. I cannot figure out what's going on with web pack or how to include a specific js file. Do you have any tips?

miketheman:
We have a pattern to generate a separate file for zxcvbn and include the compiled version like this
https://github.com/pypi/warehouse/blob/b1d0f63b849f15c11ad704b14249ae162ca2a523/webpack.config.js#L103

https://github.com/pypi/warehouse/blob/b1d0f63b849f15c11ad704b14249ae162ca2a523/warehouse/templates/accounts/register.html#L224

miketheman (approved):

PR #18414, opened 2025-07-24
Add functionality for editing organizations via admin

ewdurbin:
⚠️ this was predominantly authored by Claude under my prompting, I will be reviewing it as if I had not written it, please do the same. ⚠️ 

- Add admin-specific OrganizationForm for editing organization details
- Update organization_detail view to handle GET/POST requests
- Implement inline editing in organization detail template
- Add proper permission controls (AdminOrganizationsWrite)
- Add comprehensive test coverage for form validation and view behavior
- Support staff and superusers can edit; moderators have read-only access

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -957,14 +957,14 @@ msgstr ""
 
 #: warehouse/templates/404.html:30 warehouse/templates/500.html:18
 #: warehouse/templates/accounts/two-factor.html:35
-#: warehouse/templates/base.html:319 warehouse/templates/base.html:325
-#: warehouse/templates/base.html:331 warehouse/templates/base.html:337
-#: warehouse/templates/base.html:353 warehouse/templates/base.html:359
-#: warehouse/templates/base.html:384 warehouse/templates/base.html:390
-#: warehouse/templates/base.html:399 warehouse/templates/base.html:412
-#: warehouse/templates/base.html:421 warehouse/templates/base.html:427
-#: warehouse/templates/base.html:433 warehouse/templates/base.html:446
-#: warehouse/templates/base.html:463
+#: warehouse/templates/base.html:337 warehouse/templates/base.html:343
+#: warehouse/templates/base.html:349 warehouse/templates/base.html:355
+#: warehouse/templates/base.html:371 warehouse/templates/base.html:377
+#: warehouse/templates/base.html:402 warehouse/templates/base.html:408
+#: warehouse/templates/base.html:417 warehouse/templates/base.html:430
+#: warehouse/templates/base.html:439 warehouse/templates/base.html:445
+#: warehouse/templates/base.html:451 warehouse/templates/base.html:464
+#: warehouse/templates/base.html:481
 #: warehouse/templates/includes/accounts/profile-callout.html:17
 #: warehouse/templates/includes/file-details.html:129
 #: warehouse/templates/index.html:98 warehouse/templates/index.html:105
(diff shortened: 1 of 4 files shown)


warehouse/admin/views/organizations.py
```diff
@@ -146,23 +196,62 @@ def organization_list(request):
 
 @view_config(
     route_name="admin.organization.detail",
-    require_methods=False,
```
miketheman, line 149: thought: curious that claude decided to reorder the parameters. I wonder if we should expect more line churn as a result.

warehouse/admin/views/organizations.py
```diff
@@ -18,9 +20,57 @@
     OrganizationApplicationStatus,
     OrganizationType,
 )
+from warehouse.subscriptions.interfaces import IBillingService
 from warehouse.utils.paginate import paginate_url_factory
 
 
+class OrganizationForm(wtforms.Form):
```
miketheman, line 27: This appears to be functionally equivalent to `warehouse.manage.forms.SaveOrganizationForm` - can we not use that again, and save the duplication? Other than a single difference I can see in the choices for `orgtype`, which is a refactor that could be made to `SaveOrganizationForm`, unless there's some other reasoning?

warehouse/admin/templates/admin/organizations/detail.html
```diff
@@ -86,51 +86,76 @@ <h4 class="modal-title" id="renameModalLabel">
...
-                <a href="{{ organization.link_url }}" rel="nofollow noopener noreferrer" target="_blank"><input class="form-control" value="{{ organization.link_url }}" readonly></a>
+              <div class="form-group">
+                <label for="link_url" class="col-sm-5 control-label">
+                  Organization URL <i class="fa-solid fa-arrow-up-right-from-square"></i>
+                </label>
+                <div class="col-sm-7">
+                  {{ form.link_url(class="form-control", value=organization.link_url, disabled=(not request.has_permission(Permissions.AdminOrganizationsWrite))) }}
+                  {% if form.link_url.errors %}
+                    <div class="text-danger">{{ form.link_url.errors[0] }}</div>
+                  {% endif %}
```
miketheman: With this change we lose the ability to click an a href to visit the URL.

suggestion: move the href to the `label`'s icon right above the form

miketheman (commented):
I testing the admin UI, works as advertised so far. Some comments about improvements inline

warehouse/admin/views/organizations.py
ewdurbin: I think it saw this as a convention in other decorators. seem innocuous

warehouse/admin/views/organizations.py
ewdurbin: our convention is to have forms defined in admin. Claude initially reused the form and when I asked if that was a convention in admin, it correctly determined it was not.

warehouse/admin/templates/admin/organizations/detail.html
ewdurbin: will have the robot fix this.

ewdurbin: f5dfb10ac

ewdurbin: Screenshot 2025-07-25 at 11 39 37 AM

warehouse/admin/templates/admin/organizations/detail.html
```diff
@@ -86,51 +86,83 @@ <h4 class="modal-title" id="renameModalLabel">
...
-              <div class="col-sm-7">
-                <a href="{{ organization.link_url }}" rel="nofollow noopener noreferrer" target="_blank"><input class="form-control" value="{{ organization.link_url }}" readonly></a>
+              <div class="form-group">
+                <label for="link_url" class="col-sm-5 control-label">
+                  Organization URL
+                </label>
+                <div class="col-sm-7">
+                  <div class="input-group">
+                    {{ form.link_url(class="form-control", value=organization.link_url, disabled=(not request.has_permission(Permissions.AdminOrganizationsWrite))) }}
+                    <span class="input-group-btn">
```
miketheman: Correct the UI so it doesn't look weird

```suggestion
                    <span class="input-group-append input-group-btn">
```

miketheman (approved):

PR #18422, opened 2025-07-25
feat: implement PEP 792 project status markers

### LINKED ISSUE #18415: Implement PEP 792 (Project status markers)
PEP 792 was approved a few weeks back: https://peps.python.org/pep-0792/

PyPI is already in a pretty good position to support it, since it already has two status markers (archived and quarantined) and could add others in the future if desired. 

Some TODOs for adding it:

- [x] Update the HTML and JSON simple index renders to include project statuses
- [x] Unit and integration tests

Unless anybody would prefer I hold off, I'll likely take a stab at this tomorrow 🙂 

CC @miketheman for viz

woodruffw:
This adds PEP 792 status markers to the HTML and JSON simple APIs.

The approach I went with here is to add a new `ProjectStatusMarker` enum that maps from `LifecycleStatus` -- I figured that would easier to follow/maintain than trying to shoehorn these statuses more deeply into PyPI's concept of "lifecycle," which is a behavioral superset of what PEP 792 intends to model.

I've also gone ahead and unconditionally rendered the project status in both APIs, instead of omitting it in the `active` case (which the PEP permits) -- this makes the implementation a bit shorter and (IMO) easier to follow. But I'm happy to tweak this as well, of course 🙂 

I'll also include some updates to the API docs as part of this PR.

Closes #18415.


woodruffw:
(NB: I started with unit tests, but I'm happy to add functional/integration tests here too. It might be nice to use this opportunity to add some [inline snapshots](https://pypi.org/project/inline-snapshot/) to the functional tests.)

woodruffw:
Gentle ping for review 🙂 

(+CC @miketheman and @di in particular)

PR #18427, opened 2025-07-28
Proxy protection

ewdurbin:
This adds some JavaScript, which attempts to detect if the site that has loaded the page matches a set of allowed domains configured in the backend. If the domain detected does not match that list, a security model overlays the entire screen, which can be dismissed leaving a persistent banner.

Some light cryptography is implemented, which is not difficult to bypass, but does mean that an active proxy must be used, which can perform the hashing necessary and rewrite the correct strings. It's not bulletproof, but it will require a slightly more advanced infrastructure to perform similar proxy attacks.

We can also alter the hashing scheme relatively quickly in order to break proxies that make assumptions if another attack occurs or if they modified their proxy during an attack.

Screenshot 2025-07-28 at 10 37 33 AM
Screenshot 2025-07-28 at 10 38 29 AM

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -333,6 +333,7 @@ def test_configure(monkeypatch, settings, environment):
         "gcloud.service_account_info": {},
         "warehouse.forklift.legacy.MAX_FILESIZE_MIB": 100,
         "warehouse.forklift.legacy.MAX_PROJECT_SIZE_GIB": 10,
+        "warehouse.allowed_domains": [],
     }
     if environment == config.Environment.development:
         expected_settings.update(
@@ -394,6 +395,7 @@ def test_configure(monkeypatch, settings, environment):
         ]
         + [
             pretend.call(".logging"),
+            pretend.call(".request"),
             pretend.call("pyramid_jinja2"),
             pretend.call(".filters"),
             pretend.call("pyramid_mailer"),
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -670,3 +670,34 @@ def test_root_factory_access_control_list():
             ),
         ),
     ]
+
+
+class TestWarehouseAllowedDomains:
+    def test_allowed_domains_parsing(self):
+        """Test that allowed domains are parsed correctly."""
+
+        # Test the lambda function used in maybe_set
+        def parser(s):
+            return [d.strip() for d in s.split(",") if d.strip()]
+
+        # Test normal case
+        assert parser("pypi.org, test.pypi.org, example.com") == [
+            "pypi.org",
+            "test.pypi.org",
+            "example.com",
+        ]
+
+        # Test with empty strings
+        assert parser("pypi.org,,, test.pypi.org, ") == ["pypi.org", "test.pypi.org"]
+
+        # Test with only commas
+        assert parser(",,,") == []
+
+        # Test single domain
+        assert parser("pypi.org") == ["pypi.org"]
+
+        # Test with extra spaces
+        assert parser("  pypi.org  ,   test.pypi.org  ") == [
+            "pypi.org",
+            "test.pypi.org",
+        ]
--- a/tests/unit/test_request.py
+++ b/tests/unit/test_request.py
@@ -1,0 +1,96 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import pretend
+
+from warehouse import request
+
+
+class TestCreateNonce:
+    def test_generates_unique_nonces(self):
+        """Test that each request gets a unique nonce."""
+        req1 = pretend.stub()
+        req2 = pretend.stub()
+
+        nonce1 = request._create_nonce(req1)
+        nonce2 = request._create_nonce(req2)
+
+        # Nonces should be strings
+        assert isinstance(nonce1, str)
+        assert isinstance(nonce2, str)
+
+        # Nonces should have reasonable length (base64url encoded 32 bytes)
+        assert len(nonce1) >= 32
+        assert len(nonce2) >= 32
+
+        # Nonces should be unique
+        assert nonce1 != nonce2
+
+    def test_nonce_is_url_safe(self):
+        """Test that nonces are URL-safe."""
+        req = pretend.stub()
+        nonce = request._create_nonce(req)
+
+        # Should only contain URL-safe characters
+        # base64url uses A-Z, a-z, 0-9, -, _
+        import re
+
+        assert re.match(r"^[A-Za-z0-9_-]+$", nonce)
+
+
+class TestCreateHashedDomains:
+    def test_hashes_domains_with_nonce(self):
+        """Test that domains are hashed using the nonce."""
+        req = pretend.stub(
+            nonce="test-nonce-123",
+            registry=pretend.stub(
+                settings={"warehouse.allowed_domains": ["pypi.org", "test.pypi.org"]}
+            ),
+        )
+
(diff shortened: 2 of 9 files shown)


tests/unit/test_config.py
```diff
@@ -670,3 +670,34 @@ def test_root_factory_access_control_list():
...
     ]
+
+
+class TestWarehouseAllowedDomains:
+    def test_allowed_domains_parsing(self):
+        """Test that allowed domains are parsed correctly."""
+
+        # Test the lambda function used in maybe_set
+        def parser(s):
+            return [d.strip() for d in s.split(",") if d.strip()]
```
miketheman, line 683: Curious why the secondary implementation is done here, since if this code changes and the code in config doesn't, we have a false sense of security.

Would it make more sense to extract to a `coercer` function and pass that to `maybe_set` instead, thus keeping them in sync?

warehouse/request.py
```diff
@@ -0,0 +1,37 @@
...
+    """Generate a cryptographically secure random nonce for the request."""
+    return secrets.token_urlsafe(32)
+
+
+def _create_hashed_domains(request):
+    """Create a list of hashed allowed domains using the request nonce as salt."""
+    allowed_domains = request.registry.settings.get("warehouse.allowed_domains", [])
+
+    if not allowed_domains:
+        return ""
```
miketheman, line 18: If I'm reading this correctly, if no domains exist in the config, no detection is applied - is that desirable? Or, should the default then fall back to `warehouse.domain` value so that there's always something in this path of protection?

warehouse/request.py
```diff
@@ -0,0 +1,37 @@
...
+        domain_bytes = domain.encode("utf-8")
+        hashed = hmac.new(nonce_bytes, domain_bytes, hashlib.sha256).hexdigest()
+        hashed_domains.append(hashed)
+
+    return ",".join(hashed_domains)
+
+
+def includeme(config):
+    # Add a nonce to every request
+    config.add_request_method(_create_nonce, name="nonce", reify=True)
```
miketheman, line 34: I estimate this to add some microseconds to each request, which should not be a problem, but something to be aware of.

warehouse/request.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Add hashed domains to every request
+    config.add_request_method(_create_hashed_domains, name="hashed_domains", reify=True)
```
miketheman, line 37: I have not measured the impact of this call yet.

tests/unit/test_request.py
```diff
@@ -0,0 +1,96 @@
...
+        assert nonce1 != nonce2
+
+    def test_nonce_is_url_safe(self):
+        """Test that nonces are URL-safe."""
+        req = pretend.stub()
+        nonce = request._create_nonce(req)
+
+        # Should only contain URL-safe characters
+        # base64url uses A-Z, a-z, 0-9, -, _
+        import re
```
miketheman, line 35: No linter about the import not being at the top of the test file? Curious about that...

warehouse/static/js/warehouse/utils/proxy-protection.js
```diff
@@ -0,0 +1,221 @@
...
+      </span>
+    `;
+    document.body.insertBefore(banner, document.body.firstChild);
+  }
+}
+
+// Initialize proxy protection when DOM is ready
+docReady(checkProxyProtection);
+
+export default checkProxyProtection;
```
miketheman, line 200: Is this being exported for use in any other code?

The pattern often used is to import the module in `index.js`, and the perform the `docReady(checkProxyProtection)` in that file instead. Removes the need to import `docReady` here.

warehouse/static/js/warehouse/utils/proxy-protection.js
```diff
@@ -0,0 +1,221 @@
...
+  const hashArray = Array.from(new Uint8Array(signature));
+  return hashArray.map(b => b.toString(16).padStart(2, "0")).join("");
+}
+
+async function checkProxyProtection() {
+  // Get the allowed domains and nonce from data attributes
+  const proxyProtectionElement = document.querySelector("[data-allowed-domains][data-request-nonce]");
+  if (!proxyProtectionElement) {
+    return;
+  }
```
miketheman: From this it appears that if the element is not on the page, none of the detection logic will happen. If I strip out `data-request-nonce` from the HTML, then the check passes. It would be better to not fail silently.

warehouse/static/js/warehouse/utils/proxy-protection.js
```diff
@@ -0,0 +1,205 @@
...
+  
+  // Try to get the allowed domains and nonce from data attributes
+  const proxyProtectionElement = document.querySelector("[data-allowed-domains][data-request-nonce]");
+  
+  if (proxyProtectionElement) {
+    const hashedDomainsStr = proxyProtectionElement.dataset.allowedDomains;
+    const nonce = proxyProtectionElement.dataset.requestNonce;
+    
+    if (hashedDomainsStr && nonce) {
+      // Parse the hashed domains
```
miketheman, line 43: Similar note about if the elements are missing

warehouse/static/js/warehouse/utils/proxy-protection.js
```diff
@@ -0,0 +1,205 @@
...
+      let hashedDomains = [];
+      try {
+        hashedDomains = hashedDomainsStr.split(",").filter(h => h.length > 0);
+      } catch (e) {
+        console.error("Failed to parse hashed domains:", e);
+      }
+
+      if (hashedDomains.length > 0) {
```
miketheman, line 51: another case of a problem failing open

tests/unit/test_config.py
ewdurbin: Computer made that decision 🤷

warehouse/request.py
ewdurbin: https://github.com/pypi/warehouse/pull/18427/commits/bd9ce90aed39f41afdb8a16de21b085939324c7d

warehouse/static/js/warehouse/utils/proxy-protection.js
ewdurbin: https://github.com/pypi/warehouse/pull/18427/commits/bd9ce90aed39f41afdb8a16de21b085939324c7d

warehouse/static/js/warehouse/utils/proxy-protection.js
ewdurbin: https://github.com/pypi/warehouse/pull/18427/commits/bd9ce90aed39f41afdb8a16de21b085939324c7d

warehouse/static/js/warehouse/utils/proxy-protection.js
ewdurbin: https://github.com/pypi/warehouse/pull/18427/commits/bd9ce90aed39f41afdb8a16de21b085939324c7d

warehouse/static/js/warehouse/utils/proxy-protection.js
ewdurbin: I think I know what you mean OK: 126b3708a20be073fffe4191c67d5d842939ee0d

tests/unit/test_config.py
ewdurbin: Yeah, testing itself doesn't make much sense. I'll work on repairing this.

ewdurbin:
Will follow up with outstanding review comments

PR #18440, opened 2025-07-29
fix: do not pyupgrade or flake8 on venv

JacobCoffee:
~~Whenever I run `make lint` or `make reformat` it churns through `venv/` on a fresh install inside the base container and that makes it not fast.~~

~~this ignores that path for flake8 and pyupgrade~~

i am bad


ewdurbin:
I think you want to do this: https://github.com/pypi/warehouse/pull/14739

PR #18443, opened 2025-07-29
fix(search): limit description for search

miketheman:
If a release description is larger than the transport size, the indexing will fail, as it cannot pass more than a single 10MB request.

Fixes WAREHOUSE-PRODUCTION-29W


webknjaz:
xref https://github.com/pypi/support/issues/7117

PR #18483, opened 2025-08-05
Add organization membership and role management in admin interface

ewdurbin:
- Add view functions for adding, updating, and deleting organization roles
- Add routes for the new role management endpoints
- Update organization detail template to show interactive role management UI
- Add dropdown for role changes with explicit save button
- Add confirmation modal for removing members
- Add comprehensive tests for all new functionality
- Consolidate all organization-related code into existing files

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tests/unit/admin/views/test_organizations.py
+++ b/tests/unit/admin/views/test_organizations.py
@@ -10,6 +10,8 @@ from warehouse.admin.views import organizations as views
 from warehouse.organizations.interfaces import IOrganizationService
 from warehouse.organizations.models import (
     OrganizationApplicationStatus,
+    OrganizationRole,
+    OrganizationRoleType,
     OrganizationType,
 )
 from warehouse.subscriptions.interfaces import IBillingService
@@ -18,6 +20,7 @@ from ....common.db.accounts import UserFactory
 from ....common.db.organizations import (
     OrganizationApplicationFactory,
     OrganizationFactory,
+    OrganizationRoleFactory,
     OrganizationStripeCustomerFactory,
 )
 from ....common.db.subscriptions import StripeCustomerFactory
@@ -267,12 +270,11 @@ class TestOrganizationList:
 
 class TestOrganizationDetail:
     @pytest.mark.usefixtures("_enable_organizations")
-    def test_detail(self):
-        organization = pretend.stub(
-            id=pretend.stub(),
+    def test_detail(self, db_request):
+        organization = OrganizationFactory.create(
             name="example",
             display_name="Example",
-            orgtype=pretend.stub(name="Company"),
+            orgtype=OrganizationType.Company,
             link_url="https://www.example.com/",
             description=(
                 "This company is for use in illustrative examples in documents "
@@ -280,33 +282,30 @@ class TestOrganizationDetail:
                 "coordination or asking for permission."
             ),
             is_active=False,
-            status=OrganizationApplicationStatus.Submitted,
         )
         organization_service = pretend.stub(
             get_organization=lambda *a, **kw: organization,
         )
         billing_service = pretend.stub()
-        request = pretend.stub(
-            flags=pretend.stub(enabled=lambda *a: False),
-            find_service=lambda iface, **kw: {
-                IOrganizationService: organization_service,
-                IBillingService: billing_service,
-            }[iface],
-            matchdict={"organization_id": pretend.stub()},
-            method="GET",
-        )
 
-        result = views.organization_detail(request)
+        db_request.find_service = lambda iface, **kw: {
+            IOrganizationService: organization_service,
+            IBillingService: billing_service,
+        }[iface]
+        db_request.matchdict = {"organization_id": str(organization.id)}
+        db_request.method = "GET"
+
+        result = views.organization_detail(db_request)
         assert result["organization"] == organization
         assert isinstance(result["form"], views.OrganizationForm)
+        assert result["roles"] == []
 
     @pytest.mark.usefixtures("_enable_organizations")
-    def test_detail_is_approved_true(self):
-        organization = pretend.stub(
-            id=pretend.stub(),
+    def test_detail_is_approved_true(self, db_request):
+        organization = OrganizationFactory.create(
             name="example",
             display_name="Example",
-            orgtype=pretend.stub(name="Company"),
+            orgtype=OrganizationType.Company,
             link_url="https://www.example.com/",
             description=(
                 "This company is for use in illustrative examples in documents "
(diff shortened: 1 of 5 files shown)


warehouse/admin/views/organizations.py
```diff
@@ -249,9 +253,18 @@ def organization_detail(request):
...
+        .join(User)
+        .filter(OrganizationRole.organization == organization)
+        .order_by(User.username)
+        .all()
+    )
+
     return {
         "organization": organization,
         "form": form,
+        "roles": roles,
```
miketheman: Interesting! The bot added another query, which I don't think is necessary? Iterating through `organization.roles` relationship appears sufficient:

```suggestion
    return {
        "organization": organization,
        "form": form,
        "roles": organization.roles,
```

If the main utility of the function is to have the results sorted by `User.username`, that can be implemented more efficiently on the `OrganizationRole` relationship (in `warehouse/organizations/models.py`):

```python
    user: Mapped[User] = relationship(back_populates="organization_roles", lazy=False, order_by=User.username)
```

With this change, the view now passes "the same" data twice to the template - once as part of the `organization` object, and once as part of `roles` - probably not an issue yet, as the volume of data for org responses is probably small, and this is admin-only. I'm fine to preserve the dual query to get this out the door.

warehouse/admin/templates/admin/organizations/detail.html
```diff
@@ -201,19 +201,71 @@ <h3 class="card-title">Members</h3>
...
-                <td>{{ role.role_name.value }}</td>
+                <td>
+                  <form method="POST" action="{{ request.route_path('admin.organization.update_role', organization_id=organization.id, role_id=role.id) }}" style="display: inline;">
+                    <input name="csrf_token" type="hidden" value="{{ request.session.get_csrf_token() }}">
+                    <div class="input-group input-group-sm">
+                      <select class="form-control" name="role_name" {{ 'disabled' if not request.has_permission(Permissions.AdminRoleAdd) }}>
+                        <option value="Member" {% if role.role_name.value == "Member" %}selected{% endif %}>Member</option>
+                        <option value="Manager" {% if role.role_name.value == "Manager" %}selected{% endif %}>Manager</option>
+                        <option value="Billing Manager" {% if role.role_name.value == "Billing Manager" %}selected{% endif %}>Billing Manager</option>
+                        <option value="Owner" {% if role.role_name.value == "Owner" %}selected{% endif %}>Owner</option>
```
miketheman: thought: I'm kinda surprised we don't use [`OrganizationRoleType`](https://github.com/pypi/warehouse/blob/f59df186dc62274b5831a72f639a1e92bbe3f94c/warehouse/organizations/models.py#L49-L53) in any HTML templates yet - seems like a perfect use case? These options are following an established pattern, so it's fine, but it'd be good to refactor that at some point if we ever envision adding another role. No action needed right now.

tests/unit/admin/views/test_organizations.py
```diff
@@ -267,106 +270,97 @@ def test_is_invalid_query(self, db_request):
...
-            orgtype=pretend.stub(name="Company"),
+            orgtype=OrganizationType.Company,
             link_url="https://www.example.com/",
             description=(
                 "This company is for use in illustrative examples in documents "
                 "You may use this company in literature without prior "
                 "coordination or asking for permission."
             ),
             is_active=False,
-            status=OrganizationApplicationStatus.Submitted,
```
miketheman, line 283: oh, the dangers of stubs! nice cleanup.

tests/unit/admin/views/test_organizations.py
```diff
@@ -267,106 +270,97 @@ def test_is_invalid_query(self, db_request):
...
-            }[iface],
-            matchdict={"organization_id": pretend.stub()},
-            method="GET",
-        )
 
-        result = views.organization_detail(request)
+        db_request.find_service = lambda iface, **kw: {
+            IOrganizationService: organization_service,
+            IBillingService: billing_service,
+        }[iface]
```
miketheman: `db_request` fixture should have [realish service fixtures already for both orgs and billing](https://github.com/pypi/warehouse/blob/f1820e0614784d24468c6be0f44c603d431775a0/tests/conftest.py#L495-L510) - if you wanted to further reduce the amount of stubs for these tests, I'd support that! I think that would mean removing this assignment and the stub services. I understand if that's more surgery than you want for this though.

tests/unit/admin/views/test_organizations.py
```diff
@@ -541,6 +535,74 @@ def test_does_not_update_with_invalid_form(self, db_request):
...
 
+    @pytest.mark.usefixtures("_enable_organizations")
+    def test_detail_with_roles(self, db_request):
+        """Test that organization detail view includes roles"""
+        organization = OrganizationFactory.create(name="pypi")
+
+        # Create some users with roles
+        user1 = UserFactory.create(username="alice")
+        user2 = UserFactory.create(username="bob")
+        user3 = UserFactory.create(username="charlie")
```
miketheman: thought: A true sort test would be to create these in non-alpha order, so there's more guarantee of non-orders, since often the default sort order is insertion order, lacking any other modifiers.

```suggestion
        # Intentionally not ordered to test order later
        user3 = UserFactory.create(username="charlie")
        user2 = UserFactory.create(username="bob")
        user1 = UserFactory.create(username="alice")
```

miketheman (approved):
Looks like a generally good 80% accurate for a bot, not too shabby.

UI looks nice, too.

Was also generally surprised to not see `require_csrf=True` on any of the views, but maybe this doesn't need it? Other views do 🤔 .

I like that there's an Org Event emitted, maybe a future PR will surface those?

Approving with some improvements and suggestions inline - feel free to tag me for a re-review if you like.

warehouse/admin/templates/admin/organizations/detail.html
ewdurbin: c6062c6a0ce6f6c2795d06b3faa3df5da9558f8c, a489188f6450049770541dfcccd49e55bb9d1441

warehouse/admin/views/organizations.py
ewdurbin: middle ground for now: 79c75396feb

ewdurbin:
> Was also generally surprised to not see `require_csrf=True` on any of the views, but maybe this doesn't need it? Other views do 🤔 .

Yeah, it's funny that it doesn't choose to copy pasta everything 🙃 99f29700813680519bee6a0a9362eb60d4fc1481

tests/unit/admin/views/test_organizations.py
ewdurbin: https://github.com/pypi/warehouse/pull/18483/commits/57cd23652fef08da1afbe6dbda2d03cd7204ec1d

ewdurbin: this was an excellent task for da robot!

PR #18500, opened 2025-08-08
feat(admin): lookup emails sent to address

miketheman:
Styles buttons a bit, and moves unverify reason to new row.

Refs: https://github.com/pypi/warehouse/pull/18034#pullrequestreview-2795461946


miketheman:
Before:

image


After:

image

PR #11884, opened 2022-07-19
docs: Fix a few typos

timgates42:
There are small typos in:
- tests/frontend/viewport_toggle_controller_test.js
- warehouse/email/ses/models.py
- warehouse/legacy/api/json.py
- warehouse/rate_limiting/__init__.py

Fixes:
- Should read `vulnerabilities` rather than `vulnerabilties`.
- Should read `unverified` rather than `unverfied`.
- Should read `starting` rather than `satarting`.
- Should read `skip` rather than `skipp`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md

PR #18553, opened 2025-08-15
lint: fix typos and include lint checking

miketheman:
Supports both `reformat` and `lint`.

Ref: https://pypi.org/project/codespell/

Closes #11568
Closes #11884

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,21 @@
+[tool.codespell]
+ignore-multiline-regex = "codespell:ignore-begin.*codespell:ignore-end"
+ignore-words-list = "deriver"
+skip = [
+  "*.po",
+  "./coverage/*",
+  "./dev/*",
+  "./docs/dev/_build/*",
+  "./docs/blog-site/*",
+  "./docs/dev-site/*",
+  "./docs/user-site/*",
+  "./htmlcov/*",
+  "./node_modules/**",
+  "./package-lock.json",
+  "./warehouse/admin/static/dist/*",
+  "./warehouse/static/dist/*",
+]
+
 [tool.coverage.run]
 branch = true
 dynamic_context = "test_function"
--- a/tests/unit/forklift/test_metadata.py
+++ b/tests/unit/forklift/test_metadata.py
@@ -178,7 +178,7 @@ class TestValidation:
     def test_deprecated_classifiers_with_replacement(self, backfill):
         data = (
             b"Metadata-Version: 2.1\nName: spam\nVersion: 2.0\n"
-            b"Classifier: Natural Language :: Ukranian\n"
+            b"Classifier: Natural Language :: Ukranian\n"  # codespell:ignore
         )
 
         if not backfill:
@@ -187,7 +187,9 @@ class TestValidation:
             _assert_invalid_metadata(excinfo.value, "classifier")
         else:
             meta = metadata.parse(data, backfill=True)
-            assert meta.classifiers == ["Natural Language :: Ukranian"]
+            assert meta.classifiers == [
+                "Natural Language :: Ukranian"  # codespell:ignore
+            ]
 
     @pytest.mark.parametrize("backfill", [True, False])
     def test_deprecated_classifiers_no_replacement(self, backfill):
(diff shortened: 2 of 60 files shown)


warehouse/integrations/vulnerabilities/osv/views.py
```diff
@@ -43,7 +43,7 @@ def report_vulnerabilities(request):
         vulnerability_reports = request.json_body
     except json.decoder.JSONDecodeError:
         metrics.increment(
-            "warehouse.vulnerabilties.error.payload.json_error", tags=["origin:osv"]
+            "warehouse.vulnerabilities.error.payload.json_error", tags=["origin:osv"]
```
ewdurbin, line 46: is there a way for us to merge the metrics if this ships?

tests/unit/oidc/forms/test_activestate.py
```diff
@@ -128,7 +128,7 @@ def test_lookup_actor_404(self, monkeypatch):
         )
         requests = pretend.stub(
             post=pretend.call_recorder(lambda o, **kw: response),
-            expception=_requests.exceptions,
```
ewdurbin, line 131: seems odd to me that this doesn't change the behavior of this test

ewdurbin:
I understand the intent here, I want to understand if this will auto correct spelling, or if it will be a failure case in check that needs to be manually resolved.

warehouse/integrations/vulnerabilities/osv/views.py
miketheman: Sadly there's no way to merge metrics yet.

I investigated and this specific metric hasn't reported at least once in our metrics retention history, nor does it appear on any dashboards.

tests/unit/oidc/forms/test_activestate.py
miketheman: Thanks for noting. Looks like the assertions never care about this specific property, so I can remove these from being set, if they have no impact. I'll remove them completely in a following commit.

bin/reformat
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -ex
 
+# Fix typos before other changes, as they may affect formatting.
+codespell --write
```
ewdurbin: I am -1 on this. seems it could be a footgun if an undesirable change to spelling is automatically written.

miketheman: Happy to remove, whenever there is an ambiguity of spelling, codespell won't change it, rather continue to error out during a lint phase.

Happy to remove if you think this is a serious footgun, but otherwise it felt appropriate to help the user similar to other reformat directives.

miketheman: bc4fe53

tests/unit/oidc/forms/test_activestate.py
miketheman: f9821a3

tests/unit/forklift/test_metadata.py
```diff
@@ -178,7 +178,7 @@ def test_invalid_classifier(self):
     def test_deprecated_classifiers_with_replacement(self, backfill):
         data = (
             b"Metadata-Version: 2.1\nName: spam\nVersion: 2.0\n"
-            b"Classifier: Natural Language :: Ukranian\n"
+            b"Classifier: Natural Language :: Ukranian\n"  # codespell:ignore
```
JacobCoffee, line 181: is not `Ukrainian`?

pyproject.toml
```diff
@@ -1,3 +1,21 @@
...
+ignore-multiline-regex = "codespell:ignore-begin.*codespell:ignore-end"
+ignore-words-list = "deriver"
+skip = [
+  "*.po",
+  "./coverage/*",
+  "./dev/*",
+  "./docs/dev/_build/*",
+  "./docs/blog-site/*",
+  "./docs/dev-site/*",
+  "./docs/user-site/*",
```
JacobCoffee, line 11: are there just too. many isseus for us to scan right now or? seems like these would be a good place to make sure we spell check

JacobCoffee (approved):

tests/unit/forklift/test_metadata.py
miketheman: It is, and that's literally what the test is testing for - the misspelling is the deprecated classifier

JacobCoffee: oh i see

pyproject.toml
miketheman: No, these are the output / built renderings, so the check is duplicative

Issue #1244, opened 2016-06-07
JSON API for Trove Classifiers


dstufft:
Similar to #1241 but provide a JSON API, this will depend on figuring out a solution to #284.

PR #18574, opened 2025-08-20
feat: list_classifiers endpoint returns all classifiers

woodruffw:
This is a funky case: the `list_classifiers` API is from legacy PyPI. It used the `sorted_classifiers` API from `trove_classifiers`, but seemingly did so primarily because the concept of a "deprecated" classifier didn't exist at the time.

To accomodate deprecated classifiers, `trove_classifiers` added an `all_classifiers` API. However PyPI never switched to this API _here_, although it _did_ switch to it in the `warehouse classifiers sync` CLI:

https://github.com/pypi/warehouse/blob/bbcabd28a904dc8bc1508be48da4d69a3811a5fa/warehouse/cli/classifiers.py#L5

As a result, there's now a slight discrepancy between what PyPI accepts (all classifiers, including deprecated ones) and what PyPI advertises in this API (only non-deprecated classifiers). 

On a more practical note: I have no idea if this is worth changing/fixing 😅. It solves a very specific niche problem I have, but I think a very reasonable reaction to this PR would be "absolutely not."

CC @dstufft

### DIFF
--- a/tests/unit/legacy/api/test_pypi.py
+++ b/tests/unit/legacy/api/test_pypi.py
@@ -4,7 +4,7 @@ import pretend
 import pytest
 
 from pyramid.httpexceptions import HTTPBadRequest, HTTPMovedPermanently, HTTPNotFound
-from trove_classifiers import sorted_classifiers
+from trove_classifiers import all_classifiers
 
 from warehouse.legacy.api import pypi
 
@@ -61,7 +61,7 @@ def test_list_classifiers(db_request):
     resp = pypi.list_classifiers(db_request)
 
     assert resp.status_code == 200
-    assert resp.text == "\n".join(sorted_classifiers)
+    assert resp.text == "\n".join(all_classifiers)
 
 
 def test_search():
--- a/warehouse/legacy/api/pypi.py
+++ b/warehouse/legacy/api/pypi.py
@@ -2,7 +2,7 @@
 
 from pyramid.httpexceptions import HTTPGone, HTTPMovedPermanently, HTTPNotFound, HTTPOk
 from pyramid.view import forbidden_view_config, view_config
-from trove_classifiers import sorted_classifiers
+from trove_classifiers import all_classifiers
 
 from warehouse.classifiers.models import Classifier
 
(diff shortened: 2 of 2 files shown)


di:
I'm not 100% sure on why people are using this endpoint (is it to find classifiers to then search for? or valid classifiers to publish with?) but I think this might become confusing if we're not able to say that a given classifier is deprecated in the context of the response here.

Given https://github.com/pypi/warehouse/issues/1244, I think instead of making changes to the way this API works, we should probably just take this as an opportunity to turn it down instead.

It might be worth noting that the `/classifiers` view also does not include deprecated classifiers (namely, because we don't want people to use them): https://github.com/pypi/warehouse/blob/bbcabd28a904dc8bc1508be48da4d69a3811a5fa/warehouse/views.py#L347-L353

woodruffw:
> I'm not 100% sure on why people are using this endpoint (is it to find classifiers to then search for? or valid classifiers to publish with?) but I think this might become confusing if we're not able to say that a given classifier is deprecated in the context of the response here.

I'm not sure about other users, but in my case it was to get the exact "live" set of classifiers accepted by PyPI (deprecated or not). But this is a direct proxy for `trove_classifiers` anyways, so it arguably doesn't really accomplish anything on my part 😅 

> Given #1244, I think instead of making changes to the way this API works, we should probably just take this as an opportunity to turn it down instead.

Works for me! I'm curious if that endpoint sees any real traffic anyways, given that it's been undocumented for a long time.

di:
> accepted by PyPI

For what, upload or search?

woodruffw:
> > accepted by PyPI
> 
> For what, upload or search?

Upload only in my case.

di:
Since we don't accept deprecated classifiers for upload, what's the motivation for that? 

https://github.com/pypi/warehouse/blob/bbcabd28a904dc8bc1508be48da4d69a3811a5fa/warehouse/forklift/metadata.py#L174-L189

woodruffw:
> Since we don't accept deprecated classifiers for upload, what's the motivation for that?

...oh 😅. For some reason I thought deprecated classifiers were still accepted on upload, and I was working from my incorrect memory. 

This is all noise, then. Sorry!

PR #18583, opened 2025-08-22
fix: add non-breaking space to error codes in 403, 404, 410, and search result templates

### LINKED ISSUE #18552: Missing space between word "projects" and "for" in PyPI search results
**Describe the bug**
The search results page on pypi.org is missing a space in the header that describes the number of matches found. For example, it says:
*5,799 projectsfor "fastapi"*
It should say:
*5,799 projects for "fastapi"*

**To Reproduce**
Go to pypi.org and search for a term that has results.

**Additional context**
The bug can be easily corrected by adding a space before the word "for" at [line 147 (currently) in /warehouse/templates/search/results.html](https://github.com/pypi/warehouse/blob/52b9fd0a9296cf6ec1c7a7356ab75bc31f6be243/warehouse/templates/search/results.html#L147).

**Example 1: Bug**
Image

**Example 2: Where To Correct**
Image

hasansezertasan:
Still a draft; feedback very welcome 🙂

Closes #18552

### DIFF
--- a/warehouse/templates/403.html
+++ b/warehouse/templates/403.html
@@ -6,4 +6,4 @@
 {% block error_heading %}
   {% trans %}You don't have permission to view this page{% endtrans %}
 {% endblock %}
-{% block error_code %}403{% endblock %}
+{% block error_code %}&nbsp;403{% endblock %}
--- a/warehouse/templates/404.html
+++ b/warehouse/templates/404.html
@@ -6,7 +6,7 @@
 {% block error_heading %}
   {% trans %}We looked everywhere but couldn't find this page{% endtrans %}
 {% endblock %}
-{% block error_code %}404{% endblock %}
+{% block error_code %}&nbsp;404{% endblock %}
 {% block extra_css %}<link href="https://fonts.googleapis.com/css?family=Ewert" rel="stylesheet">{% endblock %}
 {% block easter_egg %}
   <div class="viewport-section viewport-section--ee">
--- a/warehouse/templates/410.html
+++ b/warehouse/templates/410.html
@@ -6,4 +6,4 @@
 {% block error_heading %}
   {% trans %}The page you are looking for has been removed{% endtrans %}
 {% endblock %}
-{% block error_code %}410{% endblock %}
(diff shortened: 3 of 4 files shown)


warehouse/templates/403.html
```diff
@@ -6,4 +6,4 @@
 {% block error_heading %}
   {% trans %}You don't have permission to view this page{% endtrans %}
 {% endblock %}
-{% block error_code %}403{% endblock %}
+{% block error_code %}&nbsp;403{% endblock %}
```
di: Probably makes more sense to add this once in the place where the block is used here: https://github.com/pypi/warehouse/blob/bbcabd28a904dc8bc1508be48da4d69a3811a5fa/warehouse/templates/error-base.html#L23-L25

hasansezertasan: Thank you @di for your quick review 🙏.

> Probably makes more sense to add this once in the place where the block is used here:

Definetly 👍. I already tried that and it didn't give a proper result. I tried several variations:

```html
<strong>{% trans %}Error code{% endtrans %}
  &nbsp;{% block error_code %}{% endblock %}
</strong>
```

```html
<strong>{% trans %}Error code{% endtrans %}&nbsp;
  {% block error_code %}{% endblock %}
</strong>
```

```html
<strong>{% trans %}Error code{% endtrans %}&nbsp;{% block error_code %}{% endblock %}
</strong>
```

The first two doesn't work as expected but the later works well. But then the `make reformat` (djlint) would turn it into:

```html
<strong>{% trans %}Error code{% endtrans %}&nbsp;
  {% block error_code %}{% endblock %}
</strong>
```

---

I think we need to decide on:

- Use the current implementation.
- Suppressing this behavior using djlint for that block.

I think the best way to go is this:

```html
{# djlint:off #}
<strong>{% trans %}Error code{% endtrans %}&nbsp;{% block error_code %}{% endblock %}</strong>
{# djlint:on #}
```

miketheman (approved):
Thanks, looks like it works!

hasansezertasan:
> Thanks, looks like it works!

Happy to hear this 😇.

Issue #8456, opened 2020-08-19
2FA bruteforce check


ewjoachim:
**What's the problem this feature will solve?**
Currently, there's a rate limiter on 2FA to avoid abuse, but an attacker patient enough can hope to correctly guess the code (>50% chance) in 3 months. There's no limit on the maximum number of failed tries before a protection measure is added. Having a 6 digits code can only be a thing if you can't bruteforce it, otherwise it's just a very weak code.

**Describe the solution you'd like**
A sufficient (to determine) number of failed 2FA tries should raise a protective measure (to determine).

**Additional context**
I reported the problem via the official Security mailing list back in March 2020, but (AFAICT) the security team didn't have the time to handle it in a private patch. I recently got the confirmation from @di that it was OK to post it publicly, and develop the patch publicly. I might be interested in helping and/or coding but I'm currently involved in #7124 and I'd much rather prefer to finish the big works before starting on another one. So feel free to grab the issue and improve PyPI's security :)

Here's the original report as sent to security@ in March:

> This one may be harder than usual to reproduce but, here I go :)
> Again, not sure this qualifies as an important security problem, but you're the judge.
>
> Looking at the TOTP code in warehouse, I noticed it uses an anti-bruteforce policy which is a moving window of (unless I'm wrong) 10 attempts per 5 minutes. Each attempt counts for 3 because the TOTP validates with 3 30-second windows: the current one, the previous one and the next one, and we can fit 10*3 independent 30-second windows in 5 minutes. I'll be using 9 attempts below rather than 10 so that the victim won't notice if they try to login too.
>
> As it stands, if I have someone's credentials, I have more than 50% chance of success bypassing TOTP without raising any alarm by quietly having a script run 9 tries every 5 minutes for 3 months. In 1 year time, I have >95% chances of success. [1]
>
> So, I know 3 months is not, like, 3 minutes, but we're clearly not looking at a case of "3000 years" nor "3 months with access to a supercomputer", so that might be pulled off easily (assuming user doesn't change their password in 3 months, but it's not very long in terms of how often we should do it.). Also, it's just 50% chance, flip of a coin. Also, it assumes that the attacker has working credentials in the first place (but that's the whole point of 2FA). And also, I wasn't able to try this one out, obviously, I didn't sit on it for 3 months to check if it was correct :) [Note: I haven't tried since]
>
[...]

PR #18584, opened 2025-08-22
feat: add rate limiting to 2fa attempts

miketheman:
Limits specifically calls to the 2FA-related actions:

- checking a recovery code
- checking a TOTP value
- checking a Webauthn value

The rate limits were selected to be a balance of usability vs how long it would take a slow-roll actor to continue trying.

Metrics are emitted for monitoring and alerting purposes.

Refs: #8456

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -602,6 +602,7 @@ def test_root_factory_access_control_list():
                 Permissions.AdminObservationsWrite,
                 Permissions.AdminOrganizationsRead,
                 Permissions.AdminOrganizationsWrite,
+                Permissions.AdminOrganizationsNameWrite,
                 Permissions.AdminProhibitedEmailDomainsRead,
                 Permissions.AdminProhibitedProjectsRead,
                 Permissions.AdminProhibitedProjectsRelease,
--- a/warehouse/admin/templates/admin/organizations/detail.html
+++ b/warehouse/admin/templates/admin/organizations/detail.html
@@ -35,7 +35,7 @@
               type="button"
               class="btn btn-danger btn-block"
               data-toggle="modal"
-              data-target="#renameModal" {{ 'disabled' if not request.has_permission(Permissions.AdminOrganizationsWrite) }}
+              data-target="#renameModal" {{ 'disabled' if not request.has_permission(Permissions.AdminOrganizationsNameWrite) }}
               title="Rename"
             >
               <i class="icon fa fa-rotate"></i> Rename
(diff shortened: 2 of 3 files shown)


warehouse/accounts/__init__.py
```diff
@@ -178,6 +178,19 @@ def includeme(config):
     config.register_service_factory(
         RateLimit(global_login_ratelimit_string), IRateLimiter, name="global.login"
     )
+    # Register separate rate limiters for 2FA attempts
+    twofa_user_ratelimit_string = config.registry.settings.get(
```
ewjoachim, line 182: I'd tend to suggest "mfa" as a synonyme of 2fa that doesn't start with a digit, but we've used 2fa elsewhere, so that's just a suggestion.

ewjoachim (approved):
The change in the ratelimit, unless I'm mistaken, means attacks will be 2.4 times slower (so >50% chance of success in 7 months)

The new metrics will surely help measuring the normal "noise" level of users failing to 2FA (assuming those kinds of attacks are rare) (maybe we're about to discover something huge :D )

This will definitely help setting the right thresholds when later pulling out the effective measures that have been discussed (exponential backoff, auto-lockout, ...)

👍 for me and thanks for taking the time to act on this 5-year-old issue :) (issue birthday was just a few days ago) <3

warehouse/accounts/__init__.py
miketheman: Maybe a codebase-wide refactor? Going to keep it consistent for now, but agreed - multi-factor-authentication is more applicable.

miketheman:
@ewjoachim Thanks for the review!

> The change in the ratelimit, unless I'm mistaken, means attacks will be 2.4 times slower (so >50% chance of success in 7 months)

By my math, the "50 per day" is the most restrictive, so you'd need some 20k days to hit all combinations, so even if we assume that you get there in half of the entries, that's still some ~25 years. Unless I can't do math?

ewjoachim:
I trust your maths more than I trust mine. I think I saw the 50/h and missed the 50/d, but let's redo the maths :)

There are 1M possible TOTP codes (000000 to 999999)
Each code [counts as 3 codes](https://github.com/pypi/warehouse/blob/main/warehouse/utils/otp.py#L78-L82) because we check the 30 secs before and after, so each code has 1/333333 chances to be right (you could actually try the code `000000` every half hour instead of trying to hit them all)
The probability of having 0 success on n tries is `(333332/333333)^n` which goes under 50% after `n=math.ceil(math.log(0.5)/math.log(333332/333333))=231049` tries, which divided by 50 makes 4620 days which is 12.6 years. So an attacker would get >50% of success after 12.6 years.

ewdurbin (approved):

PR #18638, opened 2025-09-05
Remove all organization feature flag gating

ewdurbin:
Organizations features are now fully deployed and no longer need feature flag gating. This removes all DISABLE_ORGANIZATIONS flag checks and related infrastructure.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/warehouse/admin/flags.py
+++ b/warehouse/admin/flags.py
@@ -9,7 +9,6 @@ from warehouse.utils.db.types import bool_false
 
 
 class AdminFlagValue(enum.Enum):
-    DISABLE_ORGANIZATIONS = "disable-organizations"
     DISABLE_PEP740 = "disable-pep740"
     DISALLOW_DELETION = "disallow-deletion"
     DISALLOW_NEW_PROJECT_REGISTRATION = "disallow-new-project-registration"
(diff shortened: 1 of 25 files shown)


warehouse/admin/flags.py
```diff
@@ -9,7 +9,6 @@
 
 
 class AdminFlagValue(enum.Enum):
-    DISABLE_ORGANIZATIONS = "disable-organizations"
```
miketheman, line 12: Once merged, do you have a plan to remove the flag from the DB?

miketheman (approved):
Looks generally good.

There's some leftovers in tests that I _think_ won't be necessary any more, for example https://github.com/pypi/warehouse/pull/18638/files#diff-62c21c73d9a012104a029f268b8102f3f475fe63f8387afa38824b2473910fd7L123

There's also calls to it in templates, that as long as there's still utility to performing `request.organization_access` checks prior to other conditions, then they are probably fine to leave, but I haven't audited.

warehouse/admin/flags.py
ewdurbin: i'll include that in this PR, as a missing flag row will result in `flags.enabled` returning False, which is equivalent to the current value.

PR #18658, opened 2025-09-09
Handle too-long metadata filenames

di:
Fixes [WAREHOUSE-PRODUCTION-2A7](https://python-software-foundation.sentry.io/issues/6858098772/events/34ca5210251b4477b3ab65b033a0bcd6/).

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2046,7 +2046,10 @@ class TestFileUpload:
         release = ReleaseFactory.create(project=project, version="1.0")
         RoleFactory.create(user=user, project=project)
 
-        filename = f"{project.normalized_name.replace('-', '_')}-{release.version}-py2.py3-none-any.whl"
+        filename = "{}-{}-py2.py3-none-any.whl".format(
+            project.normalized_name.replace("-", "_"),
+            release.version,
+        )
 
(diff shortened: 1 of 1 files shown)


warehouse/forklift/legacy.py
```diff
@@ -1536,8 +1536,22 @@ def file_upload(request):
                         filename=filename, metadata_filename=metadata_filename
                     ),
                 )
-            with open(temporary_filename + ".metadata", "wb") as fp:
-                fp.write(wheel_metadata_contents)
+            try:
+                with open(temporary_filename + ".metadata", "wb") as fp:
+                    fp.write(wheel_metadata_contents)
```
miketheman, line 1541: question: do we know what filesystem max length this is supporting? This is only one limit, and we have other limits in play: S3/B2 max length is `<full object path> + filename` must be below 1024 - does our filesystem allow for longer than that?

di: Since this limit is dependent on the filesystem in use (i.e. it _could_ change in the future) I think we still should guard against this error here like this. If we want to additionally guard against the S3/B2 limits, we should do that as well (but note that the filename & path in the linked issue is _much_ shorter than 1024 characters, so it wouldn't have caught this specific instance).

miketheman (approved):

PR #18660, opened 2025-09-10
update json api reference with new fields

ucodery:
The API docs for the JSON endpoints no longer reflect the actual body returned by pypi. This is what comes back today from curling the example project (nearly; some embellishments were left in, such as some `null`s being shown as `""`, some use of ellipses, and none of the project's versions actually have an dynamic fields).


di (approved):
Thanks!

PR #18661, opened 2025-09-10
Move 'show password' widget on Login UI

### LINKED ISSUE #14132: Updating UI for 'show password' on PyPI main website login form

image

'Show password' is in a similar position to 'Remember me' used by most login forms. For the more reflexive among us (it's me, sorry.), this can result in accidentally showing your password off to anyone who might be around.

Moving the 'Show password' checkbox in-line with the "Password" label, or anywhere else that it can reduce mis-identification.

nlhkabu:
Closes #14132

- Moves "Show password" widget to above the password field, as on the registration page
- Moves "Forgot password" to next to the login button

### **Before:**
image

### **After**
Screenshot From 2025-09-10 06-59-21

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -1140,7 +1140,7 @@ msgid "Sponsors"
 msgstr ""
 
 #: warehouse/templates/accounts/login.html:7
-#: warehouse/templates/accounts/login.html:80 warehouse/templates/base.html:44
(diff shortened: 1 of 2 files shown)


miketheman:
This change modified the count of `tab` presses a user would have to perform to get to the password input - from one to two.
Is there a way to control the display order separately from the keyboard navigation order?

di:
I assume we could set `tabindex`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex

PR #18689, opened 2025-09-12
Require email confirmation for TOTP-based logins

### LINKED ISSUE #18425: Email confirmation for TOTP-based logins
### **What's the problem this feature will solve?**

A successful phishing attack can lead to account takeover if an attacker obtains both a user's password and a valid TOTP code.

### **Describe the solution you'd like**

To mitigate this risk, we should require confirmation via the primary email address for users who rely on TOTP for two-factor authentication. The proposed solution involves the following changes to the login flow:

1.  For any account using TOTP to log in, after the user enters their credentials and a valid TOTP code, they will not be immediately logged in.
2.  Instead, an email with a confirmation link will be sent to their primary email address.
3.  The user must click the confirmation link in the email to complete the login process. Login should be prevented if the user does not click the link.
4.  For each successful login, PyPI will aggregate user details (IP address and possibly device type) for security monitoring purposes.
5.  Users who authenticate with **WebAuthn** and **passkeys** will not be subject to this email confirmation step, as these methods are already considered more secure against phishing.

### **Additional context**

The following files and functions in the `warehouse` codebase will likely need to be modified to implement this feature:

* **Login and TOTP Handling:**
    * `warehouse/accounts/views.py`: The `login` view will need to be modified to trigger the email confirmation flow after a successful TOTP verification.
    * `warehouse/accounts/forms.py`: The `LoginForm` may need to be adjusted to handle the new confirmation step.
    * `warehouse/two_factor/views.py`: The views for handling TOTP will need to be updated to send the confirmation email instead of immediately logging the user in.
    * `warehouse/two_factor/forms.py`: The `TOTPForm` will need to be adjusted to accommodate the new workflow.
* **Email Confirmation:**
    * A new email template will need to be created for the login confirmation email.
    * A new view will be required to handle the confirmation link from the email.
    * `warehouse/email.py`: The email sending logic will need to be updated to handle the new confirmation email.
* **User Details Aggregation:**
    * The `login` view in `warehouse/accounts/views.py` should be updated to log the user's IP address and user agent string upon successful login.
* **WebAuthn and Passkeys:**
    * The views and forms related to WebAuthn and passkey authentication will need to be checked to ensure that the email confirmation flow is *not* triggered for these login methods.
        * `warehouse/webauthn/views.py`
        * `warehouse/webauthn/forms.py`

di:
Fixes #18425.

This PR maintains a record of device information across logins for each user:
- For TOTP logins, confirmation via a link sent to the primary email is required for each new device;
- For non-TOTP logins, no confirmation is required.

### DIFF
--- a/tests/functional/manage/test_account_publishing.py
+++ b/tests/functional/manage/test_account_publishing.py
@@ -7,10 +7,10 @@ from http import HTTPStatus
 import pytest
 import responses
 
+from tests.common.constants import REMOTE_ADDR
 from tests.common.db.accounts import UserFactory, UserUniqueLoginFactory
 from warehouse.accounts.models import UniqueLoginStatus
 from warehouse.utils.otp import _get_totp
-from tests.common.constants import REMOTE_ADDR
 
 
 @pytest.mark.usefixtures("_enable_all_oidc_providers")
--- a/tests/functional/manage/test_organization_publishing.py
+++ b/tests/functional/manage/test_organization_publishing.py
@@ -7,12 +7,12 @@ from http import HTTPStatus
 import pytest
 import responses
 
+from tests.common.constants import REMOTE_ADDR
 from tests.common.db.accounts import UserFactory, UserUniqueLoginFactory
 from tests.common.db.organizations import OrganizationFactory, OrganizationRoleFactory
 from warehouse.accounts.models import UniqueLoginStatus
 from warehouse.organizations.models import OrganizationRoleType
 from warehouse.utils.otp import _get_totp
-from tests.common.constants import REMOTE_ADDR
 
 
 @pytest.mark.usefixtures("_enable_all_oidc_providers")
--- a/tests/functional/test_login.py
+++ b/tests/functional/test_login.py
@@ -5,9 +5,8 @@ import time
 from http import HTTPStatus
 
 from tests.common.db.accounts import UserFactory
-from warehouse.accounts.models import UserUniqueLogin
+from warehouse.accounts.models import UniqueLoginStatus, UserUniqueLogin
 from warehouse.utils.otp import _get_totp
-from warehouse.accounts.models import UniqueLoginStatus
 
 
 def test_unrecognized_login_with_totp(webtest):


warehouse/config.py
```diff
@@ -410,6 +410,7 @@ def configure(settings=None):
     maybe_set(settings, "token.email.secret", "TOKEN_EMAIL_SECRET")
     maybe_set(settings, "token.two_factor.secret", "TOKEN_TWO_FACTOR_SECRET")
     maybe_set(settings, "token.remember_device.secret", "TOKEN_REMEMBER_DEVICE_SECRET")
+    maybe_set(settings, "token.confirm_login.secret", "TOKEN_CONFIRM_LOGIN_SECRET")
```
miketheman, line 413: I don't see this used anywhere. Leftover from an idea? If removed, also remove from `dev/environment`

warehouse/templates/email/unrecognized-login/body.txt
```diff
@@ -0,0 +1,12 @@
+A login attempt was made from an unrecognized device.
+
+To complete your login, please visit the following link:
+
+{{ request.route_url('accounts.confirm-login', _query={'token': token}) }}
```
miketheman, line 6: I think we need to add the host here if we expect this to work on TestPyPI as well:
```
..., _host=request.registry.settings.get('warehouse.domain')
```

warehouse/email/__init__.py
```diff
@@ -983,6 +983,16 @@ def send_recovery_code_reminder_email(request, user):
     return {"username": user.username}
 
 
+@_email("unrecognized-login")
+def send_unrecognized_login_email(request, user, *, ip_address, user_agent, token):
+    return {
+        "username": user.username,
+        "ip_address": ip_address,
```
miketheman, line 990: suggestion: include geo data as well

warehouse/accounts/views.py
```diff
@@ -401,20 +411,101 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
...
-            user_service.update_user(userid, last_totp_value=form.totp_value.data)
+            user = user_service.get_user(userid)
 
-            resp = HTTPSeeOther(redirect_to)
-            _set_userid_insecure_cookie(resp, userid)
+            unique_login = (
+                request.db.query(UserUniqueLogin)
+                .filter(
+                    UserUniqueLogin.user_id == userid,
+                    UserUniqueLogin.ip_address == request.remote_addr,
```
miketheman, line 423: Since these appear to be the "couplet" of conditions used in most lookups, there should probably be a composite index for both columns.
Here's an example of how that's done: https://github.com/pypi/warehouse/blob/e9b70d0bd7b637e4aa5e459b83b46ba32f0031c9/warehouse/packaging/models.py#L581-L583

Unless the UniqueConstraint satisfies this already?

warehouse/accounts/views.py
```diff
@@ -401,20 +411,101 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
...
+                if unique_login.status == UniqueLoginStatus.CONFIRMED:
+                    # We've seen this device before for this user and they've
+                    # confirmed it, log in the user
+                    two_factor_method = "totp"
+                    _login_user(
+                        request, userid, two_factor_method, two_factor_label="totp"
+                    )
+                    user_service.update_user(
+                        userid, last_totp_value=form.totp_value.data
+                    )
```
miketheman, line 438: question: would it make sense to also update a `unique_login.last_used` or something like that when we see it again?

warehouse/accounts/views.py
```diff
@@ -401,20 +411,101 @@ def two_factor_and_totp_validate(request, _form_class=TOTPAuthenticationForm):
...
+
+                # Get User Agent Information
+                user_agent_info_data = {}
+                if user_agent_str := request.headers.get("User-Agent"):
+                    try:
+                        parsed = linehaul_user_agent_parser.parse(user_agent_str)
+                        if (
+                            parsed
+                            and parsed.installer
+                            and parsed.installer.name == "Browser"
```
miketheman, line 494: Should we exclude anything that doesn't parse as a Browser? It be interesting to learn if someone's user agent doesn't parse as a browser and is used for authentication.

warehouse/accounts/views.py
```diff
@@ -955,6 +1046,87 @@ def _error(message):
...
+        return _error(request._("Invalid token: no token supplied"))
+
+    # Check whether this token is being used correctly
+    if data.get("action") != "login-confirmation":
+        return _error(request._("Invalid token: not a login confirmation token"))
+
+    # Check whether a user with the given user ID exists
+    user = user_service.get_user(uuid.UUID(data.get("user.id")))
+    if user is None:
+        return _error(request._("Invalid token: user not found"))
```
miketheman, line 1177: question: if I spam this endpoint with enough tokens, is it possible for me to enumerate user existence?
Asking another way, should this endpoint have rate limits?

warehouse/accounts/views.py
```diff
@@ -955,6 +1046,87 @@ def _error(message):
...
+        request.db.query(UserUniqueLogin)
+        .filter(UserUniqueLogin.id == unique_login_id)
+        .one_or_none()
+    )
+
+    if unique_login is None:
+        return _error(request._("Invalid login attempt."))
+
+    if unique_login.ip_address != request.remote_addr:
+        return _error(request._("Device details didn't match, please try again"))
```
miketheman: question: what user behavior would change after a "try again" here?

warehouse/accounts/views.py
```diff
@@ -1442,6 +1614,30 @@ def _login_user(request, userid, two_factor_method=None, two_factor_label=None):
...
+    )
+    if unique_login is None and two_factor_method != "totp":
+        # We haven't seen this login before. Create a new one and mark it as confirmed
+        # if this is non-TOTP.
+        unique_login = UserUniqueLogin(
+            user_id=userid,
+            ip_address=request.remote_addr,
+            status=UniqueLoginStatus.CONFIRMED,
+        )
+        request.db.add(unique_login)
```
miketheman: question: should this logic trigger if the user used recovery codes?

warehouse/accounts/models.py
```diff
@@ -475,3 +480,42 @@ class ProhibitedUserName(db.Model):
...
+
+    user_id: Mapped[UUID] = mapped_column(
+        PG_UUID(as_uuid=True),
+        ForeignKey("users.id", onupdate="CASCADE", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    user: Mapped[User] = orm.relationship(back_populates="unique_logins")
+
+    ip_address: Mapped[str] = mapped_column(String, nullable=False)
```
miketheman, line 512: question: Should this be its own string, or a relationship to `warehouse.events.models.IpAddress`?

tests/unit/accounts/test_views.py
```diff
@@ -847,39 +911,49 @@ def test_get_returns_recovery_code_status(self, pyramid_request, redirect_url):
...
+            last_login=(
+                datetime.datetime.now(datetime.UTC) - datetime.timedelta(days=1)
+            ),
+        )
+        monkeypatch.setattr(
+            type(user),
+            "has_recovery_codes",
+            property(lambda u: has_recovery_codes),
+        )
+        user.record_event = pretend.call_recorder(lambda *a, **kw: None)
```
miketheman, line 998: thought: instead of monkeypatching, we could pass something like:
```python
recovery_codes=[RecoveryCode(code="fake")] if has_recovery_codes else [],
```
to UserFactory.create() - but since you're following an existing pattern, that's fine to defer.

warehouse/templates/email/unrecognized-login/body.html
miketheman, line 1: When running a manual verification, the email sent was only text, not HTML. Any clue as to why?

warehouse/templates/email/unrecognized-login/body.txt
miketheman, line 1: This file is missing the license header. I don't know why the detector didn't catch it.

warehouse/config.py
di: It's automatically used by the `TokenServiceFactory`: https://github.com/pypi/warehouse/blob/f891447aa5fb0375dc963f1adf7d99b26ddb3aa9/warehouse/accounts/services.py#L807

based on the name of the service, which is `confirm_login` https://github.com/pypi/warehouse/pull/18689/files#diff-fd646332e15b1650ac3a05f944bf6ae36335637b687d06676bb67f7d6f3b1fefR107

warehouse/templates/email/unrecognized-login/body.txt
di: We don't do it here and it seems to work? https://github.com/pypi/warehouse/blob/f891447aa5fb0375dc963f1adf7d99b26ddb3aa9/warehouse/templates/email/verify-email/body.html#L8

I think this is only necessary when emails are being sent from a task, because then the request is missing the original host.

warehouse/accounts/views.py
di: I'm not sure what that would give us. Do we want to "expire" trusted devices after some period of time?

warehouse/accounts/views.py
di: I guess we could provide the raw user agent here instead if it doesn't parse, is that what you're suggesting?

warehouse/templates/email/unrecognized-login/body.txt
```diff
@@ -0,0 +1,12 @@
+A login attempt was made from an unrecognized device.
```
di, line 2:
```suggestion
{# SPDX-License-Identifier: Apache-2.0 -#}
A login attempt was made from an unrecognized device.
```

warehouse/templates/email/unrecognized-login/body.html
di: Are you talking about the `ConsoleAndSMTPEmailSender` in local dev? I don't think it displays HTML, only text: https://github.com/pypi/warehouse/blob/f891447aa5fb0375dc963f1adf7d99b26ddb3aa9/warehouse/email/services.py#L157

warehouse/config.py
miketheman: Derp, thanks. That's something I'll have to remember.

warehouse/templates/email/unrecognized-login/body.html
miketheman: I am - one line above it is the note that you can visualize the HTML version in MailDev, which usually displays something richer for other emails.

warehouse/templates/email/unrecognized-login/body.txt
miketheman: That's correct, I'm having trauma induced memories from past task-related emails 😭

warehouse/accounts/views.py
di: I think the token would need to be valid because the token checks come before this? So using this as an oracle for user IDs would mean the attacker would need to be able to create a valid token, which would be a larger problem, and generally shouldn't be possible.

warehouse/accounts/views.py
di: Presumably this would happen to a legitimate user if they followed the confirmation link on a different device than what they originally used to log in with. Perhaps this should say:

```suggestion
    if unique_login.ip_address != request.remote_addr:
        return _error(request._("Device details didn't match, please try again from the device you originally used to log in."))
```

warehouse/accounts/models.py
di: Given that that model is primarily used for banning IPs, I didn't want to overload it, but I think if we need to unify these in the future we could.

warehouse/accounts/views.py
di: 9f16701429579e6e2347e193249543472d52eebb

warehouse/accounts/views.py
di: b0bcc60a46d4dd46529ea37bb24bb0881462f4b9

warehouse/accounts/views.py
di: b625aa9343c8e31814d58b7479fdeac2603c5a2c

warehouse/accounts/models.py
miketheman: Fair enough, it's not the worst duplicated data in the system, and having it persisted has its advantages as well.

warehouse/accounts/views.py
miketheman: There's likely going to be user agents that parse, but maybe not parse as a browser, and if that happens, we get no data at all to pass to the email - is that better?

warehouse/accounts/views.py
miketheman: Let me try to rephrase.

By the time we reach here, the token must be valid - I think we're on the same page for that.

If the user id in the token data is not valid, but the token is, it's a curiosity - was the user get removed after the token was created, or was the token generated by something capable of generating valid-yet-incorrect tokens?

Maybe rate limits should be on invalid/expired/missing tokens, to prevent brute force token attempts? Or maybe that needs to go down into the TokenServiceFactory layer as a general approach?

I've looked at other patterns using tokens and it seems that this is not a new pattern, so we can defer solving that.

warehouse/accounts/views.py
miketheman: I think having a "90 days since last used" window or something like that might be a decent balance between prompting the user experience, and giving us a way to invalidate these records if unused.

warehouse/accounts/views.py
di: > was the user get removed after the token was created

I think this would always be the case.

> or was the token generated by something capable of generating valid-yet-incorrect tokens

This shouldn't be possible unless the token secret has leaked, which means we have a bigger problem that an attacker being able to use this as an oracle for user IDs.

Given that the same is true for other token services, marking this as resolved.

warehouse/accounts/views.py
di: 14a7b4b69c31d8c51e8aae8386f3ba777596ce64

warehouse/accounts/views.py
di: 9740bf4

warehouse/templates/email/unrecognized-login/body.html
di: 7353e69

warehouse/email/__init__.py
di: 264e0ac

warehouse/migrations/versions/4c20f2342bba_add_useruniquelogin.py
```diff
@@ -0,0 +1,81 @@
...
+Create Date: 2025-07-29 00:55:39.682180
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "4c20f2342bba"
+down_revision = "a6994b8bed95"
```
miketheman:
```suggestion
down_revision = "daf71d83673f"
```

warehouse/migrations/versions/4c20f2342bba_add_useruniquelogin.py
```diff
@@ -0,0 +1,81 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Refactor UserUniqueLogin to use mapped_column
+
+Revision ID: 4c20f2342bba
+Revises: a6994b8bed95
```
miketheman:
```suggestion
Revises: daf71d83673f
```

di: 6017ef5

warehouse/migrations/versions/4c20f2342bba_add_useruniquelogin.py
di: 6017ef5

warehouse/admin/templates/admin/users/detail.html
```diff
@@ -988,6 +988,41 @@ <h3 class="card-title">Pending OpenID Connect Publishers</h3>
...
+              <thead>
+                <tr>
+                  <th scope="col">Created</th>
+                  <th scope="col">IP address</th>
+                  <th scope="col">Status</th>
+                  <th scope="col">Device Information</th>
+                </tr>
+              </thead>
+              <tbody>
+                {% for login in user.unique_logins %}
```
ewdurbin, line 1009: we are almost certainlty going to need to paginate this... but for now seems fine since its admin.

warehouse/templates/accounts/unrecognized-device.html
```diff
@@ -0,0 +1,17 @@
...
+{% block content %}
+  <div class="horizontal-section">
+    <div class="site-container">
+      <h1 class="page-title">Unrecognized device</h1>
+      <p>We did not recognize this device. Please check your email for a login confirmation link.</p>
+      <p>
+        You should have received an email from <strong>noreply@pypi.org</strong> with the subject line "<strong>Unrecognized login to your PyPI account</strong>".
+      </p>
+      <p>
+        If you did not make this change, you can email <a href="mailto:admin@pypi.org">admin@pypi.org</a> to communicate with the PyPI administrators.
```
ewdurbin: "If you did not make this change"? Seems wrong?

warehouse/templates/email/unrecognized-login/body.html
```diff
@@ -0,0 +1,26 @@
...
+    </li>
+    <li>
+      <strong>Location:</strong> {{ ip_address.geo_ip or "Unknown" }}
+    </li>
+    <li>
+      <strong>User agent:</strong> {{ user_agent }}
+    </li>
+  </ul>
+  <p>
+    If you did not make this change, you can email admin@pypi.org to
```
ewdurbin: "If you did not attempt this login" instead?

di: 8992f0f

warehouse/templates/accounts/unrecognized-device.html
di: 821531c

PR #18690, opened 2025-09-14
Improved the user's experience!

PortalWalkerCodes:
Improved user experience!

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -39,8 +39,8 @@ from warehouse.events.tags import EventTag
 from warehouse.i18n import localize as _
 
 # Common messages, set as constants to keep them from drifting.
-INVALID_EMAIL_MESSAGE = _("The email address isn't valid. Try again.")
-INVALID_PASSWORD_MESSAGE = _("The password is invalid. Try again.")
+INVALID_EMAIL_MESSAGE = _("The email address isn't valid. Please try again.")
+INVALID_PASSWORD_MESSAGE = _("The password is invalid. Please Try again.")
 INVALID_USERNAME_MESSAGE = _(
     "The username is invalid. Usernames "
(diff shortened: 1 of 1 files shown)


woodruffw (changes requested):
The admins have the final say here, but I don't think this PR is ultimately an improvement: these messages are translated into dozens of languages, so changing their English version will invalidate all other ones for a relatively small benefit (saying "please" more). That results in a strictly worse user experience for a lot of people.

warehouse/accounts/forms.py
```diff
@@ -59,7 +59,7 @@ class PreventNullBytesValidator:
 
     def __init__(self, message=None):
         if message is None:
-            message = _("Null bytes are not allowed.")
+            message = _("Please enter a value.")
```
woodruffw, line 62: This is a loss of information: this validator is *specifically* checking for null bytes (i.e. ASCII NULL), so "Please enter a value" doesn't describe what it's actually rejecting.

Or in other words: `\x00invalid` *is* a value, it's just not a valid one. But with this change we'd emit a confusing error message suggesting that it isn't a value.

warehouse/accounts/forms.py
```diff
@@ -191,7 +191,7 @@ def validate_password(self, field):
             except TooManyFailedLogins as err:
                 raise wtforms.validators.ValidationError(
                     _(
-                        "There have been too many unsuccessful login attempts. "
+                        f"There have been too many unsuccessful login attempts. "
```
woodruffw, line 194: This f-string has no effect, as far as I can tell? It might actually break the code below, since it'll format `{time}` instead of letting the i18n machinery take care of it.

ewdurbin:
As @woodruffw points out, unnecessary churn in translated strings is not worth it here.

PR #18774, opened 2025-09-29
oidc/gitlab: Support single-character workflow filenames

Turbo87:
I'm currently working on GitLab support for the Trusted Publishing implementation of crates.io and noticed that the regular expression that PyPI uses does not accept single character workflow filenames like `a.yml`.

This PR should hopefully fix the problem if CI confirms it... 🤞

### DIFF
--- a/warehouse/oidc/models/gitlab.py
+++ b/warehouse/oidc/models/gitlab.py
@@ -33,8 +33,8 @@ _WORKFLOW_FILEPATH_RE = re.compile(
                       # component of the claim.
 
     (                 # our capture group
-        .+            # match one or more of any character, including slashes
-        [^/]          # match at least one non-slash character, to prevent
+        .*            # match zero or more of any character, including slashes
+        [^/]          # match exactly one non-slash character, to prevent
                       # empty basenames (e.g. `foo/.yml`)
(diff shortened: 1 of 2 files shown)


warehouse/oidc/models/gitlab.py
```diff
@@ -33,8 +33,8 @@
                       # component of the claim.
 
     (                 # our capture group
-        .+            # match one or more of any character, including slashes
-        [^/]          # match at least one non-slash character, to prevent
+        .*            # match zero or more of any character, including slashes
+        [^/]+         # match at least one non-slash character, to prevent
```
di:
```suggestion
        [^/]         # match exactly one non-slash character, to prevent
```

I don't think we need this change? The comment should be updated though.

Turbo87: yeah, I guess without `+` should probably also work.

adding another suggestion to fix the whitespace alignment 😉 

```suggestion
        [^/]          # match at least one non-slash character, to prevent
```

Turbo87: and forgot the comment change... 🤦 

let me fix this properly 😅

di (approved):

PR #18776, opened 2025-09-30
Restyle org request page

nlhkabu:
**Behaviour change:**

- If the PyPI admin has sent several information requests, the user now only sees a response form under the last message. This makes it appear as though the conversation is a single thread.
- Show information requests as a single conversation thread in the PyPI admin

**Screenshots:**

localhost_manage_organizations_application_fbd5ea4d-82e8-453c-bfeb-87fdc72f9033_ (2)

localhost_admin_organization_applications_fbd5ea4d-82e8-453c-bfeb-87fdc72f9033_

### DIFF
--- a/warehouse/admin/templates/admin/organization_applications/detail.html
+++ b/warehouse/admin/templates/admin/organization_applications/detail.html
@@ -336,7 +336,7 @@
       </div>
     </div>
 
-    {% set information_requests = organization_application.information_requests %}
+    {% set information_requests = organization_application.information_requests|reverse %}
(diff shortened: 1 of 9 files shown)


ewdurbin:
Once linting is clear, this is good to go!

ewdurbin:
Hm, something about this seems to have broken the admin side. The "new response" icon now shows for all applications, even if there is not a new response.

Screenshot 2025-10-03 at 8 35 29 AM

nlhkabu:
Ok, thanks for flagging @ewdurbin - I'll take a look and see what is going on

PR #18793, opened 2025-10-02
user-docs: import PEP 541 policy as a page

woodruffw:
From conversation in https://github.com/psf/policies/pull/41 -- I've moved this into the PyPI user docs instead. I *think* the "project management" hierarchy is the best place for it, since we have similar policy-like stuff under it already (like quotas and yanking).

I've left the page verbatim from that other PR, which also means it doesn't really read exactly like the language in the other pages (it's more PEP-y). I wasn't sure how much I should edit it for style given that it's standard language 🙂 

xref #17999 for additional context on what's motivating this.

CC @Thespi-Brain @ewdurbin @miketheman from the original PR

### DIFF
--- a/docs/mkdocs-user-docs.yml
+++ b/docs/mkdocs-user-docs.yml
@@ -73,6 +73,7 @@ nav:
   - "Project Management":
       - "project-management/storage-limits.md"
       - "project-management/yanking.md"
+      - "project-management/name-retention.md"
   - "Organization Accounts":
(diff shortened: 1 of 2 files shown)


miketheman:
Generally looks good to me, assigning to @Thespi-Brain for a deeper review.

Thespi-Brain (approved):
Looks good!

PR #18798, opened 2025-10-03
test: add types to oidc.* modules

miketheman:
Increase safety when working with these complex modules, while learning how they interoperate.


miketheman:
Almost ready, there's some more precision to add in `warehouse.oidc.forms.*`, ~~`warehouse.oidc.tasks`~~ - that's coming next.

PR #18805, opened 2025-10-06
fix: disallow uppercase values in GitLab publishers

### LINKED ISSUE #18797: GitLab Trusted Publisher case sensitivity
**Describe the bug**
Setting up a trusted publisher for GitLab CI requires case sensitivity that does not match the actual spelling.

I have a project on GitLab where the organization has a captialized letter in it: `coBib`. The URL is https://gitlab.com/cobib/cobib.
I configured a trusted publisher for the `cobib` package specifying the owner as `coBib` and project as `cobib`.
This resulted in the following error:
```
ERROR    TrustedPublishingFailure: The token request failed; the index server   
         gave the following reasons:                                            
                                                                                
         * `invalid-publisher`: valid token, but no corresponding publisher     
         (Publisher with matching claims was not found)  
```

It was only after playing around a bit, that I tried to set owner to `cobib`, after which my trusted publisher config worked.

**Expected behavior**
The `owner` field of a trusted publisher should not be case sensitive.

I thought, that this is already the case, but I still ran into this problem on Wednesday of this week.

**To Reproduce**
- Create a project on GitLab with an organization name using a capitalized letter
- Configure a trusted publisher for `.gitlab-ci.yml` where the correct capitalized spelling is used
- Observe the token failure

**My Platform**
- Project URL: https://gitlab.com/cobib/cobib
- GitLab CI config at the time of failure: https://gitlab.com/cobib/cobib/-/blob/v5.5.3/.gitlab-ci.yml (no changes to this file were necessary to make v5.5.5 publish successfully, as I only had to update the trusted publisher config as described above)

**Additional context**
There are some (semi-)related issues which all appear to be resolved:
- https://github.com/pypi/warehouse/issues/14622
- https://github.com/pypi/warehouse/issues/15498
- https://github.com/pypi/warehouse/issues/18330

Looking at the GitLab `oidc` model, I would even expect the current code to work, which further adds to my confusion of why it did not:
https://github.com/pypi/warehouse/blob/5dec3cdedd4fa0bbba0d796d27020b8314ea50a4/warehouse/oidc/models/gitlab.py#L129

miketheman:
The API and claims generated from GitLab will be desensitized when
generating JWTs.

Instead of updating the regular expression and using the same generic
error message, create a custom validator with a custom message.

Resolves #18797

### DIFF
--- a/tests/unit/oidc/forms/test_gitlab.py
+++ b/tests/unit/oidc/forms/test_gitlab.py
@@ -152,6 +152,16 @@ class TestGitLabPublisherForm:
             },
             {"project": "some", "namespace": "some", "workflow_filepath": None},
             {"project": "some", "namespace": "some", "workflow_filepath": ""},
+            {
+                "namespace": "some-owner",
+                "project": "CaseSensitive",
+                "workflow_filepath": "some-workflow.yml",
+            },
+            {
+                "namespace": "CaseSensitive",
+                "project": "some-repo",
+                "workflow_filepath": "some-workflow.yml",
+            },
         ],
     )
     def test_validate_basic_invalid_fields(self, data):
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -394,7 +394,7 @@ msgid "Select project"
 msgstr ""
 
 #: warehouse/manage/forms.py:507 warehouse/oidc/forms/_core.py:36
-#: warehouse/oidc/forms/gitlab.py:47
+#: warehouse/oidc/forms/gitlab.py:60
 msgid "Specify project name"
 msgstr ""
 
@@ -677,7 +677,7 @@ msgid "Expired invitation for '${username}' deleted."
 msgstr ""
 
 #: warehouse/oidc/forms/_core.py:38 warehouse/oidc/forms/_core.py:49
-#: warehouse/oidc/forms/gitlab.py:50 warehouse/oidc/forms/gitlab.py:54
+#: warehouse/oidc/forms/gitlab.py:64 warehouse/oidc/forms/gitlab.py:68
 msgid "Invalid project name"
 msgstr ""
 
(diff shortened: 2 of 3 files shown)


miketheman:
I elected to update the form to reject and thus guide the user to input what becomes the ground truth value, rather than lower-case it during query time, so that we are not manipulating the ground truth, and reduce the time-to-discover a problem with the configuration.

miketheman:
Looks like:
image

di:
I think if GitLab is case-insensitive, we should just be case-insensitive as well?

miketheman:
> I think if GitLab is case-insensitive, we should just be case-insensitive as well?

I previously wrote: 

> I elected to update the form to reject and thus guide the user to input what becomes the ground truth value, rather than lower-case it during query time, so that we are not manipulating the ground truth, and reduce the time-to-discover a problem with the configuration.

Expanding on this - if we do lowercase at query-time, our machinery would still provide the end user a "publisher not found" message when incorrect, and then kick off the questions of "what's the ground truth configured, what did we get in the claims, etc". Whereas prompting them to input the value that we will expect to find later at the outset removes a set of concerns, especially as I look at other behaviors related to validation.

The way I see it, if there's data stored in the DB, that's the ground truth. We should store what folks give us, instead of lowercasing at insert-time, since then we modify the value they gave us, creating a mismatch of expectations.

If we store the user-supplied, case-sensitive value (like we do today), then we must remember that **every time** we access that value, we have to lowercase it, even during manual debugging in a SQL prompt. My memory isn't that good, so I'm preferring to have the user control the input and provide valid values and reduce potential confusion down the line.

Do you see an inherent problem with this approach?

di:
Yeah, I think the issue is that if the user puts in `caseSensitiveProject` (because that's what they see displayed on GitLab and we reject it because we're being case-sensitive and overly restrictive, that's somewhat confusing:
- If the user knows and expects GitLab to be case-insensitive, it's confusing because it's the opposite of what they know.
- If the user *doesn't* know that GitLab is case-insensitive, it's confusing because it seems like we're telling them to put in something other than what their repo is named.

> The way I see it, if there's data stored in the DB, that's the ground truth. We should store what folks give us, instead of lowercasing at insert-time, since then we modify the value they gave us, creating a mismatch of expectations.

I agree, we should not mutate what we've been originally given by the user before storing it.

> If we store the user-supplied, case-sensitive value (like we do today), then we must remember that every time we access that value, we have to lowercase it, even during manual debugging in a SQL prompt. My memory isn't that good, so I'm preferring to have the user control the input and provide valid values and reduce potential confusion down the line.

I disagree, I don't think this is worth making users jump through extra hoops here just to do normalization for us.

I think either way when doing a manual query, we'll need to remember that there might be a difference between the repo name a user provides us and what's in the database (e.g. if they come to us with a repo named `https://gitlab.com/coBib/cobib`, we'll have to remember that this will be stored as `cobib` and not `coBib` in the database.

Additionally, this doesn't help any of the users that have a mis-configured publisher currently, whereas updating the query to do normalization would make their publishers work immediately:

```
warehouse=> SELECT count(*) FROM gitlab_oidc_publishers WHERE namespace ~ '[A-Z]' OR project ~ '[A-Z]';
 count
-------
    66
(1 row)
```

ewdurbin:
I'm in favor of us accepting whatever the user gives, more or less full-stop. this is a recipe for frustration and errors, not dissimilar from a banking app refusing to let me paste in an account number from my password manager and forcing me to type it in piece by piece and opening the door to typos.

Example: a user wants to paste in `PyPI/Receiver` and is refused they may very well type in `pypi/reciever` and introduce a typo (my favorite typo in fact).

woodruffw:
Yeah, I *think* PyPI should probably be permissive of casing decisions made by the user at TP enrollment/provisioning time -- it's frustrating that CI/CD services don't tend to document these details super well, but IMO the user experience is a lot better if PyPI behaves *like* GitHub and GitLab (which are case-insensitive-but-preserving AFAICT in this context) and allows matching regardless of how the user originally cased their inputs.

In other words, IMO we should allow uppercase values here and accept a small amount of additional complication at lookup/matching time. 

(Relatedly, I think it was a mistake on my part to do a "normalization" trick for GitHub where we replace the user's given username with the one the GitHub API responds with -- similarly to here, we probably *should* just take whatever the user puts in as long as it's valid and honor their case choices.)

Ref: https://github.com/pypi/warehouse/blob/8b77163904ee0ba8900185919c0795b9e353726f/warehouse/oidc/forms/github.py#L128-L130

miketheman:
Thanks for the explanations and reasonings.
I've created https://github.com/pypi/warehouse/pull/18824 as an alternative, where we continue to allow users to submit what they will, and lowercase the stored values at query time.

I suspect that this may still be confusing when a publisher is not found, as the error message of `* `invalid-publisher`: valid token, but no corresponding publisher` and the output claims (from the JWT) will not match what is configured, but that's a different problem to solve.

PR #13187, opened 2023-03-10
Reorganize view-code related to Organizations

miketheman:
As detailed in #13180, here's a working proof of concept of splitting apart some of the largest files in the codebase.

Looking for feedback on:

- file/folder naming structure
- view helper concept
- anything else you think is relevant?

PR #18811, opened 2025-10-07
refactor: extract oidc management views to module

miketheman:
`warehouse/manage/views/__init__.py` is already a complex module, with
over 3k LOC. Split out any OIDC code (conveniently in a single class) to
a new module.
This should allow for more focused edits when there are changes.


di:
Makes sense to me to pull these out. I would expect us to use the same organizational pattern we use elsewhere though, where we create a `views` directory instead, which would contain `__init__.py`, `publishers.py`, etc. (I would also expect to pull the related tests out into a separate file since these are generally 1:1).

miketheman:
I agree - however these are `manage/` views, and are already collected under that namespace, following their routes. We have `oidc/views.py` for the non-management views.

I'll move the tests out in another commit.

ewdurbin:
> I'll move the tests out in another commit.

another commit or another PR?

miketheman:
> > I'll move the tests out in another commit.
> 
> another commit or another PR?

commit - [7069bbc](https://github.com/pypi/warehouse/pull/18811/commits/7069bbc4784b49121d2d442af11ab308bdbfd0dd)

ewdurbin:
> Makes sense to me to pull these out. I would expect us to use the same organizational pattern we use elsewhere though, where we create a `views` directory instead, which would contain `__init__.py`, `publishers.py`, etc. (I would also expect to pull the related tests out into a separate file since these are generally 1:1).

Yeah, If we're going to further churn this, `git mv warehouse/manage/views.py warehouse/manage/views/__init__.py` along with the others seems like a sane thing to just rip all the bandaids at once.

di:
> Yeah, If we're going to further churn this, git mv warehouse/manage/views.py warehouse/manage/views/__init__.py along with the others seems like a sane thing to just rip all the bandaids at once.

I'm doing a double take because I thought `warehouse/manage/views.py` existed but it doesn't, nevermind me!

miketheman:
Confused - `git mv warehouse/manage/views.py` - I did that way back in [`546f16c` (#13187)](https://github.com/pypi/warehouse/pull/13187/commits/546f16c0a5421cc2590ba836ea02f81f5b90f30d) ?

di (approved):

ewdurbin:
> Confused - `git mv warehouse/manage/views.py` - I did that way back in [`546f16c` (#13187)](https://github.com/pypi/warehouse/pull/13187/commits/546f16c0a5421cc2590ba836ea02f81f5b90f30d) ?

always look at the code, not the chatter :) my bad.

PR #18840, opened 2025-10-10
feat(cli): add search index management tools

miketheman:
Two tools:

- print out current state of cluster indices
- remove older, non-current-alias indices


warehouse/cli/search.py
```diff
@@ -22,3 +24,68 @@ def reindex(config):
...
+
+    aliases = client.indices.get_alias(index="production*,staging*")
+    click.echo("\nCurrent Aliases:")
+    for index, alias in aliases.items():
+        for a in alias["aliases"]:
+            click.echo(f"{a} -> {index}")
+
+
+@search.command()
+@click.argument("env_name", type=str, required=True)
```
JacobCoffee:
```suggestion
@click.argument("env_name", type=click.Choice(['production', 'staging'], required=True)
```
could you do this to not allow other entries

miketheman: Good idea! I'll add `development` too

PR #18841, opened 2025-10-10
feat: Add custom OIDC issuers to orgs

### LINKED ISSUE #18831: Register a custom OIDC Issuer URL for an Organization via PyPI Admin
To support GitHub Enterprise Server (GHES) and GitLab Self-Managed (GLSM), the issuers for these services will not be the same as the `iss` from a JWT claims that we would see from the publicly-available service.

To support this approach, create data models, admin views, and associated relevant behaviors for an admin to add an issuer URL, which can be queried for later when a JWT with it is passed to `mint-token` API call.

miketheman:
Allows storing a specific `issuer_url` associated to an org, for later lookup.

Resolves #18831

### DIFF
--- a/tests/unit/admin/test_routes.py
+++ b/tests/unit/admin/test_routes.py
@@ -70,6 +70,16 @@ def test_includeme():
             domain=warehouse,
         ),
         pretend.call(
+            "admin.organization.add_oidc_issuer",
+            "/admin/organizations/{organization_id}/oidc-issuers/add/",
+            domain=warehouse,
+        ),
+        pretend.call(
+            "admin.organization.delete_oidc_issuer",
+            "/admin/organizations/{organization_id}/oidc-issuers/{issuer_id}/delete/",
+            domain=warehouse,
+        ),
+        pretend.call(
             "admin.organization_application.list",
             "/admin/organization_applications/",
             domain=warehouse,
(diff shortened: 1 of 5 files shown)


warehouse/organizations/models.py
```diff
@@ -190,6 +190,54 @@ class OrganizationStripeCustomer(db.Model):
     customer: Mapped[StripeCustomer] = relationship(lazy=False)
 
 
+class OIDCIssuerType(enum.StrEnum):
+    GitHub = "github"
+    GitLab = "gitlab"
+    Google = "google"
+    ActiveState = "activestate"
```
di, line 197: A nit, maybe this should live in `/warehouse/oidc/` instead? Also, should we find some way to ensure this remains consistent with the `polymorphic_identity` used for each of the individual publisher model types?

miketheman: I put it here since it's only currently used by this module, but will happily move it later if this approach works well.
I like the idea of constraining the potential values, but will defer for now, since it'll need to modify the mixin to be an enum choice, not a string, and update here:
https://github.com/pypi/warehouse/blob/9cfbbf940af85786d479a5e2f0c486355b2ff96a/warehouse/oidc/models/_core.py#L118-L127

PR #18846, opened 2025-10-10
Further lock-down ZIP archive features

sethmlarson:
This further locks down the ZIP archive checks we do so that non-standardized ZIP archives or archives using uncommon/unnecessary features aren't uploaded to PyPI


tests/unit/utils/test_zipfiles.py
```diff
@@ -70,8 +82,8 @@ def test_bad_zips(filename, error):
 @pytest.mark.parametrize("filename", list(os.listdir(ZIPDATA_DIR / "accept")))
 def test_good_zips(filename):
     result = zipfiles.validate_zipfile(zippath(f"accept/{filename}"))
-    assert result[0] is True
     assert result[1] is None
+    assert result[0] is True
```
miketheman: question: curious why the order changed here - is it material?

miketheman (approved):
I'm no expert on the specific zip semantics, but looks overall pretty good.

Any ideas on improving the end-user experience, such as when they receives an error message, there's a little more content/guidance on what to do next? It's fine to come later, but having some action for folks to self-serve/debug would be helpful to diagnose issues like as reported in #18848

tests/unit/utils/test_zipfiles.py
sethmlarson: It's to see the message of the error if the test fails. The message `AssertionError: False is not True` is less helpful

miketheman: Oh, interesting! One pattern I've used with `assert` is to set the output as the secondary, so something like:

```python
assert result[0] is True, result[1]
```

sethmlarson: Done in https://github.com/pypi/warehouse/pull/18846/commits/51d815d2a7a35d5d5c943bbd04d5f98b736b9a5c

PR #18847, opened 2025-10-11
Link to the stable version of the SPDX spec

SpecLad:
Currently, the link goes to a draft. Point it to the latest stable release instead.


SpecLad:
I don't know what's going on with the build there - it's saying I should run `make translations`, but I did and there's nothing to commit. Gonna rebase just in case.

PR #18886, opened 2025-10-16
Don't permit pull_request_target for GitHub publishers

di:
(Draft, until we can determine how many users this will affect)

### DIFF
--- a/warehouse/oidc/models/github.py
+++ b/warehouse/oidc/models/github.py
@@ -153,7 +153,7 @@ def _check_sub(
 
 
 def _check_event_name(
-    ground_truth: str, signed_claim: str, _all_signed_claims: SignedClaims, **_kwawrgs
+    ground_truth: str, signed_claim: str, _all_signed_claims: SignedClaims, **_kwargs,
 ) -> bool:
     if signed_claim == "pull_request_target":
         raise InvalidPublisherError(


mohamadoaz:
في الخميس، ١٦ أكتوبر ٢٠٢٥, ٧:٥١ م Dustin Ingram ***@***.***>
كتب:

warehouse/oidc/models/github.py
```diff
@@ -152,6 +152,18 @@ def _check_sub(
     return f"{org}:{repo}".lower() == ground_truth.lower()
 
 
+def _check_event_name(
```
webknjaz, line 155: I think this should have a more specific name.
```suggestion
def _disallow_pull_request_target_events(
```

warehouse/oidc/models/github.py
```diff
@@ -152,6 +152,18 @@ def _check_sub(
...
+    ground_truth: str, signed_claim: str, _all_signed_claims: SignedClaims, **_kwawrgs
```
webknjaz:
```suggestion
    ground_truth: str, signed_claim: str, _all_signed_claims: SignedClaims, **_kwargs,
```

warehouse/oidc/models/github.py
di: This follows the convention of our other OIDC claim check functions, `_check_` + the claim name.

miketheman:
This needs to be rebased/deconflicted with the measurement code to be removed, but overall 👍

Issue #18865, opened 2025-10-14
Look up custom issuer URLs during token minting


miketheman:
With self-managed instances, the `iss` (issuer) value on the inbound JWT is going to differ based on the URL of the service.

Probably the place to add a lookup:
https://github.com/pypi/warehouse/blob/9cfbbf940af85786d479a5e2f0c486355b2ff96a/warehouse/oidc/utils.py#L69-L96

PR #18892, opened 2025-10-16
feat: look up custom issuers by inbound URL

miketheman:
When receiving an inbound JWT, handle cases where the inbound issuer is not one of the "standard" service URLs, such as in the case of GitLab Self-Managed or GitHub Enterprise Server instances.

Resolves: #18865

~~Calling `PendingGitLabPublisher.reify()` with a custom `issuer_url` has not been implemented yet.~~

### DIFF
--- a/tests/unit/oidc/models/test_gitlab.py
+++ b/tests/unit/oidc/models/test_gitlab.py
@@ -937,3 +937,13 @@ class TestPendingGitLabPublisher:
         # it is returned and the pending publisher is marked for deletion.
         assert existing_publisher == publisher
         assert pending_publisher in db_request.db.deleted
+
+    def test_reify_with_custom_issuer_url(self, db_request):
+        custom_issuer_url = "https://gitlab.custom-domain.com"
+        pending_publisher = PendingGitLabPublisherFactory.create(
+            issuer_url=custom_issuer_url
+        )
+        publisher = pending_publisher.reify(db_request.db)
+
+        assert publisher.issuer_url == custom_issuer_url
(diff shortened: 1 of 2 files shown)


warehouse/oidc/views.py
```diff
@@ -135,8 +139,16 @@ def mint_token_from_oidc(request: Request):
         )
 
     # Associate the given issuer claim with Warehouse's OIDCPublisherService.
+    # First, try the standard issuers
     service_name = OIDC_ISSUER_SERVICE_NAMES.get(unverified_issuer)
```
miketheman, line 143: If nobody ever uses a custom issuer, there's no lookup penalty, since it'll match the static dictionary value. This should be the case for most lookups.

warehouse/oidc/views.py
```diff
@@ -147,7 +159,7 @@ def mint_token_from_oidc(request: Request):
             request=request,
         )
 
-    if request.flags.disallow_oidc(OIDC_ISSUER_ADMIN_FLAGS[unverified_issuer]):
+    if request.flags.disallow_oidc(OIDC_ISSUER_ADMIN_FLAGS.get(unverified_issuer)):
```
ewdurbin, line 162: I don't understand this change. Will we be creating flags for each issuer?

ewdurbin (approved):

warehouse/oidc/views.py
miketheman: No flags per issuer, but prevents a KeyError on lookup for a custom issuer that doesn't exist in the flags.

PR #18913, opened 2025-10-21
Add documentation about archive issues

sethmlarson:
Follow-up from #18846 cc @miketheman

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -346,7 +346,7 @@ def _is_valid_dist_file(filename, filetype):
             if not zip_ok:
                 return False, (
                     f"ZIP archive not accepted: {zip_error}. "
-                    f"See {request.user_docs_url("/archives")} for more information"
+                    f"See https://docs.pypi.org/archives for more information"
                 )
 
         except zipfile.BadZipFile:  # pragma: no cover


docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
+# Archives
```
miketheman: question: would this be clearer as `Archive Formats`, or even `File Formats`, `Artifacts` or something else to distance the concept from Project Archival?
I like the short-one-name for shorter URLs, but fear it may add confusion.

docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
...
+
+Wheels and source distributions use the ZIP and tar format to distribute
+multiple files within a single artifact. To avoid archive differential / confusion
+attacks due to complexities of archive formats, PyPI rejects archives which
```
miketheman, line 5: question: is there any external links that describe more what this kind of attack is? I'm going to guess it'll be new for most folks reading this.

docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
+# Archives
+
+Wheels and source distributions use the ZIP and tar format to distribute
```
miketheman, line 3:
```suggestion
Wheels and source distributions use the ZIP and tar formats respectively to distribute
```

docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
...
+multiple files within a single artifact. To avoid archive differential / confusion
+attacks due to complexities of archive formats, PyPI rejects archives which
+use unnecessary and uncommon features of archives, such as archives that are
+designated for multiple disks or archives that are constructed to intentionally
+confuse archive implementations.
+
+This page details the archive format features that PyPI rejects so if you
```
miketheman:
```suggestion
This page details some of the the archive format features that PyPI rejects so if you
```

docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
...
+encounter an error you can debug the issue and upload the fixed archive to PyPI.
+
+## Multiple or malformed central directory
+
+Archives often support having multiple central directories
+or indexes to allow for "append-only" updates, this is not allowed in archives
```
miketheman:
```suggestion
Archives often support having multiple central directories
or indexes to allow for "append-only" updates. This is not allowed in archives
```

docs/user/archives.md
```diff
@@ -0,0 +1,59 @@
...
+
+There is two or more ZIP extra metadata field 
+with the same ID that have security relevance, such
+as marking a ZIP as ZIP64 or defining the Unicode filename.
+
+## Trailing data or comments
+
+Many archives support trailing or prepended data
+or comments within records. PyPI disallows these features
+to avoid smuggling other archive records within comments.
```
miketheman, line 59: suggestion: add a section below this like "further reading", "for more details on archive format, here's some other stuff"

warehouse/forklift/legacy.py
```diff
@@ -344,7 +344,10 @@ def _is_valid_dist_file(filename, filetype):
             # to avoid parser differentials.
             zip_ok, zip_error = zipfiles.validate_zipfile(filename)
             if not zip_ok:
-                return False, f"ZIP archive not accepted: {zip_error}"
+                return False, (
+                    f"ZIP archive not accepted: {zip_error}. "
+                    f"See https://docs.pypi.org/archives for more information"
```
miketheman, line 349: We have a helper!

```suggestion
                    f"See {request.user_docs_url("/archives")} for more information"
```

(this also actually uses the `f` part of this string)

miketheman (commented):
I like this a lot, thanks! A bunch of comments inline. Let me know what you think!

docs/user/archives.md
miketheman: Unless this is the exhaustive list?

sethmlarson: It's exhaustive at the moment.

docs/user/archives.md
sethmlarson: Not necessarily, source distributions can be ZIPs, too iirc?

docs/user/archives.md
sethmlarson: I'm not sure if it's relevant for readers of this guide? Presumably they just want to know why we rejected their upload.

docs/user/archives.md
sethmlarson: I can rename it to "Archive Formats".

docs/user/archives.md
miketheman: Not according to the spec? https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-format

I think historically the spec might have allowed that? PyPI doesn't reject them, so.... nothing to change here for now.

warehouse/forklift/legacy.py
miketheman: I take this back, `request` isn't available in this scope, so your intial approach was better 🤕

miketheman (approved):

PR #18916, opened 2025-10-21
feat: allow user to choose custom issuer for project

### LINKED ISSUE #18871: Allow configuring a new GitLab Trusted Publisher with a custom issuer
If an org has a custom issuer **already provisioned**, allow that to be selected via drop-down when configuring a new trusted, defaulting to the public service.

miketheman:
If the project is owned by an organization and there is a custom issuer
configured, populate the form with the choices available.

Resolves #18871

### DIFF
--- a/warehouse/oidc/models/gitlab.py
+++ b/warehouse/oidc/models/gitlab.py
@@ -346,19 +346,16 @@ class GitLabPublisherMixin:
             List of unique issuer URLs, with the default gitlab.com first.
             If organization is None, only returns the default issuer.
         """
-        # Start with the default issuer
-        issuer_urls = {GITLAB_OIDC_ISSUER_URL}
-
-        if organization:
-            # Get custom issuers from the organization
-            for oidc_issuer in organization.oidc_issuers:
-                if oidc_issuer.issuer_type == OIDCIssuerType.GitLab:
-                    issuer_urls.add(oidc_issuer.issuer_url)
-
-        # Return as sorted list with default first
-        result = [GITLAB_OIDC_ISSUER_URL]
-        result.extend(sorted(issuer_urls - {GITLAB_OIDC_ISSUER_URL}))
-        return result
+        issuer_urls = (
+            {
+                oidc_issuer.issuer_url
+                for oidc_issuer in organization.oidc_issuers
+                if oidc_issuer.issuer_type == OIDCIssuerType.GitLab
+            }
+            if organization
+            else set()
+        )
+        return [GITLAB_OIDC_ISSUER_URL] + sorted(issuer_urls)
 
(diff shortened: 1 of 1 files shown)


warehouse/oidc/models/gitlab.py
```diff
@@ -329,6 +332,34 @@ def admin_details(self) -> list[tuple[str, str]]:
...
+        if organization:
+            # Get custom issuers from the organization
+            for oidc_issuer in organization.oidc_issuers:
+                if oidc_issuer.issuer_type == OIDCIssuerType.GitLab:
+                    issuer_urls.add(oidc_issuer.issuer_url)
+
+        # Return as sorted list with default first
+        result = [GITLAB_OIDC_ISSUER_URL]
+        result.extend(sorted(issuer_urls - {GITLAB_OIDC_ISSUER_URL}))
+        return result
```
di: Am I missing something or can this be simplified?

```suggestion
        issuer_urls = set()

        if organization:
            # Get custom issuers from the organization
            for oidc_issuer in organization.oidc_issuers:
                if oidc_issuer.issuer_type == OIDCIssuerType.GitLab:
                    issuer_urls.add(oidc_issuer.issuer_url)

        # Return as sorted list with default first
        return [GITLAB_OIDC_ISSUER_URL] + sorted(issuer_urls)
```

warehouse/templates/manage/project/publishing.html
```diff
@@ -195,6 +195,21 @@ <h3 class="callout-block__heading">{% trans %}Proceed with caution!{% endtrans %
   </p>
   <div id="environment-errors">{{ field_errors(gitlab_publisher_form.environment) }}</div>
 </div>
+{% if gitlab_publisher_form.issuer_url.choices|length > 1 %}
```
di, line 198: Should we just always display this for all organizations, even if the only choice is the default? It seems like this would be a good place to indicate to users that other options are available, and I could see it being confusing that the issuer is not visible for a user that knows we support self-hosted gitlab but isn't sure how to do it.

miketheman: Maybe after a beta testing phase, and we have more details as to how it behaves? I prefer to not clutter a user interface with details that are not applicable to them.

warehouse/oidc/models/gitlab.py
miketheman: Thanks for the pointer, I have an even simpler approach with set comprehension incoming.

miketheman: e3d391e45d04c3048d2878ebc6dab0ffa54de5d2

di (approved):

PR #18925, opened 2025-10-23
fix(forklift): Distinguish deleted project filename error (#18671)

### LINKED ISSUE #18671: Distinguish between 'deleted' and 'live' filenames when refusing upload for duplicate filename
In #18662, @freakboy3742 said:

> At the very least, the error message should highlight that the issue with a "filename already in use" could be a deleted project.

In addition to #17942, we should also update the error message accordingly here.

---

Delta17920:
Update upload validation to check project lifecycle_status for file conflicts. Provides a more specific error message when the conflict is with an archived project. Fixes #18671

## Proof / Test Output

This ss shows the new test (`test_upload_fails_with_deleted_project_conflict`) and all other tests in the file passing after the code change:
image


di:
Thanks for the PR.

Note that whether or not the project is archived has no bearing on whether the given file has been deleted -- an archived project can have 'live' files and an active project can have 'deleted' files.

The actual error message we're trying to improve is this one: https://github.com/pypi/warehouse/blob/5f7e35c922e18378f096c6f0344bdb99e15ccdde/warehouse/forklift/legacy.py#L1145-L1159

Given that the previous checks are looking for "live" files (the `File` model) we should only arrive at this error message in the case if the filename is not 'live' but was previously used and is recorded (the `Filename` model).

So, we are already distinguishing between 'deleted' and 'live' files when returning this error message, we just need to be more clear in this error message that the file has since been 'deleted' and is no longer 'live'.

Delta17920:
on it

Delta17920:
Thanks for the review, @di ! I've pushed the update. I just changed the error message string in legacy.py and updated the matching test, as you suggested.

warehouse/forklift/legacy.py
```diff
@@ -1151,8 +1151,8 @@ def file_upload(request):
             )
             raise _exc_with_message(
                 HTTPBadRequest,
-                "This filename has already been used, use a "
-                "different version. "
+                "This filename was used by a file that has since been deleted. "
```
woodruffw: Noting as a small thing: the upload API isn't standardized, so a lot of clients sniff the error strings themselves to figure out what happened. 

(I took a look at both `twine` and `uv publish` and neither sniffs this specific string, so it's probably OK. But noting for completeness' sake 🙂)

di: I think the "File already exists" is the important one for this, but yeah, good call.

Delta17920: Thanks for the quick review and the helpful discussion, @di and @woodruffw! I'm glad to know this change doesn't impact the primary "File already exists" logic. Is there anything else you need from me for this to be approved? As this is my first contribution, I'm really looking forward to getting it merged.
Thanks again!

PR #18934, opened 2025-10-27
refactor(packaging): Use SQLAlchemy event for Filename registry (Fixes #576)

### LINKED ISSUE #576: Add filenames to registry via trigger/hook/event
Instead of adding filenames to the filename registry inline in the upload view, we should be doing this via some sort of hook or trigger. This is probably a good candidate for just using a database level trigger for.

Delta17920:
Fixes #576.

This PR refactors the logic for adding new filenames to the `file_registry` (`Filename` model).

As suggested by the issue, this logic was previously handled inline in the `file_upload` view, which is not ideal. This PR moves the logic to a SQLAlchemy event listener.

**Changes:**

1.  Removed the `request.db.add(Filename(filename=filename))` call from `warehouse/forklift/legacy.py`.
2.  Added an `@event.listens_for(File, "after_insert")` hook to `warehouse/packaging/models.py`.

This new hook automatically inserts the `File.filename` into the `file_registry` table when a new `File` is created.

To resolve the `SAWarning` about modifying a session during a flush, the hook uses a direct `connection.execute()` operation.

---
**Proof / Test Output:**

The existing tests in `tests/unit/forklift/test_legacy.py` still pass, which confirms the filename logging logic was successfully moved.

image

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -25,12 +25,12 @@ from sqlalchemy import (
     Text,
     UniqueConstraint,
     cast,
+    event,
     func,
     or_,
     orm,
     select,
     sql,
-    event,
 )
 from sqlalchemy.dialects.postgresql import (
     ARRAY,
@@ -1217,6 +1217,7 @@ class AlternateRepository(db.Model):
     url: Mapped[str]
     description: Mapped[str]
 
+
 @event.listens_for(File, "after_insert")
 def add_filename_to_registry(mapper, connection, target):
     """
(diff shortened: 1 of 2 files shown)


Delta17920:
@dstufft @di

warehouse/packaging/models.py
```diff
@@ -1215,3 +1216,20 @@ class AlternateRepository(db.Model):
...
+    This event listener is triggered *after* a new `File` object is
+    successfully inserted into the database.
+
+    We use a direct connection-level insert (`connection.execute()`)
+    instead of `session.add(Filename(...))` to avoid an `SAWarning`.
+    Modifying the session *during* the flush process (which is when
+    this hook runs) is not a supported operation. This method
+    bypasses the session's unit-of-work tracking and is safe here.
+    """
+    connection.execute(Filename.__table__.insert(), {"filename": target.filename})
```
ewdurbin, line 1235: My assumption is that this still occurs inside the wrapping transaction from [our transaction manager](https://github.com/pypi/warehouse/blob/6ff631b2486fead392e16baeeba43c564d9353a6/warehouse/tasks.py#L119), so 👍🏼

Delta17920:
@di ready to merge

PR #18960, opened 2025-10-29
Add GitHub Actions workflow for Python application

Kelleretoro:
This workflow installs Python dependencies, runs linting with flake8, and executes tests using pytest.


Kelleretoro (commented):
OK

PR #18977, opened 2025-10-31
chore(deps): replace sphinx with mkdocs

miketheman:
Review commit-by-commit for easier review.

Note: This doesn't attempt to preserve the theme, nor fix/reorganize content, rather end up with the same content and functionality.

There's another effort to reorder content by @nlhkabu , that should happen after this is merged.

Ultimately, this unblocks the conflict on dependencies seen in #18971 and convert everything to use `mkdocs` as the documentation choice.


miketheman:
Preview here: https://warehouse--18977.org.readthedocs.build/

PR #19005, opened 2025-11-05
feat: pending publishers to organizations

miketheman:
Enable an Organization to create a Pending Trusted Publisher, similar to an account-level Trusted Publisher.
The main advantage in this approach is that it skips the need to transfer the project once it's created to the organization.

### DIFF
--- a/warehouse/manage/views/organizations.py
+++ b/warehouse/manage/views/organizations.py
@@ -1857,8 +1857,8 @@ class ManageOrganizationPublishingViews:
         if publisher_already_exists:
             self.request.session.flash(
                 self.request._(
-                    "This trusted publisher has already been registered. "
-                    "Please contact PyPI's admins if this wasn't intentional."
+                    "This publisher has already been registered in your organization. "
+                    "See your existing pending publishers below."
                 ),
                 queue="error",
             )
--- a/tests/unit/manage/views/test_organizations.py
+++ b/tests/unit/manage/views/test_organizations.py
@@ -3646,8 +3646,8 @@ class TestManageOrganizationPublishingViews:
         assert result == view.default_response
         assert db_request.session.flash.calls == [
             pretend.call(
-                "This trusted publisher has already been registered. "
-                "Please contact PyPI's admins if this wasn't intentional.",
+                "This publisher has already been registered in your organization. "
+                "See your existing pending publishers below.",
                 queue="error",
             )
         ]
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -354,7 +354,6 @@ msgid "The trusted publisher could not be registered"
 msgstr ""
 
 #: warehouse/accounts/views.py:1764
-#: warehouse/manage/views/organizations.py:1860
 msgid ""
 "This trusted publisher has already been registered. Please contact PyPI's"
 " admins if this wasn't intentional."
@@ -691,6 +690,12 @@ msgstr ""
 #: warehouse/manage/views/organizations.py:1192
 #, python-brace-format
 msgid "Expired invitation for '${username}' deleted."
+msgstr ""
+
+#: warehouse/manage/views/organizations.py:1860
+msgid ""
+"This publisher has already been registered in your organization. See your"
+" existing pending publishers below."
 msgstr ""
 
 #: warehouse/oidc/forms/_core.py:38 warehouse/oidc/forms/_core.py:49


warehouse/manage/views/organizations.py
```diff
@@ -1705,3 +1720,297 @@ def transfer_organization_project(project, request):
...
+
+        if publisher_already_exists:
+            self.request.session.flash(
+                self.request._(
+                    "This trusted publisher has already been registered. "
+                    "Please contact PyPI's admins if this wasn't intentional."
+                ),
+                queue="error",
+            )
+            return self.default_response
```
di, line 1865: For non-organization publishers, we let conflicting pending publishers exist and clean them up when the 'publish' happens, why is the behavior different here?

warehouse/oidc/views.py
```diff
@@ -217,12 +218,32 @@ def mint_token(
...
+                        )
+                    )
+                else:
+                    # For user-owned projects, create normally
+                    new_project = project_service.create_project(
+                        pending_publisher.project_name,
+                        pending_publisher.added_by,
+                        request,
+                        ratelimited=False,
+                    )
```
di, line 246: I feel like this should probably become something like:
```suggestion
                # Add the project to the organization
                new_project = project_service.create_project(
                    pending_publisher.project_name,
                    pending_publisher.added_by,
                    request,
                    ratelimited=False,
                    organization_id=pending_publisher.organization_id,
                )
```
And we should handle the `creator_is_owner` difference and the `OrganizationProject` creation there instead.

warehouse/manage/views/organizations.py
di: Specifically, this would let any user "squat" any pending publishers that an organization may want to create, without recourse.

miketheman: This code mimics the per-account behavior in https://github.com/pypi/warehouse/blob/19e612847ae7daa00f86b69d7e2555502b31fd1b/warehouse/accounts/views.py#L1754-L1769

Both behave similarly in that they create the filters based on the form inputs, which are likely to differ between setups, but prevent the same account/org from registering "true duplicates".

The differences are found in the `make_existence_filters` - project name (same) + repo name (likely different) + repo owner (likely different), filename (could be the same across repos), env (also could be the same).
Any one of these being different will not trigger the "already exists" branch here.
See the test added in d4e3fa5682c5c163b99487ab033f38426af0bab2 - no code changes are necessary, and the [same cleanup behavior should be invoked during project upload/creation](https://github.com/pypi/warehouse/blob/19e612847ae7daa00f86b69d7e2555502b31fd1b/warehouse/packaging/services.py#L683-L703).

di: Good point. We don't have a way to distinguish between an org/non-org pending publisher at upload time either, so it wouldn't be possible to support this anyways.

I'd suggest changing the flash message here at least: "This trusted publisher has already been registered. Please contact PyPI's admins if this wasn't intentional." makes it seem like there's nothing else for the user to do, but they could just go remove that conflicting publisher to create a org publisher.

warehouse/oidc/views.py
miketheman: I took a look at what this might entail, and it's a little more surgery than is probably warranted to make this work, so will defer for now. 

A future refactor could be to remove or replace `creator_is_owner`, and instead pass the desired owner ID and owner role type instead of trying to infer it.

Likely includes changes to https://github.com/pypi/warehouse/blob/d9bce6a2da57570dbeeec1f1d39b0764402d27fa/warehouse/manage/views/organizations.py#L871-L891

di: Happy to do it for you if you'd prefer. I think the fact that our OIDC logic has conditionals relating to organizations is a bit of a smell here (also it seems likely that we'd like to create org projects via non-OIDC routes as well?)

warehouse/manage/views/organizations.py
miketheman: Message updated in ed9ae0f9d36a22125bb9dc44c7fffc49b1359ae6

warehouse/oidc/views.py
miketheman: If you'd like to refactor once this is out, by all means - I'd like to get this launched.
I think we have more than a couple of different project and role association creation patterns, so consolidating those is probably worthwhile, but shouldn't block this going out.

di: https://github.com/pypi/warehouse/pull/19027

PR #19012, opened 2025-11-06
feat: warn users about username vs email

miketheman:
Users open supports cases when trying to log in using their email
address, and end up opening up support cases.
Provide them instant feedback instead.


di:
Any reason we can't just support usernames _or_ emails in this field?

miketheman:
> Any reason we can't just support usernames _or_ emails in this field?

Haven't explored that, and am unlikely to do so anytime soon - but this was something I could get out relatively quickly.

di (approved):

PR #19018, opened 2025-11-07
allow uploads of wheels targeting macOS 26

btucker:
macOS 26 was released on 2025-09-15. PyPI should allow wheels targeting it.


di:
Given that Apple has [switched to year-based versioning](https://www.engadget.com/apps/ios-26-is-official-apple-changes-from-version-numbers-to-years-for-its-os-names-172129166.html), we might need to update the spec at https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#macos to account for this first.

btucker:
> Given that Apple has [switched to year-based versioning](https://www.engadget.com/apps/ios-26-is-official-apple-changes-from-version-numbers-to-years-for-its-os-names-172129166.html), we might need to update the spec at https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#macos to account for this first.

@di How are you thinking the spec should change?

di:
Right now it just says:

> They must both be positive integers, with the `x` value being `>= 10`.

It would be nice if this said something like:

> They must both be positive integers, with the `x` value being either `10 <= x <= 15`, or  `>=26` and correspond to the year of the macOS release.

btucker:
@di Sounds good. I just opened https://github.com/pypa/packaging.python.org/pull/1943

PR #19027, opened 2025-11-10
Move organizaton-specific logic out of OIDC view

di:
per https://github.com/pypi/warehouse/pull/19005#discussion_r2510998881


warehouse/packaging/interfaces.py
```diff
@@ -4,6 +4,8 @@
 
 import typing
 
+from uuid import UUID
+
```
miketheman: nitpick: Move to type checking import

Issue #18882, opened 2025-10-16
Trusted publishing: Support for SemaphoreCI


cchristous:
**What's the problem this feature will solve?**

I'd like to be able to publish to pypi from my self hosted SemaphoreCI instance using OIDC.

**Describe the solution you'd like**

SemaphoreCI already supports OIDC:
* https://docs.semaphore.io/using-semaphore/openid
* https://docs.semaphore.io/reference/openid

I believe it meets the requirements defined on https://docs.pypi.org/trusted-publishers/internals/#how-do-i-become-a-trusted-publishing-provider.

Maybe SaaS support would be a good place to start, but it doesn't benefit my use case.

**Additional context**

PR #19048, opened 2025-11-14
Add SemaphoreCI OIDC trusted provider support

cchristous:
This attempts to implement https://github.com/pypi/warehouse/issues/18882. 

The automated tests are all passing, but I don't have confidence this actually works because I don't know how to test the actual integration. I have access to and familiarity with Semaphore, and I could test using the 2 SaaS services, if the warehouse were deployed. Though, I am hoping there is an easier way to test this, so I could use some guidance on how to proceed.

### DIFF
--- a/warehouse/migrations/versions/7a97c540ed60_add_semaphoreci_oidc_models.py
+++ b/warehouse/migrations/versions/7a97c540ed60_add_semaphoreci_oidc_models.py
@@ -3,7 +3,7 @@
 Add SemaphoreCI OIDC models
 
 Revision ID: 7a97c540ed60
-Revises: a25f3d5186a9
+Revises: b8e6e0867168
 Create Date: 2025-11-04 00:00:00.000000
 """
 
@@ -13,7 +13,7 @@ from alembic import op
 from sqlalchemy.dialects import postgresql
 
 revision = "7a97c540ed60"
-down_revision = "a25f3d5186a9"
+down_revision = "b8e6e0867168"
 
 
 def upgrade():


warehouse/migrations/versions/7a97c540ed60_add_semaphoreci_oidc_models.py
```diff
@@ -0,0 +1,52 @@
...
+Create Date: 2025-11-04 00:00:00.000000
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "7a97c540ed60"
+down_revision = ("d0f67adbcb80", "daf71d83673f")
```
di:
```suggestion
Revises: 6c0f7fea7b1b
Create Date: 2025-11-04 00:00:00.000000
"""

import sqlalchemy as sa

from alembic import op
from sqlalchemy.dialects import postgresql

revision = "7a97c540ed60"
down_revision = "6c0f7fea7b1b"
```

warehouse/oidc/models/semaphore.py
```diff
@@ -0,0 +1,240 @@
...
+        # This should be unreachable since we already checked ":repo:" is in the string
+        return False
+
+    # The repo value is between :repo: and the next :
+    repo_and_rest = parts[1]
+    repo_parts = repo_and_rest.split(":", 1)
+    repo_in_sub = repo_parts[0]
+
+    if not repo_in_sub:
+        return False
```
di: I think this can be simplified a bit:
```suggestion
    # There should only be one ':repo:' in the sub:
    if signed_claim.count(":repo:") != 1:
        return False

    _, repo_and_rest = signed_claim.split(":repo:", 1)
    repo_in_sub, _ = repo_and_rest.split(":", 1)

    if not repo_in_sub:
        return False
```

warehouse/oidc/models/semaphore.py
```diff
@@ -0,0 +1,240 @@
...
+        "sub": _check_sub,
+        "repo_slug": check_claim_binary(str.__eq__),
+        "jti": check_existing_jti,
+    }
+
+    __unchecked_claims__ = {
+        "org",
+        "org_id",
+        "prj",
+        "prj_id",
```
di: I'm not familiar with SemaphoreCI but I'm surprised we wouldn't want to restrict the token based on these claims? (Even if they are duplicated in the subject?) Do you have any more details about them?

Specifically, I would assume that the `_id` claims could help us prevent resurrection attacks, unless this is not possible with SemaphoreCI?

At the very least, it would be good to check that these match what's configured in addition to checking the subject and repo_slug.

warehouse/oidc/models/semaphore.py
```diff
@@ -0,0 +1,240 @@
...
+            raise InvalidPublisherError("Missing 'repo_slug' claim")
+
+        # repo_slug format: owner/repository
+        if "/" not in repo_slug:
+            raise InvalidPublisherError(
+                f"Invalid 'repo_slug' claim format: {repo_slug!r}, "
+                "expected 'owner/repository'"
+            )
+
+        organization, project = repo_slug.split("/", 1)
```
di: Is there a reason to use `repo_slug` here rather than the `org`/`prj` claims?

warehouse/oidc/models/semaphore.py
```diff
@@ -0,0 +1,240 @@
...
+    PendingOIDCPublisher,
+    check_claim_binary,
+    check_existing_jti,
+)
+
+if typing.TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+
+SEMAPHORE_OIDC_ISSUER_URL_SUFFIX = ".semaphoreci.com"
```
di, line 30: This is interesting! Generally our model has been either to support a specific issuer URL or to support custom issuer URLs on a per-organization basis.

Does this mean that there is a different issuer for every SemaphoreCI project? Or generally, what are the expected values here?

di:
> The automated tests are all passing, but I don't have confidence this actually works because I don't know how to test the actual integration. I have access to and familiarity with Semaphore, and I could test using the 2 SaaS services, if the warehouse were deployed. Though, I am hoping there is an easier way to test this, so I could use some guidance on how to proceed.

I think the best way to test this would be to put it behind a feature flag and enable it only for test.pypi.org first.

warehouse/oidc/models/semaphore.py
cchristous: > Does this mean that there is a different issuer for every SemaphoreCI project? Or generally, what are the expected values here?

Yes, the expected values are of the format `https://<org-name>.semaphoreci.com` where org-name is the name of the Semaphore organization (not the GitHub organization). Ref https://docs.semaphore.io/reference/openid#reference.

warehouse/oidc/models/semaphore.py
cchristous: This wasn't correct/sufficient. I've updated the logic. Previously, it used git organization + repo name. The new logic uses Semaphore organization name, Semaphore project name, and repo slug (which contains Git organization + repo name)

warehouse/oidc/models/semaphore.py
cchristous: I've added org_id and prj_id to the checked claims to prevent resurrection attacks

warehouse/oidc/models/semaphore.py
cchristous: Simplified

warehouse/migrations/versions/7a97c540ed60_add_semaphoreci_oidc_models.py
```diff
@@ -0,0 +1,68 @@
...
+Create Date: 2025-11-04 00:00:00.000000
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "7a97c540ed60"
+down_revision = "6c0f7fea7b1b"
```
di:
```suggestion
Revises: a25f3d5186a9
Create Date: 2025-11-04 00:00:00.000000
"""

import sqlalchemy as sa

from alembic import op
from sqlalchemy.dialects import postgresql

revision = "7a97c540ed60"
down_revision = "a25f3d5186a9"
```

di:
@cchristous FYI you have linting errors here

cchristous:
> @cchristous FYI you have linting errors here

I thought had everything passing, but clearly not. Sorry about that. I ran `make tests` and `make lint` and fixed issues, and now they are both passing.

cchristous:
@di could you help me get this over the line?

PR #19053, opened 2025-11-14
feat: add account associations data model

### LINKED ISSUE #19052: Create data model to store account associations
Create a new table to store per-user account associations, with a reference to the user account

Store things like:

- the service name (eg `github`)
- external user id (string, since could be anything)
- external username (eg `miketheman`)
- ~~tokens from provider~~ (deferring for now, since we might not ever need them, preferring just-in-time user verifications)
- extra metadata (unstructured, as we learn more we can evolve the data model

miketheman:
Resolves #19052

### DIFF
--- a/warehouse/accounts/models.py
+++ b/warehouse/accounts/models.py
@@ -189,6 +189,13 @@ class User(SitemapMixin, HasObservers, HasObservations, HasEvents, db.Model):
         )
     )
 
+    account_associations: Mapped[list[AccountAssociation]] = orm.relationship(
+        back_populates="user",
+        cascade="all, delete-orphan",
+        lazy=True,
+        order_by="AccountAssociation.created.desc()",
+    )
+
     @property
     def primary_email(self):
         primaries = [x for x in self.emails if x.primary]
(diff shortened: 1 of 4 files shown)


miketheman:
A couple of todos:
- [x] resolve conflicts with `main`
- [x] generalize the AccountAssociations model to abstract it away, and allow for a subclass of OAuthAccountAssociations with the specifics, so we can add other types of account associations in the future

warehouse/accounts/models.py
```diff
@@ -539,3 +546,101 @@ def __repr__(self):
...
+
+    __repr__ = make_repr("service", "external_username")
+
+    id: Mapped[UUID] = mapped_column(
+        PG_UUID(as_uuid=True),
+        ForeignKey("account_associations.id", ondelete="CASCADE"),
+        primary_key=True,
+    )
+
+    service: Mapped[str] = mapped_column(
```
ewdurbin, line 631: I don't know if it is too big brained to further subclass this to GitHubOAuthAccountAssociation, GitLabOAuthAccountAssociation, GoogleOAuthAccountAssociation etc rather than map to a string?

I doubt they will ever have dramatically different implementations, but it would avoid "just a string" headaches down the line.

Perhaps an enum is a good middle ground in the short term, so that we at least have a definitive list of known values.

ewdurbin (approved):

warehouse/accounts/models.py
miketheman: I think the instinct is correct, considering we did something similar to allow for the OIDC implementations.
In those cases, they are discrete enough and have their own extensive behaviors, having the concrete classes makes sense. Since in this phase of the OAuth associations we're currently only storing the external user_id and username, there's not much else specific.

If we find a need for more concrete classes, we would likely add another class (but not associated underlying database table) to refer to these as distinct types.

PR #19055, opened 2025-11-14
fix: update all journals in bulk

### LINKED ISSUE #19033: User account cannot be deleted: 503 error
When trying to delete a user account, [pypi.org](https://pypi.org/) fails with a "Sorry, something went wrong" screen. The HTTP response is 503 Service Unavailable.

The user account in question (`girder-robot`) appears to have no associated entities (no owned packages, no organizations, no API tokens, etc.).

I have tried repeatedly over the course of several days (Nov 7, 2025 - Nov 11, 2025) to perform the deletion.

---

This may be related to #16073 / #16439 , but appears to have slightly different details than those issues.

miketheman:
Issues with the previous code:

- all journal objects were loaded into memory, bloating the process for
  accounts with high volumes, only to be updated, and not used otherwise
- once loaded, iterating though each to update is a classic n+1
  performance issue, and in some cases hitting a 60 second timeout

With this code:

- perform a single SQL query to update all relevant records
- doesn't load any journal objects into memory
- since nothing is in memory, we don't need to synchronize with
  sqlalchemy session map, making this even more efficient
- no longer needs a manual flush

Fixes #19033
Fixes WAREHOUSE-PRODUCTION-2BN
Fixes WAREHOUSE-PRODUCTION-28N

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -164,7 +164,7 @@ msgstr ""
 msgid "Successful WebAuthn assertion"
 msgstr ""
 
-#: warehouse/accounts/views.py:725 warehouse/manage/views/__init__.py:862
+#: warehouse/accounts/views.py:725 warehouse/manage/views/__init__.py:855
 msgid "Recovery code accepted. The supplied code cannot be used again."
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


di (approved):

miketheman:
Test that currently covers the user-facing manage behavior:
https://github.com/pypi/warehouse/blob/8c8ff104824e04efe7b3e75b443ec5fb0a7bd1b2/tests/unit/manage/test_views.py#L924-L931
And the one for admin: https://github.com/pypi/warehouse/blob/fd5b2defcbd03588bdc123e5eeb62837ccec2a0a/tests/unit/admin/views/test_users.py#L411-L424

Helpful docs: https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#selecting-a-synchronization-strategy

PR #19120, opened 2025-12-02
load donate banner from donate.python.org

ewdurbin:
This is no-op until we go live with the fundraiser. can be previewd using `?banner_preview=true`


corneliusroemer:
In the future, it might be helpful to include a screenshot of what the banner looks like to give reviewers a chance to see it directly on the PR without having to follow instructions.

PR #19124, opened 2025-12-02
Data migration for `UserUniqueLogin`/`IpAddress`

di:
Follow-on to #19123. This creates `IpAddress` rows where missing and backfills `UserUniqueLogin.ip_address_id` where missing.

Batching because:
```
warehouse=> select count(*) from user_unique_logins where ip_address_id is NULL;
 count
-------
 19078
(1 row)
```

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -133,9 +133,9 @@ pip-api==0.0.34 \
     --hash=sha256:8b2d7d7c37f2447373aa2cf8b1f60a2f2b27a84e1e9e0294a3f6ef10eb3ba6bb \
     --hash=sha256:9b75e958f14c5a2614bae415f2adf7eeb54d50a2cfbe7e24fd4826471bac3625
     # via -r requirements/dev.in
-pip-tools==7.5.1 \
-    --hash=sha256:a051a94794ba52df9acad2d7c9b0b09ae001617db458a543f8287fea7b89c2cf \
-    --hash=sha256:f5ff803823529edc0e6e40c86b1aa7da7266fb1078093c8beea4e5b77877036a
+pip-tools==7.5.2 \
+    --hash=sha256:2d64d72da6a044da1110257d333960563d7a4743637e8617dd2610ae7b82d60f \
+    --hash=sha256:2fe16db727bbe5bf28765aeb581e792e61be51fc275545ef6725374ad720a1ce
     # via -r requirements/dev.in
 plaster==1.1.2 \
     --hash=sha256:42992ab1f4865f1278e2ad740e8ad145683bb4022e03534265528f0c23c0df2d \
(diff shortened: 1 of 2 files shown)


warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
```diff
@@ -0,0 +1,101 @@
...
+                        FROM user_unique_logins
+                        WHERE ip_address_id IS NULL
+                        ORDER BY id
+                        LIMIT :batch_size
+                    )
+                """
+            ),
+            {"batch_size": BATCH_SIZE},
+        )
+        bind.commit()
```
ewdurbin, line 101: Does this close out the transaction wrapping transaction that we are in during the migration? If so, cool! I think that means that this migration can be run repeatedly if it timeout... which is nice.

warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
```diff
@@ -0,0 +1,101 @@
...
+            """
+        )
+    ).scalar_one()
+
+
+def upgrade():
+    op.execute("SET statement_timeout = 120000")
+    op.execute("SET lock_timeout = 120000")
+
+    op.create_index(
```
ewdurbin, line 44: related to question about commit, if we make this "CREATE INDEX IF NOT EXISTS" I think we can pretty confidently ship this and it will eventually complete even if it takes longer than the timeouts.

warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+
+    bind = op.get_bind()
+    BATCH_SIZE = 1000
+
+    while _get_remaining_ips_to_insert(bind) > 0:
+        bind.execute(
+            sa.text(
+                """
+                INSERT INTO ip_addresses (ip_address)
```
ewdurbin: We have WAREHOUSE_IP_SALT in the environment, so we could calculate the hashed_ip_address at the same time. `sha256(CONCAT(ip_address, "WAREHOUSE_IP_SALT")`

This may also surface addresses that already exist in the table.

di: `IpAddress.ip_address` is non-nullable so I don't think there are any missing IPs that exist as only hashes? (or am I missing something?).

I'll add the hashes.

warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
di: Yeah I think so?!

warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
di: 6055695da41f9e34d2bedcd1c812178b2d04e329

warehouse/migrations/versions/be443e514e3e_data_migration_for_useruniquelogin_ip_.py
di: 5696067659b12995474a02a0639fa7a4c55a7638

PR #19188, opened 2025-12-17
chore(deps): update Python version to 3.13.11

miketheman:


sethmlarson:
So the "failure" here is that prior to 3.13.11 (specifically https://github.com/python/cpython/commit/333d4a6f4967d3ace91492a39ededbcf3faa76a6) CPython didn't understand ZIP64 EOCD extensible data at all. Now it does, so the zipfile module can actually process these ZIP archives now. But this feature isn't useful for Python packages so I'm inclined to still reject archives with ZIP64 EOCD extensible data explicitly.

miketheman:
> But this feature isn't useful for Python packages so I'm inclined to still reject archives with ZIP64 EOCD extensible data explicitly.

Thanks for boiling it down - I agree, we should definitely reject "exotic" distributions if they don't conform to how we expect them to behave.

Issue #19103, opened 2025-11-26
Create routes, views for creating account associations


miketheman:
Create the routes, views, forms, and templates for associating an account with an external OAuth provider, starting with GitHub as a first.

Use GitHub's OAuth flow to ensure the user has authenticated on GitHub and return their user id and username back to warehouse, storing an association.
Allow the user to also remove their existing association.

Should also send an email, like other security-related operations, letting the account holder know this has happened, and add events to their user activity.

miketheman:
GitHub account associations are working well so far, now on to GitLab associations.

miketheman:
I'm going to rename this for GitHub-scope and close out, and open a fresh one to track GitLab and others.

PR #19197, opened 2025-12-18
feat: associate a PyPI account with a GitHub account

miketheman:
Currently only implemented for GitHub, will update UIs once another OAuth provider is added.
Not used yet for anything in recovery flow, surfaces the data in Admin UI for human review.

Still needs:

- [x] send user email on addition/deletion
- [x] a real GitHub App created on `pypi` and config set in cabotage

Refs #19103

### DIFF
--- a/warehouse/accounts/oauth.py
+++ b/warehouse/accounts/oauth.py
@@ -10,12 +10,15 @@ from __future__ import annotations
 
 import secrets
 import urllib.parse
+import warnings
 
 from typing import TYPE_CHECKING, Self
 
 import requests
 
 from zope.interface import Interface, implementer
+
+from warehouse.utils.exceptions import NullOAuthProviderServiceWarning
 
 if TYPE_CHECKING:
     from pyramid.request import Request
@@ -177,7 +180,11 @@ class GitHubAppClient:
 @implementer(IOAuthProviderService)
 class NullOAuthClient:
     """
-    Null OAuth client for testing and development.
+    Null OAuth client for testing and development ONLY.
+
+    WARNING: This service should NEVER be used in production environments.
+    It creates fake user associations without any actual OAuth verification,
+    which would allow anyone to associate arbitrary external accounts.
 
     Returns mock data without making real OAuth requests.
     Useful for local development without configuring OAuth apps.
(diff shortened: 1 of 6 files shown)


warehouse/config.py
```diff
@@ -467,6 +467,7 @@ def configure(settings=None):
     maybe_set_compound(settings, "breached_emails", "backend", "BREACHED_EMAILS")
     maybe_set_compound(settings, "breached_passwords", "backend", "BREACHED_PASSWORDS")
     maybe_set_compound(settings, "domain_status", "backend", "DOMAIN_STATUS_BACKEND")
+    maybe_set_compound(settings, "github.oauth", "backend", "GITHUB_OAUTH_BACKEND")
```
ewdurbin, line 470: namespace these if we can? `account_association.github.oauth`... just for future

warehouse/accounts/oauth.py
```diff
@@ -0,0 +1,243 @@
...
+@implementer(IOAuthProviderService)
+class NullOAuthClient:
+    """
+    Null OAuth client for testing and development.
+
+    Returns mock data without making real OAuth requests.
+    Useful for local development without configuring OAuth apps.
+    """
+
+    def __init__(
```
ewdurbin, line 193: could we potentially warn (or even error) if this is attempted to be used outside of development? my understanding is if we deployed this and forgot configuration it would lead to littering mockuser_NNNNNN in test/prod dbs

ewdurbin (approved):
Overall very nice! Just some simple thoughts about orderliness inline below.

Questions:

Can you detail the minimal configuration necessary for the GitHub Application we will need to configure? Specifics like "Expire user authorization tokens", do we need webhooks (I doubt it), and any specific Permission grants would be great!

warehouse/config.py
miketheman: Is namespacing required? I envision us using this connection in the future to offer a better Trusted Publishing onboarding where the user can pick a repo from a drop down

warehouse/accounts/oauth.py
miketheman: Definitely a good idea. I think an even better approach would be to not allow the "maybe" aspect of configuration and require the setting, and not fall back to anything.

And in this service, implement a warning like we do with some other Null services:

https://github.com/pypi/warehouse/blob/36a581985da26cbff4274b7b96deff95fb327a5f/warehouse/oidc/services.py#L41-L46

miketheman: Updated in 516cee59

miketheman:
> Can you detail the minimal configuration necessary for the GitHub Application we will need to configure? Specifics like "Expire user authorization tokens", do we need webhooks (I doubt it), and any specific Permission grants would be great!

Settings:

- **Name**: PyPI
- **Homepage URL**: `https://pypi.org`
   - **Callback URL**: `https://pypi.org/manage/account/associations/github/callback`
   - **Webhook**: Uncheck "Active" (not needed initially)
   - **Permissions** (Repository):
     - `Metadata`: Read (for future repository listing)
   - **Where can this app be installed?**: Any account

PR #19203, opened 2025-12-19
chore(deps): update JavaScript dependencies

miketheman:
- Update everything within an `npm update` scope per rules
- Selectively update test related utils (jest & friends)
- Update Node CI version to match Dockerfile


miketheman:
Dropped from 1012 packages to 996 packages total.

PR #19213, opened 2025-12-22
feat: Add IP address banning and unbanning functionality

di:


tests/unit/admin/views/test_ipaddresses.py
```diff
@@ -75,3 +77,71 @@ def test_ip_address_found_with_unique_logins(self, db_request):
...
+        resp = ip_views.ban_ip(db_request)
+
+        assert isinstance(resp, HTTPSeeOther)
+        assert resp.location == f"/admin/ip-addresses/{ip_address.ip_address}"
+        assert ip_address.is_banned
+        assert ip_address.ban_reason == BanReason.ADMINISTRATIVE
+        assert ip_address.ban_date is not None
+
+
+class TestUnbanIpAddress:
```
miketheman, line 142: question: worth adding a test case and handling for unbanning an already-unbanned entry, or are we fine with doing the unban operation regardless of current state?

warehouse/admin/templates/admin/ip_addresses/detail.html
```diff
@@ -38,9 +38,24 @@ <h3 class="card-title"><code>IpAddress</code> Record</h3>
       <dd class="col-sm-8">{{ ip_address.ban_date }}</dd>
 
       <dt class="col-sm-4">Ban Reason:</dt>
-      <dd class="col-sm-8">{{ ip_address.ban_reason.value }}</dd>
+      <dd class="col-sm-8">{{ ip_address.ban_reason.value if ip_address.ban_reason is not none else 'None' }}</dd>
```
miketheman, line 41: I was reading more about [Jinja2 filters](https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.default) recently, and I think this could also be expressed as:

```suggestion
      <dd class="col-sm-8">{{ ip_address.ban_reason.value | default('None') }}</dd>
```
but that means that the `.value` call would have failed a lookup on None.

Maybe we should instead change the BanReason from an Enum to a StrEnum, like we do in other places, and thus don't need to refer to it by .value?

A larger refactor, but probably worthwhile to consider as a follow-up.

tests/unit/admin/views/test_ipaddresses.py
di: f50b905

warehouse/admin/templates/admin/ip_addresses/detail.html
di: Agreed that this should be a follow-up.

Issue #10718, opened 2022-02-10
Explicitly typed in project name is not the first in relevance order


GeorgeFischhof:
**Describe the bug**

I checked my package, typed in the exact package name into search box: pluggable-info-monitor and the first 3 pages of the results do not contain my project. (I did not check all the 500 pages...
The project exist on this url:
https://pypi.org/project/pluggable-info-monitor/

**Expected behavior**
The project is listed at first position, because default relevance order is used

**To Reproduce**
type in the search box:
pluggable-info-monitor

on other tab / window go to url:
https://pypi.org/project/pluggable-info-monitor/

**My Platform**
Windows 8.1 and Windows 10, Firefox latest version used

BR,
George

dil-gfischhof:
It is good :) 
In the meantime I was searching for another package: selfupdate, and searched with "self-update" this dashed phrase is similar: if I write with dash, the package without dash is not shown on first some pages

miketheman:
In attempts to reproduce this behavior in the development environment, I wasn't able to, as we don't have the same packages, and I struggled a bit to try and create the same-named, so I found [`zero-downtime-migrations`](https://pypi.org/project/zero-downtime-migrations/).

Searching for a double-hyphenated package produces the desired results: https://pypi.org/search/?q=zero-downtime-migrations&o= We can see other packages like `zero`, `django-downtime`, and `migrations` later on on the results, but the most matched one surfaces first.

Here's the JSON payload we generate and submit to the ES service:

<details>

```json
{
  "query": {
    "bool": {
      "should": [
        {
          "bool": {
            "must": [
              {
                "multi_match": {
                  "fields": [
                    "author",
                    "author_email",
                    "description^5",
                    "download_url",
                    "home_page",
                    "keywords^5",
                    "license",
                    "maintainer",
                    "maintainer_email",
                    "normalized_name^10",
                    "platform",
                    "summary^5"
                  ],
                  "query": "zero-downtime-migrations",
                  "type": "best_fields"
                }
              }
            ]
          }
        },
        {
          "prefix": {
            "normalized_name": "zero-downtime-migrations"
          }
        }
      ]
    }
  },
  "suggest": {
    "name_suggestion": {
      "text": "zero-downtime-migrations",
      "term": {
        "field": "name"
      }
    }
  }
}
```

</details>

Notable is that we're using the `normalized_name` value in the search, which is boosted by 10x than others, so it makes sense that. match on this field would produce desired results.


Results (lengthy):

<details>

[...]

PR #19224, opened 2025-12-27
fix: boost full name search results

miketheman:
To enable matching the full input name before any tokenized or split
versions, add a `should` clause to the search queries.


miketheman:
Should resolve:

- https://github.com/pypi/warehouse/issues/2877
- https://github.com/pypi/warehouse/issues/10718

Need to confirm that the use cases described are resolved once deployed.

Isaac9710009 (approved):

JacobCoffee (approved):

Issue #19222, opened 2025-12-25
Login fails after 2FA: “Session invalidated” shown and user remains logged out (303 redirect from /account/two-factor)


chigwell:
**Describe the bug**
When logging in on PyPI via `https://pypi.org/account/login/`, the login flow appears to succeed through username/password and the second factor code entry, but the session is immediately invalidated. After submitting the 2FA code, I am redirected to a page that still shows **“Log in”** and **“Register”** (no account/menu visible), with a banner message at the top: **“Session invalidated”**. In the browser network tab, the request to `https://pypi.org/account/two-factor` returns **HTTP 303**.

**Expected behavior**
After successfully entering username/password and a valid 2FA code, I should be logged in and see my account/session (e.g., account menu / profile link), without any “Session invalidated” message.

**To Reproduce**

1. Go to `https://pypi.org/account/login/`
2. Enter PyPI username (nickname) and password, submit
3. Enter the 2FA authentication code, submit
4. Observe that the page still shows “Log in” / “Register” and a banner “Session invalidated”
5. Open DevTools → Network: request to `https://pypi.org/account/two-factor` shows **303** status

**My Platform**

* Browser: Google Chrome Version 143.0.7499.170 (Official Build) (arm64)

**Additional context**

* The visible UI after completing 2FA looks like an unauthenticated state (no account information shown).
* Network: `https://pypi.org/account/two-factor` returns **303** during the 2FA step.
* Nickname: `eugene.evstafev`

LiwaaCoder:
I've been looking into the `warehouse` source code regarding the 2FA flow, and I believe I found a logic error that might explain why the session is invalidated during the redirect.

In `warehouse/accounts/views.py`, the `login` view correctly captures the headers returned by `_login_user` and passes them to the response:

View: login (around line 335)
headers = _login_user(request, userid, ...)
 ...
return HTTPSeeOther(redirect_to, headers=dict(headers)) # Correct: Headers are passed

However, in the two_factor_and_totp_validate view, _login_user is called, but the returned headers (which contain the authentication cookies) seem to be ignored/dropped before the redirect:

LiwaaCoder:
If _login_user invalidates the old session and generates a new one, failing to pass these headers in the HTTPSeeOther response means the browser never receives the Set-Cookie header for the new session. When the browser follows the 303 redirect to the dashboard, it arrives unauthenticated.

It seems Chrome 143 might be stricter about how it handles sessions during redirects, exposing this missing header issue.

[...]

PR #19223, opened 2025-12-25
Fix: Pass auth headers in 2FA and recovery redirects

LiwaaCoder:
Summary: In two_factor_and_totp_validate and recovery_code, the headers returned by _login_user (which contain the new session/auth cookies) were being ignored. This caused the session to be invalidated immediately after the HTTP 303 redirect.

Fix: captured the headers from _login_user and passed them to HTTPSeeOther.

Related Issue: "Fixes #19222"

PR #19243, opened 2026-01-04
fix: pass auth headers in 2FA and recovery redirects

chigwell:
Closes: https://github.com/pypi/warehouse/issues/19222


di:
This appears to just be a duplicate of #19223.

PR #19250, opened 2026-01-05
remove donate.python.org banner loading code

ewdurbin:


ewdurbin:
Will merge on confirmation from fundraising team!

PR #19251, opened 2026-01-05
host fonts ourselves

ewdurbin:
removes a few more CSP configurations, reduces tracking surface for users


warehouse/static/js/vendor/fonts-ewert.js
```diff
@@ -0,0 +1,4 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+// Ewert - Display font for 404 page only
+import "@fontsource/ewert/400.css";
```
miketheman: Dart Sass will break in the future if we don't use `@use` instead of `@import`, and emits build-time warnings.

See https://sass-lang.com/documentation/breaking-changes/import/

warehouse/static/js/vendor/fonts.js
```diff
@@ -0,0 +1,17 @@
...
+import "@fontsource/source-sans-3/400.css";
+import "@fontsource/source-sans-3/400-italic.css";
+import "@fontsource/source-sans-3/600.css";
+import "@fontsource/source-sans-3/600-italic.css";
+import "@fontsource/source-sans-3/700.css";
+import "@fontsource/source-sans-3/700-italic.css";
+
+// Source Code Pro - Monospace/code font
+// Weight: 500
+import "@fontsource/source-code-pro/500.css";
```
miketheman: Same note about `@use`

webpack.config.js
```diff
@@ -146,6 +146,10 @@ module.exports = [
 
       /* Vendor Stuff */
       fontawesome: "./warehouse/static/sass/vendor/fontawesome.scss",
+
+      /* Self-hosted fonts via Fontsource */
+      fonts: "./warehouse/static/js/vendor/fonts.js",
+      "fonts-ewert": "./warehouse/static/js/vendor/fonts-ewert.js",
```
miketheman: It's weird to me  to see `.js` extensions for `.scss`-style code - any reason instead of `warehouse/static/sass/vendor/*.scss`?

ewdurbin: because I do not know how to web pack :) resolving.

ewdurbin: b597f608e

warehouse/static/js/vendor/fonts.js
ewdurbin: b597f608e

warehouse/static/js/vendor/fonts-ewert.js
ewdurbin: b597f608e

webpack.config.js
```diff
@@ -182,7 +186,14 @@ module.exports = [
                     },
                   },
                 },
-                "sass-loader",
+                {
+                  loader: "sass-loader",
+                  options: {
+                    sassOptions: {
+                      loadPaths: [path.resolve(__dirname, "node_modules")],
```
miketheman, line 193: This inclusion of explicit paths feels wrong, but I can't explain why. If it works, it works, and we can revisit webpack configs at a later time.

Per https://fontsource.org/docs/getting-started/sass-import it looks like could be simpler, but I'm not certain how right now.

webpack.config.js
```diff
@@ -146,6 +146,10 @@ module.exports = [
 
       /* Vendor Stuff */
       fontawesome: "./warehouse/static/sass/vendor/fontawesome.scss",
+
+      /* Self-hosted fonts via Fontsource */
+      fonts: "./warehouse/static/sass/vendor/fonts.scss",
+      "fonts-ewert": "./warehouse/static/sass/vendor/fonts-ewert.scss",
```
miketheman, line 152: Nice splitting them so we only incur ewert loading on 404 pages

webpack.config.js
ewdurbin: lemme try again without, I moved some things around

ewdurbin: hmmm, that `pkg:` style syntax was causing a lot of headaches to move to... but adding the `~` prefix like font awesome uses seems incompatible with 'use'.

outta my depth on the webpack/sass/scss chain here....

PR #19261, opened 2026-01-06
Don't return empty headers from `_login_user`

di:
Our usage of [`remember`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/security.html#pyramid.security.remember) with session-based auth does not actually set any headers, the result will always be an empty list, so passing the empty list through to the response/redirect is unnecessary.

This is an alternative to #19223, which is unnecessary, to reduce confusion here in the future.

Our usage of `remember` never calls it with the additional, optional kwargs to set as headers:
https://github.com/pypi/warehouse/blob/c3df5911a46f64c41710f2e7eb5d2c4e0a6f26ca/warehouse/accounts/views.py#L1524-L1525

And our security policy uses `SessionAuthenticationHelper`: https://github.com/pypi/warehouse/blob/c3df5911a46f64c41710f2e7eb5d2c4e0a6f26ca/warehouse/accounts/security_policy.py#L100-L101

which always returns an empty list:  https://github.com/Pylons/pyramid/blob/eb0e01a0fd7aff84c345b1c43bd4fa9d2fe04441/src/pyramid/authentication.py#L1265-L1268

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -174,171 +174,171 @@ msgid ""
 "#admin-intervention for details."
 msgstr ""
 
-#: warehouse/accounts/views.py:899
+#: warehouse/accounts/views.py:898
 msgid "Expired token: request a new password reset link"
 msgstr ""
 
-#: warehouse/accounts/views.py:901
+#: warehouse/accounts/views.py:900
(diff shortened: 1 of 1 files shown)


miketheman (approved):
Seems reasonable, especially considering that the upstream choice to return an empty list always.

PR #19267, opened 2026-01-07
chore: update Python version to 3.14.2

miketheman:
Refs: https://www.python.org/downloads/release/python-3142/


ewdurbin (approved):
Dunno if it really matters one way or another, but I have been:

```
$ touch requirements/*.in
$ make requirements/*.txt
```

Just to keep the files "fresh". In this case it removes lots of hashes! Not sure why!

miketheman:
> Dunno if it really matters one way or another, but I have been:
> 
> ```
> $ touch requirements/*.in
> $ make requirements/*.txt
> ```
> 
> Just to keep the files "fresh". In this case it removes lots of hashes! Not sure why!

Good idea, thanks! Done in d1ecede

PR #19282, opened 2026-01-08
fix: prevent login redirect to CSI include endpoints

JacobCoffee:
when logging in, sporadically, i would land on a CSS_less page:
image

the login link's next parameter was using the csi endpoints url instead of the actual page url

```
The main_navigation_logged_out macro was using request.url for the login link's "next" parameter. When rendered via CSI (client-side include), request.url was the CSI endpoint URL (/_includes/authed/current-user-indicator/) rather than the actual page the user was viewing.

This caused users to be redirected to raw HTML fragments after login, resulting in a CSS-less page in quirks mode.

Fix: Add optional next_url parameter to the macro. The CSI include template calls it without args (no redirect), while the fallback content passes the correct request.url.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
```

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -1072,14 +1072,14 @@ msgstr ""
 
 #: warehouse/templates/404.html:33 warehouse/templates/500.html:18
 #: warehouse/templates/accounts/two-factor.html:35
-#: warehouse/templates/base.html:345 warehouse/templates/base.html:351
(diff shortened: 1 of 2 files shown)


JacobCoffee:
i had an unrelated, new file `warehouse/templates/manage/organization/analytics.html` that was soaked up in the translation gen that made it fail 🙃

PR #19283, opened 2026-01-08
fix: set a statement timeout on uploads

miketheman:
Prevent upload exhaustion by setting a maximum timeout. When uploads take a while (for whatever reason - slowdowns on storage or database lock acquisition), the connection may stall indefinitely waiting for locks to be resolved.

Other web endpoints have a natural 60 second cutoff through other machinery, uploads have no such guarantees.

10 minutes is above the maximum observed latency from the past 90 days.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -30,7 +30,7 @@ from pyramid.httpexceptions import (
 )
 from pyramid.request import Request
 from pyramid.view import view_config
-from sqlalchemy import and_, exists, func, orm
+from sqlalchemy import and_, exists, func, orm, text
 from sqlalchemy.exc import MultipleResultsFound, NoResultFound
 
 from warehouse.admin.flags import AdminFlagValue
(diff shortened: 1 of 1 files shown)


warehouse/forklift/legacy.py
```diff
@@ -664,6 +664,14 @@ def file_upload(request):
         request.metrics.increment("warehouse.upload.failed", tags=["reason:no-file"])
         raise _exc_with_message(HTTPBadRequest, "Upload payload does not have a file.")
 
+    # Set a statement timeout for this connection to prevent long-running
+    # transactions from holding database locks indefinitely. The upload
+    # operation holds a global advisory lock for journal monotonicity,
+    # so limiting transaction duration prevents lock pile-up.
+    # 600 seconds (10 minutes) allows for large file uploads while
+    # providing an upper bound on lock duration.
+    request.db.execute(text("SET statement_timeout = '600s'"))
```
JacobCoffee, line 673: would it be better to do this in https://github.com/pypi/warehouse/blob/e486f615e2931816d645635baf3145d8d41d6735/warehouse/db.py#L187-L193 or do we have things that run over 10 mins elsewhere on purpose

JacobCoffee (approved):

warehouse/forklift/legacy.py
miketheman: We have background tasks that may run longer - specifically the full search reindex

PR #19298, opened 2026-01-12
feat: associate gitlab accounts to pypi accounts

miketheman:
Resolves: #19350


miketheman:
Blocked on getting new Application (`pypi`, `testpypi`) configured on the appropriate GitLab Group

miketheman:
To unblock this code and merge, I updated the logic to only activate when we set the environment variable, so we can merge it and apply the environment variable when we have the GitLab application set up.

PR #19308, opened 2026-01-13
Validate admin form email field

JacobCoffee:
- Validates on frontend and backend
- Moves `no_deliverability_check` fixture into public namespace to be used in other modules

Fixes https://python-software-foundation.sentry.io/issues/7187418804

### DIFF
--- a/tests/unit/admin/views/test_users.py
+++ b/tests/unit/admin/views/test_users.py
@@ -834,7 +834,7 @@ class TestUserRecoverAccountInitiate:
         assert account_recovery.additional == {"status": "initiated"}
 
     def test_user_recover_account_initiate_override_email(
-        self, db_request, db_session, monkeypatch, no_deliverability_check
+        self, db_request, db_session, monkeypatch, _no_deliverability_check
     ):
         admin_user = UserFactory.create()
         user = UserFactory.create(
@@ -905,7 +905,7 @@ class TestUserRecoverAccountInitiate:
         assert account_recovery.additional == {"status": "initiated"}
 
     def test_user_recover_account_initiate_override_email_exists(
-        self, db_request, db_session, monkeypatch, no_deliverability_check
+        self, db_request, db_session, monkeypatch, _no_deliverability_check
     ):
         admin_user = UserFactory.create()
         user = UserFactory.create(
@@ -979,7 +979,7 @@ class TestUserRecoverAccountInitiate:
         assert account_recovery.additional == {"status": "initiated"}
 
     def test_user_recover_account_initiate_override_email_exists_wrong_user(
-        self, db_request, db_session, monkeypatch, no_deliverability_check
+        self, db_request, db_session, monkeypatch, _no_deliverability_check
     ):
         admin_user = UserFactory.create()
         user = UserFactory.create(
--- a/tests/unit/admin/views/test_users.py
+++ b/tests/unit/admin/views/test_users.py
@@ -1057,7 +1057,9 @@ class TestUserRecoverAccountInitiate:
         db_request.method = "POST"
         db_request.user = admin_user
         db_request.POST["project_name"] = ""
-        db_request.POST["support_issue_link"] = "https://github.com/pypi/support/issues/1"
+        db_request.POST["support_issue_link"] = (
+            "https://github.com/pypi/support/issues/1"
+        )
         db_request.POST["override_to_email"] = "ivalid-email"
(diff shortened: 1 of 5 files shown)


warehouse/admin/views/users.py
```diff
@@ -516,6 +517,19 @@ def user_recover_account_initiate(user, request):
             )
 
             if override_to_email is not None:
+                try:
+                    email_validator.validate_email(
+                        override_to_email, check_deliverability=False
```
miketheman: Might actually make sense to flip this to True here, since we actually want to ensure we can deliver to the desired address

tests/unit/admin/views/test_users.py
```diff
@@ -1035,6 +1035,52 @@ def test_user_recover_account_initiate_override_email_exists_wrong_user(
...
+        user = UserFactory.create(
+            totp_secret=b"aaaaabbbbbcccccddddd",
+            webauthn=[
+                WebAuthn(
+                    label="fake", credential_id="fake", public_key="extremely fake"
+                )
+            ],
+            recovery_codes=[
+                RecoveryCode(code="fake"),
+            ],
```
miketheman: question: were these necessary for the test to pass? The [base UserFactory sets a `totp_secret`](https://github.com/pypi/warehouse/blob/7678eadfefaa68ef38c26f6f33a31becf0637fc2/tests/common/db/accounts.py#L80), but not the other two.

miketheman (changes requested):
Nice stuff - one change requested, one question

tests/unit/admin/views/test_users.py
JacobCoffee: i dont think so,, i just copied the pattern from existing tests

warehouse/admin/views/users.py
```diff
@@ -516,6 +517,19 @@ def user_recover_account_initiate(user, request):
...
+                    )
+                except email_validator.EmailNotValidError:
+                    request.session.flash("Invalid email address format", queue="error")
```
miketheman: If we _do_ check deliverability, then the message might need to change - since the exception covers both undeliverable and bad syntax

tests/unit/admin/views/test_users.py
miketheman: Cool - probably good to remove if not needed.

warehouse/admin/views/users.py
JacobCoffee: [7dbdb70](https://github.com/pypi/warehouse/pull/19308/commits/7dbdb7077ab1fea03b8e7fd0f4782c1d53fa5aab)

tests/unit/admin/views/test_users.py
JacobCoffee: [b6e8a11](https://github.com/pypi/warehouse/pull/19308/commits/b6e8a115468879ebff3300b83c01082e3ba4f431)

warehouse/admin/views/users.py
JacobCoffee: [d249bd8](https://github.com/pypi/warehouse/pull/19308/commits/d249bd8c35c94df622e86391df65ce8959db549d)

JacobCoffee:
i dont like that [PT004](https://github.com/pypi/warehouse/actions/runs/20976270317/job/60291667208?pr=19308) recommends moving a function to private if it doesn't return. that seems wrong? anyway, [28cfef9](https://github.com/pypi/warehouse/pull/19308/commits/28cfef9ee3522862036da710afbb61acd15ed872) fixes that to conform

JacobCoffee:
Merged `main` into this branch to clear a conflict in `tests/conftest.py`.

Two things had changed underneath it:

- Main landed its own `no_email_deliverability_check` fixture, which does the same job as this branch's `_no_deliverability_check`. Rather than keep both, I dropped this branch's copy and repointed its three call sites in `tests/unit/admin/views/test_users.py` at main's fixture.
- Main deleted the `_event` helper, so I took that deletion.

`tests/conftest.py` therefore drops out of this PR entirely. The net diff against main is now just the email validation change itself — the `type="email"` input, the backend deliverability check in `user_recover_account_initiate`, and its tests.

_🤖 Done via scheduled babysitting job._

miketheman (approved):
Seems like this ought to work.

We should start to think about how we extract more functionality from views for reuse, since we already have a `NewEmailMixin` designed for forms, but admin views don't always forms, but that's a much bigger architectural change.

Issue #13888, opened 2023-06-07
Trusted publishing: Support for CircleCI


di:
Per https://circleci.com/docs/api/v2/index.html#tag/OIDC-Token-Management, CircleCI now supports a customizable `aud` claim which means we could support it as a trusted publisher.

This is likely blocked on https://github.com/pypi/warehouse/issues/13887 given that OIDC tokens from CircleCI might be intentionally shared across multiple third party providers.

dstufft:
It also sounds like we shouldn't support CircleCI currently?

Even if we resolve https://github.com/pypi/warehouse/issues/13887 so that we verify that `aud` is *only* our audiences, from https://github.com/sigstore/fulcio/issues/591#issuecomment-1579417445 it sounds like CircleCI only lets you have a single OIDC token, which means that you only have a single set of audiences per workflow.

I think that means if we restrict `aud` to only be PyPI, then they cannot auth that workflow with OIDC anywhere else, because they only get the one token. It feels janky to support a provider where we would have to expect to be the *only* thing that authenticates with OIDC? At least I can assume that we will very quickly get an issue asking us to relax the `aud` requirements because someone wants to auth to two things from the same workflow.

jbialy:
Hi @di and @dstufft,

I wanted to post an update that CircleCI has recently rolled out support for generating ID token with [custom audience claim at the job/step level](https://circleci.com/docs/oidc-tokens-with-custom-claims/#create-oidc-token-in-a-job-with-custom-claims). This unblocks anyone who needs a token restricted to a single audience and gets around the limitation of having to share multiple audience claims in a [single pipeline token](https://circleci.com/docs/openid-connect-tokens/#openid-connect-id-token-availability).

We've also added a reply in the original sigstore issue https://github.com/sigstore/fulcio/issues/591#issuecomment-1831318416.

woodruffw:
Thanks @jbialy! I believe this fully unblocks things here; we've made corresponding changes in `id` here: https://github.com/di/id/pull/144

miketheman:
I was reading a [recent CircleCI blog post](https://circleci.com/blog/openid-connect-identity-tokens/), and it looks like there's some added complexity in that their provider is CircleCI org-specific, so the URL pattern would be:

> `https://oidc.circleci.com/org/<YOUR ORGANIZATION ID>/.well-known/openid-configuration`

[...]

PR #19309, opened 2026-01-13
feat: add circleci as trusted publisher [wip]

meeech:
https://github.com/pypi/warehouse/issues/13888

Will collect open questions here as I review. There are def some details I don't know the nuance around, but this should get pretty close.

Open design question: the truth is, we should be able to make trusted publisher revolve exclusively around pipeline-definition-id as the only required piece of information, since that is bound to a project and org. its not just a string like 'workflow.yml'. requiring the orgid and project id are just redundant. Looking at the google form, it seems a minimal amount of required info would be ok? [ after thinking about it, we will keep org-id, project-id - it just makes things easier for users if we ever need to go back to the source, and its easy info to have when setting things up ] 

we can make contexts optional (they would satisfy the same usecase as envs since you make a context and restrict access to certain users or branches)

MUST
* we must reject any publish where the ssh-rerun claim is true. like we dont even allow the exchange for an api token to do the publish

### DIFF
--- a/warehouse/manage/views/oidc_publishers.py
+++ b/warehouse/manage/views/oidc_publishers.py
@@ -787,8 +787,10 @@ class ManageOIDCPublisherViews:
             return response
 
         # CircleCI OIDC publishers are unique on the tuple of
-        # (circleci_org_id, circleci_project_id, pipeline_definition_id),
+        # (circleci_org_id, circleci_project_id, pipeline_definition_id, context_id),
         # so we check for an already registered one before creating.
+        # Empty string is used to represent "no context" (unconstrained).
+        context_id = form.context_id.data or ""
         publisher = (
             self.request.db.query(CircleCIPublisher)
             .filter(
@@ -796,6 +798,7 @@ class ManageOIDCPublisherViews:
                 CircleCIPublisher.circleci_project_id == form.circleci_project_id.data,
                 CircleCIPublisher.pipeline_definition_id
                 == form.pipeline_definition_id.data,
+                CircleCIPublisher.context_id == context_id,
             )
             .one_or_none()
         )
@@ -804,6 +807,7 @@ class ManageOIDCPublisherViews:
                 circleci_org_id=form.circleci_org_id.data,
                 circleci_project_id=form.circleci_project_id.data,
                 pipeline_definition_id=form.pipeline_definition_id.data,
+                context_id=context_id,
             )
 
             self.request.db.add(publisher)
--- a/warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
+++ b/warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
@@ -22,6 +22,7 @@ def upgrade():
         sa.Column("circleci_org_id", sa.String(), nullable=False),
         sa.Column("circleci_project_id", sa.String(), nullable=False),
         sa.Column("pipeline_definition_id", sa.String(), nullable=False),
+        sa.Column("context_id", sa.String(), nullable=True),
         sa.ForeignKeyConstraint(
             ["id"],
             ["oidc_publishers.id"],
@@ -31,6 +32,7 @@ def upgrade():
             "circleci_org_id",
             "circleci_project_id",
             "pipeline_definition_id",
+            "context_id",
             name="_circleci_oidc_publisher_uc",
         ),
     )
@@ -40,6 +42,7 @@ def upgrade():
         sa.Column("circleci_org_id", sa.String(), nullable=False),
         sa.Column("circleci_project_id", sa.String(), nullable=False),
         sa.Column("pipeline_definition_id", sa.String(), nullable=False),
+        sa.Column("context_id", sa.String(), nullable=True),
         sa.ForeignKeyConstraint(
             ["id"],
             ["pending_oidc_publishers.id"],
@@ -49,6 +52,7 @@ def upgrade():
             "circleci_org_id",
             "circleci_project_id",
             "pipeline_definition_id",
+            "context_id",
             name="_pending_circleci_oidc_publisher_uc",
         ),
     )
--- a/warehouse/oidc/forms/circleci.py
+++ b/warehouse/oidc/forms/circleci.py
@@ -7,7 +7,12 @@ from warehouse.oidc.forms._core import PendingPublisherMixin
 
 
 class CircleCIPublisherBase(wtforms.Form):
-    __params__ = ["circleci_org_id", "circleci_project_id", "pipeline_definition_id"]
+    __params__ = [
+        "circleci_org_id",
+        "circleci_project_id",
+        "pipeline_definition_id",
+        "context_id",
+    ]
 
     circleci_org_id = wtforms.StringField(
         validators=[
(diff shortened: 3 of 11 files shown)


warehouse/oidc/models/circleci.py
```diff
@@ -0,0 +1,166 @@
...
+    def publisher_name(self) -> str:
+        return "CircleCI"
+
+    @property
+    def publisher_base_url(self) -> str | None:
+        # CircleCI doesn't have a predictable public URL pattern for projects
+        # based on org-id/project-id (they're UUIDs)
+        return None
+
+    def publisher_url(self, claims: SignedClaims | None = None) -> str | None:
```
meeech, line 95: q: right now, we can just produce a url to a circleci.com/api to get a json blob representing the config that produced the artifact (api_url + "/pipeline/" + pipeline-id + "/config") or (api_url + "/workflow/" + workflow-id + "/job")

working on setting up something like app.circleci.com/workflow/{{workflow-uuid}} or app.circleci.com/job/{{job-uuid}}

warehouse/oidc/forms/circleci.py
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
```
meeech, line 1: TODO: create all the required fields

warehouse/manage/views/oidc_publishers.py
```diff
@@ -732,6 +740,117 @@ def add_activestate_oidc_publisher(self):
 
         return HTTPSeeOther(self.request.path)
 
+    @view_config(
```
meeech, line 743: Q: I'm just a tourist here, but wondering is it time to break these apart into sep files and import?  since it will keep growing as new pubs are added? though maybe its so rare it doesn't really matter.

willing to do it, but i would want a maintainer with more xp in the code to approve that

warehouse/manage/views/oidc_publishers.py
```diff
@@ -732,6 +740,117 @@ def add_activestate_oidc_publisher(self):
...
+        if not form.validate():
+            self.request.session.flash(
+                self.request._("The trusted publisher could not be registered"),
+                queue="error",
+            )
+            return response
+
+        # CircleCI OIDC publishers are unique on the tuple of
+        # (circleci_org_id, circleci_project_id), so we check for an already
+        # registered one before creating.
```
meeech: todo/q - check into this tuple - what is the uniqueness requirements - how granular do we need to go? how does this interact with the one top level iss?

meeech: based on whaty Im seeing with the github publisher, what we have is too broad. will probably want to go all the way down to the pipeline definition id

warehouse/templates/manage/project/publishing.html
```diff
@@ -353,6 +353,64 @@ <h3 class="callout-block__heading">{% trans %}Proceed with caution!{% endtrans %
   </div>
 </form>
 {% endmacro %}
+{% macro circleci_form(request, circleci_publisher_form) %}
+  {{ prefilled_warning() if prefilled_provider == "circleci" }}
+  <p>
+    {% trans href="https://circleci.com/docs/openid-connect-tokens/" %}
```
meeech, line 359: this is an old url., which we have a redirect on. but prolly best to give new url `https://circleci.com/docs/guides/permissions-authentication/openid-connect-tokens/`

warehouse/templates/manage/project/publishing.html
```diff
@@ -353,6 +353,64 @@ <h3 class="callout-block__heading">{% trans %}Proceed with caution!{% endtrans %
   </div>
 </form>
 {% endmacro %}
+{% macro circleci_form(request, circleci_publisher_form) %}
```
meeech, line 356: todo: add the actual fields we need

warehouse/templates/manage/account/publishing.html
```diff
@@ -373,6 +373,82 @@
   </div>
 </form>
 {% endmacro %}
+{% macro circleci_form(request, pending_circleci_publisher_form) %}
+  <p>
+    {% trans href="https://circleci.com/docs/openid-connect-tokens/" %}
```
meeech, line 378: use newer URL - this is a redirect

warehouse/templates/manage/account/publishing.html
```diff
@@ -373,6 +373,82 @@
...
+    <div id="project_name-errors">{{ field_errors(pending_circleci_publisher_form.project_name) }}</div>
+  </div>
+  <div class="form-group">
+    <label for="circleci_org_id" class="form-group__label">
+      {% trans %}Organization ID{% endtrans %}
+      {% if pending_circleci_publisher_form.circleci_org_id.flags.required %}
+        <span class="form-group__required">{% trans %}(required){% endtrans %}</span>
+      {% endif %}
+    </label>
+    {{ pending_circleci_publisher_form.circleci_org_id(placeholder=gettext("xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx") ,
```
meeech, line 416: todo: proper fields

warehouse/templates/manage/organization/publishing.html
```diff
@@ -386,6 +386,82 @@
   </div>
 </form>
 {% endmacro %}
+{% macro circleci_form(request, pending_circleci_publisher_form, organization) %}
+  <p>
+    {% trans href="https://circleci.com/docs/openid-connect-tokens/" %}
```
meeech, line 391: old url redirect. use latest

warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
```diff
@@ -0,0 +1,67 @@
...
+
+revision = "549bdbefa6bd"
+down_revision = "a6cae8e65f1a"
+
+
+def upgrade():
+    op.create_table(
+        "circleci_oidc_publishers",
+        sa.Column("id", sa.UUID(), nullable=False),
+        sa.Column("circleci_org_id", sa.String(), nullable=False),
```
meeech, line 22: todo/q: what is the scope here for these columns.

warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
```diff
@@ -0,0 +1,67 @@
...
+Revision ID: 549bdbefa6bd
+Revises: a6cae8e65f1a
+Create Date: 2026-01-13 00:00:00.000000
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "549bdbefa6bd"
```
meeech, line 14: todo: look closer here - good chance this is wrong

warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
```diff
@@ -0,0 +1,67 @@
+# SPDX-License-Identifier: Apache-2.0
```
meeech, line 1: revist once we know all the fields.

warehouse/migrations/versions/549bdbefa6bd_add_circleci_oidc_publisher.py
meeech: update: maybe not! cursory look seems close to accurate (the down hash was in dec)

meeech:
i fucked up a scrub 🤦🏼 sorry cleaning up

meeech:
I am declaring bankruptcy on this PR. Will open a new one.

PR #19315, opened 2026-01-14
add validator for null bytes to Org Application form

JacobCoffee:
ensure we validate nullbytes on org app form


tests/unit/manage/test_forms.py
```diff
@@ -919,6 +919,15 @@ def test_validate_name_with_organization(self):
             pretend.call("my_organization_name")
         ]
 
+    def test_validate_name_with_null_bytes(self):
+        form = forms.CreateOrganizationApplicationForm(
+            MultiDict({"name": "test\x00name"}),
+            organization_service=pretend.stub(),
+            user=pretend.stub(),
+        )
+        form.validate()
```
miketheman: Ensure the form doesn't get validated
```suggestion
        assert not form.validate()
```

tests/unit/manage/test_forms.py
```diff
@@ -814,7 +814,7 @@ def test_validate_name_with_no_organization(self):
             organization_service=organization_service,
             user=user,
         )
-        field = pretend.stub(data="my_organization_name")
+        field = pretend.stub(data="my_organization_name", errors=[])
```
miketheman:
```suggestion
        field = pretend.stub(data="my_organization_name")
```

tests/unit/manage/test_forms.py
```diff
@@ -838,7 +838,7 @@ def test_validate_name_with_existing_application(self, db_session):
...
```
miketheman:
```suggestion
        field = pretend.stub(data="my_organization_name")
```

tests/unit/manage/test_forms.py
```diff
@@ -910,7 +910,7 @@ def test_validate_name_with_organization(self):
             organization_service=organization_service,
             user=pretend.stub(),
         )
-        field = pretend.stub(data="my_organization_name")
+        field = pretend.stub(data="my_organization_name", errors=[])
```
miketheman:
```suggestion
        field = pretend.stub(data="my_organization_name")
```

tests/unit/manage/test_forms.py
```diff
@@ -938,6 +951,21 @@ def test_save(self, pyramid_request):
...
+            MultiDict({"name": "test"}),
+            organization_service=organization_service,
+            user=pretend.stub(),
+        )
+        field = pretend.stub(data="test", errors=["Existing error"])
+
+        form.validate_name(field)
+
+        assert organization_service.find_organizationid.calls == []
+
```
miketheman: If the inline check uis removed, this test is no longer necessary.
```suggestion
```

warehouse/manage/forms.py
```diff
@@ -472,7 +474,9 @@ class OrganizationNameMixin:
     organization_id = None
 
     def validate_name(self, field):
-        # Find organization by name.
+        if field.errors:
+            return
+
```
miketheman: I don't think this check is necessary, since the validation chain should stop before this validation happens.

```suggestion
# Find organization by name.
```

miketheman (changes requested):
Some more inline changes - even simpler.

Likely needs a `make translations` after.

PR #19332, opened 2026-01-16
name our rate limiters

ewdurbin:
Previously since we had no top-level "identifiers" configured in our use of the limits library, all rate limiters with the same rate limit string were hit/tested for different purposes.

This scopes all rate limits to the configured name.

See two commits: a manual fix that totally worked, and what the 🤖 said to do to stop from repeating ourselves.


miketheman (approved):
Looking like a nice refactor, thanks!

PR #19340, opened 2026-01-16
fix(#19338): handle missing submitter in role verification flows

JacobCoffee:
Fix issues surfaced from fixing [WAREHOUSE-PRODUCTION-2D6](https://python-software-foundation.sentry.io/issues/7195794231/?alert_rule_id=409168&alert_type=issue&notification_uuid=7c9c5557-ff14-462e-a96e-26a0e65487a3&project=51408&referrer=slack) that would throw errors on data access when a user is deleted


ewdurbin (commented):
Alternative, rather than reaching this point, we could treat "The inviting user exists" as a condition for an invitation to be acceptable.

After [line 1182](https://github.com/pypi/warehouse/pull/19340/changes#diff-6ffb5ead6d79fe07ebbabcc9c25227926583a5e15a2baff1ce3ffd580deae633R1182), we could add a check for "inviting user still exists" and `return _error(request._("Invalid Invitation: The inviting user no longer exists. Please ask a remaining Owner to reissue your invitation"))` or similar...

PR #19352, opened 2026-01-21
feat: auto-quarantine young projects from one observer

miketheman:
If a single trusted observer reports a new project, place it in quarantine for review.

We don't onboard folks as trusted observers that often, and newer projects can be removed from the index quicker this way.


JacobCoffee:
i think i like it. would it make sense to go back across all observations to date and see how many observations were made by a single person to a project <24hrs old to see what that would've allowed? and see some ratio of correct:incorrect observations? 

then maybe it'd be easier to say "usually our observers are right, so this makes sense"

miketheman:
Thanks, good suggestion!

I had some help writing a complex SQL query, which uses the logic found in the observation insights to help qualify false-positives.

Query:

<details>

```sql
WITH trusted_observer_reports AS (
    -- Start small: only malware reports from trusted observers
    SELECT o.id                                   AS observation_id,
           o.related_id,
           o.related_name,
           (regexp_match(o.related_name,
                         'name=''([^'']+)'''))[1] AS project_name,
           o.created                              AS report_time,
           o.actions
    FROM project_observations o
             JOIN observers obs ON o.observer_id = obs.id
             JOIN users u ON u.observer_association_id = obs._association_id
    WHERE o.kind = 'is_malware'
      AND u.is_observer = true),
     single_observer_projects AS (
         -- Only check observer counts for projects we care about
         SELECT related_name
         FROM project_observations
         WHERE kind = 'is_malware'
           AND
             related_name IN (SELECT related_name FROM trusted_observer_reports)
         GROUP BY related_name
         HAVING COUNT(DISTINCT observer_id) = 1),
     filtered_reports AS (
         -- Now we have a small set: trusted observer + single observer
         SELECT tor.*
         FROM trusted_observer_reports tor
         WHERE tor.related_name IN
               (SELECT related_name FROM single_observer_projects)),
     project_creation_times AS (
         -- Only look up creation times for the projects we actually need
         SELECT name                AS project_name,
                MAX(submitted_date) AS created_time
         FROM journals
         WHERE action = 'create'
           AND name IN (SELECT project_name
                        FROM filtered_reports
                        WHERE project_name IS NOT NULL)
         GROUP BY name),
     classified AS (SELECT fr.observation_id,
                           fr.project_name,
                           fr.report_time,
                           pct.created_time AS project_created,
                           EXTRACT(EPOCH FROM
                                   (fr.report_time - pct.created_time)) /
                           3600             AS project_age_hours,
                           CASE
                               WHEN fr.related_id IS NULL THEN 'true_positive'
                               WHEN fr.actions::text LIKE '%remove_malware%'
                                   THEN 'true_positive'
                               WHEN fr.actions::text LIKE '%verdict_not_malware%'
                                   THEN 'false_positive'
                               ELSE 'pending'
                               END          AS verdict
                    FROM filtered_reports fr
                             LEFT JOIN project_creation_times pct
                                       ON pct.project_name = fr.project_name)
SELECT COUNT(*)                                           AS total_would_auto_quarantine,
       COUNT(*) FILTER (WHERE verdict = 'true_positive')  AS true_positives,
       COUNT(*) FILTER (WHERE verdict = 'false_positive') AS false_positives,
       COUNT(*) FILTER (WHERE verdict = 'pending')        AS pending,
       ROUND(
                       COUNT(*)
                       FILTER (WHERE verdict = 'true_positive')::numeric *
                       100.0 /
                       NULLIF(COUNT(*)
                              FILTER (WHERE verdict IN ('true_positive', 'false_positive')),
                              0),
                       1
       )                                                  AS accuracy_pct
FROM classified
WHERE project_age_hours < 24
  AND project_age_hours >= 0;
```

</details>

| total\_would\_auto\_quarantine | true\_positives | false\_positives | pending | accuracy\_pct |
| :--- | :--- | :--- | :--- | :--- |
| 740 | 736 | 3 | 1 | 99.6 |

Seems like a pretty good ratio.

PR #19358, opened 2026-01-21
test: add proper typing to Observers and Events

miketheman:
Mitigates the need to ignore attributes when used.

Also refactors usage of sqlalchemy's `UUID` to use native UUIDs whenever possible.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -1017,32 +1017,32 @@ msgstr[1] ""
 msgid "Just now"
 msgstr ""
 
-#: warehouse/subscriptions/models.py:25
+#: warehouse/subscriptions/models.py:24
 #: warehouse/templates/manage/project/history.html:209
 msgid "Active"
 msgstr ""
 
-#: warehouse/subscriptions/models.py:26
+#: warehouse/subscriptions/models.py:25
 msgid "Past Due"
 msgstr ""
 
-#: warehouse/subscriptions/models.py:27
+#: warehouse/subscriptions/models.py:26
 msgid "Unpaid"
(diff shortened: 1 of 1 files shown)


tests/unit/cli/test_db.py
```diff
@@ -325,16 +324,16 @@ def test_dbml_command(monkeypatch, cli):
...
 }
 
 Table _clan_member {
   name varchar [not null]
-  clan_id varchar
+  clan_id uuid
   joined datetime [not null, default: `now()`]
   departed datetime
-  id varchar [pk, not null, default: `gen_random_uuid()`]
+  id uuid [pk, not null, default: `gen_random_uuid()`]
```
miketheman, line 336: I double-checked, and dbml app does indeed support `uuid` data type

warehouse/observations/models.py
```diff
@@ -138,6 +142,15 @@ class Observation(AbstractConcreteBase, db.Model):
         "polymorphic_identity": "observation",
     }
 
+    if typing.TYPE_CHECKING:
```
ewdurbin, line 145: I am curious how we can/will ensure that these manually written attributes stay in sync.

I _assume_ that if an attribute is added or removed from the subclasses then the type checker would detect it?

miketheman: This is definitely a challenge I tried really hard to avoid. Other solutions were typing.Protocol - but that didn't work since the base classes didn't line up. 
If the attribute is removed from the **subclass**, the type checker won't know about it, since the "stub" here is telling the type checker what to do.
Thankfully we only use this table-per-related pattern in two places today.

PR #19371, opened 2026-01-23
Add RFC 9457 Problem Details support for Simple API errors 

luisgcoding:
This PR implements [RFC9457](https://www.rfc-editor.org/rfc/rfc9457.html) support for PyPI's Simple API error responses, replacing plain text error messages with standardized, machine-readable JSON error responses. 

This is part of a Pre-PEP work discussed [here](https://discuss.python.org/t/pre-pep-discussion-rfc-9457-error-responses-for-package-registries/105453).

Current PyPi behaviour when executing command `curl -i http://localhost:80/simple/nonexistent-package/ `:
Screenshot 2026-01-28 at 7 01 43 PM

Now with the change proposed here:
Screenshot 2026-01-28 at 6 48 02 PM


warehouse/rfc9457.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Additional extension members can be provided via the extensions dict.
+    """
+
+    status: int
+    title: str
+    detail: str | None = None
+    type: str = "about:blank"
+    instance: str | None = None
+    extensions: dict[str, Any] = field(default_factory=dict)
```
warsaw: When I was looking at the test, it seemed odd that extensions were passed in as a dictionary, but got flattened in the JSON result.  I was put off by the different shape, but now I see why; it's because `ProblemDetails` is a dataclass, which doesn't have a way to specify `**kws`.  I think that's okay, except that the way it's written, it's possible for keys in `extensions` to override the required arguments, e.g.

```python
pd = ProblemDetails(400, "Bad request", extensions={"status": 200})
```

Maybe that's not terrible, but maybe it makes sense to do some extra checking before the `result.update()` on line 66?

I also wonder is whether `extensions` should be a `kw_only` field?

warehouse/rfc9457.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Returns a dictionary representation following RFC 9457 structure.
+        The 'type' field is omitted when it equals 'about:blank' (default).
+        """
+        result: dict[str, Any] = {
+            "status": self.status,
+            "title": self.title,
+        }
+
+        if self.type != "about:blank":
```
warsaw, line 58: My [reading](https://www.rfc-editor.org/rfc/rfc9457.html#name-aboutblank) of `about:blank` says it *can* be implicit but doesn't need to be.  Should we default to explicit?

Also should we support the **SHOULD** "the title SHOULD be the same as the recommended HTTP status phrase for that code"?

warehouse/rfc9457.py
```diff
@@ -0,0 +1,106 @@
...
+        if self.detail:
+            parts.append(f"\n{self.detail}")
+
+        return " ".join(parts) if parts else "Unknown problem"
+
+
+def problem_details_from_exception(
+    exc: HTTPException, detail: str | None = None, **extensions: Any
+) -> ProblemDetails:
+    problem_detail = detail
```
warsaw: Could be:

```python
    if detail is None:
        problem_detail = getattr(...
    else:
        problem_detail = detail
```

or a ternary expression, but I don't think that would make things more readable.

warehouse/rfc9457.py
```diff
@@ -0,0 +1,106 @@
...
+    return ProblemDetails(
+        status=exc.status_code,
+        title=exc.title or exc.status,
+        detail=problem_detail,
+        extensions=extensions,
+    )
+
+
+def accepts_problem_json(request) -> bool:
+    acceptable = request.accept.acceptable_offers([RFC9457_CONTENT_TYPE])
```
warsaw: There was some [question on DPO](https://discuss.python.org/t/pre-pep-discussion-rfc-9457-error-responses-for-package-registries/105453/12) whether we wanted to limit this response to only requests that explicitly accept `application/problem+json`, although now that I'm looking I don't see `accepts_problem_json()` called anywhere.

I'm unsure so maybe @woodruffw can chime in?

warsaw (changes requested):

warehouse/rfc9457.py
luisgcoding: I added validation and made the field `kw_only` [here](7bcb466fe4452a2e3b61461bc393c3d97d438097).

warehouse/rfc9457.py
luisgcoding: I added the SHOULD [here](7bcb466fe4452a2e3b61461bc393c3d97d438097) in method `problem_details_from_exception`. I think is better to be implicit but happy to change it otherwise.

warehouse/rfc9457.py
luisgcoding: Fixed [here](7bcb466fe4452a2e3b61461bc393c3d97d438097)

warehouse/rfc9457.py
luisgcoding: I removed the accepts_problem_json for the moment. We can add it back if we decide to use it for backwards compatibility.

warehouse/views.py
```diff
@@ -94,11 +98,13 @@ def httpexception_view(exc, request):
...
         if isinstance(exc, HTTPNotFound) and request.path.startswith("/simple/"):
-            response = HTTPNotFound(
-                body="404 Not Found",
-                content_type="text/plain",
+            # Return RFC 9457 Problem Details for all /simple/ errors.
+            problem = problem_details_from_exception(exc)
+            response = type(exc)(
+                body=problem.to_json(),
+                charset="utf-8",
+                content_type=RFC9457_CONTENT_TYPE,
```
zsol, line 107: AFAICT this unconditionally returns the new content type, but wouldn't it be better if this was negotiated using `Accept` headers?

zsol (commented):
Feel free to ignore my unsolicited review, just a couple of questions I had while reading this.

Would it make sense to extend these responses to:
- 503 responses (`views.service_unavailable` currently returns HTML)
- various 400 responses (mostly defined in `sanity.py`)
- maaaybe the 405 error coming from `warehouse.csrf.require_method_view`
?

PR #19380, opened 2026-01-23
granian for local dev

ewdurbin:
Just playing around to see how it do.


ewdurbin:
not a huge or obvious win and would need quite a bit of work to validate before shipping.

PR #19381, opened 2026-01-23
fix: store last checked time for visibility

miketheman:
We should update the last checked time even though the status may not have changed in order to track that something has indeed been checked.

With the current behavior it appears that nothing has been checked recently, even though it might have been already checked.

### DIFF
--- a/tests/unit/accounts/test_tasks.py
+++ b/tests/unit/accounts/test_tasks.py
@@ -244,12 +244,8 @@ def test_update_email_domain_status_retries_failures_in_7_days(
 
     # Timestamp should be set to ~23 days ago for retry in ~7 days
     assert fail_check.domain_last_checked is not None
-    expected_retry_offset = datetime.now(tz=timezone.utc) - timedelta(days=23)
-    # Allow 1 minute tolerance for test execution time
-    assert (
-        abs((fail_check.domain_last_checked - expected_retry_offset).total_seconds())
-        < 60
-    )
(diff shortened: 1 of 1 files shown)


tests/unit/accounts/test_tasks.py
```diff
@@ -237,7 +242,15 @@ def test_update_email_domain_status_does_not_update_if_not_needed(
 
     domain_status_service.get_domain_status.assert_called_once_with(fail_check.domain)
 
-    assert fail_check.domain_last_checked is None
+    # Timestamp should be set to ~23 days ago for retry in ~7 days
+    assert fail_check.domain_last_checked is not None
+    expected_retry_offset = datetime.now(tz=timezone.utc) - timedelta(days=23)
+    # Allow 1 minute tolerance for test execution time
+    assert (
+        abs((fail_check.domain_last_checked - expected_retry_offset).total_seconds())
```
ewdurbin: would checking that `fail_check.domain_last_checked.date() == expected_retry_offset.date()` be sufficient?

FWIW `expected_retry_offset` isn't the best name here since it's actually a datetime, not an offset?

```python
expected_retry_date = (datetime.now(tz=timezone.utc) - timedelta(days=23)).date()
assert fail_check.domain_last_checked.date() == expected_retry_date.date()
```

is... probably fine?

miketheman: Thanks, normalizing to a date works well here and is simpler, will refactor!

PR #19411, opened 2026-01-28
Validate WHEEL file tags

konstin:
It can happen that the filename and the dist-info `WHEEL` file mismatch, either because the author manually changed the wheel filename, or due to a bug in the build backend implementation.

This caused a number of bugs in uv and pip, which expected the `WHEEL` file to contain accurate tags in the basic `Key: Value` format:

* https://github.com/astral-sh/uv/issues/2149
* https://github.com/astral-sh/uv/issues/6615
* https://github.com/astral-sh/uv/issues/15832
* https://github.com/astral-sh/uv/issues/16581
* https://github.com/astral-sh/uv/issues/16823
* https://github.com/pypa/pip/issues/13709
* https://github.com/astral-sh/uv/issues/17711

Another notable example is https://github.com/ziglang/zig-pypi/pull/40, where the author wasn't aware of the expanded tags requirement.

This PR adds validation that enforces the basic `Key: Value` format, that the tags in the `WHEEL` file are expanded and that the tags match.

---

Speaking of https://github.com/ziglang/zig-pypi/pull/40, https://github.com/pypi/support/issues/8180 would really valuable to have, zig is an important part of the toolchain of building manylinux-compliant python packages using rust.

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -10,6 +10,7 @@ import zipfile
 
 import packaging.requirements
 import packaging.specifiers
+import packaging.tags
 import packaging.utils
 import packaging.version
 import packaging_legacy.version
@@ -347,19 +348,6 @@ def _is_valid_dist_file(filename, filetype, metrics, *, scan=True):
                         zfp.read(target_file)
                     except KeyError:
                         return False, f"PKG-INFO not found at {target_file}"
-                if filename.endswith(".whl"):
-                    try:
-                        name, version, _ = os.path.basename(filename).split("-", 2)
-                    except ValueError:
-                        return (
-                            False,
-                            "Unable to parse name and version from wheel filename",
-                        )
-                    target_file = os.path.join(f"{name}-{version}.dist-info", "WHEEL")
-                    try:
-                        zfp.read(target_file)
-                    except KeyError:
-                        return False, f"WHEEL not found at {target_file}"
 
                 # Scan archive members for YARA rule matches while open
                 if scan:
@@ -431,6 +419,61 @@ def _is_valid_dist_file(filename, filetype, metrics, *, scan=True):
 
     # If we haven't yet decided it's not valid, then we'll assume it is and
     # allow it.
+    return True, None
+
+
+def _is_valid_wheel_file(
+    contents: bytes, tags: frozenset[packaging.tags.Tag]
+) -> tuple[bool, str | None]:
+    """
+    Check whether the WHEEL file matches the wheel filename tags.
+
+    A mismatch can happen when manually editing the wheel filename without also editing
+    the dist-info WHEEL file.
+    """
+    filename_tags = {str(tag) for tag in tags}
+    dist_info_tags = set()
+    try:
+        decoded = contents.decode("utf-8")
+    except UnicodeDecodeError:
+        return False, "WHEEL file is not valid UTF-8"
+
+    for line in decoded.splitlines():
+        if line.startswith("Tag:"):
+            tag = line.removeprefix("Tag:").strip()
+            if "." in tag:
+                return (
+                    False,
+                    f"Tags in WHEEL file must be expanded, not compressed: {tag}",
+                )
+            dist_info_tags.add(tag)
+
+    # Handle the case that there's no overlap, usually two different tags.
+    if not (dist_info_tags & filename_tags):
+        formatted_filename_tags = ", ".join(sorted(filename_tags))
+        formatted_dist_info_tags = ", ".join(sorted(dist_info_tags))
+        return (
+            False,
+            "Wheel filename and WHEEL file tags mismatch: "
+            + f"{formatted_filename_tags} vs. {formatted_dist_info_tags}",
+        )
+
+    only_dist_info_tags = dist_info_tags - filename_tags
+    if only_dist_info_tags:
+        formatted_tags = ", ".join(sorted(only_dist_info_tags))
+        return (
+            False,
+            f"WHEEL file has tags not in wheel filename: {formatted_tags}",
+        )
+
+    only_filename_tags = filename_tags - dist_info_tags
+    if only_filename_tags:
+        formatted_tags = ", ".join(sorted(only_filename_tags))
+        return (
+            False,
+            f"Wheel filename has tags not in WHEEL file: {formatted_tags}",
+        )
+
     return True, None
 
 
@@ -1555,6 +1598,27 @@ def file_upload(request):
             metadata_file_hashes = {
                 k: h.hexdigest().lower() for k, h in metadata_file_hashes.items()
             }
+
+            with zipfile.ZipFile(temporary_filename) as zfp:
+                target_file = os.path.join(f"{name}-{version}.dist-info", "WHEEL")
+                try:
+                    wheel_file = zfp.read(target_file)
+                    valid, reason = _is_valid_wheel_file(wheel_file, tags)
+                except KeyError:
+                    valid = False
+                    reason = f"WHEEL not found at {target_file}"
+
+                if not valid:
+                    request.metrics.increment(
+                        "warehouse.upload.failed",
+                        tags=[
+                            "reason:invalid-distribution-file",
+                            f"filetype:{form.filetype.data}",
+                        ],
+                    )
+                    raise _exc_with_message(
+                        HTTPBadRequest, f"Invalid distribution file. {reason}"
+                    )
 
         # If the user provided attestations, verify them
         # We persist these attestations subsequently, only after the
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import datetime
+import email.parser
+import email.policy
 import hashlib
 import hmac
 import os.path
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -390,6 +378,61 @@ def _is_valid_dist_file(filename, filetype):
...
+
+    for line in decoded.splitlines():
+        if line.startswith("Tag:"):
+            tag = line.removeprefix("Tag:").strip()
+            if "." in tag:
+                return (
+                    False,
+                    f"Tags in WHEEL file must be expanded, not compressed: {tag}",
+                )
+            dist_info_tags.add(tag)
```
woodruffw: I think we should probably use an RFC 2822 header parser here, like the one that's used to parse metadata elsewhere.

warehouse/forklift/legacy.py
```diff
@@ -390,6 +378,61 @@ def _is_valid_dist_file(filename, filetype):
     return True, None
 
 
+def _is_valid_wheel_file(
+    contents: bytes, tags: frozenset[packaging.tags.Tag]
+) -> tuple[bool, str | None]:
```
woodruffw: Maybe we can raise a ValueError from this validator instead of returning a flag and message here?

warehouse/forklift/legacy.py
konstin: Note that the email parser doesn't enforce UTF-8 anymore, it accepts non-UTF-8, too.

woodruffw: Hmm, maybe we could take some of the tricks from `packaging`'s `parse_email`?

https://github.com/pypa/packaging/blob/6fe1d5b93364827ff6d1bb78ffc6bf34c27afe3f/src/packaging/metadata.py#L350

Alternatively, I think we'd be fine to enforce that the input is a `str` (thus UTF-8). In practice other WHEEL encodings shouldn't really happen and wouldn't be well supported anyways.

konstin: I've changed it to actively enforcing UTF-8 and returned the test to its previous assertion.

warehouse/forklift/legacy.py
```diff
@@ -1500,6 +1539,35 @@ def file_upload(request):
                 k: h.hexdigest().lower() for k, h in metadata_file_hashes.items()
             }
 
+            with zipfile.ZipFile(temporary_filename) as zfp:
+                target_file = os.path.join(f"{name}-{version}.dist-info", "WHEEL")
```
woodruffw, line 1599: Curious but not a blocker: is there somewhere else we can do this where the wheel is already open for reading? IIRC `file_upload` already opens and parses metadata from distributions elsewhere, might be nice to single-source with those here.

woodruffw (approved):

warehouse/forklift/legacy.py
konstin: I had an initial version where I restructured the validation pass to be sorted by source dist vs. wheel (there's `.zip` source dist logic still in there) and made it open the archive only once, but abandoned it for too much churn for adding a single validation pass.

ewdurbin: yes... this is just one more of.... I think 6 or 7 such open/reads.

warehouse/forklift/legacy.py
```diff
@@ -390,6 +380,55 @@ def _is_valid_dist_file(filename, filetype):
...
+    """
+    filename_tags = {str(tag) for tag in tags}
+    dist_info_tags = set()
+
+    try:
+        content = contents.decode("utf-8")
+    except UnicodeDecodeError:
+        raise ValueError("WHEEL file is not UTF-8 encoded")
+
+    data = email.parser.Parser(policy=email.policy.compat32).parsestr(content)
```
woodruffw: Nit: not super critical, but you can use `HeaderParser` here to clarify that `WHEEL` only contains headers, and we never parse an email-style trailing body from it.

woodruffw (approved):

ewdurbin (approved):

di:
Just want to say, I think I would like this to eventually land in https://github.com/pypa/packaging/pull/1009 instead, but since progress has (admittedly) been slow there, I have no problems adding support here first.

konstin:
For some context on the impact here, uv is basically a `WHEEL` test harness for invalid tags, as they cause a reinstall loop. In the description is listed the cases that were reported over the months, including the notable recent one where torch accidentally published a broken wheel, which would have been caught be this PR (https://github.com/astral-sh/uv/issues/17711). uv also validates the `Version` field (https://github.com/astral-sh/uv/blob/6b2d6f2c401aefb9d97cff180df0110c99e3493e/crates/uv-install-wheel/src/wheel.rs#L273-L304)

konstin:
Here's the results for the first 50k projects from a hacky script that check the latest release for all projects: https://gist.github.com/konstin/755aead7facf690efb476b90ecee8fe7. I've split them into maturin and non-maturin, since maturin produces a lot of wheels and has since been fixed. I'll update it when i got a full run.

miketheman:
@konstin for impact analysis reporting - something that would help contextualize the impact of this change would be adding the file's date to the output, CSV-style - so that these could be inspected relatively easily.

konstin:
Like this? https://gist.github.com/konstin/56afd979b79ed86afa944ef196d27d7d

miketheman:
> Like this? [gist.github.com/konstin/56afd979b79ed86afa944ef196d27d7d](https://gist.github.com/konstin/56afd979b79ed86afa944ef196d27d7d)

Yes, thanks!

konstin:
Full file: https://gist.github.com/konstin/b299bfd20489758e51fe8750dc52ca20

miketheman:
Thanks @konstin - I was looking to generate some idea of how much this may impact existing upload / builders.

Here's a chart where I only look at the past couple of years of uploads in the data you furnished:

COUNTA of project vs  upload_date (Since 2024-01-01)
[source, data massaged a bit for parsing](https://docs.google.com/spreadsheets/d/1kKP_ufczSm4krz-othsALHqM4i-RCOnSul-k8WXCXsI/edit?usp=sharing)

Those numbers aren't as low as I'd hope/expect them to be.

Slicing another way, it looks like maturin is the biggest contributor to the issue, with over 87% of all problematic uploads in the past year attributed to this generator.

generator | COUNTA of project | COUNTA of project
-- | -- | --
maturin | 31370 | 87.46%
setuptools | 2217 | 6.18%
bdist_wheel | 750 | 2.09%
hatchling | 516 | 1.44%
poetry-core | 409 | 1.14%
skbuild | 134 | 0.37%
scikit-build-core | 118 | 0.33%
xmake | 53 | 0.15%
(blank)  | 31 | 0.09%

...

Looks like something was fixed, since the most recent error is up to maturin v1.9.6, but looks like a lot of folks are still using it, but there's a path forward for those users - upgrade maturin and keep going?

For setuptools, the most recent release of 82.0.0 is reporting similar behavior - one example: https://pypi.org/project/OpenGeode-core/15.31.4rc1/#opengeode_core-15.31.4rc1-cp310-cp310-manylinux_2_28_x86_64.whl

Do those users have a path forward with their tool of choice? Or will their upload be rejected without a path forward with their chosen tool?

Generally speaking the code changes are fine, I'm just worried that if we ship this without having a clear answer for users impacted, it will increase failures and end user confusion.

konstin:
For maturin you can update, the releases should be highly compatible (at least I'm not aware of any blockers to upgrading from the issue tracker). It's kinda embarrassing that this went uncaught for so long.

For the other cases, I'm not clear on whether they are actually build system bugs, or if people are manually renaming wheels and aren't are that they also have to update the `WHEEL` file (or really, shouldn't edit this by hand at all).

miketheman:
> ... It's kinda embarrassing that this went uncaught for so long.

heh, we try our best 😉 

> For the other cases, I'm not clear on whether they are actually build system bugs, or if people are manually renaming wheels and aren't are that they also have to update the `WHEEL` file (or really, shouldn't edit this by hand at all).

Yeah, that would make sense - if the tool generates one thing and the publishing process renames files, that would probably lead to this mismatch. So what can we offer folks in the way of understanding the problem and resolving it? Do we need to have a Pinned Issue here for a while, improve the messaging they receive from an upload failure, or something else?

ewdurbin:
At least in the example provided for opengeode-core it appears to be renaming of files post build:

https://github.com/Geode-solutions/actions/blob/97d31ebb48d31dbcb706296cc313b5c56c181892/.github/workflows/cpp-deploy-linux-python.yml#L122-L125

Screenshot 2026-02-18 at 9 07 41 AM

I think its good to be cautious, but we cannot debug everyones builds for them preemptively.

Perhaps we could have a brief window where we notify people that their builds are bad?

di:
Seems to me like this should have a deprecation period with notification via email before it becomes an outright block.

miketheman:
> Perhaps we could have a brief window where we notify people that their builds are bad?

> Seems to me like this should have a deprecation period with notification via email before it becomes an outright block.

These both seem reasonable, but I'm also curious about the infrequent publisher who hasn't published in the deprecation window hitting this would experience. Say they only publish a new version once a year.

In my mind, having a GH Issue with a "you hit this error during publishing, here's a few things you can explore to resolve, like upgrade maturin past x version, stop renaming files, etc" and link the issue in the error message provides a longer-term solution that folks can use when they hit this condition to self-resolve.

konstin:
Do we have new data since how the upload rate of such packages has developed?

konsti-openai:
Sloppy plots, but it should roughly hold.

image

> Recent contributors: setuptools 49.5%, maturin 12.3%, hatchling 11.1%.

image

Still looking worse than I'd hoped :/

PR #19412, opened 2026-01-28
refactor: extract base class for null OAuth clients

miketheman:
Prepares OAuth infrastructure for multi-provider support:
- Extract _NullOAuthClientBase from NullOAuthClient
- Rename NullOAuthClient to NullGitHubOAuthClient for clarity
- Add _provider and _callback_route class attributes for customization
- Simplify constructor to only require redirect_uri

No functional changes - existing GitHub OAuth behavior is preserved.

### DIFF
--- a/tests/unit/oidc/models/test_github.py
+++ b/tests/unit/oidc/models/test_github.py
@@ -80,11 +80,6 @@ def test_extract_workflow_filename(workflow_ref, expected):
     assert github._extract_workflow_filename(workflow_ref) == expected
 
 
-@pytest.mark.parametrize("claim", ["", "repo", "repo:"])
-def test_check_sub(claim):
-    assert github._check_sub(pretend.stub(), claim, pretend.stub()) is False
-
-
 class TestGitHubPublisher:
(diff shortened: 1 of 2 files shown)


miketheman:
Extracted from #19298 - this is ready to ship on its own, so we can benefit from the refactor and potentially add other services.

tests/unit/accounts/test_oauth.py
```diff
@@ -194,49 +194,34 @@ def test_get_user_info_http_error(self):
...
-        assert client.redirect_uri == "http://localhost/callback"
-
-    def test_initialization_with_params(self):
-        client = NullOAuthClient(
-            client_id="custom_id",
-            client_secret="custom_secret",
-            redirect_uri="http://localhost/callback",
-        )
-        assert client.client_id == "custom_id"
-        assert client.client_secret == "custom_secret"
```
miketheman, line 223: These tests are no longer needed, as the subclasses construct the shape needed.

ewdurbin (approved):

PR #19501, opened 2026-02-10
fix: show disabled reason to authenticated users

miketheman:
In the event that a user's account has been administratively frozen, share that with a message of what to do next, **after** the user has successfully proven ownership of a given account.

### DIFF
--- a/tests/unit/accounts/test_security_policy.py
+++ b/tests/unit/accounts/test_security_policy.py
@@ -408,7 +408,7 @@ class TestSessionSecurityPolicy:
         assert request.session.flash.calls == [
             pretend.call(
                 "Your account has been suspended. "
-                "Please contact admin@pypi.org for assistance.",
+                "Please contact security@pypi.org for assistance.",
                 queue="error",
             )
         ]
--- a/warehouse/accounts/security_policy.py
+++ b/warehouse/accounts/security_policy.py
@@ -82,7 +82,7 @@ class SessionSecurityPolicy:
             if disabled_reason == DisableReason.AccountFrozen:
                 request.session.flash(
                     "Your account has been suspended. "
-                    "Please contact admin@pypi.org for assistance.",
+                    "Please contact security@pypi.org for assistance.",
                     queue="error",
                 )
             else:


warehouse/accounts/security_policy.py
```diff
@@ -75,10 +76,17 @@ def identity(self, request):
...
         # User may have been frozen or disabled since the session was created.
-        is_disabled, _ = login_service.is_disabled(userid)
+        is_disabled, disabled_reason = login_service.is_disabled(userid)
         if is_disabled:
             request.session.invalidate()
-            request.session.flash("Session invalidated", queue="error")
+            if disabled_reason == DisableReason.AccountFrozen:
+                request.session.flash(
+                    "Your account has been suspended. "
+                    "Please contact admin@pypi.org for assistance.",
```
di: My only question is whether this should be security@ instead, in my experience these are almost always due to security-related actions and end up landing in the security@ inbox when initially sent to admin@.

tests/unit/accounts/test_security_policy.py
```diff
@@ -377,7 +378,68 @@ def test_identity_is_disabled(self, monkeypatch):
...
+        assert request.find_service.calls == [pretend.call(IUserService, context=None)]
+        assert user_service.get_user.calls == [pretend.call(userid)]
+        assert request.session.password_outdated.calls == []
+        assert user_service.get_password_timestamp.calls == []
+        assert user_service.is_disabled.calls == [pretend.call(userid)]
+        assert request.session.invalidate.calls == [pretend.call()]
+        assert request.session.flash.calls == [
+            pretend.call(
+                "Your account has been suspended. "
+                "Please contact admin@pypi.org for assistance.",
```
miketheman:
```suggestion
                "Please contact security@pypi.org for assistance.",
```

warehouse/accounts/security_policy.py
miketheman:
```suggestion
                    "Please contact security@pypi.org for assistance.",
```

miketheman: Excellent point, updated!

warehouse/accounts/security_policy.py
```diff
@@ -75,10 +76,17 @@ def identity(self, request):
...
         if is_disabled:
             request.session.invalidate()
-            request.session.flash("Session invalidated", queue="error")
+            if disabled_reason == DisableReason.AccountFrozen:
+                request.session.flash(
+                    "Your account has been suspended. "
+                    "Please contact security@pypi.org for assistance.",
+                    queue="error",
+                )
+            else:
```
JacobCoffee, line 88: we don't want to give other reasons they might've been disabled?

miketheman: It depends - but this is a good cautious approach.

[OWASP guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#authentication-and-error-messages) have guidelines on what should be shared or not - so it's possible that in other scenarios we may want to provide more context, but let's ease into it.

PR #19513, opened 2026-02-11
Add a check constraint for pending publisher project names

di:
Follow on to https://github.com/pypi/warehouse/pull/19502.

### DIFF
--- a/tests/unit/oidc/models/test_core.py
+++ b/tests/unit/oidc/models/test_core.py
@@ -4,7 +4,7 @@ import pretend
 import psycopg
 import pytest
 
-from warehouse.accounts.models import User
+from tests.common.db.oidc import PendingGitHubPublisherFactory
 from warehouse.oidc import errors
 from warehouse.oidc.models import _core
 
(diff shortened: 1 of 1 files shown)


tests/unit/oidc/models/test_core.py
```diff
@@ -31,6 +33,40 @@ def test_check_claim_invariant():
...
+        db_session.flush()
+
+        valid_project_name = "good-name_123"
+        publisher = _core.PendingOIDCPublisher(
+            project_name=valid_project_name,
+            added_by_id=user.id,
+        )
+        db_session.add(publisher)
+        db_session.flush()
+        assert publisher.project_name == valid_project_name
```
miketheman, line 47: This can be rewritten  more succinctly with factories:

```suggestion
class TestPendingOIDCPublisher:
    def test_project_name_constraint(self, db_session):
        invalid_project_name = "İnspect"

        with pytest.raises(psycopg.errors.CheckViolation):
            PendingGitHubPublisherFactory(project_name=invalid_project_name)

    def test_project_name_constraint_valid(self, db_session):
        valid_project_name = "good-name_123"
        publisher = PendingGitHubPublisherFactory(project_name=valid_project_name)

        assert publisher.project_name == valid_project_name
```

tests/unit/oidc/models/test_core.py
```diff
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pretend
+import psycopg
 import pytest
 
+from warehouse.accounts.models import User
```
miketheman:
```suggestion
from tests.common.db.oidc import PendingGitHubPublisherFactory
```

miketheman (approved):
Generally looks good - looks like there's another test case `test_mint_token_with_invalid_name_fails` that is getting triggered by this new constraint, which is great, since the conditions prevented here should remove the need for that test since it "can't happen".

I also rewrote the new tests to be more inline with testing approach with factories that do most of the work

PR #19525, opened 2026-02-13
feat: emit pypi ownership data in JSON API

miketheman:
This replaces the XML-RPC `package_roles` method with a modern,
cacheable alternative.

### DIFF
--- a/docs/user/api/json.md
+++ b/docs/user/api/json.md
@@ -269,7 +269,14 @@ Accept: application/json
               "yanked_reason": null
             }
         ],
-        "vulnerabilities": []
+        "vulnerabilities": [],
+        "ownership": {
+            "roles": [
+                {"role": "Owner", "user": "theacodes"},
+                {"role": "Maintainer", "user": "pypa-bot"}
+            ],
+            "organization": "pypa"
+        }
     }
     ```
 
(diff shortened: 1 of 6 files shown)


docs/user/api/json.md
```diff
@@ -430,10 +437,30 @@ Accept: application/json
...
           }
         ],
-        "vulnerabilities": []
+        "vulnerabilities": [],
+        "ownership": {
+            "roles": [
+                {"role": "Owner", "user": "theacodes"},
+                {"role": "Maintainer", "user": "pypa-bot"}
+            ],
+            "organization": "pypa"
```
miketheman, line 446: Chat review note:

@ewdurbin wanted to explore whether this should have more structure to it, in the future case if there are orgs-owning-orgs and what that might look like as an API change

ewdurbin: I suppose this can be done with some delimiter down the line, or a new key to denote "parent" or "child" organizations (though I can't imagine how child orgs would be relevant)

ewdurbin (approved):

tuuputamadre (approved):

tuuputamadre (changes requested):
Funcionar

PR #19528, opened 2026-02-15
Link Inspector from file details view

### LINKED ISSUE #19032: [FR] Link inspector from the `File details` view

I was at https://pypi.org/project/ssh/#ssh-1.8.0.tar.gz and figured I wanted to look at what's inside.

I know how to navigate to https://inspector.pypi.io/project/ssh/1.8.0/packages/81/ca/b8a41ee0d9753f4ed1c33cb8f37c5aaeb71c12077c7a1b0d9007c83fc72d/ssh-1.8.0.tar.gz/ manually but it's a bit annoying because I keep forgetting what the correct domain for this service is. I'm pretty sure most folks don't even know this exists.

I want a small link/button that would take me to the correspondint inspector page with a single click. This shouldn't be difficult to implement and probably is a `good first issue`, which is why I'm adding this label. Plus it does not seem like a controversial UX improvement.

Trighap52:
Closes #19032

## Summary
- add an Inspector link to the project file details metadata section
- use the same inspector URL pattern already used in admin release detail
- add a functional test to verify the inspector link is rendered for a release file

## Testing
- T=tests/functional/test_project.py make tests

### DIFF
--- a/tests/functional/test_project.py
+++ b/tests/functional/test_project.py
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from http import HTTPStatus
+
+from tests.common.db.packaging import FileFactory, ProjectFactory, ReleaseFactory
+
+
+def test_project_release_file_details_links_to_inspector(webtest):
+    project = ProjectFactory.create(name="sampleproject")
+    release = ReleaseFactory.create(project=project, version="1.2.3")
+    file = FileFactory.create(
+        release=release,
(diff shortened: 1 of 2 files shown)


woodruffw:
I'll let the actual maintainers opine, but I don't think there was consensus in #19032 about this being something that PyPI *wants* to formally expose in the UI -- the inspector is a debugging tool and exposing it directly in PyPI's UI may mislead users about it being a first-class feature of PyPI itself.

PR #19551, opened 2026-02-19
Sort sponsors

JacobCoffee:
Sort by level and then name

### DIFF
--- a/tests/unit/sponsors/test_init.py
+++ b/tests/unit/sponsors/test_init.py
@@ -3,10 +3,8 @@
 import pretend
 
 from celery.schedules import crontab
-from sqlalchemy import true
 
 from warehouse import sponsors
-from warehouse.sponsors.models import Sponsor
 from warehouse.sponsors.tasks import update_pypi_sponsors
 
 from ...common.db.sponsors import SponsorFactory
@@ -47,16 +45,10 @@ def test_do_not_schedule_sponsor_api_integration_if_no_token():
 
 
 def test_list_sponsors(db_request):
-    SponsorFactory.create_batch(5)
+    expected = SponsorFactory.create_batch(5)
     SponsorFactory.create_batch(3, is_active=False)
 
     result = sponsors._sponsors(db_request)
-    expected = (
-        db_request.db.query(Sponsor)
-        .filter(Sponsor.is_active == true())
-        .order_by(Sponsor.level_order, Sponsor.name)
-        .all()
-    )
 
-    assert result == expected
     assert len(result) == 5
+    assert set(result) == set(expected)


JacobCoffee:
if i understnad what you're asking @miketheman the query uses what the task pulls in , i don't think it was being used until this pr

tests/unit/sponsors/test_init.py
```diff
@@ -51,7 +51,12 @@ def test_list_sponsors(db_request):
     SponsorFactory.create_batch(3, is_active=False)
 
     result = sponsors._sponsors(db_request)
-    expected = db_request.db.query(Sponsor).filter(Sponsor.is_active == true()).all()
+    expected = (
+        db_request.db.query(Sponsor)
+        .filter(Sponsor.is_active == true())
+        .order_by(Sponsor.level_order, Sponsor.name)
+        .all()
+    )
```
miketheman: We have an opportunity to improve this test, since previously it reflected the original query, and now it's performing the same query, duplicating the behavior.

Instead, how about casting the first `create_batch` to a variable like `expected` and then we don't have to query a second time?

warehouse/sponsors/__init__.py
```diff
@@ -8,7 +8,12 @@
 
 
 def _sponsors(request):
-    return request.db.query(Sponsor).filter(Sponsor.is_active == true()).all()
+    return (
+        request.db.query(Sponsor)
+        .filter(Sponsor.is_active == true())
+        .order_by(Sponsor.level_order, Sponsor.name)
+        .all()
+    )
```
miketheman, line 16: Setting this explicit order will impact anywhere we use `request.sponsors` and don't re-sort.

warehouse/templates/pages/sponsors.html
```diff
@@ -164,7 +164,7 @@ <h2 class="centered-heading__title">Infrastructure sponsors</h2>
         <hr class="centered-heading__rule">
       </div>
       <div class="sponsor-grid sponsor-grid--bottom-margin">
-        {% for sponsor in request.sponsors | sort(attribute="name") %}
+        {% for sponsor in request.sponsors %}
```
miketheman, line 167: This is only removing the name sort order from the infra sponsors, is that intentional? I suspect you'd want the "main" sponsors section instead / as well.

JacobCoffee: i think this was preferable so that we ensure db-level sortnig for all instead of just infra

warehouse/sponsors/__init__.py
JacobCoffee: i think its just sponsors.html and the sponsors footer (which i cant seem to see that we actually use/render...?) but this is the preferred thing so that we keep consistencies and sponsor expectations on ordering

tests/unit/sponsors/test_init.py
JacobCoffee: [15ad28e](https://github.com/pypi/warehouse/pull/19551/commits/15ad28e9c9e14a0bf0ad22d2439d9f12d6f3417f) wdyt about this

warehouse/sponsors/__init__.py
JacobCoffee: oh i have adblock so i missed these. that'd be a separate issue but i feel like we should do this in a way to not be blocked by ad-blockers

tests/unit/sponsors/test_init.py
miketheman: Better, thanks!

warehouse/sponsors/__init__.py
miketheman: I don't know if you're responding to this comment, or to the one in the HTML template - but it's unclear from the PR description on what the intent here is, so it's hard to review without the context.

It'd help if you viewed the current page without ad blockers, as well as the proposed changes so that you can see the differences. However, the sample sponsor data doesn't contain levels/order - I added that in https://github.com/pypi/warehouse/pull/19553 so that might helpful to merge first and then see how this renders for you, and make the desired changes.

JacobCoffee:
pref https://github.com/pypi/warehouse/pull/19803 since we need to have sponsors supply white logos to go with theme

PR #19584, opened 2026-02-25
feat: scan files on upload for pyarmor

miketheman:
Instead of detection this violation post-upload, triggering external scanners, getting reports, quarantining, handling comms to reporters and end users, reject these at the outset.

I split these into reviewable commits.

This brings back the concept of YARA scanning that was originally brought in #7377 but retired due to too many false positives.

Using the newer, faster `yara-x`, I was able to process a 67MB wheel with 50k Python files in it under 4 seconds (inside a container), so this should be a relatively low impact on `file_upload()` - and there's an Admin Flag to flip scanning off if we find out that it's too strenuous and need to stop doing it.

~~I'm still working out a protection for very large files being uploaded, please review some of the initial logic while I work out the rest.~~ Handled, now reviewable.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -881,6 +881,19 @@ class TestFileUpload:
                 },
                 "keywords: Should not be a tuple.",
             ),
+            # local version error
+            (
+                {
+                    "metadata_version": "1.2",
+                    "name": "example",
+                    "version": "1.0+local",
+                    "md5_digest": "bad",
+                    "filetype": "sdist",
+                },
+                "The use of local versions in '1.0+local' is not allowed. "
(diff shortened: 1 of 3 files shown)


warehouse/utils/scanner.py
```diff
@@ -0,0 +1,279 @@
...
+                logger.info(
+                    "Skipping oversized file in YARA scan",
+                    member=name,
+                    member_size=size,
+                    archive=archive_name,
+                    max_size=_SCAN_MAX_FILE_SIZE,
+                )
+                continue
+
+            if overflow:
```
JacobCoffee, line 131: how does overflow get set here if its false above

JacobCoffee: ahh nvm the looop :D

PR #19593, opened 2026-02-27
refactor: optimize query counts for Organization Accounts page

miketheman:
Currently the page can produce N+1 queries based on how many organization roles exist, making this more painful for larger orgs.

- Refactor to call for permissions once in the base template
- Refactor to query less for sole owners
- Add a functional test to backstop query growth - down from 38 to 13


miketheman:
Thanks to #19544 we can do these kinds of things now!

PR #19349, opened 2026-01-20
feat: add circleci as trusted publisher

meeech:
https://github.com/pypi/warehouse/issues/13888

Related PRs:

https://github.com/pypi/pypi-attestations/pull/166 Add CircleCI to pypi-attestations
https://github.com/meeech/warehouse/pull/1 Stacked pr off this one - integrates attestations. Question: would it be better to just merge this into have the one pr? I was trying to be mindful of making things easier to review. 

I have tried to keep commits bite sized. I am open to split this up into multiple prs if that preferred (though not sure where the 'fault lines' would be for this since its all a bit related - forms, model, etc? let me know

MUST
* we must reject any publish where the ssh-rerun claim is true. like we don't even allow the exchange for an api token to do the publish ✅

while i worked with a LLM to generate this PR, i (the human) have reviewed all this code. (amp for the curious)
tbf it did a good job doing all the boilerplate. 

Basically I try to adhere to https://github.com/ghostty-org/ghostty/blob/main/AI_POLICY.md

PR #19602, opened 2026-03-02
fix(tests): stabilize timeline trend week-boundary assertion

meeech:
👋🏼 hit this last night when I was working on #19349 

This test flaked locally (Sunday ~10pm EST) because relative timestamps could straddle a week boundary and produce two labels instead of one.

Use the most recent Wednesday noon anchor so the test remains in the same week while still falling within the rolling cutoff window used by _fetch_observations.

Amp-Thread-ID: https://ampcode.com/threads/T-019cac83-e053-704b-ab1f-39513bcdce9a


meeech:
@miketheman yeah no problem. thanks for the reference pr. will get that up

meeech:
@miketheman I think this is what you meant? just to avoid using the word `now` and give more meaningful names?
 otherwise let me know - i looked at the pr you referenced, but it looks like datetime.now is still used for the inital value on your times? so I wasn't entirely sure what you meant about static values.

PR #19618, opened 2026-03-04
feat(admin): improve prohibited name release behavior

### LINKED ISSUE #19609: Unclear Input Box in Prohibited Project Name Page UI
The Prohibited Project Page currently has an input box under the Action column. However, it is unclear what input is needed for that input box. Is it asking for the PyPI username of the person releasing the prohibited project name as a record of evidence? Or is it the name of the prohibited project itself? Without admin guidance or diving deep into the codebase, the functionality and input is unclear in the UI and providing clarity would be helpful. 

Some possible resolutions could be (a) provide a hovering tooltip to instruct support/admin what the functionality of the box is such as "Type new project owner name" (as it turns out that is the input that the box is looking for in order to release the package name) or (b) add a label under the input box indicating what input is needed. 

---

Image

miketheman:
Instead of a relatively opaque set of actions, improve the overall behavior with icons.

Each icon now has a label, a tooltip, and pops open a modal that explains a little more and provides the operator more context and an opportunity to confirm/cancel.

Resolves #19609

### DIFF
--- a/warehouse/admin/templates/admin/prohibited_project_names/list.html
+++ b/warehouse/admin/templates/admin/prohibited_project_names/list.html
@@ -54,23 +54,75 @@
         <td>{{ prohibited_project_name.comment }}</td>
         <td>{{ prohibited_project_name.observation_kind }}</td>
         <td>
-          <form method="POST" action="{{ request.route_path('admin.prohibited_project_names.remove') }}">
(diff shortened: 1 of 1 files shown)


miketheman:
Some screenshots from dev:

image

Confirmation modals:

image

Screenshot 2026-03-04 at 12 12 32

Issue #17444, opened 2025-01-17
Upgrade JavaScript dependency `eslint` to 9


miketheman:
The current version of eslint 8 uses the `eslintConfig` from `package.json` and is incompatible with eslint 9.

Read, then follow relevant steps of the [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide) - some may not be simple 1-to-1 steps.

The overall outcome is:

- `package.json` is updated for the right eslint version
- `package.json` no longer has any `eslint` directives in it
- `package-lock.json` only contains only the necessary updates
- an new config file has the config directives
- all lints and tests pass

---

**Good First Issue**: This issue is good for first time contributors. If you've already contributed to Warehouse, work on [another issue without this label](https://github.com/pypi/warehouse/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+-label%3A%22good+first+issue%22) instead. If there is not a corresponding pull request for this issue, it is up for grabs. For directions for getting set up, see our [Getting Started Guide](https://warehouse.pypa.io/development/getting-started/).

If you are working on this issue and have questions, feel free to ask them here, in the [PyPA Discord](https://discord.gg/pypa) or on the [Discourse](https://discuss.python.org/c/packaging/14).

D3athSkulll:
Hey, Can I solve this issue?

D3athSkulll:
Hey , I am willing to work on this issue, can you please allocate it to me?

miketheman:
> Hey , I am willing to work on this issue, can you please allocate it to me?

Feel free to submit a PR, we do not routinely assign issues to non-members.

D3athSkulll:
I have made the required changes and made a esLint_Upgrade branch in my forked repo and pushed the code there. How do i submit a PR ?

D3athSkulll:
Hey I have submitted a PR, can you pls verify it.

D3athSkulll:
Hey ,can you please verify my submitted pr?

miketheman:
Hi @D3athSkulll , I reviewed the PR, and added a bunch of notes.
Please consider that not everyone is working as the same time as you, and we all have different response times and commitments.

Ivan-houzhiwen:
Hi! I would like to work on this issue. I've created a PR #19626 that upgrades ESLint from v8 to v9 with the flat config format. Please let me know if you have any feedback!

PR #19626, opened 2026-03-05
chore: upgrade eslint from v8 to v9

Ivan-houzhiwen:
This PR upgrades ESLint from version 8 to version 9, addressing issue #17444.

Changes:
1. Updated eslint from ^8.36.0 to ^9.0.0
2. Migrated config to eslint.config.js (flat config)
3. Configured @stylistic/js plugin for flat config

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "copy-webpack-plugin": "^12.0.2",
     "css-loader": "^7.1.2",
     "css-minimizer-webpack-plugin": "^7.0.0",
-    "eslint": "^8.36.0",
+    "eslint": "^9.0.0",
     "gettext-parser": "^8.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "copy-webpack-plugin": "^12.0.2",
     "css-loader": "^7.1.2",
     "css-minimizer-webpack-plugin": "^7.0.0",
-    "eslint": "^8.36.0",
+    "eslint": "^9.0.0",
```
miketheman, line 57: Looks like we're missing an update to package-lock.json after this, did you forget to add it to the commit scope?

miketheman (changes requested):
Thanks for a good start! I think the lockfile update is missing, preventing a proper build

PR #19628, opened 2026-03-05
fix: require URL equality for retroactive URL verification

miketheman:
Retroactive verification of home_page and download_url on existing releases
only checked whether the new upload's URL was verified,
not whether it matched the URL already stored on the release.
A second upload with a different, legitimately verifiable URL
could grant a verified badge to the original unrelated URL.

Add the same equality check that project_urls already uses.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -494,6 +494,29 @@ class TestFileValidation:
             None,
         )
 
+    def test_tarfile_zipfile_polyglot(self, tmpdir):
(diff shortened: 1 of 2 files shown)


miketheman:
I spent some time to manually verify the cases where this would change an existing behavior, and all of them (20 projects in totla) happen to be in the cases of:

- user used to use manual uploads, now uses Trusted Publishing
- user used to use Trusted Publishing, now uses manual uploads
- user used Trusted Publishing for some of the files in a release, manual for others

PR #19662, opened 2026-03-10
feat: allow new user to change unverified email

### LINKED ISSUE #18079: Add ability for users to change typo-ed email addresses before verification
There have been a ton of issues/cases where users have accidentally typo-ed their primary email addresses and ended up not able to verify their email addresses since it's a typo one. (For e.g. thespibrain@gmil.com instead of thespibrain@gmail.com). This has created a weird loop where a user needs to verify their primary address in order to add a second correct one, but because their first email address is a typo, they are unable to verify. 

Creating some kind of a "check" for users before verification to make sure that the email address they typed is the one they want will be super helpful to users, (e.g. a "do you want to add thespibrain@gmil.com" msg maybe) so that users can double check that what they typed is exactly what they intended as their email address. 

**Additional context**
The problem is not limited to mistyping the domain part of the email address, based on previous issues I have worked on. So, adding a domain check won't be enough, as users have also typo-ed other parts of their email addresses.

miketheman:
During registration, if the user selected an unverifiable-by-them email address, they get stuck and may have to contact support to continue.

By allowing a user with no projects, no 2FA setup, and otherwise "nothing" account to change their unverified email address to a new one, we reduce the support ticket burden and allow the user to self-service this condition.

Resolves #18079

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -100,7 +100,7 @@ msgid ""
 msgstr ""
 
 #: warehouse/accounts/forms.py:422 warehouse/manage/forms.py:132
-#: warehouse/manage/forms.py:820
+#: warehouse/manage/forms.py:824
 msgid "The name is too long. Choose a name with 100 characters or less."
 msgstr ""
 
@@ -164,7 +164,7 @@ msgstr ""
 msgid "Successful WebAuthn assertion"
 msgstr ""
 
-#: warehouse/accounts/views.py:631 warehouse/manage/views/__init__.py:860
+#: warehouse/accounts/views.py:631 warehouse/manage/views/__init__.py:948
 msgid "Recovery code accepted. The supplied code cannot be used again."
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


warehouse/templates/manage/unverified-account.html
```diff
@@ -235,7 +235,7 @@ <h1 class="page-title">{{ title }}</h1>
 <p>
   {% trans help_url=help_url %}
   If you cannot verify the primary email address listed,
-  please see <a href="{{ help_url }}">{{ help_url }}</a>
+  you can correct it below or see <a href="{{ help_url }}">{{ help_url }}</a>
```
JacobCoffee: if we provide the ability to self-service, unless we HAVE to have help_url for some reason arguably we should remove the URL here so that we reduce support case loads

miketheman: Fair enough, we can always restore it if we get too many cases of "I didn't know how to contact you"

Issue #787, opened 2015-11-20
Implement/Integrate statistics


dstufft:
The new design has a pane for statistics, however we need to actually implement and integrate them.

nlhkabu:
This tab will be removed for launch (see #1310) - we should consider this a long term 'todo' :)

For reference, this is the design.

![screenshot from 2016-06-30 07-15-57](https://cloud.githubusercontent.com/assets/3323703/16478615/8a72833a-3e92-11e6-89c9-bca36a6b8cf5.png)

reiinakano:
Hello guys, I know you're all super busy and not getting paid for this, but I just want to ask whether you have an ETA? The last activity on the issue seems to be more than a year ago.

No rush though. Thanks for all your work! :)

brainwane:
Related: #3400 & #3434 on https://libraries.io/pypi/ and https://packaging.python.org/guides/analyzing-pypi-package-downloads/ . The right thing to do here, like in #991, might be to aid @andrew in getting download stats into libraries.io (https://github.com/librariesio/libraries.io/issues/1916), then pull a selected stat or two into the display in Warehouse but still point people to libraries.io and the Google BigQuery dataset for more details.

cc @a-wakeel and @micurd who opened related issues.

andrew:
Related issue on the Libraries.io repo about download stats, will be put back on the roadmap within a few months I expect: https://github.com/librariesio/libraries.io/issues/53

di:
https://github.com/pypa/pypi-legacy/issues/254 says:

>as a package owner/maintainer, it'd be fantastic if there was any way to get something more fine-grained than "downloads per month/week/day".
>
>i recognize that privacy concerns that put some limitations on here, but a few things i think might still be doable, in priority order:
>
>1/ what package install led to me being installed?
>
>this one would be huge -- if i'm maintaining a package that's a common dependency for several packages, which ones are driving my installs? in particular, if i need to change/deprecate something, which dependencies am i most likely to break?
>
>2/ breakdowns by OS/python version
>
>the biggest win here would of course be py2 vs. py3 -- how many of my users have already switched? on top of that, OS information (even just linux/mac/win) would be good, esp. around testing priorities.
>
>3/ support for user-provided data in the install request
>
[...]

PR #19663, opened 2026-03-10
Link to pypistats on a project

JacobCoffee:
Adds a link to see the project stats for the PSF-hosted `pypistats` package:

image


miketheman:
See #787 for existing conversations on the topic

JacobCoffee:
@miketheman Can this just be a step in that direction? Linking to them, that is. instead of pulling them. We can work towards that

miketheman:
Possibly? I currently distrust the download counts as they are, and they are trivially inflatable, so linking directly to them adds a degree of "we bless this source of data with these results".
See the note here: https://pypi.org/stats/#:~:text=Third%2Dparty%20statistics

That's one reason (I believe) we haven't done so in the past, but that was also before the PSF adopted running pypistats. I still don't know the governance/ownership model, and there's a few open issues on the repo pointing out inaccuracies.

I think the topic of whether to link this service should be discussed over in that issue to get a better set of perspectives.

PR #19684, opened 2026-03-13
fix: isolate advisory lock from other flush operations to prevent deadlocks

miketheman:
`ensure_monotonic_journals` acquired a global advisory lock in `before_flush`, then the same flush executed all pending INSERTs/UPDATEs.
When a `File` INSERT's unique constraint check waited on a concurrent transaction that was itself waiting for the advisory lock, a deadlock cycle formed.

Refactor the listener to automatically defer `JournalEntries` when other objects are also pending. The `before_flush` handler expunges them, and an `after_flush` handler restores them for a subsequent flush cycle where they're the only pending objects.
This ensures the advisory lock is never held while unrelated operations execute.

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -1120,17 +1120,41 @@ def ensure_monotonic_journals(config, session, flush_context, instances):
     # The way this works, not even the SERIALIZABLE transaction types give
     # us this property. Instead we have to implement our own locking that
     # ensures that each new journal entry will be serialized.
-    for obj in session.new:
-        if isinstance(obj, JournalEntry):
-            session.execute(
-                select(
-                    func.pg_advisory_xact_lock(
-                        cast(cast(JournalEntry.__tablename__, REGCLASS), Integer),
-                        _MONOTONIC_SEQUENCE,
-                    )
-                )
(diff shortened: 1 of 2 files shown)


warehouse/packaging/models.py
```diff
@@ -1120,17 +1120,41 @@ def ensure_monotonic_journals(config, session, flush_context, instances):
...
+    if has_other_pending:
+        # This flush contains both JournalEntries and other pending changes.
+        # Acquiring the advisory lock here would hold it while non-journal
+        # INSERTs/UPDATEs execute (e.g., File INSERT unique constraint checks),
+        # which can deadlock with concurrent transactions waiting for the same
+        # advisory lock. Defer the JournalEntries to a subsequent flush where
+        # they'll be the only pending objects, minimizing lock hold scope.
+        for je in journal_entries:
+            session.expunge(je)
+        session.info.setdefault("_deferred_journals", []).extend(journal_entries)
```
JacobCoffee, line 1139: sorry this is hard for me to grok, but do we dump the session if we roll back or would this make it so that we get stale journals in reused sessions?

miketheman: Sessions are per-request, not pooled, so they cannot be reused as part of another session. Is that what you mean by "dump"?

PR #19695, opened 2026-03-16
feat(admin): improve project search and malware handling

miketheman:
A few improvements to the admin UI.

- allow searching projects by UUID
- improve projects main page with some stats and remove heavy paginated list loading
- enable closing of orphaned malware reports

### DIFF
--- a/tests/unit/accounts/test_services.py
+++ b/tests/unit/accounts/test_services.py
@@ -37,6 +37,7 @@ from warehouse.accounts.models import (
     DisableReason,
     ProhibitedUserName,
     TermsOfServiceEngagement,
+    User,
     UserTermsOfServiceEngagement,
 )
 from warehouse.events.tags import EventTag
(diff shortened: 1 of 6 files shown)


warehouse/admin/views/malware_reports.py
```diff
@@ -320,7 +319,6 @@ def not_malware_for_project(request):
     """
     observation_id = request.matchdict.get("observation_id")
     observation = request.db.get(Observation, observation_id)
```
JacobCoffee, line 321: could this ever b enull? or does get here do like get_or_404? or should we raise HTTPNotFound or something if missing

warehouse/admin/views/projects.py
```diff
@@ -45,6 +48,74 @@
 def project_list(request):
     q = request.params.get("q")
 
+    if q and q.startswith("id:"):
+        try:
+            project_id = uuid.UUID(q[3:])
```
JacobCoffee: this isnt used until like ~120, would it be bad to move one of them closer to each other?

warehouse/admin/views/malware_reports.py
miketheman: It's possible, but unlikely, since the route is linked to from the Admin UI, and provides the inbound `observation_id`.

For admin routes, we're often a little less guarded, as can be seen by most of the other views in this module and others.

An improvement could be to have a `factory` pattern on the route, which pre-hydrates the object from the DB before entering the view, like we do for Users:

https://github.com/pypi/warehouse/blob/584505583701038f8ee20f39ceb6285743163734/warehouse/admin/routes.py#L114-L120
https://github.com/pypi/warehouse/blob/837de24849af4af7b8819ac8d24fedad353a0195/warehouse/accounts/models.py#L52-L60

That's a larger refactor for items outside the scope of this change.

warehouse/admin/views/projects.py
miketheman: Good idea, moving it!

PR #19698, opened 2026-03-16
feat: record new device attempts

miketheman:
When a user authenticates with correct credentials (password + TOTP/recovery-code) from a new IP address, device_is_known() returns False, redirects to email confirmation, but records no event.

The `LoginSuccess` event only fires later when the user confirms via email — and if they never confirm, there's no audit trail at all.

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -164,181 +164,181 @@ msgstr ""
 msgid "Successful WebAuthn assertion"
 msgstr ""
 
-#: warehouse/accounts/views.py:631 warehouse/manage/views/__init__.py:948
+#: warehouse/accounts/views.py:633 warehouse/manage/views/__init__.py:948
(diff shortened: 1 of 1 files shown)


warehouse/events/tags.py
```diff
@@ -64,6 +64,7 @@ class Account(EventTagEnum):
         EmailReverify = "account:email:reverify"
         EmailVerified = "account:email:verified"
         LoginFailure = "account:login:failure"
+        LoginNewDevice = "account:login:new_device"
```
JacobCoffee, line 67: how do these tags work?

miketheman: We refer to the tag like `EventTag.Account.LoginNewDevice` and the string's value is what's persisted in database records. The base class has some decent inline documentation here: https://github.com/pypi/warehouse/blob/0fad09ad3cc9eca3a9f5baaea28eca59bfbc7303/warehouse/events/tags.py#L6-L31

PR #19705, opened 2026-03-18
Tạo pylint.yml

cauba0978:
Chạy dự án


miketheman:
It's unclear what this is doing, and I'd advise you to open an issue to discuss the change you want to have happen before trying this again.

PR #19708, opened 2026-03-18
feat(dev): add traefik in front of services

miketheman:
Mirrors production in that we can now expose all of the web-like services via the same machinery.
web, camo, files can all be accessed directly via `localhost` and the appropriate paths.

### DIFF
--- a/dev/environment
+++ b/dev/environment
@@ -27,7 +27,7 @@ SESSION_SECRET="an insecure development secret"
 BILLING_BACKEND=warehouse.subscriptions.services.MockStripeBillingService api_base=http://stripe:12111 api_version=2020-08-27 domain=localhost
 # BILLING_BACKEND=warehouse.subscriptions.services.StripeBillingService api_version=2020-08-27 publishable_key=pk_test_123 secret_key=sk_test_123 webhook_key=whsec_123 domain=localhost
 
-CAMO_URL={request.scheme}://{request.domain}:9000/
+CAMO_URL={request.scheme}://camo.localhost/
 CAMO_KEY=insecurecamokey
 
 # Example of a Live Reload URL for remote development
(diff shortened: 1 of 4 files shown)


dev/environment
```diff
@@ -27,7 +27,7 @@ SESSION_SECRET="an insecure development secret"
 BILLING_BACKEND=warehouse.subscriptions.services.MockStripeBillingService api_base=http://stripe:12111 api_version=2020-08-27 domain=localhost
 # BILLING_BACKEND=warehouse.subscriptions.services.StripeBillingService api_version=2020-08-27 publishable_key=pk_test_123 secret_key=sk_test_123 webhook_key=whsec_123 domain=localhost
 
-CAMO_URL={request.scheme}://{request.domain}:9000/
+CAMO_URL={request.scheme}://camo.localhost/
```
dstufft, line 30: I have a weird environment where I'm running warehouse on top of Ubuntu that is running inside of a WSL2 container on my Windows desktop, so I'm not sure if this is a weirdness related to my setup or if it's a general issue, so just mentioning it.

Does this actually work? When I try to resolve ``camo.localhost`` my DNS resolver does not resolve it to anything, so none of these actually work for me. I haven't actually tried this branch, so I'm not sure if maybe docker is doing something to automatically add those hosts (but I can't see how it would be)?

Obviously I could adjust my `/etc/hosts` to alias ``*.localhost`` to ``127.0.0.1``, which I don't mind doing, but if it doesn't work in general without that (and it's not just my weird environment) then we might want to consider a different approach? At a minimum we should at least document needing to do that?

miketheman: I don't think you're meant to try to navigate to camo.localhost yourself any longer - rather the proxy handles that for you transparently

dstufft: Maybe I'm misunderstanding what this is doing, but isn't traefik routing based on hostname, so I need to visit ``camo.localhost`` so that my browser sets ``Host: camo.localhost``, so that traefik can see the hostname and route to the correct backend?

miketheman: Hmm, you're right - I'll need to think about this more. Taking it back to draft, will continue to poke. Thanks for testing!

miketheman: I think I sussed it out now - please feel free to take another pass!

JacobCoffee (approved):

PR #19710, opened 2026-03-19
Fix 2FA page columns overflowing at high display scale

### LINKED ISSUE #18270: 2FA page is too wide
**Describe the bug**
Here's what the PyPI 2FA page looks like for me:

Image

The text on the right side is cut off. It might not be obvious from the screenshot, but the word "Enter" in "Enter this code" cannot be seen, and the 2FA entry textbox extends past the edge of the screen.

Also, it just looks weird. Even though authenticating with a USB device and with a code are equivalent operations, the first column takes up 3/4 of the horizontal space.

**Expected behavior**
All content should be visible without horizontal scrolling.

Ideally, the two columns should be of the same width.

**To Reproduce**
Add an authentication app and a security key to your account. Log in.

**My Platform**
I'm using Firefox on Windows 11. My screen resolution is 1920x1080, and my display scale level is 150%.

**Additional context**
N/A

Redslayer112:
Closes #18270

The first column of the 2FA page was using `flex-shrink: 0`, preventing
it from shrinking to share space with the second column. At high display
scale factors this caused the second column to overflow off screen.

Replaced with `flex: 1` on both columns so they share horizontal space
equally regardless of content size or display scale.

Note: reproducing this bug locally requires an account with both a security
key and an authenticator app configured. The fix is a one-line CSS change
replacing `flex-shrink: 0` on the first column with `flex: 1` on both columns.


warehouse/static/sass/blocks/_twofa-login.scss
```diff
@@ -1,5 +1,4 @@
 /* SPDX-License-Identifier: Apache-2.0 */
-
```
miketheman, line 2: Curious - why did the whitespaces get removed? Are you using a linter/formatter that the project doesn't use?

Redslayer112:
Sorry about that — my editor's auto-formatter stripped those blank lines. I've restored them in the force-push; the diff should now only contain the intentional changes.

PR #19727, opened 2026-03-24
Restrict publishing new artifacts to releases after 14 days

### LINKED ISSUE #19726: Restrict new uploads to existing releases after a certain amount of time

Once API tokens are found by attackers, they can publish new artifacts to existing releases making the cleanup process more time consuming for maintainers and successfully attacking users that aren't using lock files with checksums.

After a certain amount of time, disallow future uploads to an existing release.

**Additional context**

This has been discussed here: https://discuss.python.org/t/restricting-open-ended-releases-on-pypi/43566

sethmlarson:
Closes https://github.com/pypi/warehouse/issues/19726

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -76,8 +76,7 @@ from warehouse.utils.wheel import (
 PATH_HASHER = "blake2_256"
 
 # After a release has been published for this
-# number of days, send a warning message to
-# users that this may be rejected in the future.
+# number of days reject new uploaded files.
 MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS = 14
 MAXIMUM_AGE_FOR_NEW_UPLOADS_SECONDS = datetime.timedelta(
     days=MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
+import datetime
 import hashlib
 import hmac
 import os.path
@@ -73,6 +74,14 @@ from warehouse.utils.wheel import (
 )
 
 PATH_HASHER = "blake2_256"
+
+# After a release has been published for this
+# number of days, send a warning message to
+# users that this may be rejected in the future.
+MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS = 14
+MAXIMUM_AGE_FOR_NEW_UPLOADS_SECONDS = datetime.timedelta(
+    days=MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS
+).total_seconds()
 
 COMPRESSION_RATIO_MIN_SIZE = 64 * ONE_MIB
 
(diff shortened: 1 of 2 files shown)


warehouse/forklift/legacy.py
```diff
@@ -1072,6 +1078,22 @@ def file_upload(request):
         #       at least this should be some sort of hook or trigger.
         _sort_releases(request, project)
 
+    # Check if the release should be closed to
+    # new artifacts due to age.
+    if not is_new_release and release.created is not None:
```
dstufft: Note, you can sort of get around this by deleting the release and recreating it.

That's a heavy hammer though, as you won't be able to re-upload files with the same name again, but that's trivial to work around in the wheel case, but sdists are much harder.

sethmlarson: Hmm, this is good to know. I didn't realize we did hard deletes of `Release` for user-triggered deletes. Any suggestions how to implement this without creating another database model?

Not to leave a trap for us later for when this changes, but API tokens today don't allow deleting releases, so in a way it's still effective against token compromises.

dstufft: Thinking about this more, sdists might not be much harder. I know that most of the tooling normalizes versions, but I don't recall if the filename check on PyPI normalizes versions or not. If it does not, then working around it for sdists would be easy too, you just modify your filename to add a `.0` to the version and upload that.

Different filename (assuming we don't normalize), so it'd be allowed, but all the consuming tooling will treat it as equivalent. 

To answer your question. I don't think there is a way without adding another database model. That's what I had to do for the filename registry, and I can't think of a different option for releases. We don't have soft deletes in PyPI, so deletion is "real", including deleting the whole project and recreating it. Anything that we want to survive past that has to be on a different table that is setup not to cascade on delete (and otherwise handles deletions somehow-- often times by not using foreign keys / relational data but using the "natural" key and manually querying the model to see).

warehouse/forklift/legacy.py
```diff
@@ -72,6 +73,11 @@
 
 PATH_HASHER = "blake2_256"
 
+# After a release has been published for this
+# number of days, restrict publishing new
+# artifacts to the same release.
+MAXIMUM_RELEASE_AGE_IN_SECS_BEFORE_CLOSING = 14 * 24 * 60 * 60  # 14 days in seconds.
```
miketheman:
```suggestion
MAXIMUM_RELEASE_AGE_IN_SECS_BEFORE_CLOSING = datetime.timedelta(days=14).total_seconds()
```

warehouse/forklift/legacy.py
sethmlarson: I've opted to focus on rejecting uploads using the `Release.created` datetime instead of using the journal or a separate model. This approach is much simpler and easier to test and get "right".

This changes this PR to primarily protect against compromised tokens or workflows instead of a compromised account. This is the bulk of watering hole attacks that we see, so this should be of benefit still.

warehouse/forklift/legacy.py
```diff
@@ -74,6 +75,14 @@
 
 PATH_HASHER = "blake2_256"
 
+# After a release has been published for this
+# number of days, send a warning message to
```
woodruffw: Comment here might be out of date? It says warning message, but we fail hard below 🙂

sethmlarson: Yep, I'll fix this.

sethmlarson: https://github.com/pypi/warehouse/pull/19727/commits/3141730fcdce9a3b6d6b5d0765c91d91319eb21f

ewdurbin (approved):
Love the straightforward approach! Covers all concerns I would have (emitting a metric specifically).

As a policy decision I am less confident, so I'd love for more of the @pypi/warehouse-admins to weigh in!

warehouse/forklift/legacy.py
```diff
@@ -1054,6 +1062,24 @@ def file_upload(request):
...
+    # users deleting and recreating releases in the UI, only
+    # against uploads through compromised API tokens or workflows.
+    oldest_release_age_allowed = datetime.datetime.now() - datetime.timedelta(
+        seconds=MAXIMUM_AGE_FOR_NEW_UPLOADS_SECONDS
+    )
+    if release.created < oldest_release_age_allowed:
+        request.metrics.increment(
+            "warehouse.upload.failed", tags=["reason:closed-release"]
+        )
+        raise _exc_with_message(
```
ewdurbin, line 1077: This will notify the uploader, but not necessarily be detectable by other maintainers. Should we consider a Project event that stores the attempt to append a release in the log?

miketheman: I think that's a good idea for a follow on if we ever see this become a confusion point

warehouse/forklift/legacy.py
```diff
@@ -74,6 +75,13 @@
 
 PATH_HASHER = "blake2_256"
 
+# After a release has been published for this
+# number of days reject new uploaded files.
+MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS = 14
+MAXIMUM_AGE_FOR_NEW_UPLOADS_SECONDS = datetime.timedelta(
+    days=MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS
+).total_seconds()
```
miketheman, line 83: Really like the separation and reuse, so feasibly this could become a variable that can be adjusted with one value.

miketheman (approved):
Let's go!

Issue #19752, opened 2026-03-28
Incorrect documentation of passkey support


jayqi:
**Describe the bug**

As part of the unrecognized device flow while logging in today, I saw the following message:

> In the future, you can automatically trust new devices without needing email verification by using a security device or passkey as your 2FA method.

From a [quick search for "passkey"](https://github.com/search?q=repo%3Apypi%2Fwarehouse%20passkey&type=code) in the codebase, this is indeed documented in several templates and also in the [November 14, 2025 blog post](https://blog.pypi.org/posts/2025-11-14-login-verification/).

I was quite excited to see that I could set up a passkey, only to not actually find any way to set up a passkey. Based on https://github.com/pypi/warehouse/issues/13764, it seems like there actually isn't any passkey support? 

**Expected behavior**

If passkeys are not actually supported, the UI copy and blog post should not claim that users can use passkeys.

PR #19762, opened 2026-03-30
Remove misleading passkey reference

Mephimeow:
Fixes issue: removes passkey mention from unrecognized device templates since passkeys are not actually supported (only WebAuthn/security device is available).


Mephimeow:
https://github.com/pypi/warehouse/issues/19752 this issue

Mephimeow:
Added missing body.txt template fix

PR #19778, opened 2026-04-01
Use direct SQL UPDATE for macaroon `last_used` to avoid deadlocks

JacobCoffee:
Remove macroon dirty state from play to see if we fix deadlocking from https://python-software-foundation.sentry.io/issues/6960944491/?project=51408&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h

### DIFF
--- a/warehouse/macaroons/services.py
+++ b/warehouse/macaroons/services.py
@@ -144,15 +144,15 @@ class DatabaseMacaroonService:
             # skipping is okay as last_used value is imprecise (python dt, not DB clock)
             # and informational in the ui
             self.db.execute(
-              update(Macaroon)
-              .where(
-                  Macaroon.id.in_(
-                      select(Macaroon.id)
-                      .where(Macaroon.id == dm.id)
-                      .with_for_update(skip_locked=True)
-                  )
-              )
-              .values(last_used=datetime.datetime.now())
+                update(Macaroon)
+                .where(
+                    Macaroon.id.in_(
+                        select(Macaroon.id)
+                        .where(Macaroon.id == dm.id)
+                        .with_for_update(skip_locked=True)
+                    )
+                )
+                .values(last_used=datetime.datetime.now())
             )
 
             return True


warehouse/macaroons/services.py
```diff
@@ -133,7 +134,15 @@ def verify(self, raw_macaroon: str, request, context, permission) -> bool:
 
         verified = caveats.verify(m, dm.key, request, context, permission)
         if verified:
-            dm.last_used = datetime.datetime.now()
+            # Use a direct UPDATE instead of ORM attribute mutation to avoid
+            # marking the macaroon dirty. A dirty macaroon causes autoflush
+            # during Project.__acl__() evaluation, which can deadlock with the
+            # advisory lock held by ensure_monotonic_journals.
+            self.db.execute(
```
ewdurbin, line 146: I'm feeling a lil un-easy about this. I'm not sure if we consider this breaking our transaction model... the macaroon _was_ used but we'll never have an associated event I don't think if something goes wrong. I don't know how much (if at all) we rely on the accuracy of this value.

JacobCoffee: wdyt of [02e97da](https://github.com/pypi/warehouse/pull/19778/commits/02e97dadfb35d74221fb44601ba65e6bfdad60fd)

warehouse/macaroons/services.py
```diff
@@ -133,7 +134,26 @@ def verify(self, raw_macaroon: str, request, context, permission) -> bool:
...
+                select(Macaroon.id)
+                .where(Macaroon.id == dm.id)
+                .with_for_update(skip_locked=True)
+            ).scalar()
+            if row:
+                self.db.execute(
+                    update(Macaroon)
+                    .where(Macaroon.id == dm.id)
+                    .values(last_used=datetime.datetime.now())
+                )
```
provinzkraut: I think you should be able to merge these into just one statement, something like:

```suggestion
            self.db.execute(
              update(Macaroon)
              .where(
                  Macaroon.id.in_(
                      select(Macaroon.id)
                      .where(Macaroon.id == dm.id)
                      .with_for_update(skip_locked=True)
                  )
              )
              .values(last_used=datetime.datetime.now())
            )

```
saves you one DB roundtrip. It should still be safe, as the locked rows are determined by the result of the subquery; at least that's my understanding of how the `SELECT` and `UPDATE` work together here:

> When a locking clause appears in a sub-SELECT, the rows locked are those returned to the outer query by the sub-query.

https://www.postgresql.org/docs/18/sql-select.html#SQL-FOR-UPDATE-SHARE

warehouse/macaroons/services.py
```diff
@@ -133,7 +134,27 @@ def verify(self, raw_macaroon: str, request, context, permission) -> bool:
...
+            # which can deadlock with the journal advisory lock under
+            # concurrent uploads. SKIP LOCKED ensures that if another
+            # transaction is already updating this row (same token used
+            # concurrently), we just skip — the other transaction will
+            # set last_used anyway.
+            # skipping is okay as last_used value is imprecise (python dt, not DB clock)
+            # and informational in the ui
+            self.db.execute(
+              update(Macaroon)
+              .where(
```
dstufft: Can’t this be simplified to just be Macaroon.id == dm.id?

I think this is currently doing a sub select to select s single id… that we already have?

dstufft (approved):
Small change, but overall looks good to me.

warehouse/macaroons/services.py
JacobCoffee: doesnt skip locked require a select stmt here? maybe i am misunderstanding

provinzkraut: Yeah, I suggested this change because you can't (to my knowledge) use a `SKIP LOCKED` directly on the `UPDATE`, which was the original intent here. This sort of emulates that, by trying to acquire the lock in a subquery, and then using the queried ID for the actual `UPDATE` clause. If the lock can be acquired in the subquery, it will return the one ID, propagate the lock to the outer `UPDATE`, where the condition will then match. If the subquery can't acquire a lock, it will return an empty row set, so the outer `UPDATE` will be skipped.

JacobCoffee: reading the pg docs i think this is right @dstufft but lmmk if you want me to simplify it post-merge with a followup PR

dstufft: Ah right.

That's what I get for reviewing things on my phone :)

PR #19785, opened 2026-04-02
litellm blog: fix typos

JelleZijlstra:


docs/mkdocs-blog.yml
```diff
@@ -35,6 +35,8 @@ markdown_extensions:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
+  - pymdownx.highlight:
+      use_pygments: false
```
JelleZijlstra, line 39: This is what makes fenced code blocks in the blog post work. If we don't add `use_pygments: false` it crashes. Codex tells me upgrading pymdown-extensions may fix this but I'll consider that out of scope for now.

JelleZijlstra: Actually seems like the fix is not in a released version of pymdown-extension yet.

JelleZijlstra: To clarify, this is my understanding:
* The current state means we get nice fenced blocks, but no syntax highlighting (from pygments)
* Currently if we set use_pygments to true rendering crashes, because pymdownx is passing a None title to pygments and pygments assumes it's a str.
* This is fixed in the latest pymdownx master but not released yet.
* Once it's released we can upgrade pymdownx, turn on pygments here and get syntax highlighting

(This is mostly from what Codex told me.)

PR #19794, opened 2026-04-04
Added dependencies for uv.

LastWindBender:
Added dependencies for uv.
Add url_icon filter for dynamic URL icons in templates.

TODO:
Dockerfile implementation can be done.
Ruff implementation can be done


pyproject.toml
```diff
@@ -1,3 +1,10 @@
+[project]
+name = "warehouse"
+version = "1.0.0"
+description = "pypi warehouse"
+readme = "README.rst"
+requires-python = ">=3.13.12"
```
LastWindBender:
```suggestion
requires-python = ">=3.13.12"
dependencies = [
    "ruff>=0.15.9",
]
```

uv.lock
```diff
@@ -0,0 +1,5030 @@
...
+    { url = "https://files.pythonhosted.org/packages/ed/76/104b2aa106a208da8b17a2fb72e033a5a9d7073c68f7e508b94916ed47a9/markupsafe-3.0.3-cp313-cp313-musllinux_1_2_riscv64.whl", hash = "sha256:795e7751525cae078558e679d646ae45574b47ed6e7771863fcc079a6171a0fc", size = 21588, upload-time = "2025-09-27T18:36:48.82Z" },
+    { url = "https://files.pythonhosted.org/packages/b5/99/16a5eb2d140087ebd97180d95249b00a03aa87e29cc224056274f2e45fd6/markupsafe-3.0.3-cp313-cp313-musllinux_1_2_x86_64.whl", hash = "sha256:8485f406a96febb5140bfeca44a73e3ce5116b2501ac54fe953e488fb1d03b12", size = 23041, upload-time = "2025-09-27T18:36:49.797Z" },
+    { url = "https://files.pythonhosted.org/packages/19/bc/e7140ed90c5d61d77cea142eed9f9c303f4c4806f60a1044c13e3f1471d0/markupsafe-3.0.3-cp313-cp313-win32.whl", hash = "sha256:bdd37121970bfd8be76c5fb069c7751683bdf373db1ed6c010162b2a130248ed", size = 14543, upload-time = "2025-09-27T18:36:51.584Z" },
+    { url = "https://files.pythonhosted.org/packages/05/73/
```
LastWindBender:
```suggestion
[package.metadata]
requires-dist = [{ name = "ruff", specifier = ">=0.15.9" }]
```

miketheman:
I don't think we're currently interested in migrating to uv.

For work like this, please open an issue to discuss prior to working on implementation, to avoid wasted effort.

JacobCoffee:
For posterity we did end up discussing and have since moved ~fully to ruff and may look at something that helps DX like pixi or uv for quicker development local loops but this is something the team will PR @LastWindBender so if you see another PR to do similar work then here is your reasoning

LastWindBender:
I saw #19953 PR. You've contributed more than I have. If there's anything I can do, I'll add it to the PR.

PR #19803, opened 2026-04-06
Rework footer, Use Sponsor Logos

JacobCoffee:
To be done with https://github.com/python/pythondotorg/pull/2979.

We are required to display sponsor logos in the PyPI footer per contract obligations. We can continue to display infrastructure/in-kind sponsors below them, though.

~~Latest round of edits also moves the language switcher into a dropdown and moves it to the header (easier/better UX for people that need to swap right away)~~ Moved to https://github.com/pypi/warehouse/pull/20400

reorganizes the footer to not be so tall

image

Closes PYPI-19


warehouse/sponsors/tasks.py
```diff
@@ -50,8 +50,11 @@ def update_pypi_sponsors(request):
         sponsor.service = sponsor_info["description"]
         sponsor.link_url = sponsor_info["sponsor_url"]
         sponsor.color_logo_url = sponsor_info["logo"]
+        if sponsor_info.get("white_logo"):
+            sponsor.white_logo_url = sponsor_info["white_logo"]
```
miketheman: Wouldn't this be more succinct as:

```suggestion
        sponsor.white_logo_url = sponsor_info.get("white_logo")
```

since the template already handles with `if sponsor.white_logo_url...`?

warehouse/templates/includes/sponsors-footer.html
```diff
@@ -14,28 +14,36 @@
...
+         rel="noopener"
+         href="{{ sponsor.link_url }}">
+        {% if sponsor.white_logo_url %}
+          {{ sponsor.white_logo_img|camoify|safe }}
+        {% else %}
+          {{ sponsor.color_logo_img|camoify|safe }}
+        {% endif %}
+        <span class="sponsors__name">{{ sponsor.name }}</span>
+        <span class="sponsors__service">{{ sponsor.service }}</span>
+      </a>
```
miketheman: This change is not reflected in the "why" this is happening - and is more extensive than allowing the sponsor's white logo if supplied.

Why is this change desired?

More importantly, this effectively duplicates the entire logic block, iterating on `request.sponsors`. A better refactor would be to calculate all of the ordering in Python, and pass an already-ordered collection of `sponsors` to be rendered with a single logic block, instead.

miketheman (changes requested):
See inline comments, but also address failing test coverage once done with any changes.

warehouse/templates/includes/sponsors-footer.html
JacobCoffee: moved things to python, and i added reasoning in the OP

JacobCoffee:
https://github.com/pypi/warehouse/actions/runs/24110246717/job/70343004070?pr=19803
am confused why we dont fix line length issues on `make reformat` :\

warehouse/sponsors/__init__.py
```diff
@@ -8,12 +8,32 @@
...
+        (s for s in all_sponsors if s.infra_sponsor),
+        key=lambda s: s.name,
+    )
+    return psf + infra
 
 
 def includeme(config):
     # Add a request method which will allow to list sponsors
     config.add_request_method(_sponsors, name="sponsors", reify=True)
+    config.add_request_method(_footer_sponsors, name="footer_sponsors", reify=True)
```
miketheman, line 36: What if instead of a new request method, `request.sponsors` performs the query, and returns a well-structured dictionary that is sorted, ordered, etc and can be used in a template?

Maybe a shape like:

```python
sponsors = {
    "infrastructure": [...],
    "all": [...],
}
```

Then we remain with a single request method request.sponsors that returns the presorted, filtered objects for the templates to iterate?

warehouse/sponsors/tasks.py
```diff
@@ -50,8 +50,10 @@ def update_pypi_sponsors(request):
         sponsor.service = sponsor_info["description"]
         sponsor.link_url = sponsor_info["sponsor_url"]
         sponsor.color_logo_url = sponsor_info["logo"]
+        sponsor.white_logo_url = sponsor_info.get("white_logo")
         sponsor.level_name = sponsor_info["level_name"]
         sponsor.level_order = sponsor_info["level_order"]
         sponsor.is_active = True
         sponsor.psf_sponsor = True
+        sponsor.footer = sponsor_info["level_name"] in {"Visionary", "Sustainability"}
```
miketheman, line 58: This condition is no longer used in the template

tests/unit/sponsors/test_init.py
```diff
@@ -42,16 +41,40 @@ def test_do_not_schedule_sponsor_api_integration_if_no_token():
...
-    SponsorFactory.create_batch(5)
+    expected = SponsorFactory.create_batch(5)
     SponsorFactory.create_batch(3, is_active=False)
 
     result = sponsors._sponsors(db_request)
-    expected = db_request.db.query(Sponsor).filter(Sponsor.is_active == true()).all()
 
-    assert result == expected
     assert len(result) == 5
+    assert set(result) == set(expected)
```
miketheman, line 56: Nice refactor

miketheman (dismissed):
Can you also share a screenshot what looks like with some richer data?

JacobCoffee:
image

JacobCoffee:
i think all review comments have been addressed now

warehouse/sponsors/tasks.py
JacobCoffee: now used [4495b3e](https://github.com/pypi/warehouse/pull/19803/commits/4495b3e7f6840276fdc5dbe0692b61489695777e)

warehouse/sponsors/__init__.py
JacobCoffee: resolved in [9bdcf97](https://github.com/pypi/warehouse/pull/19803/commits/9bdcf976078f0cb05235a1d0ed03b395cc1982b8)

ewdurbin:
Is the tiering required? Can the additional sponsor logos not just be displayed first left-to-right with "Visionary Sponsor of the PSF" or "Sustainability Sponsor of the PSF" under their logo/name?

This would create significantly less empty space and keep from driving remaining sponsors further down the page.

JacobCoffee:
@ewdurbin it isnt required. i would also prefer the footer not grow but shrink, something like this
 
image

ewdurbin:
> @ewdurbin it isnt required. i would also prefer the footer not grow but shrink, something like this
> 
> image

i like this a lot, but for a larger redesign, i think @nlhkabu's input would be really valuable.

JacobCoffee:
> i think @nlhkabu's input would be really valuable.
agree! asked nicole via slack yesterday

some good questions out of it:

> bigger questions are:
> 
> what links should actually be in the footer?
> how should they be grouped?

but i think these can be done in follwup PRs once her work merging help/docs goes through we can prune and reorganize foote rlinks

JacobCoffee:
Latest commit is
image
which is a little broadening of scope but i think its pushing us in a good direciton

JacobCoffee:
image

 Latest after pulling out the language switcher sstuff

nlhkabu:
@JacobCoffee how can I preview the actual sponsor logos locally?

nlhkabu:
Initial thoughts:

1. I like the rework, overall :)
1. I'm not a fan of the way that some of the links wrap - I think we could make the columns a teeny bit wider to accommodate the longer links (in English)
2. I would suggest extending `footer__brand` to 100% wide as soon as the links wrap below it (screenshot below)
3. `footer_menus` should stack to a single column on mobile
4. We should use standard button size for the donate button, instead of writing custom CSS here. At the same time the "switch to desktop" button could be resized and moved to the left, to align with the rest of the content

image

JacobCoffee:
@nlhkabu re: preview local, i just load up placeholders. images are pulled with an api key from python.org api. 
i guess it would technically be possible to go the sponsor page and view page source and grab the URLs.  

alternatively and probably easier you can add manual logos via the pypi admin panel

JacobCoffee:
image

image

noting that language bar will go away in #20400

nlhkabu:
Hi @JacobCoffee 

I've now tested this with real logos and made some adjustments.  Here is the zip of reformatted (cropped) colour pngs:

[sponsor-logos.zip](https://github.com/user-attachments/files/31292613/sponsor-logos.zip)

------------------

## Desktop
image

## Large tablet
image

## Small tablet
image

## Mobile
image

nlhkabu (approved):
HTML and CSS all ok :)

nlhkabu:
@JacobCoffee I've pushed changes, please review.

JacobCoffee:
@nlhkabu looks great to me!
image
ill merge on your final approval

nlhkabu (approved):

nlhkabu:
Looks good to me!

ewdurbin (approved):
if @nlhkabu is happy, I'm happy.

PR #19804, opened 2026-04-06
Add PyEmscripten support per PEP 783

hoodmane:
[PEP 783](https://peps.python.org/pep-0783/#package-indexes) was accepted, so this adds PyEmscripten support.


warehouse/utils/wheel.py
```diff
@@ -54,6 +54,10 @@ class InvalidWheelRecordError(Exception):
         re.compile(r"^ios_(\d+)_(\d+)_(.*?)_iphonesimulator$"),
         lambda m: f"iOS {m.group(1)}.{m.group(2)}+ {_normalize_arch(m.group(3))} Simulator",  # noqa: E501
     ),
+    (
+        re.compile(r"^pyemscripten_(\d+)_(\d+)_wasm32$"),
+        lambda m: f"PyEmscripten {m.group(1)}.{m.group(2)} wasm32",  # noqa: E501
```
JacobCoffee, line 59: just a nit, but all over have an indicator of forwards compat like 
`ioS 26.0+` or `Android API level 21+`
im not sure on pyemcripten versioning but it has no `+`

JacobCoffee (approved):

warehouse/utils/wheel.py
hoodmane: No + yes it's not forward compatible.

hoodmane: See discussion in the pep around here: 
https://github.com/python/peps/blob/32d17e43552a29d4c95d4311f017c2914cdabc7d/peps/pep-0783.rst?plain=1#L61

ryanking13:
@JacobCoffee Can we merge this? We would like to announce the users that PEP 783 is accepted and people can now upload wheels to PyPI. Thanks!

henryiii:
Packaging 26.1 just shipped about 10 minutes ago with pyemscripten wheel tag support!

https://github.com/pypa/packaging/releases/tag/26.1

warehouse/forklift/legacy.py
```diff
@@ -177,6 +177,10 @@
 _manylinux_arches = _jointlinux_arches | {"ppc64"}
 _musllinux_arches = _jointlinux_arches
 
+_pyemscripten_platform_re = re.compile(
+    r"pyemscripten_(?P<major>\d+)_(?P<minor>\d+)_wasm32"
```
miketheman, line 181: observation: looks like their's potential for upstream emscripten memory64 support, so at the point that PyEmscripten supports wasm64, this might need to be extracted to an `arches` style set, and the `if m:` clause below will need to adapt.

miketheman (approved):
Thanks! I read through and found that we don't directly rely on `packaging.tags` for much more than type hints, so we don't need to wait on upgrading the `packaging` library.
The tests for `_valid_platform_tag` are sufficient to prove the string will be allowed through validation, and we make our own pretty tags (maybe that's something to also move upstream? 🤔)

An inline note for our future selves, but otherwise, good to go.

warehouse/forklift/legacy.py
agriyakhetarpal: I believe Wasm64 currently suffers severe performance penalties, which will be exacerbated by the CPython VM already running in the WASM VM. It would indeed require these changes, but I don't really see this happening anytime soon because of the fact that it's probably overkill for browsers, as their per-tab limits are way smaller than 4 GiB. Out-of-browser use cases, maybe.

PR #19817, opened 2026-04-07
feat: add workflows to deploy draft PRs to TestPyPI

miketheman:
Add two GitHub Actions workflows for managing a long-lived `testpypi` branch that cabotage can deploy to test.pypi.org:

- `testpypi-deploy.yml`: when the `deploy-to-testpypi` label is added to a draft PR, merges the PR's changes into the `testpypi` branch and comments on the PR with the result.
- `testpypi-reset.yml`: on every push to `main`, automatically syncs the `testpypi` branch to match `main` - unless a draft PR with the label is active. Also supports manual dispatch.


miketheman:
@dstufft asked: How can we prevent longer drifts from `main`?
@JacobCoffee asked: what about database drift if they get reverted? Do we need to exclude those, or downgrade?

PR #19846, opened 2026-04-12
feat: enable CORS headers on RSS feeds

miketheman:
Browser-based feed utilities balk at accessing this already-public, well-cached data via JavaScript.

Relax the browser `same-origin` policy - no affect to server-side or non-browser access.
As these are read-only GET endpoints with no authentication, `CORS *` doesn't enable any new attack surface.
The added `X-PyPI-Last-Serial` header is useless to readers, but keeps the settings consistent across all usages.

### DIFF
--- a/tests/unit/rss/test_views.py
+++ b/tests/unit/rss/test_views.py
@@ -10,6 +10,16 @@ from warehouse.rss import views as rss
 from ...common.db.packaging import ProjectFactory, ReleaseFactory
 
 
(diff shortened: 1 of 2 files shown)


miketheman:
Easy enough to test in a browser's Console:

```javascript
fetch("https://pypi.org/rss/updates.xml").then(r => r.text()).then(console.log, console.error)
```

ewdurbin (approved):

PR #19855, opened 2026-04-13
feat: add callout to discourage security research

miketheman:

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -2128,6 +2128,54 @@ msgid ""
 msgstr ""
 
 #: warehouse/templates/accounts/register.html:214
+msgid "Important: Prohibited actions put PyPI at risk"
+msgstr ""
+
+#: warehouse/templates/accounts/register.html:216
+#, python-format
+msgid ""
+"%(title)s is critical infrastructure for the Python community, with high "
+"volunteer engagement in keeping it secure. Unnecessary violations of our "
+"terms waste limited time we could dedicate to securing the ecosystem."
+msgstr ""
+
+#: warehouse/templates/accounts/register.html:220
+msgid ""
+"PyPI <strong>does not authorize</strong> using our infrastructure for "
+"<strong>any kind of security research</strong>, including probing "
+"vulnerabilities through dependency confusion or \"beacons\"."
+msgstr ""
+
+#: warehouse/templates/accounts/register.html:223
+msgid ""
+"Prohibited security research includes <strong>bug bounty "
+"programs</strong>, <strong>pentesting</strong>, and <strong>red team "
+"engagements</strong>, unless explicitly authorized by PyPI "
+"administration."
+msgstr ""
+
+#: warehouse/templates/accounts/register.html:226
+msgid ""
+"Violations will result in <strong>immediate account suspension</strong> "
+"and may include <strong>removal of all associated projects and "
+"files</strong>."
+msgstr ""
+
+#: warehouse/templates/accounts/register.html:230
+#, python-format
+msgid ""
+"If you are a security researcher, please refer to our <a "
+"href=\"%(security_page)s\">published security policy</a> and coordinate "
+"through proper disclosure channels."
+msgstr ""
+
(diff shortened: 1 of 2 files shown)


miketheman:
superseded

<details>

What it looks like:

image

</details>

nlhkabu:
@miketheman - wondering what this looks like as part of the overall registration flow?
My concern is for the 99.9% of users who are not abusing PyPI being hit in the face with this :)

Could we instead introduce a registration field that asks the user why they are setting up their account?

Something like: 
Label: "Account purpose"
Checkboxes: "Upload packages for distribution to others", "Security research", "Other", etc. (we would need to come up with a good list here)

Then if the user selects "Security research" _then_ we hit them with this?

kam193:
This is a good point. However, I was thinking that we have forgotten about two popular abuses that are not limited to the security research:

* obfuscation
* using PyPI as large file storage (yeah... Uploading 100s of ebooks or splitting and uploading 10s of files to host a LLM model is a thing) 

I think that an alternative to selective displaying the message could be adding a point about them, and maybe going from red to yellow frame wit black text, to indicate it's a preventive warning instead of hit in a face? 😅

miketheman:
> @miketheman - wondering what this looks like as part of the overall registration flow? My concern is for the 99.9% of users who are not abusing PyPI being hit in the face with this :)
> 
> Could we instead introduce a registration field that asks the user why they are setting up their account?
> 
> Something like: Label: "Account purpose" Checkboxes: "Upload packages for distribution to others", "Security research", "Other", etc. (we would need to come up with a good list here)
> 
> Then if the user selects "Security research" _then_ we hit them with this?

I intended to place this right above the "submit" button at the bottom of the registration page to make it very apparent.

On the adding a selection box, maybe? In my emails with some folks, they don't consider what they are doing to be "Security research" - so maybe that's part of it. I often get a "but it's authorized by Company A!" - which doesn't matter here. They may also state "it's not research, it's actually a supply chain attack simulation" 😮‍💨 So I'm dubious that adding a selector would move the needle, but am happy to entertain it. A challenge with that approach is that then we should probably store that selected data somewhere, which makes this implementation harder than a big callout block.


> This is a good point. However, I was thinking that we have forgotten about two popular abuses that are not limited to the security research:
> 
> * obfuscation
> * using PyPI as large file storage (yeah... Uploading 100s of ebooks or splitting and uploading 10s of files to host a LLM model is a thing)
> 
> I think that an alternative to selective displaying the message could be adding a point about them, and maybe going from red to yellow frame wit black text, to indicate it's a preventive warning instead of hit in a face? 😅

Thanks for noting those cases - absolutely should be added. I used the `callout-block__danger` class for the redness - can totally change to [`callout-block__warning`](https://github.com/pypi/warehouse/blob/ff29f6f700e03fa4582685543764678e26b9cbc7/warehouse/static/sass/blocks/_callout-block.scss#L33-L34) (similar to what quarantined projects display today).

kam193:
I was thinking about the flow and other use cases. I believe the purpose selection won't work (for reasons you mentioned), and I'm more and more in the direction of "making it applicable for everyone". And as part of that, I'd suggest one more change: instead of _I assert that I will not use this account for security research or testing._, maybe  go in the direction of just _I agree to respect PyPI's Terms of Service and Acceptable Use Policy_. 

It's not an uncommon pattern to have such agreement explicitly, it still draws attention to additional explanations of the AUP, and should make it less strange for people who use PyPI for genuine use cases.

miketheman:
> And as part of that, I'd suggest one more change: instead of _I assert that I will not use this account for security research or testing._, maybe go in the direction of just _I agree to respect PyPI's Terms of Service and Acceptable Use Policy_.

We already have a line on the registration page:

> By registering, you agree to the PyPI Terms of Service.

I like the idea of echoing that as a checkbox with links, and would expand to:

> By registering, you agree to the PyPI Terms of Service and Acceptable Use Policy.

miketheman:
Updated, now looks like:

image

... and:

Screenshot 2026-09-17 at 11 12 44

JacobCoffee (approved):

nlhkabu (approved):

PR #19860, opened 2026-04-14
fix: disallow TP uploads to quarantined projects

miketheman:


tests/unit/packaging/test_models.py
```diff
@@ -268,7 +268,9 @@ def test_acl_for_quarantined_project(self, db_session):
             team=team, project=project, role_name=TeamProjectRoleType.Owner
         )
 
-        publisher = GitHubPublisherFactory.create(projects=[project])
+        # Publishers should not appear in the ACLs, since quarantined
+        # projects should not allow uploads from trusted publishers
+        GitHubPublisherFactory.create(projects=[project])
```
dstufft, line 273: small nit: feel free to ignore. I was confused by this change until it clicked that we needed a publisher to exist to make sure it didn't appear in the ACLs, might be nice to expand on that a tad bit more in the comment.

dstufft (approved):

PR #19909, opened 2026-04-20
REST API for managing trusted publishers

### LINKED ISSUE #14456: Allow configuring of trusted publisher through an API (or other bulk mechanism)
From https://github.com/pypi/pypi-oidc-private-beta-community/issues/34:

> I feel like the long-term API tokens should allow setting up Trusted Publishing via API. Would it be possible to expose a corresponding endpoint? This can be useful for mass-maintained sets of projects that follow the same structure, like the ones relying on https://github.com/jaraco/skeleton or similar. Such an API could enable setting up new projects from templates and complement things like GitHub repository templates or cookiecutter templates. For projects migrating from the project-scoped API tokens, it might be useful to have a self-destruct API endpoint, that would expire the current token which could be used for such migration automation. Is it hard to implement such an interface?

Allows maintainers with dozens or hundreds of projects to readily enroll with trusted publishing (and apply the requisite change upstream) without the toil of clicking through pages of web pages.

An addition to the REST API that lets users configure trusted publishing would readily allow automation of this process.

Alternatives for consideration:
- have trusted publishing enabled by default (maybe based on package metadata)
- allow user to bulk-configure projects, by supplying a list of projects or selecting checkboxes of all projects

thatch:
Add POST/GET/DELETE endpoints under /danger-api/projects/{name}/trusted-publishers
authenticated by a project-owner API token (macaroon), supporting all four
providers: GitHub, GitLab, Google, and ActiveState.

Also adds bin/configure-trusted-publisher, a standalone CLI that reads credentials
from ~/.pypirc like twine, auto-detects the provider and repo from the current git
checkout, and lets you select a workflow file interactively.

Closes https://github.com/pypi/warehouse/issues/14456

The configure-trusted-publisher script should go into its own project (but that should be owned by pypa, not me).  Comments welcome on whether this is "danger-api"-worthy.  I'll test manually more once the basic idea is validated.

### DIFF
--- a/bin/configure-trusted-publisher
+++ b/bin/configure-trusted-publisher
@@ -1,0 +1,345 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""
+Configure a PyPI trusted publisher via the API.
+
+Reads credentials from ~/.pypirc (like twine), auto-detects the provider and
+repo from the current git checkout, and lets you pick a workflow file.
+
+Usage:
+    configure-trusted-publisher <project> [options]
+
+Examples:
+    configure-trusted-publisher mypackage
+    configure-trusted-publisher mypackage --dry-run
+    configure-trusted-publisher mypackage --environment release
+    configure-trusted-publisher mypackage --api-url https://test.pypi.org
+"""
+
+from __future__ import annotations
+
+import argparse
+import configparser
+import glob
+import json
+import os
+import re
+import subprocess
+import sys
+import urllib.error
+import urllib.request
+from pathlib import Path
+
+
+CONTENT_TYPE = "application/vnd.pypi.api-v0-danger+json"
+DEFAULT_PYPI_URL = "https://pypi.org"
+DEFAULT_REPOSITORY = "pypi"
+
+
+def _read_pypirc(config_file: str | None, repository: str) -> dict:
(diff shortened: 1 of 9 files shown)


bin/configure-trusted-publisher
```diff
@@ -0,0 +1,353 @@
...
+    token = config.get("password")
+    if token:
+        return token
+
+    print(
+        "Error: No API token found.\n"
+        "Provide via --token, PYPI_TOKEN env var, or ~/.pypirc [pypi] password field.",
+        file=sys.stderr,
+    )
+    sys.exit(1)
```
jaraco, line 88: I keep my token safely stored encrypted using `keyring` ([here's how twine does it](https://github.com/pypa/twine/blob/4f20c0dc374bbd54fe7cddc341e3c5ba7479917b/twine/auth.py#L259)). Ideally, this script would provide that as an option (or maybe I just pass in `--token $(keyring get ...)`).

jaraco (approved):
Looks good to me. I'm fairly confident this will have the intended effect of making it possible to bulk enable trusted publishing across any number of repos.

bin/configure-trusted-publisher
thatch: My hope is that this script either becomes part of twine, or its own project, and we can sort that out there.  It's a demo here because without an api client it's kind of hard to show that it works :)

PR #19914, opened 2026-04-21
feat(admin): implement datatables server-side pagination

miketheman:
As more records are added, the page takes longer to load, as it's currently loading *everything* before passing it to a renderer, and then dumping everything for client-side browsing.

Introduces a server-side parser for DataTables so that we can fetch a subset and display, as well as search for items beyond what has already been fetched.

If this works well enough, I may extract and leverage for other large/slow admin pages for better load times and searchability.

Refs: https://datatables.net/manual/server-side


ewdurbin (approved):

JacobCoffee (approved):
datatables are cool

PR #19915, opened 2026-04-21
remove alternate-locations functionality

ewdurbin:
PEP 708 has been rejected, ref: https://discuss.python.org/t/proposal-withdraw-reject-pep-708-extending-the-repository-api-to-mitigate-dependency-confusion-attacks-due-to-lack-of-adoption/106800/13

this just removes the code, a second PR will followup to remove the database tables.

NOTE: I decided to leave associated [events](https://github.com/pypi/warehouse/blob/39f86591a3bce47add4a5fea05dd7f597563b195/warehouse/events/tags.py#L98-L99) in order to not break the history of modifications made

NOTE: Database Consistency Check will be an expected failure in this part. Follow up PR will be a migration which removes the actual tables.


woodruffw (approved):
Nice 🙂

PR #19917, opened 2026-04-21
Correctly purge the organization profile when updating an org's projects

### LINKED ISSUE #19911: Stale organization project lists
**Describe the bug**

The [Astral org](https://pypi.org/org/astral/) has at least 4 projects enrolled against it, but only three are currently rendered in the `/orgs/{orgname}` view:

Image

(Specifically `ty` is missing, even though it was added 10+ minutes before the screenshot above).

I *suspect* this is a stale cache entry or similar for the organization project list view.

**Expected behavior**

The organization view that lists projects should show a timely listing of the organization's current projects, i.e. should be invalidated whenever the organization's project list changes.

**To Reproduce**

The easiest reproducer for this is probably to add a project to an org and observe staleness in this view.

**My Platform**

* macOS 26.2, Safari Version 26.2 (21623.1.14.11.9), but I believe this is purely a server-side bug

**Additional context**

N/A

woodruffw:
~~The bug here was actually kind of subtle: we use `key_factory` to build the purge keys, which can silently fail if the underlying attribute access fails (e.g. `Project.organization` returns `None` or raises `AttributeError`).~~

~~This in turn wouldn't happen in most normal cases, but _can_ happen if we build the purge keys during `after_flush`, since at that point relationships like `Project.organization` haven't been materialized yet (if the `Project` was initialized with `organization_id` instead).~~

My original theory was wrong here.

The bug here stems from an interaction between `flag_dirty` and #19898. The basic problem is that `add_organization_project` calls `organization.record_event`, which adds `events` to the organization's `committed_state`. This in turn trips the optimization in #19898, since `committed_state == {'events'}`, causing us to skip the purges.

(`committed_state` is otherwise empty, since `flag_dirty` itself doesn't add anything else.)

Fixes #19911.

### DIFF
--- a/tests/unit/cache/origin/test_init.py
+++ b/tests/unit/cache/origin/test_init.py
@@ -175,9 +175,7 @@ def test_store_purge_keys_organization_project_add_survives_events_committed_sta
     # Construct an OrganizationEvent with `source=organization` to force
     # `events` into Organization's committed_state before the next flush.
     # This is the exact state the manage view produces via `record_event`.
-    db_request.db.add(
-        organization.Event(source=organization, tag="test:regression")
-    )
+    db_request.db.add(organization.Event(source=organization, tag="test:regression"))
 
     db_request.db.flush()
 
--- a/tests/unit/cache/origin/test_init.py
+++ b/tests/unit/cache/origin/test_init.py
@@ -149,21 +149,8 @@ def test_store_purge_keys_organization_project_add_survives_events_committed_sta
 ):
     # Regression guard for issue #19911.
     #
-    # `store_purge_keys` skips emitting a purge for a dirty object whose
-    # `committed_state` is a subset of `_NON_CACHE_RELEVANT_ATTRS` (the
-    # perf optimizations in PRs #19898 and #19910). Before this fix,
-    # `add_organization_project` relied on `flag_dirty(org)` to trigger an
-    # Organization-dirty purge — but the manage view then recorded an event
-    # on the organization, setting `event.source = org` and back-populating
-    # the `events` dynamic collection, which put `'events'` in the
-    # Organization's committed_state. The skip then masked the only purge
-    # for `org/{orgname}`, leaving the CDN entry stale.
-    #
-    # The fix makes the purge come from the new OrganizationProject row
-    # (session.new, not session.dirty) — the skip applies only to dirty
-    # objects, so a `session.new` row is unaffected. This test pins down
-    # that behavior: the `org/{name}` purge must still fire even when the
-    # Organization has `events` in its committed_state at flush time.
+    # We should purge `org/{name}` even when the Organization's
+    # `committed_state` includes `events` (or any other non-cache-relevant attr).
     organization = OrganizationFactory.create()
     project = ProjectFactory.create()
     db_request.db.flush()
--- a/tests/unit/cache/origin/test_init.py
+++ b/tests/unit/cache/origin/test_init.py
@@ -4,6 +4,10 @@ import pretend
 import pytest
 
 from tests.common.db.accounts import UserFactory
+from tests.common.db.organizations import (
+    OrganizationFactory,
+    OrganizationProjectFactory,
+)
 from tests.common.db.packaging import (
     FileFactory,
     ProjectFactory,
@@ -16,6 +20,7 @@ from warehouse.cache import origin
 from warehouse.cache.origin.derivers import html_cache_deriver
 from warehouse.cache.origin.interfaces import IOriginCache
 from warehouse.observations.models import ObservationKind
+from warehouse.organizations.services import database_organization_factory
 
 
 def test_store_purge_keys():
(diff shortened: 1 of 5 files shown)


warehouse/organizations/services.py
```diff
@@ -556,16 +556,16 @@ def add_organization_project(self, organization_id, project_id):
...
         """
+        from warehouse.packaging.models import Project
+
+        # We need to pass the full objects instead of just their IDs
+        # to avoid silently breaking our purges that rely on `if_attr_exists`.
         organization_project = OrganizationProject(
-            organization_id=organization_id,
-            project_id=project_id,
+            organization=self.db.get(Organization, organization_id),
+            project=self.db.get(Project, project_id),
```
woodruffw, line 565: Flagging: this is the core of the fix, but I'm also kind of unhappy with it (since now we're fetching an entire row from the DB). Maybe not a huge deal in context though, given that adding a project to an org isn't exactly a hot path operation.

woodruffw:
NB: This effectively "fixes" the purge behavior by bypassing the optimization in #19898 entirely. I'm not sure if that's the best approach 😅

tests/unit/cache/origin/test_init.py
```diff
@@ -113,6 +118,80 @@ def test_store_purge_keys_skips_audit_only_collection_changes(
...
+):
+    # Adding a project to an organization via `add_organization_project` must
+    # purge both the project profile (`project/{name}`) and the organization
+    # profile (`org/{name}`), otherwise the cached `/org/{orgname}` project
+    # list goes stale.
+    #
+    # Exercising `add_organization_project is load-bearing: the
+    # purge-key factory uses `if_attr_exists`, which resolves to `None` during
+    # `after_flush` if the row was built with only FK ids instead of
+    # objects, causing us to silently drop the purges.
```
miketheman, line 132: Claude insists on using comment blocks over docstrings, I don't know why.

warehouse/packaging/services.py
```diff
@@ -667,10 +667,17 @@ def create_project(
...
+            # so the `OrganizationProject` purge-key factory can resolve the
+            # relationship during `after_flush` and actually emit the
+            # `org/{name}` and `project/{name}` purge keys.
+            from warehouse.organizations.models import Organization
+
             self.db.add(
                 OrganizationProject(
-                    organization_id=organization_id, project_id=project.id
+                    organization=self.db.get(Organization, organization_id),
+                    project=project,
```
miketheman, line 680: I reverted this change locally, and tests pass. So either we don't need this change, or the tests aren't expressing the condition well enough.

warehouse/organizations/services.py
```diff
@@ -556,16 +556,16 @@ def add_organization_project(self, organization_id, project_id):
         """
         Adds an association between the specified organization and project
         """
+        from warehouse.packaging.models import Project
```
miketheman, line 559: I'm still not sure why our isort config doesn't flag inline imports without a disable comment - but this should be placed at top-of module. (Hey Claude, remember that preference).

warehouse/organizations/services.py
miketheman: If we don't watch to fetch the record, we can add a `flag_dirty` for the `organization_project.project` instead, but that also feels clunky.

Considering that the upstream caller already has both the Organization and Project objects, I think it's a better choice to pass the objects that we can then mutate, instead of working with IDs, and then we avoid some of the manual db.add,flush,flag_dirty, and let the ORM do its thing.

warehouse/packaging/__init__.py
```diff
@@ -173,6 +173,7 @@ def includeme(config):
         OrganizationProject,
         purge_keys=[
             key_factory("project/{attr.normalized_name}", if_attr_exists="project"),
+            key_factory("org/{attr.normalized_name}", if_attr_exists="organization"),
```
miketheman, line 176: I _think_ this is the main fix - since it was previously unregistered.

tests/unit/cache/origin/test_init.py
```diff
@@ -113,6 +118,80 @@ def test_store_purge_keys_skips_audit_only_collection_changes(
...
+    # Construct an OrganizationEvent with `source=organization` to force
+    # `events` into Organization's committed_state before the next flush.
+    # This is the exact state the manage view produces via `record_event`.
+    db_request.db.add(organization.Event(source=organization, tag="test:regression"))
+
+    db_request.db.flush()
+
+    purges = db_request.db.info.get("warehouse.cache.origin.purges", set())
+    assert f"org/{organization.normalized_name}" in purges
+    assert f"project/{project.normalized_name}" in purges
```
miketheman, line 171: This test passes for the wrong reasons, and also believe it shouldn't work in its current behavior.

Adding an Event shouldn't dirty an object for cache purge, which is why it's excluded. Dirties that include `events` and other attributes continue to be purged correctly. One logged example from a few minutes ago:

```json
{
  "obj_class": "File",
  "state": "dirty",
  "keys": [
    "project/ctboost"
  ],
  "changed_attrs": [
    "events",
    "provenance"
  ],
  "event": "cache_purge_keys_generated",
  ...
}
```
The session purge shows that the object had `events` + <anything else>, and thus got a cache purge. see https://github.com/pypi/warehouse/blob/e2dd240f3924fc89572a1ddba4ba72ff1ab78530/warehouse/cache/origin/__init__.py#L69-L70
maybe the word "only" needs to be bolded :wink:

The reason this test is currently passing is that there's no "session purge pop" after `organization_service.add_organization_project(...)` call - so those changes will trigger the purge.

Moving `db_request.db.flush() / db_request.db.info.pop("warehouse.cache.origin.purges", None)` to be after the service call ensures the session state is "clean" before attempting the Event addition, and shows the failure.

Generally I think removing this test case is best, as the general idea is covered by this test: https://github.com/pypi/warehouse/blob/25bc4041168dc0adf8bae53ee86e805f72aae191/tests/unit/cache/origin/test_init.py#L100-L118
Maybe adding some more parameterizations there could be helpful, but might be more complexity than worth it? Your choice.

miketheman (changes requested):
Thanks for digging in, and there's a few things inline - let me know if anything doesn't make sense.

PR #19934, opened 2026-04-24
perf: remove an n+1 pattern from account publishing page

miketheman:
When loading the /manage/account/publishing/ page, the template loader
pattern iterated on every project for the user, lazy-loading `oidc_publishers`.
The more projects with OIDC enabled, the slower the page.

Instead, load all of the Project's `name` column only, since that's
all what's needed for the template.

----

perf: wrap default_response with a cached_property

When called multiple times in a request, no need to reload the same data
each time.

### DIFF
--- a/warehouse/accounts/views.py
+++ b/warehouse/accounts/views.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import datetime
+import functools
 import hashlib
 import json
 import uuid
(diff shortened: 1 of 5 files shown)


ewdurbin (approved):

warehouse/accounts/views.py
```diff
@@ -1767,9 +1768,20 @@ def _check_ratelimits(self):
                 )
             )
 
-    @property
+    @functools.cached_property
```
dstufft, line 1771: Is this actually doing something? The class only lives for a single request and I'd only expect to call this property once per request?

miketheman: See the removals in the test case in fcc20540ad6cb78f2775f3c70634c71fd76a2ca7 - the property can be called multiple times in a single request

PR #19938, opened 2026-04-24
add a frontend team that is scoped to working on HTML/CSS/etc

dstufft:
I've added a @pypi/warehouse-frontend team which has the same level of permissions as @pypi/warehouse-committers, so they're allowed to merge PRs to `main`.

However, @pypi/warehouse-frontend is *not* a sub-team of @pypi/warehouse-reviewers, so because we require approval from code owners to merge a PR, that means we can more finely scope the effective permissions that @pypi/warehouse-frontend has, by only giving that team codeowner over a subset of the files.

This PR updates our `CODEOWNERS` file so that files that are relevant to working on the front end has both @pypi/warehouse-frontend and @pypi/warehouse-reviewers as owners, which GitHub will implement as requesting reviews from *both* teams, but will only require a review from *ONE* of the codeowners.

That "OR" handling is implemented per file, so if a PR touches both files that touch the frontend *and* files that are not part of the front end, then the @pypi/warehouse-frontend team will be able to approve for the frontend files, but not for the additional files. However @pypi/warehouse-reviewers would be able to approve for  both (since they are listed as owners for both-- if we ever added a file that @pypi/warehouse-reviewers  isn't an owner on, then they wouldn't be to approve for that).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,27 @@
+# By default, everything is owned by our reviewers team.
 * @pypi/warehouse-reviewers
+
+# Frontend files are owned by both our reviewers team AND the frontend team,
+# which allows either team to approve a Pull Request, but it does not require
+# both to approve it.
+#
+# We have to keep @pypi/warehouse-reviewers here, because GitHub will only match
+# a path to a single line in the CODEOWNERS file, whichever line it matches last.
+.github/ISSUE_TEMPLATE/~visual-design.md @pypi/warehouse-reviewers @pypi/warehouse-frontend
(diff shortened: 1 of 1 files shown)


JacobCoffee (approved):

miketheman (approved):
Thanks for picking this up.
I validated the structure, and had to refresh my brain that last-match wins.

nlhkabu:
Thanks for this @dstufft 

Could we add `/docs/user`  and `/docs/dev` to the scope here?
It would make sense to be able to:

1. Update user facing documentation
2. Update developer documentation specific to the FE

Otherwise, this looks good to me!

dstufft:
That seems reasonable to me! I've gone ahead and done that. I'll merge this once tests pass and flip you to the new team.

PR #19940, opened 2026-04-25
ignore a .venv directory if one exists

dstufft:
This just allows someone to use tools that want to add a `.venv` directory without it causing issues. This should be a no-op otherwise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .state
 .idea
 .envrc
+.venv
 
 docs/dev/_build/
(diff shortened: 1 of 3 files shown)


JacobCoffee (approved):
big +1, i really dislike developing in docker. its painful, and use uv for deving on warehouse when i can. also im too dumb to know how to make my IDE see my docker env, but with .venv it auto-finds it and does all the linking and import magic instead of red/yellow squigglies everywhere

PR #19941, opened 2026-04-25
ignore a few missing imports in pyright

dstufft:
We ignore these in the mypy configuration, unfortunately pyright doesn't check that configuration and so it warns on these which has been keeping my editor yelling at me for awhile now.

Adding a handful of "noise" comments seems pretty low cost to me and makes working on warehouse less annoying for me personally 😄, but I'm happy to just keep trying to remember to exclude this "floating patch" that I've had applied to my local environment for awhile now if people feel strongly about not adding them.

### DIFF
--- a/tests/unit/test_db.py
+++ b/tests/unit/test_db.py
@@ -8,7 +8,7 @@ import psycopg
 import pytest
 import sqlalchemy
 import venusian
-import zope.sqlalchemy  # type: ignore
+import zope.sqlalchemy  # pyright: ignore
 
 from sqlalchemy import event
 from sqlalchemy.exc import DBAPIError, OperationalError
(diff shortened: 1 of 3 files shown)


JacobCoffee (approved):
since we arent a library whats the point here of running both?

dstufft:
In abstract there's little point to running both, but the project runs mypy while my editor happens to run pyright so unless either the project switches to pyright, or I reconfigure my editor to use mypy instead of pyright. Last time I tried to use mypy in my editor it was unbearably slow.

I know there's also `ty` now, which I think is supposed to be even faster? But I haven't really tried it yet.

dstufft:
IOW, 6 "noise" comments seemed like a reasonably low cost to support editors that are using pyright :) Which I think is common with VSCode?

JacobCoffee:
+1 to `ty` or basedpyright or anything not slow as a turtle

PR #19942, opened 2026-04-25
use fd instead of find

dstufft:
We have a few uses of the venerable `find` command to locate files to run quality checks against such as formatting and linting. We've always had this awkward situation with `find` that we want to run these things against all of *our* files, but not against other files that might inadvertently exist in the directory. To try to ensure this we've combined trying to scope our find calls to search from `.`, but exclude a number of patterns that are commonly not "our" files.

This kinda sucks? It makes things error prone; if an unexpected "kind" of file gets added, we have to go and update all of those find commands to add yet another ignore pattern. This is compounded by the fact that often times these files are not files that the project in general works with, but rather are files that the local developer's workflow might end up producing (like a `.venv` directory).

Rather than try and maintain these multiple lists of files to ignore, this PR switches us to use [fd](https://github.com/sharkdp/fd) which is a find replacement which will automatically ignore files that match our `.gitignore` pattern as well as files that are hidden using the `.` prefix.

This removes the need to maintain (some of) the duplicated set of "paths to ignore", reducing friction just a tiny bit.

It's likely pretty obvious, but just to call out the flags that are being used here:

- `--type` searches for paths with a given type, `f` means file like in `find`.
- `--extension` searches for files with a given extension, replacing `-name '*.ext'`.
- `--exclude` excludes paths matching a particular glob.
- `--exec-batch` works like `-exec` in find with the `+` flag that collects all of the entries and runs the executable once with all of the arguments (`-exec` exists as well but runs it once per).
- `-exec` and `-exec-batch` default to adding `{}` to the end of the command, so I've omitted it.


miketheman (commented):
I'm a little hesitant since `fd` is not included by default on macOS system today - is this a replacement for `find` on any other popular distributions? I noticed you included Debian's `fdfind` to rename to `fd` - maybe that's the signal that this isn't 100% baked into the wider ecosystems yet?

dstufft:
AFAIK it's not replaced `find` on any distribution because it's not a drop in replacement, it performs a similar task but with different CLI args and behavior (one of which is the reason this PR exists-- ignoring `.gitignore` directories by default).

Debian renamed it from `fd` to `fdfind` because there's already a [fd command on debian for some random tool](https://packages.debian.org/source/trixie/fdclone). It's not unusual for debian to end up having to give CLIs slightly different names when their names clash, typically whichever came second "loses" unless the maintainers for the two packages come to an agreement it's more important for the second one to have the name.

Since all of our scripts are intended to only run inside of the docker container, where this PR ensures `fd` is installed, does it not shipping by default in macOS matter much? If someone is running those scripts outside of the docker container, they're already doing something unsupported yea?

woodruffw (approved):
LGTM; I looked for a way to do this with `update-alternatives` but linking seems right in this context 🙂

PR #18175, opened 2025-05-22
chore(deps): replace curlylint with djlint

miketheman:
Suggestion: review this PR with "[ignore whitespace](https://github.com/pypi/warehouse/pull/18175/files?w=1)".

Some rules were disabled to make this achievable in a short term, and can be resolved after this is completed.

Locally, `djlint` is faster than `curlylint` - ~3 seconds vs ~14 seconds for the same amount of files.

PR #19943, opened 2026-04-27
Use ruff to format and lint

### LINKED ISSUE #17570: Investigate replacing `pyupgrade`, `black`, `isort` and `flake8` with `ruff`
Since I couldn't find any discussion about `ruff` in existing issues, I'm opening this one to see if there's interest and to document what I've found so far.

Currently, Python linting and formatting (via `make lint` and `make reformat`) is done via `black`, `isort`, `pyupgrade` and `flake8`.

`ruff` can be used as an (almost) drop-in replacement for most of these. The only thing not supported is custom `flake8` plugins, like the one defined [here](https://github.com/pypi/warehouse/blob/main/dev/flake8/checkers.py).

I have a branch where I tested the migration by:
- Replacing `pyupgrade` with `ruff`'s [pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
- Replacing `isort` with `ruff'`'s [isort rules](https://docs.astral.sh/ruff/rules/#isort-i)
- Replacing `black` with `ruff format`
- Replacing `flake8`'s default rules with `ruff check`
- Replacing `flake8-pytest-style` default rules with `ruff`'s [flake8-pytest-style rules](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt)
- Still running `flake8`, but only selecting the custom `WH` plugin mentioned above.

After migrating, running with a warm cache I get around a 10 second improvement in both `make` commands, and `make reformat` is now almost instant:
```
Before (black+pyupgrade+isort+flake8):

make reformat  0.08s user 0.04s system 1% cpu 10.757 total

make lint  0.15s user 0.07s system 0% cpu 33.592 total

---------------

After (ruff+flake8):

make reformat  0.07s user 0.04s system 14% cpu 0.754 total

make lint  0.14s user 0.07s system 0% cpu 23.736 total
```

cc @miketheman @woodruffw

dstufft:
This switches Warehouse from using black, isort, pyupgrade, flake8, and a collection of flake8 plugins to using (almost) only ruff (`flake8` is kept to *only* run our custom flake8 plugin until https://github.com/astral-sh/ruff/issues/283 is solved or another solution is found).

In #17570, @miketheman mentioned that he didn't believe it was worth changing tooling for the benefits at the time. I can understand that, but I think today the benefits far outweigh the costs.

When I run `make lint` on my MacBook, it takes around 40s to complete, about 25s (~65%) of that is taken up by pyupgrade, black, isort, and flake8, with black alone being about 10s of that time. This ends up cascading into my editor, where I have VSCode setup to run black when I save files in Warehouse, which can end up making an operation which typically is instant, take up to 10s on particularly large files like `tests/unit/forklift/test_legacy.py`. That kind of lag makes is super frustrating for me to work on the project 😦.

I tried switching my editor to ruff locally, which worked but added its own set of frustrations. While ruff is mostly a drop in replacement for the above tools (they claim 99% compatibility with black), it's not a 100% drop in replacement. This ends up having my editor and `make reformat` fighting over what the formatting should be, going back and forth depending on which one I ran last. Then I have to be careful to not commit these spurious formatting changes, otherwise my PR will fail in CI, which I'll then fix and repost, but it adds significant round trip latency to that process.

As far as I can tell, there's little reason *not* to use `ruff` here, the justification in #17570 was mostly that the motivation to migrate wasn't compelling but not that there was a reason that `ruff` wouldn't be just fine.

So this PR implements using ruff for linting and formatting, but keeps around `flake8` just to run our custom plugin. This ends up reducing that 25s of time down to basically 6s (of which, almost all of that is spent invoking flake8 for our custom plugin). It does not reduce the remaining 15s that comes from `sphinx-lint`, `djlint`, `mypy`, and `bin/flushes`.

Importantly, it means `ruff` can be used in an editor, giving near instant formatting on save without adding frustration around disagreeing formatters or saving a file randomly taking 10s to complete.

In #17570 @miketheman was worried about having to maintain the ruff cache to benefit from the speed ups. I'm happy to report that is basically not an issue. Each invocation of ruff takes about 110ms if I run them with `--no-cache` (to disable the cache completely), which drops down to about 35ms without `--no-cache` with a warm cache available.

So I've just let CI not worry about the cache, because while the cache is a win in relative terms, it'll probably take more time to save and restore the cache then we'd actually save.

For local development, I've followed the same pattern we use for mypy and pylint, I've configured ruff to put it's cache in `dev/` instead of the project root, so the cache will be stored in the host and shared between runs.

As I was setting this up and turning on rules, I realized there's a few non-performance benefits as well:

- The ruff editor integration understands not just formatting but linting as well, shifting linting issues further left so you'll get that feedback in real time as you edit, rather than later on when you run `make lint` (or even worse, when CI fails).
- It ignores things that are ignored by `.gitignore` by default, so no more maintaining one of the duplicated list of things to ignore, it just lints/checks the whole project by default.
- It supports the magic GitHub output that will annotate the PR with the lint failures, making it easier to see where the issues are in any given PR.
- It just seems to work better?
    - Some of the changes to imports are things that `isort` should have been catching, but just wasn't?
    - The `pyupgrade` functionality found changes to make (using `datetime.UTC`, f-strings, etc) that pyupgrade claims to support but just didn't make?
    - It appears flake8 doesn't notice naming issues in a `lambda`, so a number of them weren't caught.
    - ruff is sometimes able to merge short strings that are split over multiple lines, or split long strings that are on a single line, automating something that a human currently has to do (and sometime if you don't, black gives you `("first half " "second half")`. This feature exists in black, but has been unstable for years and so isn't generally usable.
    - Enabling new lints is easy since they're all baked in-- I tried to enable all of the ones that seemed reasonable and didn't cause a lot of failures with the current code base, which lead to me finding an actual `gettext` bug in the code.

I'll be honest too, even just looking at where ruff and black deviate in this code base, I pretty much find the ruff version in this PR easier to read, so that feels like a win too.

Unless folks are against this, I hope to merge it once it gets a review as it'll make warehouse a lot less frustrating for me to work on it 😄.

Closes #17570


warehouse/manage/forms.py
```diff
@@ -735,8 +734,9 @@ def validate__max_apps(self, field):
             self.form_errors.append(
                 _(
                     "You have already submitted the maximum number of "
-                    f"Organization requests ({self.max_applications})."
```
dstufft, line 738: Notice this is in a `gettext`, so this isn't actually valid :)

JacobCoffee (approved):
we will also want a `.git-blame-ignore-revs` but i dont guess you can do that before squash merging to get the real commit

woodruffw:
> we will also want a `.git-blame-ignore-revs` but i dont guess you can do that before squash merging to get the real commit

My only feedback too! Otherwise this LGTM.

woodruffw (approved):

pyproject.toml
```diff
@@ -132,3 +126,88 @@ filterwarnings = [
...
+  "SLOT",
+  # flake8-logging; enforce rules around good logging use
+  "LOG",
+  # flake8-logging-format; enforce the use of proper logging format strings
+  "G",
+  # flake8-raise; remove extraneous parens on exception raises
+  "RSE",
+  # ruff; grab bag of ruff specific linting rules
+  "RUF",
+]
```
miketheman, line 156: curious - would these comments not follow cleanly on the same line as the election codes? Feels like this could be compressed to be a little more visual-parsing-friendly

pyproject.toml
```diff
@@ -132,3 +126,88 @@ filterwarnings = [
...
+"tests/unit/email/test_services.py" = ["PT009"]
+"tests/unit/test_views.py" = ["PT009"]
+# TID252: Pretty much all of our tests use relative imports.
+"tests/**/*.py" = ["TID252"]
+
+[tool.ruff.lint.isort]
+combine-as-imports = true
+lines-between-types = 1
+# Even though `cgi` isn't a stdlib module anymore, treat it as one for now
+extra-standard-library = ["cgi"]
```
miketheman, line 183: That's nice syntax

tests/unit/accounts/test_services.py
```diff
@@ -246,7 +246,9 @@ def test_check_password_ip_rate_limited(self, user_service, metrics):
     def test_check_password_invalid(self, user_service, metrics):
         user = UserFactory.create()
         user_service.hasher = pretend.stub(
-            verify_and_update=pretend.call_recorder(lambda L, r: (False, None))
+            verify_and_update=pretend.call_recorder(
+                lambda L, r: (False, None)  # noqa: N803
+            )
```
miketheman, line 251: Instead of disabling these, fix with a lowercase character? What does `L` mean here, and could we use something smarter and not have to disable? This should apply to most N803 cases - if we're already touching these lines, might as well fix them - or ship that as a fix prior to this patch if the intent is to preserve history for most non-lint-specific changes.

dstufft:
Yea, I was going to add a `.git-blame-ignore-revs`, but I can't until it's merged 😅 . Well and I also want to test it to make sure it's not going to do something silly on the lines that are actually added by this PR (such as the configuration in `pyproject.toml`.

miketheman (commented):
Historical context was that the `lint` step was often slowed by [curlylint which I replaced last year](https://github.com/pypi/warehouse/pull/18175). It also predated warehouse's use of `.git-blame-ignore-revs` - which [I added as part of another large refactor](https://github.com/pypi/warehouse/pull/18219) that would make git history messy.

My perspective back then was probably that the seconds adding up weren't worth the historical context loss and line churn for what was not a demonstrable boost of speed to the overall process, so it wasn't "worth it" at the time.

Thanks for digging into the specifics of the shared caching.

I ran a comparison locally to prove the speedup, and it does indeed look useful today:

```shellsession
❯ hyperfine \
    --warmup 1 \
    --command-name flake8isort \
    --prepare 'git switch main && make build' \
    'make lint' \
    --command-name ruff \
    --prepare 'git switch use-ruff && make build' \
    'make lint'
Benchmark 1: flake8isort
  Time (mean ± σ):     14.310 s ±  0.313 s    [User: 0.173 s, System: 0.080 s]
  Range (min … max):   13.848 s … 14.971 s    10 runs

Benchmark 2: ruff
  Time (mean ± σ):      5.873 s ±  0.246 s    [User: 0.165 s, System: 0.075 s]
  Range (min … max):    5.617 s …  6.421 s    10 runs

Summary
  ruff ran
    2.44 ± 0.12 times faster than flake8isort
```

A few inline notes, but otherwise generally good to go.

pyproject.toml
dstufft: I don't think I even tried TBH, I'm happy to do it that way though.

tests/unit/accounts/test_services.py
dstufft: Honestly, I just didn't want to go through and figure out which ones were required to be that way or not. They're mostly in stubs, and I would have to go through and figure out what the parameters of the thing we're stubbing is and whether they're treating them as keyword arguments or as positional. 

I also kind of tried to (mostly) take the approach to try and minimize "risky" changes, particularly to the test suite, and in my head changing the casing of a stub is risky, because it's harder to know from just looking at it whether or not it needs to be that way or not.

I do plan to spend some time going through the `noqa` comments we have in the code base, and figuring out if the issue that the lint is detecting can actually be fixed, and if so fixing it, but I wanted that to be separate from getting this landed.

pyproject.toml
dstufft: This is done now, I think it does read nicer with the comments on the same line. Only weirdness is `E` and `F`  both coming from `pycodestyle`. I just duplicated the comment because all of the options seemed less than ideal in different ways, and duplication seemed less worse to me.

dstufft:
https://github.com/pypi/warehouse/pull/19949 adds the `.git-blame-ignore-revs`.

PR #19952, opened 2026-04-27
fix: surface project permission errors before email/2FA checks on upload

### LINKED ISSUE #18575: On upload, check for verified email / 2FA after check for project permissions
**Describe the bug**
Releasing a new version of bird_tool_utils is disallowed - 
```
         User 'mitchac' does not have a verified primary email address. Please  
         add a verified primary email before attempting to upload to PyPI. See  
         https://pypi.org/help/#verified-email for more information.    
```
This was true yesterday, but I've since removed mitchac as a maintainer, and retried. I'm not aware of any other way that user is associated with the project.

**Expected behavior**
Release should work.

**To Reproduce**
GitHub action at  https://github.com/wwood/bird_tool_utils-python/actions/runs/17088791740/job/48537427150

TIA.

miketheman:
The verified-email and 2FA checks previously ran in the `ensure_uploads_allowed` decorator, ahead of the project permission check in the upload view. A user uploading with a token whose owner doesn't have permission to the target project (and who also lacks a verified email or 2FA) would see a confusing "verified email" error instead of the more useful "not allowed to upload to project X".

Move the user-only email and 2FA checks out of the decorator and into the view as `_ensure_user_can_upload`, called after the project permission check. For brand-new projects, the helper is also called before `create_project()` so we don't leave an empty project record behind on rejection.

Closes #18575

### DIFF
--- a/tests/unit/forklift/test_decorators.py
+++ b/tests/unit/forklift/test_decorators.py
@@ -65,11 +65,7 @@ class TestEnsureUploadsAllowed:
         req = pretend.stub(
             flags=pretend.stub(enabled=lambda f: False),
             identity=pretend.stub(),
-            user=pretend.stub(
-                username="some-user",
-                primary_email=pretend.stub(email="foo@example.com", verified=True),
-                has_two_factor=True,
-            ),
+            user=pretend.stub(),
         )
         resp = pretend.stub()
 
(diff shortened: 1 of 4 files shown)


dstufft:
Hmm. I do agree that the 2fa/email error coming before the permission is a little confusing in the specific case that you're using an API key for a user other than your own. 

That being said, that *feels* like a pretty niche situation, particularly now that trusted publishing exists? In order to get to the part where we're able to check permissions against a `Project`, we have to do work in the `file_upload` view that we're ultimately going to throw away. 

Now that we only support API keys and trusted publishing, I wonder if we couldn't just improve the error message to say something like `The user 'foo' associated with the API key does not have...` to make it more clear that the link to the user is coming through the API key?

Then we can continue to reject the upload before we have to invoke any machinery to even figure out what project they're trying to upload to, saving the view from having to do work that we can figure out ahead of time it's going to throw away.

WDYT?

miketheman:
> Hmm. I do agree that the 2fa/email error coming before the permission is a little confusing in the specific case that you're using an API key for a user other than your own.

I think that's missing the main issue - the fact that this user account that owns the API Token no longer has permissions on the project is the "problem" - the fact that the account doesn't have other security guards in place isn't helpful to the end user at this juncture.

> That being said, that _feels_ like a pretty niche situation, particularly now that trusted publishing exists? In order to get to the part where we're able to check permissions against a `Project`, we have to do work in the `file_upload` view that we're ultimately going to throw away.
> 
> Now that we only support API keys and trusted publishing, I wonder if we couldn't just improve the error message to say something like `The user 'foo' associated with the API key does not have...` to make it more clear that the link to the user is coming through the API key?

Can definitely add the "user via API token..." messaging to make that clearer, but that doesn't solve the order-of-operations problem raised in the original issue.

> Then we can continue to reject the upload before we have to invoke any machinery to even figure out what project they're trying to upload to, saving the view from having to do work that we can figure out ahead of time it's going to throw away.

The `file_upload` view doesn't use the Permissions approach used in most places in the view configuration - that's probably the "right solution" overall.
Maybe the longer-term solution for file_upload is to adopt the pyramid authorization approach and lean into a consolidated authZ approach - that's a lot more surgery than is probably justified to solve the User Experience of getting a helpful, actionable error message.

I'll add a commit for improving the message, but I can't see a spot yet where we check for permissions that doesn't throw away "some" work without trying to reshape everything to match the authorization policy approach.

miketheman:
Dug into what hoisting the project lookup ahead of form+metadata validation would actually save us. Form.validate() is regex and required-field checks. metadata.parse() is heavier - classifier set ops, Requirement parsing, URL/email syntax - but it's all CPU. No queries, no external calls. ~10-40ms of CPU per rejected-by-permission request, that's it.

Not worth restructuring the view into two paths for that, especially since rate limits already cap the bad-token-volume scenario where it'd start mattering.

Going to leave the lookup where it is and keep the message clarification. We can come back to hoist it higher if it becomes a performance issue.

dstufft (approved):

PR #19953, opened 2026-04-27
(POC) Investigate uv

JacobCoffee:
**Very POC** to see how well we could fare from having modern tooling that wasn't super slow in local development.

Test swappin pip -> uv for local dev. Could also try pixi or something else with the uv resolver backend like pdm but was most familiar with this and we did the whole ruff-ification recently.

Initial commit (4e424ba) uses same `requirements/*.txt` format, no lockfile change for now. Would prefer modern pyproject.toml + lockfile if we were to proceed. Later commits modernizes further. but you can just checkout the initial.

Cold-cache, no-docker-cache benchmarks on my machine:

| What | pip / pip-compile | uv | speedup |
|---|---:|---:|---:|
| `make deps_upgrade_all` (resolve all 8 .in files) | **83.7s** | **7.4s** | **11.3x** |
| └ `main.in` alone | 39.2s | 1.9s | 20.3x |
| `make build` base image — install steps | 52.3s | 10.0s | 5.2x |
| `make build` docs image — install steps | 11.4s | 5.5s | 2.1x |
| Python install per teardown+rebuild (base+docs) | **92.3s** | **74.6s** | **−17.7s** |

bottleneck here is the frontend stuff, could swap to [rspack](https://rspack.rs/)

history here is pretty linear.
- [4e424ba](https://github.com/pypi/warehouse/pull/19953/commits/4e424ba3e7b048bd741b87b071c21980a8b8f6ff) will work standalone (not ci) to see just the uv change
- [958e1c2](https://github.com/pypi/warehouse/pull/19953/commits/958e1c27dee1964aab5f44d76f7ff841286e40a8) swapps to pyproject toml and lock file
- [ed45666](https://github.com/pypi/warehouse/pull/19953/commits/ed456669ea1c47caaee8ee43d4aed3205795e27b) swaps to docker stuff to frozen sync w/ groups trying to address @dstufft's PyPI circular-dep concern
- [b76e18f](https://github.com/pypi/warehouse/pull/19953/commits/b76e18f84ff21dd7b8bb38ebc186b005fd5b7654) domt forge tto relock but really we should NEVER have to do that bc uv tooling ensure relocks but "just in case" someone edits pyproject.toml manually ig uess?
- [78f3a53](https://github.com/pypi/warehouse/pull/19953/commits/78f3a53f4c58389655c47f38d52723902311db0b) for removing old style requirements in favor of pyproj. toml


JacobCoffee:
we continued poking at this, to see what ci and other things would look like.

**CI wall-clock comparison: pre-ruff baseline → ruff PR → uv migration**

GitHub Actions, same workflow (`.github/workflows/ci.yml`), same `depot-ubuntu-24.04-arm` runners. All three runs ran the same set of jobs and all passed.

| Job | #19941 (pre-ruff) | #19943 (ruff merged) | #19953 (uv migration) | uv vs pre-ruff |
|---|---:|---:|---:|---:|
| **build** (Docker image) | **197 s** | 173 s | **71 s** | **−64% / 2.77× faster** |
| Tests                    | 244 s | 228 s | 211 s | −14% |
| Lint                     | 100 s | 65 s  | 62 s  | −38% |
| Check DB Consistency     | 99 s  | 86 s  | 78 s  | −21% |
| User Documentation       | 85 s  | 52 s  | 51 s  | −40% |
| Developer Documentation  | 69 s  | 52 s  | 52 s  | −25% |
| Licenses                 | 69 s  | 64 s  | 52 s  | −25% |
| Translations             | 86 s  | 52 s  | 53 s  | −38% |
| Dependencies             | 95 s  | 76 s  | 50 s  | −47% |
| **Total wall (build + slowest matrix)** | **~441 s (7m21s)** | **~401 s (6m41s)** | **~282 s (4m42s)** | **−36% / 1.56× faster** |

The matrix jobs run in parallel, so total wall = build + slowest matrix job (Tests).

**Where the time went:**
- **Ruff alone (#19943)**: ~24 s off build (smaller lint deps to install) and ~35 s off Lint (one tool replaces four). Total wall: ~40 s.
- **uv on top (#19953)**: another ~102 s off build (`uv sync --frozen` reads URLs+hashes from the lock — no PyPI index lookups, no resolver round-trip), ~17 s off Tests (faster image = faster job startup), ~26 s off Dependencies (`uv lock --check` is sub-second vs the old `bin/deps`). Total wall: another ~119 s.

Stacked: **~159 seconds saved per CI run** vs pre-ruff. PRs running CI 5+ times across review iterations recover ~13 minutes of wall-clock per PR.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "types-s3transfer==0.16.0",
+    "types-setuptools==82.0.0.20260408",
+    "types-stripe==3.5.2.20240106",
+    "types-webencodings==0.5.0.20260408",
+    "types-webob==1.8.0.20260408",
+    "types-wtforms==3.2.1.20260408",
+    "types-zxcvbn==4.5.0.20260408",
+    "typing-extensions==4.15.0",
+    "typing-inspection==0.4.2",
+    "tzdata==2026.1",
```
dstufft:
```suggestion
```

If you delete this and re-run `uv lock`, it'll select one or two versions back and then will work on windows as well. Not the biggest deal either way, but nice to support

pyproject.toml
```diff
@@ -1,3 +1,480 @@
+[project]
+name = "warehouse"
+version = "0.0.0"
+description = "PyPI's web application"
+requires-python = ">=3.13"
```
dstufft:
```suggestion
requires-python = "==3.13.*"
```

we only support a single version at a time, so there's no reason to let other versions influence our lock file.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+dependencies = [
+    "alembic>=0.7.0",
+    "alembic-postgresql-enum",
+    "Automat",
+    "argon2-cffi",
+    "b2sdk",
+    "Babel",
+    "bcrypt",
+    "boto3",
+    "celery[redis]>=5.6.2,<5.7",
```
dstufft:
```suggestion
    "celery[redis]>=5.6.2",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "github-reserved-names>=1.0.0",
+    "google-cloud-bigquery",
+    "google-cloud-storage",
+    "hiredis",
+    "html5lib",
+    "humanize",
+    "itsdangerous",
+    "Jinja2>=2.8",
+    # https://github.com/jazzband/pip-tools/issues/1577
+    "kombu[redis]>=5.6,<5.7",
```
dstufft:
```suggestion
```

this was added due to a `pip-tools` bug that uv doesn't have so there's no reason to keep this.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "pytz",
+    "PyJWT[crypto]>=2.8.0",
+    "readme-renderer[md]>=36.0",
+    "requests",
+    "requests-aws4auth",
+    "redis>=2.8.0,<7.0.0",
+    "rfc3986",
+    "sentry-sdk",
+    "pypi-attestations==0.0.29",
+    "sqlalchemy[asyncio]>=2.0,<3.0",
```
dstufft:
```suggestion
    "sqlalchemy[asyncio]>=2.0",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "stdlib-list",
+    "stripe",
+    "structlog",
+    "tldextract",
+    "transaction",
+    "trove-classifiers",
+    "ua-parser",
+    "urllib3",
+    "webauthn>=1.0.0,<3.0.0",
```
dstufft:
```suggestion
    "webauthn>=1.0.0",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "WTForms[email]>=2.0.0",
+    "yara-x",
+    "zope.sqlalchemy",
+    "zxcvbn",
+]
+
+[dependency-groups]
+deploy = [
+    "gunicorn<24.0.0",  # pin to below 24 to avoid a restart error
+    "ddtrace==4.7.0",
```
dstufft:
```suggestion
    "ddtrace",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+]
+dev = [
+    "asyncudp>=0.7",
+    "hupper>=1.9",
+    "pyramid_debugtoolbar>=2.5",
+    "watchdog",
+]
+ipython = ["ipython>=9.12.0"]
+lint = [
+    "ruff==0.15.12",
```
dstufft:
```suggestion
    "ruff",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "boto3-stubs",
+    "types-certifi",
+    "types-first",
+    "types-html5lib",
+    "types-itsdangerous",
+    "types-passlib",
+    "types-python-slugify",
+    "types-pytz",
+    "types-redis",
+    "types-requests==2.33.0.20260408",
```
dstufft:
```suggestion
    "types-requests",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

For this pin specifically, it was added to work around an issue while we couldn't upgrade urllib3, but we have since upgraded urllib3 anyways and we've been upgrading this regularly.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+]
+tests = [
+    "coverage",
+    "factory_boy",
+    "freezegun",
+    "pretend",
+    "pytest>=3.0.0",
+    "pytest-icdiff",
+    "pytest-mock",
+    "pytest-postgresql>=3.1.3,<9.0.0",
```
dstufft:
```suggestion
    "pytest-postgresql>=3.1.3",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "cairosvg",
+]
+
+[tool.uv]
+package = false
+# Scope the lock to linux only; we deploy on linux (Docker) and the previous
+# requirements/*.txt files were single-platform too. Cross-platform locking
+# can be enabled later but currently surfaces win32 marker conflicts that
+# were silently ignored by pip-compile.
+environments = ["sys_platform == 'linux'"]
```
dstufft:
```suggestion
# Require that the package ships wheels for the following platforms:
required-environments = [
  # Linux on x86_64 (Intel/AMD)
  "sys_platform == 'linux' and platform_machine == 'x86_64'",
  # macOS on Apple Silicon (ARM)
  "sys_platform == 'darwin' and platform_machine == 'arm64'",
  # Windows on x86_64 (Intel/AMD)
  "sys_platform == 'win32' and platform_machine == 'AMD64'",
]
```

with the other changes, we no longer need to limit ourselves to only linux, and we can make it easier to install the dependencies on windows/macOS as well for people who want their editor to be able to find the deps by default.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+# Constrain all transitive deps to the versions pinned by the pre-migration
+# requirements/*.txt files. The migration to pyproject.toml + uv.lock should
+# not silently bump versions; future upgrades happen via deliberate PRs that
+# remove a constraint and re-lock.
+constraint-dependencies = [
```
dstufft, line 177: Once the `uv.lock` file has been added with these constraints, you can delete this from the `pyproject.toml`. 

Creating the `uv.lock` with these constraints will ensure the lock file is generated with these constraints in mind, and then once we have that `uv.lock`, uv will ensure that we don't change versions without an explicit upgrade happening.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
...
+    "python-slugify",
+    "pytz",
+    "PyJWT[crypto]>=2.8.0",
+    "readme-renderer[md]>=36.0",
+    "requests",
+    "requests-aws4auth",
+    "redis>=2.8.0,<7.0.0",
+    "rfc3986",
+    "sentry-sdk",
+    "pypi-attestations==0.0.29",
```
dstufft:
```suggestion
    "pypi-attestations",
```

Currently we're not really using these pins, we're regularly updating the upper bounds (via dependabot) so we're tracking the latest version. With `uv` an upper bounds won't get automatically bumped liked that, because the lock file is what keeps you from upgrading.

Makefile
```diff
@@ -38,7 +38,7 @@ default:
 	@echo
 	@exit 1
 
-.state/docker-build-base: Dockerfile package.json package-lock.json requirements/main.txt requirements/deploy.txt requirements/lint.txt requirements/tests.txt requirements/dev.txt
+.state/docker-build-base: Dockerfile package.json package-lock.json pyproject.toml uv.lock
```
dstufft, line 41: We probably only need this to depend on `uv.lock` and not `pyproject.toml`, the `uv.lock` file depends on the relevant content in `pyproject.toml`, but there is also irrelevant content in `pyproject.toml` that would needlessly trigger a rebuild.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
+[project]
+name = "warehouse"
+version = "0.0.0"
+description = "PyPI's web application"
```
dstufft:
```suggestion
```

this is a nit, but this field isn't required and there's no real benefit to specifying it that I can tell.

pyproject.toml
```diff
@@ -1,3 +1,480 @@
+[project]
+name = "warehouse"
+version = "0.0.0"
```
dstufft:
```suggestion
version = "1"
```

warehouse is stable 😉

.github/workflows/ci.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         with:
           path: |
               dev/.mypy_cache
-          key: ${{ runner.os }}-mypy-${{ env.pythonLocation }}-${{ hashFiles('requirements.txt', 'requirements/*.txt') }}
+          key: ${{ runner.os }}-mypy-${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml', 'uv.lock') }}
```
dstufft:
```suggestion
          key: ${{ runner.os }}-mypy-${{ env.pythonLocation }}-${{ hashFiles('uv.lock') }}
```

The `uv.lock` is enough to know if the dependencies has changed and avoids unrelated changes to `pyproject.toml` invalidating the cache.

dstufft:
You probably want to update the dependabot config to tell it to use uv, and the `docker-compose.yml` file to make sure that we're giving uv a cache location in `.cache` to persist with the other caches at runtime (used for `make deps` etc).

bin/lint
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
+# Fail if pyproject.toml has changed without regenerating uv.lock.
+uv lock --check
```
dstufft, line 5: This is really fast, so it's probably not a big deal either way, but given `make deps` still exists/works, that `uv` itself will generally enforce this, and CI has the Dependencies matrix item still, is this providing much value here?

I'm ok either way, just asking if it is :)

Dockerfile
```diff
@@ -43,6 +43,9 @@ RUN NODE_ENV=production npm run build
 # We'll build a light-weight layer along the way with just docs stuff
 FROM python:${PYTHON_IMAGE_VERSION} AS docs
 
+# Pull the uv binary in so we can use it as a faster pip / pip-compile.
+COPY --from=ghcr.io/astral-sh/uv:0.11.7 /uv /uvx /usr/local/bin/
```
dstufft, line 47: Can dependabot update this? 

Either way, we reference the hardcoded version several times, it would be good to extract that into a variable like was done with `PYTHON_IMAGE_VERSION`.

notatallshaw:
What version of pip was this benchmarked against?

I don't know where bottlenecks are in this CI but there were big improvements to metadata processing, resolver heuristics, and memory consumption in 26.1. I'm not saying to not go with uv, it has a lot of functionality pip never will, but my long term aim is performance parity.

JacobCoffee:
it may have been 26.0, ill retry with latest

PR #19954, opened 2026-04-28
(POC): Investigate webpack and frontend tooling alternatives

JacobCoffee:
A (fully vibed) continuation POC of [pypi/warehouse#19953](https://github.com/pypi/warehouse/pull/19953) but for frontend tooling:
- `npm` → `bun`
- `webpack` → `rspack`
- `eslint` → `oxlint` + `oxfmt`

| Swap | Before | After | Notes |
|---|---:|---:|---|
| Custom `WebpackLocalisationPlugin` (200 LOC, webpack-internal Parser hooks) | n/a | **30 LOC `defineLocaleConstants()` + `DefinePlugin`** | Door-opener for *any* modern bundler. |
| `npm ci` (cold) | ~14 s | **~7-8 s** (bun) | `bun.lock` (270 KB) replaces `package-lock.json` (540 KB). |
| `webpack` build | ~24-25 s | **~2 s rspack** + ~21 s post-build = **~23 s total** | `mini-css-extract` / `copy-webpack` / `webpack-manifest` / compression / image-min all swapped or replaced (inline manifest plugin + `bin/post-build.mjs`). |
| `eslint` | ~700-800 ms lint, ~1 s wall | **9 ms oxlint + 27 ms oxfmt --check, ~1 s wall** | Stylistic rules deferred to `oxfmt` (one-shot reformat of 53 of 71 files). `eslint.config.mjs` deleted; `@stylistic/eslint-plugin`, `eslint`, `@eslint/js`, `globals` dropped. |

**oxlint caught 5 real bugs eslint missed:**

1. `removeEventListener("submit", this.check.bind(this))` — `.bind()` returns a *new* function, so the listener never matches → never gets removed (memory leak + hidden double-firing risk if reconnected).
2-5. `expect(...).toBeNull` (no parens) → reads the matcher as a property, never *calls* it → 4 tests silently passed without asserting anything.

All five fixed in this POC.

**Post-build (gzip/brotli/sharp/svgo) is now the long pole**, not rspack itself — that work used to live inside webpack's plugin chain, now runs in a separate node script (`bin/post-build.mjs`) and is parallelisable as a follow-up if needed.

Closing now, but just for ref. later because I imagine we'd want to move off of slow webpack into vite or rspack, into tooling like https://oxc.rs/ suite or biome.js, etc.


JacobCoffee:
**CI wall-clock comparison: `poc/uv-localdev` (uv migration only) → `poc/frontend-modernize` (uv + bun + rspack + oxlint/oxfmt).**

Same workflow (`.github/workflows/node-ci.yml`), same `ubuntu-24.04` runners. Three matrix jobs, run in parallel.

| Job | uv-localdev baseline | frontend-modernize | Δ |
|---|---:|---:|---:|
| Static Lint (`bin/static_lint`: `bun run lint` + `bun run stylelint`) | 34 s | **24 s** | **−29% / 1.4×** |
| Static Tests (`bin/static_tests`: jest 92/92) | 29 s | **21 s** | **−28% / 1.4×** |
| Static Pipeline (`bin/static_pipeline`: build + post-build + sourcemaps) | 110 s | **54 s** | **−51% / 2.0×** |
| **Total Node CI wall (max parallel job)** | **110 s (1m50s)** | **54 s (54s)** | **−51% / 2.0×** |

**Where the time went:**
- **Static Pipeline** is where it shows up most — `bun install` (~7 s vs npm's ~14 s) + `rspack build` (~2 s vs webpack's ~25 s) + `bin/post-build.mjs` (~21 s, new — the gzip/brotli/sharp/svgo work that webpack did in-band).
- **Static Lint** runs `oxlint` (9 ms) instead of `eslint` (~700 ms); rest is `stylelint` + Node startup, unchanged.
- **Static Tests** is jest in both cases; the small win comes from `bun run` having faster startup than `npm run`.

**Stacked vs main:** with PR #19953 already merging the uv work for −36% on the main CI workflow (7m21s → 4m42s), this PR's frontend work piles ~−1 minute onto Node CI specifically. Combined: PR-iteration cycles see a much smaller frontend feedback loop on every commit (54 s vs 110 s = back to your editor 56 s sooner per push).

PR #19963, opened 2026-04-29
Use ast-grep to implement custom linting

dstufft:
This uses `ast-grep`, which can pattern match against the AST in a tree, to implement most of our custom linting rather than using custom shell scripts or flake8.

The only thing that flake8 is doing now is implementing the `@view_config` linter that checks for a missing template.

That's not possible to implement in `ast-grep` since it requires more than just greping the AST (you have to pull the filename out), but I think we can move that to be a test that uses Pyramid's [introspector](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/introspector.html), which would allow us to ditch flake8 completely then.


dstufft:
FWIW here's the timings for each step of `bin/lint` on my desktop computer with this PR (the desktop is faster than my laptop, so it runs flake8 in just under 4s instead of 6s):

* `ast-grep` at 0.255s
* `ruff check` at 0.080s
* `ruff format` at 0.065s
* `python -m flake8` at 3.765s
* `sphinx-lint` at 0.594s
* `python -m djlint` at 3.947s
* `python -m mypy` at 7.281s.

Here's the timings with the `main` branch:

* `ruff check` at 0.056s
* `ruff format` at 0.050s
* `python -m flake8` at 3.483s
* `sphinx-lint` at 0.564s
* `python -m djlint` at 3.754s
* `python -m mypy` at 7.485s
* `bin/flushes` at 0.398s

On my desktop this PR itself is roughly neutral in terms of time spent in linting in the dev iteration loop, but it's a step towards a 22% (on my desktop) to 30% (on my laptop) reduction in the loop.

`ast-grep` also has editor plugins that surface the lints it finds as LSP red (or yellow if you want) squiggle, which means for things like `bin/flushes` we enable shifting those issues closer to the time of development, so you can see those issues as you're writing the code rather than at some point in the future when you run the linters.

sgconfig.yml
```diff
@@ -0,0 +1,6 @@
+ruleDirs:
+  - dev/ast-grep/rules
+testConfigs:
+  - testDir: dev/ast-grep/rule-tests
+utilDirs:
+  - dev/ast-grep/utils
```
miketheman, line 6: Is there anything in this directory, or is this for some future effort?

warehouse/accounts/services.py
```diff
@@ -779,6 +782,7 @@ def device_is_known(
                 + datetime.timedelta(seconds=token_service.max_age),
             )
             request.db.add(unique_login)
+            # ast-grep-ignore: db-flush
```
miketheman: The main point of `bin/flushes` was to avoid needing to mark each one with an explicit ignore, rather ensure that there's a comment following it so at the very least we had some context. Can ast-grep not detect that?

bin/flushes
```diff
@@ -1,31 +0,0 @@
...
-    )
-    self.db.add(my_object)
-    self.db.flush()  # flush the db now so my_object.id is available
-    method_that_depends_on_id(my_object)
-
-is a common pattern when auto-generated ids are needed.
-
-Parts of the warehouse codebase (namely our cache_key/purge_key machinery)
-react to flushes and may have unintended side-effects as a result of calls to
-flush().
```
miketheman, line 25: This helpful description and context is lost, making it less clear to the next contributor on why this is the case.

miketheman (commented):
Timings are cool. See inline with some questions about the implementation for flushes

sgconfig.yml
dstufft: Nothing in there, it was just created by the `ast-grep init` command to place utility rules, but we don't have any of those. We could delete it, I just didn't bother since it felt harmless to have it.

warehouse/accounts/services.py
dstufft: A comment is just an ast node, so the rule can be marked as having to precede a comment node, but it feels better to me to have an explicit ignore-- we're assuming that a comment following it is providing context, but there's no reason that it has to be? Something like:

```python

self.db.flush()  # typing: ignore
```

or

```python

self.db.add(my_object)  # first add
self.db.flush() # then flush
```

would be allowed by `bin/flushes` with no real indication that it even *meant* to ignore the lint (or that it even knew that there was a lint that they were suppressing).

Maybe those are unlikely to happen, but they're definitely *possible* (and I think it's generally considered a less than ideal practice to suppress lints implicitly, or even broadly with just `# noqa` or what not).

On the flip side, I don't see much downside to having an explicit "ignore this lint" marker so that we know that the person *meant* to ignore that lint, and it wasn't just accidental. The only thing I can think it is adds a tiny little bit of extra noise, but that feels pretty minor given how few cases of them we have?

bin/flushes
dstufft: Image

updated!

warehouse/accounts/services.py
miketheman: I think the "tiny bit of noise" is the thing I'm trying to defend against. I count over 20 of these adding a line, where we could have it trail inline and less flow-breaking.

Maybe a happy medium is that the `# ast-grep-ignore: db-flush` needs to be on the same line as the db.flush() call - that way it doesn't increase overall line count, and trails behind the comment that explains what is happening?

```
request.db.flush()  # To get the ID for the token # ast-grep-ignore: db-flush
```
That's very targeted and explicit, and is getting closer to a line length limit, but maybe that's an indicator to use more succinct language?

Instead of:

```python
        self.db.flush()  # flush the db now so email.id is available # ast-grep-ignore: db-flush
```
we write something like:

```python
        self.db.flush()  # Generate email.id # ast-grep-ignore: db-flush
```
72 characters

or:
```python
        self.db.flush()  # Generate email.id from DB # ast-grep-ignore: db-flush
```
A perfect 80, but will be longer for objects that have longer names

dev/ast-grep/rules/db-flush.yml
```diff
@@ -0,0 +1,31 @@
...
+
+severity: error
+message: manually flushing the database session
+note: |
+  Use of flush() is _often_ unnecessary and should be annotated with a comment:
+
+      obj = MyObject()
+      self.db.add(obj)
+      # ast-grep-ignore: db-flush
+      self.db.flush()  # flush the db now so obj.id is available
```
miketheman:
```suggestion
      self.db.flush()  # Generate obj.id # ast-grep-ignore: db-flush
```

See discussion in warehouse/accounts/services.py

bin/reformat
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 set -ex
 
+# We do this before ruff so that ruff will be able to format any fixes that
+# come out of this.
+#
+# NOTE(dstufft): We use || true so that we don't fail if there's a lint rule
+#                that can't be fixed. The lint step will catch that but this
+#                means the rest of this can run without issue.
```
miketheman: Do you have an example of a rule that can't be fixed that we'd allow to pass?

bin/lint
```diff
@@ -5,16 +5,17 @@ set -ex
 # When on GitHub Actions, configure ruff to use the github output style
 if [ -n "$GITHUB_ACTIONS" ]; then
   export RUFF_OUTPUT_FORMAT=github
+  export EXTRA_AST_GREP_ARGS="--format=github"
 fi
 
 # Actually run our tests.
+ast-grep scan --error=unused-suppression --error=no-suppress-all $EXTRA_AST_GREP_ARGS
```
miketheman, line 12: This executes deceivingly fast, which made me question whether it was working at all. Useful:

```shellsession
$ ast-grep scan --inspect summary
sg: summary|project: isProject=true,projectDir=/Users/miketheman/workspace/miketheman/warehouse
sg: summary|file: scannedFileCount=871,skippedFileCount=0
sg: summary|rule: effectiveRuleCount=3,skippedRuleCount=0

$ git ls-files "*.py" | wc -l
     871
```

Matches! Super fast, very nice.

bin/reformat
dstufft: Maybe it'd be clearer to say *automatically* fixed?

Currently none of the `ast-grep` rules have a `fix:` section, because all of them require a human to do something:

1. The `db-flushes` rule requires a human to evaluate whether or not the flush is actually warranted.
2. The `backref` rule requires a human to setup the other side of the relationship so that back_populates can work.
3. The metrics tag one.. maybe it could apply a fix if you use the wrong kind of sequence (e.g. tuple to a list?), but in general if you pass like a dictionary or something there's probably not a good auto fix to use for it?

So right now reformat calling into `ast-grep` is kind of silly/pointless, but I added it because it's super fast and it means if someone adds a rule that has a `fix:` section it just works without having to remember to make this change in the future.

miketheman: That makes more sense now. I guess then if someone does as a rule-with-fix and it magically works, then the comment is outdated and if the fix fails it won't error? Maybe it's better to leave a note uncomment `# ast-grep scan --update-all` when we have a rule with a fix, and delete the existing comment, so we don't end up with competing docs vs behaviors

dstufft: I'll just drop it completely for now and when/if we get something that we can automatically fix, we can figure out what the right behavior is then instead of trying to guess now.

warehouse/utils/project.py
```diff
@@ -177,6 +177,7 @@ def remove_project(project, request, flash=True):
     )
 
     request.db.delete(project)
+    # ast-grep-ignore: db-flush
```
dstufft: I left this as a "preceding line" ignore comment because adding it made the line too long, and I'm not sure I understand what the existing comment actually means to be able to shorten it to something without possibly changing the meaning.

warehouse/accounts/services.py
dstufft: That seems reasonable to me, I went ahead and updated the recommendation and the suppressions that I added (except for one of them, noted in a different comment).

warehouse/utils/project.py
miketheman: Origin: https://github.com/pypi/warehouse/pull/2977/changes/6c94e8570b11f1021b7033adf407e1125d7a5522 probably to support this loop: https://github.com/pypi/warehouse/pull/2977/changes/79d832dc905798067c13305f1cd2c899b9837114#diff-ceb956331d4c25c5969a84f065ac96d4f802b383dfb8ee781614626d0dbf41a9R141-R143
Loop removed in https://github.com/pypi/warehouse/commit/b378e8ede89416a32f968a55f14e8efed9001320#diff-ceb956331d4c25c5969a84f065ac96d4f802b383dfb8ee781614626d0dbf41a9L149-R158

Probably safe to remove the flush now. Easy enough to test in dev - nuke a user with multiple projects.

miketheman (changes requested):
One final change and good to go.

warehouse/utils/project.py
dstufft: Updated, two of the tests depended on the flush to happen to update the state of the database, so I added the flush to those tests (in the real world our transaction tween makes sure that happens).

miketheman (approved):

PR #19966, opened 2026-04-29
make the base locale file unowned

dstufft:
In https://github.com/pypi/warehouse/pull/19938 we made a frontend team so that the frontend permissions could be more finely grained scoped to just what the frontend actually needed, with a secondary goal of reducing the amount of noise that someone working only on the frontend would actually get.

That secondary goal ended up not working very well due to the fact that we wanted to allow the frontend team to update the translations without needing a *second* review from the general reviewers team, which meant that they get pulled into any PR that updates `warehouse/locale/messages.pot`.

Unfortunately *most* PRs end up updating `warehouse/locale/messages.pot` because it embeds line numbers from where it pulled the translatable strings from, which meant that most PRs still ended up tagging the front end team.

However, `warehouse/locale/messages.pot` is deterministically generated from the contents of the repository and our test suite ensures that it matches what is expected, so we don't actually need a codeowner for this file at all. The files that we extract translations from will have a codeowner to enforce review, and then our test suite will enforce that this file matches what is expected.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -22,13 +22,17 @@ package.json                             @pypi/warehouse-reviewers @pypi/warehou
 webpack.config.js                        @pypi/warehouse-reviewers @pypi/warehouse-frontend
 webpack.plugin.localize.js               @pypi/warehouse-reviewers @pypi/warehouse-frontend
 
-# This is not *technically* a frontend file, but editing the frontend will likely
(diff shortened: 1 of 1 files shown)


dstufft:
@nlhkabu for visibility :)

miketheman (approved):

miketheman:
Related to https://github.com/pypi/warehouse/issues/6631

PR #19974, opened 2026-04-29
pyproject.toml indentation and linting

dstufft:
* The `pyproject.toml` file had inconsistent indentation, so it was impossible to auto format it (which many editors do by default) without a bunch of unrelated changes, so fix it so that it has consistent indentation (I picked 2 because most of the file was using 2).
* Add `tombi` to lint/format the `pyproject.toml` (~150ms) to keep it well formatted into the future.
* Use a shared `.cache` directory for everything instead of individual directories to make managing them easier and store it in a docker volume instead of on the host so cross platform issues don't arise.


bin/lint
```diff
@@ -15,3 +15,7 @@ sphinx-lint --enable=all --disable=line-too-long README.rst CONTRIBUTING.rst doc
 python -m djlint --check --lint ./warehouse/templates ./docs/blog
 python -m mypy -p warehouse
 ./bin/flushes
+
+# lint our pyproject.toml file (as well as any other toml files)
+tombi lint --error-on-warnings
```
JacobCoffee, line 20: could we ever move into `[project.scripts]` or entrypoints instead of these bin calls or is that bad?

JacobCoffee (approved):

dstufft:
Forgot to update the CI cache dirs, so doing that real quick as well :)

bin/lint
dstufft: Maybe? I don't recall what the limitations of `[project.scripts]` is tbh, not all of these things are python though, like `bin/static_lint` is calling `npm` stuff.

JacobCoffee: oh true!

.gitignore
```diff
@@ -27,11 +27,6 @@ dev/example.sql
 dev/prod.sql
 dev/prod.sql.xz
 dev/.coverage*
-dev/.mypy_cache/
-dev/.pip-cache/
-dev/.pip-tools-cache/
-dev/.pytest_cache/
-dev/.ruff_cache/
```
miketheman, line 34: why'd all the cache directories move?

dstufft: When I add tombi it had yet another cache directory, felt silly to individually maintain an ignore for each one and the fact we're sharing the caches with the host means that if any of those caches have platform specific stuff in them, then we can end up in awkward positions where two different platforms are writing to the same, platform specific cache directory.

We also had the `.stylelintcache` which wasn't being shared with the host, and was stored in a completely different place than the other caches.

So a single `.cache` directory means we don't have to maintain a list of what caches we have, and it's stored in a docker volume (which is what stylelint was setup to do) so that we don't share it with the host system so there's no cross platform issues that can arise.

PR #19986, opened 2026-05-01
more linting

dstufft:
I guess I lied about being done messing with linting for a bit 😅 , turns on a few more of the disabled lints, again I think it's made some improvements, and also has caught some actual issues.

### DIFF
--- a/gunicorn-prod.conf.py
+++ b/gunicorn-prod.conf.py
@@ -28,4 +28,4 @@ control_socket_disable = True
 
 
 def when_ready(server):
-    open("/tmp/app-initialized", "w").close()
+    open("/tmp/app-initialized", "w").close()  # noqa: S108
--- a/gunicorn-prod.conf.py
+++ b/gunicorn-prod.conf.py
@@ -25,7 +25,3 @@ statsd_host = "localhost:8125"
 # Instead of overwriting `XDG_RUNTIME_DIR` to a writable target,
 # which could affect other code, disable the interface as we are unlikely to need it.
 control_socket_disable = True
-
-
-def when_ready(server):
-    open("/tmp/app-initialized", "w").close()  # noqa: S108
--- a/gunicorn-uploads.conf.py
+++ b/gunicorn-uploads.conf.py
@@ -31,4 +31,4 @@ control_socket_disable = True
 
 
 def when_ready(server):
-    open("/tmp/app-initialized", "w").close()
+    open("/tmp/app-initialized", "w").close()  # noqa: S108
--- a/gunicorn-uploads.conf.py
+++ b/gunicorn-uploads.conf.py
@@ -28,7 +28,3 @@ statsd_host = "localhost:8125"
 # Instead of overwriting `XDG_RUNTIME_DIR` to a writable target,
 # which could affect other code, disable the interface as we are unlikely to need it.
 control_socket_disable = True
-
-
-def when_ready(server):
-    open("/tmp/app-initialized", "w").close()  # noqa: S108
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -163,7 +163,7 @@ ignore = [
   "TD",  # this wants TODO comments to have author and link, which we don't do
   "ERA",  # this has a REALLY high false positive rate
   "ANN",  # requires typing annotations, which has almost 15k errors for us
-  "S",  # high false positives rate, checks for security issues so would be nice to use
+  "S101",  # We use assert in the test suite, but more importantly for type narrowing
   "FBT",  # rejects boolean params that change function behavior, requires api changes
   "B019",  # we have some uses of this, we should probably fix them
   "B904",  # it's silly to require raise E from ... when Python mostly does it by default
@@ -195,7 +195,8 @@ ignore = [
 "tests/unit/email/test_services.py" = ["PT009"]
 "tests/unit/test_views.py" = ["PT009"]
 # TID252: Pretty much all of our tests use relative imports.
-"tests/**/*.py" = ["TID252"]
+# S: The test suite doesn't need to worry about security issues.
+"tests/**/*.py" = ["TID252", "S"]
 # INP001: Our migrations are scripts not files that are part of a package
 "warehouse/migrations/**/*.py" = ["INP001"]
 # The warehouse cli can use print to give feedback to the user.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -171,7 +171,6 @@ ignore = [
   "DTZ",  # prevents using naive datetimes, probably useful, but needs manual review
   "TRY",  # would require reworking some of our exceptions to make this work
   "EM",  # would require reworking some of our exceptions to make this work
-  "PLC",  # a lot of rules, some might be good, a lot are noise but needs review
   "PLR2004",  # we use a lot of magic values, maybe we shouldn't but for now allow it
   "PLR0915",  # too many statements; I'm suspect on the value of the "too many X" rules
   "PLR0912",  # too many branches; I'm suspect on the value of the "too many X" rules
@@ -199,7 +198,8 @@ ignore = [
 # INP001: Our migrations are scripts not files that are part of a package
 "warehouse/migrations/**/*.py" = ["INP001"]
 # The warehouse cli can use print to give feedback to the user.
-"warehouse/cli/**/*.py" = ["T201"]
+# The warehouse cli also defers imports on purpose.
+"warehouse/cli/**/*.py" = ["T201", "PLC0415"]
 # INP001: These are just scripts as well.
 # T201: These scripts/helpers are fine to call print
 "bin/*.py" = ["INP001", "T201"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,39 +189,36 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 # N805: zope.interface definitions look like classes, but are not
 "interfaces.py" = ["N805"]
-# PT009: These files use a lot of custom assertion tools that trip up this detection.
-"tests/unit/email/test_init.py" = ["PT009"]
-"tests/unit/email/test_services.py" = ["PT009"]
-"tests/unit/test_views.py" = ["PT009"]
-# TID252: Pretty much all of our tests use relative imports.
-# S104: tests use 0.0.0.0 as a sentinel value sometimes
-# S105: tests aren't used in production and can use hardcoded pws for tests
-# S106: tests aren't used in production and can use hardcoded pws for tests
-# S108: tests use hardcoded tmp paths that get stubbed out
-# S303: strength of the hash functions used in tests doesn't matter
-# S311: tests don't need cryptographically strong randomness
-# S324: strength of the hash functions used in tests doesn't matter
-"tests/**/*.py" = [
-  "TID252",
-  "S104",
-  "S105",
-  "S106",
-  "S108",
-  "S303",
-  "S311",
-  "S324"
-]
+
 # INP001: Our migrations are scripts not files that are part of a package
 "warehouse/migrations/**/*.py" = ["INP001"]
+
 # The warehouse cli can use print to give feedback to the user.
 # The warehouse cli also defers imports on purpose.
 "warehouse/cli/**/*.py" = ["T201", "PLC0415"]
+
 # INP001: These are just scripts as well.
 # T201: These scripts/helpers are fine to call print
 "bin/*.py" = ["INP001", "T201"]
 "dev/*.py" = ["INP001", "T201"]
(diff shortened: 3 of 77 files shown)


tests/unit/email/test_services.py
```diff
@@ -337,12 +336,10 @@ def test_last_sent(self, db_session):
             aws_client, sender="DevPyPI <noreply@example.com>", db=db_session
         )
         for address in [to, "somebody_else@example.com"]:
-            for subject in [subject, "I do not care about this"]:
+            for s in [subject, "I do not care about this"]:
```
dstufft, line 339: It took me longer than I'd like to admit to even understand what the original code was doing to make sure that this change made sense, so I think this is a real big improvement in readability.

tests/unit/integration/secrets/test_utils.py
```diff
@@ -28,7 +28,7 @@ def test_disclosure_origin_serialization(someorigin):
 
 
 def test_disclosure_origin_equivalence(someorigin):
-    assert someorigin == someorigin
+    assert someorigin == someorigin  # noqa: PLR0124
```
dstufft: I think this could probably be removed? but it's not clear if it's actually doing something that I'm missing, so I left it in since it's just in the test suite so relatively harmless.

tests/unit/oidc/models/test_github.py
```diff
@@ -606,7 +606,7 @@ def test_github_publisher_job_workflow_ref(self, claim, ref, sha, valid, expecte
             assert check(publisher.job_workflow_ref, claim, claims) is True
         else:
             with pytest.raises(errors.InvalidPublisherError) as e:
-                check(publisher.job_workflow_ref, claim, claims) is True
+                check(publisher.job_workflow_ref, claim, claims)
```
dstufft, line 609: This looks like it's missing an `assert`, but that's actually a trick because `check()` is expected to raise and so the `is True` is meaningless even if there was an `assert`.

tests/unit/oidc/models/test_gitlab.py
```diff
@@ -678,7 +678,7 @@ def test_gitlab_publisher_ci_config_ref_uri(
             assert check(publisher.ci_config_ref_uri, claim, claims) is True
         else:
             with pytest.raises(errors.InvalidPublisherError) as e:
-                check(publisher.ci_config_ref_uri, claim, claims) is True
+                check(publisher.ci_config_ref_uri, claim, claims)
```
dstufft, line 681: This looks like it's missing an `assert`, but that's actually a trick because `check()` is expected to raise and so the `is True` is meaningless even if there was an `assert`.

tests/unit/test_filters.py
```diff
@@ -298,7 +298,7 @@ def test_format_email(meta_email, expected_name, expected_email):
     ("inp", "expected"),
     [
         ("foo", "foo"),  # no change
-        (" foo  bar ", " foo  bar "),  # U+001B : <control> ESCAPE [ESC]
+        (" foo \x1b bar ", " foo  bar "),  # U+001B : <control> ESCAPE [ESC]
```
dstufft: there was an invalid escape sequence in this string, which meant that what it thought I was testing it actually wasn't (and now it correctly matches what the comment says).

pyproject.toml
```diff
@@ -145,15 +145,22 @@ ignore = [
   "TD",  # this wants TODO comments to have author and link, which we don't do
   "ERA",  # this has a REALLY high false positive rate
   "ANN",  # requires typing annotations, which has almost 15k errors for us
-  "S",  # high false positives rate, checks for security issues so would be nice to use
+  "S101",  # We use assert in the test suite, but more importantly for type narrowing
```
miketheman, line 166:
```suggestion
  "S101",  # We use assert for type narrowing
```
Already covered by the exclusion later on

pyproject.toml
```diff
@@ -171,11 +178,13 @@ external = ["WH"]  # These come from our internal flake8 plugin.
 "tests/unit/email/test_services.py" = ["PT009"]
 "tests/unit/test_views.py" = ["PT009"]
 # TID252: Pretty much all of our tests use relative imports.
-"tests/**/*.py" = ["TID252"]
+# S: The test suite doesn't need to worry about security issues.
```
miketheman: I mean, some of them it might? I'd hate for us to miss a sleeper `exec()` somewhere, so if we're taking a moment to fine-tune things, let's not explicity ignore all possible security findings, just because it's in tests.

pyproject.toml
```diff
@@ -171,11 +178,13 @@ external = ["WH"]  # These come from our internal flake8 plugin.
...
+"tests/**/*.py" = ["TID252", "S"]
 # INP001: Our migrations are scripts not files that are part of a package
 "warehouse/migrations/**/*.py" = ["INP001"]
 # The warehouse cli can use print to give feedback to the user.
-"warehouse/cli/**/*.py" = ["T201"]
+# The warehouse cli also defers imports on purpose.
+"warehouse/cli/**/*.py" = ["T201", "PLC0415"]
```
miketheman, line 198: 👏👏👏

pyproject.toml
```diff
@@ -145,15 +145,22 @@ ignore = [
...
   "TRY",  # would require reworking some of our exceptions to make this work
   "EM",  # would require reworking some of our exceptions to make this work
-  "PL",  # a lot of rules, some might be good, a lot are noise but needs review
+  "PLR0911",  # too many return statements; I'm suspect on the value of the "too many X" rules
+  "PLR0912",  # too many branches; I'm suspect on the value of the "too many X" rules
+  "PLR0913",  # too many arguments; I'm suspect on the value of the "too many X" rules
+  "PLR0915",  # too many statements; I'm suspect on the value of the "too many X" rules
+  "PLR2004",  # we use a lot of magic values, maybe we shouldn't but for now allow it
+  "PLW1641",  # we don't need every object to implement __hash__
+  "PLW2901",  # it's fine to overwrite a for loop variable
```
miketheman: Curious - for these, is it that there's too high a finding count, or that you think it's not worth it? I personally have been confused by `PLW2901` a few times before, and renaming helped both my brain and I think the type checker didn't like reassignment

pyproject.toml
```diff
@@ -145,15 +145,22 @@ ignore = [
   "TD",  # this wants TODO comments to have author and link, which we don't do
   "ERA",  # this has a REALLY high false positive rate
   "ANN",  # requires typing annotations, which has almost 15k errors for us
-  "S",  # high false positives rate, checks for security issues so would be nice to use
+  "S101",  # We use assert in the test suite, but more importantly for type narrowing
   "FBT",  # rejects boolean params that change function behavior, requires api changes
-  "B",  # catches a lot of funky patterns, but needs reviewed to make sure they're ok
+  "B019",  # we have some uses of this, we should probably fix them
+  "B904",  # it's silly to require raise E from ... when Python mostly does it by default
```
miketheman: We raise from None a bit: https://github.com/search?q=repo%3Apypi%2Fwarehouse+%2Fraise+.*+from+None%2F&type=code (regex isn't perfect, might not catch multiline raises)

I'm kind of a fan of making it a little more explicit whenever we re-raise, but I suspect that doing this too liberally could cut off context that sentry gets? I don't know well enough how the exceptions are tracked and handled with their middleware vs a `raise from None`

gunicorn-uploads.conf.py
```diff
@@ -24,4 +24,4 @@
 
 
 def when_ready(server):
-    open("/tmp/app-initialized", "w").close()
+    open("/tmp/app-initialized", "w").close()  # noqa: S108
```
miketheman: This is a holdover from heroku era, and the entire functions may no longer be needed.

warehouse/forms.py
```diff
@@ -21,9 +21,11 @@ class URIValidator:
     def __init__(
         self,
         require_scheme=True,
-        allowed_schemes={"http", "https"},
+        allowed_schemes=None,
         require_authority=True,
     ):
+        if allowed_schemes is None:
+            allowed_schemes = {"http", "https"}
```
miketheman, line 28: oh sneaky, caught!

tests/unit/integration/secrets/test_utils.py
miketheman: Seems likely to me as well. It kind of follows the test pattern above - assert something about `someorigin` and then assert other things. So probably better to remove.

```suggestion
```

tests/unit/test_filters.py
miketheman: Discussed on call, remove example as it's functionally equivalent to the test below it.

```suggestion
```

miketheman (commented):
A bunch of line comments and suggestions. I didn't get through all of it, or reason specifically about the implications of each change.

pyproject.toml
dstufft: This rule *requires* you to use either `raise E from None` OR `raise E from exc` when you're inside of an except handler like:

```python
try:
    something()
except Exception as exc:
    raise Thing  # not allowed
    raise Thing from exc  # allowed
    raise Thing from None  # allowed
```

Python treats `raise E` and `raise E from exc` as almost functionally identical (obviously it does not do that for `raise R from None`.

There's ~99 cases where we do `raise E` in an except handler, and skimming through them it looked like they were generally cases where it really didn't matter (HTTPExceptions that get turned into responses, form validation errors that get turned into errors on the form, etc).

There's maybe a few where `raise E from None` would be more clear?

Turning this lint on would require you to think about whether the context of the original exception matters or not, and I do think that has some value, but it's a hard one to make a blanket mechanical change for.

pyproject.toml
dstufft: I think this was just the first one I ignored and I assumed it had more findings than it did, and it didn't personally bother me very much. It's only 3 findings, so easy enough to fix.

pyproject.toml
dstufft: I went ahead and scoped the ignores down even in the test suite.

pyproject.toml
dstufft: Dropped this from the list of ignores and fixed the 3 places we were doing that.

pyproject.toml
dstufft: I left this in since 3f3e15ca4 now means that `S` is not ignored for the entire test suite.

gunicorn-uploads.conf.py
dstufft: removed!

pyproject.toml
miketheman: > Turning this lint on would require you to think about whether the context of the original exception matters or not, and I do think that has some value, but it's a hard one to make a blanket mechanical change for.

Then maybe we want the inverse? Annotate why someone is raising from None which is not the norm?

It's entirely possible my hesitance is that the choice of descriptor in the disable line - I don't find it at all silly to ask developers to be explicit - it's in the zen after all.

miketheman (commented):
I commented inline, that should "count" as a review but didn't

pyproject.toml
dstufft: I can reword the description here if that makes it better? Essentially I think the default behavior of `raise E` is basically fine for us in every case, where `raise E from exc` is better it's like, `0.01%` better, and where `raise E from None` is better it's like, 2% better.

The problem is you can't mechanically know if a given `raise E` should be `raise E from exc` or `raise E from None`, so this lint treats `raise E` as a problem, forcing the case where it's `0.01%` better to adapt to cover the case where it's 2% better, and I don't think either case is important enough for us to worry about whether someone did it or not.

In libraries or other code bases where the exception itself is part of the public API that you expect end users to see, I think I'd probably weigh that benefit a little higher?

In either case, I can re-word the rationale a bit here.

dstufft: done!

miketheman (approved):

lauedaitech50:
d39a5d12f4e9ae85058f0f2bdc01cc0d99678bdc

tests/unit/test_filters.py
lauedaitech50: d39a5d12f4e9ae85058f0f2bdc01cc0d99678bdc

lauedaitech50:
_script()google-antigravity 0.1.0pip install google-antigravity==0.1.0

tests/unit/cache/origin/test_fastly.py
natalejoe, line 246: @dstufft

gunicorn-prod.conf.py
natalejoe, line 1: ****

natalejoe (changes requested):

gunicorn-prod.conf.py
miketheman: Please stop spamming comments with no context. Otherwise we'd have to block your user account

PR #19993, opened 2026-05-01
chore(deps): unpin inputs where possible

miketheman:
Dependabot will run `pip-compile` for updates, and when it does, it bumps the values in the .in files as well as the rendered .txt files.

Touching the .in files impacts local developers needlessly running the same `pip-compile` steps that have already been done via CI, slowing them down when nothing has effectively changed.

This should minimize the amount of locally-run pip-compile.

### DIFF
--- a/requirements/deploy.in
+++ b/requirements/deploy.in
@@ -1,2 +1,2 @@
 gunicorn<24.0.0  # pin to below 24 to avoid a restart error
-ddtrace==4.7.0
+ddtrace
--- a/requirements/lint.in
+++ b/requirements/lint.in
@@ -1,4 +1,4 @@
-ruff==0.15.12
+ruff
 ast-grep-cli
 djlint
 flake8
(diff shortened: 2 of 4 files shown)


dstufft (commented):
There's a couple more of these that I found in the uv poc:

- `sqlalchemy[asyncio]>=2.0,<3.0`
- `webauthn>=1.0.0,<3.0.0`
- `pytest-postgresql>=3.1.3,<9.0.0`
- `pypi-attestations==0.0.29`
- `kombu[redis]>=5.6,<5.7`

All of those are ones where we're on the latest version and the upper bounds is not excluding anything today (so it's not adding any value at the moment), and most of them are ones where we've had dependabot bumping the upper bounds whenever a new release came out that was excluded by the pin.

That leaves only `redis>=2.8.0,<7.0.0`, which we are legitimately excluding versions using that pin.

Thanks for picking this up though!

dstufft (approved):

Issue #19985, opened 2026-04-30
Add secret scanning during file uploads


miketheman:
Originally discussed in #12412 

With the new YARA-X scanning added in #19584 we now have a richer ability to scan things and reject them on upload-time, preventing them from ever reaching long-term storage, mirrors, etc.

Some folks still make mistakes and either add things like a `.env` file, or paste their token in a text file/readme.

Today the scanner only scans specific file types, so this would probably mean we'd want to expand the scanning ability to pretty much "anything but large binaries".
We could look at doing a almost-everything file-extraction operation to spill items to disk, so that yara can manage their own scanning limitations, but that seems like a riskier move. We should probably only do that if other options don't satisfy our needs.

The actions to take when a secret is identified:
- reject the upload with a clear message why it was rejected
- invalidate the token, similar to the other token invalidation integrations we have today
- notify the Token owner via email about what happened (also similar to the secret scanning invalidation we have today)

kam193:
The first step could be just a sanity YARA check by extending scanned file extensions with the most typical places the tokens are left (*.md, *.txt, *.env, project description, ...) and rejecting uploads with the appropriate note.

I'd also note that the scope should be limited to PyPI own tokens only as we e.g. cannot easily determine whether secrets to other services were uploaded intentionally or not.

PR #19994, opened 2026-05-01
feat: prevent uploading PyPI tokens in common places

kam193:
The minimal implementation of #19985 to prevent users from uploading tokens in distribution archives. As the current scanner requires selecting extensions to scan, the list was extended with common places where they are left behind.

This change does not perform automatic token revoking. Only tokens for pypi.org are detected.

### DIFF
--- a/tests/unit/utils/test_scanner.py
+++ b/tests/unit/utils/test_scanner.py
@@ -5,9 +5,13 @@ import struct
 import tarfile
 import zipfile
 
+from uuid import uuid4
+
+import pymacaroons
 import pytest
 import yara_x
 
+from warehouse.macaroons import caveats
 from warehouse.utils import scanner
 
 
@@ -16,6 +20,41 @@ def rules():
     compiled = scanner.compile_rules()
     assert compiled is not None
     return compiled
+
+
+def _generate_token(domain="pypi.org", projects_scope=False):
+    raw_macaroon = pymacaroons.Macaroon(
+        location=domain,
+        identifier=str(uuid4()),
+        key=b"fake key",
+        version=pymacaroons.MACAROON_V2,
+    )
+
+    if projects_scope:
+        caveats_ = [caveats.ProjectID(project_ids=[str(uuid4()) for _ in range(3)])]
+    else:
+        caveats_ = [
+            caveats.ProjectName(normalized_names=[f"project-{i}" for i in range(3)]),
+            caveats.RequestUser(user_id=str(uuid4())),
+        ]
+    for caveat in caveats_:
+        raw_macaroon.add_first_party_caveat(caveats.serialize(caveat))
+
+    return f"pypi-{raw_macaroon.serialize()}"
+
+
+@pytest.fixture(
+    scope="module", params=[False, True], ids=["user-scope", "projects-scope"]
+)
+def pypi_token(request):
+    return _generate_token(domain="pypi.org", projects_scope=request.param)
+
+
+@pytest.fixture(
+    scope="module", params=[False, True], ids=["user-scope", "projects-scope"]
+)
+def localhost_token(request):
+    return _generate_token(domain="localhost", projects_scope=request.param)
 
 
 def _make_wheel(tmp_path, files_dict, name="fake_package", version="1.0"):
@@ -248,22 +287,22 @@ class TestScanArchive:
         )
         assert scanner.scan_archive(whl, rules=rules) == []
 
-    def test_skips_non_python_files_in_wheel(self, tmp_path, rules):
+    def test_skips_excluded_files_in_wheel(self, tmp_path, rules):
         whl = _make_wheel(
             tmp_path,
             {
                 "pkg/data.json": "__pyarmor__(__name__, __file__, b'x')",
-                "pkg/readme.txt": "__pyarmor_enter__()",
+                "pkg/module.so": b"__pyarmor_enter__()",
             },
         )
         assert scanner.scan_archive(whl, rules=rules) == []
 
-    def test_skips_non_python_files_in_tarball(self, tmp_path, rules):
+    def test_skips_excluded_files_in_tarball(self, tmp_path, rules):
         tar = _make_tarball(
             tmp_path,
             {
                 "fake-1.0/data.json": "__pyarmor__(__name__, __file__, b'x')",
-                "fake-1.0/readme.txt": "__pyarmor_enter__()",
+                "fake-1.0/module.so": b"__pyarmor_enter__()",
             },
         )
         assert scanner.scan_archive(tar, rules=rules) == []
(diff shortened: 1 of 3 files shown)


warehouse/utils/scanner_rules/pypi_token.yar
```diff
@@ -0,0 +1,24 @@
+rule secrets_pypi_token
+{
+	meta:
+		description = "Detects PyPI API tokens exposed in source code."
+		author = "Kamil Mankowski"
+		message = "We have detected a PyPI API token exposed in the uploaded file. Publishing it would allow anyone to perform actions on your behalf. For your own security, please revoke the token immediately."
```
miketheman: Probably should link to something like https://pypi.org/help/#compromised-token

warehouse/utils/scanner.py
```diff
@@ -16,9 +16,21 @@
 # YARA rules directory
 _RULES_DIR = Path(__file__).parent / "scanner_rules"
 
-# Extensions to scan inside archives. Python source (.py) for source-level
-# rules (e.g. pyarmor), and .pye for SourceDefender-encrypted files.
-_SCAN_EXTENSIONS = {".py", ".pye"}
+# Extensions to scan inside archives.
+_SCAN_EXTENSIONS = {
```
miketheman: Seems like the variable name and comment are no longer accurate, with the inclusion of the last two filenames.
Maybe a better name would be `_SCAN_ALLOWLIST` or `_SCANNER_TARGET_FILES`?

warehouse/utils/scanner_rules/pypi_token.yar
```diff
@@ -0,0 +1,24 @@
...
+		// TODO: look if there are test tokens in use we should exclude
+		// $test_token = "pypi-AgEIcHlwaS5vcmcCJxxx"
+
+	condition:
+		$pypi_token
+		// If we want to allow some test-only tokens, we can use:
+		// and for all i in (1 .. #pypi_token) : (
+		// 	not $test_token at @pypi_token[i]
+		// )
+}
```
miketheman: https://github.com/pypi/warehouse/blob/4a1d36e695c1fcaa4272a48cd5ffc9aaa51cf49a/.editorconfig#L9

Might want to set up [editorconfig](https://editorconfig.org/) locally?

tests/unit/utils/test_scanner.py
```diff
@@ -18,6 +22,41 @@ def rules():
     return compiled
 
 
+def _generate_token(domain="pypi.org", projects_scope=False):
```
miketheman, line 25: I can imagine a future factory-style generator to create tokens for testing, so this paves that future nicely.

tests/unit/utils/test_scanner.py
```diff
@@ -396,3 +435,48 @@ def test_spoofed_file_size_does_not_bypass_scan(self, tmp_path, rules):
...
+    "README.md",
+    "PUBLISHING.RST",
+    "publish.sh",
+    "info.txt",
+    ".env",
+]
+
+
+class TestPyPITokenDetection:
+    # TODO: separated METADATA/PKG-INFO tests with correct paths
```
miketheman: Can you clarify a bit what this TODO represents? Is it specifically to have the files be located in a subdirectory?

miketheman (changes requested):
Excellent start, thanks for taking a pass at it! I've dropped some comments inline.

tests/unit/utils/test_scanner.py
kam193: We want to catch tokens left in the project description in the METADATA/PKG-INFO. Current tests don't put them in e.g. `dist-info`, so it's a lie that it tests project description

warehouse/utils/scanner_rules/pypi_token.yar
kam193: (I definitely didn't hope you'd come with the link :roll_eyes:)

warehouse/utils/scanner_rules/pypi_token.yar
kam193: yeah, looks like I didn't install the extension on the new device :thinking:

kam193:
From looking for test tokens to exclude, the closest seem to be https://github.com/mongodb/kingfisher/blob/b2811107a814c117f6a54fd8d334aef52409bf8f/src/validation.rs#L1924 or https://github.com/cider-security-research/cicd-goat/blob/0ed10925f3983857cf219b2ac1c327b861fcccca/solutions/duchess.md?plain=1#L35 but the first won't land in any scanned file, and the second is not a PyPI project.

However, I've discovered how people present a placeholder for PyPI tokens, and for regression prevention, I've added them to the test base. This is handled well just with the length in the Yara rule, but I've left the guidelines to exclude other sample tokens specifically, in case one needs it in the future.

warehouse/utils/scanner.py
```diff
@@ -16,9 +16,25 @@
 # YARA rules directory
 _RULES_DIR = Path(__file__).parent / "scanner_rules"
 
-# Extensions to scan inside archives. Python source (.py) for source-level
-# rules (e.g. pyarmor), and .pye for SourceDefender-encrypted files.
-_SCAN_EXTENSIONS = {".py", ".pye"}
+# Extensions to scan inside archives.
+_SCAN_TARGETS = {
```
woodruffw, line 20: Nit: there might be a better way to express these targets? Maybe as globs, to make it clearer which ones are filenames versus suffixes?

warehouse/utils/scanner.py
```diff
@@ -16,9 +16,25 @@
...
+    # Python source for source-level rules (e.g. pyarmor)
+    ".py",
+    # .pye for SourceDefender-encrypted files
+    ".pye",
+    # Different textual files for common places where PyPI tokens are accidentally left.
+    ".md",
+    ".rst",
+    ".env",
```
woodruffw, line 28: Nit: it might be good to match _anything_ with `*env*` in the filename (including suffixes), since `.env.prod`, `.prod.env`, `env.py`, etc. are all conceivable.

warehouse/utils/scanner.py
kam193: I was thinking about globs (also to cover your second comment), but I decided it would be better to leave it for a separate change to better research what would be the best way for that. My personal preference would be scan every file under a size limit, but I think it requires more considerations

woodruffw: Makes sense, thanks. I'll defer to whatever others think makes sense for an initial merge here.

warehouse/utils/scanner.py
kam193: Yes - see the previous comment. As a bit of context, the current list is based on where I saw tokens recently.

warehouse/utils/scanner_rules/pypi_token.yar
```diff
@@ -0,0 +1,24 @@
+rule secrets_pypi_token
+{
+	meta:
+		description = "Detects PyPI API tokens exposed in source code."
```
miketheman, line 4: question: I think the other rules in the repo use spaces - which is more adopted by the community? whichever it is, it should be set it in the root `.editorconfig` to be consistent.

warehouse/utils/scanner_rules/pypi_token.yar
```diff
@@ -0,0 +1,24 @@
...
+		author = "Kamil Mankowski"
+		message = "We have detected a PyPI API token exposed in the uploaded file. Publishing it would allow anyone to perform actions on your behalf. For your own security, please revoke the token immediately. See https://pypi.org/help/#compromised-token for additional help."
+
+	strings:
+		// Regex adapted from trufflehog's PyPI token detector
+		// Intentionally not derived from the official Token format definition to spare unnecessary matches.
+		// Pre-computed head ensures we match actual pypi.org tokens
+		// https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/pypi/pypi.go
```
miketheman, line 12: great addition to using a reference - better to make it a permalink like https://github.com/trufflesecurity/trufflehog/blob/05a583290be8c8f79852eb8741f5042920b47d00/pkg/detectors/pypi/pypi.go#L28

warehouse/utils/scanner_rules/pypi_token.yar
```diff
@@ -0,0 +1,24 @@
...
+		$pypi_token = /pypi-AgEIcHlwaS5vcmcCJ[a-zA-Z0-9-_]{150,157}/
+
+		// TODO: look if there are test tokens in use we should exclude
+		// $test_token = "pypi-AgEIcHlwaS5vcmcCJxxx"
```
miketheman, line 16: I don't think there are any that I'd expect to be included in packages?

I have found examples of the exact same regex included in some pacakges that do their own scanning/detection - some examples:
https://github.com/langchain-ai/langsmith-sdk/blob/76e3468b8a83d2e3855b07f9ef92fde11a93f7e5/python/langsmith/anonymizer.py#L270
https://github.com/mistralai/client-python/blob/b0613c7bc6bbd87e3991c3673235d5f89a49db72/src/mistralai/extra/observability/redaction_policies.py#L141

I'd want to allow those through - how would this filter work for them? Probably good examples to add to the test cases.

miketheman (commented):
Coming along nicely - I've added some more notes inline, let me know if they aren't clear

Issue #5869, opened 2019-05-16
Feature flag system


brainwane:
**What's the problem this feature will solve?**
It's annoying to roll out changes to Test PyPI and not PyPI.org, or to only certain users -- we're running into that in https://github.com/pypa/warehouse/issues/5661#issuecomment-490267708 . For instance, we can't currently change configuration to indicate "this should be enabled for all users on test.pypi.org, including new users created after we flip the switch, but only per-individual on pypi.org".

**Describe the solution you'd like**
A feature flag system.

* Most important is the ability to, either via a config file or a GUI interface for PyPI admins, enable a feature for all users on test.pypi.org, including new users created after we flip the switch, but not have that affect default users on pypi.org.
* Nice-to-have: a way for users to opt in to beta test new features in their account settings, so we don't have to create and run separate Google forms and similar mechanisms for beta periods.

di:
Bumping this. I'd like us to experiment with using a standalone feature flagging service, and https://flagsmith.com/ looks like a good option:

- we can stand it up ourselves for free
- it's open-source
- it's Python

Key features we'll want to check for when evaluating this:

- ability to set feature flags per-user
- ability to segment users based on some arbitrary characteristics (e.g. what domain they have verified emails for, whether they have 2FA enabled, etc)
- ability to fail-closed in a reasonable way

Steps towards resolving this:

- [ ] prototype using this in local development by adding it to our docker compose services
- [ ] stand up an an instance in our staging environment
- [ ] deploy this to staging
- [ ] stand up an an instance in our production environment
- [ ] deploy this to production

PR #19996, opened 2026-05-01
feat(oidc): expire pending trusted publishers after a TTL

miketheman:
Pending OIDC publishers that are never used can accumulate indefinitely. Expire them after `PENDING_PUBLISHER_EXPIRY_DAYS` (30 days) to limit the window during which a stale or speculative pending publisher can be reified into a real project.

- daily task purges expired rows and notifies the registrant by email
- daily task sends a one-shot reminder `PENDING_PUBLISHER_REMINDER_DAYS` (5 days) before expiry, gated by an `expiration_reminded` flag

The cutoff is computed by a single helper, `pending_publisher_cutoff(days)`, returning a naive UTC datetime to match the naive `created` column for in-Python comparisons.

### DIFF
--- a/warehouse/migrations/versions/e8a83da04d40_add_expiry_columns_to_pending_oidc_publishers.py
+++ b/warehouse/migrations/versions/e8a83da04d40_add_expiry_columns_to_pending_oidc_publishers.py
@@ -1,0 +1,41 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Add expiry columns to pending_oidc_publishers
+
+Revision ID: e8a83da04d40
+Revises: b8e6e0867168
+Create Date: 2026-04-09 00:00:00.000000
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision = "e8a83da04d40"
+down_revision = "b8e6e0867168"
+
+
+def upgrade():
+    op.add_column(
+        "pending_oidc_publishers",
+        sa.Column(
+            "created",
+            sa.DateTime(),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+    )
+    op.add_column(
+        "pending_oidc_publishers",
+        sa.Column(
+            "expiration_reminded",
+            sa.Boolean(),
+            server_default=sa.false(),
+            nullable=False,
+        ),
+    )
+
+
+def downgrade():
+    op.drop_column("pending_oidc_publishers", "expiration_reminded")
+    op.drop_column("pending_oidc_publishers", "created")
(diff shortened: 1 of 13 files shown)


miketheman:
Note for reviewers: The migration sets the default date on any existing pending publishers as "now" and will notify them en-masse in 25 days following this merge.

warehouse/migrations/versions/e8a83da04d40_add_expiry_columns_to_pending_oidc_publishers.py
```diff
@@ -0,0 +1,41 @@
...
+            sa.DateTime(),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+    )
+    op.add_column(
+        "pending_oidc_publishers",
+        sa.Column(
+            "expiration_reminded",
+            sa.Boolean(),
```
JacobCoffee, line 32: would it be better to be a timestamp

warehouse/oidc/tasks.py
```diff
@@ -1,13 +1,41 @@
...
 from warehouse.metrics import IMetricsService
-from warehouse.oidc.models import OIDCPublisher
+from warehouse.oidc.models import OIDCPublisher, PendingOIDCPublisher
 from warehouse.packaging.models import File, Project, Release
 
+if typing.TYPE_CHECKING:
+    from pyramid.request import Request
+
+# Pending publishers expire after this many days.
+PENDING_PUBLISHER_EXPIRY_DAYS = 30
```
JacobCoffee, line 26: would we ever expose things like this in admin so we can live-update

JacobCoffee (approved):

warehouse/oidc/tasks.py
miketheman: Probably not for a while - there's been a few ideas over the years for more dynamic feature flags thank what we have - see https://github.com/pypi/warehouse/issues/5869 for background. It's been unclear exactly what major benefit these would have over existing norms - we could easily externalize this to a config value and change it there, which only needs a new release/deploy, and it's clear what state the value was in when, and who changed it, etc. More dynamic configs and feature flags make that harder to reason about, so they have to be worth it.

warehouse/migrations/versions/e8a83da04d40_add_expiry_columns_to_pending_oidc_publishers.py
miketheman: Dubious value, it's going to be deleted in 5 days. What would you use that for? We can always change it later if it makes more sense.

PR #19997, opened 2026-05-02
feat(domain-status): add Fastly Domain Research API support

case-fastly:
As discussed in Slack with @miketheman, the legacy Domainr API will be deprecated later this year, and the Fastly Domain Research API is its replacement.

This PR adds support for the new API alongside the old one, for testing and confirmation purposes.


warehouse/accounts/services.py
```diff
@@ -1239,3 +1239,46 @@ def get_domain_status(self, domain: str) -> list[str] | None:
...
+
+    def get_domain_status(self, domain: str) -> list[str] | None:
+        """
+        Check if a domain is available or not.
+        See https://www.fastly.com/documentation/reference/api/domain-management/domain-research/
+        """
+
+        # bail early if no api key is set, so we don't send failing requests
+        if not self.api_key:
+            return None
```
miketheman: Instead of guarding, why not set the api_key to be required on instantiation, and then we have confidence it it already there? Could also drop `test_no_api_key_returns_none` as a result

miketheman (commented):
Looks directionally good - one question inline

miketheman (approved):
Nice work! I'll go ahead and ship this, and we can control the rollout by setting the different envvar for each environment as we wish. No timeline on that - likely after Labor Day in the USA

case-fastly:
> Nice work! I'll go ahead and ship this, and we can control the rollout by setting the different envvar for each environment as we wish. No timeline on that - likely after Labor Day in the USA

Yep sounds good, this way you can disable the legacy functionality when you're ready.

Keep us posted!

PR #19998, opened 2026-05-02
fix: `subject_type` for `"foo:bar"` event tags

divbzero:
The `EventTagEnum` docstring indicates that omitted `subject_type` should fallback to `source_type`:

https://github.com/pypi/warehouse/blob/53babdb2bc7a52445fb60c562e8d5a8f49581752/warehouse/events/tags.py#L26

However, a typo resulted in `EventTagEnum("foo:bar")` to have `"f"` instead of `"foo"` for `subject_type`.

### DIFF
--- a/warehouse/events/tags.py
+++ b/warehouse/events/tags.py
@@ -38,11 +38,11 @@ class EventTagEnum(enum.StrEnum):
 
     # Name = "source_type:subject_type:action"
     def __new__(cls, value: str):
-        values = value.split(":")
+        values: list[str] = value.split(":")
         obj = str.__new__(cls, value)
         obj._value_ = value
         obj.source_type = values[0]
-        obj.subject_type = ":".join(values[1:-1]) or value[0]
+        obj.subject_type = ":".join(values[1:-1]) or values[0]
         obj.action = values[-1]
         return obj
 


miketheman (commented):
How does this manifest? Feels odd to see a fix like this with no failing or updated tests?

divbzero:
@miketheman I don’t think it actually manifests. `subject_type` is defined for `EventTagEnum` but isn’t used anywhere in the code base.

I contributed to pypi/warehouse awhile back and was reading the code again as a reference for a different project I’m working on. Just happened to spot this typo.

Does the fix look sensible as is? Would we want to add tests, or perhaps remove `source_type`, `subject_type`, and `action` from `__new__` if they aren’t used anywhere?

miketheman:
> Does the fix look sensible as is? Would we want to add tests, or perhaps remove `source_type`, `subject_type`, and `action` from `__new__` if they aren’t used anywhere?

I'm thinking that if we don't actually make the distinction anywhere, and use it as the raw string, the parsing/assignment could be removed? But on the off chance we _do_ use it in some weird way, it's probably safer to keep it, so adding tests is a better approach - maybe the tests would show the intended usage.

PR #19999, opened 2026-05-02
add make patch to run all the things

dstufft:
Whenever you're making a PR there are a few things that you're supposed to run like linting, formatting, and translations. Currently these are treated as distinct steps (which is useful when you only want one of them) but you're really supposed to run all 3 before you make a PR.

However, it's easy to forget to run one of them (often times translations!) and the order that you run them in can matter (if you run linting first, you may get errors that reformat would fix, if you don't run translations last you may need to re-run it if reformating changed the line numbers around).

This wraps all of that up in a single command, `make patch` that you can run and it'll do the formatting, linting, and translations all in one shot. It re-uses the existing `bin/whatever` commands so we're not re-defining what it means to format/lint/translate/etc, just giving an easy way to run them all in the ideal order.


miketheman (commented):
This feels like it's duplicating 2/3 of the current pre-commit configuration https://github.com/pypi/warehouse/blob/994ebdd30a581e595f93d3f0ef9ad39ab11f5cf0/.pre-commit-config.yaml

Why not consolidate on that approach, and suggest users use either prek or pre-commit instead?

dstufft:
Users would need to install `prek` then, or we'd need to add `prek` to the development dependencies and then users would still need to do `make patch` or something and it didn't seem worthwhile to worry about it versus a 3 command shell script that will rarely ever change.

dstufft:
FWIW I've had this as a local patch for a few years now and I don't think it's ever changed since I first wrote it 😅

miketheman (approved):
Let's ship this so folks can use it if they want to, and not have to hold a local patch

docs/dev/development/getting-started.md
```diff
@@ -745,6 +745,13 @@ This is useful in scenarios like passing a
 TESTARGS="--randomly-seed=1234" make tests
 ```
 
+You can run the common tools (formaters, linting, translations) to make sure
```
miketheman: Yay, spelling linter caught this!
```suggestion
You can run the common tools (formatters, linting, translations) to make sure
```

miketheman (approved):

PR #20000, opened 2026-05-02
don't execute warehouse just to compile pycs

dstufft:
Currently the `Dockerfile` runs `python -m warehouse db -h` to trigger compilation of the `.pyc` files so that containers can boot up faster, however that adds ~14s to the container build process on my desktop, the bulk of which has nothing to do with compiling pyc files and mostly has to do with the fact that we're executing warehouse's code (and all of its imports), which has the *side effect* of compiling the pyc files.

Rather than doing that, we can just run `python -m compileall` (with the `-j 0` flag to parallelize it) to compile all of the files in `/opt/warehouse/src/`, which will cover warehouse and the tests (but not warehouse's dependencies).

pip compiles the `*.pyc` files for everything it installs, which we went out of our way to delete, so to get the `*.pyc` files for our dependencies, we just stop deleting those `*.pyc` files that pip already created for us.

The `python -m compileall` command takes about 2s on my desktop, the difference in deleting or not deleting the `*.pyc` files appears to basically be just noise.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,8 +96,7 @@ RUN --mount=type=cache,target=/root/.cache/pip \
             -r /tmp/requirements/docs-dev.txt \
             -r /tmp/requirements/docs-user.txt \
             -r /tmp/requirements/docs-blog.txt \
-    && pip check \
-    && find /opt/warehouse -name '*.pyc' -delete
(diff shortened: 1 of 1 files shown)


dstufft:

PR #20001, opened 2026-05-02
use less layers in the docker image

dstufft:
This doesn't really meaningfully change the time it takes to build, but it does make the resulting image smaller due to less layers used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,17 +68,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # We create an /opt directory with a virtual environment in it to store our
-# application in.
-RUN set -x \
-    && python3 -m venv /opt/warehouse
+# application in, we'll use --upgrade-deps to make sure we have the latest
+# version of pip.
+RUN --mount=type=cache,target=/root/.cache/pip \
+    set -x \
+        && python3 -m venv --upgrade-deps /opt/warehouse
 
 # Now that we've created our virtual environment, we'll go ahead and update
 # our $PATH to refer to it first.
 ENV PATH="/opt/warehouse/bin:${PATH}"
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -68,18 +68,16 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # We create an /opt directory with a virtual environment in it to store our
-# application in.
-RUN set -x \
-    && python3 -m venv /opt/warehouse
+# application in, we'll use --upgrade-deps to make sure we have the latest
+# version of pip.
+RUN --mount=type=cache,target=/root/.cache/pip \
```
dstufft, line 77: `venv` can upgrade pip for us now, so we'll let it do that (and also give it the same cache we use elsewhere).

dstufft: Forgot to mention, this makes our image smaller because we don't have layer 1 holding version X of pip, and then layer 2 deleting that and replacing it with version Y.

Dockerfile
```diff
@@ -169,7 +155,8 @@ RUN --mount=type=cache,target=/root/.cache/pip \
             install --no-deps --only-binary :all: \
                     -r /tmp/requirements/deploy.txt \
                     -r /tmp/requirements/main.txt \
-                    $(if [ "$DEVEL" = "yes" ]; then echo '-r /tmp/requirements/tests.txt -r /tmp/requirements/lint.txt'; fi) \
+                    $(if [ "$DEVEL" = "yes" ]; then echo '-r /tmp/requirements/dev.txt -r /tmp/requirements/tests.txt -r /tmp/requirements/lint.txt'; fi) \
```
dstufft: this doesn't really do much, but it means all of the deps get installed as a single layer, we weren't benefiting from multiple layers here (because the `COPY requirements` is when these would get invalidated anyways) and this puts it all in a single layer.

dstufft:

PR #20002, opened 2026-05-03
actually cache things with apt

dstufft:
We went through all this trouble to have persistent cache directories for apt, and then we ran `apt-cache clean` and `rm -rf` to go ahead and delete all of the apt caches anyways so we weren't actually caching anything.

Saves about 5s per apt invocation on my laptop (though we rarely have these layers un-cached, so it rarely occurs in local dev).

I verified that the layers looked the same before and after using `dive`.


dstufft:

PR #20003, opened 2026-05-03
set a TLDEXTRACT_CACHE environment variable

dstufft:
We run `tldextract --update` to update the cache of the public suffix list, and then we disable tldextract's built in mechanisms to update that cache at runtime.

Unfortunately when we run `tldextract --update` it runs as root, which means the cache is being saved in `$HOME/.cache`, which means it's at `/root/.cache`. When we run the image we're running as the `nobody` user, which has a `$HOME` set to `/nonexistant`, so this will be looking for a cache at `/nonexistant/.cache`.

This still works because tldextract bundles a copy of the public suffix list and will fall back to using it.

tldextract supports setting an environment variable that points it to look in a specific location for the cache, so we set that env var to have it stored at `/var/cache/tldextract` (it has 644 permissions, so anyone can read it), and that env var is part of the docker image, so it should carry over to runtime as well.


miketheman (commented):
In production, the `tldextract` cache is written to a /tmp/ directory, like this:
```
/tmp/cache/python-tldextract/3.13.13.final__warehouse__6ade0b__tldextract-5.3.1/
```

I suspect you're seeing a dev-only issue, since we run as `root` in dev. I tried nobody in the past, but it got reverted here: https://github.com/pypi/warehouse/pull/15444 (I was trying to speed up the postgres-stage of test cycle - to be able to clone the migrated database instance so each worker doesn't have to repeat that work.

dstufft:
Where did you get that path at? Are we able to grab the production docker image?

miketheman:
> Where did you get that path at? Are we able to grab the production docker image?

shell on prod:

```shellsession
nobody@pypi-warehouse-worker-5d45f855b5-rbntr:/opt/warehouse/src$ python -c "import tldextract; print(tldextract.cache.get_cache_dir())"
/tmp/cache/python-tldextract/3.13.13.final__warehouse__6ade0b__tldextract-5.3.1
```

dstufft:
I don't believe I remember how to get that shell 😅 , is there actually cached files in there? That would be where `tldextract` is looking for the cache when it's running as the `nobody` user, but if we build with `root` then I think the `tldextract --update` is still writing to `/root/.cache`?

dstufft:
Nevermind, found my notes!

dstufft:
No I'm right I think, I just got the path it was going to use wrong, the `tldextract --update` in our dockerfile is doing nothing currently.

The image build runs as root (`ls -la /opt/warehouse/src/` is all owned by root and the docker file doesn't do any `chmod`, so it's getting that ownership "naturally" by nature of running as root. Which means that when the image build runs `tldextract --update` it's writing the cache file to `/root/.cache/python-tldextract`.

At runtime tldextract can't see `/root/.cache` since it runs as the `nobody` user, so it's using `/tmp/cache` instead, which is *not* where there `tldextract --update` in the image build wrote the file out.

When the container starts, `/tmp/cache/python-tldextract/3.13.13.final__warehouse__6ade0b__tldextract-5.3.1` is going to be empty, and we'll never fetch an updated PSL at runtime because we pass `suffix_list_urls=()` into the `TLDExtract`.

However, `/tmp/cache` still gets populated, but it gets populated with the PSL that is bundled inside of tldextract, not the content that was fetched at image build time with `tldextract --update`.

tldextract calls [get_suffix_lists()](https://github.com/john-kurkowski/tldextract/blob/f06eec7f1c86322554bcb4031b72a74699f04cf6/tldextract/suffix_list.py#L73-L92) to get the suffix list and handle caching, updating, etc, which has no public urls passed into it due to `suffix_list_urls=()`. That ends up calling (and caching the reuslts of) [_get_suffix_list()](https://github.com/john-kurkowski/tldextract/blob/f06eec7f1c86322554bcb4031b72a74699f04cf6/tldextract/suffix_list.py#L95-L118), which tries to call [find_first_response()](https://github.com/john-kurkowski/tldextract/blob/f06eec7f1c86322554bcb4031b72a74699f04cf6/tldextract/suffix_list.py#L28-L59), which will raise `SuffixListNotFound` because there are no urls configured. That will cause [_get_suffix_list() to fall back to the snapshot](https://github.com/john-kurkowski/tldextract/blob/f06eec7f1c86322554bcb4031b72a74699f04cf6/tldextract/suffix_list.py#L107-L114) stored inside the package itself, which will then be cached in `/tmp/cache/python-tldextract/`.

So the `tldextract --update` is still doing nothing, but we might have to point it to a writeable directory (I'll test this) if we want to actually have the PSL updated at build time.

Otherwise if we're happy to continue to use the snapshot that is embedded inside of `tldextract`, we should just remove the `tldextract --update`.

miketheman (approved):
Thanks for the thorough investigation - we definitely want the updated list, so this should fix us not getting the updates.

PR #20005, opened 2026-05-03
make docker builds faster, particularly on version bumps

dstufft:
I've done everything I can think of to speed up docker builds here, particularly in the case where a developer already has images built but something has changed invalidating part of the cache.

My primary test case was trying to speed up:

```
$ docker system prune -a --volumes            # reset the system to a blank state
$ docker build . -t wh --build-arg DEVEL=yes  # build everything, warming the cache
$ update-dev-txt                              # update build to 1.4.4 in requirements/dev.txt
$ docker build . -t wh --build-arg DEVEL=yes  # build everything, see how long it takes
```

When I do that with the `main` branch, the build take 107.9s seconds, and with my PR it takes 67.7s, for a savings of roughly 40s.

<details>
  <summary>main branch</summary>
  
  ```
 [+] Building 107.9s (32/32) FINISHED                                                                                                                                                docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 9.76kB                                                                                                                                                        0.0s
 => [internal] load metadata for docker.io/library/python:3.13.13-slim-bookworm                                                                                                               0.2s
 => [internal] load metadata for docker.io/library/node:25.8.1-bookworm                                                                                                                       0.2s
 => [internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 184B                                                                                                                                                             0.0s
 => [build 1/7] FROM docker.io/library/python:3.13.13-slim-bookworm@sha256:eabbb62836ee44c18d350821e9f78488bcf65134bf763ae9989d63e611fa04d9                                                   0.0s
 => [internal] load build context                                                                                                                                                             0.3s
 => => transferring context: 210.51kB                                                                                                                                                         0.3s
 => [static-deps 1/4] FROM docker.io/library/node:25.8.1-bookworm@sha256:f0dfc4847f46d231e44948ded37521492439fd4ac19a6ef2597b9c79fe7a03a0                                                     0.0s
 => CACHED [build 2/7] RUN set -x     && python3 -m venv /opt/warehouse                                                                                                                       0.0s
 => CACHED [build 3/7] RUN pip --no-cache-dir --disable-pip-version-check install --upgrade pip                                                                                               0.0s
 => [build 4/7] COPY requirements /tmp/requirements                                                                                                                                           0.1s
 => [build 5/7] RUN --mount=type=cache,target=/root/.cache/pip     set -x     && if [ "yes" = "yes" ]; then pip --disable-pip-version-check install -r /tmp/requirements/dev.txt; fi          8.3s
 => [build 6/7] RUN --mount=type=cache,target=/root/.cache/pip     set -x     && if [ "yes" = "yes" ] && [ "no" = "yes" ]; then pip --disable-pip-version-check install -r /tmp/requirements  0.4s
 => [build 7/7] RUN --mount=type=cache,target=/root/.cache/pip     set -x     && pip --disable-pip-version-check             install --no-deps --only-binary :all:                     -r /  59.7s
 => CACHED [stage-4  2/10] WORKDIR /opt/warehouse/src/                                                                                                                                        0.0s
 => CACHED [stage-4  3/10] RUN set -eux;     rm -f /etc/apt/apt.conf.d/docker-clean;     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache;          0.0s
 => CACHED [stage-4  4/10] RUN --mount=type=cache,target=/var/cache/apt,sharing=locked     --mount=type=cache,target=/var/lib/apt,sharing=locked     set -x     && if [ "yes" = "yes" ]; the  0.0s
 => CACHED [static-deps 2/4] WORKDIR /opt/warehouse/src/                                                                                                                                      0.0s
 => CACHED [static-deps 3/4] COPY package.json package-lock.json babel.config.js /opt/warehouse/src/                                                                                          0.0s
 => CACHED [static-deps 4/4] RUN --mount=type=cache,target=/root/.npm,sharing=locked     npm ci                                                                                               0.0s
 => CACHED [static 1/6] COPY warehouse/static/ /opt/warehouse/src/warehouse/static/                                                                                                           0.0s
 => CACHED [static 2/6] COPY warehouse/admin/static/ /opt/warehouse/src/warehouse/admin/static/                                                                                               0.0s
 => CACHED [static 3/6] COPY warehouse/locale/ /opt/warehouse/src/warehouse/locale/                                                                                                           0.0s
 => CACHED [static 4/6] COPY webpack.config.js /opt/warehouse/src/                                                                                                                            0.0s
 => CACHED [static 5/6] COPY webpack.plugin.localize.js /opt/warehouse/src/                                                                                                                   0.0s
 => CACHED [static 6/6] RUN NODE_ENV=production npm run build                                                                                                                                 0.0s
 => CACHED [stage-4  5/10] COPY --from=static /opt/warehouse/src/warehouse/static/dist/ /opt/warehouse/src/warehouse/static/dist/                                                             0.0s
 => CACHED [stage-4  6/10] COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehouse/src/warehouse/admin/static/dist/                                                 0.0s
 => [stage-4  7/10] COPY --from=build /opt/warehouse/ /opt/warehouse/                                                                                                                        10.8s
 => [stage-4  8/10] COPY . /opt/warehouse/src/                                                                                                                                                1.5s
 => [stage-4  9/10] RUN tldextract --update                                                                                                                                                   2.4s
 => [stage-4 10/10] RUN python -m warehouse db -h                                                                                                                                            15.5s
 => exporting to image                                                                                                                                                                        5.5s
 => => exporting layers                                                                                                                                                                       5.5s
 => => writing image sha256:2010254e300204427161c4dc311c2f4ad9ae4887d0c410c6ea880feb386870d4                                                                                                  0.0s
 => => naming to docker.io/library/wh           
```
</details>

<details>
  <summary>with pr</summary>
  
  ```
[+] Building 67.7s (33/33) FINISHED                                                                                                                                                 docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 8.69kB                                                                                                                                                        0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                                                     0.2s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:2780b5c3bab67f1f76c781860de469442999ed1a0d7992a5efdf2cffc0e3d769                                                               0.0s
 => [internal] load metadata for docker.io/library/node:25.8.1-bookworm                                                                                                                       0.1s
 => [internal] load metadata for docker.io/library/python:3.13.13-slim-bookworm                                                                                                               0.1s
 => [internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 1.01kB                                                                                                                                                           0.0s
 => [internal] load build context                                                                                                                                                             0.2s
 => => transferring context: 149.84kB                                                                                                                                                         0.2s
 => [static-deps 1/4] FROM docker.io/library/node:25.8.1-bookworm@sha256:f0dfc4847f46d231e44948ded37521492439fd4ac19a6ef2597b9c79fe7a03a0                                                     0.0s
 => [base 1/5] FROM docker.io/library/python:3.13.13-slim-bookworm@sha256:eabbb62836ee44c18d350821e9f78488bcf65134bf763ae9989d63e611fa04d9                                                    0.0s
 => CACHED [base 2/5] COPY bin/docker/* /usr/local/bin/                                                                                                                                       0.0s
 => CACHED [base 3/5] RUN set -eux;     rm -f /etc/apt/apt.conf.d/docker-clean     && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache              0.0s
 => CACHED [base 4/5] RUN python -m compileall /usr/local/lib -j 0                                                                                                                            0.0s
 => CACHED [base 5/5] WORKDIR /opt/warehouse/src/                                                                                                                                             0.0s
 => CACHED [build 1/2] RUN --mount=type=cache,id=pkg,target=/root/.cache         create-venv /opt/warehouse                                                                                   0.0s
 => [build 2/2] RUN --mount=type=cache,id=pkg,target=/root/.cache     --mount=type=bind,src=requirements/,dst=/opt/warehouse/src/requirements/     pip-install         -r requirements/depl  37.5s
 => CACHED [stage-5 1/8] RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt,sharing=locked     --mount=type=cache,id=apt-lib,target=/var/lib/apt,sharing=locked     set -x     && if   0.0s
 => CACHED [static-deps 2/4] WORKDIR /opt/warehouse/src/                                                                                                                                      0.0s
 => CACHED [static-deps 3/4] COPY package.json package-lock.json babel.config.js /opt/warehouse/src/                                                                                          0.0s
 => CACHED [static-deps 4/4] RUN --mount=type=cache,target=/root/.npm,sharing=locked     npm ci                                                                                               0.0s
 => CACHED [static 1/6] COPY warehouse/static/ /opt/warehouse/src/warehouse/static/                                                                                                           0.0s
 => CACHED [static 2/6] COPY warehouse/admin/static/ /opt/warehouse/src/warehouse/admin/static/                                                                                               0.0s
 => CACHED [static 3/6] COPY warehouse/locale/ /opt/warehouse/src/warehouse/locale/                                                                                                           0.0s
 => CACHED [static 4/6] COPY webpack.config.js /opt/warehouse/src/                                                                                                                            0.0s
 => CACHED [static 5/6] COPY webpack.plugin.localize.js /opt/warehouse/src/                                                                                                                   0.0s
 => CACHED [static 6/6] RUN NODE_ENV=production npm run build                                                                                                                                 0.0s
 => [stage-5 2/8] COPY --from=build /opt/warehouse/ /opt/warehouse/                                                                                                                           9.4s
 => [stage-5 3/8] RUN python -m compileall /opt/warehouse/lib/ -j 0                                                                                                                           6.9s
 => [stage-5 4/8] COPY --from=static /opt/warehouse/src/warehouse/static/dist/ /opt/warehouse/src/warehouse/static/dist/                                                                      0.3s
 => [stage-5 5/8] COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehouse/src/warehouse/admin/static/dist/                                                          0.1s
 => [stage-5 6/8] COPY --exclude=requirements      --exclude=bin      --exclude=docs      --exclude=babel.config.js      --exclude=eslint.config.mjs      --exclude=package-lock.json      -  0.8s
 => [stage-5 7/8] RUN python -m compileall warehouse/ -j 0                                                                                                                                    1.1s
 => [stage-5 8/8] RUN tldextract --update                                                                                                                                                     2.9s
 => exporting to image                                                                                                                                                                        5.0s
 => => exporting layers                                                                                                                                                                       5.0s
 => => writing image sha256:081e1cadc0882b3ed465a42c55faa1731a309086e22695d3cb117ac5acc92464                                                                                                  0.0s
 => => naming to docker.io/library/wh  
  ```
</details>

I'm not 100% sure exactly which changes bring in the most win here, and some of them depend on each other, but a summary of the changes here:

- Fix `apt` caching, because we were mounting cache volumes for apt, but executing `apt-cache clean` and `rm -rf` to blow away everything we would have cached anyways.
    - This doesn't impact my test case above because that's almost always cached anyways, but still beneficial when it's not.
- Reorganize the dockerfile steps so that the order is more optimal:
    - ARGs cause cache invalidation when they change, so put things that aren't depending on their value before them.
    - Also remove some ARGs from stages that weren't using them to avoid more cache invalidation.
    - The `COPY` from the build stage to the runtime stage takes ~10s, whereas the `COPY` from the static stage takes 0.2s each, so put the slower `COPY` first so we don't re-copy if the static files change but not the dependencies.
- Compress things into single layers when they all get invalidated together anyways so they'll never be independently cache-able.
- Don't have pip compile the pyc files for what it installs, it's faster to do it explicitly using `python -m compileall`.
    - We do this 3 times, once for the stdlib (in our "base" image so that is shared), once for the deps (right after we copy them over from the build stage), and then once for warehouse itself.
- Use `python -m compileall` to compile the pycs for warehouse instead of `python -m warehouse db -h`, the latter has to execute warehouse, which just happens to have compiling pycs as a side effect, and takes ~15s while the former only has to compile the pycs and can do it in parallel.
- Avoid adding layers for `COPY` the `requirements/*.txt` files, and instead use bind mounts during build-- caching will still work correctly with changes, but we avoid a (small) pointless layer.
- Avoid layer churn by having the call to create the virtual environment upgrade to the latest pip instead of doing it as a second step, which causes the "venv creation" layer to have pip vX, and then the following layer to have tombstone markers for pip vX and also add pipvY's files + pay the cost for installing pip twice (and uninstalling it once).
- Aggressively filter the build context (if it's not in the build context, it can't possibly cause a spurious cache invalidation).
- Aggressively filter what we install from the build context into the final image to only what is actually needed at runtime (again, if we're not `COPY` a file into the container, then it can't cause a cache invalidation)
   - This means that we're pretty much only installing warehouse itself, no requirements files, no test suite, no `.git` repo, no `bin/` dir, nothing that isn't needed at runtime. We re-add those things in the `docker-compose` file to support development (which matches what happens in CI as well).
   - Note: The `sitecustomize.py` that we use to install coverage was being ran in production afaict, though presumably doing nothing because we didn't have coverage installed, but this fixes that as well.
- Add helpers for `apt-install`, `pip-install`, and `create-venv` to remove visual noise in the `Dockerfile` and make it easier to make sure all the right flags are being passed everywhere.

The remaining time taken is basically:

- 38s for `pip install`
- 10s to `COPY` the virtual environment from the build stage to the runtime stage.
- 7s to compile the `*.pyc` files for the dependencies in the virtual environment.
- 2s to compile the `*.pyc` files for warehouse itself.
- 3s to pre-cache the public suffix list (`tldextract --update`).
- 5s to export the layers to an image.

We might be able to drop the 10s to `COPY` the virtual environment if we have the runtime inherit from the build stage instead of using the copy, but we'd want to move the `apt-get install` up above the dependency stuff-- that would make a cold build take longer [^1] because it wouldn't be able to parallelize the `apt-get install` and the `pip install` but it would make a warm build *anytime the `apt-get install` layer is cached) 10s faster.

I didn't make that change because it's not a clear performance win in every case and if we ever need build only dependencies we'll have to restructure the docker file and bring that back-- but if we were OK with that, it would bring our build down to 53.2s.

<details>
  <summary>with pr and no build stage</summary>
  
  ```
[+] Building 53.2s (32/32) FINISHED                                                                                                                                                 docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 8.75kB                                                                                                                                                        0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                                                     0.3s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:2780b5c3bab67f1f76c781860de469442999ed1a0d7992a5efdf2cffc0e3d769                                                               0.0s
 => [internal] load metadata for docker.io/library/node:25.8.1-bookworm                                                                                                                       0.1s
 => [internal] load metadata for docker.io/library/python:3.13.13-slim-bookworm                                                                                                               0.1s
 => [internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 1.01kB                                                                                                                                                           0.0s
 => [internal] load build context                                                                                                                                                             0.2s
 => => transferring context: 149.84kB                                                                                                                                                         0.2s
 => [static-deps 1/4] FROM docker.io/library/node:25.8.1-bookworm@sha256:f0dfc4847f46d231e44948ded37521492439fd4ac19a6ef2597b9c79fe7a03a0                                                     0.0s
 => [base 1/5] FROM docker.io/library/python:3.13.13-slim-bookworm@sha256:eabbb62836ee44c18d350821e9f78488bcf65134bf763ae9989d63e611fa04d9                                                    0.0s
 => CACHED [base 2/5] COPY bin/docker/* /usr/local/bin/                                                                                                                                       0.0s
 => CACHED [base 3/5] RUN set -eux;     rm -f /etc/apt/apt.conf.d/docker-clean     && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache              0.0s
 => CACHED [base 4/5] RUN python -m compileall /usr/local/lib -j 0                                                                                                                            0.0s
 => CACHED [base 5/5] WORKDIR /opt/warehouse/src/                                                                                                                                             0.0s
 => CACHED [build 1/4] RUN --mount=type=cache,id=pkg,target=/root/.cache         create-venv /opt/warehouse                                                                                   0.0s
 => CACHED [build 2/4] RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt,sharing=locked     --mount=type=cache,id=apt-lib,target=/var/lib/apt,sharing=locked     set -x     && if [   0.0s
 => CACHED [static-deps 2/4] WORKDIR /opt/warehouse/src/                                                                                                                                      0.0s
 => CACHED [static-deps 3/4] COPY package.json package-lock.json babel.config.js /opt/warehouse/src/                                                                                          0.0s
 => CACHED [static-deps 4/4] RUN --mount=type=cache,target=/root/.npm,sharing=locked     npm ci                                                                                               0.0s
 => CACHED [static 1/6] COPY warehouse/static/ /opt/warehouse/src/warehouse/static/                                                                                                           0.0s
 => CACHED [static 2/6] COPY warehouse/admin/static/ /opt/warehouse/src/warehouse/admin/static/                                                                                               0.0s
 => CACHED [static 3/6] COPY warehouse/locale/ /opt/warehouse/src/warehouse/locale/                                                                                                           0.0s
 => CACHED [static 4/6] COPY webpack.config.js /opt/warehouse/src/                                                                                                                            0.0s
 => CACHED [static 5/6] COPY webpack.plugin.localize.js /opt/warehouse/src/                                                                                                                   0.0s
 => CACHED [static 6/6] RUN NODE_ENV=production npm run build                                                                                                                                 0.0s
 => [build 3/4] RUN --mount=type=cache,id=pkg,target=/root/.cache     --mount=type=bind,src=requirements/,dst=/opt/warehouse/src/requirements/     pip-install         -r requirements/depl  38.1s
 => [build 4/4] RUN python -m compileall /opt/warehouse/lib/ -j 0                                                                                                                             6.1s
 => [stage-5 1/5] COPY --from=static /opt/warehouse/src/warehouse/static/dist/ /opt/warehouse/src/warehouse/static/dist/                                                                      0.3s
 => [stage-5 2/5] COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehouse/src/warehouse/admin/static/dist/                                                          0.1s
 => [stage-5 3/5] COPY --exclude=requirements      --exclude=bin      --exclude=docs      --exclude=babel.config.js      --exclude=eslint.config.mjs      --exclude=package-lock.json      -  0.8s
 => [stage-5 4/5] RUN python -m compileall warehouse/ -j 0                                                                                                                                    1.1s
 => [stage-5 5/5] RUN tldextract --update                                                                                                                                                     1.0s
 => exporting to image                                                                                                                                                                        4.8s
 => => exporting layers                                                                                                                                                                       4.8s
 => => writing image sha256:a32c68ac89079f50873c9d4616bac09a2be9d06264d46e8c34e6825d39ded0e2                                                                                                  0.0s
 => => naming to docker.io/library/wh                                                                                                                                                         0.0s
  ```
</details>

I don't think there's anything left to do to make the python build faster without removing stuff (the pyc precompilation or the tldextract) or more invasive tooling changes.

[^1]: Technically it would only make it take longer if the static build was cached but the python build was not, because the fully cold build's slowest task is now the static build, which does happen in parallel and takes longer than doing the python build without parallelization does.

### CODE REVIEW

**Comment by dstufft** [comment]
I thought about it some more, and did the removal of the build stage since I think it's a win in every case except when you have the static stage cached but you *don't* have the `apt-get` layer cached *and* you're building with `DEVEL=yes`, but in every other case it's a ~10s win.

I also realized that compiling the pyc files for warehouse itself isn't useful in a `DEVEL=yes` build, because that's pretty much always going to have the checkout mounted over top of it anyways and so won't be used, and I also think a devel build doesn't need updated PSL cache.

So that brings it down to 57s

<details>
  <summary>updated pr</summary>
  
  ```
[+] Building 57.3s (32/32) FINISHED                                                                                                                                                 docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                          0.0s
 => => transferring dockerfile: 8.69kB                                                                                                                                                        0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                                                     0.3s
 => CACHED docker-image://docker.io/docker/dockerfile:1@sha256:2780b5c3bab67f1f76c781860de469442999ed1a0d7992a5efdf2cffc0e3d769                                                               0.0s
 => [internal] load metadata for docker.io/library/python:3.13.13-slim-bookworm                                                                                                               0.2s
 => [internal] load metadata for docker.io/library/node:25.8.1-bookworm                                                                                                                       0.2s
 => [internal] load .dockerignore                                                                                                                                                             0.0s
 => => transferring context: 1.01kB                                                                                                                                                           0.0s
 => [static-deps 1/4] FROM docker.io/library/node:25.8.1-bookworm@sha256:f0dfc4847f46d231e44948ded37521492439fd4ac19a6ef2597b9c79fe7a03a0                                                     0.0s
 => [internal] load build context                                                                                                                                                             0.2s
 => => transferring context: 150.72kB                                                                                                                                                         0.2s
 => [base 1/5] FROM docker.io/library/python:3.13.13-slim-bookworm@sha256:eabbb62836ee44c18d350821e9f78488bcf65134bf763ae9989d63e611fa04d9                                                    0.0s
 => CACHED [base 2/5] COPY bin/docker/* /usr/local/bin/                                                                                                                                       0.0s
 => CACHED [base 3/5] RUN set -eux;     rm -f /etc/apt/apt.conf.d/docker-clean     && echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache              0.0s
 => CACHED [base 4/5] RUN python -m compileall /usr/local/lib -j 0                                                                                                                            0.0s
 => CACHED [base 5/5] WORKDIR /opt/warehouse/src/                                                                                                                                             0.0s
 => CACHED [stage-4 1/9] RUN --mount=type=cache,id=pkg,target=/root/.cache         create-venv /opt/warehouse                                                                                 0.0s
 => CACHED [stage-4 2/9] RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt,sharing=locked     --mount=type=cache,id=apt-lib,target=/var/lib/apt,sharing=locked     set -x     && if   0.0s
 => [stage-4 3/9] RUN --mount=type=cache,id=pkg,target=/root/.cache     --mount=type=bind,src=requirements/,dst=/opt/warehouse/src/requirements/     pip-install         -r requirements/de  42.0s
 => CACHED [static-deps 2/4] WORKDIR /opt/warehouse/src/                                                                                                                                      0.0s
 => CACHED [static-deps 3/4] COPY package.json package-lock.json babel.config.js /opt/warehouse/src/                                                                                          0.0s
 => CACHED [static-deps 4/4] RUN --mount=type=cache,target=/root/.npm,sharing=locked     npm ci                                                                                               0.0s
 => CACHED [static 1/6] COPY warehouse/static/ /opt/warehouse/src/warehouse/static/                                                                                                           0.0s
 => CACHED [static 2/6] COPY warehouse/admin/static/ /opt/warehouse/src/warehouse/admin/static/                                                                                               0.0s
 => CACHED [static 3/6] COPY warehouse/locale/ /opt/warehouse/src/warehouse/locale/                                                                                                           0.0s
 => CACHED [static 4/6] COPY webpack.config.js /opt/warehouse/src/                                                                                                                            0.0s
 => CACHED [static 5/6] COPY webpack.plugin.localize.js /opt/warehouse/src/                                                                                                                   0.0s
 => CACHED [static 6/6] RUN NODE_ENV=production npm run build                                                                                                                                 0.0s
 => [stage-4 4/9] RUN python -m compileall /opt/warehouse/lib/ -j 0                                                                                                                           7.2s
 => [stage-4 5/9] COPY --from=static /opt/warehouse/src/warehouse/static/dist/ /opt/warehouse/src/warehouse/static/dist/                                                                      0.3s
 => [stage-4 6/9] COPY --from=static /opt/warehouse/src/warehouse/admin/static/dist/ /opt/warehouse/src/warehouse/admin/static/dist/                                                          0.1s
 => [stage-4 7/9] COPY --exclude=requirements      --exclude=bin      --exclude=docs      --exclude=babel.config.js      --exclude=eslint.config.mjs      --exclude=package-lock.json      -  0.8s
 => [stage-4 8/9] RUN if [ "yes" != "yes" ]; then python -m compileall warehouse/ -j 0; fi                                                                                                    0.3s
 => [stage-4 9/9] RUN if [ "yes" != "yes" ]; then tldextract --update; fi                                                                                                                     0.5s
 => exporting to image                                                                                                                                                                        5.1s
 => => exporting layers                                                                                                                                                                       5.1s
 => => writing image sha256:3164c8446ac61736b526afffa89a07b43f7e19f81662b59bd31f12d642cbfd4a                                                                                                  0.0s
 => => naming to docker.io/library/wh                                                                                                                                                         0.0s
```
</details>

.gitignore
```diff
@@ -6,6 +6,7 @@
 
 .cache/
 .coverage
+.coverage.*
```
miketheman: Can these be collapsed to a single entry?
```suggestion
.coverage*
```

Dockerfile
```diff
@@ -115,128 +113,97 @@ USER docs
 
 # Now we're going to build our actual application, but not the actual production
 # image that it gets deployed into.
```
miketheman: This comment is now incorrect, since there's no other image after this point that I can see.

Dockerfile
```diff
@@ -40,66 +40,64 @@ RUN NODE_ENV=production npm run build
...
 
 
-# We'll build a light-weight layer along the way with just docs stuff
-FROM python:${PYTHON_IMAGE_VERSION} AS docs
+# Create a base image that contains some helpers and settings for our python
+# stages to inherit from.
+FROM python:3.13.13-slim-bookworm AS base
+
+# Copy our helpers over into the base image
+COPY bin/docker/* /usr/local/bin/
```
miketheman, line 48: We don't appear need these in the production image, and they likely will fail due to running as `nobody` - any reason to keep them?

I'm also not totally sold on the script wrappers extraction - if only used 2x in the Dockerfile - if there were 3 or more, sure, but these don't feel like the indirection is justified yet

Dockerfile
```diff
@@ -115,128 +113,97 @@ USER docs
...
-ENV PYTHONUNBUFFERED 1
-ENV PYTHONPATH /opt/warehouse/src/
-ENV PATH="/opt/warehouse/bin:${PATH}"
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONPATH=/opt/warehouse/src/
 
-WORKDIR /opt/warehouse/src/
+# We create an /opt directory with a virtual environment in it to store our
+# application in, we'll use --upgrade-deps to make sure we have the latest
+# version of pip.
```
miketheman: The new comment about `--upgrade-deps` seems out of place on this line, there's no install step here?

miketheman (commented):
I went through for as long as I could spare, and surfaced some questions and changes.
I'd be interested in seeing a side-by-side diff of the resulting container image layouts, with ARG CI=yes, DEVEL=no - those kinds of flags to ensure that the resulting image is indeed the same everywhere we care about it to be.

Dockerfile
dstufft: Well it's not any different than keeping `pip` itself installed right? Executing `pip` will likely fail in the production image due to running as `nobody`.

There's little harm in keeping them in the production image because permissions are going to keep them from doing anything, and this layer collectively adds 329 bytes to the 610MB production image (and removing them from the image later would actually increase that size, unless bind mounts were used to bring them into the image), so it doesn't seem particularly worth it to me to worry about them being left behind.

As far as the script wrappers in general, I think they make the docker file much easier to read, it takes lines like

```
RUN --mount=type=cache,id=pkg,target=/root/.cache \
    --mount=type=bind,src=requirements/,dst=/opt/warehouse/src/requirements/ \
    set -x \
            install --no-deps --only-binary :all: --no-compile \
            -r /tmp/requirements/docs-dev.txt \
            -r /tmp/requirements/docs-user.txt \
            -r /tmp/requirements/docs-blog.txt \
    && pip check
```

and turns it into

```
RUN --mount=type=cache,id=pkg,target=/root/.cache \
    --mount=type=bind,src=requirements/,dst=/opt/warehouse/src/requirements/ \
    pip-install \
        -r requirements/docs-dev.txt \
        -r requirements/docs-user.txt \
        -r requirements/docs-blog.txt
```

Removing a lot of visual noise when reading the docker file as well as giving a single place to update these commands when needed (for instance, adding the `--no-compile` flag like this PR does).

.dockerignore
```diff
@@ -1,14 +1,60 @@
...
+.pre-commit-config.yaml
+.python-version
+.readthedocs.yml
+.stylelintrc.json
+.weblate.ini
+CONTRIBUTING.rst
+Dockerfile
+README.rst
+babel.cfg
+bin
```
miketheman, line 44: I think we need `bin/release` for release step?

Dockerfile
```diff
@@ -1,10 +1,10 @@
-# Set variables reused in Dockerfile
-ARG PYTHON_IMAGE_VERSION=3.14.5-slim-trixie
+# syntax=docker/dockerfile:1
```
miketheman, line 1: Are we certain cabotage's buildkit version knows how to handle this directive?

miketheman (commented):
Two more inlines

Dockerfile
dstufft: I'm pretty sure this directive has existed at least since the introduction of BuildKit, and I believe even earlier than that, we don't actually require this directive to make anything work, it's just best practice to specify it.

.dockerignore
dstufft: Yea, I missed that when I checked the `Procfile` to make sure we weren't using any of the `bin/worker` etc style scripts we used to use.

miketheman (approved):

PR #20016, opened 2026-05-04
add a test to ensure templates all exist

dstufft:
This drops the last remaining use of `flake8` in the code base, saving 6s in `make lint` as well as removing the `setup.cfg` file as everything left is configured in `pyproject.toml` or their own configuration file.

The template exists static lint worked by reading the AST looking for a `@view_config()` decorator with a renderer keyword and they attempted to figure out if that was a template path using heuristics against the string, and then if it was, tried to parse out the actual template name (because many of our templates use asset paths so they look like `warehouse:template/path.html`) and then checked to see if they existed in one of our two template locations.

This had a number of assumptions baked into it:

- All of our templates got added via `@view_config()`
   - We have some that get added via `config.add_template_view()`.
   - Wouldn't hold true if we have a wrapper around `@view_config()`, but I'm not aware of any uses of that.
- Assumes that all of our templates come from warehouse itself (currently true, not required to be true).
- Requires updating anytime we add another non-template renderer (doesn't happen very often).
- Assumes the templates in warehouse always come from one of two folders (currently true, likely will always be true, but not actually required to be true).

The test added here works by taking a configured instance of the application and asking Pyramid to give it a list of all of the templates, and then gets the actual Jinja2 Environment associated with that template and asks the Environment to locate the template for us. This matches what happens at runtime, works regardless of how the template is configured, where it comes from, etc.

It does make the assumption that all of the templates (not all of the renderers, that's a different question you can ask Pyramid) are Jinja2 templates which is currently true, but isn't required to be true, but will fail with a descriptive error if that ever changes.

All in all, the big upsides to this PR are:

- 6s less time in `make lint`
- checking for existing templates is more thorough.

The cost of this PR is:

- Non-existent templates can't be caught at lint time (which can show up in your editor as you're editing) but will be caught at test time instead.

I think new templates/views are added infrequently enough that making detecting errors in `@view_config()` get deferred to test time is worth saving 6s off `make lint` and the test is more thorough to boot.


dev/flake8/checkers.py
miketheman, line 1: `WH`, we hardly knew ye. Rest well

miketheman (approved):

PR #13164, opened 2023-03-08
Organizations: Rework invitation expiry handling

ewdurbin:
- adds events to account and org event logs
- displays expired invitations with "delete" or "resend" actions

~depends on #13149~
closes #12425

PR #20018, opened 2026-05-04
Fix: Use RFC-5735 TEST-NET addresses in unit tests

### LINKED ISSUE #13186: Update all IP Addresses in tests to use RFC-5735 TEST-NET-* addresses.

64johnlee:
## Summary

Replace non-RFC-5735 compliant private IP addresses with properly reserved TEST-NET addresses in unit tests:

-  →  (TEST-NET-1, 192.0.2.0/24) — in tests/unit/accounts/test_views.py (13 occurrences)
-  →  (TEST-NET-2, 198.51.100.0/24) — in tests/unit/cache/origin/test_fastly.py (24 occurrences)

Per [RFC-5735](https://www.rfc-editor.org/rfc/rfc5735), the TEST-NET blocks are:
- 192.0.2.0/24 (TEST-NET-1)
- 198.51.100.0/24 (TEST-NET-2)  
- 203.0.113.0/24 (TEST-NET-3)

## Motivation

As noted in [#13164 (comment)](https://github.com/pypi/warehouse/pull/13164#discussion_r1131763629), using RFC-5735 TEST-NET addresses in tests is preferred over private addressing space (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16).

Closes #13186

### DIFF
--- a/tests/unit/accounts/test_views.py
+++ b/tests/unit/accounts/test_views.py
@@ -2946,7 +2946,7 @@ class TestVerifyOrganizationRole:
         db_request.method = "POST"
         db_request.GET.update({"token": "RANDOM_KEY"})
         db_request.route_path = pretend.call_recorder(lambda *a, **kw: "/")
-        db_request.remote_addr = "192.168.1.1"
+        db_request.remote_addr = "192.0.2.1"
         db_request.session.flash = pretend.call_recorder(lambda *a, **kw: None)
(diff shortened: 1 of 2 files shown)


woodruffw:
I see basically no harm to doing this, although I don't think our current use of non-test addresses in the test suite is particularly harmful either.

@64johnlee Could you fix the lint failure?

PR #20020, opened 2026-05-05
Add GitHub Actions workflow for Python linting and testing

p5hzehxa:
This workflow installs Python dependencies, runs linting with flake8, and executes tests with pytest.


JacobCoffee:
We already do this
https://github.com/pypi/warehouse/blob/3b89d353d7d26b356ffb39fed7938c6d803e1065/.github/workflows/ci.yml#L46-L51

PR #20027, opened 2026-05-05
Fix: Use RFC-5735 TEST-NET addresses in unit tests

64johnlee:
## Description
Uses RFC-5735 TEST-NET IP addresses (192.0.2.0/24, 198.51.100.0/24, 203.0.113.0/24) instead of real/public IPs in unit tests.

## Changes
- Replaced hardcoded test IP addresses with RFC-5735 compliant TEST-NET addresses
- Updated affected test files

## Testing
- All unit tests pass with updated IP addresses

## Related Issue
Addresses review feedback from closed PR #20018


JacobCoffee:
image
Please do not reopen or I will have to block you from contributing to the organization.

PR #20034, opened 2026-05-06
chore(dependabot): test grouped updates

miketheman:
Try grouping "everything else" again.
Attempted two years ago in #15920 and reverted in #15932 after unsuccessful runs.

If this works consistently for a couple of days, remove the "Combine PRs" manual action, and remove any other pieces for it.

Refs: #15909

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,6 +26,10 @@ updates:
       # Keep both psycopg and psycopg-c together
       patterns:
         - "psycopg*"
+    python-packages:
+      # Group all remaining packages together
+      patterns:
+        - "*"
   cooldown:
     default-days: 7
 - package-ecosystem: github-actions


miketheman:
I tested the config locally with the `dependabot` CLI, and it took ~1 hour to run, but that's in serial, and also doesn't actually handle the resulting Pull Request generation - that's handled on GH side.

JacobCoffee (approved):
would we want to set it to non-major (patch, minor) so that majors get a special pr to ensure breakage isnt a big thing? just an idea, lgtm to test

miketheman:
> would we want to set it to non-major (patch, minor) so that majors get a special pr to ensure breakage isnt a big thing? just an idea, lgtm to test

I was reading the docs, and it looked like even on grouped updates we could issue an `ignore` command, so let's see what it does tomorrow.

PR #20035, opened 2026-05-06
fix: close verified-URL open-redirect chain

miketheman:
User-submitted URLs (project metadata, trusted publisher claims, route parameters) can be crafted so that our URL parser sees one URL and the browser visits another. Three places where that mismatch leaks:

1. _verify_url_pypi and verify_url_from_reference normalize with rfc3986, which leaves "\" as a path character. Browsers (per WHATWG) treat "\" as "/", so "/project/foo/..\x/../bar/" verifies as /project/foo/ even though the browser visits /project/bar/. Reject any URL containing a backslash.

2. redirect_view_factory drops percent-decoded matchdict values into the Location header. Encoding "\", "?" and friends as %5C, %3F then lets a permissive route like /p/{name}/{version}/ get bent to /account/logout/?next=... Refuse backslashes in the constructed target.

3. is_safe_url accepted "/\t/evil.com/" because urllib3.util.parse_url keeps the tab, but webob's _make_location_absolute later calls urllib.parse.urlsplit, which (Python 3.10+) strips tab/CR/LF before parsing. The URL collapses to "//evil.com/" and redirects off-host. Strip the same characters in is_safe_url so we judge the URL the user actually gets served.


ewdurbin (approved):

digitalresistor:
@ewdurbin @miketheman I don't bite...

https://github.com/Pylons/webob/security/advisories/GHSA-fh3h-vg37-cc95 -- https://pypi.org/project/WebOb/1.8.10/

Feel free to reach out to me in the future so we can try and coordinate (especially so it doesn't come while I am really sick and unable to do a quick release).

miketheman:
> I don't bite...

@digitalresistor I should hope not?

I don't think the fixes here were specific to webOb, were they? This was mostly about what user-provided URL values in metadata are used to represent "verification" behaviors for misdirection/reputation farming in a browser.

PR #20036, opened 2026-05-07
chore(ci): upgrade update-wow-action

miketheman:
It finally broke using older node20 runtime.
Hope to be able to replace with grouped updates after existing open ones are merged.


miketheman:
Closing, going to swap with a first-party action from GH instead.

PR #20037, opened 2026-05-07
chore(ci): swap to actions/create-github-app-token

miketheman:
wow-actions finally broke due to node20. Instead of upgrading, drop wow-actions/use-app-token in favor of GitHub's first-party action. One fewer third-party touching our App's private key, and the installation token gets revoked when the job ends instead of sitting valid for an hour.

Also dropped the GITHUB_TOKEN fallback. If the app token mint fails, GITHUB_TOKEN opens a combined PR that can't trigger downstream CI — which is the reason we mint an app token in the first place. Better to fail loud.

### DIFF
--- a/.github/workflows/combine-prs.yml
+++ b/.github/workflows/combine-prs.yml
@@ -25,17 +25,16 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - name: Use GitHub App Token
-        uses: wow-actions/use-app-token@9e8487c993ab4085b2dd8cb90ab446b6a18cf834 # v2.1.1
+      - name: Generate GitHub App Token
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-prs.yml
```diff
@@ -25,17 +25,16 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - name: Use GitHub App Token
-        uses: wow-actions/use-app-token@9e8487c993ab4085b2dd8cb90ab446b6a18cf834 # v2.1.1
+      - name: Generate GitHub App Token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
```
miketheman, line 29: This is also their first version that uses immutable releases!
https://github.com/actions/create-github-app-token/releases/tag/v3.1.1

Image

dstufft (approved):

PR #20052, opened 2026-05-08
fix: silence gunicorn startup noise

miketheman:
We don't need the new interface that showed up in 25.1.0, and it's noisy.

Fixes WAREHOUSE-PRODUCTION-2GK

### DIFF
--- a/gunicorn-prod.conf.py
+++ b/gunicorn-prod.conf.py
@@ -19,6 +19,13 @@ access_log_format = (
 
 statsd_host = "localhost:8125"
 
+# Disable remote control of the daemon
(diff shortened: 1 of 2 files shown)


JacobCoffee (approved):

JacobCoffee:
i dont see in their docs a way to change path, interesting

miketheman:
> i dont see in their docs a way to change path, interesting

Yeah, I had to dig into the code - apparently they auto-detect `XDG_RUNTIME_DIR` and use that, fall back to `$HOME` - and our `XDG_RUNTIME_DIR` is not unset, it resolves to an empty string. I don't know what that's about, and don't need to know right now 😄

PR #20065, opened 2026-05-12
Add `_uploaded_via` and `_trusted_publishing` to json api

thatch:
These are only accessible in the version HTML today, which requires a fastly client challenge to view.  These fields are useful for automation.

### DIFF
--- a/warehouse/packaging/utils.py
+++ b/warehouse/packaging/utils.py
@@ -120,6 +120,8 @@ def _simple_detail(project, request):
                     if file.provenance
                     else None
                 ),
+                "_uploaded_via": file.uploaded_via or None,
+                "_trusted_publishing": file.uploaded_via_trusted_publisher,
             }
             for file in files
         ],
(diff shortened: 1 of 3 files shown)


woodruffw:
Hi @thatch, can you say a bit about your use case for these? Historically PyPI has not exposed "bare" TP information because it would mislead people into treating it as a security signal/differentiator.

(For context, you can retrieve attestation information from the standard index APIs, and since attestations imply TP, you can use the attestation's principal to reconstruct the TP principal.)

warehouse/packaging/utils.py
```diff
@@ -116,8 +125,10 @@ def _simple_detail(project, request):
                     if file.provenance
                     else None
                 ),
+                "_uploaded_via": file.uploaded_via or None,
+                "_trusted_publishing": is_trusted_publisher,
```
woodruffw, line 129: Independent of other concerns, I don't think PyPI can/should modify the standard index APIs without a PEP. If accepted, I think this would fit better in the PyPI specific JSON API?

PR #20066, opened 2026-05-12
chore(deps): upgrade to Python 3.14.5

miketheman:
Let's try this again with the older garbage collector.


pyproject.toml
dstufft:
```suggestion
target-version = "py314"
```

Do we want to do this now? Or should we wait until we make sure 3.14 works without needing to be reverted?

miketheman: Seems safe to me since it changed nothing?

dstufft: Did you forget to push the change?

miketheman: No? Can you be clearer about what you're expecting to see and are not? I get no changes with `ruff format` or `ruff check` - maybe something is different on your local end?

dstufft: Oh, I mean mean updating the target version for ruff to 314. Even if there's no changes, it seems like we'd want to update the version there as well? Otherwise we'll be marking Python 3.14 everywhere _but_ ruff?

miketheman: Oh, I get it now. Updated.

dstufft (approved):

PR #20074, opened 2026-05-14
fix(oidc): roll back aborted transaction after pending-publisher conflict

### LINKED ISSUE #20006: 500 error when adding second Pending Trusted Publisher to a single account
## Summary

The "Add a new pending publisher" form at
https://pypi.org/manage/account/publishing/ returns a generic 500 error
page ("Sorry, something went wrong / PyPI is down for maintenance or is
having an outage") when submitting **the second** pending publisher
under a single account, regardless of project name.

The first submission for the same account succeeds normally.

## Account

PyPI username: `zx22413`

## Reproduction

1. Logged in fresh, 2FA enabled.
2. Submitted form with project name `forget-rag` and these values:
   - Owner: `zx22413`
   - Repository: `forget-rag`
   - Workflow: `publish.yml`
   - Environment: `pypi`
   → **Succeeded**, listed under Pending publishers.
3. Submitted form again with project name `mem-broom`, all other
   fields identical.
   → **500 page**.
4. Waited 24h, retried `mem-broom`. Same 500.
5. Tried different name `forget-rag-broom` (no naming overlap with
   any existing PyPI package). Same 500.
6. Tried unrelated random name `zxtest-pending-QQQ`. Same 500.

PyPI status page (https://status.python.org/) reports all services
operational throughout. `pypi.org/pypi/<name>/json` and the homepage
return 200 for me; only the pending-publisher submission fails.

## Expected

Either accept the second pending publisher, or return a specific
validation error explaining what's wrong (typosquat conflict, rate
limit, etc.) instead of a 500.

## Browser

Tried both regular and incognito mode in the same browser session.
Same result. (Network is wired desktop, not behind any unusual proxy.)

## Why this matters / impact

I'm shipping a small two-package release (`forget-rag` + `mem-broom`)
this week. The first package registered fine; the second can't, so I'm
deferring it to a git-install-only release until this is resolved.
Happy to provide additional logs or test more names if useful.

Fyko:
## What

Rolls back the aborted transaction in the `except UniqueViolation` handler for pending trusted publishers, in both the account- and org-scoped views.

## Why

Registering a second pending publisher that points at the same repo/owner/workflow/environment as an existing one (under a different project name) fails the table's unique constraint on INSERT. #19972 added a handler to surface that conflict but never rolled back the now-aborted transaction, so the next DB access (rendering the user's pending publishers, or the end-of-request commit) blew up with `PendingRollbackError` and the user got a generic 500/503.

Fixes #20006


woodruffw (approved):
The approach here makes sense to me. Needs rebase/deconflict but otherwise LGTM.

ewdurbin (approved):

Fyko:
yay! thanks yall

PR #20081, opened 2026-05-18
Validate entry points in wheels

woodruffw:
When validating wheels on upload, we now additionally validate any entry_points.txt that's present.

This includes checking that the file is well-formed according to the Entry Points specification, and that any `console_scripts` or `gui_scripts` defined have names that match the suggested pattern in the specification. Entry point names outside of those sections are not currently validated, as their semantics are application-defined rather than installer-defined.

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -4024,6 +4024,69 @@ class TestFileUpload:
         ]
         assert resp.status_code == 200
 
+    def test_upload_fails_with_invalid_entrypoints_wheel(
+        self, monkeypatch, pyramid_config, db_request
+    ):
+        """
+        Uploading a wheel fails if the entry_points.txt file is invalid.
+        """
+
+        user = UserFactory.create()
+        pyramid_config.testing_securitypolicy(identity=user)
+        db_request.user = user
+        db_request.user_agent = "warehouse-tests/6.6.6"
+        EmailFactory.create(user=user)
+        project = ProjectFactory.create()
+        release = ReleaseFactory.create(project=project, version="1.0")
+        RoleFactory.create(user=user, project=project)
+
+        temp_f = io.BytesIO()
+        project_name = project.normalized_name.replace("-", "_")
+        with zipfile.ZipFile(file=temp_f, mode="w") as zfp:
+            zfp.writestr("some_file", "some_data")
+            zfp.writestr(f"{project_name}-{release.version}.dist-info/METADATA", "")
+            zfp.writestr(
+                f"{project_name}-{release.version}.dist-info/entry_points.txt",
+                "[console_scripts]\n/bin/evil = evil:main\n",
+            )
+            zfp.writestr(
+                f"{project_name}-{release.version}.dist-info/RECORD",
+                dedent(
+                    f"""\
+                    some_file,
+                    {project_name}-{release.version}.dist-info/METADATA,
+                    {project_name}-{release.version}.dist-info/entry_points.txt,
+                    {project_name}-{release.version}.dist-info/RECORD,
+                    """,
+                ),
+            )
+
+        filename = f"{project_name}-{release.version}-cp34-none-any.whl"
+        filebody = temp_f.getvalue()
+
+        db_request.POST = MultiDict(
+            {
+                "metadata_version": "1.2",
+                "name": project.name,
+                "version": release.version,
+                "filetype": "bdist_wheel",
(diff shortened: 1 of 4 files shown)


ewdurbin (approved):

woodruffw:
Some notes for review:

- Thie validation is intentionally conservative: we only allow bare names for entrypoints. If this causes problems for users, we could relax it to allowing relative names (i.e. relative to the installation prefix).
- This doesn't currently perform other restrictions that installers _may_ perform, such as preventing entrypoints with names like `python`, `python3.exe`, etc. We may wish to do this in the future.
- This repeats an unfortunate pattern we use elsewhere, which is to re-open the wheel file (as a ZIP) just to handle/validate a single member. In the future we should optimize this by aggregating the members we wish to validate so that we only need to open the ZIP once.

warehouse/forklift/legacy.py
```diff
@@ -1450,6 +1452,15 @@ def file_upload(request):
                     filename=filename,
                 )
 
+            try:
+                validate_entrypoints(temporary_filename)
+            except InvalidWheelEntryPointsError:
+                raise _exc_with_message(
+                    HTTPBadRequest,
+                    f"Wheel '{filename}' has invalid entry points defined in "
+                    "the entry_points.txt file",
```
miketheman, line 1461: This message drops the specifics of what the problem was in InvalidWheelEntryPointsError - is that meant to be obscured?

warehouse/forklift/legacy.py
```diff
@@ -1450,6 +1452,15 @@ def file_upload(request):
                     filename=filename,
                 )
 
+            try:
+                validate_entrypoints(temporary_filename)
+            except InvalidWheelEntryPointsError:
```
miketheman, line 1457: Should this new validator have some metric timings and counts around it? It's fine for that to be a fast-follower.

warehouse/utils/wheel.py
```diff
@@ -217,6 +222,74 @@ def validate_record(wheel_filepath: str) -> bool:
...
+    (matching the Entry Points specification's constraints) and
+    that all `console_scripts` and `gui_scripts` entry points have names
+    that do not contain absolute or relative path components.
+
+    Validation errors are not currently reported via email.
+    """
+
+    # See: <https://packaging.python.org/en/latest/specifications/entry-points/#file-format>
+    class CaseSensitiveConfigParser(configparser.ConfigParser):
+        optionxform = staticmethod(str)  # type: ignore[assignment]
```
miketheman, line 254: Should this add `interpolation=None` ? minor tweak...

miketheman (approved):

tests/unit/forklift/test_legacy.py
```diff
@@ -4024,6 +4024,69 @@ def test_upload_record_check_exclusions_only_in_dist_info(
         ]
         assert resp.status_code == 200
 
+    def test_upload_fails_with_invalid_entrypoints_wheel(
+        self, monkeypatch, pyramid_config, db_request
+    ):
+        """
+        Uploading a wheel fails if the entry_points.txt file is invalid.
+        """
```
miketheman, line 4032: Is there any test for a wheel with an entrypoints, or a missing entrypoints.txt file?

warehouse/forklift/legacy.py
woodruffw: Yeah, I made it generic for an initial land. Happy to have it include more details in a follow-up if you think it'd be helpful (I was also thinking that might make sense for an email task, but didn't want to expand the PR too much for the initial fix).

warehouse/utils/wheel.py
woodruffw: That's directly from the EP spec, but I _think_ probably yes. The spec doesn't explicitly forbid interpolation, but in practice I think it should be absolutely verboten 🙂

tests/unit/forklift/test_legacy.py
woodruffw: Yeah, that's covered indirectly via our other wheel tests (which don't include entrypoints at all).

warehouse/forklift/legacy.py
woodruffw: Yeah, will fast-follow for those, thanks!

sethmlarson (approved):
If this emits a track-able metric to DD then this LGTM. Thank you @woodruffw and @miketheman!

woodruffw:
Thanks all! I'll work on those follow-ups in a bit.

PR #20096, opened 2026-05-22
refactor: allow `auth_methods` on  route configurations

### LINKED ISSUE #7266: Refactor authentication mechanisms

miketheman:
See #13854 for historical approach to this idea, but this approach uses route predicates instead.

`auth_methods` is a route predicate, not a view predicate.

The security policies read this in their `identity()` methods (`SessionSecurityPolicy` and `BasicAuthSecurityPolicy`).
`identity()` runs at **authentication** time, right after Pyramid matches the route.
The matched route is sitting there to inspect via `request.matched_route.predicates`.
The view isn't - it hasn't been resolved yet, and there's no request attribute that exposes view predicates anyway. Put this on a view predicate and the policy is blind to it at the one moment it needs the answer.

It's also where the concept belongs. Which credentials an endpoint accepts is a property of the route, not of one view callable. A route can fan out to several views (different methods, accept headers), and all of them should answer to the same auth contract - set once, where the route is defined. This is different than permissions, which may vary based on the HTTP verbs (GET vs POST) - so they can be combined without having to recall the auth_method on each `view_config` (and it's "too late" anyhow).

The predicate is storage-only: `__call__` always returns `True`,
so it never influences routing or view selection.
We already do this with the `domain` route predicate,
which is the same kind of thing: route-level metadata that infrastructure reads.

Fixes #7266

### DIFF
--- a/docs/dev/security/security-policy-flow.md
+++ b/docs/dev/security/security-policy-flow.md
@@ -1,0 +1,497 @@
+# Security Policy Flow
+
+This document traces how an HTTP request is authenticated and authorized in
+Warehouse, starting from the moment gunicorn hands a WSGI request to the
+Pyramid application. It covers how Pyramid's internals call into our security
+policies, how identities are resolved, and how ACLs and the other gates (route
+`auth_methods`, two-factor, verified email, macaroon caveats) combine to allow
+or deny a request.
+
+> This list is the single place that maps the concepts below to files (linked on
+> `main`). The prose and diagrams name symbols, not paths, so moving a file only
+> means updating an entry here.
+
+- [`warehouse/accounts/__init__.py`](https://github.com/pypi/warehouse/blob/main/warehouse/accounts/__init__.py): registers the policy stack (`config.set_security_policy`)
+- [`warehouse/utils/security_policy.py`](https://github.com/pypi/warehouse/blob/main/warehouse/utils/security_policy.py): `MultiSecurityPolicy`, `AuthenticationMethod`, `principals_for`, and the `PERMISSION_AUTH_METHODS` table with its permission/method gate
+- [`warehouse/accounts/security_policy.py`](https://github.com/pypi/warehouse/blob/main/warehouse/accounts/security_policy.py): `SessionSecurityPolicy`, `BasicAuthSecurityPolicy`, and the session authorization gates (verified email, two-factor)
+- [`warehouse/accounts/utils.py`](https://github.com/pypi/warehouse/blob/main/warehouse/accounts/utils.py): `UserContext`
+- [`warehouse/macaroons/security_policy.py`](https://github.com/pypi/warehouse/blob/main/warehouse/macaroons/security_policy.py): `MacaroonSecurityPolicy` (API tokens and Trusted Publishing)
+- [`warehouse/oidc/utils.py`](https://github.com/pypi/warehouse/blob/main/warehouse/oidc/utils.py): `PublisherTokenContext`
+- [`warehouse/predicates.py`](https://github.com/pypi/warehouse/blob/main/warehouse/predicates.py): `AuthMethodsPredicate` and the `auth_methods_for_route` helper
+- [`warehouse/routes.py`](https://github.com/pypi/warehouse/blob/main/warehouse/routes.py) and [`warehouse/forklift/__init__.py`](https://github.com/pypi/warehouse/blob/main/warehouse/forklift/__init__.py): route declarations tagged with `auth_methods`
+- [`warehouse/config.py`](https://github.com/pypi/warehouse/blob/main/warehouse/config.py): `RootFactory` (the default context and the root ACL)
+- [`warehouse/utils/wsgi.py`](https://github.com/pypi/warehouse/blob/main/warehouse/utils/wsgi.py): `ProxyFixer` (trusted client IP) and `VhmRootRemover` WSGI middleware
+- [`warehouse/errors.py`](https://github.com/pypi/warehouse/blob/main/warehouse/errors.py): `WarehouseDenied`
+- [`warehouse/authnz/_permissions.py`](https://github.com/pypi/warehouse/blob/main/warehouse/authnz/_permissions.py): the `Permissions` enum used throughout the ACLs
+
+## Diagram legend
+
+The diagrams below share one color scheme:
+
+- <span style="background-color:#fde4cf; border: 1px solid #f98131; padding: 2px 6px;">Pyramid</span> core machinery (router, traversal, view derivers, ACL helper, per-request cache)
+- <span style="background-color:#cde6f5; border: 1px solid #006dad; padding: 2px 6px;">Warehouse</span> code (our security policies and the gates they wrap around the ACL)
+- <span style="background-color:#fef3c7; border: 1px solid #b45309; padding: 2px 6px;">Service data</span> via `IUserService`/`IMacaroonService`, Postgres/Redis DBs, etc
+- <span style="background-color:#f1f5f9; border: 1px solid #475569; padding: 2px 6px;">External actors</span> (gunicorn, browsers, CLI clients such as `twine`)
+
+The colors apply to the flowcharts. The interaction diagrams are drawn as Mermaid
+sequence diagrams, which Material renders with a single box fill, so those are left
+uncolored; read the participant names for the same categories.
+
+## 1. From gunicorn to the security policy
+
+gunicorn speaks WSGI to the Warehouse application object.
+That object is the Pyramid `Router`, wrapped in our own stack of WSGI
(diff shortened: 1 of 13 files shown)


warehouse/predicates.py
```diff
@@ -27,6 +28,38 @@ def __call__(self, info, request):
...
+        return True
+
+
+def auth_methods_for_route(route) -> frozenset[AuthenticationMethod] | None:
+    """Return the ``auth_methods`` predicate value for a route, if any.
+
+    Returns ``None`` when the route has no ``auth_methods`` predicate
+    configured. Callers decide their own default behavior in that case.
+    """
+    if route is None:
```
ewdurbin: I'm curious why this needs to silently return None, I wouldn't expect success here. Is this _just_ for test stubs?

warehouse/utils/security_policy.py
```diff
@@ -22,6 +23,37 @@ class AuthenticationMethod(enum.Enum):
...
+    Permissions.ProjectsUpload: frozenset(
+        {AuthenticationMethod.MACAROON, AuthenticationMethod.BASIC_AUTH}
+    ),
+    # Migration debt: move to the new API auth surface, then delete.
+    Permissions.APIEcho: frozenset({AuthenticationMethod.MACAROON}),
+    Permissions.APIObservationsAdd: frozenset({AuthenticationMethod.MACAROON}),
+}
+
+
+def permission_allows(
```
ewdurbin: this is named a little generically in my reading. we're realistically checking if an authentication method allows a permission, not if a permission allows something.

`permission_allowed_by_authenticaiton_mechanism` or `mechanism_allows_permission`?

my concern here is that while this is a solid simplification and centralization of the concern, it wasn't initially clear to me that this was "warehouse" and not some other pyramid authnz thing.

warehouse/predicates.py
```diff
@@ -27,6 +28,38 @@ def __call__(self, info, request):
...
+    def text(self):
+        return f"auth_methods = {sorted(m.value for m in self.val)}"
+
+    phash = text
+
+    def __call__(self, info, request):
+        return True
+
+
+def auth_methods_for_route(route) -> frozenset[AuthenticationMethod] | None:
```
ewdurbin, line 49: would it make sense to use `add_request_method` so that `request.route_auth_methods` can be used instead?

warehouse/predicates.py
```diff
@@ -27,6 +28,38 @@ def __call__(self, info, request):
...
+    """Return the ``auth_methods`` predicate value for a route, if any.
+
+    Returns ``None`` when the route has no ``auth_methods`` predicate
+    configured. Callers decide their own default behavior in that case.
+    """
+    if route is None:
+        return None
+    for predicate in getattr(route, "predicates", ()):
```
ewdurbin: is this getattr also guarding against stubs without predicates in tests?

warehouse/routes.py
```diff
@@ -635,13 +635,15 @@ def includeme(config):
     config.add_route(
         "api.echo",
         "/danger-api/echo",
+        auth_methods={"macaroon"},
```
ewdurbin, line 638: this is so clean ✨

ewdurbin (commented):
Overall I'm a fan of the outcome but got confused a few times in review (probably more to do with my chronic lack of understanding of Pyramid authnz 😭, than your implementation)

A few questions inline.

warehouse/predicates.py
miketheman: The only place this is currently used is in the security policy itself.

What use case do you see having this on the request already, since at that point we're "too late" as we've already resolved the route and dispatched to the view?

warehouse/predicates.py
miketheman: Left in from early testing, removing. Nicely spotted!

warehouse/predicates.py
miketheman: Yes, another test-centric guard. Removing, and updating test conditions. One day, pretend, one day.

warehouse/utils/security_policy.py
miketheman: Good feedback - I'll rename to `permission_allowed_by_authentication_method` to not introduce a new concept of "mechanism" as the existing polices use `*Method` in them.

miketheman:
I've also added a new docs page, since this is getting complex and I found myself returning to my notes extensively, so why not share them. See https://warehouse--20096.org.readthedocs.build/security/security-policy-flow/

warehouse/predicates.py
ewdurbin: fair, just felt weird having the function related to a request's route not more directly associated to the request!

ewdurbin:
> I've also added a new docs page, since this is getting complex and I found myself returning to my notes extensively, so why not share them. See https://warehouse--20096.org.readthedocs.build/security/security-policy-flow/

That's actually super helpful, thanks. I am a bit confused why Macaroons "engage" by default, that was suprising to me.

ewdurbin (approved):

docs/dev/security/security-policy-flow.md
```diff
@@ -0,0 +1,497 @@
...
+- [`warehouse/authnz/_permissions.py`](https://github.com/pypi/warehouse/blob/main/warehouse/authnz/_permissions.py): the `Permissions` enum used throughout the ACLs
+
+## Diagram legend
+
+The diagrams below share one color scheme:
+
+- <span style="background-color:#fde4cf; border: 1px solid #f98131; padding: 2px 6px;">Pyramid</span> core machinery (router, traversal, view derivers, ACL helper, per-request cache)
+- <span style="background-color:#cde6f5; border: 1px solid #006dad; padding: 2px 6px;">Warehouse</span> code (our security policies and the gates they wrap around the ACL)
+- <span style="background-color:#fef3c7; border: 1px solid #b45309; padding: 2px 6px;">Service data</span> via `IUserService`/`IMacaroonService`, Postgres/Redis DBs, etc
+- <span style="background-color:#f1f5f9; border: 1px solid #475569; padding: 2px 6px;">External actors</span> (gunicorn, browsers, CLI clients such as `twine`)
```
merwok, line 34: Note for future improvement: declaring background colours without explicit font colours results in poor contrast in dark mode.

warehouse/utils/security_policy.py
```diff
@@ -22,6 +23,37 @@ class AuthenticationMethod(enum.Enum):
     BASIC_AUTH = "basic-auth"
     SESSION = "session"
     MACAROON = "macaroon"
+    # Placeholder for the dedicated API auth surface that will replace
+    # danger-api's macaroon abuse. No policy implements it yet.
```
merwok, line 27: Who or what is danger-api who is abusing macaroons?

miketheman: Me

docs/dev/security/security-policy-flow.md
miketheman: Thanks! I don't use dark mode, and know that mkdocs-material has some support for that, so if you send me a PR showing the right way to do that, happy to adapt!

PR #20101, opened 2026-05-26
fix: close files once done

miketheman:
Without a context manager, file handles remains open and become resource leaks.

Surfaced by turning all test-time warnings into errors, and handling each case.


ewdurbin (approved):
good stuff.

as an aside, unless there's a use-case remaining (mine is long gone) the DBML stuff could probably just... go

miketheman:
Thanks - tracked in https://github.com/pypi/warehouse/issues/20102

PR #20120, opened 2026-06-03
test: replace pretend in the shared test fixtures

miketheman:
Swap the pretend-based conftest fixtures for real objects:

- `metrics` wraps the real `NullMetrics`, so `with metrics.timed(...)` returns a working context manager and method signatures are checked instead of silently accepted.
- `pyramid_request` drops its pretend task, log, and authentication_method stubs; `send_email` follows the same task pattern.
- The OIDC service fixtures take a real cache URL string in place of a `pretend.stub(`).

A `_CallRecorder` wraps each recorded method so one object answers both the old `obj.method.calls == [pretend.call(...)]` assertions and the unittest.mock API. Existing tests keep passing untouched, which lets the rest of the removal land file by file rather than in one sweep.


JacobCoffee (approved):

tests/conftest.py
```diff
@@ -77,50 +79,67 @@
...
+    Wraps a callable -- a real bound method (``metrics``) or a lambda stand-in
+    (``request.task`` / ``request.log`` / ``send_email``) -- so a single object
+    satisfies BOTH the legacy ``pretend``-style
+    ``obj.method.calls == [pretend.call(...)]`` assertions and the modern
+    ``unittest.mock`` API (``assert_called_once_with`` / ``assert_has_calls`` /
+    ``assert_not_called`` / ``call_args_list``). This lets the pretend removal
+    proceed file-by-file rather than as one big-bang sweep; delete it once no
+    ``.calls`` assertions against these fixtures remain.
+
+    See: dev/pretend-migration.md
```
ewdurbin: was this intended to be committed somewhere?

ewdurbin (approved):
Love the support for migrating!

tests/conftest.py
miketheman: Nope, personal notes 🤦 . Will remove!

PR #20126, opened 2026-06-04
Fix re-auth allowing other user's credentials

facutuesca:
Currently, when a user's session is stale and they get the re-auth form, this form has a hidden input field for the username. If one changes this hidden input before submitting the form, one can re-auth the expired session of user `A` using the credentials of user `B`.

Concretely:
1. User Alice is logged in
2. Once Alice's session goes stale after 30 minutes, it requires reauthentication to access certain re-auth protected views
3. On the reauth form, one can change the hidden `username` field from `alice` to another account, e.g. `bob`.
4. Bob's valid password is entered and the form is submitted
5. Alice session is re-authenticated using Bob's credentials, and one is redirected to the protected view of Alice's account

This happens because the password validation during re-auth is not tied to the authenticated session user. This PR fixes that by removing the username hidden field and defining a `validate_password` function that uses the session's user ID instead.

cc @miketheman

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -560,16 +560,12 @@ class WebAuthnAuthenticationForm(WebAuthnCredentialMixin, _TwoFactorAuthenticati
 
 class ReAuthenticateForm(PasswordMixin, wtforms.Form):
     __params__ = [
-        "username",
         "password",
         "next_route",
         "next_route_matchdict",
         "next_route_query",
     ]
 
-    username = wtforms.fields.HiddenField(
-        validators=[wtforms.validators.InputRequired()]
-    )
     next_route = wtforms.fields.HiddenField(
         validators=[wtforms.validators.InputRequired()]
     )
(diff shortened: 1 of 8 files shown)


warehouse/accounts/forms.py
```diff
@@ -580,10 +576,40 @@ class ReAuthenticateForm(PasswordMixin, wtforms.Form):
         validators=[wtforms.validators.InputRequired()]
     )
 
-    def __init__(self, *args, user_service, **kwargs):
+    def __init__(self, *args, user_id, user_service, **kwargs):
         super().__init__(*args, **kwargs)
+        self.user_id = user_id
         self.user_service = user_service
 
+    def validate_password(self, field):
```
ewdurbin, line 584: Thoughts on changing the call signature of `validate_password` from `PasswordMixin` to accept an explicit user_id instead of duplicating the logic inside the `if userid is not None` conditional?

ewdurbin (approved):
Very straightforward fix to a subtle issue, I have a suggestion to consolidate implementation of `validate_password`, and a merge conflict needs resolved.

warehouse/accounts/forms.py
miketheman: This is a good idea, and fine to defer as a follow-up refactor. I'll capture as an issue.

miketheman (approved):

warehouse/accounts/forms.py
miketheman: See https://redirect.github.com/pypi/warehouse/issues/20321

PR #20133, opened 2026-06-08
feat(admin): optionally notify on token deletion

miketheman:
When an admin deletes a macaroon (token), provide the administrator the facility to also send a notification to the user, and allow them to also set a reason to inform the user of why.

### DIFF
--- a/tests/unit/admin/views/test_macaroons.py
+++ b/tests/unit/admin/views/test_macaroons.py
@@ -2,13 +2,13 @@
 
 import uuid
 
-import pretend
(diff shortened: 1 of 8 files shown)


tests/unit/admin/views/test_macaroons.py
```diff
@@ -2,13 +2,13 @@
 
 import uuid
 
-import pretend
```
ewdurbin, line 5: love to see it.

ewdurbin (approved):

PR #20134, opened 2026-06-08
only open uploaded sdist/whl (tar.gz/zip) once

ewdurbin:
we have repeatedly kicked the can down the road on this every time we add another check during upload that accesses file contents.

this moves our opening of the archive into a context manager, so that we don't have to re-open and re-scan tar headers or the zip central-directory every time.

in some extremely specific local benchmarks around the validation steps in the current upload, this resulted in approx 40-50% improvement...


warehouse/forklift/legacy.py
```diff
@@ -272,7 +274,22 @@ def _validate_filename(filename, filetype):
         )
 
 
-def _is_valid_dist_file(filename, filetype, metrics, *, scan=True):
+def _open_dist_file(filename, stack: ExitStack):
```
JacobCoffee, line 285: ive never seen this done this way, so interesting

ewdurbin: passing around an ExitStack is way more fun than writing thick context managers!!!!

JacobCoffee (approved):

Issue #18880, opened 2025-10-16
Tests: Over-mocking leads to errors surfaced in production


miketheman:
One example:
https://github.com/pypi/warehouse/blob/ee2fe4154807fdf21823213939db33e60349ca8e/tests/unit/manage/views/test_oidc_publishers.py#L1190-L1348

These tests don't exercise the stack down to the data layer, leading to issues discovered in production, like #18879, as well as noted in related changes here: https://github.com/pypi/warehouse/pull/18862#pullrequestreview-3336254308

The action here is to start with the trusted publishing tests, and refactor to use less `pretend`, and lean into using `db_request` and factory fixtures as necessary, replacing `pretend` in favor of real objects, with factory fixtures whenever possible.

miketheman:
We have `tests/functional` that exercises about 48% of the overall code today.

However, a lot of those test still use fixtures and stubs, so converting them to use `webtest`-only style testing should happen first, so we can then evaluate what amount of the codebase is handled via webtesting.

miketheman:
Also worth noting: Most tests also do not exercise aspects of view configurations - like HTTP methods, permissions, and HTML logic, tween/middleware behavior, security policy, etc.
The functional web test path is useful to express testing expectations that are otherwise very hard to test.

PR #20141, opened 2026-06-09
test: replace pretend in the organizations, subscriptions, and api tests

miketheman:
Migrate the test suites under `tests/unit/{organizations,subscriptions,api}` off the `pretend` library, using `pytest-mock` and `factory_boy` instead.

- Stubs that stood in for real objects now use a factory build or an existing `conftest` fixtures. Opaque values use `mocker.sentinel`; the few genuinely classless shapes use `types.SimpleNamespace`.
- `call_recorder` + `monkeypatch.setattr` give way to `mocker.patch.object` and `mocker.spy`, and `x.calls == [pretend.call(...)]` checks become `assert_called_once_with`, `call_args_list`, or `assert_not_called`.
- `includeme` tests pass a `mocker.Mock(spec=[...])` as the config, so an unexpected directive still raises.

Refs: #18880

### DIFF
--- a/tests/unit/api/test_config.py
+++ b/tests/unit/api/test_config.py
@@ -1,76 +1,70 @@
 # SPDX-License-Identifier: Apache-2.0
 
+import types
+
 import orjson
-import pretend
 import pytest
 
 from warehouse.api import config
 
 
-def test_api_set_content_type():
-    response = pretend.stub()
-
-    @pretend.call_recorder
+def test_api_set_content_type(mocker):
     def view(context, request):
-        return response
+        return mocker.sentinel.response
 
-    info = pretend.stub(options={"api_version": "v1"})
+    info = types.SimpleNamespace(options={"api_version": "v1"})
     wrapped_view = config._api_set_content_type(view, info)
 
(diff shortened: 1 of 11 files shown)


tests/unit/api/test_config.py
```diff
@@ -1,76 +1,70 @@
...
-    response = pretend.stub()
-
-    @pretend.call_recorder
+def test_api_set_content_type(mocker):
     def view(context, request):
-        return response
+        return mocker.sentinel.response
 
-    info = pretend.stub(options={"api_version": "v1"})
+    info = types.SimpleNamespace(options={"api_version": "v1"})
```
ewdurbin, line 15: TIL, this is a rather cool lil class!

ewdurbin (approved):

tests/unit/api/test_config.py
miketheman: Indeed it is! It's my "last resort" for piece of data that has no specific shape, so I try to use it sparingly. I prefer using more "real" shaped objects whenever possible

PR #20149, opened 2026-06-11
test: replace pretend in the top-level unit test modules

miketheman:
Migrate eleven top-level modules under `tests/unit/` off the `pretend` library, using `pytest-mock` and existing fixtures instead.

- Stubs that stood in for the request now use the real `pyramid_request` fixture; opaque values use `mocker.sentinel`, and the genuinely classless shapes (a view-deriver info, a route, a WTForms field) use `types.SimpleNamespace`.
- `call_recorder` + `monkeypatch.setattr` give way to `mocker.patch.object(autospec=True)` and `mocker.spy`, and `x.calls == [pretend.call(...)]` checks become `assert_called_once_with`, `call_args_list`, or `assert_not_called`.
- `includeme` tests pass a `mocker.Mock(spec=[...])` as the config, so an unexpected directive still raises.

Refs: #18880


JacobCoffee (approved):
in place of pretend, is it better to mock things vs like in https://github.com/pypi/warehouse/pull/20154/changes i used real obbjects. which is preferred? for my own edification

miketheman:
> in place of pretend, is it better to mock things vs like in [#20154 (changes)](https://github.com/pypi/warehouse/pull/20154/changes) i used real obbjects. which is preferred? for my own edification

At some point, I'll likely publish some testing guidelines in the dev docs, but generally speaking, use fixtures, factories, "real" object shapes before mocking unless it's simple/trivial enough that the test doesn't need full shapes.

PR #20154, opened 2026-06-12
Record size limit changes

JacobCoffee:
When an organizations size limit is changed we don't log this in the journal which makes it awkward to try and find who did the change, when, etc.


woodruffw (approved):
Seems reasonable to me 🙂

warehouse/admin/views/organizations.py
```diff
@@ -1059,8 +1059,19 @@ def set_upload_limit(request):
         )
 
     # Form validation has already converted to bytes or None
+    old_upload_limit = organization.upload_limit
     organization.upload_limit = form.upload_limit.data
 
+    organization.record_event(
+        request=request,
+        tag="admin:organization:set_upload_limit",
```
ewdurbin: Should these be added to the enum? https://github.com/pypi/warehouse/blob/main/warehouse/events/tags.py#L157

ewdurbin (commented):
I think we should be adding a tag to the enums and referencing it via that Enum rather than having a raw string.

Also, we should probably include the modifications to Project level limits in this work.

warehouse/admin/views/organizations.py
JacobCoffee: [1fb5012](https://github.com/pypi/warehouse/pull/20154/commits/1fb501283a856b0092f9c513701f4c383dee47bc)

ewdurbin (approved):

morekpnpaksl-afk:
🌹 ❤️

morekpnpaksl-afk:
Uploading VID_20260618_003434.mp4…

PR #20157, opened 2026-06-13
ci: run static lint in pre-commit when frontend files change

JacobCoffee:
Add a `static-lint` pre-commit hook that runs `make static_lint` (eslint + stylelint via the static container) for `.scss`/`.js` changes, plus a `static_lint` Make target mirroring `static_tests`.


JacobCoffee:
Merged `main` into this branch to clear a conflict in `Makefile`: main added `## Run static asset tests` help text to `static_tests`, while this branch added the `static_lint` target directly above it. Kept both and gave `static_lint` matching help text so it shows up in `make help`.

Net diff against main is unchanged otherwise — still just the `static_lint` target and the `static-lint` pre-commit hook.

_🤖 Done via scheduled babysitting job._

miketheman (approved):
Sure

PR #20161, opened 2026-06-15
Add email for company orgs

JacobCoffee:
A lot of company orgs are not using their approved org and just squatting on them.
Only 11% of our company orgs currently have a seat.

Add email to notify orgs that they will need to have an active subscription. 

Does not mention the need for a project although I would prefer that as that would be "real" usage but capturing the seat to confirm they intend on actually using it is a good first step imo.

Future steps would be
- different emails for orgs with files but no seat
- eventual quarantine/releasing the name of orgs that continue with no files OR seat
A lot of people use orgs but an even great number just seem to be name squatting or their venture fell through so it remains empty.


warehouse/cli/organizations.py
```diff
@@ -0,0 +1,126 @@
...
+    subscription and no active manual activation, reminding them that
+    continued use requires purchasing at least one seat.
+    """
+    # avoid circular imports
+    from warehouse.email import send_organization_subscription_required_email
+    from warehouse.organizations.models import Organization, OrganizationType
+
+    env = pyramid.scripting.prepare(registry=config.registry)
+    request = env["request"]
+    session = request.db
```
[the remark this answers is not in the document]
JacobCoffee: maybe, but this may not stay as a CLI as we want it to be continuous.. investigating

warehouse/organizations/tasks.py
```diff
@@ -79,3 +84,34 @@ def update_organziation_subscription_usage_record(request):
...
+def notify_organizations_requiring_subscription(request):
+    """
+    Email owners of company orgs that have no active subscription
+    (or manual activation) that 1 seat is required for paid orgs.
+    """
+    organizations = (
+        request.db.query(Organization)
+        .filter(
+            Organization.is_active.is_(True),
+            Organization.orgtype == OrganizationType.Company,
```
ewdurbin, line 103: Should we filter this to Company orgs older than some threshold? I think it would be rather annoying to ping people within the first N days/weeks of approval. Perhaps we should give a month or so of leeway?

warehouse/templates/email/organization-subscription-required/body.html
```diff
@@ -0,0 +1,22 @@
...
+    Your Company organization <strong>{{ organization_name }}</strong> on PyPI was approved, but it does not currently have an active subscription.
+  </p>
+  <p>
+    Continued use of a Company organization requires a paid subscription with at least 1 seat purchased. Without an active subscription, your organization's features (such as uploads and member invitations) may be restricted.
+  </p>
+  <p>
+    <strong><a href="{{ request.route_url('manage.organization.activate_subscription', organization_name=organization_name) }}">Activate your subscription</a></strong>
+  </p>
+  <p>
+    If your organization is no longer using PyPI Organizations features, or if you believe you qualify for a Community organization instead, please reach out to us at <a href="mailto:support@pypi.org">support@pypi.org</a>.
```
ewdurbin: just a callout that it would be good to get a 👍🏼 from @Thespi-Brain here on copy/call-to-action since we're directing folks there.

might also be good to add a +orgs to the mailto so that its tagged automatically by helpscout?

warehouse/email/__init__.py
```diff
@@ -641,6 +641,22 @@ def send_organization_deleted_email(request, user, *, organization_name):
     }
 
 
+@_email(
+    "organization-subscription-required",
+    repeat_window=datetime.timedelta(days=30),
```
ewdurbin, line 646: TIL we had a repeat_window helper here, that's nice! So we only bother folks once a month after first notice, nice!

ewdurbin (approved):
Seems good to me, a few thoughts/questions/suggestions.

warehouse/templates/email/organization-subscription-required/body.html
JacobCoffee: [6cfe4f5](https://github.com/pypi/warehouse/pull/20161/commits/6cfe4f542b200923dd8b9152ad4e31d15dac984a)

warehouse/organizations/tasks.py
JacobCoffee: [e1c0b96](https://github.com/pypi/warehouse/pull/20161/commits/e1c0b961fcdefbfd5e7baf43a46a47ebc3a77a19)

warehouse/templates/email/organization-subscription-required/body.txt
```diff
@@ -0,0 +1,22 @@
...
+{% block content %}
+Hello {{ username }},
+
+Your Company organization '{{ organization_name }}' on PyPI was approved, but it does not currently have an active subscription.
+
+Continued use of a Company organization requires a paid subscription with at least 1 seat purchased. Without an active subscription, your organization's features (such as uploads and member invitations) may be restricted.
+
+To activate a subscription, visit your organization's billing page:
+
+{{ request.route_url('manage.organization.activate_subscription', organization_name=organization_name) }}
```
miketheman, line 13: This, and the one in .html are showing `localhost` as their URL, since this is missing the `_host` param with the correct domain.

Folks have been emailing in with this as a concern.

PR #20175, opened 2026-06-18
Add checks if a user is only owner of org

### LINKED ISSUE #20174: Organizations can be orphaned
Currently if an organization has one owner but that account is removed the organization remains, but ownerless.

JacobCoffee:
1. Disallow users from deleting their account of they still retain an org.
2. If admins delete an account, record it and deactivate it

The second option im not sure if its the right move or not. I would think either, in order of most preferred: 
- disallow nuking an account with orgs until they are dealt with (or optionally double nuke do delete all orgs that user has?)
- do the recording but transfer orgs to the deleting admin
- or push to a system account

Also for ref. there are currently ~22 orphaned orgs, one being of significance (in name)

Closes #20174


warehouse/manage/views/__init__.py
```diff
@@ -551,6 +556,12 @@ def delete_account(self):
             )
             return self.default_response
 
+        if self.sole_organizations:
```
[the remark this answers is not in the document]
JacobCoffee: [c1b1803](https://github.com/pypi/warehouse/pull/20175/commits/c1b18037458ae8e52f5721a4e2d66d41a90e53f4)

warehouse/admin/views/users.py
```diff
@@ -384,6 +390,46 @@ def _nuke_user(user, request):
...
+            tag=EventTag.Organization.OrganizationRoleRemove,
+            request=request,
+            additional={
+                "submitted_by_user_id": str(request.user.id),
+                "role_name": OrganizationRoleType.Owner.value,
+                "target_user_id": str(user.id),
+                "reason": "nuked",
+            },
+        )
+        organization.is_active = False
```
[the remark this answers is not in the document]
JacobCoffee: [0cf43b3](https://github.com/pypi/warehouse/pull/20175/commits/0cf43b3b171f8ac1c6280219787f33d2c19d55bb)

warehouse/manage/views/__init__.py
```diff
@@ -356,6 +357,16 @@ def account_associations(self):
...
         return user_projects(request=self.request)["projects_sole_owned"]
 
+    @property
+    def sole_organizations(self):
+        return [
+            organization
+            for organization in user_organizations(request=self.request)[
+                "organizations_with_sole_owner"
+            ]
+            if organization.is_in_good_standing()
```
ewdurbin: do we want to gate this on account standing? we probably don't want people to orphan lapsed or unactivated orgs either?

ewdurbin (approved):
overall feels OK/right to me... i'm also fairly iffy on the admin side of things here. might be a good idea to add a confirm modal in the scenario in which a user being nuked owns an org?

warehouse/manage/views/__init__.py
JacobCoffee: https://github.com/pypi/warehouse/pull/20234 now allows lapsed/unactivated orgs so we can do this

JacobCoffee: [36a22a6](https://github.com/pypi/warehouse/pull/20175/commits/36a22a6ace852fa0fe317183fb0b1097ca8d1dcc) and [13e69a2](https://github.com/pypi/warehouse/pull/20175/commits/13e69a24a13e9d45ee765e8a8edeae638678e60e) for your suggested ahnges

JacobCoffee:
want to test the ui bits more on recent commits

JacobCoffee:
It works nicely. 

image
image
image

PR #20177, opened 2026-06-18
Add helpful org status filters

JacobCoffee:
Adds more filters to organizations

also adds real data for subscription status, before we just had an empty column so good standing and subscription were shifted over, rendering on the wrong columns

image


warehouse/admin/views/organizations.py
```diff
@@ -212,17 +245,34 @@ def organization_list(request):
...
             elif field == "type":
                 if "company".startswith(value):
                     filters.append(Organization.orgtype == OrganizationType.Company)
                 elif "community".startswith(value):
                     filters.append(Organization.orgtype == OrganizationType.Community)
+            elif field == "activity":
+                has_projects, has_recent_release = _organization_activity_predicates()
+                active = Organization.is_active.is_(True)
+                if "inactive".startswith(value):
+                    filters.append(Organization.is_active.is_(False))
```
ewdurbin, line 257: do we have _any_ Organizations which have `is_active` set to False? i do not think there is any mechanism to flip that bit manually or automatically.

ewdurbin (approved):
i think too much weight is being given to the `is_active` flag on Organizations... but otherwise approve if this fits needs of psf staff

warehouse/admin/views/organizations.py
JacobCoffee: There soon will be through https://github.com/pypi/warehouse/pull/20175/ but nothing else at the moment. More things soon (?)

PR #20180, opened 2026-06-18
drop alternate_repositories table

ewdurbin:
followup to #19915


ewdurbin:
Considered a backup of the table, but as #19915 retained the relevant Project events, the data can be reconstructed if it was for some reason urgent. Given that 708 is dead, I do not think it is worth further effort.

woodruffw (approved):
LGTM! Might want to remove the generated Alembic comments, but minor.

ewdurbin:
> LGTM! Might want to remove the generated Alembic comments, but minor.

i never do this, it would be a good idea to do it, 😂

PR #20181, opened 2026-06-18
Schedule S3 -> B2 Syncs

### LINKED ISSUE #19704: Many files are not synced across archive and cache
Noticed today, when investigating other file-related behaviors

Not all files are in both the `archive` (s3) and `cache` (b2) storages:

```sql
warehouse=> SELECT COUNT(*) AS file_count FROM release_files WHERE NOT cached;
 file_count
------------
    7725023
(1 row)

-- Even newer files are missing:

warehouse=> SELECT date_trunc('month', upload_time) AS month,
         COUNT(*) AS file_count
  FROM release_files
  WHERE NOT cached
    AND upload_time >= date_trunc('month', now()) - interval '3 years'
  GROUP BY month
  ORDER BY month;
        month        | file_count
---------------------+------------
 2023-03-01 00:00:00 |     178073
 2023-04-01 00:00:00 |     202665
 2023-05-01 00:00:00 |      81463
 2023-06-01 00:00:00 |        593
 2023-07-01 00:00:00 |        165
 2023-08-01 00:00:00 |        232
 2023-09-01 00:00:00 |        146
 2023-10-01 00:00:00 |        389
 2023-11-01 00:00:00 |        276
 2023-12-01 00:00:00 |        186
 2024-01-01 00:00:00 |        150
 2024-02-01 00:00:00 |         19
 2024-03-01 00:00:00 |        486
 2024-04-01 00:00:00 |       3173
 2024-05-01 00:00:00 |       1982
 2024-06-01 00:00:00 |        749
 2024-07-01 00:00:00 |        498
 2024-08-01 00:00:00 |        424
 2024-09-01 00:00:00 |        983
 2024-10-01 00:00:00 |       1097
 2024-11-01 00:00:00 |        858
 2024-12-01 00:00:00 |        209
 2025-01-01 00:00:00 |        199
 2025-02-01 00:00:00 |        456
 2025-03-01 00:00:00 |        320
 2025-04-01 00:00:00 |       1259
 2025-05-01 00:00:00 |        106
 2025-06-01 00:00:00 |        177
 2025-07-01 00:00:00 |         82
 2025-08-01 00:00:00 |         94
 2025-09-01 00:00:00 |        115
 2025-10-01 00:00:00 |         75
 2025-11-01 00:00:00 |        231
 2025-12-01 00:00:00 |        139
 2026-01-01 00:00:00 |        587
 2026-02-01 00:00:00 |         97
 2026-03-01 00:00:00 |         67
(37 rows)
```

The metric `warehouse.packaging.files.not_cached` has been pretty static for the past 12+ months:

Image

So the problem doesn't appear to be getting materially worse, which is great, but we still have fresher files not being synced to cache. This issue isn't currently concerned with why not, rather how to proceed when this situation happens.

In #13651 the `archived` column fell out of use so it naturally has many `false` entries, so I didn't use it as part of the query.

There doesn't seem to be a periodic task to reconcile - was there ever?

I've run a few iterations in a prod shell to confirm that the numbers change, but since the "batch" will only update once the task is completed, it's risky to set a long-running command with more than ~1000 at a time, since if interrupted it'll have to start over.

A batch of 1000 takes ~11 minutes, so if we ran it continuously, we'd eventually get there in ~2-3 months.

1. Are we fine with having a persistent fallback, and not everything synced to cache?

    <figure>
      Image
      <figcaption><i>Fastly fetching from fallback</i></figcaption>
    </figure>

2. Should we set up a periodic reconcile task every 15 minutes to catch us up, and then ease off the periodicity once caught up to run hourly/daily/etc?

3. Something else entirely?

JacobCoffee:
Closes #19704

### DIFF
--- a/warehouse/packaging/__init__.py
+++ b/warehouse/packaging/__init__.py
@@ -21,6 +21,7 @@ from warehouse.packaging.tasks import (
     compute_2fa_metrics,
     compute_packaging_metrics,
     compute_top_dependents_corpus,
+    reconcile_file_storages,
     update_description_html,
 )
 
@@ -170,6 +171,9 @@ def includeme(config):
     )
 
     config.add_periodic_task(crontab(minute="*/1"), check_file_cache_tasks_outstanding)
+
+    # Sync S3 to B2
+    config.add_periodic_task(crontab(minute="*/15"), reconcile_file_storages)
 
     config.add_periodic_task(crontab(minute="*/5"), update_description_html)
     config.add_periodic_task(crontab(minute="*/5"), update_role_invitation_status)
--- a/tests/unit/packaging/test_init.py
+++ b/tests/unit/packaging/test_init.py
@@ -18,11 +18,12 @@ from warehouse.packaging.models import File, Project, Release, Role
 from warehouse.packaging.services import project_service_factory
 from warehouse.packaging.tasks import (
     check_file_cache_tasks_outstanding,
+    reconcile_file_storages,
     update_description_html,
 )
 
 
-def test_includeme(monkeypatch):
+def test_includeme(monkeypatch, mocker):
     storage_class = pretend.stub(
         create_service=pretend.call_recorder(lambda *a, **kw: pretend.stub())
     )
@@ -51,7 +52,7 @@ def test_includeme(monkeypatch):
         registry=pretend.stub(settings=settings),
         register_origin_cache_keys=pretend.call_recorder(lambda c, **kw: None),
         get_settings=lambda: settings,
-        add_periodic_task=pretend.call_recorder(lambda *a, **kw: None),
+        add_periodic_task=mocker.Mock(),
     )
 
     packaging.includeme(config)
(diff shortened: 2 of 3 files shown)


ewdurbin:
Out of curiosity, have you run the reconciler manually to ensure it's working as expected? Not a blocker... just curious.

JacobCoffee:
yeah, I ran it earlier with a small batch size

```
/opt/warehouse/src$ python -m warehouse storages reconcile --batch-size 1
Running reconcile_file_storages with batch_size 1...
Checking File<...> (16/2c/project-here-0.0.1.tar.gz)...
$project pulled from archive ⬆
{"obj_class": "File", "state": "dirty", "keys": ["project/project-here"], "changed_attrs": ["cached"], "event": "cache_purge_keys_generated", "logger": "warehouse.cache.origin", "level": "info"}
{"count": 1, "keys": ["project/project-here"], "event": "cache_purge_executing", "logger": "warehouse.cache.origin", "level": "info"}
/opt/warehouse/src$
```

warehouse/packaging/__init__.py
```diff
@@ -171,6 +172,9 @@ def includeme(config):
 
     config.add_periodic_task(crontab(minute="*/1"), check_file_cache_tasks_outstanding)
 
+    # Sync S3 to B2
+    config.add_periodic_task(crontab(minute="*/15"), reconcile_file_storages)
```
ewdurbin, line 176: default batch size is 100, (not sure if we have a setting overriding it in config)

that's _probably_ no problem every 15 minutes, but IIRC this task isn't really designed to run concurrently. should we have a guard to make sure two running instances of reconcile_file_storages (for whatever reason) don't conflict?

JacobCoffee: how do you find [de7e1a0](https://github.com/pypi/warehouse/pull/20181/commits/de7e1a0e1ad4e4183c058e1277172caf25ddd278)?

ewdurbin: I find it extremely elegant.

ewdurbin (approved):

PR #20186, opened 2026-06-19
Admin org billing view

JacobCoffee:
image

Will want a followup to record events and show them in ui

Closes #20176

Issue #20176, opened 2026-06-18
Organization Admin Billing View


JacobCoffee:
Admins can't see or manage org subscriptions besides via Stripe or raw DB access

### Org detail page — add a Billing section
- [ ] List each subscription: status, plan/product, price + interval, seats, `subscription_id` (link to Stripe)
- [ ] Show customer id + billing email
- [ ] Cancel subscription button (`cancel_subscription`)
- [ ] Open Stripe billing portal link (`create_portal_session`)

### Org list
- [ ] Fix the broken "Subscription" column (header renders, no cell) — show active-sub status or remove it

### Report
- [ ] List Company orgs that are ownerless or inactive but still have an active subscription (billing leaks)

PR #20188, opened 2026-06-19
Organization billing events

JacobCoffee:
Followup for #20186, part of #20176 

Adds historics table for billing 

image


JacobCoffee:
Could also do with showing who started the subscription but i think we'd have to stuff some more metadata in `create_checkout_session` or somewhere to make the trip to/from stripe, so maybe a followup PR

ewdurbin (approved):

PR #20189, opened 2026-06-19
Add path-based PR labeling

### LINKED ISSUE #20187: Auto-label PRs based on paths touched
It would be nice to label PRs based on existing labels based on the paths of files touched in a PR.

Touch `.github/` -> `github_actions`
Touch `warehouse/admin` -> `admin`
etc.

This will save time manually adding the labels.

GitHub has an official action available for this:
https://github.com/actions/labeler

---

omkar-334:
Adds GitHub's official pull request labeler so Warehouse PRs get existing component labels based on changed file paths. The config covers common areas like GitHub Actions, admin, docs, organizations, trusted publishing, JavaScript, CSS, Docker, and dependencies.

- Uses only labels that already exist in `pypi/warehouse`.
- Skips broad low-signal labels such as `python`, `testing`, and `HTML`.
- Pins `actions/labeler` to the v6.1.0 commit SHA.
- Runs on `pull_request_target` without checking out or executing PR code.

Resolves #20187


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,23 @@
+name: Pull Request Labeler
+
+on:
+  pull_request_target:
```
JacobCoffee, line 4: this is unsafe :)

JacobCoffee (changes requested):

.github/labeler.yml
```diff
@@ -0,0 +1,206 @@
...
+          - ".github/**"
+
+i18n:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "babel.cfg"
+          - "warehouse/i18n/**"
+          - "tests/unit/i18n/**"
+
+javascript:
```
JacobCoffee, line 117: is there a glob for just *.js?

JacobCoffee: in the event any of these files are deleted or new .js files are added, it doesnt scale well. same concept applies to other things in this file too

.github/workflows/labeler.yml
JacobCoffee, line 1: I'd 
- convert file-type labels to extension globs. (*.html, *.js, etc.) 
- convert any enumerated file to a directory glob where a directory exists like 
the rest can stay maybe? but its a maintenance cost

omkar-334:
I pushed this to my fork and opened a dummy pr to check if it works - https://github.com/omkar-334/warehouse/pull/1...

.github/workflows/labeler.yml
JacobCoffee: more details https://research.jfrog.com/post/part-1-pull-request-target-exploitation/ , and you can search around. we've removed all instances of pull_request_taget for the `python` org, dont want to introduce it here

JacobCoffee:
image

image
yeah looks okay so far :)

.github/workflows/labeler.yml
omkar-334: oh wow , sorry, i didn't know this.... i'll look into an alternative rather than `pull_request_target`

omkar-334:
hi @JacobCoffee 
I went through the https://github.com/actions/labeler code - and it does not read, checkout or execute the contents of any files. It mostly uses the github typescript api to get the list of files changed in the PR and the PR metadata (pr number, existing labels.    
The only user-facing inputs are the filenames and branch names - which are again only used in glob/regex commands for matching. Another doubt is - I agree `pull_request_target` is a vulnerability, but wouldnt the `labeler` repo switch to something safer if it was a concern in that action specifically?  

 -----  
 
However, I was able to get an alternative method working - with 2 workflows - 

- A no-permission `pull_request` workflow records the PR number as an artifact
- a trusted workflow_run workflow downloads and validates that number before applying labels.

we would need the following for this approach - `actions/labeler`, `actions/upload-artifact`, and `actions/download-artifact`.   

I didnt commit it to this PR yet, but i tested it out on my fork's main branch, along with a few test PRs - it looks to be working fine... Can you take a look at the changes so we can decide how to procced - https://github.com/pypi/warehouse/compare/main...omkar-334:warehouse:main  
or specifically this commit - https://github.com/omkar-334/warehouse/commit/77344b8194c5bf6b19a3ad5830c4851a7e2003e6

Dummy PRs - https://github.com/omkar-334/warehouse/pull/6, https://github.com/omkar-334/warehouse/pull/7, https://github.com/omkar-334/warehouse/pull/9, https://github.com/omkar-334/warehouse/pull/9

.github/labeler.yml
omkar-334: understood. Changed this

gubonlucid-com (approved):

gubonlucid-com (approved):

woodruffw (changes requested):
Unfortunately I don't think there's a safe way to do this in GHA -- the safe(r) trigger here is `pull_request`, but that won't run with the appropriate permissions on third-party PRs.

I think the "right" way to do this is to use a GitHub App, but that requires a lot more infrastructure and engineering than just a GitHub Actions workflow.

omkar-334:
@woodruffw agreed, with `pull_request`, the workflow won't run on third-party PRs, i tried it out.  
However, I've seen `pull_request_target` being used by other organizations for this specific labelling workflow - 
example -Intel -  https://github.com/open-edge-platform/geti/blob/develop/.github/workflows/labeler.yaml

woodruffw:
> However, I've seen `pull_request_target` being used by other organizations for this specific labelling workflow -
> example -Intel - [open-edge-platform/geti@`develop`/.github/workflows/labeler.yaml](https://github.com/open-edge-platform/geti/blob/develop/.github/workflows/labeler.yaml?rgh-link-date=2026-07-07T14%3A41%3A35Z)

The fact that other organizations use an insecure trigger is not, in my opinion, a good reason for us to use it.

(It happens to be the case that `actions/labeler` is one of the few safe uses of `pull_request_target`, but IMO the risk is still not worth it, particularly given that we would need to lift out current org-wide ban on `pull_request_target`.)

omkar-334:
Understood. Definietly, its not worth lifting the ban just for this workflow. 

What do you think of my workaround described above? Is it worth doing it like this
> However, I was able to get an alternative method working - with 2 workflows -
> 
> * A no-permission `pull_request` workflow records the PR number as an artifact
> * a trusted workflow_run workflow downloads and validates that number before applying labels.

woodruffw:
`workflow_run` is approximately as insecure as `pull_request_target` (in some sense it's even more dangerous, since it's easy to spoof the "origin" trigger). I think we should _probably_ ban `workflow_run` at the org level as well, if we haven't already.

Unfortunately, I think we're not going to be able to make progress on this with a purely GHA-based approach. An app is probably the right approach, but per above I think that's a nontrivial undertaking that probably needs to be scoped out first (to determine if it's worth doing).

JacobCoffee:
thank you so much @omkar-334 for the work but as @woodruffw mentioend there's no nice way to do this. 

i'm gonna go ahead and close for now and later explore a github app POC we can deploy across all orgs in the python enterprise if repo maintainers would find it useful.

PR #20193, opened 2026-06-21
Add core-metadata (PEP 658/714) hashes to the JSON API

cooperlees:
## What & why

The Simple/Index API already exposes the hash of each file's [Core Metadata](https://packaging.python.org/en/latest/specifications/core-metadata/) (the `.metadata` file served alongside a distribution) via the `core-metadata` key, as standardized by [PEP 658] and [PEP 714]. For example, [`/simple/black/`](https://pypi.org/simple/black/) carries it today.

The legacy JSON API (`/pypi/<project>/json`) does **not**, which means consumers that mirror PyPI — notably [bandersnatch](https://github.com/pypa/bandersnatch) — cannot discover and serve PEP 658/714 Core Metadata from the JSON API and have to fall back to the Simple API. This then adds an extra PyPI call to get the simple API adding load on PyPI and more network calls for clients we'd love to not need to do.

This PR adds a `core-metadata` key to every file object returned by both JSON routes:

- `GET /pypi/<project>/json` (under `urls` and `releases`)
- `GET /pypi/<project>/<version>/json` (under `urls`)

## Format

It mirrors the Simple/Index API format exactly so it can be passed straight through by mirrors:

- `false` when no metadata file is available
- `{"sha256": "<hex>"}` when it is

```json
{
  "filename": "black-...-py3-none-any.whl",
  "digests": { "md5": "...", "sha256": "...", "blake2b_256": "..." },
  "core-metadata": { "sha256": "..." },
  ...
}
```

The hyphenated key name `core-metadata` is intentional: it matches the PEP 714 standardized field name and the existing Simple/Index API key, rather than the JSON API's usual snake_case, so mirrors can reuse it without translation.

## Tests & docs

- Updated `tests/unit/legacy/api/test_json.py` to assert the new key, covering both the populated (`{"sha256": ...}`) and absent (`false`) branches. `warehouse/legacy/api/json.py` remains at 100% coverage.
- Updated the user-facing JSON API docs (`docs/user/api/json.md`) with a description of the field and example payloads.

Refs https://github.com/pypa/bandersnatch/issues/1124

[PEP 658]: https://peps.python.org/pep-0658/
[PEP 714]: https://peps.python.org/pep-0714/

🤖 Generated with [Claude Code](https://claude.com/claude-code)


miketheman (approved):
Seems like a low-impact change to me - especially on the non-standard JSON endpoint.

@cooperlees curious - can bandersnatch tolerate a failure/missing JSON endpoint and fall back to using Simple Index endpoints? I don't think that will matter to this specific change, but more along the lines of "we probably need to design a better mirroring protocol and endpoints in the future" kind of thing.

cooperlees:
> can bandersnatch tolerate a failure/missing JSON endpoint and fall back to using Simple Index endpoints?

No. It's all designed around the JSON API and generating its own Simple HTML/JSON today. But I am super open to changing that to just pull the SimpleAPI artifacts directly and what else we come up with.

At least it no longer needs xmlrpc.

> "we probably need to design a better mirroring protocol and endpoints in the future" kind of thing.

Yeah, we need to workout all the metadata that's needed from the JSON API and we can build from there.

Sadly tho, plugins use many elements in the JSON API metadata to allow people to filter what packages to mirror today.

PR #20195, opened 2026-06-22
Allow support to burn codes

JacobCoffee:


warehouse/admin/views/users.py
```diff
@@ -695,7 +695,7 @@ def user_recover_account_complete(user: User, request):
 @view_config(
     route_name="admin.user.burn_recovery_codes",
     require_methods=["POST"],
-    permission=Permissions.AdminUsersWrite,
+    permission=Permissions.AdminUsersAccountRecoveryWrite,
```
ewdurbin: while this works, I think it would be better to have a named permission for this action, and add that to the admin/support role

JacobCoffee: tht makes sense, i did it in [de41cc7](https://github.com/pypi/warehouse/pull/20195/commits/de41cc739d3da0bae9b68f6b6e1516e6727ac432)

ewdurbin (approved):

Issue #20125, opened 2026-06-04
Allow users to "burn" tokens


woodruffw:
TL;DR: it's sometimes nice to be able to "burn" (i.e. revoke) an API token, using just the token itself as the latent credential. Some use cases for this:

- Trusted Publishing uses a short-lived API token, but we can make it even shorter-lived by burning it after the user is done with it. For `gh-action-pypi-publish`, for example, this could be done with a post-step in the action. Some other indices (notably crates.io) do this for their implementation of Trusted Publishing.
- Third parties that accidentally discover a credential can get it proactively revoked. The main example of this is security vendors/researchers who find PyPI credentials online in various places; at the moment they do this either manually (backchanneling the credential to us) or indirectly (e.g. through one of our secret scanning partner integrations), but we could make that smoother/more reliable.

A rough design sketch:

* `POST /_/secrets/burn-token` should accept our normal authentication headers and no payload. Like with the secret scanning endpoints, we'd return HTTP 204 to acknowledge the request but not comment on its outcome.
* If the credential is valid, we'd then (1) revoke it and (2) take credential-specific reporting steps. For TP API tokens this would probably be nothing else; for user API tokens it would probably be some kind of revocation email (similar to the one we send ATM for secret scanning)

Some other notes/risks:

* What about bulk burns, e.g. if someone has found 1000+ API tokens and needs to revoke them? 
* Does this pose a risk in terms of backend load?
* Should we allow this endpoint to convey a revocation hint as well ("I'm submitting this because X")? It's not clear we'd get much value from it.

CC @sethmlarson @miketheman per discussions on Slack 🙂 

See https://github.com/pypi/warehouse/issues/13409#issuecomment-1544430725 for context.

woodruffw:
Another related use case: https://github.com/pypi/warehouse/issues/12765#issuecomment-1462345065

(That one is arguably more marginal, since it better fits the shape/role of a management API.)

woodruffw:
Prior art: https://docs.github.com/en/rest/credentials/revoke?apiVersion=2026-03-10#revoke-a-list-of-credentials

Edit: The GitHub API design is nice since it allows bulk revocation. They also do HTTP 202 instead, which seems fine and probably more semantic than 204.

guedou:
I would really like to have access to this API.

In my experience, it’s necessary to be able to add a custom field. For example, one called “message", so that I can explain where the issue occurred.

woodruffw:
[...]

Issue #16785, opened 2024-09-24
Improve runtime safety by checking untyped definitions


miketheman:
We use mypy to help prevent some runtime errors for passing the wrong types of objects around.

In a recent attempt to replace integer math with a timedelta duration, the change from an `int` to ` float` returned type worked in almost every case, except for when the value was being [passed into a redis `setex` command as the expiration](https://github.com/pypi/warehouse/blob/5a4d2cadec641b5d6a6847d0127940e0f532f184/warehouse/sessions.py#L298-L305), which only [accepts integers or durations](https://github.com/redis/redis-py/blob/7d73d742d6df4b3d6f6a1f0ec73e72d4a02add74/redis/typing.py#L27).

The hope was that with that change, if the typed function received the incorrect type, a dev-time error would be raised, however nothing was raised, so we felt it in production, which was rolled back.

Giving mypy enough "hints" to trigger its checks was this diff, which wasn't added to the code yet:

```diff
diff --git a/warehouse/sessions.py b/warehouse/sessions.py
index 902e55dcb..591948688 100644
--- a/warehouse/sessions.py
+++ b/warehouse/sessions.py
@@ -224,7 +224,7 @@ class SessionFactory:
     cookie_name = "session_id"
     max_age = 12 * 60 * 60  # 12 hours

-    def __init__(self, secret, url):
+    def __init__(self, secret, url) -> None:
         self.redis = redis.StrictRedis.from_url(url)
         self.signer = crypto.TimestampSigner(secret, salt="session")

@@ -275,7 +275,7 @@ class SessionFactory:

         return session

-    def _process_response(self, request, response):
+    def _process_response(self, request, response) -> None:
         # If the request has an InvalidSession, then the view can't have
         # accessed the session, and we can just skip all of this anyways.
         if isinstance(request.session, InvalidSession):
```

Had these been in place, the check would have failed, and we may have prevented that issue.

However, applying hints selectively everywhere isn't a good enough safety net, at least until we have a 100% type coverage, which in and of itself isn't always a great goal.

Taking a hint from one of the `--strict` settings of mypy, we can enable checking the bodies of untyped signatures with [`--check-untyped-defs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs) so the checks would happen even if we don't explicitly add types, leaning on type inference, even if we don't add type hints everywhere.

[...]

PR #20197, opened 2026-06-22
Allow burning of TP tokens

woodruffw:
This implements the (hopefully) uncontroversial part of #20125: we don't expose a general-purpose revocation API for now, just one that's specific to Trusted Publishing flows.

The basic idea: clients (like twine and uv) that implement Trusted Publishing can choose to explicitly "burn" their temporary token immediately after it's used. This adds a layer of protection atop TP tokens already being short-lived, since the publishing client can proactively invalidate them before they'd otherwise expire.

For prior art: crates.io's implementation of Trusted Publishing supports this, and their default integration (`rust-lang/crates-io-auth-action`) performs it by default as a post-hook when invoked:

https://github.com/rust-lang/crates-io-auth-action/blob/a1c979fa8eec9cf00f2c01904184fe133deb0ce9/src/post.ts#L29-L47

Signed-off-by: William Woodruff <william@yossarian.net>

### DIFF
--- a/tests/unit/oidc/test_views.py
+++ b/tests/unit/oidc/test_views.py
@@ -1202,7 +1202,8 @@ def test_burn_oidc_issued_token_user_macaroon(metrics, monkeypatch):
 
 
 def test_burn_oidc_issued_token_success(metrics):
-    publisher = pretend.stub(publisher_name="GitHub")
+    projects = [pretend.stub(record_event=pretend.call_recorder(lambda *a, **kw: None))]
+    publisher = pretend.stub(publisher_name="GitHub", projects=projects)
     macaroon = pretend.stub(
         id="fake-macaroon-id",
         oidc_publisher=publisher,
@@ -1238,6 +1239,14 @@ def test_burn_oidc_issued_token_success(metrics):
             "warehouse.oidc.burn_oidc_issued_token",
             tags=["status:success", "publisher_name:GitHub"],
         ),
+    ]
+
+    assert macaroon.oidc_publisher.projects[0].record_event.calls == [
+        pretend.call(
+            tag=EventTag.Project.ShortLivedAPITokenRevoked,
+            request=request,
+            additional={},
+        )
     ]
 
 
--- a/tests/unit/oidc/test_views.py
+++ b/tests/unit/oidc/test_views.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
+import http
 import json
 
 from datetime import datetime
@@ -96,7 +97,7 @@ def test_mint_token_from_oidc_not_enabled(token, service_name, request):
     )
 
     response = views.mint_token_from_oidc(request)
-    assert request.response.status == 422
+    assert request.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert response == {
         "message": "Token request failed",
         "errors": [
@@ -140,7 +141,7 @@ def test_mint_token_from_oidc_invalid_payload(body):
     req = Request()
     resp = views.mint_token_from_oidc(req)
 
-    assert req.response.status == 422
+    assert req.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert resp["message"] == "Token request failed"
     assert isinstance(resp["errors"], list)
     for err in resp["errors"]:
@@ -179,7 +180,7 @@ def test_mint_token_from_oidc_invalid_payload_malformed_jwt(body):
     req = Request()
     resp = views.mint_token_from_oidc(req)
 
-    assert req.response.status == 422
+    assert req.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert resp["message"] == "Token request failed"
     assert isinstance(resp["errors"], list)
     for err in resp["errors"]:
@@ -212,7 +213,7 @@ def test_mint_token_from_oidc_jwt_decode_leaky_exception(monkeypatch):
         pretend.call("jwt.decode raised generic error: oops")
     ]
 
-    assert req.response.status == 422
+    assert req.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert resp["message"] == "Token request failed"
     assert isinstance(resp["errors"], list)
     for err in resp["errors"]:
@@ -245,7 +246,7 @@ def test_mint_token_from_oidc_unknown_issuer(metrics):
     req = Request()
     resp = views.mint_token_from_oidc(req)
 
-    assert req.response.status == 422
+    assert req.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert resp["message"] == "Token request failed"
     assert isinstance(resp["errors"], list)
     for err in resp["errors"]:
@@ -329,7 +330,7 @@ def test_mint_token_from_trusted_publisher_verify_jwt_signature_fails():
     response = views.mint_token(
         oidc_service, DUMMY_GITHUB_OIDC_JWT, claims["iss"], request
     )
-    assert request.response.status == 422
+    assert request.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert response == {
         "message": "Token request failed",
         "errors": [
@@ -365,7 +366,7 @@ def test_mint_token_trusted_publisher_lookup_fails():
     response = views.mint_token(
         oidc_service, DUMMY_GITHUB_OIDC_JWT, claims["iss"], request
     )
-    assert request.response.status == 422
+    assert request.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert response == {
         "message": "Token request failed",
         "errors": [
@@ -409,7 +410,7 @@ def test_mint_token_duplicate_token():
     response = views.mint_token(
         oidc_service, DUMMY_GITHUB_OIDC_JWT, claims["iss"], request
     )
-    assert request.response.status == 422
+    assert request.response.status == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert response == {
         "message": "Token request failed",
         "errors": [
@@ -443,7 +444,7 @@ def test_mint_token_pending_publisher_project_already_exists(db_request):
     resp = views.mint_token(
         oidc_service, DUMMY_GITHUB_OIDC_JWT, claims["iss"], db_request
     )
-    assert db_request.response.status_code == 422
+    assert db_request.response.status_code == http.HTTPStatus.UNPROCESSABLE_ENTITY
     assert resp == {
         "message": "Token request failed",
         "errors": [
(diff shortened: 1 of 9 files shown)


warehouse/oidc/views.py
```diff
@@ -435,3 +442,62 @@ def should_send_environment_warning_email(
...
     return publisher.environment == "" and claims_env is not None and claims_env != ""
+
+
+class BurnPayload(BaseModel):
+    token: StrictStr
+
+
+@view_config(
+    route_name="oidc.burn_token",
+    require_methods=["POST"],
```
woodruffw, line 454: Noting: I did this as `POST`, but `DELETE` is arguably more semantically appropriate? If we do it that way, I'll probably need to rework this view to accept the macaroon via our normal Authn/z stack rather than as a body payload, since `DELETE` isn't well-defined with bodies.

warehouse/oidc/views.py
```diff
@@ -435,3 +442,62 @@ def should_send_environment_warning_email(
...
+    """
+
+    metrics = request.find_service(IMetricsService, context=None)
+    metrics.increment("warehouse.oidc.burn_oidc_issued_token.attempt")
+
+    try:
+        payload = BurnPayload.model_validate_json(request.body)
+        unverified_macaroon = payload.token
+    except ValidationError:
+        return _accepted(request=request)
```
ewdurbin, line 478: add metric.increment for `warehouse.oidc.burn_oidc_issued_token.failure` with tag `failure_reason:invalid_payload`?

warehouse/oidc/views.py
```diff
@@ -435,3 +442,62 @@ def should_send_environment_warning_email(
...
+
+    macaroon_service: DatabaseMacaroonService = request.find_service(
+        IMacaroonService, context=None
+    )
+
+    try:
+        macaroon: Macaroon = macaroon_service.find_from_raw(unverified_macaroon)
+    except InvalidMacaroonError:
+        return _accepted(request=request)
```
ewdurbin, line 497: add metric.increment for `warehouse.oidc.burn_oidc_issued_token.failure` with tag `failure_reason:invalid_macaroon`?

warehouse/oidc/views.py
```diff
@@ -435,3 +442,62 @@ def should_send_environment_warning_email(
...
+
+    if macaroon.oidc_publisher is None:
+        # This macaroon was issued to a principal other than a Trusted Publisher,
+        # which means this endpoint can't burn it.
+        # NOTE: mypy can't see that `user` and `oidc_publisher` are disjoint.
+        username = macaroon.user.username  # type: ignore[union-attr]
+        sentry_sdk.capture_message(
+            f"Tried to burn an API token corresponding to a user: {username!r}"
+        )
+        return _accepted(request=request)
```
ewdurbin, line 511: add metric.increment for `warehouse.oidc.burn_oidc_issued_token.failure` with tag `failure_reason:not_oidc_publisher`?

ewdurbin (approved):
i'm +1 to this feature and fairly happy with this implementation being limited to only valid trusted publishing issued macaroons.

while reporting to sentry is a good signal of misuse, i would like to see failures recorded as a metric as well to watch for broader patterns over time.

calling this a "provisional" approval as i'd like input from other admins, specifically @miketheman before merge.

warehouse/oidc/views.py
woodruffw: Sounds good! I was thinking failure could be subtractive (i.e. `attempt` minus `success`) but having it be explicit seems good too.

warehouse/oidc/views.py
```diff
@@ -79,6 +80,12 @@ def _invalid(errors: list[Error], request: Request) -> JsonResponse:
     }
 
 
+def _accepted(request: Request) -> JsonResponse:
+    request.response.status = 202
```
miketheman: Prefer semantic names over numbers
```suggestion
    request.response.status = http.HTTPStatus.ACCEPTED
```

warehouse/oidc/views.py
```diff
@@ -79,6 +80,12 @@ def _invalid(errors: list[Error], request: Request) -> JsonResponse:
     }
 
 
+def _accepted(request: Request) -> JsonResponse:
```
miketheman, line 84: I like the reuse, but this also feels like we should be subclassing `pyramid.httpexceptions.HTTPAccepted` here somehow instead? Then the return call looks like `return TokenBurnAccepted()` or something like that?

warehouse/oidc/views.py
miketheman: Yeah, this is a bit of a challenge, isn't it. Putting on a REST Purist hat for a moment, you're not actually asking the endpoint to "delete this resource" - since then you'd have a path with the resource you're asking to delete - aka `DELETE /_/oidc/tokens/d4d8a706-51a1-4844-a49f-12d4bc3db6f9` (or whatever path...) - which you'd have to deserialize from the raw token, which is possible via `pypitoken.Token.load(...).identifier` - maybe that's actually a better approach? because then, we're actively using the token as auth, and the token id as a resource? I'm not certain about that yet - but wanted to call it out as a maybe different approach.

warehouse/oidc/views.py
```diff
@@ -435,3 +442,74 @@ def should_send_environment_warning_email(
...
+
+    Downstream integrators can call this endpoint to expedite the invalidation of a
+    Trusted Publishing-issued token.
+
+    Unlike our other Trusted Publishing APIs, this API only ever returns an HTTP 202
+    (indicating receipt, but not communicating the outcome).
+    """
+
+    metrics = request.find_service(IMetricsService, context=None)
+    metrics.increment("warehouse.oidc.burn_oidc_issued_token.attempt")
```
miketheman: Prefer using `request.metrics.increment(...` instead of needing to import and find as it's already on `request` object.

warehouse/oidc/views.py
```diff
@@ -435,3 +442,74 @@ def should_send_environment_warning_email(
...
+
+    try:
+        payload = BurnPayload.model_validate_json(request.body)
```
miketheman, line 471: If I understand this correctly, this is creating a pydantic model to verify that the inbound is a str? Is this machinery worth it to validate? The single test doesn't help me understand what an invalid str would be.

warehouse/oidc/views.py
```diff
@@ -435,3 +442,74 @@ def should_send_environment_warning_email(
...
+        )
+        metrics.increment(
+            "warehouse.oidc.burn_oidc_issued_token.failure",
+            tags=["failure_reason:not_oidc_publisher"],
+        )
+        return _accepted(request=request)
+
+    macaroon_service.delete_macaroon(macaroon.id)
+    metrics.increment(
+        "warehouse.oidc.burn_oidc_issued_token.success",
```
miketheman: Thinking about metrics and naming, I'm thinking it may be beneficial to collapse all statuses under the same metric name, and disambiguate them by tags, (aka `status:success`, `status:failure` etc.). That way, we can plot "the thing" and use tags to split them out, instead of having to plot .attempt vs .<every other status> to determine if we're counting correctly.

warehouse/oidc/views.py
```diff
@@ -435,3 +442,74 @@ def should_send_environment_warning_email(
...
+    except InvalidMacaroonError:
+        metrics.increment(
+            "warehouse.oidc.burn_oidc_issued_token.failure",
+            tags=["failure_reason:invalid_macaroon"],
+        )
+        return _accepted(request=request)
+
+    if macaroon.oidc_publisher is None:
+        # This macaroon was issued to a principal other than a Trusted Publisher,
+        # which means this endpoint can't burn it.
```
miketheman, line 501: If restricting to only TP issued, does that change the semantics of the endpoint/route name? The view function is specific, the route is not.

warehouse/oidc/__init__.py
```diff
@@ -76,6 +76,7 @@ def includeme(config: Configurator) -> None:
     # NOTE: This is a legacy route for the above. Pyramid requires route
     # names to be unique, so we can't deduplicate it.
     config.add_route("oidc.github.mint_token", "/_/oidc/github/mint-token", domain=auth)
+    config.add_route("oidc.burn_token", "/_/oidc/burn-token", domain=auth)
```
miketheman, line 79: I think the `/_/` prefix has thus far been reserved for system-to-system integrations, or internal behaviors like client-side includes. Considering this is an endpoint that would be readily available to external tools, not our own website or integrations we've carefully selected, is this the right place for it, or would `/manage/account/token/` or something closer to that be more appropriate, since this is no longer system-to-system, rather tool-to-system?

miketheman (changes requested):
Interesting idea, plenty of comments inline.
I like the 202 response for anything, yay!

There's no rate limit protections I can see, that's probably something good to add - and since we expect it to be TP-centric, not by IP, rather by token identity? Not sure about that yet, but there's nothing protecting from a DoS run on this endpoint right now. The integration one has verification of the submitter identity before we hit the DB for token lookups.

There's no Event emitted for the token removal, which there would be for a regular token. I realize that our "automatic reaper" doesn't emit those, rather relies on the created time to emit "Expired" vs actually knowing that it's gone. That's an oddball case that we may want to add into this scope, since we're not relying on the passage of time and a task to run successfully, we're taking an actual action.

But generally moving in a positive direction!

warehouse/oidc/__init__.py
woodruffw: I think the rationale for putting TP APIs under `/_/` is that we've historically classified them as system-to-system integrations. It's definitely a fuzzy category though, since in the case of TP it's "system-to-system, mediated by a client on one side of the other system."

My 0.02c would be for keeping it here, if only because all of the other TP APIs are already under `/_/` and it would be odd/inconsistent to have just one that isn't. If [PEP 807](https://peps.python.org/pep-0807/) ever gets pushed forwards then moving all of these off of `/_/` would be pretty easy.

warehouse/oidc/views.py
woodruffw: This was intended to match the existing pattern in this file, i.e. we have `_invalid` for returning a 422.

(IIRC, the reason we did it this way originally is because the `view_config` sets `renderer="json"`, so we need to plumb into `request.response` to get the body to encode correctly.)

warehouse/oidc/views.py
woodruffw: Yeah, I think the big problem there is that it requires TP clients to actually peek into the macaroon, which none of them currently do. I guess they could, but that seems like a more troublesome long-tail change (for twine, gh-action-pypi-publish, pdm, uv, etc) than just a single new request (for the Python ones because pypitoken would be a new dep, for uv because we'd need to come up with an equivalent to pypitoken).

I think my preference here would be to either keep it this way _or_ do a slightly less idiomatic `DELETE`, e.g. `DELETE /_/oidc/tokens` with a `Bearer` header. The latter feels pretty bad to me as well though 🙂

warehouse/oidc/views.py
woodruffw: Yep, exactly. We already do that for `TokenPayload` above, which happens to have the exact same shape (I'm not reusing it because that's just a coincidence, although we _could_ reuse it until they diverge).

(My 0.02c is that this is worth it, since the open-coded version of pulling the JSON body, checking for a key, checking that it's a string, etc. would be longer. I can certainly add some tests demonstrating the value of `StrictStr` though.)

warehouse/oidc/views.py
woodruffw: Hmm, I might be missing what you mean? `/_/oidc/...` is our prefix for TP routes, and the endpoint is `burn_oidc_issued_token`.

(I think the generic version of this route would be `/_/burn-token` or `/api/burn-token` or similar.)

woodruffw:
> not by IP, rather by token identity? Not sure about that yet, but there's nothing protecting from a DoS run on this endpoint right now. The integration one has verification of the submitter identity before we hit the DB for token lookups.

Hmm, are you imagining something like a ratelimit key that's `hash(macaroon)`? That seems doable to me!


> There's no Event emitted for the token removal, which there would be for a regular token. I realize that our "automatic reaper" doesn't emit those, rather relies on the created time to emit "Expired" vs actually knowing that it's gone. That's an oddball case that we may want to add into this scope, since we're not relying on the passage of time and a task to run successfully, we're taking an actual action.

Yeah, this is kind of subtle: I think there's a chance this will be pretty noisy, since (in the happy case) there's a direct 1-1 correspondence between a TP token being issued and being "burned" before it would naturally expire. So we'd effectively double the number of TP-associated events in a user's security feed for something that's already happening in the background implicitly.

My weak preference would be for deferring that, but I'm curious what you (and @ewdurbin) think. I think a cool longer-term thing we could do is "group" events properly to reduce the visual noise, which would allow us to emit these without spamming users.

ewdurbin:
> > There's no Event emitted for the token removal, which there would be for a regular token. I realize that our "automatic reaper" doesn't emit those, rather relies on the created time to emit "Expired" vs actually knowing that it's gone. That's an oddball case that we may want to add into this scope, since we're not relying on the passage of time and a task to run successfully, we're taking an actual action.
> 
> Yeah, this is kind of subtle: I think there's a chance this will be pretty noisy, since (in the happy case) there's a direct 1-1 correspondence between a TP token being issued and being "burned" before it would naturally expire. So we'd effectively double the number of TP-associated events in a user's security feed for something that's already happening in the background implicitly.
> 
> My weak preference would be for deferring that, but I'm curious what you (and @ewdurbin) think. I think a cool longer-term thing we could do is "group" events properly to reduce the visual noise, which would allow us to emit these without spamming users.

Since this is a new type of interaction/event, I think I am actually in the camp of emitting rather than suppressing. We have a lot of control over how events are surfaced to users which means that improvements that allow for filtering//suppressing serverside or clientside are probably preferable over dropping a potentially interesting event on the floor.

woodruffw:
Sounds good, I'll work on adding the appropriate event for it.

warehouse/templates/manage/project/history.html
```diff
@@ -224,6 +224,8 @@ <h2>{% trans %}Security history{% endtrans %}</h2>
       <small>{% trans %}Workflow:{% endtrans %} {{ event.additional.workflow }}</small>
     {% endif %}
   </small>
+{% elif event.tag == EventTag.Project.ShortLivedAPITokenRevoked %}
+  <strong>{% trans %}Short-lived API token revoked{% endtrans %}</strong>
```
woodruffw, line 228: NB: It _might_ make sense to just not render this event at the moment, since it's kind of confusing in the current layout: it corresponds to an automatic action (e.g. `gh-action-pypi-publish` proactively revoking a credential instead of letting it passively expire) and doesn't currently contain a lot of useful end-user context.

In the future, we could refactor the security history a bit to make this more interesting: for example we could allow events to be nested, so that the user would see something like:

```
+------------------------------------+
|  Short-lived token added           |
|  Expires at: ...                   |
|  Proactively revoked at: ...       |
+------------------------------------+
```

...but this would (1) require us to track the macaroon IDs in token events (we don't currently) and (2) would require us to join/aggregate individual events together 🙂

warehouse/oidc/views.py
miketheman: Were any tests added to demonstrate the value of StrictStr?

tests/unit/oidc/test_views.py
```diff
@@ -1115,6 +1116,149 @@ def test_should_send_environment_warning_email(
...
+        pretend.call(
+            "warehouse.oidc.burn_oidc_issued_token",
+            tags=["status:failure", "failure_reason:invalid_macaroon"],
+        ),
+    ]
+
+
+def test_burn_oidc_issued_token_user_macaroon(metrics, monkeypatch):
+    macaroon = pretend.stub(
+        id="fake-macaroon-id",
```
miketheman: A real `delete_macaroon` call will fail open with a string input.

https://github.com/pypi/warehouse/blob/e531d366e822259ca391972a0dc939265a65db8b/warehouse/macaroons/services.py#L102-L110

Here's what I [hate about pretend](https://github.com/pypi/warehouse/issues/18880) is that this isn't what a Macaroon id would be - it's a uuid, not a str - so this test is passing today, but would likely fail in reality.

This would have probably been exposed by a functional test or using [publisher factories](https://github.com/pypi/warehouse/blob/e531d366e822259ca391972a0dc939265a65db8b/tests/common/db/oidc.py#L27)/[service fixtures](https://github.com/pypi/warehouse/blob/e531d366e822259ca391972a0dc939265a65db8b/tests/conftest.py#L558)?

I suspect [more type hints](https://github.com/pypi/warehouse/issues/16785) around these interfaces and methods would have helped identify this as well.

(same comment applies to other test cases here using strings as ID)

warehouse/oidc/views.py
```diff
@@ -435,3 +443,81 @@ def should_send_environment_warning_email(
...
+            tags=["status:failure", "failure_reason:invalid_payload"],
+        )
+        return _accepted(request=request)
+
+    macaroon_service: DatabaseMacaroonService = request.find_service(
+        IMacaroonService, context=None
+    )
+
+    try:
+        macaroon: Macaroon = macaroon_service.find_from_raw(unverified_macaroon)
```
miketheman: find_from_raw() only resolves the database identifier; it does not authenticate the macaroon.

Should we call `macaroon_service.verify()` here to ensure that this is signed by us? A forged macaroon carrying a known UUID burns the real token, and since the identifier is str(dm.id) in the clear, that's a non-zero possibility.

The stated "valid TP token" restriction needs to verify validity.

warehouse/oidc/views.py
```diff
@@ -435,3 +443,81 @@ def should_send_environment_warning_email(
...
+class BurnPayload(BaseModel):
+    token: StrictStr
+
+
+@view_config(
+    route_name="oidc.burn_token",
+    require_methods=["POST"],
+    renderer="json",
+    require_csrf=False,
+)
```
miketheman, line 457: At the very least, should we add `accept="application/json"` here to prevent any other weird submissions?

miketheman (changes requested):
Thanks-  many of the previous reviews have been addressed, but there's some new things that should be resolved first.
Let me know if you disagree!

warehouse/oidc/views.py
woodruffw: > Should we call `macaroon_service.verify()` here to ensure that this is signed by us? A forged macaroon carrying a known UUID burns the real token, and since the identifier is str(dm.id) in the clear, that's a non-zero possibility.

I think this is pretty niche in practice, since someone would need to get the real (i.e. sensitive) macaroon in order to discover its UUID. But yeah, it's unintentional, I'll change to validating.

woodruffw: Ah, I remmeber why I didn't do this -- our `macaroon_service.verify` also checks the caveats, which don't apply here. I think I need to add `verify_signature_only` (which is what we need going forwards anyways, for opaque scoped).

woodruffw: Okay, added `verify_signature_only` for this.

warehouse/oidc/views.py
woodruffw: Mind if we fast-follow for that? I think we should do this consistently for all of the OIDC API routes, in which case it'd sprawl the diff a bit to do it here.

warehouse/oidc/views.py
woodruffw: I'll add some. 

(We already demonstrate the value in `test_mint_token_from_oidc_invalid_payload` and elsewhere, but no harm I guess in showing it in other endpoints.)

woodruffw: 77825fb82

tests/unit/oidc/test_views.py
woodruffw: Thanks! This caught what would have been a hard failure in burning: 85eb6aac6

(I looked, and the services/ifaces are pretty well typed here already, but mypy locally doesn't seem to propagate the type that's hinted for `find_service`. I think that would have caught this.)

miketheman (approved):
There's still opportunities here like a functional test for the happy path.

@woodruffw please keep a close eye on this as it rolls out.

PR #20207, opened 2026-06-24
Schedule OpenSearch index cleanup

JacobCoffee:
Schedule deletion task for OpenSearch index cleanup so manual runs don't have to be done


warehouse/search/tasks.py
```diff
@@ -204,14 +220,42 @@ def reindex(self, request):
...
+                    _delete_index(client, index_to_delete)
             else:
                 client.indices.put_alias(name=index_base, index=new_index_name)
     except redis.exceptions.LockError as exc:
         sentry_sdk.capture_exception(exc)
         raise self.retry(countdown=60, exc=exc)
 
 
+@tasks.task(bind=True, ignore_result=True, acks_late=True)
+def delete_older_indices(self, request: Request) -> None:
```
JacobCoffee, line 284: should probs pull out the CLI into a helper and just let this and the CLI call that since we know the CLI works ✨ flawlessly ✨ since i have to use it every other day at 3am 😭

JacobCoffee: [afb5598](https://github.com/pypi/warehouse/pull/20207/commits/afb5598042f6ae79f0a216379fe1df3ce6d80f20)

JacobCoffee:
this has been suuuper quiet lately so I think the index cleanup is working well enough. if i see a page again for it ill re-open

PR #20208, opened 2026-06-24
Add missing burn permission to ACL

JacobCoffee:
Missed in https://github.com/pypi/warehouse/pull/20195


JacobCoffee:
Keeping the two ACLs in sync is very weird to me can we fix that?

ee-oai (approved):

ewdurbin (approved):

PR #20211, opened 2026-06-25
fix(admin): only select the reported version

miketheman:
Previously would use substring select, so multiple versions could be selected by default if the matched, and if the list was long enough, one might be missed and then inadvertently removed.


JacobCoffee (approved):
jinja squiggly semantics 😵‍💫

miketheman:
> jinja squiggly semantics 😵‍💫

cat-wiggle

PR #20218, opened 2026-06-26
perf(manage): scope project history file events via FK join

miketheman:
Filtering file events on the JSONB `additional['project_id']` key forced a full seq scan of `file_events` on every project history page load. As more records are added, the page load time increases.

Avg Duration (1 year)

Join through `release_files` -> `releases` and filter on `Release.project_id`, which uses existing indexes.

Also switch `union` to `union_all`, since project and file events are never duplicate rows.

Refs: #13205


JacobCoffee (approved):

tests/unit/manage/test_views.py
```diff
@@ -6060,9 +6061,10 @@ def test_first_page(self, db_request, user_service):
         file_events_query = (
             db_request.db.query(File.Event)
             .join(File.Event.source)
-            .filter(File.Event.additional["project_id"].astext == str(project.id))
+            .join(File.release)
+            .filter(Release.project_id == project.id)
```
cofin, line 6065: These are definitely good changes.  I'm guessing this can now leverage the index properly.

cofin (approved):
Sorry for the drive by review! But for what it's worth, this looks good to me.

tests/unit/manage/test_views.py
miketheman: Thanks - that's the intention, for sure.

miketheman:
Since we like pretty pictures:

Latency on service warehouse, resource GET manage project history in env prod

We're in the sub-second range now. Zoomed in top post-change:

Latency on service warehouse, resource GET manage project history in env prod (1)

PR #20221, opened 2026-06-29
docs: update TP and add dev doc for account associations

miketheman:


docs/dev/development/account-associations.md
```diff
@@ -0,0 +1,94 @@
...
+But `create_service` reads `settings["gitlab.oauth.client_id"]`,
+so the first time a user clicks "Connect", the request 500s with `KeyError: 'gitlab.oauth.client_id'`.
+Always pass both kwargs with the real client.
+This applies to every provider.
+
+## Adding a new provider
+
+Follow the `GitLabOAuthClient` pattern:
+
+1. Add the client class (and a `Null*OAuthClient`) in
```
JacobCoffee:
```
1. Add the client class (and a `Null*OAuthClient`) in
```
i think the github client falls out of convention here? (*AppClient instead)

JacobCoffee (approved):
just 1 question if we want/need to change the naming convention on our GH client calss

docs/dev/development/account-associations.md
JacobCoffee: maybe i am cionfusing the real client <-> null client, though.

miketheman: Good callout - and that's actually a small distinction in the app protocols. GitHub has both GitHub OAuth Apps, as well as GitHub Apps (nice and confusing!).
We're using the GitHub App, which implements the `IOAuthProviderService` interface, so I'll update the naming here to `Null*Client`

miketheman:
```suggestion
1. Add the client class (and a `Null*Client`) in
```

JacobCoffee (approved):

PR #20234, opened 2026-06-30
Allow org deletes if empty and not subscribed

### LINKED ISSUE #20232: Allow organization owners to delete their organizations if billing not active

Currently, owners of an empty organization cannot delete their organization if not in good standing (billing not activated is one case of this, and the one this issue is pinpointing)

Allow option to delete an organization here or figure out a way to get them into the org settings page with limited options (either activate billing or delete org)

Image

**Additional context**

JacobCoffee:
Company organizations can't currently delete their org if they have not yet opted in to billing (a new requirement based on https://github.com/pypi/warehouse/pull/20161) 

This PR adds the ability to manage parts of the organization, where before the `Manage` button was hidden until billing was activated.

image

If a user tries to go to any of the critical parts of the org (teams, projects, members) they are redirected with a flash saying please activate billing first

image

Allows self-serving so that support is not hammered with requests.

Resolves #20232


warehouse/manage/views/organizations.py
```diff
@@ -320,7 +320,7 @@ def manage_organization_application_submit(self):
     context=Organization,
     renderer="warehouse:templates/manage/organization/settings.html",
     uses_session=True,
-    require_active_organization=True,
+    require_active_organization=False,
```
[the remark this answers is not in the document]
JacobCoffee: that's kindve the trade-off here, though..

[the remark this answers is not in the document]
JacobCoffee: i think the allowed items (namely editing display name, url and description) are fine to leave open to avoid having a bigger diff. Cant change org type or org official name

warehouse/predicates.py
```diff
@@ -100,6 +100,12 @@ def __call__(self, context: Organization | Team, request):
 
         if organization.is_in_good_standing():
             return True
+        request.session.flash(
```
ewdurbin, line 103: great addition along the way!

warehouse/predicates.py
```diff
@@ -100,6 +100,12 @@ def __call__(self, context: Organization | Team, request):
...
+            "This organization's billing is inactive. Activate billing to "
+            "manage its projects, teams, and members.",
```
miketheman, line 105: Curious - how is this related to allowing deletion?

ewdurbin (approved):
In testing, I noticed that I could still reach `/manage/organization/ORG/publishing/`, which I think should likely also be gated behind good standing. Once that's in place, I think this is GTG.

warehouse/manage/views/organizations.py
miketheman: suggestion: Convention is currently to add a comment after the `False` to explain why an active org is not required

warehouse/templates/manage/organizations.html
```diff
@@ -112,6 +112,12 @@ <h3 class="organization-snippet__title">
                    title="{% trans %}Activate billing for this organization{% endtrans %}">
                   {% trans %}Activate Billing{% endtrans %}
                 </a>
+                {% if organization.is_active and organization.name in organizations_owned %}
+                  {# Owners can still manage (e.g. delete) an org with inactive billing #}
+                  <a class="button"
+                     href="{{ request.route_path('manage.organization.settings', organization_name=organization.normalized_name) }}"
+                     title="{% trans %}Manage this organization{% endtrans %}">{% trans %}Manage{% endtrans %}</a>
+                {% endif %}
```
miketheman: This block appears to be duplicating the same block later on (`{# Show "Manage" button for managers and owners #}`) - do the conditions need to be re-sorted or something else?

miketheman (commented):
Generally looking good.
Some inline questions and suggestions

miketheman:
Some more thoughts:

- What happens if an org has a Stripe cancelled billing? 
- There's no test that POSTs `confirm_organization_name` and asserts the org is deleted while billing-inactive, none for the canceled-subscription case, and no negative test that billing managers/members can't reach the delete

warehouse/predicates.py
JacobCoffee: To delete you need to go into `/manage/organization/$orrg/settings/` and we expose the other menu items so this is just to say hey you cant do that either activate billing or delete your org

warehouse/templates/manage/organizations.html
JacobCoffee: [f01131e](https://github.com/pypi/warehouse/pull/20234/commits/f01131e6257d3be7296a185d3634a0916619a5fc)

JacobCoffee:
> What happens if an org has a Stripe cancelled billing?

we still receive the webhook from the match statement so if cancelled itll show activate billing re-shows and the delete button is still there after it has been webhook fires (may be delayed a little, and i think its at period-end so theylls till be considered active until then)

PR #20236, opened 2026-06-30
fix(admin): accept multiple selected releases in malware reports

### LINKED ISSUE #20235: Admin: Multiple releases cannot be selected in a Malware Report for quarantine or removal
**Describe the bug**

Path: admin/malware_reports/<report_id>/

Currently, all warehouse admin pages reject duplicate POST keys via the `reject_duplicate_post_keys_view`, with one legacy exception.

When attempting to select and submit multiple releases to either quarantine or remove in the Detail View for a Malware Report, the submission will return an error:

```
400 Bad Request
The server could not comply with the request since it is either malformed or otherwise incorrect.

POST body may not contain duplicate keys (URL: '<host>/admin/malware_reports/<report-id>/quarantine_release/')
```

**Expected behavior**
On the malware report page, it should be possible to select multiple releases to be marked for quarantine or removal.

**To Reproduce**
1. Login to an account that's marked as an observer
1. Create a malware report for <project-name> at project/<project-name>/submit-malware-report/
1. From admin/malware_reports/ click through to the malware report page for the project

Issue 1: Failure to quarantine multiple releases:

1. Click on Quarantine Release
2. Select multiple releases
3. Submit
4. Results in a 400 error

Issue 2: Failure to remove multiple releases:
1. Click on Remove Release
2. Select multiple releases, enter a reason, confirm by typing the project name
3. Submit
4. Results in a 400 error

**My Platform**

Verified in a local development environment.

**Additional context**

nnja:
Fixes #20235 for both release quarantine and release deletion, taking a
slightly different approach than #20222.

Now `permit_duplicate_post_keys` takes a set of key names, so the two verdict views only allow dupes for `versions` instead of opting out of the check.

Also made a few fixes based on feedback:
 - the 400 tells you which keys are duplicated, and passing in anything that's not `True` or a collection of strings to `permit_duplicate_post_keys` fails at config time
 - beefed up tests to check for that the DB reflects the expected state
 - put back `if len(keys) != len(set(keys))` as the top level check in `config.py` in case of unexpected performance issues

@miketheman let me know what you think. I'm looking for feedback and insight into what's preferred for Warehouse.


tests/functional/admin/test_malware_reports.py
```diff
@@ -128,3 +128,123 @@ def test_remove_release_accepts_multiple_versions(self, webtest):
...
+        success_message = flash_messages.html.find(
+            "span", {"class": "notification-bar__message"}
+        )
+        assert success_message is not None
+        assert "Removed release(s)" in success_message.text
+
+        # Only the selected releases are removed, the third is not.
+        assert first.version in success_message.text
+        assert second.version in success_message.text
+        assert third.version not in success_message.text
```
miketheman, line 146: Nice additional assertion, but I don't think this asserts anything from the DB-side of things, since the message reflects data from the POST. Maybe add a DB query to assert that 1,2 versions are removed and 3 is not?

sethmlarson (approved):
Seems good to me, thanks @nnja! I'll let @miketheman review too as primary user.

tests/functional/admin/test_malware_reports.py
```diff
@@ -118,8 +118,8 @@ def test_remove_release_accepts_multiple_versions(self, webtest):
 
         body = MultiDict()
         body.add("csrf_token", csrf_token)
-        body.add("versions", "0.0.4")
-        body.add("versions", "0.0.3")
+        body.add("versions", first.version)
+        body.add("versions", second.version)
```
miketheman, line 123: I like the reuse of the version string ID from the object, that's helpful.

The order was explicitly reversed so that the form exercised what is likely to happen from the UI. Swap these two lines should be enough.

warehouse/config.py
```diff
@@ -278,18 +279,24 @@ def from_base64_encoded_json(configuration):
...
+    # permit_duplicate_post_keys = True, or for an exception-only view.
+    permit = info.options.get("permit_duplicate_post_keys")
+    if permit is True or info.exception_only:
         return view
 
-    # If this isn't an exception or hasn't been permitted to have duplicate
-    # POST keys, wrap the view with a check
+    # Passing a collection to permit_duplicate_post_keys will allow
+    # those keys to be duplicated.
+    allowed_duplicate_post_keys = set(permit) if permit else set()
```
miketheman, line 303: This works today because the tests are behaving.
What if instead of `permit_duplicate_post_keys={"versions"}`, I accidentally set `permit_duplicate_post_keys="versions"` instead? Should we validate the shape of `permit` when it's not True?
Admittedly that's a dev/test time concern, but if incorrectly tested can lead to a 400 in production.

warehouse/config.py
```diff
@@ -278,18 +279,24 @@ def from_base64_encoded_json(configuration):
...
     def wrapped(context, request):
         if request.POST:
-            # Determine if there are any duplicate keys
-            keys = list(request.POST.keys())
-            if len(keys) != len(set(keys)):
+            counts = Counter(request.POST.keys())
+            duplicated_post_keys = {key for key, count in counts.items() if count > 1}
+            if duplicated_post_keys - allowed_duplicate_post_keys:
                 return HTTPBadRequest(
                     f"POST body may not contain duplicate keys (URL: {request.url!r})"
```
miketheman: I'm also realizing that the output does nothing to help user understand what the duplicate keys were - that's probably a good spot to improve.

tests/functional/admin/test_malware_reports.py
```diff
@@ -128,3 +128,123 @@ def test_remove_release_accepts_multiple_versions(self, webtest):
...
+        # permit_duplicate_post_keys should fail.
+        body.add("foo", "bar")
+        body.add("foo", "baz")
+
+        resp = webtest.post(
+            f"/admin/malware_reports/{report.id}/quarantine_release/",
+            params=body,
+            status=HTTPStatus.BAD_REQUEST,
+        )
+        assert "POST body may not contain duplicate keys" in resp.body.decode()
```
miketheman: Combined with the note out output, this doesn't assert which keys were duplicated that triggered the violation, so we can tighten that with better output.

warehouse/config.py
```diff
@@ -278,18 +279,24 @@ def from_base64_encoded_json(configuration):
...
 
+    # Wrap the view with a check that rejects any duplicate keys
+    # that aren't explicitly permitted.
     @functools.wraps(view)
     def wrapped(context, request):
         if request.POST:
-            # Determine if there are any duplicate keys
-            keys = list(request.POST.keys())
-            if len(keys) != len(set(keys)):
+            counts = Counter(request.POST.keys())
```
miketheman: This change now runs the code block for every POST that isn't an exception or explicitly permitted as True. I don't think this should be a material performance issue because Counter is pretty fast, but it'll slow down with more and more keys. I don't expect that to be an issue today, but worth noting.

miketheman (changes requested):
Thanks for working through this! I've commented inline - let me know if anything could use clarification.

Also needs a merge/rebase against `main`, we'll need that anyways before merging.

Once updated, feel free to click "Re-request review" to ping that this is ready for re-review.

tests/functional/admin/test_malware_reports.py
nnja: Good call. I originally left the DB assertions out since they were covered by `test_malware_reports.py`. But I agree that the functional tests should prove that the full flow works as expected. Both instances are updated.

tests/functional/admin/test_malware_reports.py
nnja: I didn't catch the order was reversed on purpose. Fixed, and added a comment for future context.

warehouse/config.py
nnja: Good catch. Added checks at config time that the value is either `True` or a collection of strings, and a unit test to validate. A malformed value will now throw a `ConfigurationError`.

warehouse/config.py
nnja: More descriptive error messages are definitely more user friendly. This now returns: `POST body may not contain duplicate keys (duplicated: ['foo'], URL: '...')`.

tests/functional/admin/test_malware_reports.py
nnja: Added validation that the incorrectly duplicated key is flagged, while the allowed key is not.

warehouse/config.py
nnja: Good call out. `Counter` should be fast, but since this touches every POST there's no harm in having an additional check. Now the `len(keys) != len(set(keys)):` check is at the top level, so performance should not be affected. And the `Counter` check is only used when duplicates are found.

nnja:
@miketheman Good catches! I think I addressed all of your feedback, but let me know if I missed anything.

di (approved):
LGTM once merge conflicts are resolved!

nnja:
@di I fixed the merge conflicts, and updated my tests to use the new helpers introduced in #20446. Let me know if any other changes or updates are needed.

warehouse/config.py
```diff
@@ -282,21 +284,44 @@ def from_base64_encoded_json(configuration):
...
+    # misconfigured.
+    if permit is not None and (
+        isinstance(permit, str)
+        or not isinstance(permit, Collection)
+        or not all(isinstance(key, str) for key in permit)
+    ):
+        raise ConfigurationError(
+            "permit_duplicate_post_keys must be True or a collection "
+            f"of POST key names, got: {permit!r}"
+        )
```
miketheman: **issue:** `permit_duplicate_post_keys=False` now raises `ConfigurationError` at app boot.
`False` isn't `None`, isn't a `str`, and `bool` isn't a Collection, so it falls through to the `raise`.

**suggestion:** `if permit and (...)` fixes it, and then the `if permit else set()` on L306 becomes redundant

tests/functional/admin/test_malware_reports.py
```diff
@@ -91,3 +93,139 @@ def test_remove_release_accepts_multiple_versions(self, webtest, login_admin):
...
+            related=project,
+            observer=observer,
+            kind="is_malware",
+            payload={"inspector_url": ""},
+        )
+
+        detail_page = webtest.get(
+            f"/admin/malware_reports/{report.id}/", status=HTTPStatus.OK
+        )
+        csrf_token = detail_page.forms["quarantine"]["csrf_token"].value
```
miketheman, line 141: question: Shouldn't this be picking the CSRF from the `quarantine-release` form instead? In practice these are usually the same value, but just in case the form moves or some other change moves things, I wouldn't expect this test to fail as a result.

```suggestion
        csrf_token = detail_page.forms["quarantine-release"]["csrf_token"].value
```

tests/functional/admin/test_malware_reports.py
```diff
@@ -91,3 +93,139 @@ def test_remove_release_accepts_multiple_versions(self, webtest, login_admin):
...
+            f"/admin/malware_reports/{report.id}/quarantine_release/",
+            params=body,
+            status=HTTPStatus.BAD_REQUEST,
+        )
+        body_text = resp.body.decode()
+        assert "POST body may not contain duplicate keys" in body_text
+
+        # The error names only the duplicated key not in the permitted list.
+        assert "duplicated: ['foo']" in body_text
+        assert "duplicated: ['versions']" not in body_text
```
miketheman, line 231: **suggestion:** change the assertion to match what would be emitted if the allowed condition was `permit_duplicate_post_keys={"foo", "versions"}`, otherwise this assertion would never be true.

```suggestion
        assert "duplicated: ['foo', 'versions']" not in body_text
```

warehouse/config.py
```diff
@@ -282,21 +284,44 @@ def from_base64_encoded_json(configuration):
...
+
+    allowed_duplicate_post_keys = set(permit) if permit else set()
```
miketheman, line 306: **issue:** `permit_duplicate_post_keys=False` now raises `ConfigurationError` at app boot.
`False` isn't `None`, isn't a `str`, and `bool` isn't a Collection, so it falls through to the `raise`.

**suggestion:** `if permit and (...)` fixes it, and then the `if permit else set()` on L306 becomes redundant

miketheman (changes requested):
Thanks for taking another pass, I did another in-depth review of the changes and found some questions and issues. We're pretty close now!

PR #20242, opened 2026-07-02
feat(manage): segment archived projects on the projects page

miketheman:
Active projects were getting lost in a flat list mixed with old, archived ones. Archived projects now live in their own ~~collapsed~~ section at the bottom.


miketheman:
older pics

<details>

## Default

Screenshot 2026-07-02 at 16 37 30


## Opened

image

</details>

image

warehouse/templates/manage/projects.html
```diff
@@ -53,78 +109,37 @@ <h3 class="package-snippet__title">
...
+        <div class="callout-block no-top-margin">
+          <p>
+            {% trans href='https://packaging.python.org/' %}You have not uploaded any projects to PyPI, yet. To learn how to get started, visit the <a href="{{ href }}" target="_blank" rel="noopener">Python Packaging User Guide</a>{% endtrans %}
+          </p>
+        </div>
+      {% endif %}
     {% endfor %}
   </div>
+  {% if projects_archived %}
+    <details class="callout-block"
```
ewdurbin: Question: Is there anyway to make this default to _not_ collapsed? My concern is that `Ctl+f` searching on page will not surface projects under the details fold.

miketheman: In an earlier version of this I had it not even bounded by a box, only a section with a header and a list like above. I can absolutely revert to that approach. Part of me was thinking about folks with long project pages, but I'll let that become a user driven request instead.

miketheman: See f549822fa62e9ff0fa99b5632fbec3e27afd2caa

ewdurbin (approved):

woodruffw (approved):
Noice

watkinsoley007-star (approved):

PR #20257, opened 2026-07-09
migrate to mailpit for local development SMTP

ewdurbin:
[Mailpit](https://mailpit.axllent.org/) is a bit more feature rich than maildev in my experience.

This also adds persistence across container restarts while I'm futzing.

I chose to map the web UI to the same port that maildev used for muscle memory purposes


miketheman (commented):
Seems generally reasonable, but I'm curious - mailpit is owned by a single maintainer, whereas maildev is owned by an org - any thoughts on long-term maintainability here?

ewdurbin:
can always go back to maildev i guess. i also haven't evaluated an updated version of maildev, we're pretty far behind on version :)

miketheman:
> can always go back to maildev i guess. i also haven't evaluated an updated version of maildev, we're pretty far behind on version :)

We're on the most recent release of maildev 2.2.1 as of #18200 
3.0.0 is only a release candidate right now.

miketheman (approved):
Low risk enough, and we can swap back if needed in the future.

PR #20258, opened 2026-07-09
feat(admin): make journals pages usable via Tabulator

### LINKED ISSUE #14541: admin Journals page is no longer particularly useful
The page doesn't load due to the volume of paginations, and times out https://pypi.org/admin/journals/

Instead, provide a search interface that allows searching on well-indexed columns (e.g. submitted by, project name) so admins can perform journal lookups without needing to look at alternate solutions.

miketheman:
The journals page ran a `count(*)` over the whole journals table (hundreds of millions of rows) on every load and timed out before rendering anything.

Replace the global and per-project journals pages with a Tabulator table fed by a shared JSON endpoint that never counts:

- fetch one row past the page size to detect a next page, and cap pagination depth at 10,000 rows
- size unfiltered pagination from the pg_class.reltuples estimate
- filter only on indexed columns: name and version exact, submitted_by exact (CITEXT, so case-insensitive), action by prefix, and a date filter for entries on or before a given day
- sort only on indexed columns (NULLs last), newest first by default
- run page queries under a 10s statement timeout so an unindexable filter degrades to a clear error instead of a hung worker

Filters can be pre-set from the query string, so the "view all journals" button on user detail pages deep-links to a filtered table, and the per-project page pins its name filter onto the same endpoint.

Tabulator is the jQuery-free table library AdminLTE 4 adopted after dropping DataTables, so this is also a first incremental step in that direction.
Refs: https://adminlte.io/themes/v4/docs/integrations.html
Refs: https://www.tabulator.info/

Closes #14541

### DIFF
--- a/warehouse/admin/views/projects.py
+++ b/warehouse/admin/views/projects.py
@@ -200,6 +200,18 @@ def project_detail(project, request):
         .order_by(JournalEntry.submitted_date.desc(), JournalEntry.id.desc())
         .limit(30)
     )
+    # The card lists only the newest entries; the badge shows the true
+    # count. A short first page already proves the total; otherwise
+    # count(*) is an index-only scan on journals_name_idx.
+    if len(journal) < 30:
+        journal_count = len(journal)
+    else:
+        journal_count = (
+            request.db.query(func.count())
+            .select_from(JournalEntry)
+            .filter(JournalEntry.name == project.name)
+            .scalar()
+        )
     observations = list(
         request.db.query(project.Observation)
         .options(joinedload(project.Observation.observer))
@@ -213,6 +225,7 @@ def project_detail(project, request):
         "releases": releases,
         "maintainers": maintainers,
         "journal": journal,
+        "journal_count": journal_count,
         "oidc_publishers": project.oidc_publishers,
         "ONE_MIB": ONE_MIB,
         "MAX_FILESIZE": MAX_FILESIZE,
@@ -534,7 +547,11 @@ def release_render(release, request):
     uses_session=True,
 )
 def journals_list(project, request):
-    q = request.params.get("q")
+    """Render the shell for a project-scoped journals Tabulator table.
+
+    The table itself is fed by the shared JSON endpoint behind
+    `admin.journals.list`, pre-filtered on this project's name.
+    """
     project_name = request.matchdict["project_name"]
 
     if project_name != project.normalized_name:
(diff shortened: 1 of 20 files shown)


miketheman:
A visual

journals-admin

warehouse/admin/views/projects.py
```diff
@@ -200,6 +200,18 @@ def project_detail(project, request):
         .order_by(JournalEntry.submitted_date.desc(), JournalEntry.id.desc())
         .limit(30)
     )
+    # The card lists only the newest entries; the badge shows the true
+    # count. A short first page already proves the total; otherwise
+    # count(*) is an index-only scan on journals_name_idx.
+    if len(journal) < 30:
+        journal_count = len(journal)
+    else:
```
JacobCoffee, line 208: do we want a timeout incase counts are giant

JacobCoffee (approved):
it would be nice to have a lower bound for dates so we could scope like `Jan 1-March 15` and a (porbbaly) bigger chnage to allow version scoping like `<1,>2` or similar but can add later

warehouse/admin/views/projects.py
miketheman: Excellent question! I ran a query on prod to figure this out:

```sql
warehouse=> \timing
Timing is on.
warehouse=> SELECT name, COUNT(*) AS journal_count
FROM journals
GROUP BY name
ORDER BY journal_count DESC
LIMIT 20;
           name           | journal_count 
--------------------------+---------------
 duckdb                   |         94133
 ccxt                     |         56506
 xgcondb                  |         51864
 ddtrace                  |         45217
 tf-nightly               |         39614
 pyagrum-nightly          |         37016
 lalsuite                 |         35584
 spanishconjugator        |         28131
 kuzu                     |         25460
 aim                      |         25310
 certora-cli-alpha-master |         23431
 tf-nightly-cpu           |         23164
 simple-equ               |         22253
 setuptools               |         18891
 coiled                   |         18104
 simplepyble              |         15231
 tensordict-nightly       |         14856
 fbgemm-gpu-nightly-cpu   |         13761
 Deliverance              |         13726
 tf-nightly-gpu           |         13567
(20 rows)

Time: 7205.786 ms (00:07.206)
```

So it takes ~7 seconds to get all, since there's an index on `journals.name` already.

To use the worst case here with the query proposed:

```sql
warehouse=> SELECT COUNT(*) FROM journals WHERE journals.name = 'duckdb';
 count 
-------
 94133
(1 row)

Time: 11.790 ms
```
A [detailed explain analyze visualization](https://explain.datadoghq.com/?dbms=postgres&plan=H4sIAAAAAAAAE-1WXW_aMBT9K5Zf1k4oCxAKibQH1q1qpa50gNSHGlXGuQ1eHTtynAGq-t8nO6RJIWjSnjapjznnnvvhm-ME3z8TfCuoJDh6JvhGxYDm2wwIjggeJ4mGhBoguEPwzGhqINk66lZQLh18S7XhVKDvKi5VM55mAiqOCgECjddUW_aRihw6BI_zrWTonGZ0KZr4zFBtigydq9wQHA37g17ohWGH4LkyVDTwvu_5fqfsHU3VOic46lbPdzw2K4KjkS3FTEEFqjLPeWoLdoee73drukzfJHs12Ui_Q66VyipoUpissF3dE8xUIc3Jx1OCF3aaFdUQo0tu0Beh2JNVhAM_rKkp0Ljm_Jr4yrXh0M7daW4MyLfctWJUvC1Vw4dlSry1Skm1FplDmrUkc3CrYNfx1adJKdsdsN3dHm3lsM-XrRxT12yr2HV1TPtKtkrtS2RHuD-wxJWMYYMmUmzRjNFXB4A0aAqCGq5kvuKZi50UBvRf-oBRabcDzGZ02S6UXlMdu3xlFzc0LZv6qQotqcgfJE3hgccbF1T1cxhX2pmtIKUOz4ql4MyhY8Fp3hL81pe-FwSHnuyNfO9ssOfJMBz2-_u-9I_60vf8bnDMlj2v3wsObBkGXVfhD9aU7hQWr6d3rmTsBj2pJvVsCPqMPsQFe4qXH6LIwMacloep1jmaQqp-QYyWW1TmmAJbAXvaTXQJNEMXYNgKXANno8Ho_Rb4T2-Bl8VLh-AfBegtuopBGv7IrZuj8CzoDcOwF4R-Pwj6g1H1dksuk_Ir2rZx_33d__S6G0tsEAMn0zxJQLsPgr0_vm2AFe5mbfwvdIcvC_wbeOKfFE4JAAA).

So I'm pretty confident that even the worst lookups will be pretty fast here

mhna55729-coder (approved):

PR #20260, opened 2026-07-09
Don't break --skip-existing for old releases

### LINKED ISSUE #20251: No-op reupload of old package now leads to "HTTPError: 400 Bad Request"
**Describe the bug**

In order to reduce the chances that our users will inadvertently install malicious software, Semgrep typosquats on a few similar names, e.g. `samgrep`. During our release process, we (re)publish such dummy packages with version 0.1.0. Up until today's release, re-publishing the old packages was a harmless no-op and the CI logs showed the following message:

```
WARNING  Skipping samgrep-0.1.0.tar.gz because it appears to already exist 
```

As of today, we instead get:

```
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
```

(note that we dry-run the release on test.pypi.org before publishing to the production instance).

**Expected behavior**

I expect the previous behavior where publishing an already-released version is a no-op rather than an error.

This seems likely to be related to #19727. My request would be to relax that check such that releases retain idempotence. It's very reasonable to prevent people from publishing new artifacts to old releases, but as implemented it also prevents use cases like I've described. Of course, we can adjust our release process to check if a package already exists before uploading, but it's a small amount of additional complexity that we would prefer to avoid.

**To Reproduce**

Most likely, this is reproducible by uploading a package, waiting 14 days, and then attempting to upload it again.

**My Platform**

GitHub Actions, Ubuntu 24.04.4 runner, uploading using `pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e`

**Additional context**

sethmlarson:
Closes #20251

### DIFF
--- a/warehouse/forklift/legacy.py
+++ b/warehouse/forklift/legacy.py
@@ -1193,23 +1193,6 @@ def file_upload(request):
                 + " for more information.",
             )
 
-        # Check to see if uploading this file would create a duplicate sdist
-        # for the current release.
-        if (
-            form.filetype.data == "sdist"
-            and request.db.query(
-                request.db.query(File)
-                .filter((File.release == release) & (File.packagetype == "sdist"))
-                .exists()
-            ).scalar()
-        ):
-            request.metrics.increment(
-                "warehouse.upload.failed", tags=["reason:duplicate-sdist"]
-            )
-            raise _exc_with_message(
-                HTTPBadRequest, "Only one sdist may be uploaded per release."
-            )
-
         # Check that the release is either new or that the release
         # is still within the window allowing new files to be published.
         # Note that this feature explicitly doesn't protect against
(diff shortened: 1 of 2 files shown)


JacobCoffee (approved):

warehouse/forklift/legacy.py
```diff
@@ -1228,6 +1210,24 @@ def file_upload(request):
...
+        if release.created < oldest_release_age_allowed:
+            request.metrics.increment(
+                "warehouse.upload.failed", tags=["reason:closed-release"]
+            )
+            raise _exc_with_message(
+                HTTPBadRequest,
+                f"Uploading new files to releases older than "
+                f"{MAXIMUM_AGE_FOR_NEW_UPLOADS_DAYS} days is not allowed.",
+            )
+
```
miketheman, line 1213: Could this block be placed right after `is_duplicate` check above, and exit earlier? Could the now-first "Only one sdist may be uploaded per release" error can lead a user to delete a closed release's only sdist, which can then never be restored?

miketheman (approved):
I suspect there's some timing or order of operation bugs in here - but I can't pin them down.

The entire view code needs further simplification, since it's really hard to follow the logic, but that shouldn't hold up this change.

miketheman (approved):
Yay!

PR #20263, opened 2026-07-11
feat: pyramid_components for HTML elements

miketheman:
`pyramid_components` lets a UI unit be defined as a typed Props contract plus a
co-located template, and rendered from Jinja2 with a `{% component "name", kw=val %}`
tag.
It ships a `Component` base class, the props-to-dataclass binding, a
module-level registry whose `render()` is the single path all rendering goes
through, the Jinja2 extension, and a Pyramid `includeme`.

Review commit by commit to see how the package remains isolated and the leverage is in defining our own components and calling them in template code.
These are a few admin-related examples that are useful for proving the concepts, before considering more refactoring in the main code for items like common forms behaviors - that should also be explored in admin first.


miketheman:
Split into stacked PRs for review, bottom to top: #20593 -> #20594 -> #20595 -> #20596.

PR #20265, opened 2026-07-11
Update hardcoded private/reserved IPs to RFC-5735 TEST-NET ranges (py…

### LINKED ISSUE #13186: Update all IP Addresses in tests to use RFC-5735 TEST-NET-* addresses.

IIIMGDIII:
…pi/warehouse#13186)

Replace hardcoded private IP addresses in test files with TEST-NET-* ranges:
- 127.0.0.1 → 192.0.2.1 (TEST-NET-1)
- 192.168.1.x → 198.51.100.x (TEST-NET-2)
- 10.x → 203.0.113.x (TEST-NET-3)
- 172.16.0.1 → 198.51.100.1

Updated REMOTE_ADDR constant and its pre-computed SHA256 hashes. Fixed source code defaults in tasks.py and metrics/services.py.

Closes #13186


JacobCoffee:
#15470, #20018, #20027 all do this and one of them is still open but they were abandonedd so ill choose this one and see if we can actually address PR comments and merge it..

warehouse/admin/views/malware_reports.py
```diff
@@ -477,6 +477,7 @@ def _selected_releases(project: Project, versions: list[str]):
     uses_session=True,
     require_csrf=True,
     require_methods=["POST"],
+    permit_duplicate_post_keys=True,
```
JacobCoffee: why this change in a test suite IP swap?

JacobCoffee (changes requested):

warehouse/admin/views/malware_reports.py
IIIMGDIII: Sorry, slipped in by mistake, removed...

IIIMGDIII: Rebased on main and resolved the conflicts in the xmlrpc test files. The diff is now tests-only, just the TEST-NET IP normalisation, nothing in source. Should be good for another look when you get a chance.

Issue #20273, opened 2026-07-15
More stateful tests found by randomly


di:
List of test runs that expose a non-deterministic test result (failure) when run with a specific random seed (ordering).

To reproduce, use TESTARGS="--randomly-seed=<SEED> -vvv" make tests before adding to this list.

Seeds:

- [x] `--randomly-seed=511478457` (https://github.com/pypi/warehouse/actions/runs/29442507769/job/87444902531) - #20306
- [x] `--randomly-seed=2144378614` (https://github.com/pypi/warehouse/actions/runs/29313673233/job/87023037977) - #20274

ewdurbin:
* --randomly-seed=511478457 (https://github.com/pypi/warehouse/actions/runs/29442507769/job/87444902531)

seems to be related to DNS resolution... though its not at all clear why the lines reported as uncovered... are reported as such.

* --randomly-seed=2144378614 (https://github.com/pypi/warehouse/actions/runs/29313673233/job/87023037977)

appears to be due to mutation of sentinels by the code under test, see #20274.

miketheman:
Both solved for now. Leaving open for a little while to collect any more that surface as test suite gets refactored

miketheman:
A couple more surfaced today, non-deterministic seeding.

- https://github.com/pypi/warehouse/actions/runs/31717009875/job/94504742202
- https://github.com/pypi/warehouse/actions/runs/31716029427/job/94501533621

PR #20274, opened 2026-07-15
Guard unittest.mock.sentinel objects from mutation during tests

ewdurbin:
We have been introducing usage of unittest.mock.sentinel into our test suite pretty broadly.

One of the test ordering issues flagged in #20273 is failing due to mutation of the request sentinel used.

This implements an autouse pytest fixture which inspects the global cache of unittest.mock.sentinel objects and loudly fails if any unittest.mock.sentinel object leaves the test with unexpected mutations.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -107,6 +107,27 @@ class _CallRecorder:
         return getattr(self._mock, name)
 
 
+@pytest.fixture(autouse=True)
+def guard_mock_sentinels():
+    yield
+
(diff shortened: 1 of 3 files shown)


miketheman (approved):
Nice work, I think there's still probably room for us to eventually adopt some lighter `request` object vs `pyramid_request` and `db_request` fixtures, and thus reduce the _possibility_ of global singleton mutation, but this is probably a good idea regardless.

PR #20281, opened 2026-07-19
Group dependabot CodeQL workflows

JacobCoffee:
Group workflows so that they all have the same base versions of things.

I think https://github.com/pypi/warehouse/pull/20269, https://github.com/pypi/warehouse/pull/20270, and https://github.com/pypi/warehouse/pull/20256 fail because of conflicts (maybe?)


miketheman (approved):
Let's give it a shot

PR #20288, opened 2026-07-20
Add Project Size Limit Increase Requests to Warehouse Admin

alanbato:
This PR addresses https://github.com/pypi/warehouse/issues/2017, which is currently being addressed by means of using GitHub Issues in the [pypi/support](https://github.com/pypi/support/) repo. Although I believe the current process works, given the nature of the nature of the request and similar to the organization applications, I think this process could be benefited by being able to be managed 100% in the Admin (for simple cases).

Why is this relevant? According to a quick search, there were requests **649** in the last 2 years
<details>
<summary>
pypi/support Issues related to project and upload size increases
</summary>
Screenshot from 2026-07-19 08-12-50
</details>

I went ahead and used some of the existing patterns for the Organization Application process.

I did not include the individual file size limit request to make this PR smaller, and could be done in a follow-up PR, but I'm also open to having both be landed at the time, left to the reviewer's discretion.

With this change, reviewing and acting on this requests can happen without switching contexts, in order to make this process easier and more scalable.

I included several new tests, and a set of screenshots to showcase the new process.

<details>
<summary>
Walkthrough Screenshots
</summary>

Current project settings page

1. How the project settings page looks today

New project settings page

2. How it would now look, notice the new Request button.

New size limit request form

3. The button opens up a new form modal asking the questions from the current Issue template.

Form filled, returned with a success banner

4. The form is filled and submitted

New section in the Admin for project size limit requests

5.There's a new section in the admin for Project Size Limits (requests)

Request detail page

6. A request contains all of the information from the Issue Template, and is actionable

Request approve confirmation modal

7. Requests can be approved/denied with an optional message.

Request approval success

8. The request was approved!

Request approval email
Request denial email

9. There are emails sent on approval/denial

</details>


woodruffw:
Hi @alanbato, thanks for opening a PR. The issue you linked to is nearly a decade old, so I think this would benefit from some discussion there about scope and approach first. 

(In particular things like cross-index requests: it seems misleading to allow a user to request a request on PyPI and TestPyPI from a specific project on PyPI, since the two indices don't share the same namespace.)

alanbato:
Thanks for the quick reply @woodruffw!

> The issue you linked to is nearly a decade old, so I think this would benefit from some discussion there about scope and approach first.

That's very fair, the issue is indeed very old, and I should've asked how the templates had been working so far.

Can I ask where do you think the conversation's to be had? The original issue, or some other channel? Particularly:
* If the current approach is good enough, this PR can be closed until the need is revisited
* If its a matter of adding/removing fields, I think this PR could now be the place for that

With regards to the cross-index request, I tried to go with what was already on the issue template (minus information that we could deduce) and the index it's actually not necessary anymore, now that it's requested on-platform, each request will go to its own admin, and if you want both you'd do both, just like Trusted Publishing today. Given that I don't think it makes sense to keep it, I'll just go ahead and remove it.

JacobCoffee:
I think its best to keep requests in GitHub issues for now until we decide if we want to move processes into warehouse proper. 

IMO it might be nice to expose APIs for these types of thing and have an external support app manage things to keep warehouse focused 

If/when we want to pursue something like this again I'll try my best to remember to ping you and perhaps re-open your PR/salvage some of this work so it wasn't a time sink! Appreciate you

alanbato:
> I think its best to keep requests in GitHub issues for now until we decide if we want to move processes into warehouse proper.

Sounds good! Definitely wouldn't want to change a system that works today :)

> I'll try my best to remember to ping you and perhaps re-open your PR/salvage some of this work so it wasn't a time sink! 

No worries if not, it was a good challenge to learn from and get to know the codebase again.

>  Appreciate you

Likewise for your time!

PR #20297, opened 2026-07-21
Add contact links for issue templates

JacobCoffee:
Add more links to the template list so that (hopefully) people create less account recovery/541 issues on the repo


miketheman (approved):
Seems reasonable. I wish there was some sort of linter/checker for these control files, so we learn if there's anything broken before we ship it.

Issue #20131, opened 2026-06-06
Add internal notes for organization applications


JacobCoffee:
When processing org apps I want to sometimes leave a note for other team members that is not sent or visible to the requesting party.

For example:
- "Hey, I don't think we can approve this because OFAC or XYZ reason, defer for now"

<img width="425" height="404" alt="Image" src="https://github.com/user-attachments/assets/0041e120-91cd-4e4d-860a-eb4bae706b35" />

alanbato:
Hey @JacobCoffee, I'm thinking on picking this ticket up.
Were you thinking on this button appending to a list of comments? Or a per-application scratch pad that any admin can edit?
My guess is something like the former, but wanted to ask first!

JacobCoffee:
hey @alanbato 

imo it should be in the `Information Requests` area since ongoing chats are there, just rendered differently. Very similar to Helpscout's (which we use internally) internal note UI https://docs.helpscout.com/article/69-respond-to-conversations

alanbato:
How'd this look?

<img width="824" height="593" alt="Image" src="https://github.com/user-attachments/assets/e2f2ea2d-b23d-4625-b738-d40f02d3b3f5" />

JacobCoffee:
yeah thats wicked

PR #20299, opened 2026-07-21
Organization Application Internal Notes on the Admin

alanbato:
Addresses Issue #20131

Most of this is reusing Observations, which were already used for Information Requests, with the difference of now sending any emails/notifications whatsoever. They are rendered in the same list view of Information Requests.

Extracted some of the presentation logic to be used across Information Requests and Internal 

Screenshot from 2026-07-20 18-11-55

Screenshot from 2026-07-21 17-45-20

### DIFF
--- a/tests/unit/admin/views/test_organization_applications.py
+++ b/tests/unit/admin/views/test_organization_applications.py
@@ -815,3 +815,97 @@ class TestOrganizationApplicationActions:
 
         with pytest.raises(HTTPNotFound):
             views.organization_application_decline(request)
+
+    def test_addnote(self, db_request):
+        admin = UserFactory.create()
+        user = UserFactory.create()
+        organization_application = OrganizationApplicationFactory.create(
+            name="example", submitted_by=user
+        )
+
+        organization_service = pretend.stub(
+            get_organization_application=lambda *a, **kw: organization_application,
+            add_organization_application_note=pretend.call_recorder(
+                lambda *a, **kw: organization_application
+            ),
+        )
+
+        db_request.matchdict = {
+            "organization_application_id": organization_application.id
+        }
+        db_request.params = {"message": "Some internal note"}
+        db_request.user = admin
+        db_request.route_path = pretend.call_recorder(_organization_application_routes)
+        db_request.find_service = pretend.call_recorder(
+            lambda iface, context: organization_service
+        )
+        db_request.session.flash = pretend.call_recorder(lambda *a, **kw: None)
+
+        result = views.organization_application_add_note(db_request)
+
+        assert organization_service.add_organization_application_note.calls == [
+            pretend.call(organization_application.id, db_request),
+        ]
+        assert db_request.session.flash.calls == [
+            pretend.call(
+                f'Note added to "{organization_application.name}" application',
+                queue="success",
+            ),
+        ]
+        assert result.status_code == 303
+        assert (
+            result.location
+            == f"/admin/organization_applications/{organization_application.id}/"
+        )
+
+    def test_addnote_no_message(self, db_request):
+        admin = UserFactory.create()
+        user = UserFactory.create()
+        organization_application = OrganizationApplicationFactory.create(
+            name="example", submitted_by=user
+        )
+
+        organization_service = pretend.stub(
+            get_organization_application=lambda *a, **kw: organization_application,
+            add_organization_application_note=pretend.call_recorder(
+                pretend.raiser(ValueError)
+            ),
+        )
+
+        db_request.matchdict = {
+            "organization_application_id": organization_application.id
+        }
+        db_request.params = {}
+        db_request.user = admin
+        db_request.route_path = pretend.call_recorder(_organization_application_routes)
+        db_request.find_service = pretend.call_recorder(
+            lambda iface, context: organization_service
+        )
+        db_request.session.flash = pretend.call_recorder(lambda *a, **kw: None)
+
+        result = views.organization_application_add_note(db_request)
+
+        assert organization_service.add_organization_application_note.calls == [
+            pretend.call(organization_application.id, db_request),
+        ]
+        assert db_request.session.flash.calls == [
+            pretend.call("No note text provided", queue="error"),
+        ]
+        assert result.status_code == 303
+        assert (
+            result.location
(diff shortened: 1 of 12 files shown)


tests/unit/admin/views/test_organization_applications.py
```diff
@@ -815,3 +815,97 @@ def test_decline_not_found(self):
...
+    def test_addnote(self, db_request):
+        admin = UserFactory.create()
+        user = UserFactory.create()
+        organization_application = OrganizationApplicationFactory.create(
+            name="example", submitted_by=user
+        )
+
+        organization_service = pretend.stub(
+            get_organization_application=lambda *a, **kw: organization_application,
+            add_organization_application_note=pretend.call_recorder(
```
JacobCoffee: if possible all future code needs to not use pretend

https://github.com/pypi/warehouse/issues/18880 for ref.

alanbato: Done! Reduced the mocking in the tests, and did a drive-by fix of a test that was being overmocked as well.

warehouse/admin/templates/admin/organization_applications/detail.html
```diff
@@ -318,6 +340,41 @@ <h4 class="modal-title" id="declineModalLabel">
...
+                    </div>
+                    <div class="modal-body">
+                      <p>
+                        This note is for internal admin reference only. It will not be emailed to the applicant.
+                      </p>
+                      <div class="form-group">
+                        <label for="addNoteModalMessage">
+                          Note
+                        </label>
+                        <textarea id="addNoteModalMessage" class="form-control" name="message" placeholder="Internal note" rows=6></textarea>
```
JacobCoffee: make this required so we don't POST blank messages

JacobCoffee (changes requested):
lgtm besides the empty note submission

image

image

JacobCoffee:
hmm also maybe a yellow?  instead of competing gray with deferal
image

warehouse/organizations/models.py
```diff
@@ -765,6 +765,26 @@ def information_requests(self):
             reverse=True,
         )
 
+    @property
```
JacobCoffee, line 769: this is almost 1:1 for `def information_requests(self):` would it make sense to helper-it so save some lines? maybe TOO much abstraction, thinking out loud. ffine as is

warehouse/organizations/services.py
```diff
@@ -282,6 +282,29 @@ def request_more_information(self, organization_application_id, request):
...
+        Records an internal admin note on an OrganizationApplication
+        """
+        organization_application = self.get_organization_application(
+            organization_application_id
+        )
+
+        message = request.params.get("message", "")
+
+        if not message:
+            raise ValueError
```
JacobCoffee: could this be walrus'd
```suggestion
        if not (message := request.params.get("message")):
    		raise ValueError
```
mauybe not since it is similar to `request_more_information` above. just thinking out loud, again :)

warehouse/organizations/models.py
JacobCoffee: prob fine until (if) we add a third

alanbato: You're right that it's doing the same sorting + querying the same model, but filtering for different values of a column... 

Maybe having a `get_observations(kind: str)` method that both properties use? But the class doesn't have any other methods, so it feels weird to add one now...

warehouse/organizations/services.py
alanbato: I think we could do it if we flipped the early-return and left it as part of the else clause

```python
if message := request.params.get("message"):
    # Do something
    ....
else:
    raise ValueError
```

alanbato:
> hmm also maybe a yellow? instead of competing gray with deferal image

Let there be yellow! 
image

JacobCoffee (approved):

alanbato:
Are we good to merge this? :)

warehouse/organizations/services.py
JacobCoffee, line 252: Chunked the PR into frontier model to get a final review just to cross-check both of us,, only 1 meaningful thing although i think from looing at main this is already a bug before the PR?

1. Status flips on empty message (`services.py:259`) — `request_more_information` sets `status = MoreInformationNeeded` before validating the message. Empty message → `ValueError` → view catches it → transaction still commits, so the status changes with no observation or email. Pre-existing, but this PR rewrites exactly that block - May want to move the assignment inside the if message: branch.

```
The service (warehouse/organizations/services.py:251-270 on the PR branch):

def request_more_information(self, organization_application_id, request):
    organization_application = self.get_organization_application(
        organization_application_id
    )
    organization_application.status = (              # ← mutates state first
        OrganizationApplicationStatus.MoreInformationNeeded
    )

    if message := request.params.get("message", ""):
        organization_application.record_observation(...)
        send_new_organization_moreinformationneeded_email(...)
        return organization_application
    raise ValueError                                  # ← status already changed

The view (warehouse/admin/views/organizations.py, organization_application_request_more_information):

try:
    organization_service.request_more_information(...)
    ...
except ValueError:
    request.session.flash("No message provided", queue="error")

return HTTPSeeOther(...)   # normal response

Why it commits: the view swallows the ValueError and returns a normal 303. pyramid_tm only aborts the transaction when an exception propagates out of the view — a caught exception plus a normal response means commit. So the SQLAlchemy session flushes the dirty status attribute.

Net effect: an admin clicks "Request More Info" with an empty message (possible via direct POST — the textarea's rly). They see the red "Nomessage provided" flash, and reasonably assume nothing happened. But the application is now in MoreInformationNeeded —est observation, no email tothe applicant, and nothing in the thread explaining the status. The applicant was never asked for anything, and the apatever queue filters on
```

JacobCoffee (approved):
out of scope comment, just curious on your thoughts

warehouse/organizations/services.py
JacobCoffee: can be a later thing

alanbato: You're right, we mutate status before validating the existence of a message, and this was a pre-existing bug that I didn't notice when applying the walrus operator on the validation clause. Let me submit a reproducing test + fix

PR #20302, opened 2026-07-21
Fix Organization Application State Corruption on Invalid Information Requests

alanbato:
## Problem
While working on #20299, a pre-existing issue with the information request submission endpoint was found:

> Status flips on empty message (services.py:259) — request_more_information sets status = MoreInformationNeeded before validating the message. Empty message → ValueError → view catches it → transaction still commits, so the status changes with no observation or email. 

This meant that when a non-empty message was posted to the endpoint, we still changed the status of the organization application, leaving it in a corrupted state where it was awaiting information but the request of it was never sent.

This PR moves the status change to the code block where the information request is already validated.


JacobCoffee (approved):

JacobCoffee:
thanks @alanbato

Issue #18546, opened 2025-08-14
Implement the "deprecated" status marker from PEP 792


woodruffw:
PyPI has now implemented PEP 792, and includes support for three of the four standard status markers. The last one is `deprecated`, i.e. to signal that a package is considered obsolete and may have been superseded by another project. 

I think implementing this will probably follow a pretty similar course as `archived`: it'll need to be added to `LifecycleStatus`, then exposed to project owners, and then propagated like the other statuses to the index.

Related subcomponents:

* Making sure we emit a proper security event for this status change

xrefs #345 #16844 #18415

alanbato:
What does it mean to emit a proper security event in this context?


EDIT: Ah, it must mean having a `project.record_event(tag=...)` call whenever the status changes

woodruffw:
> EDIT: Ah, it must mean having a `project.record_event(tag=...)` call whenever the status changes

Yep, exactly.

PR #20304, opened 2026-07-22
Add Deprecated status to Project and Release lifecycle backend

alanbato:
This PR extends the Project/Release lifecycle state machine to add support for the `"deprecated"` status, closely following the implementation of the `archived` status, which has similar (but not equal) semantics, according to PEP 792, described in #18546

I'd figured the backend changes could be added first, laying the groundwork of adding the possibility of a new status, and then add the "frontend" portion to the warehouse frontend / index / documentation enabling project maintainers to use this status.

### DIFF
--- a/warehouse/events/tags.py
+++ b/warehouse/events/tags.py
@@ -120,6 +120,8 @@ class EventTag:
         ProjectArchiveEnter = "project:archive:enter"
         ProjectArchiveExit = "project:archive:exit"
         ProjectCreate = "project:create"
+        ProjectDeprecateEnter = "project:deprecate:enter"
+        ProjectDeprecateExit = "project:deprecate:exit"
         ProjectQuarantineEnter = "project:quarantine:enter"
         ProjectQuarantineExit = "project:quarantine:exit"
         ProjectSetTotalSizeLimit = "admin:project:set_total_size_limit"
--- a/warehouse/migrations/versions/d3dac0f3bb25_add_new_lifecyclestatus_for_deprecated.py
+++ b/warehouse/migrations/versions/d3dac0f3bb25_add_new_lifecyclestatus_for_deprecated.py
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Add new LifecycleStatus for Deprecated
+
+Revision ID: d3dac0f3bb25
+Revises: 423ffda7411f
+Create Date: 2026-07-22 00:00:00.000000
+"""
+
+from alembic import op
+from alembic_postgresql_enum import TableReference
+
+revision = "d3dac0f3bb25"
+down_revision = "423ffda7411f"
+
+
+def upgrade():
+    op.execute("ALTER TYPE public.lifecyclestatus ADD VALUE IF NOT EXISTS 'deprecated'")
+
+
+def downgrade():
+    op.sync_enum_values(
+        enum_schema="public",
+        enum_name="lifecyclestatus",
+        new_values=[
+            "quarantine-enter",
+            "quarantine-exit",
+            "archived",
+            "archived-noindex",
+        ],
+        affected_columns=[
+            TableReference(
+                table_schema="public",
+                table_name="projects",
+                column_name="lifecycle_status",
+            ),
+            TableReference(
+                table_schema="public",
+                table_name="releases",
+                column_name="lifecycle_status",
+            ),
+        ],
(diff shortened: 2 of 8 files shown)


warehouse/migrations/versions/d3dac0f3bb25_add_new_lifecyclestatus_for_deprecated.py
```diff
@@ -0,0 +1,44 @@
...
+
+from alembic import op
+from alembic_postgresql_enum import TableReference
+
+revision = "d3dac0f3bb25"
+down_revision = "423ffda7411f"
+
+
+def upgrade():
+    op.execute("ALTER TYPE public.lifecyclestatus ADD VALUE IF NOT EXISTS 'deprecated'")
```
ewdurbin, line 18: glad you used this! nice job. did we document it somewhere?

ewdurbin (approved):

warehouse/utils/project.py
```diff
@@ -348,3 +348,44 @@ def unarchive_project(project: Project, request) -> None:
...
+
+def undeprecate_project(project: Project, request) -> None:
+    if project.lifecycle_status != LifecycleStatus.Deprecated:
+        request.session.flash(
+            "Can only undeprecate a deprecated project",
+            queue="error",
+        )
+        return
+
+    project.lifecycle_status = None
```
miketheman, line 383: should this be explicitly moved to Active instead of None?

warehouse/migrations/versions/d3dac0f3bb25_add_new_lifecyclestatus_for_deprecated.py
alanbato: I was wondering how to update the possible values for the Enum with SQLAlchemy, and while searching the codebase for past instances of this I found this migration, added in a previous PR targeting lifecycle statuses:

https://github.com/pypi/warehouse/pull/17843/changes#diff-d0bd4b0a6c43cd68f075ff05914e821576af65300f00d607290d71d2b06f9bbeR30

That said, I wouldn't be able to tell you exactly what `ADD VALUE IF NOT EXISTS` does :sweat_smile: I was merely trying to hopefully follow convention

warehouse/utils/project.py
alanbato: I was wondering the same thing, but `unarchive_project` above also set its to None :thinking:, and I think the way we check for these states in the state machine is that not having a particular status set is equal to the project being active

https://github.com/pypi/warehouse/pull/20304/changes/BASE..3f9ce9f6b9f3034f853e23099e5730118683a065#diff-2706bd622e672de6d3b46b5353282111ccc34efa789289c31958ab8134f60bdcR342

Happy to set the project to active in both places, but I don't have context as to why it's set to `None` in the `unarchive` method, there may be a reason :man_shrugging:

JacobCoffee (approved):

alanbato:
Is there anything I can do to get this PR accross the finish line and have it merged? :cowboy_hat_face:

warehouse/utils/project.py
woodruffw: My memory is hazy, but IIRC the reason we clear it (instead of setting it to `active`) is purely because `None` implies `active` per PEP 792. So we save a few bytes (and cache invalidations) by not threading a redundant state through.

warehouse/events/tags.py
```diff
@@ -120,6 +120,8 @@ class Project(EventTagEnum):
         ProjectArchiveEnter = "project:archive:enter"
         ProjectArchiveExit = "project:archive:exit"
         ProjectCreate = "project:create"
+        ProjectDeprecateEnter = "project:deprecate:enter"
```
woodruffw, line 123: I think we _probably_ want to update `project/history.html` with these new states, otherwise they'll get rendered as raw events.

woodruffw: (That might be fine if we're scoping this PR intentionally to just the backend, though.)

woodruffw (commented):
Backend changes LGTM! I left one nit about a place where we might want to make a frontend change in this PR as well; the other frontend changes (like grouping with archived projects) are not as tightly bound.

warehouse/events/tags.py
alanbato: You're right! I took the liberty of also including the other states that had been added since `project/history.html` was last updated.

PR #20309, opened 2026-07-22
Stripe: Add reconciliation task

JacobCoffee:
- Add reconciliation task for stripe daily before the other task
- Dont be so aggressive on exception, shrink scope to match (caught in https://github.com/pypi/warehouse/pull/20305#discussion_r3630302189)


warehouse/organizations/tasks.py
```diff
@@ -113,6 +119,78 @@ def update_organziation_subscription_usage_record(request):
...
+@tasks.task(ignore_result=True, acks_late=True)
+def reconcile_stripe_status(request):
+    # Re-sync each subscription's status from Stripe so that state we would have
+    # learned from a webhook (e.g. a cancellation) is recovered even if the
+    # webhook was dropped. Mirrors the customer.subscription.updated handler.
+    organization_subscriptions = request.db.query(OrganizationStripeSubscription).all()
+
+    billing_service = request.find_service(IBillingService, context=None)
+    subscription_service = request.find_service(ISubscriptionService, context=None)
+    metrics = request.find_service(IMetricsService, context=None)
```
miketheman: Unneded, use `request.metrics...` instead

warehouse/subscriptions/services.py
```diff
@@ -59,6 +65,18 @@ def get_customer(self, subscription_id):
...
+        """
+        Fetch the Subscription resource from the Billing API, or None if it
+        no longer exists
+        """
+        try:
+            return self.api.Subscription.retrieve(subscription_id)
+        except stripe.error.InvalidRequestError as exc:
+            if exc.code != "resource_missing":
+                raise
+            return None
```
miketheman: This is somewhat incorrect logic here.

The reconcile task treats `None` as "the subscription was deleted, mirror the `customer.subscription.deleted` handler".
But canceled subscriptions are still retrievable from Stripe, so `resource_missing` almost never means canceled.

If we muck up the secret key, and every `.retrieve` would raise a `resource_missing`, and then reconciliation would cancel all of the subscriptions, which is probably bad.

warehouse/organizations/tasks.py
```diff
@@ -21,8 +22,9 @@
     OrganizationStripeSubscription,
     OrganizationType,
 )
-from warehouse.subscriptions.interfaces import IBillingService
+from warehouse.subscriptions.interfaces import IBillingService, ISubscriptionService
 from warehouse.subscriptions.models import StripeSubscriptionStatus
+from warehouse.subscriptions.services import TRANSIENT_STRIPE_ERRORS
```
miketheman: I think this import creates an import cycle (we have a few in the codebase, which isn't good, and ruff can't detect them yet).

The chain: `subscriptions/services.py` imports `warehouse.organizations.models`, which runs `organizations/__init__.py`, which imports `organizations.tasks`, which nowimports `TRANSIENT_STRIPE_ERRORS` back from the partially-initialized `services` module, where the tuple is not yet defined.

Today this works because `config.include(".organizations")` runs before `.subscriptions`.
Any entry point that imports `warehouse.subscriptions.services` first dies with `ImportError` at startup.

One fix idea: define the errors in `subscriptions/interfaces.py`, which `tasks.py` already imports cycle-free.

warehouse/organizations/tasks.py
```diff
@@ -113,6 +119,78 @@ def update_organziation_subscription_usage_record(request):
...
+            # can't wedge the whole run at the same row every night.
+            logger.exception(
+                "Failed to reconcile subscription %s", subscription.subscription_id
+            )
+            metrics.increment(
+                "warehouse.organizations.subscription.status.reconcile.error",
+                tags=[f"error_type:{exc.__class__.__name__}"],
+            )
+            continue
+        deleted = remote is None
```
miketheman: this might be dead code if `resource_missing` is changed.
idea: detect cancellation by `status == "canceled"` in the fetched
subscription, not by absence of the resource

warehouse/organizations/tasks.py
```diff
@@ -113,6 +119,78 @@ def update_organziation_subscription_usage_record(request):
             )
 
 
+@tasks.task(ignore_result=True, acks_late=True)
+def reconcile_stripe_status(request):
+    # Re-sync each subscription's status from Stripe so that state we would have
+    # learned from a webhook (e.g. a cancellation) is recovered even if the
+    # webhook was dropped. Mirrors the customer.subscription.updated handler.
+    organization_subscriptions = request.db.query(OrganizationStripeSubscription).all()
```
miketheman, line 129: The intent here appears to run reconciliation on every subscription daily? Seems like a slowly-growing, ever-increasing task timing?

warehouse/organizations/tasks.py
```diff
@@ -113,6 +119,78 @@ def update_organziation_subscription_usage_record(request):
             )
 
 
+@tasks.task(ignore_result=True, acks_late=True)
+def reconcile_stripe_status(request):
+    # Re-sync each subscription's status from Stripe so that state we would have
+    # learned from a webhook (e.g. a cancellation) is recovered even if the
+    # webhook was dropped. Mirrors the customer.subscription.updated handler.
```
miketheman, line 128: Are these still the same, or should the method be extracted?

The webhook `deleted-handler` records `SubscriptionCancel` unconditionally when the subscription is found, while the task records it only on status change.
The webhook passes `StripeSubscriptionStatus.Canceled` where the task
passes `.value` strings.

idea: extract one `sync_subscription_status` helper on the `subscription`
service and call it from both `warehouse/api/billing.py` and this task here

warehouse/organizations/tasks.py
miketheman, line 129: Another query for everything. This also expands into an n+1 query later on as the relates .subscription and .organization calls are loaded lazily. See `notify_organizations_requiring_subscription`  for how you did it last time. Seems like an opportunity for a refactor?

miketheman (changes requested):
I got through a bunch, but not all of the code paths. There's still some weird stuff going on here that isn't intuitive right now, so I'm hesitant to adopt this as is.

warehouse/organizations/tasks.py
JacobCoffee: [16df738](https://github.com/pypi/warehouse/pull/20309/commits/16df73872996217f43c42b59f1d713341c3d63e4) if that works, trying to keep the iface clean but i can move it there if you think thats better

warehouse/subscriptions/services.py
JacobCoffee: [b944164](https://github.com/pypi/warehouse/pull/20309/commits/b944164c578b879705534e916ab76c074501d969) but also wouldnt the raise for invalid secret do a autherror instead of the resource_missing>

warehouse/organizations/tasks.py
JacobCoffee: [29c7881](https://github.com/pypi/warehouse/pull/20309/commits/29c78811f60371d6d98bb92c222e89f562d88e16) checks for that now

warehouse/organizations/tasks.py
JacobCoffee: yep makes sense. [6dcb212](https://github.com/pypi/warehouse/pull/20309/commits/6dcb212ae8153dec4ad94524fddaad6989f9f2b0) n [72ed879](https://github.com/pypi/warehouse/pull/20309/commits/72ed879c41049fc2e6dfa6c00b3cd0a58a9cc92e)

warehouse/organizations/tasks.py
JacobCoffee: both of these seem to have `lazy=False` woudl that not make them eager?

JacobCoffee: i can maybe page stripes api https://docs.stripe.com/api/subscriptions/list to make it a little nicer

JacobCoffee: [68772e0](https://github.com/pypi/warehouse/pull/20309/commits/68772e09656081f1d2451bfb2becd7c9f7c9b077)/[8f86e11](https://github.com/pypi/warehouse/pull/20309/commits/8f86e114077dabfb0c0701612b0a390ffa2b760a)

JacobCoffee: after doing some queries we have about 1500 subs locally so with the paginated approach itll be ~15 requests -- lets stick with the paginated approach for now if thats okay? the alternative i can think of was a get call on a sub but then that hits ~1200 of the unfinished subs and thats a lot of calls.
if you have a bette ridea im all ears

```sql
SELECT
    count(*) AS total_subscriptions,
    count(*) FILTER (
        WHERE status NOT IN ('canceled', 'incomplete_expired')
    ) AS unfinished_subscriptions,
    ceil(count(*) / 100.0) AS estimated_list_requests
FROM stripe_subscriptions;
```
was my query to see the impact

PR #20323, opened 2026-07-24
fix: re-send unrecognized-login email on repeat login attempts

### LINKED ISSUE #20312: "Unrecognized login to your PyPI account" emails not sent?
**Describe the bug**

I tried logging into PyPI using 2FA, and I was asked to confirm on the confirm-login page. I didn't get any email to confirm the login.

**Expected behavior**

I expected to get a confirmation email for the login.

**To Reproduce**

Try to log in using an account with 2FA enabled. Confirm 2FA and you will be redirected to the https://pypi.org/account/confirm-login/ page. 

**My Platform**

Latest macOS (26.5.2), Firefox 153.0.

**Additional context**

miketheman:
Once a pending UserUniqueLogin existed for a user+IP, repeat logins
within the confirmation window never sent another email. If the first
email was lost or bounced, the user was stuck until the pending login
expired.

Now every unconfirmed login attempt re-sends the email, throttled to
one per 15 minutes at the email layer so retries can't flood an inbox.

The throttle is skipped for a new device or a lapsed window: the user
holds no valid token in either case, so that email must always go out.
device_is_known disables it via a new per-call repeat_window override
on the _email wrapper.

Fixes #20312

### DIFF
--- a/docs/dev/development/email.md
+++ b/docs/dev/development/email.md
@@ -45,6 +45,10 @@ Calling a function with the `_email` decorator does the following:
   send within `repeat_window`), a `warehouse.emails.skipped` metric is sent
   with the same tags plus a `reason` tag.
 
+Callers may pass a `repeat_window` keyword argument to override the
+decorator's value for a single call — pass `None` for a send that must
+always go out, e.g. notifying a newly-seen device.
+
 ## Testing e-mails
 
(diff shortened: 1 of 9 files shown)


JacobCoffee (approved):
makes sense to me!

Issue #20155, opened 2026-06-12
Organization rate limits


JacobCoffee:
**What's the problem this feature will solve?**

Currently we have global rate limits that apply to everyone.
- We have no way to allow orgs as a whole higher limits
- We have no way to granularly lift limits for an org (or user, but irrelevant here) like we do for say project size limit increases.

**Describe the solution you'd like**

Organizations are more trusted as they go through more vetting, and also they usually contain more users per project so there is more chance to hit the default rate limit for project creation - so allowing them a default higher rate limit makes sense.

It would also be nice to adjust, per-org, the limit.

**Additional context**

alanbato:
Is this something that is ready to be worked on, and up for grabs?
I see a PR that could be useful to serve as a base at https://github.com/pypi/warehouse/pull/18496

JacobCoffee:
Hey @alanbato 

I think #20046 satisfies this issue iirc but you are welcome to pick it up and finish it. I havent been able to get a review on it.

alanbato:
By "pick it up and finish it", you mean I use it as a base to create a PR of my own?
I don't think I can push to your branch in order to continue working on your current one 😅
Also wouldn't know about getting a review 🙂 Maybe we could split up some of the code into separate PRs to make them easier to review? 🤔

JacobCoffee:
> I use it as a base to create a PR of my own?

correct
I'm happy to review yours after you've wrapped it up as  I dont have time at the moment to see it through to completion (although I think its mostly there iirc?)

PR #20046, opened 2026-05-08
Configurable project creation limits

JacobCoffee:
Allows granular configurability for orgs and users that override the current global daily count

Backend creates the buckets and limiters for orgs.

Admin UI changes adds the field:
for orgs
image
for users
image

Tries to hand hold in the UI with the dropdown for d/m/yy

--

Tested locally, set org limit 5 per hour, user limit 2 per hour
Blocked on 6th org project create and 3rd user project create

also tested the per-user overrides (default 20, failed at 21) with manual override on local `di` user set to 5 (failed at 6)

Closes #20155

PR #20324, opened 2026-07-24
Allow admins to configure Project creation rate limits

alanbato:
This PR addresses Issue #20155

However, while working on such a feature, I was let know by @JacobCoffee about his previous attempt to introduce this in #20046 

The implementations are largely similar, except in two cases:

* Projects created by Org-scoped Trusted Publisher  have `ratelimited=False`, with a `TooManyProjectsCreated` except-clause that can never fire, the test for it only passes because it mocks `create_project directly`, bypassing the rate limit gate. Mine has `ratelimited=pending_publisher.organization_id is not None`, so org-scoped trusted-publisher project creation is genuinely rate-limited.
* The `Event` tag. I create an `EventTag.Organization.OrganizationSetProjectCreateRateLimit` enum member (matching the set_upload_limit/set_total_size_limit precedent) instead of a raw string literal inline.

Screenshots, which look largely the same as the previous PR, but I took them anyways during manual validation :)

For Organizations:
Screenshot from 2026-07-24 15-55-29

For Users
Screenshot from 2026-07-24 17-02-24


warehouse/organizations/models.py
```diff
@@ -354,6 +354,12 @@ class Organization(OrganizationMixin, HasEvents, db.Model):
         BigInteger,
         comment="Maximum total size limit in bytes for projects in this organization",
     )
+    project_create_ratelimit_string: Mapped[str | None] = mapped_column(
```
JacobCoffee: is there a better way to do this? @miketheman mentioned this might be the first time we've stuffed a config like this onto a model

alanbato: Yeah I was debating on whether to store this config in two different columns or like this. If this seems like a code smell, happy to de-structure it and save the bounded int and period Enum separately.

JacobCoffee:
@alanbato this looks a lot better for implementing new rate limit surface than that org rate limit PR you took over if you want to refactor FYI (also apologies for the delay in review)

https://github.com/pypi/warehouse/pull/20412

JacobCoffee:
specifically no new DB rows but maybe other things

warehouse/admin/forms.py
```diff
@@ -108,3 +116,39 @@ def validate_total_size_limit(self, field):
...
+            ),
+        ],
+    )
+    project_create_ratelimit_period = wtforms.SelectField(
+        choices=[(period.value, period.value) for period in RateLimitPeriod],
+        coerce=RateLimitPeriod,
+        default=RateLimitPeriod.Hour,
+    )
+
+    def apply_to(self, entity: User | Organization) -> str | None:
```
ewdurbin: this is a little out of sync with the way most of our interactions with User and Organization objects go...

normally rather than having this be defined on the form, we would expose a method on the UserService or OrganizationService like `apply_ratelimit` (and maybe even `clear_ratelimit`).

i think this would match the convention of the codebase better, and encapsulate the emission of the UserEvent/OrganizationEvent inside of the method.

warehouse/events/tags.py
```diff
@@ -89,6 +89,7 @@ class Account(EventTagEnum):
         RoleInvite = "account:role:invite"
         RoleRemove = "account:role:remove"
         RoleRevokeInvite = "account:role:revoke_invite"
+        SetProjectCreateRateLimit = "admin:account:set_project_create_ratelimit"
```
ewdurbin: most of our permissions do not start with the verb and have three parts starting with the object  , i'd recommend something more like 

```suggestion
        ProjectCreateRateLimitChange = "account:project_create_ratelimit:change"
```

i think i see why you picked admin as a prefix, but this seems to be a better fit to our current event tags, details like the actor (admin) are appropriate in the `additional` field and allow for reconstructing that it was an admin action.

warehouse/events/tags.py
```diff
@@ -176,6 +177,9 @@ class Organization(EventTagEnum):
         ManualActivationAdd = "admin:organization:manual_activation:add"
         ManualActivationUpdate = "admin:organization:manual_activation:update"
         ManualActivationDelete = "admin:organization:manual_activation:delete"
+        OrganizationSetProjectCreateRateLimit = (
```
ewdurbin: similar to the Account event tag

```suggestion
        OrganizationProjectCreateRateLimitChange = (
```

warehouse/events/tags.py
```diff
@@ -176,6 +177,9 @@ class Organization(EventTagEnum):
...
+            "admin:organization:set_project_create_ratelimit"
```
ewdurbin: similar to the Account event tag

```suggestion
            "organization:project_create_ratelimit:change"
```

(looking back, it seems i am the one to blame for the historical admin: prefix, so blame me for setting you off track)

warehouse/packaging/services.py
```diff
@@ -624,6 +642,7 @@ def create_project(
                     organization_id=organization_id, project_id=project.id
                 )
             )
+            orm.attributes.flag_dirty(self.db.get(Organization, organization_id))
```
ewdurbin: why is this necessary?

JacobCoffee: as i understood it purges only fire for the `Organization` when the org row itself was changed, and adding an `OrganizationProject` doesn't do that so flippin it dirty lets us purge on project creation for the org

it didnt happen before but now we have these extra buckets so i was trying to work around that (passing org ID into `create_project`)

i'll swap it to just call `add_organization_project` and it should also address your other comment about going through the services.

JacobCoffee: [dc6fdcf](https://github.com/pypi/warehouse/pull/20324/commits/dc6fdcfd7148a80e47fb877ee56a5c377ee01091) should fix that since the svc already did the pruge

warehouse/admin/forms.py
JacobCoffee: tbese should be better!

[64a7123](https://github.com/pypi/warehouse/pull/20324/commits/64a71234484d0f75090f4ef57885eec4c6aa97a9)

[f1767cc](https://github.com/pypi/warehouse/pull/20324/commits/f1767ccf5bdf80872747b56df0b3d52bbe57581d)

[0d0c7f4](https://github.com/pypi/warehouse/pull/20324/commits/0d0c7f4371969ce1ba20b227c75665c8769be445)


[4f98c54](https://github.com/pypi/warehouse/pull/20324/commits/4f98c54fd810975bb6f045fa17e70ab9e4e79052)

warehouse/accounts/models.py
```diff
@@ -107,6 +108,19 @@ class User(SitemapMixin, HasObservers, HasObservations, HasEvents, db.Model):
     )
     disabled_for: Mapped[DisableReason | None]
 
+    project_create_ratelimit_count: Mapped[int | None] = mapped_column(
```
ewdurbin, line 111: should we have a constraint that these are either not null or null together?

warehouse/organizations/models.py
```diff
@@ -355,6 +356,18 @@ class Organization(OrganizationMixin, HasEvents, db.Model):
         BigInteger,
         comment="Maximum total size limit in bytes for projects in this organization",
     )
+    project_create_ratelimit_count: Mapped[int | None] = mapped_column(
```
ewdurbin, line 359: should we have a constraint that these are either not null or null together?

warehouse/organizations/models.py
```diff
@@ -405,6 +418,14 @@ class Organization(OrganizationMixin, HasEvents, db.Model):
         back_populates="pypi_organization",
     )
 
+    @property
+    def project_create_ratelimit_string(self) -> str | None:
+        """Composed `limits`-syntax string, or None when no override is set."""
+        if self.project_create_ratelimit_count is None:
+            return None
+        period = self.project_create_ratelimit_period or RateLimitPeriod.Hour
```
ewdurbin, line 426: this feels off, if either the period or count is None, I feel like we should error/bail. noted in another comment, but a constraint of "Null together" or "Not Null together" on the models would probably simplify this.

ewdurbin (approved):
approved with a bit of a flag on the modeling of rate limit count/period columns

alanbato:
Thank you kindly @JacobCoffee and @ewdurbin for picking this up again for review, and it seems this time it was I who was late to the party 😅 

@JacobCoffee It seems you have addresses most of Ee's comments and brought the PR in sync with main again. I assume there's no other action from my part here now? Thanks for the help :)

warehouse/organizations/models.py
JacobCoffee: [b1d108e](https://github.com/pypi/warehouse/pull/20324/commits/b1d108e983318846986f5bfa1465707a42f956c1), [3659240](https://github.com/pypi/warehouse/pull/20324/commits/365924039c311e5fd991eb1d248ef54b72c4d283), and [2a66116](https://github.com/pypi/warehouse/pull/20324/commits/2a66116bdb00850686179d9f0edf9f8636b7ddda) should constrain these!

PR #20325, opened 2026-07-24
Clean up duplicated `validate_password()` definition on `ReauthenticateForm`

### LINKED ISSUE #20321: Deduplicate `validate_password` between `PasswordMixin` and `ReAuthenticateForm`
After #20126, `ReAuthenticateForm.validate_password` is a near-copy of `PasswordMixin.validate_password`. The security fix needed the re-auth form to check against the server-supplied `self.user_id` rather than a client-submitted username, and the simplest way to do that was to override the whole method.

The result is two methods that are identical except for one line. Both call `check_password`, record the same `account:*:failure` event, and raise the same `TooManyFailedLogins` lockout message. The only real difference: `PasswordMixin` resolves the user via `find_userid(self.username.data)`, and `ReAuthenticateForm` uses `self.user_id` directly.

The worry is drift. Change the lockout wording or the event tags in one method and it's easy to miss the other.

One way to fix it: pull the user-id lookup into a small overridable method on `PasswordMixin` and let both forms share a single `validate_password`. The 2FA forms already lean on `self.user_id`, so they might fold in as well.

No behavior change here, so the current tests should stay green. Low priority — #20126 is shipped and correct; this is just cleanup.

alanbato:
Issue #20321

Also created a `UserIdMixin` used by `ReathenticateForm`, but also allow us to remove the `__init__` methods of the 2FA forms.


warehouse/accounts/forms.py
```diff
@@ -174,6 +174,16 @@ def validate_username(self, field):
...
 
 
+class UserIdMixin:
+    def __init__(self, *args, user_id, user_service, **kwargs):
+        self.user_id = user_id
+        self.user_service = user_service
+        super().__init__(*args, **kwargs)
+
+    def get_user_id(self):
+        return self.user_id
```
miketheman, line 202: suggestion: add type hints for new classes and mixins, so we can prevent weirdness later. This applies to new classes, functions, etc.

warehouse/accounts/forms.py
```diff
@@ -194,11 +204,14 @@ def __init__(
...
+    def get_user_id(self):
+        return self.user_service.find_userid(self.username.data)
+
     def validate_password(self, field):
         if field.errors:
             return
 
-        userid = self.user_service.find_userid(self.username.data)
+        userid = self.get_user_id()
         if userid is not None:
```
miketheman, line 241: Now that both are using the same `validate_password` function, this clause is now being used where it wasn't before.
Now a `None` user_id would silently skip the password check instead of validating.

Maybe a guard on `UserIdMixin.__init__` would work? Exmaple:

```python
def __init__(self, *args, user_id, user_service, **kwargs):
    if user_id is None:
        raise ValueError("user_id is required")
	...
```

A test with `user_id=None` might help expose this

miketheman (changes requested):

warehouse/accounts/forms.py
alanbato: You're right, it would skip the check and pass silently!
No current call site can pass a `user_id=None` for the `ReAuthenticateForm`, and on `main` a `None` would break the form loudly I think.

Needless to say, authentication code should be coded defensively, so adding a guard is a very good idea.

alanbato:
Thanks for the review @miketheman! I've addressed both comments :)

miketheman (approved):
Thanks for working through this!

PR #20326, opened 2026-07-24
Add `Release.provenance_status`

### LINKED ISSUE #20303: Surface release provenance states for templates to consume
## Background

As part of the #19950, we need a way to surface the provenance "states" of each release.

Specifically, we need to support the following:

- No provenance
- Full provenance
- Partial provenance
- Lost provenance
- Changed provenance
- Inconsistent provenance

Note that each release can have multiple simultaneous states, e.g.

- A release that has "Lost provenance" would also have "No provenance"
- A release may have "Full provenance" but "Inconsistent provenance"

This is not dissimilar to lifecycle states, where a release can be archived, yanked, and quarantined at the same time. In this case, we expose all states to the template, and use template logic to highlight the most critical information to the user - I suggest we take a similar approach here.

### [No Provenance](https://claude.ai/public/artifacts/e5e2389c-42a8-4c92-b89a-0709d142f669)
- **Definition:** No files in the release have provenance information/attestation.
- **Data required:** file count, e.g. `0 of 10 files have provenance`

### [Full Provenance](https://claude.ai/public/artifacts/cb20a310-3edf-4a94-b701-49c419152fef)
- **Definition:** All files in the release have provenance information/attestation.
- **Data required:** file count, e.g. `10 of 10 files have provenance`

### [Partial Provenance](https://claude.ai/public/artifacts/5fe8167d-2861-421f-abde-7758c00af616)
- **Definition:** Some files have provenance, others do not.
- **Data required:** file count, e.g. `5 of 10 files have provenance`

### [Lost Provenance](https://claude.ai/public/artifacts/64df197f-8e37-422b-abdb-288ef71ce587)
- **Definition:** A previous release had partial or full provenance; this release has none.
- **Data required:**
  - File count for current release, e.g. `0 of 10 files have provenance`
  - File count for previous release, e.g. `10 of 10 files had provenance`

### [Changed Provenance](https://claude.ai/public/artifacts/be9f0886-db3d-493d-a66e-c0251b03a4af)
- **Definition:** A previous release has different provenance information than this one.
- **Data required:**
  - Publishing/build repository — current release
  - Publishing/build repository — previous release
  - Diff between current and previous repository URL
  - Publishing/build configuration — current release
  - Publishing/build configuration — previous release
  - Diff between current and previous configuration

### [Inconsistent Provenance](https://claude.ai/public/artifacts/362f331e-b727-40af-b700-13ff8db3db8c)
- **Definition:** Files *within this release* have differing provenance (repository and/or configuration).
- **Data required:**
  - Each distinct publishing/build repository, with count of files attributed to it
  - Each distinct publishing/build configuration, with count of files attributed to it

## Open questions
**What constitutes a "Previous release"?**

I suggest a release published within 14 days of the current release (this matches [pypi/warehouse#19727](https://github.com/pypi/warehouse/pull/19727)).

My thinking:
1. We should use a time delta instead of "last release" or "n releases", because a malicious actor could simply upload more than one dodgy release to bypass this warning
2. 14 days should be sufficient time for the community and/or PyPI team to flag a malicious change

di:
This PR adds `Release.provenance_status` to address #20303. When `Release.provenance_status` is present, templates can reference the following fields and properties:

### Active Provenance States
A set of string enum values representing all active attestation conditions for the release. Templates can check for specific states using standard Jinja2 in checks, e.g.:

```jinja2
{% if 'inconsistent-provenance' in release.provenance_status.states %}
```

The values are:

* `no-provenance`: 0 files in the release have attestation bundles.
* `full-provenance`: 100% of files in the release have attestation bundles.
* `partial-provenance`: Some, but not all, files in the release have attestation bundles.
* `inconsistent-provenance`: Files within this single release originate from more than one distinct repository URL or more than one distinct workflow filename.
* `lost-provenance`: The current release has 0 files with provenance, but a preceding release published within the last 14 days had attested files.
* `changed-provenance`: Both the current release and the preceding comparison release have provenance, but their sets of source repositories or workflows differ.

### File Count Ratios
Useful for rendering progress badges or ratio summaries (e.g., "3 of 4 files attested"):

* `Release.provenance_status.files_with_provenance` (`int`): Number of files in the release with attestation bundles.
* `Release.provenance_status.total_files` (`int`): Total number of files in the release.

### Source Distribution Maps
Dictionaries mapping source URLs and workflow filenames to file counts, useful for rendering repository links, workflow badges, and detailing 'inconsistent-provenance' warnings:

* `Release.provenance_status..repository_counts` (`dict[str, int]`): Maps source repository URLs (e.g., `{'https://github.com/org/repo': 2}`) to the number of files attested from each.
* `Release.provenance_status..workflow_counts` (`dict[str, int]`): Maps workflow paths/filenames (e.g., `{'publish.yml': 2}`) to the number of files attested from each.

### Comparison Release Context (for Regression Warnings)
When `lost-provenance` or `changed-provenance` is present in `Release.provenance_status.states`, these fields provide baseline context to render comparative warning cards or diff badges:

* `Release.provenance_status.comparison_release` (`Release | None`): The baseline `Release` object used for comparison (e.g., to render a link: "Compared to v1.0.0").
* `Release.provenance_status.comparison_files_with_provenance` (`int | None`): File counts from the baseline release
* `Release.provenance_status.comparison_total_files` (`int | None`): File counts from the baseline release
Source Delta Properties (set[str]):
* `Release.provenance_status.added_repositories` (`set[str]`): Sets of repository URLs added between the comparison baseline and the current release.
* `Release.provenance_status.removed_repositories` (`set[str]`): Sets of repository URLs removed between the comparison baseline and the current release.
* `Release.provenance_status.added_workflows` (`set[str]`): Sets of workflow filenames added or removed between the comparison baseline and the current release.
* `Release.provenance_status.removed_workflows` (`set[str]`): Sets of workflow filenames removed between the comparison baseline and the current release.

Fixes #20303.


miketheman (approved):

nlhkabu:
Hi @di 

I'm currently using the provenance status functionality on #20408, and I have a question regarding the added and removed sources/configurations.

Currently, `{{ release.provenance_status.removed_sources }}` returns an object like `PublisherSource(kind='GitHub', identity='seed-org/alpha')`. This provides enough context for me to build the full URL for each removed source by hardcoding the publisher URL. (The same applies to added sources.)

However, `{{ release.provenance_status.removed_workflows }}` currently returns only the filename as a string (e.g., `'release.yml'`).

How would you recommend building the full URL for workflows here?
* Is it safe to assume the `kind` from `removed_sources` applies to `removed_workflows` in all cases? (same for added)?
* Or could we theoretically have a situation where the removed source and workflows are on different platforms (e.g., one is GitHub and one is GitLab)?

I wonder whether we could instead update this to return `viewer.source` and `viewer.workflow_url`, similar to how it's handled in [PEP740AttestationViewer](https://github.com/pypi/warehouse/blob/6b62157124ab7865029afeeadf282bce27d601d9/warehouse/packaging/views.py#L32)? That might save me from having to reconstruct the URLs manually.

nlhkabu:
Also, separately, do we want to handle changes in `publisher.email` for `Google` publisher types?

di:
> Is it safe to assume the kind from removed_sources applies to removed_workflows in all cases? (same for added)?

It's an extreme edge case, but no, I think it could be ambiguous if there were multiple sources removed, we wouldn't know which workflow to attribute to which source.

> Or could we theoretically have a situation where the removed source and workflows are on different platforms (e.g., one is GitHub and one is GitLab)?

That could happen, we could also have two removed sources from the same platform (like two different GitHub repos publishing to the same PyPI project -- probably rare, but possible)

> I wonder whether we could instead update this to return viewer.source and viewer.workflow_url, similar to how it's handled in [PEP740AttestationViewer](https://github.com/pypi/warehouse/blob/6b62157124ab7865029afeeadf282bce27d601d9/warehouse/packaging/views.py#L32)? That might save me from having to reconstruct the URLs manually.

Yes, I think that would make sense, right now `removed_workflows` doesn't seem very useful.

> Also, separately, do we want to handle changes in publisher.email for Google publisher types?

Yes!

nlhkabu:
Thanks @di 

FYI, right now I've just output the data like this in https://github.com/pypi/warehouse/pull/20408

image

image

PR #20327, opened 2026-07-25
Deduplicate validate_password between PasswordMixin and ReAuthenticateForm

### LINKED ISSUE #20321: Deduplicate `validate_password` between `PasswordMixin` and `ReAuthenticateForm`

pranjali2076:
Fixes #20321

### Summary of Changes:
- Added `_get_user_id()` helper method on `PasswordMixin` to encapsulate user lookup.
- Overrode `_get_user_id()` in `ReAuthenticateForm` to return `self.user_id`.
- Removed the duplicate `validate_password` implementation in `ReAuthenticateForm` so it inherits directly from `PasswordMixin`.

### DIFF
--- a/warehouse/accounts/forms.py
+++ b/warehouse/accounts/forms.py
@@ -194,12 +194,17 @@ class PasswordMixin:
         self._check_password_metrics_tags = check_password_metrics_tags
         super().__init__(*args, **kwargs)
 
+    def _get_user_id(self):
+        """Helper to resolve the user ID to check against."""
+        return self.user_service.find_userid(self.username.data)
+
     def validate_password(self, field):
         if field.errors:
             return
 
-        userid = self.user_service.find_userid(self.username.data)
+        userid = self._get_user_id()
         if userid is not None:
+
             try:
                 if not self.user_service.check_password(
                     userid,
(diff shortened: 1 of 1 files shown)


JacobCoffee:
#20327 was opened first

PR #20329, opened 2026-07-27
2026 UI update release history

nlhkabu:
1. Adds all states (pre-release, archived, yanked, quarantined) to release history cards
2. Replaces state badges with (more obvious) card header
3. Adds file count metadata to each release card - this links directly to the file UI for each release
4. TODO: once the security page is build, we will also surface attestation status here.  This can be done in a later PR.

## Desktop

image

image

image

Screen Shot 2026-07-27 at 08 21 24

Screenshot 2026-07-27 at 08-26-03 sampleproject · Warehouse

## Mobile

image


warehouse/packaging/models.py
nlhkabu, line 1: I believe that there may be a better way to expose this data to the template???

di: Yes, #20326 introduces `Release.provenance_status.counts.total_files`, we should probably merge that first.

warehouse/packaging/models.py
```diff
@@ -460,14 +460,23 @@ def maintainers(self):
     @property
     def all_versions(self):
         session = orm_session_from_obj(self)
+        file_count = (
+            session.query(func.count(File.id))
+            .filter(File.release_id == Release.id)
+            .correlate(Release)
+            .scalar_subquery()
+        )
```
di: We should merge #20326 first instead of this change.

warehouse/templates/packaging/detail.html
```diff
@@ -25,8 +25,8 @@
         {% include "warehouse:templates/includes/packaging/project-description.html" %}
         {% if files %}
           {% include "warehouse:templates/includes/packaging/project-files.html" %}
+          {# {% include "warehouse:templates/includes/packaging/project-security.html" %} #}
         {% endif %}
-        {# {% include "warehouse:templates/includes/packaging/project-security.html" %} #}
```
di, line 39: This is commented out, should it just be removed?

di (approved):
LGTM!

JacobCoffee (approved):
lgtm, curious what causes the diff. icons in your screenshot - im not seeing it in the code or on testpypi

Image

warehouse/packaging/models.py
miketheman: This this change in play, are these values still used anywhere later?

nlhkabu:
@JacobCoffee a couple of the screenshots are stale.

The correct icon is the download arrow in a circle, as on TestPyPI.

nlhkabu:
Note: I have introduced a small change on the mobile UI - compressing the margin between cards and removing unnecessary padding:

image

warehouse/templates/includes/packaging/project-release-history.html
```diff
@@ -14,10 +14,25 @@ <h2 class="page-title split-layout">
   <div class="release-timeline">
     {% set blue_cube = request.static_url('warehouse:static/dist/images/blue-cube.svg') %}
     {% set white_cube = request.static_url('warehouse:static/dist/images/white-cube.svg') %}
+    {% set provenance_statuses = {} %}
+    {% for r in project.releases %}
+      {% set _ = provenance_statuses.update({r.version: r.provenance_status}) %}
+    {% endfor %}
```
miketheman: I suspect this is an N+1 query perf problem.

This iteration appears to only beed this iteration to get a file count, whihc should now exist on `provenance_status.counts.total_files` and others - see https://github.com/pypi/warehouse/blob/1b7402c23b086002085952d43499bc4f69af0c0f/warehouse/attestations/models.py#L82-L98

The property call costs ~3 queries per release, including a `Provenance` join that pulls every attestation JSON into memory and parses each bundle through `get_provenance_sources`. It's cached per-release only.

So this page goes from one lightweight `all_versions` row query to O(3N) queries plus N full ORM rows: ~800 queries for a 270-release project, and roughly 20k queries with thousands of parsed in-toto bundles for something like `boto3`.

warehouse/templates/includes/packaging/project-release-history.html
```diff
@@ -32,28 +47,72 @@ <h2 class="page-title split-layout">
...
+        <div class="release__card">
+          {% if hversion.lifecycle_status == "quarantine-enter" or project.lifecycle_status == "quarantine-enter" %}
+            <div class="release__card-header">
+              {% if hversion.lifecycle_status == "quarantine-enter" and hversion.lifecycle_status_changed %}
+                {% set quarantine_title %}
+                  {% trans date=hversion.lifecycle_status_changed|format_date %}Release quarantined on {{ date }}{% endtrans %}
+                {% endset %}
+              {% elif project.lifecycle_status_changed %}
+                {% set quarantine_title %}
+                  {% trans date=project.lifecycle_status_changed|format_date %}Project quarantined on {{ date }}{% endtrans %}
```
miketheman: I suspect this would misreport dates if the following conditions are true (which does happen):

1. project entered quarantine
2. project's latest release is placed in quarantine
3. project is cleared from quarantine

So in this case, the project itself is no longer quarantined, but the **release** is, and it has its own set of dates that are relevant, and the label should reflect whether it's a release or a project. Maybe that's here or in key dates fragment (or is that duplicated?)

warehouse/templates/includes/packaging/project-release-history.html
```diff
@@ -32,28 +47,72 @@ <h2 class="page-title split-layout">
...
+              <span class="release__date"><i class="fa-regular fa-calendar" aria-hidden="true"></i> {{ humanize(hversion.created) }}</span>
+              {% if provenance_statuses[hversion.version] %}
+                <span class="release__meta-divider" aria-hidden="true"></span>
+                <a href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}#files"><i class="fa-regular fa-circle-down" aria-hidden="true"></i> {% trans count=provenance_statuses[hversion.version].counts.total_files %}{{ count }} file
+                {% pluralize %}
+                {{ count }} files{% endtrans %}</a>
+              {% endif %}
+              <!-- TODO: Add provenance status
+              <span class="release__meta-divider" aria-hidden="true"></span>
+              <a href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}#security"><i class="fa-solid fa-circle-check release__provenance-icon" aria-hidden="true"></i> {% trans %}Full provenance{% endtrans %}</a> -->
```
miketheman: This code will run and emit to translations due to uit being an HTML comment, not a Jinja comment - is that desirable?

warehouse/templates/includes/packaging/project-release-history.html
nlhkabu: Updated here.
Opened an issue for key dates: https://github.com/pypi/warehouse/issues/20398

nlhkabu:
Explicitly label project vs release quarantined:

image

image

warehouse/templates/includes/packaging/project-release-history.html
nlhkabu: @di need your help here, please!

nlhkabu: To be clear - I was also planning on exposing provenance status (no provenance, full provenance, changed provenance, etc) on this page, as per this mockup:

image

_However_ if we find that displaying file count and provenance _per release_ is too expensive, I can change the design to show only the release date as metadata, as per the current page in production.

JacobCoffee: +1 on keeping file count/slsa stuff, it looks nice

di: https://github.com/pypi/warehouse/pull/20410 should resolve this, after merging this can go away:

```suggestion
```

(we will need to do some follow-on work to support provenance statuses as well)

warehouse/templates/includes/packaging/project-release-history.html
```diff
@@ -32,28 +47,82 @@ <h2 class="page-title split-layout">
...
+              <a href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}">{{ hversion.version }}</a>
+              {% if is_current %}
+                <span class="badge">{% trans %}This release{% endtrans %}</span>
+              {% endif %}
+            </p>
+            <p class="release__meta">
+              <span class="release__date"><i class="fa-regular fa-calendar" aria-hidden="true"></i> {{ humanize(hversion.created) }}</span>
+              {% if provenance_statuses[hversion.version] %}
+                <span class="release__meta-divider" aria-hidden="true"></span>
+                <a href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}#files"><i class="fa-regular fa-circle-down" aria-hidden="true"></i> {% trans count=provenance_statuses[hversion.version].counts.total_files %}{{ count }} file
```
di:
```suggestion
              {% if provenance_counts[hversion.version] %}
                <span class="release__meta-divider" aria-hidden="true"></span>
                <a href="{{ request.route_path('packaging.release', name=release.project.name, version=hversion.version) }}#files"><i class="fa-regular fa-circle-down" aria-hidden="true"></i> {% trans count=provenance_counts[hversion.version].total_files %}{{ count }} file
```

warehouse/templates/includes/packaging/project-release-history.html
miketheman: #20420 should solve this, and has a an example of the template changes

di (approved):
Seems to be working fine on TestPyPI!

di:
(@nlhkabu linting is failing)

PR #20339, opened 2026-07-29
Rework Org app UI

JacobCoffee:
- Add header banner to show overall state
- Add hotkeys for quick buttons
- Add quick reply buttons on the new header banner to quickly open modal -> fill in correct reply
- Add domain checks for various things like public sites such as code forges
- ~~Add domain checks for OFAC compliance~~
- ~~Add quick reply for OFAC non-compliant orgs to add admin note & defer~~
- Nest Orgs under a tree: Orgs/{Orgs, Applications} to help with sizing issue but also to section it off nicer

image
image
image

### DIFF
--- a/warehouse/organizations/checks.py
+++ b/warehouse/organizations/checks.py
@@ -1,0 +1,337 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Automated review checks for organization applications.
+
+Every check is advisory: nothing here approves or declines an application. A check
+returns None when it does not apply, and CheckStatus.Unknown when it applies but
+cannot reach an answer.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import enum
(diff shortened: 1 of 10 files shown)


warehouse/organizations/checks.py
```diff
@@ -0,0 +1,313 @@
...
+    MIN_SUBSTRING_LENGTH,
+    NAME_SIMILARITY_THRESHOLD,
+    OPEN_APPLICATION_STATUSES,
+)
+
+if TYPE_CHECKING:
+    from warehouse.accounts.models import User
+    from warehouse.organizations.models import OrganizationApplication
+
+_extractor = TLDExtract(suffix_list_urls=())
```
JacobCoffee, line 31: i pulled out the OFAC scan stuff for now im not that confident on it and it wasnt super accurate

PR #18843, opened 2025-10-10
feat(logging): celery structure logs

JacobCoffee:
Split off from https://github.com/pypi/warehouse/pull/18748

```
worker-1        | {"event": "beat: Acquired lock", "timestamp": "2025-10-10T20:42:30.035157Z", "level": "info", "dd.env": "development", "dd.version": "unknown"}
worker-1        | {"event": "mingle: all alone", "timestamp": "2025-10-10T20:42:31.034605Z", "level": "info", "dd.env": "development", "dd.version": "unknown"}
worker-1        | {"event": "celery@7e55216e877e ready.", "timestamp": "2025-10-10T20:42:31.043095Z", "level": "info", "dd.env": "development", "dd.version": "unknown"}
```

structures celery logs
(inject dd context will do this in another pr) 

based on https://www.structlog.org/en/stable/frameworks.html#celery

pairs with https://github.com/pypi/warehouse/pull/18844

PR #20341, opened 2026-07-29
feat: unify log rendering through ProcessorFormatter

### LINKED ISSUE #16246: Use structlog.dev.ConsoleRenderer when in dev mode
We configure the logging renderer here:

https://github.com/pypi/warehouse/blob/9a25f6975d3870bd8d6f9839af98b4a0c7b0e2db/warehouse/logging.py#L21

Explore using https://www.structlog.org/en/stable/console-output.html in local development mode, and if this mode increases clarity when running multiple containers and processes.

miketheman:
One formatter for every record: structlog, gunicorn, celery, and stdlib loggers all render as console columns in dev and JSON in prod. A tween binds request.id and celery signals bind task_id/task_name into contextvars, so ambient context reaches every log line, and can be tracked across log lines for a given request/task.

Access logs switch from `access_log_format` to `GunicornLogger`, emitting `http_request` events with queryable fields, including the `request.id` of the app logs from the same request. Their shape changes from Apache combined format to JSON, so update log consumers that parse the old format.

Resolves #16246
Closes #18843
Closes #18844

### DIFF
--- a/warehouse/tasks.py
+++ b/warehouse/tasks.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import functools
 import hashlib
-import logging
 import time
 import typing
 import urllib.parse
@@ -16,9 +15,11 @@ import celery.app.backends
 import celery.backends.redis
 import pyramid.scripting
 import pyramid_retry
+import structlog
 import transaction
 import venusian
 
+from celery import signals
 from kombu import Queue
 from pyramid.threadlocal import get_current_request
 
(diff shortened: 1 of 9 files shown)


miketheman:
The intent is to follow this with a bulk-change of logging calls once landed and confirmed to use structured logging, and include a dev doc page which describes the way to use logging in warehouse.

warehouse/tasks.py
```diff
@@ -303,10 +314,16 @@ def includeme(config: Configurator) -> None:
         REDBEAT_REDIS_URL=s["celery.scheduler_url"],
         # Silence deprecation warning on startup
         broker_connection_retry_on_startup=False,
+        # Keep the logging configured by warehouse.logging instead of letting
+        # celery replace the root logger's handlers at worker boot.
+        worker_hijack_root_logger=False,
```
ewdurbin, line 319: huh, i had no idea celery was so greedy.

ewdurbin (approved):
does what it says on the tin, looks great in dev and i also validated JSON output by forcing the console formatter off.

warehouse/tasks.py
miketheman: I know, it's not celery, but it's the closest I could recall

PR #20342, opened 2026-07-29
feat(dev): add a `warehouse db seed` command

miketheman:
Seeds one recognizable project per release provenance state (full PyPI/SLSA/both, partial, inconsistent, lost, changed) so UI work has consistent data to render.

The factories do the heavy lifting.
`ProvenanceFactory` gains optional predicate_types/repository/workflow params that build valid PEP 740 provenance through the `pypi_attestations` models, with self-signed certs carrying the Fulcio claim OIDs the file details page decodes.

Two quirks: stale `file_registry` rows are cleared first so re-seeding works after deleting the projects, and seeded projects get warmed before commit because the search reindex `after_commit` hook can't refresh expired instances.

Refs #20303


JacobCoffee (approved):
id also like to add on to this eventually for various seed data for org state

miketheman:
> id also like to add on to this eventually for various seed data for org state

Indeed - I hope that one day we can completely drop the dev db sql file and seed the database after migrations have been applied. 

In the meantime, if you can express the different orgs and states you want, they can all follow a similar pattern and live under the established  `seed-` namespace.

PR #20355, opened 2026-08-05
Add `repo_property_python_gar_access` unchecked claim

di:
Fixes [WAREHOUSE-PRODUCTION-2J8](https://python-software-foundation.sentry.io/issues/7560333845/events/d9c676820e9c4144ad3101f3fdad5efc/).

### DIFF
--- a/warehouse/oidc/models/github.py
+++ b/warehouse/oidc/models/github.py
@@ -186,6 +186,7 @@ class GitHubPublisherMixin:
         "enterprise_id",
         "ref_protected",
         "check_run_id",
+        "repo_property_python_gar_access",
     }
 
     # Get the most specific publisher from a list of publishers,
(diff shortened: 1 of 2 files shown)


warehouse/oidc/models/github.py
```diff
@@ -186,6 +186,7 @@ class GitHubPublisherMixin:
         "enterprise_id",
         "ref_protected",
         "check_run_id",
+        "repo_property_python_gar_access",
```
woodruffw, line 189: Based on https://docs.github.com/en/actions/reference/security/oidc we might need a generic `repo_property_*` pattern in the long term.

This specifically: https://docs.github.com/en/actions/reference/security/oidc#including-repository-custom-properties-in-oidc-tokens

woodruffw (approved):

warehouse/oidc/models/github.py
di: Ah, good catch, probably shouldn't have auto-merged this! 😅

di: https://github.com/pypi/warehouse/pull/20356

PR #20356, opened 2026-08-05
Handle unchecked wildcard claims

di:
Better version of #20355.


woodruffw (approved):
LGTM! Kind of sad that we have another dunder but that's my fault 😅 

(Another option would be to have `__unchecked_claims__` be `str | Callable[[] -> bool]` but that IMO that's probably way too general.)

PR #20358, opened 2026-08-07
docs: Update docs for updated "good first issue" label

ulgens:
The label name was changed to "good first issue for humans"

---

Coming from https://warehouse.pypa.io/development/getting-started/, one of the very first links 

> Then, look at our [open issues that are labelled "good first issue"](https://github.com/pypi/warehouse/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22)

goes to a filtered GitHub Issues page that fails with

Screenshot 2026-08-07 at 03 23 12 UTC@2x

It seems the label name was changed at some point in the past.


woodruffw (approved):
Thanks!

PR #20362, opened 2026-08-07
forklift: add invalid entrypoint metrics

woodruffw:
This was intended to be my "fast follow" to https://github.com/pypi/warehouse/pull/20081, sorry...

~~WIP, needs another testcase for the interpolated entrypoint scenario.~~ Removed this.


warehouse/utils/wheel.py
```diff
@@ -283,7 +283,9 @@ class CaseSensitiveConfigParser(configparser.ConfigParser):
         raise InvalidWheelEntryPointsError("entry_points.txt is not decodable as UTF-8")
 
     # The Entry Points specification requires `=` as the delimiter.
-    parser = CaseSensitiveConfigParser(delimiters=("=",))
+    # We also disable interpolation. EPS doesn't specifically forbid it but it
+    # serves no purpose in `entry_points.txt` other than introducing malleability.
+    parser = CaseSensitiveConfigParser(delimiters=("=",), interpolation=None)
```
woodruffw: Actually, I'm not sure we want this -- disabling interpolation doesn't _reject_ the interpolation syntax, it just passes it through unmutated. That's arguably wrong since downstreams will then interpret it, and _will_ accept a post-interpolated form that we _should_ have rejected.

I'm gonna remove that for now.

woodruffw: Undone.

miketheman (approved):
Coverage definitely exercises this code path, but it's always better to add an assertion if we expect this to continue working if someone else refactors the code.
Good enough for now!

PR #20364, opened 2026-08-08
Use the caveats stored in the db as additional caveats for Macaroon verification

dstufft:
Traditionally all caveats that exist for a Macaroon get embedded into the Macaroon, inflating the size and making it impossible for the system/us to further restrict the Macaroon once it's been created.

This isn't a required thing for Macaroons though, it's just a choice we made when implementing them, so this PR adjusts that choice so that instead of embedding the "system generated" caveats into the Macaroon that we give the user, we instead *only* store them in the database. When we go to verify the macaroon, we pull out any caveats that are stored in the database and dynamically append them to the Macaroon before we verify it.

This ends up meaning that the logical set of caveats changes from what is embedded in the Macaroon, to additionally include what is stored in the database.

The Macaroons stored in the database will exist in a few states:

1. Ones that predate the caveats column but whose caveats were able to be solely reflected by the legacy permissions caveat.
2. Ones that predate the caveats column but whose caveats were not able to be solely reflected by the legacy permissions caveat.
3. Ones that have been generated since the caveats column was introduced.
4. Ones that will have been generated after this change.

All of these cases will be correctly handled:

1. These macaroons should have had their system generated caveats synthesized from the legacy permissions caveat when the caveats column was added, so they will have the correct caveats stored in the caveats column, but it doesn't matter because the correct ones are embedded in the macaroon and the database caveats add to not replace the embedded caveats.
2. These macaroons do not have their system generated caveats correctly reflected in the caveats column, but it doesn't matter because the correct ones are embedded in the macaroon and the database caveats add to not replace the embedded caveats.
3. These macaroons will have their system generated caveats exist in both the database *and* embedded, so ultimately those caveats will exist twice when the macaroon is verified, but that is fine (although duplicative) because it's just asserting the same thing twice.
4. These macaroons will only have their system generated macaroons exist in the database, and the macaroon itself will not have any caveats included by the system (but they could have had ones added by the users).

In all cases, if the user has added any caveats to the macaroon, those will still be honored because the database caveats are added to whatever is embedded in the macaroon, so both "sources" of caveats are honored.

~~This PR still needs the tests updated, and we might want to still emit macaroons that have the caveats embedded to give the verification side of things some time to get run through its paces, because once we stop emitting the embedded caveats we can't revert this change or else those caveats won't be trusted.~~

### DIFF
--- a/tests/unit/macaroons/test_services.py
+++ b/tests/unit/macaroons/test_services.py
@@ -297,6 +297,10 @@ class TestDatabaseMacaroonService:
             user_id=user.id,
         )
 
+        dm = macaroon_service.find_from_raw(raw_macaroon)
+        # Add a database only caveat that has not been embedded into the macaroon
+        dm.caveats = [*dm.caveats, caveats.Expiration(expires_at=5, not_before=2)]
+
         verify = mocker.patch.object(
             caveats, "verify", autospec=True, return_value=True
         )
@@ -307,8 +311,29 @@ class TestDatabaseMacaroonService:
 
         assert macaroon_service.verify(raw_macaroon, request, context, permissions)
         verify.assert_called_once_with(
-            mocker.ANY, mocker.ANY, request, context, permissions
+            mocker.ANY, dm.key, request, context, permissions
         )
+
+        # Ensure that the macaroon that was verified is what was expected.
+        vm = verify.call_args.args[0]
+        assert vm.location == "fake location"
+        assert vm.identifier == str(dm.id).encode("utf8")
+        assert [c.to_dict() for c in vm.caveats] == [
+            # The embedded RequestUser caveat
+            {
+                "cid": f'[3,"{user.id!s}"]',
+                "cl": None,
+                "vid": None,
+            },
+            # The database stored RequestUser caveat
+            {
+                "cid": f'[3,"{user.id!s}"]',
+                "cl": None,
+                "vid": None,
+            },
+            # The database stored Expiration caveat
+            {"cid": "[0,5,2]", "cl": None, "vid": None},
+        ]
 
     def test_delete_macaroon(self, user_service, macaroon_service):
         user = UserFactory.create()
--- a/warehouse/macaroons/models.py
+++ b/warehouse/macaroons/models.py
@@ -71,13 +71,13 @@ class Macaroon(db.Model):
         JSONB,
         server_default=sql.text("'{}'"),
         comment=(
-            "The list of caveats that were attached to this Macaroon when we "
-            "generated it. Users can add additional caveats at any time without "
-            "communicating those additional caveats to us, which would not be "
-            "reflected in this data, and thus this field must only be used for "
-            "informational purposes and must not be used during the authorization "
-            "or authentication process. Older Macaroons may be missing caveats as "
-            "previously only the legacy permissions caveat were stored."
+            "The list of system generated caveats for this Macaroon. Users can add "
+            "additional caveats at any time without communicating those additional "
+            "caveats to us, which would not be reflected in this data, and thus this "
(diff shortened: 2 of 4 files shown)


tests/unit/macaroons/test_services.py
```diff
@@ -307,9 +311,30 @@ def test_verify_valid_macaroon(self, mocker, macaroon_service):
...
         verify.assert_called_once_with(
-            mocker.ANY, mocker.ANY, request, context, permissions
+            mocker.ANY, dm.key, request, context, permissions
         )
 
+        # Ensure that the macaroon that was verified is what was expected.
+        vm = verify.call_args.args[0]
+        assert vm.location == "fake location"
+        assert vm.identifier == str(dm.id).encode("utf8")
+        assert [c.to_dict() for c in vm.caveats] == [
```
dstufft, line 321: This is a strict improvement to this test, before we weren't actually testing that creating a macaroon and verifying it actually ended up having the correct caveats in it or that we were verifying the correct macaroon at all, but now we are!

warehouse/macaroons/services.py
```diff
@@ -209,6 +246,8 @@ def create_macaroon(
             key=dm.key,
             version=pymacaroons.MACAROON_V2,
         )
+        # TODO: Remove this to stop emitting embedded caveats, which are now
+        #       being stored in the database while still being verified.
```
dstufft, line 250: Whatever PR stops emitting these, we cannot revert (because once we stop emitting them, we're relying on the database stored ones being verified for any macaroons generated after that fact). So this PR continues to emit them, to give us time to make sure there's no unexpected problems with verifying the ones stored in the database.

A future PR can remove these two lines (well and the comment) and then we will no longer be emitting these for new macaroons.

warehouse/macaroons/models.py
```diff
@@ -71,13 +71,13 @@ class Macaroon(db.Model):
         JSONB,
         server_default=sql.text("'{}'"),
```
woodruffw, line 72: Just noting: we're treating this JSONB as a list, but the `server_default` is an empty dict. I suspect we never actually hit this default in practice, but perhaps we should update to:


```suggestion
        server_default=sql.text("'[]'"),
```

woodruffw (approved):
LGTM! One comment about the server default but I think it's not important in practice.

(Writing a summary below to make sure my understanding is correct.)

We currently store macaroon caveat state in two places: the serialized macaroon itself (as normal first-party caveats), and in the "DB macaroon" (i.e. our DB copy of each macaroon).

Originally, we only stored the DB side for history/event-tracking purposes, and the user-originated macaroon was the only source of ground truth for caveats. This doesn't scale well, since our caveats are (currently) linear, i.e. grow at the same order as any restrictions/scopes we might want to impose. This causes pain especially when individual macaroons begin to pass sizes that are considered unreasonable for (e.g.) HTTP headers.

The solution is to tweak our source of truth. This PR doesn't _remove_ the serialized caveats, but allows us to _augment_ them with the copy already stored in the DB before we process them for authz purposes. At the moment, these two copies of the caveats should be identical, so appending them should have no effect.

In the future, this will allow us to emit minimal macaroons as PyPI tokens, since we'll re-add any DB caveats unconditionally. This _also_ preserves our nice property that PyPI tokens are user-scopeable, since users will still receive a valid macaroon that they can append third-party caveats to.

warehouse/macaroons/models.py
dstufft: I think this is a reasonable change, but I'm going to make it in a separate PR, just because this PR touches the auth logic of macaroons, and so I'd rather avoid introducing other unrelated changes in it (even if it's small and should be entirely harmless).

Issue #19759, opened 2026-03-30
ActiveState documentation links are broken.


captn3m0:
**Describe the bug**
Links to activestate docs on various places are broken

**Expected behavior**
Links should work

**To Reproduce**
1. Go to trusted publishing settings
2. Go to activestate
3. Click the read more link.

I get a 404 on https://docs.activestate.com/platform/user/oidc/

**My Platform**

Not really relevant.

**Additional context**

The docs link to https://docs.activestate.com/platform/start/pypi/ in [multiple places](https://github.com/search?q=repo%3Apypi%2Fwarehouse%20https%3A%2F%2Fdocs.activestate.com%2Fplatform%2Fstart%2Fpypi%2F&type=code) and that should be changed to https://docs.activestate.com/platform/open-source-collaboration/pypi/ as well

I could not figure out what the correct url for https://docs.activestate.com/platform/user/oidc/ is.

miketheman:
cc @th3coop

th3coop:
Hey all, I'm no longer with ActiveState and I can't seem to tag anyone from there in this comment.  Maybe I can only tag people in the pypi org?  I'll send an email to get someone to take a look at this.

th3coop:
I've forwarded the issue to ActiveState's current tech leads.

miketheman:
Thanks @th3coop ! Hope you're on to bigger&better.

josh-activestate:
Hi @miketheman thanks for bringing this to our attention. During recent documentation updates some articles were moved or combined for improved clarity. Apologies for any issues, let's get the right links for you. 

Seems like you need a link for: 
* [creating an account](https://docs.activestate.com/platform/user/create-account)
* [creating an activestate project](https://docs.activestate.com/platform/getting-started/create-project). 
* [general information about our open source PyPI collaboration](https://docs.activestate.com/platform/open-source-collaboration/). 

Are there any other links you're looking for?

miketheman:
Thanks @josh-activestate ! Feel free to send us a pull request fixing any of your links that currently 404.

miketheman:
One fix applied in #20381

shenanigansd:
Whoops, sorry for the churn guys; I didn't see this issue before making my PR.

I see three sets of links:
- https://docs.activestate.com/platform/start/pypi/
This one was easy, its clearly defined as https://docs.activestate.com/platform/open-source-collaboration/pypi/
- https://docs.activestate.com/platform/user/oidc/
[...]

PR #20390, opened 2026-08-13
docs: update ActiveState links for OIDC

shenanigansd:
Part of https://github.com/pypi/warehouse/issues/19759

Note: Yes, I know I'm changing URLs from "OIDC" to "MFA", but the page still has "OIDC" on it, and I emailed their support and they confirmed that this is the link to use.


miketheman (approved):
Thanks!

PR #20393, opened 2026-08-13
tests: avoid a bunch of huge allocations

woodruffw:
These allocations aren't actually asserting anything, and our actual "too big" file test is unchanged in `test_upload_fails_with_too_large_file`.

On my local machine, this reduces the peak test RSS (with 18 workers) from ~12GB to ~9.5GB.

(Note: the reason I investigated this is because I kept seeing `xdist` worker crashes, which I tracked down to OOM kills in Docker for Mac's Linux VM. Raising the memory cap on the VM fixed things, but I think saving memory here is also a good call 🙂.)

### DIFF
--- a/tests/unit/forklift/test_legacy.py
+++ b/tests/unit/forklift/test_legacy.py
@@ -2499,7 +2499,7 @@ class TestFileUpload:
                 }[filetype],
                 "content": pretend.stub(
                     filename=filename,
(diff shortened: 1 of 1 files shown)


JacobCoffee (approved):

miketheman (approved):
I can confirm a smaller dip - only 12 CPUs in use here.

Did you detect this via any profilers, and should those be added to the test suite to prevent regression?

PR #20399, opened 2026-08-14
feat(macaroons): measure caveat attenuation

miketheman:
Users can add caveats to the macaroons we issue whenever they like. We have no idea how often that happens, and we'd want to know before deciding on whether to continue to use this capability.

Once a macaroon verifies, diff its embedded caveats against the ones we stored when we issued it and count verified left over caveats.

Also counts caveats that still show up in the old mapping format, which is how we'll know when the legacy adapter can go.

### DIFF
--- a/warehouse/macaroons/caveats/__init__.py
+++ b/warehouse/macaroons/caveats/__init__.py
@@ -2,6 +2,7 @@
 
 import time
 
+from collections.abc import Sequence
 from typing import Any
 
 from pydantic import StrictInt, StrictStr
@@ -28,7 +29,14 @@ from warehouse.macaroons.caveats._core import (
 from warehouse.oidc.interfaces import SignedClaims
 from warehouse.packaging.models import Project
 
-__all__ = ["deserialize", "deserialize_obj", "serialize", "serialize_obj", "verify"]
+__all__ = [
+    "attenuations",
+    "deserialize",
+    "deserialize_obj",
+    "serialize",
+    "serialize_obj",
+    "verify",
+]
 
 
 # NOTE: Under the covers, caveat serialization is done as an array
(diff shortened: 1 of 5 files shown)


warehouse/macaroons/caveats/__init__.py
```diff
@@ -144,6 +152,46 @@ def verify(self, request: Request, context: Any, permission: str) -> Result:
...
+    An empty result means every embedded caveat is one of ours, which covers both
+    of the shapes we hand out: today we embed the caveats we issue, and if we ever
+    stop, a macaroon carrying none at all is just as unattenuated.
+
+    `issued` accounts for one embedded copy of each of its caveats, so a second
+    copy of one we issued is an attenuation. Comparison is by deserialized value
+    rather than by bytes, which keeps a caveat class gaining a defaulted field
+    from reading as an attenuation on every macaroon issued before the change.
+    """
+    unmatched = list(issued)
```
woodruffw: Extreme nitpick: could probably be a `set` for O(1) presence/removal.

warehouse/macaroons/services.py
```diff
@@ -132,6 +176,14 @@ def verify(self, raw_macaroon: str, request, context, permission) -> bool:
         if dm is None:
             raise InvalidMacaroonError("deleted or nonexistent macaroon")
 
+        issued = dm.caveats
+
+        # Work out which caveats the end user added themselves before we append our
+        # own stored caveats below, which would make the two indistinguishable. We
+        # don't record this until we know the macaroon verifies, so that anyone who
+        # reads an identifier out of a token can't skew the numbers with forgeries.
+        attenuated = caveats.attenuations(m, issued) if issued else None
```
woodruffw: Nitpick: `attenuated` reads to me as a boolean, maybe `attenuations` to emphasize it's a list of appended claims?

woodruffw (approved):
This makes sense to me overall, just some nitpicks about types/naming.

warehouse/macaroons/caveats/__init__.py
miketheman: Leads to: `TypeError: cannot use 'warehouse.macaroons.caveats.ProjectName' as a set element (unhashable type: 'list')`

PR #20400, opened 2026-08-14
Move language dropdown to header

JacobCoffee:
https://github.com/pypi/warehouse/pull/19803 had this, pulled it out for easier review.

The basis of #19803 is that the footer is too tall, cluttered, and i'm attempting to clean it up.

During that, noticing that the languages are all the way at the bottom (and added to the height of the footer) i thought it'd be a good idea to move the language chooser to the header. I think this is better UX for people landing here and scanning a bunch of english may be confusing until you realize you can change...all the way at the bottom.

Collapses the list into a nice dorpdown

No more in foooter:
image

Preview:
image


nlhkabu (approved):

warehouse/templates/base.html
```diff
@@ -343,7 +336,7 @@
              class="search-form__search"
              type="text"
              name="q"
-             placeholder="{% trans %}Search projects{% endtrans %}"
+             placeholder="{% trans %}Type '/' to search projects{% endtrans %}"
```
sethmlarson: Is this change related to the wider change of "moving the language dropdown to the header"?

sethmlarson (approved):
One comment about a seemingly unrelated change, was that intentional and discussed in another issue? Otherwise leaning on Nicole's review I'll leave a LGTM.

warehouse/templates/base.html
JacobCoffee: a small unrelated change while touching the site menu reforamtting nicole did [ea2475f](https://github.com/pypi/warehouse/pull/20400/commits/ea2475f4e90988e2990dc0a6b42263623db6b5e8) it seems

miketheman: why add "type /..." on mobile?

nlhkabu: Yes, you're right - we can remove this @JacobCoffee - apologies.

JacobCoffee: It ***is*** bringing it to be consistent witht he other search bar (it shows type / on dashboard). do we want to not make it consistent? people use keyboards on their small tablets/phones sometimes

JacobCoffee: (i dont mind either way, i was fine with it for consistency, maybe we can combine the two to always mirror one another though instead of maintaining 2)

nlhkabu: lets exclude it from this PR and open an issue?  We can look into user behaviour on mobile re: search

PR #20401, opened 2026-08-15
feat: social preview cards

miketheman:
Adopt OpenGraph standards and produce images and unfurl link metadata appropriate for users and projects.


JacobCoffee (approved):
looks good!

image
image

it'd be cool in a future PR to add more detail to project pages like
image

PR #20404, opened 2026-08-17
scanner: reject sparse tar members

woodruffw:
These have no business being in sdists anyways. 

Some notes on the approach here:

- I put this in the `scanner`, since we're already iterating over tar members there. Seemed like the simplest place to put it, but arguably it mixes concerns slightly since it happens before/forecloses on YARA processing.
- I've added a `TarPolicyError` exception to propagate a useful error message + slug for both user and metrics reporting. It may also be useful in the future if we choose to reject more tar mis-features.

For tests, I've added both unit and functional tests. There's some duplication in the fixtures we could probably remove, if people feel strongly about it.

### DIFF
--- a/docs/user/archives.md
+++ b/docs/user/archives.md
@@ -72,7 +72,7 @@ format, and are subject to the following restrictions.
 Source distributions may not contain GNU or pax-style sparse
 members. These are not allowed in archives on PyPI because
 they're not (directly) part of the pax standard,
-and are handling in various inconsistent ways by different
+and are handled in various inconsistent ways by different
 tar parsers.
 
 ## Further reading
--- a/docs/user/archives.md
+++ b/docs/user/archives.md
@@ -10,7 +10,11 @@ confuse archive implementations.
 This page details some of the the archive format features that PyPI rejects so if you
 encounter an error you can debug the issue and upload the fixed archive to PyPI.
 
-## Multiple or malformed central directory
+## Wheels
+
+Wheels use the ZIP format, and are subject to the following restrictions.
+
+### Multiple or malformed central directory
 
 Archives often support having multiple central directories
 or indexes to allow for "append-only" updates. This is not allowed in archives
@@ -19,7 +23,7 @@ Additionally, central directories must be specified correctly such that
 none of the central directory can be missed or misinterpreted
 (such as the offset within the archive, size, etc).
 
-## Missing file in central directory
+### Missing file in central directory
 
 This error occurs when a file is within the archive but the
 file is not recorded in the central directory. This may occur when
@@ -28,35 +32,48 @@ contents are not removed from the archive file itself. This can also
 occur if the central directory references a file whose data is not
 within the archive.
 
-## Duplicate file entries
+### Duplicate file entries
 
 There is more than one entry that shares the same name as another entry
 within the archive, either within the "central directory" of file entries
 or multiple entries. This is disallowed as some implementations process
 duplicates differently.
 
-## Filename is not valid
+### Filename is not valid
 
 The names of files within the archive must all be UTF-8 encoded bytes
 without unprintable characters.
 Unprintable characters as the Unicode codepoints `0x00-0x20` and `0x7F`.
 
-## Negative offset
+### Negative offset
 
 One of the relative offsets specified within the archive
-is negative instead of positive. 
+is negative instead of positive.
 
-## Duplicate extra metadata
+### Duplicate extra metadata
 
-There is two or more ZIP extra metadata field 
+There is two or more ZIP extra metadata field
 with the same ID that have security relevance, such
 as marking a ZIP as ZIP64 or defining the Unicode filename.
 
-## Trailing data or comments
+### Trailing data or comments
 
 Many archives support trailing or prepended data
 or comments within records. PyPI disallows these features
 to avoid smuggling other archive records within comments.
+
+## Source distributions
+
+Source distributions use the "tarball" (gzipped tar archive)
+format, and are subject to the following restrictions.
+
+### Sparse members are not allowed
+
+Source distributions may not contain GNU or pax-style sparse
+members. These are not allowed in archives on PyPI because
+they're not (directly) part of the pax standard,
+and are handling in various inconsistent ways by different
+tar parsers.
 
 ## Further reading
 
--- a/docs/user/archives.md
+++ b/docs/user/archives.md
@@ -12,7 +12,7 @@ encounter an error you can debug the issue and upload the fixed archive to PyPI.
 
 ## Wheels
 
-Wheels use the ZIP format, and are subject to the following restrictions.
+Wheels (also known as "built distributions") use the ZIP format, and are subject to the following restrictions.
 
 ### Multiple or malformed central directory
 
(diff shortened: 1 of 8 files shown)


sethmlarson (approved):
Thanks! Seems fine if we're okay with where we've done the detection and introducing a new error type. I'm less well-placed to answer those two questions.

warehouse/forklift/legacy.py
```diff
@@ -457,6 +457,12 @@ def _is_valid_dist_file(
                         )
                         return False, yara_match.message
 
+        except scanner.TarPolicyError as exc:
+            metrics.increment(
+                "warehouse.upload.tarfile.policy_error",
+                tags=[f"reason:{exc.reason}"],
+            )
+            return False, str(exc)
```
miketheman: curious - should this be caught inside the `if scan:` nesting? Logically, there can be no catching of a scanner error if there's no scanning. 
Or maybe this function needs a bit of refactoring? Since now the scanner is doing more than only yara scanning, and the inline comments are inconsistent.

warehouse/utils/scanner.py
```diff
@@ -36,6 +36,14 @@ class YaraMatch:
     message: str  # User-facing message from rule metadata
 
 
+class TarPolicyError(Exception):
+    """Raised when a tar archive violates an sdist policy."""
+
+    def __init__(self, message: str, *, reason: str) -> None:
+        super().__init__(message)
+        self.reason = reason
```
miketheman: Non-blocking thought: this is now the third or fourth exception-with-reason we're doing. It may make more sense to factor that out to an adoptable pattern?

warehouse/utils/scanner.py
```diff
@@ -88,6 +96,15 @@ def iter_zip_members(zfp: zipfile.ZipFile) -> typing.Iterator[tuple[str, int, by
...
     """Yield (name, size, data) for scannable files in a TarFile."""
     for member in tar.getmembers():
+        # Reject sparse members (whether GNU or pax-style) outright: these aren't
+        # compatible with PEP 625 (which stipulates pure pax) and have no business being
+        # in sdists anyways.
+        if member.issparse():
+            raise TarPolicyError(
+                "sdists may not contain sparse members",
+                reason="sparse-member",
+            )
```
miketheman: I get why this is here, but it also "feels wrong" - since nothing else in here appears to raise, but `continue`.

The condition you're testing for feels more likely to be placed in the "is valid dist file" check caller -  which might be a little more code churn that you wanted, but would be more structurally accurate.

miketheman (commented):
I had some questions inline about placement and if we preserve its location, then some documentation needs to be updated.

This probably works, but doesn't gel architecturally for me just yet.

warehouse/utils/scanner.py
woodruffw: Yeah, I can do that as a follow-up. We have too many of these.

warehouse/forklift/legacy.py
woodruffw: Yeah, I think I'd keep here here, if only for consistency with the existing exceptional flow. But it's a moot point since I'll try and move all of this somewhere else, if I can find a cleaner place for it.

warehouse/utils/scanner.py
woodruffw: Yeah, the reason I didn't stash it there is because we'd then effectively be iterating through the tar stream twice, which (unlike ZIP) can be very slow since each member needs to be fully streamed. I'll see if I can figure out a good workaround.

woodruffw: Done, I've lifted this into `_is_valid_dist_file`. Turns out `getmembers` caches internally so there's no real cost to re-iteration in our case 🙂

warehouse/forklift/legacy.py
woodruffw: With the lift this is now in the right place, IMO 🙂

docs/user/archives.md
```diff
@@ -10,7 +10,11 @@ confuse archive implementations.
 This page details some of the the archive format features that PyPI rejects so if you
 encounter an error you can debug the issue and upload the fixed archive to PyPI.
 
-## Multiple or malformed central directory
+## Wheels
```
miketheman, line 13:
```suggestion
## Binary Distributions (aka Wheels)
```
for consistency?

miketheman (approved):

docs/user/archives.md
woodruffw: I said "wheels" because the first paragraph says that, but I'll add a parenthetical inside the section!

docs/user/archives.md
```diff
@@ -28,36 +32,49 @@ contents are not removed from the archive file itself. This can also
...
+
+Source distributions use the "tarball" (gzipped tar archive)
+format, and are subject to the following restrictions.
+
+### Sparse members are not allowed
+
+Source distributions may not contain GNU or pax-style sparse
+members. These are not allowed in archives on PyPI because
+they're not (directly) part of the pax standard,
+and are handling in various inconsistent ways by different
```
JacobCoffee:
```suggestion
and are handled in various inconsistent ways by different
```

warehouse/forklift/legacy.py
```diff
@@ -431,6 +431,20 @@ def _is_valid_dist_file(
...
+                # in a pax header, and are not valid in PEP 625 sdists.
+                if any(member.issparse() for member in tar.getmembers()):
+                    metrics.increment(
+                        "warehouse.upload.tarfile.policy_error",
+                        tags=["reason:sparse-member"],
+                    )
+                    return (
+                        False,
+                        (
+                            "tar not accepted: Sparse members are not allowed. "
```
JacobCoffee:
```suggestion
                            "tar archive not accepted: Sparse members are not allowed. "
```
we have similar wording for the zip archives in [line 408 above](https://github.com/pypi/warehouse/pull/20404/changes#diff-bffd49e4f943217b54b4cfcf5835737bec7a145360bd1dc2ec1541d2dcafad16R408), costmetic though
will need to update tests too

JacobCoffee (approved):

tests/unit/forklift/test_legacy.py
```diff
@@ -483,6 +483,55 @@ def test_yara_match_in_tarball(self, tmpdir, monkeypatch):
...
+
+        assert legacy._is_valid_dist_file(
+            tar_fn,
+            "sdist",
+            metrics,
+            scan=scan,
+        ) == (
+            False,
+            (
+                "tar not accepted: Sparse members are not allowed. "
```
woodruffw:
```suggestion
                "tar archive not accepted: Sparse members are not allowed. "
```

tests/functional/forklift/test_legacy.py
```diff
@@ -140,6 +165,59 @@ def test_file_upload(webtest, upload_url, additional_data):
...
+            "filetype": "sdist",
+            "metadata_version": "2.1",
+            "version": "1.0",
+        },
+        upload_files=[("content", "sampleproject-1.0.tar.gz", content)],
+        status=HTTPStatus.BAD_REQUEST,
+    )
+
+    assert response.status == (
+        "400 Invalid distribution file. tar not accepted: Sparse members are not "
```
woodruffw:
```suggestion
        "400 Invalid distribution file. tar archive not accepted: Sparse members are not "
```

warehouse/forklift/legacy.py
woodruffw: My brain reads this like "ATM machine," but I've lost that battle 🙂

michaelrodgers355-lab (approved):

PR #20405, opened 2026-08-17
Fix UA render

### LINKED ISSUE #19076: File details: "uploaded via" renders suboptimally with JSON UAs
Some uploading clients (notably `uv publish`, but potentially twine as well in the future) use JSON in their User-Agent to help feed linehaul more structured metadata. Unfortunately, this JSON payload doesn't render super nicely on the file details page 🙂 

I initially tried to fix this by wrapping `file.uploaded_via` in a `<code>` block, but the wrapping/overflowing is still a bit off with that. I'll try and fix it more thoroughly in the coming days, but I figured I should file a tracking issue.

---

Example:

Image

From: https://pypi.org/project/uv/#uv-0.9.10-py3-none-win_arm64.whl

woodruffw:
Fixes #19076.

Example:

image

### DIFF
--- a/warehouse/locale/messages.pot
+++ b/warehouse/locale/messages.pot
@@ -1069,7 +1069,7 @@ msgstr ""
 #: warehouse/templates/base.html:464 warehouse/templates/base.html:477
 #: warehouse/templates/base.html:494 warehouse/templates/base.html:502
 #: warehouse/templates/includes/accounts/profile-callout.html:5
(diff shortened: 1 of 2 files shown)


miketheman (approved):

JacobCoffee (approved):
@woodruffw is this pattern reliably detectable? may be nice in a future PR to render it in the UI prettier/more easily readable

PR #20412, opened 2026-08-18
fix: rate limit email changes

miketheman:
In #19662 I added the ability to change an email post-username registration, but that doesn't have any rate limits on failed attempts, allowing a user to brute-force different providers until one works.

Adding a rate limiter to count attempts as well as successes, scoped to a **user account** means that they already have begun a registration process, and are now trying to verify an email address.

### DIFF
--- a/warehouse/config.py
+++ b/warehouse/config.py
@@ -553,6 +553,12 @@ def configure(settings=None):
     )
     maybe_set(
         settings,
+        "warehouse.account.email_change_ratelimit_string",
+        "EMAIL_CHANGE_RATELIMIT_STRING",
+        default="5 per 5 minutes, 20 per hour",
+    )
+    maybe_set(
+        settings,
         "warehouse.account.accounts_search_ratelimit_string",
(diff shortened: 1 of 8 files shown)


warehouse/config.py
```diff
@@ -551,6 +551,12 @@ def configure(settings=None):
         "VERIFY_EMAIL_RATELIMIT_STRING",
         default="3 per 6 hours",
     )
+    maybe_set(
+        settings,
+        "warehouse.account.email_change_ratelimit_string",
+        "EMAIL_CHANGE_RATELIMIT_STRING",
+        default="5 per 5 minutes, 20 per hour",
```
ewdurbin, line 558: this seems.... pretty generous?

ewdurbin (approved):

warehouse/config.py
miketheman: I'm considering lowering it via env var, but I saw a spurt of 20 in 1 second, so I figured start here.

PR #20417, opened 2026-08-19
move back to stdlib for all json

ewdurbin:
we'll want to carefully monitor perf when this rolls out and be somewhat prepared for a rollback if we can't support the traffic without an optimized json serializer.


ww-oai (approved):

woodruffw (approved):

tests/unit/cache/test_services.py
```diff
@@ -35,27 +32,13 @@ def test_set_get_simple(self, query_results_cache_service):
 
         assert result == {"foo": "bar"}
 
-    def test_set_get_complex(self, query_results_cache_service):
-        # Construct a complex object to store in the cache
+    def test_set_get_nested(self, query_results_cache_service):
```
ewdurbin, line 35: note: the only current use of RedisQueryResults (compute_top_dependents_corpus / typo_check_project_name) has dict[str, int] shape, so i just minimized the test here rather than implementing UUID/datetime defaults

PR #20408, opened 2026-08-18
Add security tab

nlhkabu:
TODO:
 
- [x] Update security policy re: moved malware button
- [x] Add provenance section to metadata sidebar on project details page
- [x] Implement warning callout blocks
- [x] Implement filtering on the file list
- [x] Handle long project names
- [x] Handle projects with lots of files (largest file count = 307)
- [x] Mobile / responsive !
- [x] Check horizontal tabs overflow on longer languages (now that we have added an extra tab it may not wrap on mobile)
- [x] BLOCKED: Expose provenance status on release history page
- [x] Rebase to updated files PR
- [x] Show provenance state in files table
- [x] Don't show provenance top banner on yanked releases/projects
- [x] Add provenance warnings to files pages
- [ ] Lint, translate

PR #20420, opened 2026-08-19
Add `Project.provenance_statuses`

miketheman:
Adds `Project.provenance_statuses`, a project-wide mapping of version to
`ProvenanceStatus`, so a page showing one line per release can get every
release's provenance verdict in a single query.

Refs #20329
Closes #20410

## Why

Reading `Release.provenance_status` in a loop costs three queries per release.
[This review comment on #20329](https://github.com/pypi/warehouse/pull/20329/changes#r3778988535)
flagged it on the release history page, which currently assembles the mapping
in the template:

```jinja
{% set provenance_statuses = {} %}
{% for r in project.releases %}
  {% set _ = provenance_statuses.update({r.version: r.provenance_status}) %}
{% endfor %}
```

Jinja scopes `{% set %}` to the loop body, so accumulating across iterations
means mutating an object made outside it. That `{% set _ = %}` idiom appears
once elsewhere in our templates, and never to build a lookup table. The N+1 is
a symptom of assembling data in template-land, where the only way to express
"a status per release" is a loop over full ORM objects.

## What it does

`_provenance_query` fetches a release's files and their provenance together.
`_counted_releases` folds the rows. Because the result already holds every
release's date and coverage, each comparison is a walk over that list instead
of two more queries. `Release.provenance_counts` reuses the same query and
drops from two queries to one.

Both paths now break `created` ties on `_pypi_ordering` and `id`.
`Release.created` defaults to a transaction-scoped `now()`, so a batch insert
leaves every release of it sharing one, and without a tiebreak the batched and
per-release paths could name different comparison releases for the same
release and render contradictory verdicts.

## Measured

250 releases x 4 attested files, one dataset, best of 3:

| approach | queries | wall | states |
|---|---|---|---|
| `Release.provenance_status` in a loop (#20329 today) | 501 | 330 ms | all six |
| `Project.provenance_counts` (#20410) | 1 | 42 ms | four of six |
| `Project.provenance_statuses` (this PR) | 1 | 47 ms | all six |

#20410 solves the same N+1 and reads the same payloads, but has no comparison
between releases, so it cannot produce `lost-provenance` or
`changed-provenance` — two of the three warning states the cascade in #20408
puts at the top. It also names the property `Project.provenance_counts`, beside
`Release.provenance_counts`, which returns a different type.

## Against the security tab

#20408 renders one template per `ProvenanceState` through a priority cascade:

```
lost > changed > inconsistent > partial > full > no
```

`test_provenance_statuses_drive_the_security_cascade` builds one release per
state and asserts each is reachable from the mapping, then asserts every entry
equals what that release's own page renders. The release history list and the
detail page cannot disagree.

## Known cost, and the follow-up

This reads every provenance payload in the project on an uncached render,
because `inconsistent-provenance` is derived from each release's own publisher
set. Restricting the read to the comparison window would leave that state
silently absent outside it, and a jsonpath projection cannot distinguish "all
the publishers" from "the publishers I could see" — both were tried and
rejected on correctness.

The fix that removes the cost without losing states is storing publisher kind,
identity and workflow alongside the provenance row at upload. Then every state
is a cheap aggregate and this property, the security tab and the metadata
sidebar all get faster. Worth doing separately.

Coverage-only alternatives are cheaper still (one query, ~3 ms, no payload)
but reach only four of the six states, so the history badge would show a green
check on a release its own page warns about.

Pull Request description 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -6,7 +6,7 @@ import datetime
 import enum
 import typing
 
-from collections import Counter, OrderedDict
+from collections import OrderedDict
 from functools import cached_property
 from uuid import UUID
 
@@ -60,8 +60,7 @@ from warehouse.attestations.models import (
     ProvenanceComparison,
     ProvenanceCounts,
     ProvenanceStatus,
-    PublisherSource,
-    get_provenance_sources,
+    counts_from_provenance,
 )
 from warehouse.authnz import Permissions
 from warehouse.classifiers.models import Classifier
@@ -1003,25 +1002,7 @@ class Release(HasObservations, db.Model):
             .options(orm.undefer(Provenance.provenance))
             .all()
         )
-        source_counter: Counter[PublisherSource] = Counter()
-        workflow_counter: Counter[str] = Counter()
-        unreadable_files = 0
-        for provenance_object in provenance_objects:
-            extracted = get_provenance_sources(provenance_object)
-            if extracted is None:
-                unreadable_files += 1
-                continue
-            sources, workflows = extracted
-            source_counter.update(sources)
-            workflow_counter.update(workflows)
-
-        return ProvenanceCounts(
-            total_files=total_files,
-            files_with_provenance=len(provenance_objects),
-            unreadable_files=unreadable_files,
-            source_counts=dict(source_counter),
-            workflow_counts=dict(workflow_counter),
-        )
+        return counts_from_provenance(total_files, provenance_objects)
 
     @cached_property
     def provenance_status(self) -> ProvenanceStatus | None:
--- a/warehouse/packaging/models.py
+++ b/warehouse/packaging/models.py
@@ -50,6 +50,7 @@ from sqlalchemy.orm import (
     mapped_column,
     validates,
 )
+from sqlalchemy.sql import Select
 from urllib3.exceptions import LocationParseError
 from urllib3.util import parse_url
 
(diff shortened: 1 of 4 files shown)


warehouse/packaging/models.py
```diff
@@ -94,6 +94,84 @@
...
+    `_pypi_ordering` breaks those ties the way the rest of this class already
+    orders releases, and `id` breaks the remainder, since `_pypi_ordering` is
+    nullable and cannot be relied on for a total order by itself.
+    """
+    return (
+        select(Release.version, Release.created, File.id, Provenance)
+        .select_from(Release)
+        .outerjoin(File, File.release_id == Release.id)
+        .outerjoin(Provenance, Provenance.file_id == File.id)
+        .options(orm.undefer(Provenance.provenance))
```
JacobCoffee, line 116: this is saying go ahead and pull the provenance since we defer it on the original model

JacobCoffee (approved):

warehouse/packaging/models.py
JacobCoffee: (and we defer because the jsonb might be ginormous?)

di (approved):

warehouse/packaging/models.py
miketheman: That's correct!

miketheman: Here's one way to look at the table and rosw sizes in relation to each other:

| item | bytes | size | avg\_per\_row | pct\_of\_data | pct\_of\_total |
| :--- | :--- | :--- | :--- | :--- | :--- |
| provenance | 19345410765 | 18 GB | 8603.1 | 99.63 | 91.78 |
| id | 35978480 | 34 MB | 16 | 0.19 | 0.17 |
| file\_id | 35978480 | 34 MB | 16 | 0.19 | 0.17 |
| indexes | 170123264 | 162 MB | null | null | 0.81 |
| row overhead + bloat | 1490975571 | 1422 MB | null | null | 7.07 |
| TOTAL | 21078466560 | 20 GB | null | null | 100 |

So any queries that only need the id, file_id, or the logical relationship of `file` can return very quickly without having to dig up the larger JSON objects from the `provenance` column.

Query:

<details>

```sql
WITH tot AS (
    SELECT pg_total_relation_size(c.oid) AS total_bytes,
           pg_indexes_size(c.oid)        AS index_bytes
    FROM pg_class c
    WHERE c.oid = 'provenance'::regclass
),
raw AS (
    SELECT count(*)                        AS n_rows,
           sum(pg_column_size(id))         AS "id",
           sum(pg_column_size(file_id))    AS "file_id",
           sum(pg_column_size(provenance)) AS "provenance"
    FROM provenance
),
cols AS (
    SELECT r.n_rows,
           e.key                        AS item,
           (e.value #>> '{}')::bigint   AS bytes
    FROM raw r, jsonb_each(to_jsonb(r) - 'n_rows') e
),
rollup AS (
    SELECT 1 AS ord, c.item, c.bytes,
           round(c.bytes::numeric / nullif(c.n_rows, 0), 1)             AS avg_per_row,
           round(100.0 * c.bytes / nullif(sum(c.bytes) OVER (), 0), 2)  AS pct_of_data,
           round(100.0 * c.bytes / nullif(t.total_bytes, 0), 2)         AS pct_of_total
    FROM cols c CROSS JOIN tot t
    UNION ALL
    SELECT 2, 'indexes', t.index_bytes, NULL, NULL,
           round(100.0 * t.index_bytes / nullif(t.total_bytes, 0), 2)
    FROM tot t
    UNION ALL
    SELECT 3, 'row overhead + bloat',
           t.total_bytes - t.index_bytes - (SELECT sum(bytes) FROM cols),
           NULL, NULL,
           round(100.0 * (t.total_bytes - t.index_bytes
                          - (SELECT sum(bytes) FROM cols))
                 / nullif(t.total_bytes, 0), 2)
    FROM tot t
    UNION ALL
    SELECT 4, 'TOTAL', t.total_bytes, NULL, NULL, 100.00 FROM tot t
)
SELECT item, bytes, pg_size_pretty(bytes) AS size,
       avg_per_row, pct_of_data, pct_of_total
FROM rollup
ORDER BY ord, bytes DESC NULLS LAST;
```

</details>

PR #20424, opened 2026-08-20
Orgs(promo): Show survey callout on company org settings

JacobCoffee:
Add callout for orgs on individual org detail settings pages

image

---

### DIFF
--- a/warehouse/templates/manage/organization/settings.html
+++ b/warehouse/templates/manage/organization/settings.html
@@ -7,6 +7,11 @@
   {% trans organization_name=organization.name %}Manage '{{ organization_name }}' settings{% endtrans %}
 {% endblock %}
 {% block main %}
+  {% if organization.orgtype == OrganizationType.Company %}
+    {% with company_organizations=[organization] %}
+      {% include "warehouse:templates/includes/manage/company-org-survey-callout.html" %}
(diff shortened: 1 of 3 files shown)


warehouse/templates/manage/organization/settings.html
```diff
@@ -7,6 +7,11 @@
   {% trans organization_name=organization.name %}Manage '{{ organization_name }}' settings{% endtrans %}
 {% endblock %}
 {% block main %}
+  {% if organization.orgtype == OrganizationType.Company %}
+    {% with company_organizations=[organization] %}
+      {% include "warehouse:templates/includes/manage/company-org-survey-callout.html" %}
```
ewdurbin, line 12: will the dismissable status be shared between the two places this is shown?

ewdurbin (approved):

warehouse/templates/manage/organization/settings.html
JacobCoffee: yes! they include the same partial and the ID for that cookie is set as a global,

also updated just now to incl. lifetimes https://github.com/pypi/warehouse/pull/20423#pullrequestreview-5102522724 so they arent annoying and pop up every week (set to 180 days)

PR #20431, opened 2026-08-20
Add npm dependency cooldown

woodruffw:
Adds an npm cooldown, matching the Dependabot one. The idea is to guard against local updates as well, if/when those happen during dev flows.


JacobCoffee (approved):
yay gh stacks

PR #20432, opened 2026-08-20
Pin "one-shot" tools

woodruffw:
Invocations like `npx` and `pipx` should always install a specific version to prevent the latest version from being implicitly resolve. This addresses the only two instances of that I could find 🙂 

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,4 +1,4 @@
-# https://github.com/woodruffw/zizmor
+# https://github.com/zizmorcore/zizmor-action
 name: GitHub Actions Security Analysis with Zizmor
 
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/zizmor.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       - name: Run zizmor
-        run: pipx run zizmor --format sarif . > results.sarif
+        run: pipx run zizmor==1.29.0 --format sarif . > results.sarif
```
woodruffw: NB: We could consider moving to `zizmor-action` here, I don't fully remember why we don't use it.

miketheman: Probably because we predate the GH action. I'd prefer that refactor.

woodruffw: SG, I'll tweak.

.github/workflows/zizmor.yml
```diff
@@ -23,15 +23,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - name: Setup Python
-        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       - name: Run zizmor
-        run: pipx run zizmor --format sarif . > results.sarif
-      - name: Upload SARIF file
-        uses: github/codeql-action/upload-sarif@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
```
woodruffw, line 27: NB: Every version of zizmor-action is tied to a known hash-pinned set of zizmor versions, so this implicitly hash-pins zizmor as well.

miketheman (approved):

JacobCoffee (approved):

PR #20436, opened 2026-08-21
Refactor locale file handling and error reporting

mkcs28:
Updated file handling to use existsSync for missing locale files and improved error reporting for missing headers.


mkcs28 (commented):
Very Much Required

mkcs28 (commented):
.

PR #20464, opened 2026-08-29
Add `loading=lazy` to images on user-facing pages

di:
This should cut down on requests to Camo and also make page loading faster.

MDN: https://developer.mozilla.org/en-US/docs/Web/Performance/Guides/Lazy_loading
Browser support: https://caniuse.com/loading-lazy-attr

### DIFF
--- a/tests/unit/test_filters.py
+++ b/tests/unit/test_filters.py
@@ -61,7 +61,7 @@ class TestCamoify:
         assert result == (
             '<img src="https://camo.example.net/'
             "b410d235a3d2fc44b50ccab827e531dece213062/"
-            '687474703a2f2f6578616d706c652e636f6d2f696d6167652e6a7067">'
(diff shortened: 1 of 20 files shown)


JacobCoffee (approved):

nlhkabu (changes requested):
Should we be adding width and height here, where possible?

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#lazy

> While explicit [width](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#width) and [height](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#height) attributes are recommended for all images to avoid layout shift, they are especially important for lazy-loaded ones. Lazy-loaded images will never be loaded if they do not intersect a visible part of an element, even if loading them would change that, because unloaded images have a width and height of 0. It creates an even more disruptive user experience when the content visible in the viewport reflows in the middle of reading it.

nlhkabu (approved):

PR #20468, opened 2026-08-29
Add fallback icon for `source` URLs in "Project links"

Steffo99:
This PR adds an icon for `source` [well-known labels](https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels) that do not match any of the predefined code forge domains, such as `github.com`, `gitlab.com`, `codeberg.com`, `google.com` or `bitbucket.org`.

Example screenshot.

This would be useful for self-hosted code forges, like self-hosted GitLab, Gitea, and Forgejo instances, or cgit servers.


docs/user/project_metadata.md
```diff
@@ -96,15 +96,16 @@ recognition patterns are case-insensitive. Items marked with an asterisk (^*^)
...
+| Release notes    | :fontawesome-solid-newspaper:             | Curated release information | Releasenotes                                                                                           |
+| Documentation^*^ | :fontawesome-solid-book:                  | Project documentation       | Docs^*^ , a URL pointing to [Read the Docs] domains or a URL starting with `docs.` or `documentation.` |
+| Bug^*^           | :fontawesome-solid-bug:                   | Bug/Issue report location   | Issue^*^, Tracker^*^, Report^*^                                                                        |
+| Funding^*^       | :fontawesome-solid-circle-dollar-to-slot: | Sponsoring information      | Sponsor^*^, Donation^*^, Donate^*^                                                                     |
+| Source^*^        | :fontawesome-solid-code-branch:           | Source code repository      | Source^*^, Repository^*^                                                                               |
```
woodruffw: For consistency with the well-known labels, IMO we shouldn't use a prefix here: for `Source` we should treat `Repository` and `Source Code` as aliases, but not prefix match.

warehouse/templates/includes/packaging/metadata/project-links.html
```diff
@@ -49,6 +49,8 @@
     {% set icon = "fab fa-youtube" %}
   {% elif name|lower == "mastodon" %}
     {% set icon = "fab fa-mastodon" %}
+  {% elif name.lower().startswith(("source", "repository")) %}
```
woodruffw: Per above, I think this could be:


```suggestion
  {% elif name|lower in ["source", "source code", "repository"] %}
```

(Longer term we should perform the proper label normalization here, but that's definitely out of scope 🙂)

woodruffw (approved):
LGTM. I kind of suspect the `sourcecode` spelling is marginal but it's fine to keep IMO since that'll be the normalized label anyways in the future.

PR #20469, opened 2026-08-29
Add specialized icon for `releasenotes` URLs in "Project links"

Steffo99:
This PR adds an icon for URLs with `releasenotes` [well-known label](https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels) in the "Project Links" section.

Example screenshot.

The icon is different from the one used for `changelog`, as the two are considered two separate labels:

- `changelog` is for "The project’s comprehensive changelog"
- `releasenotes` is for "The project’s curated release notes"

Additionally, this fixes handling of the `whatsnew` well-known label, which isn't parsed in its lowercased version, and instead is handled only if it matches `"What's New"`.


miketheman (approved):
Thanks!

miketheman (approved):

PR #20473, opened 2026-08-31
feat: Add the UserCheck email reputation backend

miketheman:
UserCheckEmailReputationService asks usercheck.com's `/email/` endpoint about an address. The full address is checked, because an address can be disposable on a domain that is not: throwaway aliases exist on public providers like gmail.com, and only an address-level check sees those.

The check fails open on service trouble (timeouts, 5xx, 429, malformed bodies, missing API key) and fails closed when the caller has spent its per-IP budget.

Refs: https://www.usercheck.com/docs/api/email-endpoint


warehouse/accounts/services.py
```diff
@@ -1281,3 +1283,162 @@ def create_service(cls, _context, _request) -> NullEmailReputationService:
...
+                timeout=(0.25, 1),
+            )
+            resp.raise_for_status()
+        except requests.RequestException as exc:
+            # Fail open: an unavailable, unauthorized or rate-limited service
+            # must never stop somebody from using their email address. The
+            # exception repr embeds the request URL, and with it the address,
+            # so log only the exception type.
+            status_code = (
+                exc.response.status_code if exc.response is not None else "none"
```
ewdurbin, line 1384: since `requests.RequestException` will catch connect/read timeout errors as well, and we have a 250ms connect and 1s read timeout, it will be hard to differentiate between them (and other more obscure errors) in the metrics. Should we consider capturing connect and read timeouts separately so that we can set different tags?

miketheman: Thanks, added the exception name in the metrics tags

ewdurbin (approved):

PR #20477, opened 2026-09-01
feat(admin): decode as much of a token as possible

miketheman:
Tokens may show up truncated, quoted in a leak report or copied out of a log, and the decode admin page will reject them with a 400. A partial macaroon may contain enough to decide the identifier, which is enough to pull up the row.

`deserialize_partial_macaroon` reads the v2 packets itself. `pymacaroons` raises before telling you how long the payload it couldn't read was meant to be, and that length is the useful part. A fragment ending in [3," has a 42-byte caveat behind it, so it is an untouched RequestUser.

Strict deserialization is untouched. The view only reaches for the partial read once it raises, and a token with nothing readable still gets its 400. An identifier that was itself cut short is shown as-is, never looked up.

### DIFF
--- a/tests/functional/admin/test_macaroons.py
+++ b/tests/functional/admin/test_macaroons.py
@@ -2,27 +2,55 @@
 
 from http import HTTPStatus
 
-# The head of a truncated token, held apart from its `pypi-` prefix so secret
-# scanners have no token pattern to match. It carries no signature.
-_TOKEN_BODY = (
-    "AgEIcHlwaS5vcmcCJDAyZWRlM2ZlLWRjMDAtNDViOS05YTEzLTlmMTZhZDFjNDU0ZAACKlszLCJ"
-)
-TRUNCATED_TOKEN = "pypi-" + _TOKEN_BODY
+import pytest
+
+from tests.common.db import Session
+from tests.common.db.accounts import UserFactory
+from warehouse.macaroons import caveats
+from warehouse.macaroons.services import DatabaseMacaroonService
 
 
-class TestDecodeTruncatedToken:
-    def test_admin_sees_the_fields_that_could_be_read(self, webtest, login_admin):
-        """A token with no signature left still renders the fields it has."""
+@pytest.fixture
+def token():
+    """A real token and the macaroon behind it, as account settings mints them."""
+    owner = UserFactory.create()
+    return DatabaseMacaroonService(Session).create_macaroon(
+        location="pypi.org",
+        description="a token to look up",
+        scopes=[caveats.RequestUser(user_id=str(owner.id))],
+        user_id=owner.id,
+    )
+
+
+def _decode(webtest, raw_token):
+    page = webtest.get("/admin/token/decode", status=HTTPStatus.OK)
+    form = page.forms["decode-token"]
+    form["token"] = raw_token
+    return form.submit(status=HTTPStatus.OK)
+
+
+class TestDecodeToken:
+    def test_admin_sees_a_whole_token(self, webtest, login_admin, token):
+        """A whole token links to the macaroon it names."""
         login_admin()
+        raw_token, macaroon = token
 
-        page = webtest.get("/admin/token/decode", status=HTTPStatus.OK)
-        form = page.forms["decode-token"]
-        form["token"] = TRUNCATED_TOKEN
+        result = _decode(webtest, raw_token)
 
-        result = form.submit(status=HTTPStatus.OK)
+        assert f"/admin/macaroons/{macaroon.id}" in result.text
+        assert "truncated or otherwise malformed" not in result.text
+
+    def test_admin_sees_the_fields_a_truncated_token_holds(
+        self, webtest, login_admin, token
+    ):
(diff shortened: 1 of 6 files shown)


tests/functional/admin/test_macaroons.py
```diff
@@ -0,0 +1,28 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from http import HTTPStatus
+
+# The head of a truncated token, held apart from its `pypi-` prefix so secret
+# scanners have no token pattern to match. It carries no signature.
+_TOKEN_BODY = (
+    "AgEIcHlwaS5vcmcCJDAyZWRlM2ZlLWRjMDAtNDViOS05YTEzLTlmMTZhZDFjNDU0ZAACKlszLCJ"
+)
+TRUNCATED_TOKEN = "pypi-" + _TOKEN_BODY
```
woodruffw: Secret scanners hate this one weird trick

warehouse/macaroons/services.py
woodruffw, line 1: Purely a nitpick, but the partial decoding _might_ belong somewhere in `caveats` instead? I'm not sure.

warehouse/macaroons/services.py
```diff
@@ -55,6 +59,170 @@ def deserialize_raw_macaroon(raw_macaroon: str | None) -> pymacaroons.Macaroon:
...
+        elif packet.field_type == _PACKET_SIGNATURE:
+            partial.signature = _packet_text(packet)
+        elif packet.field_type == _PACKET_LOCATION and in_header:
+            partial.location = _packet_text(packet)
+        elif packet.field_type == _PACKET_IDENTIFIER:
+            if in_header:
+                partial.identifier = _packet_text(packet)
+                partial.identifier_complete = packet.complete
+            else:
+                partial.caveats.append(_packet_text(packet))
```
woodruffw, line 237: If I'm reading this right, there are two ambiguities here:

- If someone contrives a partial macaroon two have two identifier packets, the last one wins here. In a real macaroon there would only be one, so I suspect we should reject a partial macaroon that claims to have multiple identifiers.
- We start out by assuming that we're in the header, but if we never see an EOS then we can end up parsing a caveat as an identifier.

warehouse/macaroons/services.py
```diff
@@ -55,6 +59,170 @@ def deserialize_raw_macaroon(raw_macaroon: str | None) -> pymacaroons.Macaroon:
...
+
+def _packet_text(packet: _Packet) -> str:
+    """Render a packet's payload for display, with a note if it was cut short."""
+    if packet.field_type == _PACKET_SIGNATURE:
+        text = packet.payload.hex()
+    else:
+        try:
+            text = packet.payload.decode()
+        except UnicodeDecodeError:
+            text = packet.payload.hex()
```
woodruffw, line 189: I _think_ this fallback adds some malleability: if someone packs a UUID as 16 (non-UTF-8) bytes here we'll hex-encode it, which then _accidentally_ works since `uuid.UUID` allows a plain hexstring form. So this is more permissive than our normal identifier validation, which rejects non-UTF-8 outright.

I _think_ the solution here is to just reject instead of trying to recover/preserve a hexstring here.

woodruffw: (The fact that `uuid.UUID` allows a hexstring form is separately a bit nuts. But it doesn't directly affect is in the normal path since we have signatures binding over the original input there.)

woodruffw (changes requested):
Structurally this looks pretty good to me, my feedback is mostly about some probably unintentional malleability/deviation from how the Macaroon format itself would parse packets.

On a more "should we do this level," I find this kind of spooky 😅 -- fundamentally this is operating on untrusted data (and presenting it in a privileged context, i.e. the admin panel), so I have a high-level concern that it'll be easy to mis-gauge how much to trust the data we pull out here. OTOH this all only seems to happen on demand in the token decoding admin helper, so maybe it's fine?

warehouse/macaroons/services.py
miketheman: I think it's probably better here, since caveats package is purely about the caveat serde for that portion of the macaroon, and this change applies to an incomplete macaroon, that may or may not have caveats.

warehouse/macaroons/services.py
miketheman: Some tightening applied in newer commits

tests/functional/admin/test_macaroons.py
miketheman: Factories to the rescue!

warehouse/macaroons/services.py
woodruffw: As-is we could still end up misinterpreting a caveat as an identifier here, but the more I think about it the more I'm convinced that's not a security problem: we're fully unsigned in this context anyways, so someone who _wants_ to stuff an identifier here can put it wherever they'd like anyways.

woodruffw (approved):
LGTM structurally. We should continue to be careful to present anything extracted from a partial macaroon as untrustworthy, as this PR does 🙂

PR #20495, opened 2026-09-08
feat(orgs): allow pre-billing Billing Manager invites

### LINKED ISSUE #20489: Allow Orgs to invite billing managers
## What's the problem this feature will solve?

A newly approved Company organization cannot invite any members until billing is active. `Organization.is_in_good_standing()` returns `False` for Company orgs with no active subscription or manual activation, and the invite form on the Members page is hidden behind that check (`warehouse/templates/manage/organization/roles.html`), with a matching server-side guard in `_send_organization_invitation` ("Cannot invite new member. Organization is not in good standing.").

The person who requested the org may not be person who holds the company card.

So the owner has to either (a) put a personal card in and get reimbursed, (b) forward the Stripe Checkout link and have IT fill it in under the owner's session, or raise a support request which uincreases the queue workload

## Describe the solution you'd like

Allow Company org owners to invite users with the ~~**Owner** or~~ (let's only allow billing mgr invites!) **Billing Manager** role before the organization has an active subscription, so the person who will pay can be added first.

Suggested shape:

- In `_send_organization_invitation`, relax the good-standing check for Company orgs with no subscription yet: allow invites for `Billing Manager`, keep blocking `Owner`, `Manager` / `Member`.
- Show the invite form on `roles.html` in that state, with a callout explaining that only billing-capable roles can be invited until billing is activated. Invitations still count toward seats once a subscription exists.

Do **not** change the behavior for orgs whose subscription has *lapsed* (should remian still fully blocked), since that is the case the good-standing gate was written for.

## Additional context

- Gate: `warehouse/organizations/models.py::Organization.is_in_good_standing`
- Guard: `warehouse/manage/views/organizations.py::_send_organization_invitation`
- Template: `warehouse/templates/manage/organization/roles.html` (`{% if organization.is_in_good_standing() %}` around the invite form)
- Related: #20232 relaxed a similar gate for deleting orgs with inactive billing.

Drafted with 🤖

chrisriv10:
## Summary

Fixes #20489

Newly approved Company organizations currently cannot invite anyone until billing is active. This can prevent the organization owner from adding the person responsible for entering the company’s billing information.

This allows Company organizations that have never activated billing to invite a Billing Manager before billing activation while continuing to block Owner, Manager, and Member invitations.

Organizations with previous billing or manual activation history remain blocked while billing is inactive.

## Changes

- Allow the People page to be accessed for eligible pre-billing Company organizations.
- Restrict invitations in that state to the `Billing Manager` role.
- Enforce the restriction server-side in addition to limiting the available form choices.
- Use organization event history to distinguish never-activated organizations from organizations with prior billing or manual activation.
- Add a callout explaining the Billing Manager-only restriction before billing is activated.
- Preserve existing behavior for active subscriptions, lapsed subscriptions, Community organizations, manual activations, and deactivated organizations.
- Add unit and functional regression coverage.
- Regenerate the translation catalog for the new UI strings.

## Notes for reviewers

The pre-billing exception is intentionally narrow. It only applies to active Company organizations that are not currently in good standing and have no current or historical subscription or manual-activation state.

The event-history check is used because subscription and manual-activation records can be removed while organization events remain available. This prevents a previously activated organization from later being treated as newly approved.

Eligibility checks are ordered from cheapest to most expensive: scalar organization state, current billing relationships, and finally durable event history.

This PR does not change resend or revoke permissions for pending invitations.

## Testing

- `tests/unit/manage/views/test_organizations.py`: 88 passed
- `tests/unit/organizations/test_models.py`: 55 passed
- `tests/functional/manage/test_organization_roles.py`: 9 passed
  - Includes `test_get_organization_roles` with its exact 13 query assertion.
- `TestCreateOrganizationRoleForm`: 1 passed
- Pinned Ruff `0.15.20` check passes.
- Pinned Ruff `0.15.20` format check passes.
- Translation regeneration passes.
- Targeted mypy checks pass.
- `git diff --check` passes.


JacobCoffee (approved):
this has a bug where everyone shows a billing manager and some other things - approving but ill push changes to get it mergeable

thanks @chrisriv10 !

PR #20501, opened 2026-09-09
Add shield icon for security project links

BHUVANSH855:
Adds a shield icon for the `security` well-known project URL label, including its documented aliases.

From https://github.com/pypa/packaging.python.org/pull/2127#issuecomment-5587675822


miketheman (approved):
Thanks!

PR #20506, opened 2026-09-10
feat(accounts): reject disposable email addresses during validation

miketheman:
`NewEmailMixin` runs a reputation check after every other field passes, so a submission already doomed by its captcha or password doesn't spend the metered budget. A disposable verdict rejects the address, and when the domain itself is a disposable provider, `prohibit_email_domain()` also records it in `prohibited_email_domains` (gated behind the default-off `auto-prohibit-disposable-domains` flag) so the next attempt is caught without another remote call.

The check runs against the address's ASCII form, since the vendor doesn't score raw Unicode from an IDN domain correctly. The ratelimiter keys on user id when the caller is authenticated, falling back to remote address for registration, so one account hammering `add_email` can't lock out everyone behind the same office NAT. `prohibited_by` is now nullable, since the automated rows have no admin behind them.


warehouse/accounts/forms.py
```diff
@@ -430,6 +443,107 @@ def validate_email(self, field):
...
+            return False
+
+        if reputation is None or not reputation.should_block:
+            self.request.metrics.increment(
+                "warehouse.accounts.forms.validate_email_reputation",
+                tags=["result:valid"],
+            )
+            return True
+
+        if reputation.disposable_domain:
```
woodruffw, line 487: Noting: based on my read of `disposable_domain` here:

https://github.com/pypi/warehouse/blob/81db339197ea07e169fdddeddfb7d248ad4582f1/warehouse/accounts/interfaces.py#L376-L390

`self.disposable` refers to whether the _address_ is disposable (not the email provider itself).

As a result, this can snare users who have a non-public, non-relay email provider  that (inadvertently) tenants a single spammer by address. 

A specific griefing scenario:

1. I obtain an address on `mit.edu` (or whatever, any non-public, non-relay provider)
2. I get that addressed marked as disposable (by being a spammer)
3. I attempt to use that address on PyPI
4. PyPI bans all of `mit.edu` despite only my address being disposable

I _suspect_ what we want here is to check whether the entire _domain_ is disposable, in which case we need UserCheck's `/domain` endpoint rather than `/email`.

Refs:

- https://www.usercheck.com/docs/api/email-endpoint
- https://www.usercheck.com/docs/api/domain-endpoint

(That conclusion is also based on the comment below, which describes the intent based on the "whole domain operat[ing] as a disposable provider.")

warehouse/accounts/forms.py
```diff
@@ -345,13 +351,20 @@ def validate_email(self, field):
...
-        domain = extractor(resp.domain.lower()).top_domain_under_public_suffix
+        domain = tld_extractor(resp.domain.lower()).top_domain_under_public_suffix
+        self._email_domain = resp.domain.lower()
+        self._email_registrable = domain
+        # The reputation service answers on the ASCII form of the address, so
+        # hand it that rather than the raw submission: an IDN domain sent as
+        # Unicode is a different string to the vendor, and the errored lookup
+        # would silently fail open. ascii_email is None only when the local
+        # part itself is non-ASCII, which needs SMTPUTF8 anyway.
+        self._email_normalized = resp.ascii_email or resp.normalized
```
woodruffw: I believe this results in a surprising value for `_email_normalized` when both the mailbox and domain components contain non-ASCII characters: `ascii_email` is `None` in that case so we fall back to `resp.normalized`, which only uses the non-punycode form.

For example, `abc@münchen.de` correctly becomes `abc@xn--mnchen-3ya.de` but `é@münchen.de` stays as `é@münchen.de`.

I _think_ the right solution here is:


```suggestion
        self._email_normalized = f"{resp.local_part}@{resp.ascii_domain}"
```

(This is all caveated on UserCheck doing the right thing as well. Their docs aren't very clear on whether they handle Unicode mailboxes correctly at all.)

woodruffw (changes requested):
Two things; I think the email address bypass one is pretty minor but the disposable address versus domain one is probably worth fixing (since FWICT it's a genuine mixup caused by UserCheck's own API reusing the term "disposable".)

warehouse/accounts/forms.py
miketheman: Thanks - that's a good finding and covers my intent correctly. I'll be adding a condition for the disposable_provider as well - since we can get that from the /email endpoint and save the second call.

> The field is only present when the domain is disposable, is null when the operator has not been identified, and is returned on Pro plans. Gates exposes it as domain.disposable_provider.

--https://www.usercheck.com/docs/concepts/disposable

miketheman: Added to the scope

woodruffw (approved):
LGTM!

PR #14814, opened 2023-10-27
Buildkite OIDC support

sj26:
I'm sketching out support for [Buildkite OIDC tokens](https://buildkite.com/docs/agent/v3/cli-oidc) as an avenue for [trusted publishing](https://docs.pypi.org/trusted-publishers/) into PyPI. Still WIP, but sharing early for feedback :pray:

image

PR #20515, opened 2026-09-12
Add Buildkite as a trusted publishing provider

sj26:
## Why

Allow packages built on Buildkite to publish to PyPI without storing long-lived PyPI API tokens.

This revisits #14814 with a fresh implementation against Warehouse's current trusted-publisher infrastructure. The earlier PR demonstrated the integration but stalled around supporting multiple providers in the management UI. Those infrastructure changes have since landed, as noted in [the original discussion](https://github.com/pypi/warehouse/pull/14814#issuecomment-1887456636).

## What

Adds Buildkite publishers for existing projects and pending publishers for new projects, using tokens issued by `https://agent.buildkite.com`.

Users configure their Buildkite organization and pipeline slugs, with optional branch, tag, and step-key restrictions. Warehouse verifies the token and exchanges it for a short-lived, project-scoped upload credential through the existing trusted-publishing flow.

Immutable organization and pipeline IDs are pinned independently on first successful exchange when supplied. Once pinned, subsequent tokens must include the same values. Before pinning, tokens without those claims remain accepted. The management UI displays the pinned values, and the documentation explains this trust-on-first-use behavior.

The companion [`pypi-oidc` Buildkite plugin](https://github.com/sj26/pypi-oidc-buildkite-plugin) requests both immutable ID claims, preserves Buildkite's default subject, and supplies Twine credentials. A dedicated publishing step consumes artifacts from an earlier build step:

```yaml
steps:
  - label: ":python: Build distributions"
    key: build
    command: python -m build
    artifact_paths: "dist/*"

  - label: ":python: Publish to PyPI"
    key: publish
    depends_on: build
    plugin: sj26/pypi-oidc#v0.1.0
    command: |
      buildkite-agent artifact download "dist/*" .
      python -m twine upload dist/*
```

This assumes `build` and `twine` are installed on the agents. Set the publisher's step-key restriction to `publish` to match this example.

Buildkite support is disabled by default behind `disallow-buildkite-oidc`. Feedback is particularly welcome on whether the implicit ID-pinning policy meets PyPI's provider requirements, or whether immutable claims should be mandatory from the first exchange.

## Local demonstration

A real local Buildkite instance and agent built a wheel and source distribution, exchanged a signed job token through Warehouse's real OIDC service, uploaded with Twine, and installed/imported the package from the local index. The demo used a harness-only local issuer override and plugin `v0.1.0` with an unpublished fix to propagate token-exchange failures from its environment hook. No packages were uploaded to PyPI or TestPyPI.

The demo also confirmed that tokens missing an already-pinned organization ID are rejected.

Registration uses slugs, with optional restrictions and no manual immutable-ID fields:

Buildkite trusted publisher registration

After the real token exchange, both immutable IDs are recorded (this demo publisher leaves the optional restrictions unset):

Buildkite publisher after first-use ID pinning

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -705,6 +705,7 @@ def db_request(pyramid_request, db_session, tm):
 def _enable_all_oidc_providers(webtest):
     flags = (
         AdminFlagValue.DISALLOW_ACTIVESTATE_OIDC,
+        AdminFlagValue.DISALLOW_BUILDKITE_OIDC,
         AdminFlagValue.DISALLOW_GITLAB_OIDC,
         AdminFlagValue.DISALLOW_GITHUB_OIDC,
         AdminFlagValue.DISALLOW_GOOGLE_OIDC,
--- a/warehouse/admin/flags.py
+++ b/warehouse/admin/flags.py
@@ -20,6 +20,7 @@ class AdminFlagValue(enum.Enum):
     DISALLOW_GITLAB_OIDC = "disallow-gitlab-oidc"
     DISALLOW_GOOGLE_OIDC = "disallow-google-oidc"
     DISALLOW_ACTIVESTATE_OIDC = "disallow-activestate-oidc"
+    DISALLOW_BUILDKITE_OIDC = "disallow-buildkite-oidc"
     READ_ONLY = "read-only"
     AUTO_PROHIBIT_DISPOSABLE_DOMAINS = "auto-prohibit-disposable-domains"
 
--- a/warehouse/migrations/versions/a1c4b5219d5e_add_buildkite_oidc_flag.py
+++ b/warehouse/migrations/versions/a1c4b5219d5e_add_buildkite_oidc_flag.py
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Add disabled-by-default Buildkite OIDC feature flag
+
+Revision ID: a1c4b5219d5e
+Revises: 45301e80ef72
+Create Date: 2026-08-18 00:00:00.000000
+"""
+
+from alembic import op
+
+revision = "a1c4b5219d5e"
+down_revision = "45301e80ef72"
+
+
+def upgrade():
+    op.execute("""
+        INSERT INTO admin_flags(id, description, enabled, notify)
+        VALUES (
+            'disallow-buildkite-oidc',
+            'Disallow the Buildkite OIDC publisher',
+            TRUE,
+            FALSE
+        )
+        """)
+
+
+def downgrade():
+    op.execute("DELETE FROM admin_flags WHERE id = 'disallow-buildkite-oidc'")


warehouse/oidc/models/buildkite.py
```diff
@@ -0,0 +1,154 @@
...
+class BuildkitePublisherMixin:
+    """Common functionality for pending and concrete Buildkite publishers."""
+
+    organization_slug: Mapped[str] = mapped_column(String, nullable=False)
+    pipeline_slug: Mapped[str] = mapped_column(String, nullable=False)
+    buildkite_organization_id: Mapped[str] = mapped_column(String, nullable=False)
+    pipeline_id: Mapped[str] = mapped_column(String, nullable=False)
+    build_branch: Mapped[str] = mapped_column(String, nullable=False)
+    build_tag: Mapped[str] = mapped_column(String, nullable=False)
+    step_key: Mapped[str] = mapped_column(String, nullable=False)
```
di: Do we really want to make all of these required? Looking at https://buildkite.com/docs/apis/rest-api/builds, `step_key`.

At the very least, https://buildkite.com/docs/agent/cli/reference/oidc#claims says `step_key` can be `nil`, so this probably shouldn't be `nullable=False`.

Also looking at https://buildkite.com/docs/agent/cli/reference/oidc#claims, I'm confused by `build_tag`, this isn't listed in the optional tags (or in the example token) but the docs say "This claim is only included if the tag is set.", should this also be nullable?

warehouse/oidc/models/buildkite.py
```diff
@@ -0,0 +1,154 @@
...
+        if self.buildkite_organization_id:
+            details.append(("Organization ID", self.buildkite_organization_id))
+        if self.pipeline_id:
+            details.append(("Pipeline ID", self.pipeline_id))
+        if self.build_branch:
+            details.append(("Build branch", self.build_branch))
+        if self.build_tag:
+            details.append(("Build tag", self.build_tag))
+        if self.step_key:
+            details.append(("Step key", self.step_key))
```
di, line 225: These are all (currently) non-nullable so the conditionals here seem unnecessary.

warehouse/oidc/models/buildkite.py
```diff
@@ -0,0 +1,154 @@
...
+        return details
+
+
+class BuildkitePublisher(BuildkitePublisherMixin, OIDCPublisher):
+    __tablename__ = "buildkite_oidc_publishers"
+    __mapper_args__ = {"polymorphic_identity": "buildkite_oidc_publishers"}
+    __table_args__ = (
+        UniqueConstraint(
+            "organization_slug",
+            "pipeline_slug",
```
di, line 235: I think the constraint here should probably be by ID and not slug?

warehouse/oidc/models/buildkite.py
```diff
@@ -0,0 +1,154 @@
...
+    id: Mapped[UUID] = mapped_column(ForeignKey(OIDCPublisher.id), primary_key=True)
+
+
+class PendingBuildkitePublisher(BuildkitePublisherMixin, PendingOIDCPublisher):
+    __tablename__ = "pending_buildkite_oidc_publishers"
+    __mapper_args__ = {"polymorphic_identity": "pending_buildkite_oidc_publishers"}
+    __table_args__ = (  # type: ignore[assignment]
+        UniqueConstraint(
+            "organization_slug",
+            "pipeline_slug",
```
di, line 253: Ditto here about id vs slug

warehouse/oidc/models/buildkite.py
```diff
@@ -0,0 +1,154 @@
...
+BUILDKITE_OIDC_ISSUER_URL = "https://agent.buildkite.com"
+
+
+class BuildkitePublisherMixin:
+    """Common functionality for pending and concrete Buildkite publishers."""
+
+    organization_slug: Mapped[str] = mapped_column(String, nullable=False)
+    pipeline_slug: Mapped[str] = mapped_column(String, nullable=False)
+    buildkite_organization_id: Mapped[str] = mapped_column(String, nullable=False)
+    pipeline_id: Mapped[str] = mapped_column(String, nullable=False)
```
di: We should probably use the `organization_id` as well if it's available?

warehouse/oidc/models/buildkite.py
sj26: Yeah they the only required ones should be the org and pipeline slug - adjusted.

warehouse/oidc/models/buildkite.py
sj26: They've been made nullable :pray:

warehouse/oidc/models/buildkite.py
sj26: The trust-on-first-use model means we always have slugs but don't always have IDs so can't unconditionally constrain by ID. The IDs are checked later if they're already on the publisher record.

warehouse/oidc/models/buildkite.py
sj26: The pypi organization_id?

di: I'm assuming there is a similar way to transform an `organization_slug` into an `organization_id` to prevent resurrection attacks, is that the case? Are `organization_slug`s reusable?

sj26: `organization_slug` are reusable, yeah. We have no public API for turning `organization_slug` into `organization_id`, like with GitHub for instance; we try not to leak details of organization existence.

Buildkite OIDC tokens also don't include `organization_id` by default, which is why it's not a strict requirement.

The best I thought we can do here is TOFU (trust on first use) where you specify the slug and, if we see an id in a token which matches the slug, save the id beside the slug and require it for all future tokens.

We're discussing making these IDs always included by default, but the downside is much larger tokens which can cause issues in some consumers.

di: > Buildkite OIDC tokens also don't include organization_id by default, which is why it's not a strict requirement.

We can easily make it a strict requirement here by rejecting tokens without this claim.

> We have no public API for turning organization_slug into organization_id, like with GitHub for instance; we try not to leak details of organization existence.

Yes, if there's no API to do this at configuration time, the best we can do is capture it the first time we see the token. This means we will want to store an `organization_id` field on the `BuildkitePublisher` model.

Is the `pipeline_id` available to to convert from `pipeline_slug` at configuration time, or is that the same?

sj26: > We can easily make it a strict requirement here by rejecting tokens without this claim.

True. I'm planning to publish a buildkite plugin which automates the token request which asks for the required audience and can ask for the required claims too.

Better yet, `jti` is also not currently included by default but can be claimed, but we always add it when we see a rubygems audience. We can do the same for `jti`, `organization_id` and `pipeline_id` for pypi too I think - the security benefits are worth it.

It's easier for users to specify the slugs alone, imho, but then we'll always have the IDs to pin and can reject tokens without these claims.

> Is the `pipeline_id` available to to convert from `pipeline_slug` at configuration time, or is that the same?

It's the same. But, again, we can specifically include it for all OIDC tokens targeted at the pypi audience.

di: > True. I'm planning to publish a buildkite plugin which automates the token request which asks for the required audience and can ask for the required claims too.

FYI, per https://docs.pypi.org/trusted-publishers/using-a-publisher/ the audience for PyPI should always be `pypi` and `testpypi` for TestPyPI.

We will likely need to add support to https://github.com/di/id as well for fetching Buildkite credentials, this is what `twine`, etc. use.

sj26: Oh nice, I was using a buildkite plugin to prime `TWINE_PASSWORD` but it has expiry problems etc:

https://github.com/sj26/pypi-oidc-buildkite-plugin

`id` is a much nicer way to do it on demand, much like npm's token discovery gear.

sj26: Oh, `id` already supports buildkite:

https://github.com/di/id/pull/21

sj26: > Better yet, `jti` is also not currently included by default but can be claimed, but we always add it when we see a rubygems audience. We can do the same for `jti`, `organization_id` and `pipeline_id` for pypi too I think - the security benefits are worth it.

This is now shipped on our end - requesting an oidc token from buildkite for `pypi` or `testpypi` audiences automatically adds `jti`, `organization_id` and `pipeline_id` to the default claims.

sj26: I've already adjusted this PR to always require those claims, too.

sj26: We made an internal change to always populate tokens with `jti`, `organization_id` and `pipeline_id`, so there's no longer any special casing here.

sj26:
Hey folks, just checking in - I've pulled the feature flag into a separate PR (#20523) and I think resolved all of the feedback so far.

PR #20521, opened 2026-09-14
Fix migration for rate limit projects PR

JacobCoffee:
Fixes WAREHOUSE-PRODUCTION-3AB, WAREHOUSE-PRODUCTION-3AD, WAREHOUSE-PRODUCTION-3AC


JacobCoffee:
it migrated

PR #20528, opened 2026-09-15
Remove newlines in error status message

DarkaMaul:
Some validation errors (in [`file_upload`](https://github.com/pypi/warehouse/blob/98d02af987ad81389a79ee136877eb66b85de53b/warehouse/forklift/legacy.py#L1409-L1413)) include user-controlled text in the HTTP status reason phrase. 

This PR normalizes CR and newlines to spaces before constructing the status, preventing them from producing malformed response metadata.


woodruffw (commented):
This seems reasonable to me, but do you have an example of an error that does this at the moment?

(Note that this will also hopefully go away soon, since PEP 847 will change these responses.)

DarkaMaul:
> This seems reasonable to me, but do you have an example of an error that does this at the moment?

Yes - I have a local reproducer if needed, but for instance having a LICENSE file with `LICENSE\r\nX-New-Header: value` is enough.

The output is:
```
header:  X-New-Header: value does not exist in distribution file license_newline_repro_a30d049e-1.0.0.tar.gz at license_newline_repro_a30d049e-1.0.0/LICENSE, value
header:  Content-Type: text/html; charset=UTF-8
header:  Content-Length: 11
``

JacobCoffee (approved):

woodruffw (approved):

JacobCoffee:
@DarkaMaul im not sure if automerge will work until you rebase on latest main

tests/unit/forklift/test_utils.py
carineawa13-tech: 0d4f4340dc3cb54be1d59780d61542d5361ece19

carineawa13-tech: Somehow they embedded my phone with coding and I can't get it off it's like I'm in some kind of loop

PR #20530, opened 2026-09-15
Implement PEP 847

woodruffw:
WIP, not ready for review. 

This implements [PEP 847](https://peps.python.org/pep-0847/#specification), i.e. RFC 9457 style problem details. The core change here is <20 lines, everything else is tests 🙂


warehouse/views.py
```diff
@@ -75,6 +75,10 @@
 @view_config(context=HTTPException)
 @notfound_view_config(append_slash=HTTPMovedPermanently)
 @notfound_view_config(path_info=JSON_REGEX, append_slash=False)
+@forbidden_view_config(route_name="api.simple.index")
```
woodruffw: Routes HTTPForbidden when serving `/simple/`.

warehouse/views.py
```diff
@@ -75,6 +75,10 @@
...
+@forbidden_view_config(route_name="api.simple.detail")
```
woodruffw: Routes HTTPForbidden when serving `/simple/{project}/`.

warehouse/views.py
```diff
@@ -75,6 +75,10 @@
...
+@exception_view_config(PredicateMismatch, route_name="api.simple.index")
+@exception_view_config(PredicateMismatch, route_name="api.simple.detail")
```
woodruffw: I _think_ it would be safe to remove these two lines, since in practice we should never see a `PredicateMismatch` on either of these routes (given how trivial their predicates are). But they don't really hurt to include either.

PR #20538, opened 2026-09-16
Reconcile file storages by size instead of checksum

miketheman:
`reconcile_file_storages` runs every 15 minutes and has reconciled nothing since late August.
`release_files` still holds 7,493,241 rows with `cached = false`,
down from 7,725,023 when #19704 was opened, and that remaining drop came from the task's early runs rather than anything recent.

## The bug

The task compared `get_checksum` from both storages against `File.md5_digest`. On S3, `get_checksum` returned the object's ETag.
For any object S3 stored as a multipart upload, the ETag is the md5 of the
concatenated part md5s with a `-<partcount>` suffix, not the md5 of the content. `GenericS3BlobStorage.store` calls `bucket.upload_file` with no `TransferConfig`, so boto3's default 8MB `multipart_threshold` applies and every archived file above 8MB carries an ETag that cannot equal `md5_digest`. B2 compounds it by leaving `content_md5` null for anything it stores through the large file API.

The comparison fell through to the `else` branch, logged `Unable to reconcile stored file distribution`, appended to `errors`, and left `cached` false on every subsequent run.

## Evidence

Sentry `WAREHOUSE-PRODUCTION-2J9`, 720,444 occurrences since 2026-06-18.

| Date | Errors |
| --- | --- |
| 2026-07-30 | 7,776 |
| 2026-08-14 | 9,120 |
| 2026-08-21 | 9,600 |
| 2026-09-14 | 9,600 |

96 runs a day at the default batch size of 100 is 9,600 file checks. The count has sat at exactly 9,600 since 2026-08-21, so every file in every batch fails. The ramp up to that ceiling is the smaller files draining out of the pool until only multipart ones were left.

Every file I sampled from the error stream is over the threshold:

| File | Bytes |
| --- | --- |
| `cdk_lambda_subminute-2.0.145.tar.gz` | 10,738,701 |
| `deltachat-1.114.0-cp310-cp310-manylinux_2_17_x86_64.whl` | 11,537,474 |
| `maelzel-0.10.5.tar.gz` | 17,211,203 |
| `nd4j-1.0.0b3.tar.gz` | 283,381,048 |
| `tensorflow_gpu-1.13.1-cp35-cp35m-manylinux1_x86_64.whl` | 345,187,092 |

The METADATA siblings pass, because a `.metadata` file is a few KB and single part on both sides. That asymmetry shows up in the breadcrumbs of any event in the issue: `File METADATA is ok` next to `Unable to reconcile stored file distribution` for the same wheel.

## The fix

Compare sizes. `File.size` is recorded at upload, and both backends already return the object size from the HEAD request the task makes today.

- `IFileStorage.get_size` replaces `get_checksum`. S3 returns `Object.content_length`, B2 returns `FileVersion.size`, the local backend stats the file, GCS keeps raising `NotImplementedError`.
- The archive is the golden copy, but only where it agrees with `File.size`. An archive object that contradicts the database is a real problem and still errors.
- Anything the cache is missing, or holds at a different size, gets pulled from the archive.
- METADATA has no size in the database, so the archive is the only authority for it.
- `get_checksum` had no other callers and is removed.

This works against every existing object with no re-upload and no metadata backfill. The alternatives don't. `ChecksumMode=ENABLED` only returns a checksum for objects uploaded with an algorithm set, and the `meta` dict we attach at upload holds project, version, package type and python version, with no digest in it.

The trade is that size catches a missing or truncated object but not a corrupted one. If we want stronger verification going forward, `ChecksumAlgorithm="SHA256"` on the archive `store()` would give future objects a real retrievable checksum. That can land separately.

## Left for follow-ups

Things found while digging that want their own change:

- The batch query has no ordering, and rows that fail get re-picked on later runs. Less harmful once failures are rare, but a cluster of genuinely broken rows can still stall progress at the head of the scan.
- 100 files per 15 minutes against 7.5M rows is roughly 780 days. Draining the backlog needs its own sizing, separate from steady-state drift.
- ~~`_copy_file_to_cache` buffers the whole archive object in memory before writing it to the temp file.~~ Fixed in 796fd45b, taken from #20532.

## Testing

`make tests`: 6309 passed, 100% coverage. `make lint` clean.

New `TestReconcileFileStorages` covers both stores agreeing with the database, a cache that is missing or holds a differently sized copy, an archive that contradicts `File.size`, an archive missing the dist or the METADATA, and a file with no METADATA digest recorded.

The regression test is `test_both_stores_agree_with_db`. Against the old code it fails the way production does, with `unreconciled` on both `type:dist` and `type:metadata` for a file that is correct in both stores.

Refs: #19704


miketheman:
My intention is that once this is deployed and confirmed a clean run for a day or two is to increase `RECONCILE_FILE_STORAGES_BATCH_SIZE` to speed up based on already-instrumented timings - current batches of 100 take less than a minute, so could be increased up until a 10 minute duration - 10-12k per batch and burn through the backlog faster.

miketheman:
Added 796fd45b, which takes the streaming copy from #20532.

`_copy_file_to_cache` read the whole archive object into memory before writing it to the temp file, so a 345MB wheel cost the worker 345MB of RSS. It now copies through in 1MB chunks and closes the archive stream. That bug is independent of the checksum comparison and worth having either way.

The `sync_file_to_cache` tests were monkeypatching `tempfile.NamedTemporaryFile`, which is why nobody noticed. They use real streams and real temp files now.

JacobCoffee (approved):
like it better, can i have commit credit on https://github.com/pypi/warehouse/commit/796fd45bcf5869c8d046a2ddb77e57be690baf7a :D

PR #20544, opened 2026-09-17
Verifies macaroon signature before incrementing rate-limits 

DarkaMaul:
Verifies the macaroon signature instead of only looking up its UUID.
This prevents forged macaroons containing a known UUID from consuming another user’s project creation rate limit.

## More details

For a new project, `create_project` creates the project and increments the Redis backed rate limits. The user permission to create this project is only checked after. If the verification fails, the SQL transaction rolls back, but the Redis increment does not.

https://github.com/pypi/warehouse/blob/06c0ba62a95fbf17762f717790bbedbe9a623797/warehouse/forklift/legacy.py#L786-L807


woodruffw (approved):
LGTM, this is something I noticed when adding `verify_signature_only` (but considered low-priority to fix, since it requires macaroon disclosure + is only a griefing vector).

Perhaps we should add a note to `find_from_raw`'s docs saying that it doesn't perform any signature check? Alternatively, maybe we could remove `find_from_raw` entirely and keep just `verify_signature_only` 🙂

DarkaMaul:
> Alternatively, maybe we could remove `find_from_raw` entirely and keep just `verify_signature_only` 🙂

I think that's a good idea, because that's prevent an API misuse alltogether. I'll open a follow-up PR.

miketheman (changes requested):
Looks generally good. Some notes:

1. Probably want to now also update the security reporting endpoint to do the same, since the vector is similar - someone could commit a source code line with a forged UUID and achieve a simialr problem (not rate limit, but invalidation)
See https://github.com/pypi/warehouse/blob/4ee86c6e4a90b431003ec5c99e35dabdf5457683/warehouse/integrations/secrets/utils.py#L271-L278

2. There's no test I could find that confirms the rate limiting impact - a rename back to the original would not fail anything, reverting the behavior.
A test with DatabaseMacaroonService with a MacaroonFactory row plus a pymacaroons.Macaroon signed with the wrong key, asserting identity() is None is probably the way to go. Also, `test_identity_no_db_macaroon` now conflates "not in DB" with "bad signature".

3. Docs need an update for the function name change - https://github.com/pypi/warehouse/blob/main/docs/dev/security/security-policy-flow.md - if you want to generalize it a bit that's fine too.

miketheman (approved):
Thanks!

PR #20570, opened 2026-09-21
Align WebAuthn registration error handling

DarkaMaul:
Align WebAuthn registration error handling with the authentication path by returning a fixed `Invalid WebAuthn credential` message instead of the dependency exception details.

I've also updated the tests to reduce the usage of `pretend` as seen in other PRs (like 20559).


warehouse/manage/forms.py
```diff
@@ -268,8 +268,8 @@ def validate_credential(self, field):
                 rp_id=self.rp_id,
                 origin=self.origin,
             )
-        except webauthn.RegistrationRejectedError as e:
-            raise wtforms.validators.ValidationError(str(e))
+        except webauthn.RegistrationRejectedError:
+            raise wtforms.validators.ValidationError("Invalid WebAuthn credential")
```
miketheman, line 272: With this change, we lose the ability for the user to provide a more detailed exception if this happens, like we do here:
https://github.com/pypi/warehouse/blob/9d5a5ecd97ed2e632ec68f1452f16feabf703ccb/warehouse/accounts/services.py#L611

Is there a better way to expose the raised message back to the user?

warehouse/utils/webauthn.py
```diff
@@ -124,7 +124,7 @@ def verify_registration_response(response, challenge, *, rp_id, origin):
             require_user_verification=False,
         )
     except WebAuthnException as e:
-        raise RegistrationRejectedError(str(e))
+        raise RegistrationRejectedError("Invalid WebAuthn credential") from e
```
miketheman, line 127: similar note here - the raise was passing along the string from `verify_registration_response` - so does the caller now lose the ability to get those messages, and submit them to us via a support request? IO don't think they are captured or logged anywhere else.

miketheman (commented):
Nice work on converting the tests away from `pretend`.

Some questions inline - let me know if they need some more explanation.

PR #20572, opened 2026-09-21
Bump PyJWT to 2.14

JacobCoffee:


tests/common/constants.py
JacobCoffee, line 1: make the fake JWT signatures valid canonical base64url pyjwt 2.14 requires

miketheman (approved):
Nice

